PR #273, opened 2014-05-06
XWIKI-10309: Phishing Through URL Redirection

sdumitriu:
I'm not sure this is safe:
- is there any page that uses `xredirect` to redirect to a different wiki?
- is there any chance that `xredirect` uses a full URL that has a different host than what XWiki believes was requested, for example an IP vs. name, or `localhost` seen behind a proxy?


aubincleme:
We could also check the hostname against a user-defined whitelist, so that it's still possible to make that kind of customization. WDYT ?

PR #371, opened 2015-01-06
XWIKI-11486: Batch API to send large volumes of mail

bandoulyes:
* Create AbstractMailScriptService
* Minor fixes
* Update Admin UI

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/AbstractMailScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/AbstractMailScriptService.java
@@ -1,0 +1,149 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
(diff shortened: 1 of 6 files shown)


xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/AbstractMailScriptService.java
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * @version $Id $
+ * @since 6.4M3
+ */
+public abstract class AbstractMailScriptService implements ScriptService
+{
+    /**
+     * The key under which the last encountered error is stored in the current execution context.
+     */
+    static final String ERROR_KEY = "scriptservice.mailsender";
```
vmassol: not correct for the storage script service. You should have an abstract String getErrorKet() method

vmassol: Actually it depends if we want to have the same key and thus share the error between them. Do we want that $services.mailsender.getLastError() returns an error when resend() has failed? Probably not. WDYT?

PR #386, opened 2015-02-09
XWIKI-11774 : The scope icon for the group displayer is not centered in ...

vrachieru:
...flamingo

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/widgets/userpicker/userPicker.css
@@ -20,8 +20,9 @@ input.suggestGroups {
   width: auto !important;
 }
 .skin-flamingo .scope {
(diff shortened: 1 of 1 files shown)


mflorea:
Relative units are better in general, but in this particular case I think absolute units are better because the available space for the scope icon is defined in pixels:

* the input left border is set to 25px from userPicker.css, so 4px padding to the left and 5px to the right should work, considering that the icon is 16px in width
* the input height is set to 34px from the skin. I see that 8px padding top and 10px bottom look good, but unfortunately the input height is computed from LESS and can be changed from a color theme. 

An option is to set only the horizontal padding from userPicker.css (making sure it doesn't break Colibri) and put the vertical padding in the Flamingo skin, where we can compute it with LESS (see what variable is used for the input height).

@evalica WDYT?

evalica:
So after some discussions, we decided that is best if:
{code}
padding: 9px 4px; // Values calculated for a 16x16 icon, an input height of 34px and a border of 25px
{code}
Vertical: 25px - 16px = 9px / 2 = 4.5px
Horizontal: 34px - 16px = 18px / 2 = 9px

PR #389, opened 2015-03-13
Remove redondancy in the API

woshilapin:
Just remove a redundancy in the declaration of API paths (REST service `xwiki-platform-wiki` module)  since it's already defined in the [`xwiki-platform-wiki-rest-api`](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-rest/xwiki-platform-wiki-rest-api/src/main/java/org/xwiki/wiki/rest/WikiManagerREST.java#L35) module

### DIFF
--- a/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-rest/xwiki-platform-wiki-rest-default/src/main/java/org/xwiki/wiki/rest/internal/DefaultWikiManagerREST.java
+++ b/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-rest/xwiki-platform-wiki-rest-default/src/main/java/org/xwiki/wiki/rest/internal/DefaultWikiManagerREST.java
@@ -23,7 +23,6 @@ import java.net.URI;
 
 import javax.inject.Inject;
 import javax.ws.rs.POST;
-import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriBuilder;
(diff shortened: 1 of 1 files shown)


Enygma2002:
As far as I know, Java annotations are not inherited unless explicitly defined so. 

http://docs.oracle.com/javaee/6/api/javax/ws/rs/Path.html does not seem to be an inheritable annotation, so removing it in the implementation might cause the resource to not get registered. Have you tried it locally and it works?

woshilapin:
It's at least how it's done into the REST service of XWiki.  See the [API](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-api/src/main/java/org/xwiki/rest/resources/pages/PageResource.java#L37) and the [implementation](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/pages/PageResourceImpl.java#L38) of Page Resource in `xwiki-platform-rest` for example.  I made this PR only on this assumption (sorry).

Anyway, I tested it with the following `cURL` command

```
curl --verbose --data-binary "@wiki.xml" --header "Content-type: application/xml" --request POST --user Admin:admin "http://localhost:8080/xwiki/rest/wikimanager"
```

and the following file

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<wiki xmlns="http://www.xwiki.org">     
	<id>testwiki</id>
	<name>testwiki</name>
	<description>This is a new wiki</description>
	<owner>xwiki:XWiki.Admin</owner>
</wiki>
```

It works well, the subwiki is created.

Enygma2002:
You are right, it works. Maybe restlet reads the annotations from superclasses/interfaces too, don`t know.

PR #391, opened 2015-03-16
XWIKI-11897: Search not working through documents in Czech language

pabro:
Correction of language code and adding remove accents for Czech language.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
@@ -242,7 +242,7 @@
    <dynamicField name="*_ar"  type="text_ar"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_bg"  type="text_bg"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_ca"  type="text_ca"    indexed="true"  stored="true" multiValued="true" />
-   <dynamicField name="*_cz"  type="text_cz"    indexed="true"  stored="true" multiValued="true" />
+   <dynamicField name="*_cs"  type="text_cz"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_da"  type="text_da"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_de"  type="text_de"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_en"  type="text_en"    indexed="true"  stored="true" multiValued="true" />
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
@@ -242,7 +242,7 @@
    <dynamicField name="*_ar"  type="text_ar"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_bg"  type="text_bg"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_ca"  type="text_ca"    indexed="true"  stored="true" multiValued="true" />
-   <dynamicField name="*_cs"  type="text_cz"    indexed="true"  stored="true" multiValued="true" />
+   <dynamicField name="*_cs"  type="text_cs"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_da"  type="text_da"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_de"  type="text_de"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_en"  type="text_en"    indexed="true"  stored="true" multiValued="true" />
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
```diff
@@ -846,6 +846,8 @@
         <filter class="solr.LowerCaseFilterFactory"/>
         <filter class="solr.StopFilterFactory" ignoreCase="true" words="lang/stopwords_cz.txt" />
         <filter class="solr.CzechStemFilterFactory"/>       
+	<!-- Remove accents -->
+	<filter class="solr.ASCIIFoldingFilterFactory"/>
```
Enygma2002: You`re missing an extra indentation space.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
```diff
@@ -242,7 +242,7 @@
    <dynamicField name="*_ar"  type="text_ar"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_bg"  type="text_bg"    indexed="true"  stored="true" multiValued="true" />
    <dynamicField name="*_ca"  type="text_ca"    indexed="true"  stored="true" multiValued="true" />
-   <dynamicField name="*_cz"  type="text_cz"    indexed="true"  stored="true" multiValued="true" />
+   <dynamicField name="*_cs"  type="text_cz"    indexed="true"  stored="true" multiValued="true" />
```
Enygma2002: Maybe it would also be useful to rename the type "text_cz" into "text_cs", both in the usage (here) and in its definition, for correctness.

pabro:
I added indentation space and rename text_cz to text_cs in definition.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
mflorea: @Enygma2002 I'm not sure about this. The dynamic field is part of our customizations (so it's ok to change it) but the type definition comes from Solr and we didn't modify it so far. Changing it might make the upgrade a little more difficult (e.g. the person doing the upgrade might keep both ``text_cz`` and ``text_cs``). Moreover, as I commented on [XWIKI-11897](http://jira.xwiki.org/browse/XWIKI-11897) ``cz`` is the country code for Czech and maybe the Solr team convention is to use the country code on the field type definition (when there is only one language, I don't know). I prefer to keep only the customizations that are really needed.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
```diff
@@ -840,12 +840,14 @@
     </fieldType>
 
     <!-- Czech -->
-    <fieldType name="text_cz" class="solr.TextField" positionIncrementGap="100">
+    <fieldType name="text_cs" class="solr.TextField" positionIncrementGap="100">
       <analyzer> 
         <tokenizer class="solr.StandardTokenizerFactory"/>
         <filter class="solr.LowerCaseFilterFactory"/>
         <filter class="solr.StopFilterFactory" ignoreCase="true" words="lang/stopwords_cz.txt" />
```
tmortagne: Shouldn't this be modified to to be cleaner ?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
Enygma2002: My bet here is that a simple confusion was made on the Solr part between country code and language code, as you`ve mentioned.

I understand your suggestion, but my opinion is that we avoid confusions altogether (for anyone reading the config), and possibly add a comment (above the type definition) for whoever will be doing the upgrade (us, most likely) that, until Solr fixes the problem on their part, for XWiki, text_cs means Solr's text_cz :)

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/schema.xml
Enygma2002: Normally, I would agree as well, but since this means renaming the file too, it feels a bit wrong, as @mflorea pointed out before.

I`ve sent a mail on the solr list, asking about this confusion and hoping to get some clarifications: http://lucene.markmail.org/thread/fhgnyzeeaw3n2dul

pabro: Maybe the best will be let the type definition of text_cz untouched. It is based on the definition from Solr (Lucene) which use cz for Czech language and add a comment to the dynamic field that cs for XWiki means cz for Solr. If the Solr team fix it and change cz to cs, the person doing upgrade of Solr in XWiki will remove the comment and repair the dynamic field definition to include new changes from Solr. If the Solr team don't fix it, the type definition will remain same as in Solr examples which I think is cleaner.

Enygma2002: OK, I agree.

FTR, here are the upstream issues to follow for anyone interested:
https://issues.apache.org/jira/browse/LUCENE-6366
https://issues.apache.org/jira/browse/SOLR-7267

@pabro Please undo the type name change and leave it to "text_cz" as it is currently in Solr, as you`ve mentioned. As soon as it's done, please ping here so we can apply your PR.

pabro:
I undo the type name change and add a comment.

Enygma2002:
Thanks for your contribution and patience on the subject :), @pabro !

PR #395, opened 2015-03-27
XWIKI-11972 : Separate rating pages should have an author

vrachieru:
- save as minor edit
- provide a message when saving the rating document

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-api/src/main/java/org/xwiki/ratings/internal/SeparatePageRating.java
+++ b/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-api/src/main/java/org/xwiki/ratings/internal/SeparatePageRating.java
@@ -212,7 +212,7 @@ public class SeparatePageRating implements Rating
             document.setContentDirty(false);
             document.setCreatorReference(superadmin);
             document.setAuthorReference(superadmin);
-            context.getWiki().saveDocument(getDocument(), context);
+            context.getWiki().saveDocument(getDocument(), context.getMessageTool().get("rating.saveComment"), true, context);
         } catch (XWikiException e) {
             throw new RatingsException(e);
         }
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-api/src/main/java/org/xwiki/ratings/internal/SeparatePageRating.java
```diff
@@ -212,7 +212,7 @@ public void save() throws RatingsException
             document.setContentDirty(false);
             document.setCreatorReference(superadmin);
             document.setAuthorReference(superadmin);
-            context.getWiki().saveDocument(getDocument(), context);
+            context.getWiki().saveDocument(getDocument(), context.getMessageTool().get("rating.saveComment"), true, context);
```
Enygma2002: This should have been `localizationManager.getTranslationPlain("rating.saveComment")` with `@Inject private ContextualLocalizationManager localizationManager;`.

MessageTool is deprecated since 4.3M2.

See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-api/src/main/java/org/xwiki/localization/ContextualLocalizationManager.java

mflorea: I agree with @Enygma2002, but this is not a component so I guess you'll have to pass the localization manager in the constructor.

Enygma2002: Well, IMO, the real problem is a design one. It should not be the Rating's task to save/delete itself, but it should be the RatingsManager's instead. Probably a bigger refactoring will be needed here at some point.

PR #398, opened 2015-04-10
XWIKI-11922: New App from AWM should have Main.WebHome as parent page

gabrielasmeria:
* Since the default space is configurable, I used $services.model method to find out the homepage.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/AppWithinMinutesHomePage.java
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/AppWithinMinutesHomePage.java
@@ -31,11 +31,6 @@ import org.xwiki.test.ui.po.ViewPage;
  */
 public class AppWithinMinutesHomePage extends ViewPage
 {
-    /**
-     * The title of the AppWithinMinutes home page.
-     */
-    public static final String TITLE = "App Within Minutes";
-
     @FindBy(xpath = "//a[@class = 'button' and . = 'Create Application']")
     private WebElement createAppButton;
 
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
@@ -623,8 +623,8 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
           "title":$homePageTitle,
           "parent":$wikiHomePage
         }))
-        #set($queryString = $escapetool.url($queryString))
       #end
+      #set($queryString = $escapetool.url($queryString))
       $response.sendRedirect($xwiki.getURL($homePageRef, 'edit', $queryString));
     #else
       ## Redirect to view mode.
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
@@ -611,12 +611,19 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
         #set($appName = $stringtool.removeEnd($doc.title, ' Class'))
       #end
       #set($homePageRef = $services.model.createDocumentReference($doc.wiki, $appName, 'WebHome'))
-      #set($queryString = 'wizard=true')
+      #set($queryString = {"wizard" : true})
       #if(!$xwiki.exists($homePageRef))
         #set($homePageTitle = '$services.localization.render(''platform.appwithinminutes.appHomePageTitle'', [$doc.space])')
         #set($wikiHomePageRef = $services.model.resolveDocument('', 'default'))
         #set($wikiHomePage = $services.model.serialize($wikiHomePageRef, 'local'))
-        #set($queryString = "$queryString&amp;editor=inline&amp;template=AppWithinMinutes.LiveTableTemplate&amp;AppWithinMinutes.LiveTableClass_0_class=$escapetool.url($doc.fullName)&amp;title=$escapetool.url($homePageTitle)&amp;parent=$wikiHomePage")
+        #set($discard = $queryString.putAll({
+          "editor":"inline",
+          "template":"AppWithinMinutes.LiveTableTemplate",
+          "AppWithinMinutes.LiveTableClass_0_class":$doc.fullName,
+          "title":$homePageTitle,
+          "parent":$wikiHomePage
+        }))
+        #set($queryString = $escapetool.url($queryString))
       #end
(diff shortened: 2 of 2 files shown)


xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/AppWithinMinutesHomePage.java
```diff
@@ -31,11 +31,6 @@
  */
 public class AppWithinMinutesHomePage extends ViewPage
 {
-    /**
-     * The title of the AppWithinMinutes home page.
-     */
-    public static final String TITLE = "App Within Minutes";
```
sdumitriu: Why is this removed?

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
```diff
@@ -614,7 +614,9 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
       #set($queryString = 'wizard=true')
       #if(!$xwiki.exists($homePageRef))
         #set($homePageTitle = '$services.localization.render(''platform.appwithinminutes.appHomePageTitle'', [$doc.space])')
-        #set($queryString = "$queryString&amp;editor=inline&amp;template=AppWithinMinutes.LiveTableTemplate&amp;AppWithinMinutes.LiveTableClass_0_class=$escapetool.url($doc.fullName)&amp;title=$escapetool.url($homePageTitle)&amp;parent=AppWithinMinutes.WebHome")
+        #set($wikiHomePageRef = $services.model.resolveDocument('', 'default'))
```
Enygma2002: Question: Do we want the default home page or the current home page?

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/AppWithinMinutesHomePage.java
Enygma2002: I also see no point in removing this.

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
sdumitriu: What's the difference? Or do you mean "default" as in "the default default, Main.WebHome"?

Enygma2002: Default = `$services.model.resolveDocument('', 'default')`, i.e. "the default homepage of a wiki" / "the default homepage of the default space of a wiki" / etc., according to the model's configured default entities.

Current = `$services.wiki.currentWikiDescriptor.mainPageReference`, i.e. the currently configured homepage of the current wiki. See http://extensions.xwiki.org/xwiki/bin/view/Extension/Wiki+Application#HConfigureWiki

Using the default might not make sense if the homepage is configured to be something else in the current instance.

At the same time, using the current homepage might not make sense when exporting the app and installing it on a vanilla environment.

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/AppWithinMinutesHomePage.java
gabrielasmeria: That variable was used just in [these](https://github.com/gabrielasmeria/xwiki-enterprise/commit/339a00d84f3753b956dcd9caa13abbcf7113aa73) 2 functional tests to go to AWM's homepage, by clicking "App Within Minutes" from the breadcrumb.
And in order to have the current wiki homepage as a parent for the newly created app, I deleted "App Within Minutes" element from the breadcrumb (which was the parent for the newly created app). 
So, instead I used gotoPage() method to access AWM's homepage, as you can see in the link above.

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
mflorea: @Enygma2002 what @gabrielasmeria did was to follow the behaviour of the [logo link](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/src/main/resources/flamingo/companylogo.vm#L29) from the top menu bar. I'm against storing the configured home page in the parent field because the parent field won't be updated when the home page configuration changes. And saving (possibly) lots of pages when changing the home page configuration is not a good idea. Ideally we should be able to set the value of the parent field to a "token" that represents the configured home page whatever that may be. But since we don't have such a "token" the best option is IMO to store the default home page as parent. An administrator can redirect the default home page to the configured home page. He can also display the title of the configured home page in the title of the default home page using Velocity so that the breadcrumb looks nice.

Enygma2002: @mflorea the workaround sounds decent.

Note that the logo should/will use the currently configured homepage though, instead of hardcoding it.

mflorea: @Enygma2002 I agree about the logo (because the logo link is dynamic, computed on each page load).

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
```diff
@@ -614,7 +614,9 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
...
+        #set($wikiHomePage = $services.model.serialize($wikiHomePageRef, 'local'))
+        #set($queryString = "$queryString&amp;editor=inline&amp;template=AppWithinMinutes.LiveTableTemplate&amp;AppWithinMinutes.LiveTableClass_0_class=$escapetool.url($doc.fullName)&amp;title=$escapetool.url($homePageTitle)&amp;parent=$wikiHomePage")
```
mflorea: The (default) ``$wikiHomePage`` is going to be URL-safe most of the time but it's still better to URL-encode it to avoid any problems. You can quickly wrap it with ``$escapetool.url`` but a nicer fix would be to call ``$escapetool.url`` only once 2 lines below when the edit URL is created, passing a map that you have created and filled before. This will also help you better format this line (so that it doesn't exceed the 120 chars limit), making it easier to read on GitHub for instance. Ping me if you have questions.

gabrielasmeria: Fixed it. Thanks!

PR #399, opened 2015-04-13
XWIKI-6118: add boolean list to filter livetable

Pbast:
Notice: I didn't found a method to get a porperty type (here Boolean)


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3077,6 +3077,8 @@ platform.livetable.paginationResultsSingle=Result <span class="currentResultsNo"
 platform.livetable.paginationResultsMany=Results <span class="currentResultsNo">{0} - {1}</span> of <span class="totalResultsNo">{2}</span>
 platform.livetable.paginationResults=Results
 platform.livetable.paginationResultsOf=out of
+platform.livetable.false=False
+platform.livetable.true=True
```
acotiuga: From the issue's page's comments, I understand that the values used in the select should be `Yes/No`.

Pbast: yes of course (old PR)
Anyway, now I found the way to identify property type (boolean) but didn't found time to add it to avoid new 'listboolean' option...

Pbast: Finally, I think my solution is lighter instead to  retrieve property type.

Thxs

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3077,6 +3077,8 @@ platform.livetable.paginationResultsSingle=Result <span class="currentResultsNo"
 platform.livetable.paginationResultsMany=Results <span class="currentResultsNo">{0} - {1}</span> of <span class="totalResultsNo">{2}</span>
 platform.livetable.paginationResults=Results
 platform.livetable.paginationResultsOf=out of
+platform.livetable.false=No
+platform.livetable.true=Yes
```
vmassol: we already have generic keys for this which might be better to use. For example:

    no_0=No
    yesno_1=Yes

vmassol:
Apart from my comment about the translation property to use for Yes/No, I think it looks good! I can apply it if you can make this change :) Thanks Pascal

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
vmassol: Hmm I see we also have:

    yes=Yes
    no=No

I don't know which one is best but this one might be better than the other one...

FTR I see other translations for yes/no... such as:

    core.widgets.confirmationBox.button.yes=Yes
    core.widgets.confirmationBox.button.no=No
    xe.officeimporter.openoffice.yes=Yes
    xe.officeimporter.openoffice.no=No

So the question is open: should we standardize on "yes"/"no" translation keys or should we introduce specific keys as Pascal has done above "platform.livetable.false"/"platform.livetable.true"?

I'm still hesitating on my side...

tmortagne:
IMO there is no point in translating 10 times "yes" and "no" as long as it's really about "yes" and "no".

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
Pbast: Yes it was my first thinks about it. a new key allow to change livetable value ... but yes/no is generic too :-/.

Pbast:
Done! :-)

Pbast:
Hi, could you merge my PR or tell me what is wrong?
(to update livetable doc and release note in one shot :-) )
Thxs

evalica:
First of the PR has conflicts :) so it's not that easy to be merged.

Pbast:
Ok thxs for ur answer, maybe I can refresh my git local and check/resolve conflicts (probably because my PR is one year old).

Pbast:
Ok thxs for ur answer, maybe I can refresh my git local and check/resolve conflicts (probably because my PR is one year old).

Ho i make a mistake with my tentative to merge my PR. I will create a new one I think.

PR #402, opened 2015-05-05
XWIKI-10118: Allow multiple column layouts on single page

DarkKnightCZ:
Sets different class to every container which represents column numbers. By default limit is 50 columns (can be increased/decreased by explicitly specifying number of max columns as ssfx parameter).

Fixed by Tieto Czech s.r.o., Lukas Raska, lukas.raska@tieto.com


gdelhumeau:
Looks very good. I'm going to try it locally.

gdelhumeau:
Merged. Sorry for the delay.

PR #404, opened 2015-05-18
XWIKI-12095: Livetable is always empty when using Documents macro with the Parent parameter

msmeria:


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DocumentsMacro.xml
```diff
@@ -183,11 +183,12 @@ Generates:
   'rowCount' : $xcontext.macro.params.count,
   'description' : 'This table lists documents found on this wiki based on passed criteria. The columns can be sorted and some can be filtered.'
 })
-#if ("$!xcontext.macro.params.space" != "") 
-  #set ($extraParams = "&amp;space=$escapetool.url($xcontext.macro.params.space)")
-#end
-#if ("$!xcontext.macro.params.parent" != "") 
-  #set ($extraParams = "${extraParams}&amp;parent=$escapetool.url($xcontext.macro.params.parent)")
+#if ("$!xcontext.macro.params.parent" != "" && "$!xcontext.macro.params.space" != "")
```
sdumitriu: Unescaped `&` is invalid.

msmeria: Fixed in df91fbc6f9e81167868dae0b68d0214488982616.
Thanks, @sdumitriu!

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DocumentsMacro.xml
```diff
@@ -183,11 +183,12 @@ Generates:
   'rowCount' : $xcontext.macro.params.count,
   'description' : 'This table lists documents found on this wiki based on passed criteria. The columns can be sorted and some can be filtered.'
 })
-#if ("$!xcontext.macro.params.space" != "") 
-  #set ($extraParams = "&amp;space=$escapetool.url($xcontext.macro.params.space)")
-#end
-#if ("$!xcontext.macro.params.parent" != "") 
-  #set ($extraParams = "${extraParams}&amp;parent=$escapetool.url($xcontext.macro.params.parent)")
```
Enygma2002: Actually, the fix is to just add a missing `!` to `${extraParams}`, resulting in `$!{extraParams}`. This way, if *only* the parent is specified and `${extraParams}` is null, it will append an empty string (` `) instead of `${extraParams}`.

PR #411, opened 2015-06-29
updated translation since you can set a local user as an owner on a subwiki

msmeria:


evalica:
We usually have issues associated with each of our commits. Otherwise, we use `[misc]`, `[cleanup]`, etc. related tags.

PR #412, opened 2015-07-01
XWIKI-10804: 'Copy' menu entry is displayed even if the users can't ' Edit' and 'Delete'

msmeria:


xwiki-platform-core/xwiki-platform-colibri/src/main/resources/colibri/menuview.vm
```diff
@@ -297,7 +297,7 @@
 ## Page
 ##
 #set ($canView = !$doc.isNew() && $xwiki.hasAccessLevel('view'))
-#set ($canCopy = $canView && $hasCreatePage)
+#set ($canCopy = $canView && $hasCreatePage && $xwiki.hasAccessLevel('edit'))
```
mflorea: ``$hasCreatePage`` implies that the user can create a page (and thus can edit it). The problem is not here. I would investigate why ``$hasCreatePage`` is true when the user doesn't have edit right.

PR #414, opened 2015-07-07
XWIKI-12279: Add page to ActiveInstalls which allows getting the numb…

cjdelisle:
…er of current active installs with a given extension

http://jira.xwiki.org/browse/XWIKI-12279


vmassol:
Note that before I can apply this PR we need to update the functional test to prove it works, as otherwise we'll reduce the quality of this app. I'll try to add one but if you can do it then all the better :)

vmassol:
I'm doing it...

vmassol:
Applied with the following changes:
* Fixed wrong parent
* Added functional test
* Improved javadoc to follow conventions
* Hide technical page
* Change macro name to follow conventions
* Allow the user to control the variable in which to store the result
* Fix indentations

PR #424, opened 2015-08-26
XWIKI-10850: The "Create new group" field is not consistent with the …

gabrielasmeria:
…"Create group" and "Cancel" buttons

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/lightbox/lightbox.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/lightbox/lightbox.css
@@ -67,7 +67,7 @@
 #lb {
   left: 0;
   margin-left: 0;
-  width: auto;
+  width: 100%;
 }
 
 @media (min-width: 768px) { /* Should be: @screen-sm-min */
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/addnewgroup.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/addnewgroup.vm
@@ -8,9 +8,9 @@
 <input type="text" name="newgroup" value="" id="newgroupi">
 </dd>
 </dl>
-<p>
+<div class="buttons">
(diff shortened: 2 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/lightbox/lightbox.css
```diff
@@ -67,7 +67,7 @@
 #lb {
   left: 0;
   margin-left: 0;
-  width: auto;
+  width: 100%;
```
evalica: Why is this needed? The value was reset in order to fix XWIKI-10734. 
Would be nice to see how it affects the responsiveness of Modals.

Also I think width:auto rule should have been removed, since that's the default.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/addnewgroup.vm
```diff
@@ -1,10 +1,16 @@
...
+<form method="post" onsubmit="return false;" class="xform">
+<dl>
+<dt>
+<label for="newgroupi">$services.localization.render('rightsmanager.createnewgroup')</label>
+</dt>
+<dd>
+<input type="text" name="newgroup" value="" id="newgroupi">
+</dd>
+</dl>
+<p>
```
evalica: "div.buttons" could have been used. 

To be honest, I'm not sure if we should use the: 
- vertical form standard;
- bootstrap standard; 
- or both (.button + .btn classes),
since we still haven't agreed on the standard. 

From a consistency point of view and until we reach a conclusion I prefer the approach you took (the vertical form standard), especially since we extend on the old classes.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/lightbox/lightbox.css
gabrielasmeria: I had to change it to 100%, because if I would've left it auto, the modal was occupying just 50-60% of the screen. I tested the modals from the screenshots from XWIKI-10734 and they're all working.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/addnewgroup.vm
gabrielasmeria: I've added div.buttons. Thanks!

PR #425, opened 2015-09-02
XWIKI-12462: Hard to tell which label goes with which field when view…

gabrielasmeria:
…ing or editing template providers

* Fixed the indentation
* Used xwiki syntax when possible
* Added a div with xform class in order to use vertical forms properly
* Removed empty lines
* Added <label> tag for each <dt> tag
* Rearranged some code, in order to be more readable

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
@@ -42,56 +42,52 @@
   <content>{{velocity}}
 #set ($discard = $xwiki.jsx.use('XWiki.TemplateProviderSheet'))
 #set ($providerObj = $doc.getObject('XWiki.TemplateProviderClass'))
-#if ($editor == 'inline')
-  ; {{translation key="xe.templateprovider.name"/}}
-  : {{html}}&lt;input type="text" value="$!{doc.getTitle()}" name="title" /&gt;{{/html}}
-#end
-; {{translation key="xe.templateprovider.templatename"/}}
-#if ($editor == 'inline')
+{{html wiki="true"}}
+(% class="xform" %)
+(((
+  #if ($editor == 'inline')
+    ; &lt;label&gt;{{translation key="xe.templateprovider.name"/}}&lt;/label&gt;
+    : {{html}}&lt;input type="text" value="$!{doc.getTitle()}" name="title" /&gt;{{/html}}
+  #end
+  ; &lt;label&gt;{{translation key="xe.templateprovider.templatename"/}}&lt;/label&gt;
+  #if ($editor == 'inline')
+    (% class="xHint" %){{translation key="xe.templateprovider.templatename.info"/}}
+    : $doc.display('name')
+  #else
+    #set($providerName = $providerObj.getProperty('name').value)
+    : $services.localization.render($providerName)
+  #end
+  ; &lt;label&gt;{{translation key="xe.templateprovider.template"/}}&lt;/label&gt;
+  #if ($editor == 'inline')
+    : $doc.display('template')
+  #else
+    #set($templateDocName = $providerObj.getProperty('template').value)
+    : [[$templateDocName&gt;&gt;$templateDocName]]
+  #end
+  ; &lt;label&gt;{{translation key="xe.templateprovider.templatetype"/}}&lt;/label&gt;
+  #if ($editor == 'inline')
+    (% class="xHint" %){{translation key="xe.templateprovider.templatetype.info"/}}
+  #end
+  : $doc.display('type')
+  (% id="enabled-spaces" %)
+  (((
+    ; &lt;label&gt;{{translation key="xe.templateprovider.spaces"/}}&lt;/label&gt;
+    #if ($editor == 'inline')
+      (% class="xHint" %){{translation key="xe.templateprovider.spaces.info"/}}
+    #end
+    : $doc.display('spaces')
+  )))
+  ; &lt;label&gt;{{translation key="xe.templateprovider.action"/}}&lt;/label&gt;
+  #if ($editor == 'inline')
+    (% class="xHint" %){{translation key="xe.templateprovider.action.info"/}}
+  #end
+  : $doc.display('action')
 
-  {{translation key="xe.templateprovider.templatename.info"/}}
-  : $doc.display('name')
-#else
-  #set($providerName = $providerObj.getProperty('name').value)
-  : $services.localization.render($providerName)
-#end
-; {{translation key="xe.templateprovider.template"/}}
-#if ($editor == 'inline')
-  : $doc.display('template')
-#else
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
```diff
@@ -42,56 +42,52 @@
   <content>{{velocity}}
 #set ($discard = $xwiki.jsx.use('XWiki.TemplateProviderSheet'))
 #set ($providerObj = $doc.getObject('XWiki.TemplateProviderClass'))
-#if ($editor == 'inline')
-  ; {{translation key="xe.templateprovider.name"/}}
-  : {{html}}&lt;input type="text" value="$!{doc.getTitle()}" name="title" /&gt;{{/html}}
-#end
-; {{translation key="xe.templateprovider.templatename"/}}
-#if ($editor == 'inline')
+{{html wiki="true"}}
```
evalica: why do you need the 'wiki=true' parameter?

Enygma2002: It`s not a good practice to wrap everything in a `html` macro. This needs to be removed.

We want to limit its usage as much as possible and that was the whole point this page is mainly using wiki syntax instead of raw html.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
```diff
@@ -42,56 +42,52 @@
...
+(% class="xform" %)
+(((
+  #if ($editor == 'inline')
+    ; &lt;label&gt;{{translation key="xe.templateprovider.name"/}}&lt;/label&gt;
```
Enygma2002: There are 2 issues here:

1) Since you will remove the HTML macro above, you need to use a HTML macro locally, since this is the place you actually need it. You actually have an example just one line under, why not follow it and do the same for the `label` form element?
1.1) At this point, you will not be able to use the `{{translation}}` macro, since you're inside a (local) `html` macro. You do not want to set `wiki=true` on this local html macro because you risk receiving bad content from the translation (wiki syntax or something that looks like it that will result in it being interpreted and the label will look bad). The solution is to use the `$services.localization` service inside your `html` macro to render the translation.

2) All labels need to specify a `for` attribute with the value of the input ID they are associated to. This is both an usability and an accessibility best practice that we enforce to comply as much as possible with the WCAG rules. If the input does not have an ID, assign one. If the ID is assigned automatically from a `doc.display(...)` call, use the automatically assigned ID (since it always follows a pattern, usually `className_propertyName`).

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
```diff
@@ -42,56 +42,52 @@
...
-: $doc.display('type')
-(% id="enabled-spaces" %)(((
-; {{translation key="xe.templateprovider.spaces"/}}
-#if ($editor == 'inline')
-
-  {{translation key="xe.templateprovider.spaces.info"/}}
-#end
-: $doc.display('spaces')
+  #if ($xcontext.action == 'view')
+    (% class="buttonwrapper" %)[[{{translation key="xe.templateprovider.backtoadmin"/}}&gt;&gt;$xwiki.getDocument('XWiki', 'XWikiPreferences').getExternalURL('admin', 'section=Templates')]]
```
Enygma2002: This is a good opportunity to also fix the link problem here.

There is no reason to call `getExternalURL` here instead of `getURL`, since that will result in a an external link "marker" also being applied next to the link (next to the button in this case) and we don`t want that.

The reason this solution was chosen in the past is probably because either there was no [path:](http://platform.xwiki.org/xwiki/bin/view/Main/XWikiSyntax#HXWikiSyntax2.1LinkSpecification) option available to the link syntax (pre `xwiki/2.1`) or the writer did not know about it.

So you could also change this to `>>path:$xwiki...getURL(...)` to fix the bad display. (make sure the sheet document is set to 2.1 syntax for this to work).

Enygma2002:
Glad to see this handled, since I`ve noticed it but did not make the time for it! :thumbsup:

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
mflorea: @Enygma2002 I don't think it's a good idea to use wiki syntax to create UIs or forms because it wasn't meant for that. I did it in the past but I stopped. Wiki syntax is meant for content, when you don't care that much about the final HTML structure or the applied styles. On the contrary, when you write UIs/forms you want 100% control of the final HTML structure in order to enhance it with JavaScript and style it with CSS. You don't want automatically generated paragraphs or spans, and you want to control the attributes of any element (including list items).

vmassol: I agree with both of you :)

Note that having HTML macro means that some Renderers in the future may not work correctly (since HTML Macro generates Raw Blocks). That's unless we decide that all Renderers have to support Raw Blocks. This is a bit theoretical at this point since our main Renderer is the (X)HTML Renderer and it supports HTML obviously ;) If we were to write a PDF Renderer based on iText for example (I started one in a branch in the Rendering repo a long time ago), the HTML would need some special handling to convert it to HTML (as we currently do our current PDF export,- which doesn't go through XWiki Rendering).

Just to say that till now our strategy was to reduce to the maximum the usage of the HTML Macro and even to be able to not use it at all by introducing some new Macros, such as a Form Macro for example for designing forms (PDF has Form support too for example). Note that this would probably need Rendering v3, to allow adding new Events.

Now, I agree that HTML is better for UI than wiki syntax is so I think we need to discuss this on the list and set up a strategy as to when it's ok to use the HTML Macro and when it's not and the consequences of that. One idea could be that it's ok that pages with UI may not have proper Rendering for some Renderers (ie ignore the Raw Blocks).

gabrielasmeria: I'll wait for a decision to be taken regarding this and I'll update the PR afterwards.

Enygma2002: I don`t think waiting will do us any good.

Since you`ve started using the HTML macro, I suggest you use it all the way (i.e. use dd/dt, etc.) so that we can progress on this part faster, since the template provider UI needs some other work.

PR #429, opened 2015-09-08
XWIKI-10903: "Choose file" from "Upload a new package" doesn't fit with its background on mobile

gabrielasmeria:
Added:
- div with "row" class surrounding the div with id="import"
- div with class="col-xs-12 col-sm-6 col-md-6" surrounding div with id="packagelist"
- div with class="col-xs-12 col-sm-6 col-md-6" surrounding div with id="packagecontainer"

Also corrected indentation

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/importer/import.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/importer/import.css
@@ -22,11 +22,6 @@ div.loading {
 }
 
 /* Package List  styles */
-#packagelist {
-  float: left;
-  width: 51%;
-}
-
 #packagelist fieldset {
   background-color: $theme.backgroundSecondaryColor;
   padding: 5px;
@@ -71,9 +66,7 @@ div.loading {
 
 /* Package Content styles */
 #packagecontainer {
-  float: left;
   margin-left: 2%;
-  width: 46%;
 }
 
 #packageDescription {
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/importinline.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/importinline.vm
@@ -35,171 +35,177 @@
   $xwiki.ssfx.use('js/xwiki/importer/import.css', true)##
   $xwiki.jsfx.use('uicomponents/widgets/upload.js', {'forceSkinAction': true, 'language': ${xcontext.language}})##
   $xwiki.ssfx.use('uicomponents/widgets/upload.css', true)##
-  <div id="import">
-    <div id="packagelist">
-      <div class="legend">$services.localization.render('core.importer.uploadPackage')</div>
-      ## Let the user upload XAR files.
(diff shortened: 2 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/importer/import.css
```diff
@@ -71,9 +66,7 @@ div.loading {
 
 /* Package Content styles */
 #packagecontainer {
-  float: left;
   margin-left: 2%;
```
evalica: if we rely on the BS grid, than we should remove all the properties that are changing margins, width, etc.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/importinline.vm
```diff
@@ -35,171 +35,177 @@
...
-                <span class="author">$package.packageAuthor</span>
-              </div>
-              #end
-              #if("$!package.packageLicense" != '')
-              <div>
-                <span class="label">$services.localization.render('core.importer.package.licence')</span>
-                <span class="licence">$package.packageLicense</span>
-              </div>
-              #end
+  <div class="row">
```
evalica: Instead of adding a new container, a better solution would be add the class to div#import.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/importinline.vm
```diff
@@ -35,171 +35,177 @@
...
+    <div id="import">
+      <div class="col-xs-12 col-sm-6 col-md-6">
```
evalica: same here, add the classes to div#packagelist

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/importinline.vm
```diff
@@ -35,171 +35,177 @@
...
-                        <ul class="xlist pages">
-                        #foreach($node in $spaceNode.children)
+          </form>
+          <div class="legend">$services.localization.render('core.importer.availablePackages')</div>
+          <div id="packagelistcontainer">
+            #template('packagelist.vm')
+          </div> ## packagelistcontainer
+        </div> ## packagelist
+      </div>
+      <div class="col-xs-12 col-sm-6 col-md-6">
```
evalica: same

gdelhumeau:
Note that using the bootstrap classes makes this template (`importinline.vm`) specific to Flamingo.
2 solutions:

1. put the current file in `colibri` and put your new file in `flamingo`. Remove it from `templates` since it is not generic anymore.
2. do not use bootstrap classes but instead add some xwiki classes that you can overwrite in the LESS code of flamingo.

I prefer the solution 1, because Colibri is deprecated and already not usable on mobile so we don't care to fix it for this skin.

gabrielasmeria:
Thanks for the comments. I will repair everything in a cleaner PR.

evalica:
Sol 3: 
- move the `#packagecontainer`, `#packagelist` rules in `colibri.css` with a comment that they are for import
- add the `.col-*` classes in `importinline.vm`. The `col-` styling shouldn't affect `Colibri`

PR #432, opened 2015-09-09
XWIKI-12530: Remove space and document mentions from Tags

gabrielasmeria:


xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/SpaceDashboardTemplate.xml
```diff
@@ -210,7 +210,7 @@
       <position>1,2</position>
     </property>
     <property>
-      <title>$services.localization.render('platform.dashboard.wiki.tagcloud')</title>
+      <title>$services.localization.render('platform.dashboard.wiki.tagsgadget', [$doc.title])</title>
```
evalica: Use $doc.plainTitle instead of $doc.title (for documents that use translation keys in their titles) from mflorea

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3772,6 +3772,7 @@ platform.dashboard.user.displayOnMainPage=Display my dashboard on the wiki home
 platform.dashboard.wiki=Dashboard
 platform.dashboard.wiki.pages=Pages
 platform.dashboard.wiki.tagcloud=Tags
+platform.dashboard.wiki.tagsgadget=Tags for {0}
```
evalica: To be honest I don't like the key. 

Instead of 'platform.dashboard.wiki.tagsgadget' I would put something like: 
platform.dashboard.wiki.tagcloud.space.title

gabrielasmeria: I didn't want to use the word 'space' in a translation key (actually that was my first idea, too). But I'll change it.

gabrielasmeria:
Fixed it in the last commit. Thanks!

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
evalica: Wiki was wrong since the translations were grouped by wiki and space.

PR #440, opened 2015-09-24
XWIKI-12614: Headings wrapped in ordered lists during import with Lib…

AndreasJonsson:
…reOffice 4.3.3

This patch adds an html cleaning filter for the office importer to remove ordered lists produced around headings.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
+++ b/xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
@@ -1,0 +1,152 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.officeimporter.internal.filter;
+
+import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.xml.html.filter.AbstractHTMLFilter;
+
+
+/**
+ * Sometimes the heading becomes wrapped in an ordered list structure such as this during import:
+ * 1.
+ * 
+ * 11. 
+ * 111. 
+ * 1111. 
+ * 11111. (((
+ * ===== Heading text =====
+ * )))
+ *
+ * This filter removes these structures.
+ *
+ * @version $Id$
+ * @since 6.4.6
+ */
+@Component
+@Named("officeimporter/heading-in-ordered-list")
+@Singleton
+public class HeadingInOrderedListFilter extends AbstractHTMLFilter
+{
+
+    @Override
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
```diff
@@ -0,0 +1,152 @@
...
+ * 1.
+ * 
+ * 11. 
+ * 111. 
+ * 1111. 
+ * 11111. (((
+ * ===== Heading text =====
+ * )))
+ *
+ * This filter removes these structures.
```
mflorea: Would be good to indicate what is the result after the filter is applied, otherwise it's not very clear what you mean by "these structures".

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @version $Id$
+ * @since 6.4.6
```
mflorea: You made the PR for master branch, so it should be 7.3M1.

vmassol:
Hey, hi Andreas! Nice to see you again ;) Thanks for the patch, that's great!

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
AndreasJonsson: Yes, of course.  But I'd appreciate if this would be cherry picked for 6.4.6 in case there will be such a release.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
AndreasJonsson: Sorry for being unclear!  I have updated the comment.

AndreasJonsson:
Hi Vincent!  Its only a small fix, but important for a project I'm working on.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/filter/HeadingInOrderedListFilter.java
```diff
@@ -0,0 +1,166 @@
...
+ * The algorithm is as follows:
+ *
+ * For all heading tags,
+ *
+ *   Find an ancestor chain consisting of ol or li tags, the topmost being an ol.
+ *
+ *   In the DOM tree rooted at the topmost ol, remove all ol and li tags of ol lists while keeping the contents of the
+ *   tags.
+ *
+ *   Replace the topmost ol with the preserved nodes in order in the original DOM.
```
mflorea: I'm wondering if this algorithm doesn't filter too much, e.g. headings that are supposed to be inside a list item. What would be the result of filtering his:

    <ol>
      <li>
        <h4>First Category Title</h4>
        <p>First category description here ...</p>
        <ol>
          <li>
            <h5>First Sub-Category Title</h4>
            <p>First sub-category description here ...</p>
          </li>
        </ol>
      </li>
      <li>
        <h4>Second Category Title</h4>
        Second category description here ...
      </li>
    </ol>

List items can accept both block and in-line content, so having headings inside list items is syntactically correct. Semantically, I don't know. I'm not sure it's easy to know when the heading was intended to be inside the list item and when it wasn't intended. We can probably clean the ``li`` and ``ol`` when the list has only one item and the heading is the only content inside the list item, but even in this case I'm not 100% sure.

AndreasJonsson: I don't think it is possible to distingiush headings intentionally inserted inside numbered lists in the general case.  Note that sometimes the entire subsequent section will also be inserted inside the list item corresponding to the section heading.

A proper fix should maybe be made to jodconverter but it seems to be more or less abandoned.

Anyway, the bug produces terrible quality imports and it completely breaks document splitting.

mflorea: Would be nice if the list of office importer cleaning filters were configurable so that an advanced user can disable some filters if they produce bad results for a particular input document. As a start we could configure the list of filters in ``xwiki.properties`` but ideally it should be possible to select/deselect the filters from the office importer UI, using checkboxes.

This would allow the users to disable the ``HeadingInOrderedListFilter`` for documents where headings are put inside ordered lists intentionally.

But this is for another issue/commit.

PR #442, opened 2015-10-06
XWIKI-12660: With xwiki.showviewaction=0 and url.standard.hideViewAct…

AndreasJonsson:
…ion=true, requests for nested document gets mapped to UnknownAction.

This fix computes the action path using the same method as in org.xwiki.url.internal.standard.entity.AbstractEntityResourceReferenceResolver

I'm not sure about the xmlrpc special case.  But hasn't xmlrpc supporb been dropped?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiRequestProcessor.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiRequestProcessor.java
@@ -19,54 +19,65 @@
  */
 package com.xpn.xwiki.web;
 
-import java.io.IOException;
+import java.util.List;
+import java.net.URL;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.xwiki.configuration.ConfigurationSource;
 
-import com.xpn.xwiki.internal.XWikiCfgConfigurationSource;
+import org.xwiki.resource.internal.entity.EntityResourceActionLister;
+import org.xwiki.url.internal.standard.StandardURLConfiguration;
+import org.xwiki.url.ExtendedURL;
+
 
 /**
  * @version $Id$
  */
 public class XWikiRequestProcessor extends org.apache.struts.action.RequestProcessor
 {
+    private static final String VIEW_ACTION = "view";
+
+    private final EntityResourceActionLister entityResourceActionLister = Utils.getComponent(EntityResourceActionLister.class);
+
+    private StandardURLConfiguration configuration = Utils.getComponent(StandardURLConfiguration.class);
+
     protected static final Logger LOGGER = LoggerFactory.getLogger(XWikiRequestProcessor.class);
 
     @Override
     protected String processPath(HttpServletRequest httpServletRequest, HttpServletResponse httpServletResponse)
-        throws IOException
     {
-        String result = super.processPath(httpServletRequest, httpServletResponse);
+        try {
+            /**
+             * Extract the action using the same method as in {@link
+             * org.xwiki.url.internal.standard.entity.AbstractEntityResourceReferenceResolver}
+             *
+             * We cannot use the EntityResourceReferenceResolver because the execution context have not been
+             * initialized here.
+             */
+            final String url = httpServletRequest.getRequestURL().toString();
 
-        ConfigurationSource configuration =
-            Utils.getComponent(ConfigurationSource.class, XWikiCfgConfigurationSource.ROLEHINT);
-        if ("1".equals(configuration.getProperty("xwiki.virtual.usepath", "1"))) {
-            // Remove /wikiname part if the struts action is /wiki
-            if (httpServletRequest.getServletPath().equals(
-                "/" + configuration.getProperty("xwiki.virtual.usepath.servletpath", "wiki"))) {
-                int wikiNameIndex = result.indexOf("/", 1);
-                if (wikiNameIndex == -1) {
-                    result = "";
-                } else {
-                    result = result.substring(wikiNameIndex);
+            final ExtendedURL extendedURL = new ExtendedURL(new URL(url), httpServletRequest.getContextPath());
+
+            final List<String> pathSegments = extendedURL.getSegments();
+
+            String action = VIEW_ACTION;
+
+            if (pathSegments.size() > 0) {
+                action = pathSegments.get(0);
+                if (this.configuration.isViewActionHidden()
+                    && !VIEW_ACTION.equals(action)
+                    && !this.entityResourceActionLister.listActions().contains(action)) {
+                    action = VIEW_ACTION;
                 }
             }
-        }
 
-        if (StringUtils.countMatches(result, "/") <= 2) {
-            if (result.startsWith("/xmlrpc/")) {
-                return "/xmlrpc/";
-            } else {
-                return "/view/";
-            }
-        } else {
-            return result.substring(0, result.indexOf("/", 1) + 1);
+            return "/" + action + "/";
+        } catch (Exception e) {
+            throw new RuntimeException(e);
         }
     }
+
 }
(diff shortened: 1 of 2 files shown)


tmortagne:
> But hasn't xmlrpc supporb been dropped?

II's now (since 7.3M1) a deprecated optional extension. See http://jira.xwiki.org/browse/XWIKI-12605.

vmassol:
I'm handling it

vmassol:
Thanks for the PR Andreas. Yes, we need to continue supporting xmlrpc. We just removed it from platform and made it optional but it's still supposed to work.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiRequestProcessor.java
```diff
@@ -19,54 +19,65 @@
...
+
+            final List<String> pathSegments = extendedURL.getSegments();
+
+            String action = VIEW_ACTION;
+
+            if (pathSegments.size() > 0) {
+                action = pathSegments.get(0);
+                if (this.configuration.isViewActionHidden()
+                    && !VIEW_ACTION.equals(action)
+                    && !this.entityResourceActionLister.listActions().contains(action)) {
```
vmassol: I don't understand where you're handling path-based mode in this code. In path-base the first segment is going to be "wiki" and the second one will be the wiki name. The action will only be the 3rd one.

In the URL module we use a WikiReferenceExtractor component in WikiEntityResourceReferenceResolver to normalize the path segments (by removing the first 2 segments if need be).

vmassol: We also need to find a way to reuse the URL module more I think. You say the EC has not been initialized yet, maybe we can initialize it in a Filter before (something I've been wanting to do). But I agree we could do this in a second step since it's more involved and needs some deep thinking :)

AndreasJonsson: I'm sorry, I forgot about path based mode...

AndreasJonsson: Also, I'm running this fix myself and it doesn't seem to work entirely, as if you add /view/ you will go to a space named 'view'.  And I guess that you should be able to optionally add /view/, right?

vmassol: Yes /view/ should go to the wiki's home page.

There are plenty of use cases indeed and I think we really need to find a way to reuse the code from the URL module more so that we don't have several places with the same logic (which is complex to handle all cases).

I'm in meeting during the whole morning and I don't have much time this afternoon but I'll try to think about some idea too.

vmassol: I've checked the code a bit more and AFAIU ```processPath``` is used to identify the mapping in struts-config.xml and from it an Action is found and then executed. AFAICS 1) there's no "xmlrcp" mapping in struts-config.xml and thus it's not used and 2) there's a "xmlrpc" mapping in web.xml and thus when a xmlrpc URL is called it never goes through Struts.

Thus I believe it's safe to drop the xmlrpc part in XWikiRequestProcessor

Correct?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiRequestProcessor.java
```diff
@@ -19,54 +19,65 @@
...
-                if (wikiNameIndex == -1) {
-                    result = "";
-                } else {
-                    result = result.substring(wikiNameIndex);
+        try {
+            /**
+             * Extract the action using the same method as in {@link
+             * org.xwiki.url.internal.standard.entity.AbstractEntityResourceReferenceResolver}
+             *
+             * We cannot use the EntityResourceReferenceResolver because the execution context have not been
```
vmassol: Actually you shouldn't use this class because that's for the "reference" URL scheme :)

I guess you meant using ResourceReferenceResolver<ExtendedURL> (default hint).

vmassol:
One possibility is to make sure that using a ResourceReferenceResolver will work even when no EC is set. Do you know where it fails? Anyway I'll try it out somehow.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiRequestProcessor.java
AndreasJonsson: Right but it is the EntityResourceReference is the only one exposing the "action" field and thus using the EntityResourceReferenceResolver to resolve the URL should give you the action.  But maybe the action should be exposed by all ResourceReferences?

Maybe a separate component should be created to compute the action specifically so it could be used both by the RequestProcessor to produce the path and by the ResourceReferenceResolver?  I'm guessing that we want consistency in how the action is determined in both these places.

AndreasJonsson:
The failures is due to looking up properties in the server document of the current virtual wiki.

The first failure was an NPE in XWQLExecutor:

{code}
            if (query.getWiki() != null) {
                if (currentEntityReference.getType() == EntityType.WIKI) {
                    this.context.setCurrentEntityReference(new WikiReference(query.getWiki()));
                } else {
                    this.context.setCurrentEntityReference(currentEntityReference.replaceParent(
                        currentEntityReference.extractReference(EntityType.WIKI), new WikiReference(query.getWiki())));
                }
            }
{code}

Where I'm guessing it would be harmless to simply check for null:

{code}
            if (query.getWiki() != null) {
                if (currentEntityReference == null || currentEntityReference.getType() == EntityType.WIKI) {
                    this.context.setCurrentEntityReference(new WikiReference(query.getWiki()));
                } else {
                    this.context.setCurrentEntityReference(currentEntityReference.replaceParent(
                        currentEntityReference.extractReference(EntityType.WIKI), new WikiReference(query.getWiki())));
                }
            }
{code}

But it failed again later when trying to figure out if there was a custom mapping configured for the property.  Then I gave up on that idea.

AndreasJonsson:
So, the problem with using the resolver is, thus, that it also resolves the current wiki, which requires access to the wiki server documents.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiRequestProcessor.java
vmassol: All Resolvers will return an EntityResourceReference (when the passed URL is an URL pointing to an Entity)! Again the EntityResourceReferenceResolver is not the right one: it's only for the "reference" URL scheme (see http://design.xwiki.org/xwiki/bin/view/Design/AlternateURLScheme). The only resolve that must be used is @Inject ResourceReferenceResolver<ExtendedURL>. It'll return a ResourceReference which can be cast to an EntityResourceReference in case of an Entity URL. And you have a getAction() on EntityResourceReference.

vmassol:
So indeed, we're doing "too much" in the  ResourceReferenceResolver implementations for what is needed in XWikiRequestProcessor (we just need to find the action).

BTW I remember that's also one reason I had split the Resource resolving into 2 resolvers: ResourceReferenceResolver and ResourceTypeResolver. Note that in RoutingFilter (which is executed before that processor) we extract the ResourceType (which FYI we put in the request).

So we have several options but one of them would be to introduce a new EntityResourceActionResolver component interface in charge of extracting just the action from an ExtendedURL, and refactor the existing ResourceReferenceResolver implementations to use it. However it may not be easy since guessing the Action requires a lot of code. I need to analyze this tomorrow or in the coming days.

A simpler solution is to bulletproof the ResourceReferenceResolver implementations so that they work even when extracting the wiki descriptor is failing and have some default behavior in this case (I also need to analyze this to see if a default behavior makes sense).

I'm in a meeting tomorrow morning but I should be able to work on this in the afternoon. I'll post back here my findings.

OFC the longer term solution is to drop Struts (what we want to do) and thus only use the ResourceReferenceHandlerServlet.

vmassol:
ok I'm trying to implement an EntityResourceActionResolver component

vmassol:
See http://jira.xwiki.org/browse/XWIKI-12660?focusedCommentId=88314&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-88314

Thanks!

PR #443, opened 2015-10-06
XWIKI-12661: URL templates in javascript resources does not take hidd…

AndreasJonsson:
…en view action into account.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
@@ -1308,7 +1308,7 @@ XWiki.Document = Class.create({
    */
   getURL : function(action, queryString, fragment) {
     action = action || 'view';
-    var url = XWiki.Document.URLTemplate;
+    var url = action == 'view' ? XWiki.Document.ViewURLTemplate : XWiki.Document.URLTemplate;
     var spaceSegments = this.spaces.map(function(spaceSegment) {
       return encodeURIComponent(spaceSegment);
     }).join('/');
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1383,6 +1383,7 @@ if (htmlElement.readAttribute('data-xwiki-reference') != null) {
   XWiki.Document.currentDocumentReference = XWiki.currentDocument.getDocumentReference();
 }
 XWiki.Document.URLTemplate = "$xwiki.getURL('__space__.__page__', '__action__')";
+XWiki.Document.ViewURLTemplate = "$xwiki.getURL('__space__.__page__', 'view')";
```
mflorea: Thanks for this pull request! You could have also used the ``$xwiki.getURL('__space__.__page__')`` signature (without the action). I would add a short comment (besides the commit message) to explain why we need a separate template for the view URL, something like

    // We need a separate template for the view URL because the action name is missing when Short URLs are enabled.

AndreasJonsson: Please note that it is not sufficient to just cherry pick this to the 7.2 branch, as jumpToPage.js has recently been refactored.  There are separate templates for jumpToPage.js in 7.2.

mflorea: > jumpToPage.js has recently been refactored

I'm not sure what you mean. The only relevant commit on ``jumpToPage.js`` this year is https://github.com/xwiki/xwiki-platform/commit/3bfa9cfcc8b0fafe2cd8cf2326ed5b848f8c963f .

AndreasJonsson: Forget about it!  I was looking at an outdated version of the 7.2 branch...

PR #444, opened 2015-10-20
XWIKI-12671: The long submenu items are overflowing the box of the menu

KebirYacine:


evalica:
Thanks for the PR. In the end I've opted for a non-JS solution since it was simpler, see https://github.com/xwiki/xwiki-platform/commit/b216e7e381cc8798ae00ae6ed18e257509344d86

PR #448, opened 2015-11-12
XWIKI-9685: "Edit" links on the Wiki Index actually link to view mode

msmeria:


gdelhumeau:
Your pull request could be accepted as it is, but actually, it does not solve everything.
When you are in "edit" mode, on a Wiki Descriptor, you cannot add aliases, meanwhile you can do it in "view" mode.
That is why I've put a link to the "view" mode at first.
The correct fix is to do what you've done + add the ability to manage aliases in edit mode.

Thanks anyway,
Guillaume

msmeria:
Ok. I'll close this PR.

PR #451, opened 2015-11-23
XWIKI-12851:Extra "Edit Section" buttons displayed on Blog Archives p…

zoubir:
* These section editing buttons are displayed because of the html macro with wiki=true,  after verification there is no need to use the html macro here


tmortagne:
I guess you tested it but It feel a bit weird that html macro would suddenly not be needed anymore.

Enygma2002:
@tmortagne has a point. I see that the `#info` velocity macro is used and its output is HTML. That will not be rendered now that the `{{html}}` macro call is being removed.

tmortagne:
Note that you can probably use {{info}} instead of #info if you want to pursue the "get right ow html" path (which would be good, just need to make sure everything that needs to be converted is).

PR #453, opened 2015-12-08
XWIKI-8919: Add a Sandbox entry in the Application Panel

msmeria:


tmortagne:
We usually put the application panel in a dedicated ApplicationsPanelEntry page in the the space of the application.

vmassol:
Thanks Manuel! Indeed as Thomas mentioned, this is even a requirement that is defined in our best practices at http://dev.xwiki.org/xwiki/bin/view/Community/ApplicationDevelopmentBestPractices

msmeria:
Ok, I'll move the object to a separate page.

evalica:
For me Sandbox is not an Application, but a set of content pages. I'm not sure it should be in the AppBar.

evalica:
The app should be in the future the Help Center, but it should contain other pages like the syntax, etc. So the same Sandbox would not be used.

vmassol:
For me it's definitely an app: 
- it's a set of wiki pages having a common goal
- it can be installed or removed
- it's already an app in the SCM

I don't see any reason to treat it differently and considering it as an app brings advantages:  ability to hide it from the UI (from the app bar) (without having to remove it).

Any reason Caty? Also it might be better to comment in the JIRA issue or to bring it up on the list. WDYT?

xwiki-platform-core/xwiki-platform-sandbox/src/main/resources/Sandbox/ApplicationsPanelEntry.xml
```diff
@@ -0,0 +1,109 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
vmassol: missing license :)

vmassol: you should run mvn xar:format

xwiki-platform-core/xwiki-platform-sandbox/src/main/resources/Sandbox/ApplicationsPanelEntry.xml
```diff
@@ -0,0 +1,109 @@
...
+      <content></content>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.platform.panels.Applications</extensionPointId>
+    </property>
+    <property>
+      <name>platform.panels.sandboxApplication</name>
+    </property>
+    <property>
+      <parameters>label=$services.localization.render('xe.panels.quicklinks.sandbox')&#xd;
```
vmassol: Not very nice. We should instead deprecate that key and introduce a new one probably.

xwiki-platform-core/xwiki-platform-sandbox/src/main/resources/Sandbox/ApplicationsPanelEntry.xml
```diff
@@ -0,0 +1,109 @@
...
+target=Sandbox.WebHome&#xd;
+icon=icon:space</parameters>
```
vmassol: I don't think that "space" is a nice icon for sandbox ("space" icon being a folder icon). Other suggestions:
* coffee
* cut

PR #454, opened 2015-12-09
XWIKI-8830: Have the AWM application use translations from AppWithinMinutes.Translations instead of from ApplicationResources.properties

msmeria:


mflorea:
``mvn xar:format`` is missing on ``AppWithinMinutes.Translations.xml``.

msmeria:
I'll update it [hopefully] tomorrow.

msmeria:
Done in https://github.com/xwiki/xwiki-platform/pull/456

PR #457, opened 2015-12-11
XWIKI-9728 : UI issue on Panel Wizard Page about Wiki Information Panel

acotiuga:
* deleted the panel file and the parent directory


evalica:
Hi Alex. Thank for your PR. We will accept it as soon as 7.4-RC1 is released, so this will be available in 8.0.

evalica:
Thank you for your contribution!

PR #461, opened 2016-01-07
XWIKI-12974 : Fix translations in Blog Application

acotiuga:
Translation keys used in Javascript must be escaped using JavaScript String rules.
Instead of general translations (platform.core.profile.passwd.notAllowed), Blog application must use some specific translations (e.g. xe.blog.code.hide).
Replace all the hardcoded messages in the application with translations.


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2448,7 +2452,6 @@ xe.blog.manageCategories.js.add.inProgress=Adding category...
 xe.blog.manageCategories.js.add.error.401=You have been logged out, please refresh and log in
 xe.blog.manageCategories.js.add.error.403=You are not allowed to create the target page
 xe.blog.manageCategories.js.add.error.409=Target page already exists, please choose a different name
-xe.blog.manageCategories.js.add.categoryAlreadyExists=This category already exists.
```
tmortagne: We don't delete keys since they are APIs, we move them to the deprecated section at the end of the file.

acotiuga: I added that translation in a previous commit yesterday.

tmortagne: Ok then :) Sorry for the noise.

PR #470, opened 2016-03-10
XWIKI-12870:Empty Select on property of Database List linked with an integer property

zoubir:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -282,8 +285,16 @@ public String getQuery(XWikiContext context)
...
-                    fromStatements.add("StringProperty as idprop");
                     whereStatements.add("obj.id=idprop.id.id and idprop.id.name='" + idField + "'");
+                    // Get the from statements according to the type of property
+                    try {
+                        PropertyClass pc = (PropertyClass) context.getWiki().getDocument(classname, context).getXClass()
+                                .get(idField);
+                        String classType = (pc == null) ? "StringProperty" : pc.newProperty().getClass().getName();
+                        fromStatements.add(classType + " as idprop");
+                    } catch (XWikiException e) {
+                        e.printStackTrace();
```
vmassol: Exceptions need to be handled properly. The severity needs to be analyzed and a proper solution found. If the error is not important, then logging can be used (See http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HLoggingBestPractices). OTOH if the error is important and we cannot continue then an exception needs to be raised.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -294,8 +305,17 @@ public String getQuery(XWikiContext context)
...
                         whereStatements.add("obj.id=valueprop.id.id and valueprop.id.name='" + valueField + "'");
+                        // Get the from statements according to the type of property
+                        try {
+                            PropertyClass pc = (PropertyClass) context.getWiki().getDocument(classname,
+                                    context).getXClass()
+                                    .get(valueField);
+                            String classType = (pc == null) ? "StringProperty" : pc.newProperty().getClass().getName();
+                            fromStatements.add(classType + "  as valueprop");
+                        } catch (XWikiException e) {
+                            e.printStackTrace();
```
vmassol: same comment. Also this looks like code duplication so we'd need a method to be created so that we don't duplicate. Thanks

vmassol: Question: why pc == null means String Property? Maybe a comment would be nice to explain this (doesn't seem obvious to me)

vmassol: BTW (not related to your commit), it seems we have SQ-injection in this method, we need to fix this too with parametrized queries.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
zoubir: Thanks For your remarks Vincent, can I add this method

```
    private String getPropertyType(String className, String propertyName, XWikiContext context)
    {
        PropertyClass pc = null;
        try {
            pc = (PropertyClass) context.getWiki().getDocument(className, context).getXClass().get(propertyName);
        } catch (XWikiException e) {
            LOGGER.error("Failed to get document [{}]. Root cause: [{}]", className,
                    ExceptionUtils.getRootCauseMessage(e));
        }
        // we consider that the default type is StringProperty
        return pc == null ? "StringProperty" : pc.newProperty().getClass().getName();
    }
```

vmassol: Seems ok. Except that you shouldn't ExceptionUtils.getRootCauseMessage() since we want the stack trace for errors! It should be:

```
LOGGER.error("Failed to get property [{}] for document [{}]", propertyName, className, e);
````
Other comment would be to check if there's a way to access the property class without resorting to calling context.getWiki().getDocument() (haven't checked the code of DBListClass).

Then "we consider that the default type is StringProperty" is not good. You should never explain what you do (that's obvious, the reader just needs to read the code...). You must explain WHY you do it :) So in this case why do you consider that the default type is String Property?

Comment (vmassol) [inline xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java]: Actually a better solution would be to have a try/catch at the level of getQuery() with a more informative message like "Failed to compute query for DBList ....". You'd need to refactor a bit the code since it's already way too much nested.

Comment (zoubir) [inline xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java]: > Then "we consider that the default type is StringProperty" is not good. You should never explain what you do

Indeed, I agree with you, this should be intuitive for the reader , ( But  I got this information from the old behaviour of this function since the StringProperty is the most used ) , but I'll let it to NULL and hundel the exception.
I'll try to refactor the code and commit , 
Thanks Vincent

File: xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java [hunk 3]
```diff
@@ -82,6 +83,8 @@ public DBListClass()
             if (item != null) {
                 if (item instanceof String) {
                     result.add(new ListItem((String) item));
+                } else if (item instanceof Number) {
```
mflorea: This IF does the same thing as the one above: creates a new ``ListItem`` with the string value of ``item`` and adds it to the result. They could be merged.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -282,8 +285,14 @@ public String getQuery(XWikiContext context)
                     select.append("doc." + idField);
                 } else {
                     select.append("idprop.value");
-                    fromStatements.add("StringProperty as idprop");
                     whereStatements.add("obj.id=idprop.id.id and idprop.id.name='" + idField + "'");
+                    // Get the from statements according to the type of property
+                    try {
+                        String classType = getPropertyType(classname, idField,context);
```
mflorea: There's a missing space before ``context``. Have you formatted your code? You need to follow the [Java Code Style](http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -282,8 +285,14 @@ public String getQuery(XWikiContext context)
...
+                        fromStatements.add(classType + " as idprop");
+                    } catch (Exception e) {
+                        LOGGER.error("Failed to parse SQL script [{}]. Continuing with non-rendered script.", sql, e);
```
mflorea: I don't understand what this error log message has to do with the code executed in the ``try`` block above. Where are you "parsing the SQL script"? Also, I see that ``getPropertyType`` has its own try/catch block so what exceptions are you hoping to catch here?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -294,8 +303,14 @@ public String getQuery(XWikiContext context)
                         select.append(", doc." + valueField);
                     } else {
                         select.append(", valueprop.value");
-                        fromStatements.add("StringProperty as valueprop");
                         whereStatements.add("obj.id=valueprop.id.id and valueprop.id.name='" + valueField + "'");
+                        // Get the from statements according to the type of property
+                        try {
+                            String classType = getPropertyType(classname, valueField,context);
```
mflorea: Same as above. You need to format your code.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -294,8 +303,14 @@ public String getQuery(XWikiContext context)
...
+                            fromStatements.add(classType + " as valueprop");
+                        } catch (Exception e) {
+                            LOGGER.error("Failed to parse SQL script [{}]. Continuing with non-rendered script.", sql, e);
```
mflorea: Same as above, the log message doesn't seem related to the code executed in the ``try`` block above and I don't see what exceptions are you hoping to catch here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -594,4 +609,15 @@ public void displayView(StringBuffer buffer, String name, String prefix, BaseCol
...
+
+    private String getPropertyType(String className, String propertyName, XWikiContext context) throws Exception
+    {
+        PropertyClass pc = null;
+        try {
+            pc = (PropertyClass) context.getWiki().getDocument(className, context).getXClass().get(propertyName);
+        } catch (XWikiException e) {
+            LOGGER.error("Failed to get property [{}] for document [{}]", propertyName, className, e);
+        }
+        return pc.newProperty().getClass().getName();
```
mflorea: This will throw a ``NullPointerException`` if an ``XWikiException`` is caught above. You need to choose between

* falling back on a default property type in case of an ``XWikiException`` or
* let the ``XWikiException`` propagate

In general, there are two main reasons for which you would catch an exception:
* provide a fall back or
* throw a more domain-specific exception that wraps the caught exception.

mflorea:
@zoubir thanks for making this PR. It still needs a few adjustments before we merge it. Thanks for your patience.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
mflorea: It seems you copied this line from some other place in this file.. but the context here is quite different.

zoubir:
Thanks Marius for your remarks
the modifications have been done.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -80,7 +81,7 @@ public DBListClass()
             // were empty strings). This means we need to check for NULL and ignore NULL entries
             // from the list.
             if (item != null) {
-                if (item instanceof String) {
+                if (item instanceof String || item instanceof Number) {
                     result.add(new ListItem((String) item));
```
sdumitriu: This will fail. Marius was suggesting to use this instead:

    result.add(new ListItem(item.toString()));

or:

    result.add(new ListItem(String.valueOf(item)));

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -111,7 +112,14 @@ private String toStringButEmptyIfNull(Object object)
     {
         List<ListItem> list = getCachedDBList(context);
         if (list == null) {
-            String hqlQuery = getQuery(context);
+            String hqlQuery = null;
+            try {
+                hqlQuery = getQuery(context);
+            } catch (XWikiException e) {
+                LOGGER.error("Failed to get the list", e);
```
mflorea: What "list" is the error message referring to? The try block is only calling ``getQuery``, so it looks to me that the only thing that can fail there is getting the query.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -111,7 +112,14 @@ private String toStringButEmptyIfNull(Object object)
...
+                list = new ArrayList<ListItem>();
+                return list;
```
mflorea: Doesn't ``hqlQuery`` remain ``null`` if ``getQuery`` fails? And if ``hqlQuery`` remains ``null`` isn't the ``list`` assigned to an empty list just two lines below?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
zoubir: Ok I'll do this , Thanks @sdumitriu

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
zoubir: Ok @mflorea, I'll correct it

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
```diff
@@ -282,8 +290,10 @@ public String getQuery(XWikiContext context)
                     select.append("doc." + idField);
                 } else {
                     select.append("idprop.value");
-                    fromStatements.add("StringProperty as idprop");
                     whereStatements.add("obj.id=idprop.id.id and idprop.id.name='" + idField + "'");
+                    // Get the from statements according to the type of property
+                    String classType = getPropertyType(classname, idField, context);
+                    fromStatements.add(classType + " as idprop");
```
mflorea: Any reason why you swapped the ``fromStatements`` with the ``whereStatements``. Syntactically there's no problem, but semantically the FROM clause is before the WHERE clause and thus it's easier to read the code if you keep them in order.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/DBListClass.java
zoubir: Hello Marius ,
I've add a default value to null on the getQuery() function , as you can see it on my last commit , here :
https://github.com/zoubir/xwiki-platform/commit/5fffe0a254b14ad193ef1a969624ebb7e56d721b#diff-712bf30d51cacf05af1e8ef2c8898181R202

Thanks

mflorea: @zoubir you didn't fix or prevent the ``NullPointerException`` that you can have in ``getPropertyType()``. In your last commit you've added a try/catch to catch the ``NullPointerException`` from ``getPropertyType()``  as if it is an expected exception. You should prevent the ``NullPointerException`` not catch it.

acotiuga:
Closing this PR as there's an updated one https://github.com/xwiki/xwiki-platform/pull/2084

PR #471, opened 2016-03-30
XWIKI-13272  Update xwiki syntax in XWiki.RatingsTranslations to plain/1.0

acotiuga:


vmassol:
Actually I already fixed it yesterday I think :) Checking

vmassol:
ah no I didn't because it's not following the rule of being named Translations. I need to update my check.

vmassol:
Note that this page (and any other page) should be moved in a space dedicated for Ratings but that's another PR ;)

PR #473, opened 2016-03-31
XWIKI-13277: Add Rename operation Page Object & Tests

msmeria:


Enygma2002:
Reverted due to failing tests: http://ci.xwiki.org/job/xwiki-platform/2700/testReport/

PR #477, opened 2016-04-07
XWIKI-13288 : Filtering on wiki pretty name is broken in wikis index

acotiuga:


xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/WikisLiveTableResultsMacros.xml
```diff
@@ -95,6 +95,7 @@
     #end
     ## only safe version should be used in SQL queries
     #set($safe_tableAlias = "$!{tableAlias.replaceAll('[^a-zA-Z0-9_.]','')}")
```
mflorea: This line is not needed any more because ``#livetable_getTableAlias`` does

    #set ($safe_tableAlias = "$prefix$suffix.replaceAll('\W', '')")

which is equivalent.

PR #479, opened 2016-04-10
XWIIKI-13284 Add support for loading RSS feeds over HTTPS in the RSS macro

seanthegeek:
…Macro


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/DefaultRomeFeedFactory.java
```diff
@@ -43,6 +44,9 @@
      * The maximum number of milliseconds to wait when inquiring the RSS feed provider.
      */
     private static final int TIMEOUT_MILLISECONDS = 5000;
+    private static final String USER_AGENT_HEADER = "User-Agent";
+    private String version = this.getClass().getPackage().getImplementationVersion();
```
tmortagne: I feel those fields would make more sens as static, for example:

    private static final String VERSION = DefaultRomeFeedFactory.class.getPackage().getImplementationVersion();

seanthegeek:
Agreed. Fixed.

tmortagne:
Applied, thanks !

Would be great if you could add something about it in http://www.xwiki.org/xwiki/bin/view/ReleaseNotes/ReleaseNotesXWiki81M2 and also document it on xwiki.org.

seanthegeek:
Done

PR #481, opened 2016-04-13
XWIKI-13321  Add a new "visibility" xproperty to the XWiki.ConfigurableClass class

acotiuga:


acotiuga:
I close this PR because is not anymore valid. I'm creating another one.

PR #483, opened 2016-04-15
XWIKI-13092  Incorrect message on registration form when email is not configured

acotiuga:
* changed only the error code for password validation

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
@@ -637,7 +637,7 @@
   ## Handle output from the registration.
   #if($reg &amp;&amp; $reg &lt;= 0)
     {{error}}
-    #if($reg == -7)
+    #if($reg == -2)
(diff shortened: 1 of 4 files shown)


mflorea:
@acotiuga we can see from the diff that you "changed only the error code for password validation". What we can't see from the diff is why? :)

acotiuga:
@mflorea I changed it because `-2` was too general and it was returned for several cases: when password validation fails, when send validation email fails and also when the user creation fails. All of these issues were expressed to the user as password validation issue. Does this answer to the `Why` question?

mflorea:
@acotiuga afaics ``c.x.x.XWiki#createUser()`` was returning -2 *only* when:

    !password.equals(password2) || password.trim().equals("")

Changing the returned code can be viewed as API breakage. There may be existing code (an extension) that expects -2 to be returned when the password and its confirmation doesn't match.

When you say that -2 "was returned for several cases" you are actually referring to a different method, ``c.x.x.api.XWiki#createUser()``. The problem with this method is that it reuses the -2 code that can be returned by ``c.x.x.XWiki#createUser()``. Changing the returned code here can also be seen as API breakage.

I would break ``c.x.x.api.XWiki#createUser()`` because the -2 used there is for any error which probably means it's not tested explicitly by the code that calls this method (an else branch is used instead).

PR #484, opened 2016-04-18
[Misc] Fix for wrong documentation

acotiuga:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
```diff
@@ -1283,7 +1283,7 @@ public void flushCache()
     /**
      * Privileged API to create a new user from the request This API is used by RegisterNewUser wiki page
      *
-     * @return true for success/false for failure
+     * @return integer status code
```
vmassol: Thanks. Note that the best practice is to use "the", see http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HJavadocBestPractices (which points to http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide).

vmassol: Also you should indicate what the integer means or point to a place that describes it.

PR #492, opened 2016-04-22
XWIKI-13350: Add Group functional tests

msmeria:


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/GroupsPage.java
```diff
@@ -0,0 +1,100 @@
...
+    {
+        this.addGroupButton.click();
+    }
+
+    public void clickCreateGroupButton()
+    {
+        this.createGroupButton.click();
+    }
+
+    // Method to create a new Group
```
Enygma2002: You need to use multiline javadoc comments (/** **/) and not inline code comments (//).

PR #493, opened 2016-05-03
XWIKI-13395 : The DBList field always propses the blog introduction f…

DjeblounSidali:
…irst in AWM

Remove this class example since It's not a good idea to have an example document from another application here this can lead to confusion.
This can make sense only if AWM had a dependency on the blog

### DIFF
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/DBList.xml
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/DBList.xml
@@ -51,7 +51,7 @@
     <validationScript/>
     <databaseList>
       <cache>0</cache>
-      <classname>Blog.BlogPostClass</classname>
+      <classname></classname>
       <customDisplay/>
       <disabled>0</disabled>
       <displayType>select</displayType>


mflorea:
@DjeblounSidali as I explained on [XWIKI-13395](http://jira.xwiki.org/browse/XWIKI-13395) the ``Blog.BlogPostClass`` is used as an example to give you an idea of how the Database List field can be configured. I don't see the problem you are trying to solve here.

DjeblounSidali:
@mflorea technically it's bad IMO to link an application (which can be installed or not on the wiki) from another application without having a dependencies between them.

vmassol:
Indeed, that's a good point @DjeblounSidali 

@mflorea btw, what happens if the blog app is not installed?

mflorea:
@DjeblounSidali there's no dependency between AppWithinMinutes and Blog because AppWithinMinutes works fine without Blog. Again, the configuration of the Database List field is just an example. @vmassol nothing is broken if the Blog application is not installed. Otherwise users wouldn't be able to configure the Database List to use a class that doesn't have any instances or that doesn't exist (yet).

vmassol:
@mflorea  you say nothing is broken but what is displayed when the blog app is not installed? If nothing is displayed then it's broken obviously since the goal is to display something as an example.

vmassol:
FTR in maven parlance, this is called an optional runtime dependency.

acotiuga:
Closing this PR as https://jira.xwiki.org/browse/XWIKI-13395 was closed as solved by https://jira.xwiki.org/browse/XWIKI-15170.

PR #496, opened 2016-05-12
XWIKI-13040 : Velocity error displayed for XWikiGuest user when accepting invitation

acotiuga:


xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
```diff
@@ -394,9 +394,10 @@ position: relative;
     <className>XWiki.UIExtensionClass</className>
     <guid>d2420ff0-a5b9-4fae-a798-fc623aac7a6b</guid>
     <property>
-      <content>{{include reference='XWiki.RatingsMacros'/}}
-
-{{velocity}}
+      <content>{{velocity}}
+#if (!$isGuest)
```
vrachieru: I'd fix this issue by checking the rights rather than doing this.
You only "fixed" it for guest, if a logged in user with no view right for the macros page/space/whatever would have the same issue..

acotiuga: My use case: guest user with view rights. So this might be a mix of being guest or not having view rights. WDYT ?

vrachieru: Well, yes, as you said. The issue is with the view right.
You don't really care if the user is a guest or logged in, you just want the check beforehand that the current user has viewing right on the macros page and if he doesn't I guess you could skip the whole logic described in the UI extension.
If memory serves me right, it would be something in the lines of:
#set($user = $xcontext.getUser())
#set($canView = $xwiki.getDocument('XWiki.RatingsMacros').hasAccessLevel('view', $user))
#if($canView)
## do stuff
#end

vmassol: Hi Victor! Cool to see you active.

I'm not sure we want guest users to be able to rate something since we want users to rate only once AFAIR, right?

So we do need to check if the user is not guest. We also need to check if he has permissions to rate something and I don't know what's involved there (view access on the XClass but probably more rights to add the rating page).

Thanks

vrachieru: Hi, Vincent!

I was just saying that if the current user doesn't have view rights on anything related to the ratings logic (ex. XWiki.RatingsMacros) there's no point in evaluating the code in the UI extension because it's going to produce errors (like the one described in the jira issue).

As for restricting guest users voting capability, this is another issue which is handled [here.](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/RatingsMacros.xml#L59)
I thought this was just about displaying.

tmortagne: I agree with @vrachieru, XWIKI-13040 is only about right and not guest and should be handled at right level.

vmassol: ok, if preventing guests from voting is already implemented then perfect!

So yes we need to check rights but probably not just view access to the xclass but also write access to create the voting pages.

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
```diff
@@ -394,25 +394,27 @@ position: relative;
...
-#end
-##
-#set($displayRatings = $configObj.getProperty('displayRatings').value == 1)
-#set($excludedPages  = $stringtool.split($configObj.getProperty('excludedPages').value, ','))
-#set($excludedPages  = $stringtool.stripAll($excludedPages))
-##
-#if($displayRatings &amp;&amp; !$excludedPages.contains($doc.name))
-  #displayFullRating($doc.getDocumentReference() "")
+      <content>{{velocity}}
+#set($canView = $xwiki.getDocument('XWiki.RatingsMacros').hasAccessLevel('view', $xcontext.user))
```
mflorea: You should use the [security script service](http://extensions.xwiki.org/xwiki/bin/view/Extension/Security+Module#HScriptingAPI28since6.1RC129) instead.

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
```diff
@@ -394,25 +394,27 @@ position: relative;
...
+#if($canView)
+  {{include reference='XWiki.RatingsMacros'/}}
```
mflorea: The wiki syntax inside the ``{{velocity}}`` macro is evaluated after the Velocity code (because the Velocity code can generate wiki syntax) so the ``{{include}}`` macro (which is wiki syntax) is evaluated (in this case) after ``#displayFullRating()`` is called. Are you sure it works?

mflorea:
Replaced by #497.

PR #497, opened 2016-05-17
XWIKI-13040:Velocity error displayed for XWikiGuest user when accepti…

zoubir:
…ng invitation
* Adding remarks of Marius ( https://github.com/xwiki/xwiki-platform/pull/496 )

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
+++ b/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
@@ -394,27 +394,33 @@ position: relative;
     <className>XWiki.UIExtensionClass</className>
     <guid>d2420ff0-a5b9-4fae-a798-fc623aac7a6b</guid>
     <property>
-      <content>{{include reference='XWiki.RatingsMacros'/}}
-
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
```diff
@@ -394,27 +394,33 @@ position: relative;
...
-  #displayFullRating($doc.getDocumentReference() "")
-#end
-{{/velocity}}</content>
+      <content>{{velocity}}
+        #set($canView = $services.security.authorization.hasAccess("view", "XWiki.RatingsMacros"))
+        #if($canView)
+          {{include reference='XWiki.RatingsMacros'/}}
+        #end
+        {{/velocity}}
+        {{velocity}}
```
mflorea: It's probably safer to leave an empty line between the two Velocity macros because otherwise the second is considered in-line AFAIK which could break the ratings display, but I'm sure you tested and if it works then it means the empty line is not needed for now.

tmortagne: I agree it would be better to separate them.

PR #498, opened 2016-06-07
XWIKI-13461  Create tour for Homepage

acotiuga:


vmassol:
Hi Alex,

If we move the Tour app in platform, it'll need to be a subdirectory of https://github.com/xwiki/xwiki-platform/tree/master/xwiki-platform-core. In your commit, you seem to commit stuff in the distribution which doesn't seem the best place. Maybe I misunderstood something?

Thanks

acotiuga:
Hi Vincent,

Even if the Tour app will be in platform, shouldn't the tours be placed in the apps that were created for? Taking the example of the Forum app which has 3 tours: all of them are in ForumCode space and are added/removed to/from the wiki with the entire application. IMO we shouldn't keep tours if we don't have the applications. Following this rationale, I created the Homepage tour in the Main space. How does this sound to you? 
Thanks

vmassol:
Ok that's what I was missing :) This commit is not the tour app but the usage of the tour app (ie the tour itself for the home page). So yes you're correct, forget what I said.

acotiuga:
Still, this PR will be complete with the tour app commit.

evalica:
Still I think that the Help space is better than Main.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/src/main/resources/Main/HomepageTour/Translations.xml
```diff
@@ -0,0 +1,101 @@
...
+
+<xwikidoc version="1.2" reference="Main.HomepageTour.Translations" locale="">
+  <web>Main.HomepageTour</web>
+  <name>Translations</name>
+  <language/>
+  <defaultLanguage>en</defaultLanguage>
+  <translation>0</translation>
+  <creator>xwiki:XWiki.Admin</creator>
+  <creationDate>1465277605000</creationDate>
+  <parent>Main.HomepageTour.WebHome</parent>
```
vmassol: Do we want the Tour to be a Nested Page?

vmassol:
@evalica Still, one little issue is that these pages have a dependencies on Main.WebHome (i.e. on the home page), for example through:

    <targetPage>Main.WebHome</targetPage>

And the help app doesn't have a POM dependency on a module that contains Main.WebHome.

evalica:
The location of the Tour content I guess should be in enterprise, not in platform. So this PR should be redone with the new location.

acotiuga:
This PR should be recreated for xwiki-enterprise

PR #499, opened 2016-06-07
XWIKI-13478: On Homepage, replace Dashboard include with content

evalica:
- Replacing the Dashboard include with help content
- Adding a SSX to style the used table

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/src/main/resources/Main/WebHome.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/src/main/resources/Main/WebHome.xml
@@ -33,7 +33,7 @@
   <date>1321284472000</date>
   <contentUpdateDate>1321284472000</contentUpdateDate>
   <version>1.1</version>
-  <title>$services.localization.render('xe.home.title')</title>
+  <title>Home</title>
   <comment/>
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.0</syntaxId>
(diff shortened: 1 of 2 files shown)


vmassol:
Note that the SSX is not that good since the user is supposed to be able to customize that page and the SSX will remain ;)

evalica:
I've used tables in order to have some layout inside the page. 
The SSX removed the table border and margins, so the changes will not be very big. 

Now, the ssx could be removed by the admin that changes the Homepage content.

Alternatively I could have used the {{container}} macro. Unfortunately, the user would need to go through the editor's macro section in order to edit the section's content. That's why I preferred tables (but these come with some styling). Using inline styling (instead of SSX) would make the syntax more scary. 

So I don't know an alternative to that SSX.

vmassol:
"Now, the ssx could be removed by the admin that changes the Homepage content.". That won't happen it's too technical. If the admin was able to do this we wouldn't have needed to modify the home page content at all ;)

It's not a big deal, it just means the SSX will stay and it'll probably have no effect. I haven't looked at it; we just need to make sure it's not going to affect any content the user will put on the home page.

An alternative is also to move this SSX to another hidden page so that the home page stays "clean".

evalica:
If we were to move it, we will need velocity in order to include it. Now we rely on "Use this extension: on this page ..." option.

evalica:
TODO: 
- [ ] Reach a conclusion on layout
- [ ] XWiki.XWikiPreferences changes with the panels
- [ ] Remove the translation key from Title
- [ ] Add Main.WebHome in the Translations release script

PR #500, opened 2016-06-07
XWIKI-13433: Simplify the Appbar by removing some apps

evalica:
- Hidding Invitation, Panels, Scheduler, User Index from the default AppBar configuration


vmassol:
I've mentioned several times that I don't agree about this if we don't also add an Application Index entry at the same time.

evalica:
Applying since we merged https://github.com/xwiki/xwiki-platform/pull/508

PR #503, opened 2016-06-09
XWIKI-13483  After page rename, the old page redirects wrongly to Main.WebHome

acotiuga:
* I replaced the use of the compact serializer with the local serializer in order to get the full name of the new reference instead of 'WebHome'.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultModelBridge.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultModelBridge.java
@@ -83,7 +83,10 @@ public class DefaultModelBridge implements ModelBridge
     private QueryManager queryManager;
 
     /**
-     * Used to serialize a space reference in order to query the child documents.
+     * Used to serialize a space reference in order to query the child documents and to serialize the redirect location.
+     * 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultModelBridge.java
```diff
@@ -213,7 +216,7 @@ public void createRedirect(DocumentReference oldReference, DocumentReference new
             try {
                 XWikiDocument oldDocument = xcontext.getWiki().getDocument(oldReference, xcontext);
                 int number = oldDocument.createXObject(redirectClassReference, xcontext);
-                String location = this.compactEntityReferenceSerializer.serialize(newReference, oldReference);
+                String location = this.localEntityReferenceSerializer.serialize(newReference);
```
mflorea: I'm not sure about the *local* entity reference serializer. What if I move a page to a different wiki?

tmortagne: Yes using the local serializer is wrong. You should use the default serializer and be done with it, there is no reason to try to "optimize" the reference here.

acotiuga:
Is this ok now ?

tmortagne:
Fine with me but I did not worked on rename refactoring so I will let @mflorea decide.

PR #508, opened 2016-06-14
XWIKI-11423: Create an Application Index

evalica:
Have a first iteration for the Application Index, by listing all the applications that have the ''org.xwiki.platform.panels.Applications" extensionPointId.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/pom.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.platform</groupId>
+    <artifactId>xwiki-platform-application</artifactId>
+    <version>8.2-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-platform-application-ui</artifactId>
+  <name>XWiki Platform - Application - UI</name>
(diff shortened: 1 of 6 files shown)


evalica:
- [ ] Reuse Applications.Macros
We've duplicated some Macros from Panels.Applications into Applications.Macros. In the future we should: 
-- Move Panels.Applications from xwiki-platform-panels to xwiki-platform-application
-- Panels.Applications should include the Applications.Macros
-- We need also to think where to put the SSX

Currently we are using the Panels.Applications display for the Index page, but in the future, in order to be consistent with the other Indexes, we should have a livetable. 

Future:
- [ ] Add a livetable to the Applications.WebHome Index page
- [ ] Add applications Descriptors to xwiki-platform-application module

evalica:
- [ ] Make sure to update the order list from http://platform.xwiki.org/xwiki/bin/view/ExtensionPoint/DrawerItem

evalica:
- [ ] Remove #docextra from Applications.WebHome

xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.platform</groupId>
+    <artifactId>xwiki-platform-application</artifactId>
+    <version>8.2-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-platform-application-ui</artifactId>
+  <name>XWiki Platform - Application - UI</name>
+  <packaging>xar</packaging>
+  <description>Contains the Applications Index</description>
```
vmassol: I would use instead "Displays the list of installed Applications"

xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <build>
+    <!-- Needed to add support for the XAR packaging -->
+    <extensions>
+      <extension>
+        <groupId>org.xwiki.commons</groupId>
+        <artifactId>xwiki-commons-tool-xar-handlers</artifactId>
+        <version>${commons.version}</version>
+      </extension>
+    </extensions>
```
vmassol: I don't think you need this since it's already in parent poms.

xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+    <plugins>
+      <!-- The XAR packaging uses the XAR plugin and thus we need to define what version to use -->
+      <plugin>
+        <groupId>org.xwiki.commons</groupId>
+        <artifactId>xwiki-commons-tool-xar-plugin</artifactId>
+        <version>${commons.version}</version>
+      </plugin>
+    </plugins>
```
vmassol: I also don't think this is needed since it's implied by the xar packaging lifecycle.

xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/src/main/resources/Applications/Macros.xml
```diff
@@ -0,0 +1,74 @@
...
+  <date>1465912816000</date>
+  <contentUpdateDate>1465912281000</contentUpdateDate>
+  <version>1.1</version>
+  <title>Macros</title>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>{{velocity output="false"}}
+#macro (displayAppPanelEntry $params)
```
vmassol: I wouldn't use "Panel" in the macro names since it's not about the app panel and in the future we'll probably have an ApplicationEntryClass that will be defined in this module and used in the App Panel. BTW The App Panel should be moved from the Panel UI module to this module too (could be done in a second step but we shouldn't forget it).

evalica:
Vincent, thanks for the review. 
I will merge the current version + adding the Translation page. 
Alex created issues for the remaining things.

PR #513, opened 2016-06-17
XWIKI-13507 : Reuse Applications.Macros in Panels.Applications

acotiuga:
* removed duplicated code

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
@@ -122,38 +122,7 @@
       <category>Information</category>
     </property>
     <property>
-      <content>{{velocity output="false"}}
-#macro (displayAppPanelEntry $params)
-  #if ($params.icon.startsWith('icon:'))
-    #set ($icon = $services.icon.renderHTML($params.icon.substring(5)))
-  #else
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -122,38 +122,7 @@
...
-  #end
-#end
-
-#macro (displayAppPanelEntries $extensionPointId $configDoc)
-  #foreach ($uix in $services.uix.getExtensions($extensionPointId, {'sortByParameter': 'label'}))
-    #maybeDisplayAppPanelEntry($uix $configDoc)
-  #end
-#end
-{{/velocity}}
+      <content>{{include reference="Applications.Macros"/}}
```
Enygma2002: This would currently cause a circular (indirect/inexplicit) dependency between panels-ui and applications-ui (since applications-ui depends on panels-ui).

tmortagne: We really need to move all specific panels out of Panels application. It should contains nothing more than the panels "framework" and maybe a few super generic basic panels not requiring any dependency.

acotiuga: You're right. I'm closing this PR and its changes will be in another one which will also move the Panels.Application in application module, removing at the same time the dependency issue.
Thanks!

PR #515, opened 2016-06-17
XWIKI-6118: add boolean list to filter livetable

Pbast:
(merge/adapt with last macros.vm version)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -295,6 +295,15 @@
                 #end
                 </select>
                 </td>
+              #elseif("$!colprop.type" == 'listboolean' && $isFilterable)
```
gdelhumeau: Question: why do you introduce a "listboolean" type instead of "boolean"? That would be more simple.

Pbast:
I choosed listboolean to show it is not a truly boolean but a boolean list displayed.
But I fixed it like you wish.
Thxs

PR #517, opened 2016-06-23
XWIKI-12935 : Misalignment of Page rights's captcha checkbox

acotiuga:


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminSheet.xml
```diff
@@ -1064,9 +1064,17 @@ document.observe('xwiki:dom:loaded', function() {
     width: 65%;
 }
 
+dl.rights-settings {
```
evalica: not sure why you need to increase specificity by explicitly saying 'dl' - it could be removed.

evalica: We could also add 
`margin-left: 5px;`
in order to better align.

PR #518, opened 2016-06-23
XWIKI-12314 : Improper redirect when you cancel the request for adding a new user

acotiuga:
* Form.disable(window.lb.form) was performing an unusual action by redirecting to another page displaying the same form, while its own purpose was to close the pop-up window just like the "x" button is doing


vmassol:
Note: I haven't tested it; I'm trusting you on this Alex :)

acotiuga:
Thank you!

PR #520, opened 2016-07-01
XWIKI-13519 : Add in the Template Provider Sheet the Description and Icon properties

acotiuga:


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
```diff
@@ -144,6 +144,46 @@
...
+      ## Description
+      ##
+      &lt;dt&gt;
+        &lt;label #if($isEditing)for="XWiki.TemplateProviderClass_0_description"#end&gt;$escapetool.xml($services.localization.render('xe.templateprovider.description'))&lt;/label&gt;
+      &lt;/dt&gt;
+      &lt;dd&gt;
+        #if($isEditing)
+          {{/html}}
+
+          {{html wiki='true' clean='false'}}$!providerObj.get('description'){{/html}}
```
vmassol: why use clean=false here? hmm maybe for performance reasons? If it's for performance reasons, I don't recall if we've defined an official best practice, would be great to make a proposal in this case. Thanks!

vmassol: Should we support wiki syntax in description?

acotiuga: This document is full of this usage and, IMO, requires a revision. I followed the implementation of the other properties. I may take some time to investigate and clean this, if is needed. Thanks!

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3107,6 +3107,9 @@ xe.templateprovider.action=Action on create
 xe.templateprovider.action.info=The action to execute when the create button is pushed, you can configure here whether the new page is saved before it is opened for edition or not.
 xe.templateprovider.terminal=Terminal Page
 xe.templateprovider.terminal.hint=Whether or not to create terminal documents by default when using this template provider.
+xe.templateprovider.description=Description
+xe.templateprovider.icon=Icon
+xe.templateprovider.icon.info=See the {0}full list of icons{1}.
```
vmassol: Would be great to start moving the translations in xwiki-platform-administration-ui or add the new ones there. Need to check with Thomas how we can move translation keys without loosing the translations.

acotiuga: I guess it should be ok for this PR to add them here and later to move all. WDYT?

vmassol: Yes it is, especially since we're at RC1 now for 8.2... :)

Now we need to start being more carefully to stop adding technical debt. Hopefully we have some tools to migrate the translations.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderSheet.xml
vmassol: ok thanks. Maybe be done in a second step (different PR)

PR #523, opened 2016-07-14
XWIKI-13560 : Translation key xe.blog.code.blogcategories displayed when editing a blog post in the object editor

acotiuga:
* the `name` field of the `Blog.CategoryClass` class has no reason to be translation key since it's not rendered
* added `hql` statement in order to exclude `Blog.CategoryTemplate` from the list of displayed `Categories` in object mode.
* `Blog.Categories` should be also excluded from here since it's not displayed in inline mode nor in panels.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-blog/xwiki-platform-blog-ui/src/main/resources/Blog/BlogPostClass.xml
+++ b/xwiki-platform-core/xwiki-platform-blog/xwiki-platform-blog-ui/src/main/resources/Blog/BlogPostClass.xml
@@ -67,7 +67,7 @@
       <separators/>
       <size>1</size>
       <sort>value</sort>
-      <sql/>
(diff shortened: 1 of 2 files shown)


vmassol:
Alex, you said "the name field of the Blog.CategoryClass class has no reason to be translation key since it's not rendered". Do you know what the Blog.CategoryClass xobject is used for in Blog.Categories?

vmassol:
BTW it does have a translation in ApplicationResources.properties

acotiuga:
Here, in [Blog/CategoriesCode.xml#L99](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-blog/xwiki-platform-blog-ui/src/main/resources/Blog/CategoriesCode.xml#L99) is the reason for which I didn't remove in this issue the `Blog.CategoryClass` xobject from `Blog.Categories`. 

Inddeed, the translation is used. So only excluding the 2 pages from the query solves both problems (displaying template and translation key in the categories list in object mode)

PR #525, opened 2016-07-20
XWIKI-13581 : Display the number of Active installs on each extension page on e.x.o.

acotiuga:


xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-ui/src/main/resources/ExtensionCode/ExtensionSheet.xml
```diff
@@ -130,6 +130,10 @@
         #else
           $doc.authors
         #end
+      ##------- Active Installs --------
+      #if ($doc.getValue('id'))
```
sdumitriu: Arebyoubsure you're supposed to be checking for `id` and not `installedCount`?

PR #526, opened 2016-07-22
XWIKI-13581 : Display the number of Active installs on each extension page on e.x.o.

acotiuga:


xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-ui/src/main/resources/ExtensionCode/ExtensionSheet.xml
```diff
@@ -131,7 +131,7 @@
           $doc.authors
         #end
       ##------- Active Installs --------
-      #if ($doc.getValue('id'))
+      #if ($doc.getValue('installedCount') &amp;&amp; $doc.getValue('showInstalledCount'))
```
tmortagne: showInstalledCount might be null and when it's null if you behave as if it was true so you'd better do somehting like

    $doc.getValue('showInstalledCount') != 0

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-ui/src/main/resources/ExtensionCode/ExtensionSheet.xml
```diff
@@ -242,6 +242,7 @@
     |Source|$doc.display('source')
     |Display Icon Location
 (Syntax: #set($comment = '##')${comment}icon:{{html}}&lt;a href="http://www.famfamfam.com/lab/icons/silk/previews/index_abc.png" target="_blank"&gt;icon name&lt;/a&gt;{{/html}}${comment})|$doc.display("icon")
+    |Show installed count|#if($proxyExtensionObject)$doc.display('showInstalledCount', 'view')#else$doc.display('showInstalledCount')#end~
```
tmortagne: installed count has nothing to do with the proxy object. proxy object is for imported extension and you don't need the extension to be imported for it to be installed.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-ui/src/main/resources/ExtensionCode/ExtensionSheet.xml
```diff
@@ -242,6 +242,7 @@
     |Source|$doc.display('source')
     |Display Icon Location
 (Syntax: #set($comment = '##')${comment}icon:{{html}}&lt;a href="http://www.famfamfam.com/lab/icons/silk/previews/index_abc.png" target="_blank"&gt;icon name&lt;/a&gt;{{/html}}${comment})|$doc.display("icon")
+    |Show installed count|$doc.display('showInstalledCount')
```
tmortagne: Since the section label is "Active Installs" you should probably use "Show Active Installs" here, WDYT ?

PR #535, opened 2016-09-08
XWIKI-13679 : User is still added after error registration has failed due to unknown reasons error code -10

acotiuga:
- introduced a new error code for the mail sender errors

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -3396,7 +3396,12 @@ public class XWiki implements EventListener
             // Send validation mail, if needed.
             if ((result > 0) && (withValidation)) {
                 // Send the validation email
-                sendValidationEmail(xwikiname, password, email, validkey, "validation_email_content", context);
+                try {
+                    sendValidationEmail(xwikiname, password, email, validkey, "validation_email_content", context);  
+                } catch (XWikiException e) {
+                    LOGGER.error("User created. Failed to send the email to the created user", e);
+                    return -11;
+                }
             }
 
             return result;
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
@@ -57,6 +57,8 @@ import com.xpn.xwiki.web.XWikiURLFactory;
 
 public class XWiki extends Api
 {
+    private static final int MAIL_SENDER_WRONGLY_CONFIGURED = -11;
+
     /** Logging helper object. */
     protected static final Logger LOGGER = LoggerFactory.getLogger(XWiki.class);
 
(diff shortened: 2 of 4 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
```diff
@@ -57,6 +57,8 @@
 
 public class XWiki extends Api
 {
+    private static final int MAIL_SENDER_WRONGLY_CONFIGURED = -11;
```
sdumitriu: Checkstyle will complain, `private` field must come after the `protected` ones.

And I will complain that this is the first such error defined here, and there's a good reason why other haven't been defined so far: it's not the right place. This is a thin wrapper that shouldn't fix problems in the underlying class. Why not return the right error code from the privileged `com.xpn.xwiki.XWiki` class instead?

acotiuga: Thanks for remarks, I made the changes. Hope it's better now.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1475,6 +1475,7 @@ core.register.welcome=Sign up here so you can edit pages and participate in the
 core.register.passwordMismatch=Passwords are different or password is empty.
 core.register.userAlreadyExists=User already exists.
 core.register.invalidUsername=Invalid username provided. Please use only letters from the latin alphabet, numbers, and the underscore character '_'.
+core.register.mailSenderWronglyConfigured=The user has been created but the validation email has not been sent. Please check the Mail Sending Configuration and consider to recreate the user.
```
vmassol: "Please check the Mail Sending Configuration and consider to recreate the user." --> "Please check the Mail Sending Configuration and consider recreating the user."

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -3396,7 +3396,12 @@ public int createUser(boolean withValidation, String userRights, XWikiContext co
             // Send validation mail, if needed.
             if ((result > 0) && (withValidation)) {
                 // Send the validation email
-                sendValidationEmail(xwikiname, password, email, validkey, "validation_email_content", context);
+                try {
+                    sendValidationEmail(xwikiname, password, email, validkey, "validation_email_content", context);  
+                } catch (XWikiException e) {
+                    LOGGER.error("User created. Failed to send the email to the created user", e);
```
vmassol: This is not an error. ERROR means not recoverable. In this case it's a WARN and you should display a stack trace. You should use ExceptionUtils to get the root message. See other example.

I'm pretty sure I had documented this on dev.xwiki.org but can't find it ATM.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -3396,7 +3396,12 @@ public int createUser(boolean withValidation, String userRights, XWikiContext co
...
+                    return -11;
```
vmassol: Note that this look too magical and it's duplicated in 2 places. Better introduce a constant.

acotiuga:
This pull request is replaced by the #554, because I wasn't able to patch it after I lost my fork. Sorry for the inconvenience.

PR #536, opened 2016-09-19
XWIKI-13714 : Create a script service for the Instance Module

acotiuga:


xwiki-platform-core/xwiki-platform-instance/src/main/java/org/xwiki/instance/script/InstanceScriptService.java
```diff
@@ -0,0 +1,63 @@
...
+import org.xwiki.instance.InstanceId;
+import org.xwiki.instance.InstanceIdManager;
+import org.xwiki.script.service.ScriptService;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provides instance-specific scripting APIs.
+ * 
+ * @version $Id$
+ * @since 8.3-RC1
```
vmassol: we use the format "@since 8.3RC1"

xwiki-platform-core/xwiki-platform-instance/src/main/java/org/xwiki/instance/script/InstanceScriptService.java
```diff
@@ -0,0 +1,63 @@
...
+@Component
+@Named("instance")
+@Singleton
+@Unstable
+public class InstanceScriptService implements ScriptService
+{
+    /**
+     * The key under which the last encountered error is stored in the current execution context.
+     */
+    public static final String INSTANCE_ERROR_KEY = "scriptservice.instance.error";
```
vmassol: This is not needed nor used below and it's no longer a best practice since scripts should now not throw exceptions. See http://markmail.org/message/e77r6zn45teghrmz

Now, I just noticed that I forgot to update http://extensions.xwiki.org/xwiki/bin/view/Extension/Script+Module#HBestPractices which I'll do and close the thread.

xwiki-platform-core/xwiki-platform-instance/src/main/java/org/xwiki/instance/script/InstanceScriptService.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Used to access the instanceId manager.
+     */
+    @Inject
```
vmassol: No need for javadoc here, especially since the comment is not good and doesn't bring value ;)

xwiki-platform-core/xwiki-platform-instance/src/main/java/org/xwiki/instance/script/InstanceScriptService.java
```diff
@@ -0,0 +1,63 @@
...
+    private InstanceIdManager instanceIdManager;
+
+    /**
+     * @return the XWiki's instanceId
+     * @since 8.3-RC1
```
vmassol: should not be since here since it's already at the class level

xwiki-platform-core/xwiki-platform-instance/src/main/java/org/xwiki/instance/script/InstanceScriptService.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final String INSTANCE_ERROR_KEY = "scriptservice.instance.error";
+
+    /**
+     * Used to access the instanceId manager.
+     */
+    @Inject
+    private InstanceIdManager instanceIdManager;
+
+    /**
+     * @return the XWiki's instanceId
```
vmassol: I'd say something like, "@return the XWiki instance id or null if not set, see {@link InstanceIdManager#getInstanceId()}"

PR #537, opened 2016-09-22
XWIKI-13663: Update the User Profile's "Dashboard preferences" label to be in line with its behaviour

msmeria:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3627,7 +3627,7 @@ platform.linkchecker.livetable.date=Last Checked
 
 ### Dashboard Application Resources
 platform.dashboard.user.preferences=Dashboard preferences
-platform.dashboard.user.displayOnMainPage=Display my dashboard on the wiki home when I'm logged in (instead of the default dashboard)
+platform.dashboard.user.displayOnMainPage=Display my dashboard when I'm logged in (instead of the default dashboard)
```
vmassol: hmm I think it can be improved, it looks a bit strange.
What about: "Display my dashboard when viewing the Dashboard application" which is what it does AFAIK?

msmeria: That works too. So "Display my dashboard when viewing the Dashboard application" or "Display my dashboard when viewing the Dashboard application (instead of the default dashboard)"?

vmassol: Well the reason I didn't like your version is because it's not the truth ;) You don't see the dashboard when you're logged in. You only see it when you go to the dashboard app home page... 

Ok if you want to mention it's replacing the default dashboard. I'd say:

"Replace the default dashboard by my custom dashboard"

WDYT?

msmeria: I think "Replace the default dashboard with my custom dashboard" sound better.
WDYT?

vmassol: yes sorry, "with" is better ofc :)

vmassol:
Applied, thanks!

PR #540, opened 2016-09-29
XWIKI-13727 : Ratings app is unusable on XWiki 7.4.5

acotiuga:
* this commit is to fix the same issue for XWiki 8.3
* this could be applied also for XWiki 7.4.5 only if the fix for http://jira.xwiki.org/browse/XWIKI-13274 will be also applied

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
+++ b/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
@@ -446,20 +446,22 @@ position: relative;
     </class>
     <property>
       <content>{{velocity}}
-#set($canView = $services.security.authorization.hasAccess("view", "XWiki.RatingsMacros"))
+#set($canView = $services.security.authorization.hasAccess('view', 'XWiki.RatingsMacros'))
 #if($canView)
   {{include reference='XWiki.RatingsMacros'/}}
-  #set($configurationClass = 'XWiki.RatingsConfigClass')
+#end
+{{/velocity}}
+{{velocity}}
+#if($canView)
   #set($configurationDoc = $services.ratings.getConfigurationDocument($doc.documentReference))
-  #set($configurationObj = $configurationDoc.getObject($configurationClass))
+  #set($configurationObj = $configurationDoc.getObject('XWiki.RatingsConfigClass'))
   ##
   #set($displayRatings = $configurationObj.getProperty('displayRatings').value == 1)
   #set($excludedPages  = $stringtool.split($configurationObj.getProperty('excludedPages').value, ','))
   #set($excludedPages  = $stringtool.stripAll($excludedPages))
   ##
-  ##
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/main/resources/XWiki/Ratings.xml
```diff
@@ -446,20 +446,22 @@ position: relative;
     </class>
     <property>
       <content>{{velocity}}
-#set($canView = $services.security.authorization.hasAccess("view", "XWiki.RatingsMacros"))
+#set($canView = $services.security.authorization.hasAccess('view', 'XWiki.RatingsMacros'))
 #if($canView)
   {{include reference='XWiki.RatingsMacros'/}}
-  #set($configurationClass = 'XWiki.RatingsConfigClass')
+#end
+{{/velocity}}
```
sdumitriu: Why this split?

Enygma2002: It's not intuitive, as I also suggested to @evalica initially to group them, but the order of rendering is important and affected by the 2 approaches.

If they are all inside a single velocity macro, the velocity code executes first, when there is no `displayFullRating` defined and will print it as it is. After the velocity execution, the wiki syntax will be rendered and the `include` macro will be executed and it will define the `displayFullRating` macro, but it is too late by now.

If you separate them, the first velocity macro block renders completely and also includes and defines the `displayFullRating` macro, and, as a result, the separate velocity macro block`s rendering will now properly have the macro defined and can execute it when the velocity execution is done.

All of this is the result of the condition that is set on the `include` wiki macro call, which probably aims to improve performance by not defining useless macros (and whatever else is in that page).

Perhaps another option would have been to group them and, instead of the include wiki macro call, to use the old velocity `#includeMacros` macro, or something along those lines.

sdumitriu: I see, the second macro needs the `include` to be executed first. Is it terrible if the include happens anyway, i.e. move the include macro outside, before the velocity macro?

acotiuga: If the `include` happens anyway, outside the `#if($canView)`, then http://jira.xwiki.org/browse/XWIKI-13040 will occur. Hope this answers to your question.

sdumitriu: Follow up question: Why can a user see `XWiki.Ratings` but not `XWiki.RatingsMacros`? It looks like inconsistent rights to me.

New question... Why are there two files, `rating_macros.vm` and `XWiki.RatingsMacros`? Can't the wiki macros be placed in the template file?  Or the other way around?

PR #541, opened 2016-09-30
Xwiki 13753: wrong encoding page reference in "Share by email" feature

Pbast:


Pbast:
Notice: I didn't use  "mvn xar:format" because I didn't manage to launch it without error: 
"Failed to execute goal on project xwiki-platform-gwt: Could not resolve dependencies for project org.xwiki.platform:xwiki-platform-gwt:pom:8.4-SNAPSHOT: Could not transfer artifact com.google.gwt:gwt-user:jar:2.7.0 from/to xwiki-releases (http://nexus.xwiki.org/nexus/content/groups/public): Failed to transfer file: http://nexus.xwiki.org/nexus/content/groups/public/com/google/gwt/gwt-user/2.7.0/gwt-user-2.7.0.jar. Return code is: 503, ReasonPhrase:Service Unavailable. -> [Help 1]"

xwiki-platform-core/xwiki-platform-sharepage/xwiki-platform-sharepage-ui/src/main/resources/XWiki/SharePage.xml
```diff
@@ -337,7 +337,7 @@ $!{escapetool.xml($message)}
 &lt;/pre&gt;
 #end
 
-&lt;p&gt;To visit this document online, open &lt;a href="${doc.getExternalURL()}"&gt;${doc.getExternalURL()}&lt;/a&gt; in your browser.&lt;/p&gt;
+&lt;p&gt;To visit this document online, open &lt;a href="${doc.getExternalURL()}"&gt;${util.decodeURI("${doc.getExternalURL()}")}&lt;/a&gt; in your browser.&lt;/p&gt;
```
vmassol: hmm I wonder if that's correct. Imagine that we have a page named "my/Page". It would be encoded as "http://.../my%2FPage/..." but if you call util.decodeURI on it, you'll get: "http://.../my/Page/..." which obviously won't be correct...

vmassol: It's not the right way of doing this :) Once the URL is encoded you cannot unencode it... So I think we need to offer an API to not encode the URL, when required.

Pbast: yes I know. Maybe I could check your API monday to add a parameter for that (if I can)

vmassol: it's not easy (it would break backward compatibility)... I think we need to brainstorm it on the devs list. Have a good we!

sdumitriu: I don't understand why it shouldn't be encoded. It's an URL, it should be encoded, otherwise copy/pasting it wouldn't work all the time.

The only case where it would have to be decoded is it is going to be encoded a second time, but that's not the case. So, what's the problem that you're trying to solve?

vmassol: Note that if the problem is showing some technical content to users, it would be easy to replace the URL by some text. For example by replacing the text `To visit this document online, open <url> in your browser` by `You can <visit this document online> in your browser`.

sdumitriu: The URL is there for a reason. Since some email clients may remove/break links, having the full URL available for copy/paste is a common practice. Most such emails use one of the following patterns:

- "click \<a href="link">here\</a> to do something, or copy/paste the following URL in your browser if clicking doesn't work: _link_"
- "open \<a href="link">_link_\</a> to do something"

sdumitriu: Define "encoded"... The whole URL isn't encoded in any way, but parts of it are and should be `%XY` encoded, according to the URL grammar, so that illegal characters aren't used. For example, spaces in the page name, unencoded, will break the URL, so will commas, `<` and `>`, and so on. `doc.get*URL` does proper escaping of the fragments, and there shouldn't be any other escaping as far as I can see. So, I really don't see what kind of decoding is needed.

Pbast: Ok I understand. Like usually encoding problem is not so simple ;-).
Yes the problem is showing some technical content to users: i.e. like this screen:
 https://snag.gy/O9vkfc.jpg
but even it is ugly, it's working fine with url encoded. It is the main purpose.

evalica:
Closing this PR, since its associated issue (https://jira.xwiki.org/browse/XWIKI-13753) is closed as Won't fix.

PR #542, opened 2016-10-11
XWIKI-13786: Accessibility problem with rights management's checkboxes states.

gdelhumeau:
I open this PR because I don't really know if we should re-use the "labels" fields of the `MSCheckbox` class instead of creating the `alts` one.

Anyway, if we decide to display the labels in the rights editor, we would still need to add an "alt" attribute to the images. That is the best practice and the only choice to be accessible and HTML5 valid (even if our validation tool do not detect this error because it's some HTML code generated by javascript).


surli:
Closing as it's been done with #2051

PR #545, opened 2016-10-24
Update syntaxe doc without html

Pbast:


xwiki-platform-core/xwiki-platform-help/xwiki-platform-help-ui/src/main/resources/XWiki/XWikiSyntaxGroups.xml
```diff
@@ -126,7 +126,7 @@
 
 === 2.0 Groups ===
 
-Groups can be used to insert another document or div section inline directly into the current document. This allows for example to insert complex elements or style inside a list item, inside a table cell or a paragraph. Groups are delimited by the following syntactic elements: ##{{{(((...)))}}}##. One Group can contain another Group and there is no limit of imbrication.
+Groups can be used to insert another document (or a different style section) inline directly into the current document. This allows for example to insert complex elements or style inside a list item, inside a table cell or a paragraph. Groups are delimited by the following syntactic elements: ##{{{(((...)))}}}##. One Group can contain another Group and there is no limit of imbrication.
```
vmassol: It's not just about style Pascal :) It's about adding some parameter. What is done with those parameters is independent of the syntax, it's the Renderer that decide what to do with them (the behavior is not defined at the syntax level).

Technically it's all about inserting another document (there's no other use case). What a group means is that inside it you can have the same content as the content you can find in a document.

Not sure how to word it though.

PR #546, opened 2016-11-01
XWIKI-13843: Drop support for IE6

alexhenrie:
Virtually all of this was dead code, since on current browsers window.browser.isIE6x is false and the CSS "* html" does not match any elements. The entire csshover3-source.htc file was an IE6 hack, but it was not actually used anywhere. I did not delete anything in third-party components that XWiki relies on and I left all the IE7 hacks alone.


mflorea:
Thanks!

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.css
```diff
@@ -179,13 +169,6 @@ thead.xwiki-livetable-display-header td.xwiki-livetable-display-header-filter se
 tbody.xwiki-livetable-display-body tr {
   border-bottom: 1px solid $theme.borderColor;
 }
-tbody.xwiki-livetable-display-body tr:hover , tbody.xwiki-livetable-display-body tr.rowHover { /* for IE6 tr:hover fix*/
```
sdumitriu: This had both general and ie6 selectors, the rule should be put back, but only with the first selector.

mflorea: Indeed. I fixed it. Thanks @sdumitriu!

alexhenrie: Thanks for putting back "thead.xwiki-livetable-display-header tr.xwiki-livetable-display-filters:hover" too, and sorry for missing these. Without all the IE6 hacks, the code is much more understandable now. Thank you!

alexhenrie: We should also put back "thead.xwiki-livetable-display-header tr.xwiki-livetable-display-filters:hover" too, sorry for missing these. I've pushed a fix to my branch. (I thought you had fixed both, then realized that you had only fixed the one.)

mflorea: Done. Thanks!

PR #550, opened 2016-11-14
XWIKI-13869: Option to use custom file name for file download

veronikaslc:
Added "force-filename" request parameter to specify the custom file name.


mflorea:
Thanks for your contribution. Note that ``TempResourceAction`` is deprecated. The new way to serve temporary resources is through [``TemporaryResourceReferenceHandler``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-temporary/src/main/java/org/xwiki/resource/temporary/internal/TemporaryResourceReferenceHandler.java). Would be great if you could do a similar change in [``AbstractServletResourceReferenceHandler#setResponseHeaders``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java#L212).

PR #554, opened 2017-02-16
XWIKI-13679  User is still added after error registration has failed due to unknown reasons error code -10

acotiuga:
This pull request is a replacement for #535.


acotiuga:
Thank you!

PR #559, opened 2017-04-11
XWIKI-11818 : Change default value of xwiki.users.initialGroups based on xwiki.authentication.group.allgroupimplicit

krisss2121:


krisss2121:
In hierarchy.js there was some strange white character that caused build error. I replaced it with space.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/api/XWikiGroupService.java
```diff
@@ -44,6 +44,17 @@
...
     /**
+     * Check whether the configuration specifies that every user is implicitly in XWikiAllGroup. Configured by the
+     * {@code xwiki.authentication.group.allgroupimplicit} parameter in {@code xwiki.cfg}.
+     *
+     * @param context the current XWiki context
+     * @return {@code true} if the group is implicit and all users should be by default in it, {@code false} if the
+     *         group behaves as all other groups, containing only the users/subgroups that are explicitly listed inside
+     *         the document.
+     */
+    public boolean isAllGroupImplicit(XWikiContext context);
```
tmortagne: I would do the opposite actually: move isAllGroupImplicit to XWiki as a public method and call it in XWikiGroupServiceImpl#isAllGroupImplicit and XWiki#setUserDefaultGroup. Otherwise you are breaking XWikiGroupService interface and have to provide a default implementation for XWikiGroupService#isAllGroupImplicit method (with "default" keyword).

krisss2121: Ok, I'll do that.
IDE found only one implementation of XWikiGroupService - XWikiGroupServiceImpl so I thought such interface change would be seamless.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiGroupServiceImpl.java
```diff
@@ -195,19 +195,9 @@ public void flushCache()
...
-     * Check whether the configuration specifies that every user is implicitly in XWikiAllGroup. Configured by the
-     * {@code xwiki.authentication.group.allgroupimplicit} parameter in {@code xwiki.cfg}.
-     *
-     * @param context the current XWiki context
-     * @return {@code true} if the group is implicit and all users should be by default in it, {@code false} if the
-     *         group behaves as all other groups, containing only the users/subgroups that are explicitly listed inside
-     *         the document.
-     */
-    protected boolean isAllGroupImplicit(XWikiContext context)
+    private boolean isAllGroupImplicit(XWikiContext context)
```
tmortagne: You should not change the visibility. Since XWikiGroupServiceImpl is a public class you what do do here is the same as removing a method from an interface. If you build with -Pquality you should get an error.

krisss2121: Yes, sure, that was a mistake

tmortagne:
Thanks @krisss2121 ! The only thing missing is a @since on the new method XWiki#isAllGroupImplicit() but I will take care of that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/api/XWikiGroupService.java
sdumitriu: One implementation in the official xwiki-platform code, but XWiki is a framework that anybody can extend with their own modules, and these custom modules can't be refactored easily.

PR #560, opened 2017-04-14
XWIKI-14205 - Search Widget UIExtension point

ldubost:
Extension point added in SolrSearchMacros

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
@@ -824,6 +824,11 @@ $value##
 
     #end
     #displaySearchForm()
+    #foreach ($extension in $services.uix.getExtensions("org.xwiki.platform.search.widget"))
+      {{html clean=false}}
+      $services.rendering.render($extension.execute(), 'xhtml/1.0')
+      {{/html}}
+    #end
     #if ($text != '')
       #getSearchResults()
       #if ($debug)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -824,6 +824,11 @@ $value##
 
     #end
     #displaySearchForm()
+    #foreach ($extension in $services.uix.getExtensions("org.xwiki.platform.search.widget"))
```
vmassol: it seems you're adding a SOLR only extension point and not a generic search one, so I would use "solr" somewhere in the UIXP id. WDYT?

vmassol: Also, what is a search widget? Is it a Facet?

vmassol: Note that the naming is quite important because this is becoming an API that cannot be changed thereafter.

ldubost: So no it's not necessary only a solr extension point. It could be also included on other searches.. but i did not do it on lucene as I cannot test

ldubost: A search widget is something that shows up under the search form, but before anything else. You are welcome to rename the extension point if you think it's not ok.

vmassol: ok, I know this is a probably a pain for you but this needs to be discussed and agreed on the list since it's a new API and we cannot remove it thereafter. We need to decide:
* the scope (SOLR only, all search UIs)
* the name and the concept of "search widgets"
* whether it's one UIXP or 2. In the jira issue you describe 2 needs which are very different to me. You may be able to cover them with one UIXP with the current SOLR UI but an UIXP must work for any UI.

I'll comment in the jira issue. Thanks

ldubost: It's ok. It would be great if you guys can take care of this decision. I will then modify my extension to make it compatible with the integrated uiextension point

PR #563, opened 2017-05-05
XWIKI-14260: Exporting as PDF Level 1 pages, with Header option selected, displays the title of the page 2 times

aubincleme:
Fix bug by displaying the title only once if it’s the same as the space

When a PDF is exported from a page which has the same title as its space name, this code prevents the document title and the document space (which are, in this case, the same) to be printed twice.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pdfheader.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pdfheader.vm
@@ -1,6 +1,10 @@
 #set ($title = "$!{pdfdoc.display('title', 'rendered', '', $pdfobj)}")
 #if ($title == '')
-  $!{escapetool.xml($doc.space)} - $!{escapetool.xml($tdoc.plainTitle)}
+  #if ($doc.space != $tdoc.plainTitle)
+    $!{escapetool.xml($doc.space)} - $!{escapetool.xml($tdoc.plainTitle)}
+  #else
+    $!{escapetool.xml($doc.space)}
+  #end
 #else
   $escapetool.xml($title)
 #end


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pdfheader.vm
```diff
@@ -1,6 +1,8 @@
 #set ($title = "$!{pdfdoc.display('title', 'rendered', '', $pdfobj)}")
-#if ($title == '')
+#if ($title == '' && $doc.space != $tdoc.plainTitle)
   $!{escapetool.xml($doc.space)} - $!{escapetool.xml($tdoc.plainTitle)}
+#elseif ($title == '')
+  $!{escapetool.xml($tdoc.plainTitle)}
 #else
```
vmassol: Thanks for the PR!

I don't think I understand the new condition (maybe you could add some explanation?).

However the #if could be improved slightly with:

````
#if (title == '')
  #if (new condition)
    ...
  #else
   ...
#end
````

That should save an evaluation ($title == '') when the title is empty and the new condition is not satisfied.

Comment (aubincleme) [inline xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pdfheader.vm]: Fixed.
I will edit the PR description in order to add more details about this issue.

File: xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pdfheader.vm [hunk 2]
```diff
@@ -1,6 +1,10 @@
 #set ($title = "$!{pdfdoc.display('title', 'rendered', '', $pdfobj)}")
 #if ($title == '')
-  $!{escapetool.xml($doc.space)} - $!{escapetool.xml($tdoc.plainTitle)}
+  #if ($doc.space != $tdoc.plainTitle)
+    $!{escapetool.xml($doc.space)} - $!{escapetool.xml($tdoc.plainTitle)}
+  #else
+    $!{escapetool.xml($tdoc.plainTitle)}
```
vmassol: I'd use the reference in this case and not the title.

vmassol: A small improvement could also be to display the full reference, see XWIKI-14260

aubincleme: Well, that’s fixed now

PR #567, opened 2017-05-16
XWIKI-10122 and XWIKI-11084: Livetable date filter

pbondoer:
Adds pretty [date range pickers](http://daterangepicker.com) to filter dates in livetables.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
@@ -34,8 +34,7 @@
  * Displays a submit button, i.e. an <tt>input</tt> element with <tt>type=submit</tt>.
  * For styling purposes, the button is wrapped in a span element.
  * @param name The value of the <tt>name</tt> attribute
- * @param shortcut The keyboard shortcut for this button. The value will be displayed
- *        as the title of the element. The actual shortcut is attached elsewhere, in
+ * @param shortcut The keyboard shortcut for this button. The value will be displayed as the title of the element. The actual shortcut is attached elsewhere, in
  *        the javascript code.
  * @param value The value of the <tt>value</tt> attribute, i.e. the text displayed on
  *        the button.
@@ -306,6 +305,13 @@
                 </td>
               #elseif(("$!colprop.type" == 'text' || "$!colprop.type" == 'number') && $isFilterable)
                 <td class="xwiki-livetable-display-header-filter"><input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="text" #if("$!colprop.size" != '') size="${colprop.size}" title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle', [$services.localization.render("${transprefix}${colname}")]))"#end /></td>
+              #elseif("$!colprop.type" == 'date')
+                #set($ok = $xwiki.jsfx.use('js/xwiki/table/livetabledate.js', true))
+                <td class="xwiki-livetable-display-header-filter">
+                  ## Hidden input for date fields allows us to send timestamps to the server while displaying pretty dates
+                  <input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="hidden" aria-hidden="true" />
+                  <input type="text" data-type="date" #if("$!colprop.size" != '') size="${colprop.size}" title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle', [$services.localization.render("${transprefix}${colname}")]))"#end />
+                </td>
               #elseif("$!colprop.type" == 'hidden')
               #elseif($colname != '_actions' || $hasAdmin)
               <td class="xwiki-livetable-display-header-filter"></td>
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
@@ -317,7 +317,8 @@
                 <td class="xwiki-livetable-display-header-filter">
                   ## Hidden input for date fields allows us to send timestamps to the server while displaying pretty dates
                   <input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="hidden" />
-                  <input type="text" data-type="date" #if("$!colprop.size" != '') size="${colprop.size}"
+                  <input type="text" data-type="date" data-dateformat='$escapetool.html($xwiki.getXWikiPreference("dateformat",
+                  "yyyy/MM/dd HH:mm"))' #if("$!colprop.size" != '') size="${colprop.size}"
                   title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle',
                   [$services.localization.render("${transprefix}${colname}")]))"#end />
                 </td>
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
@@ -305,13 +305,21 @@
                   </select>
                 </td>
               #elseif(("$!colprop.type" == 'text' || "$!colprop.type" == 'number') && $isFilterable)
-                <td class="xwiki-livetable-display-header-filter"><input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="text" #if("$!colprop.size" != '') size="${colprop.size}" title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle', [$services.localization.render("${transprefix}${colname}")]))"#end /></td>
+                <td class="xwiki-livetable-display-header-filter">
+                  <input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="text"
+                  #if("$!colprop.size" != '') size="${colprop.size}"
+                  title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle',
+                  [$services.localization.render("${transprefix}${colname}")]))"#end />
+                </td>
               #elseif("$!colprop.type" == 'date')
                 #set($ok = $xwiki.jsfx.use('js/xwiki/table/livetabledate.js', true))
+                #set($ok = $xwiki.linkx.use($services.webjars.url('bootstrap-daterangepicker', 'css/bootstrap-daterangepicker.css'), {'type': 'text/css', 'rel': 'stylesheet'}))
                 <td class="xwiki-livetable-display-header-filter">
                   ## Hidden input for date fields allows us to send timestamps to the server while displaying pretty dates
-                  <input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="hidden" aria-hidden="true" />
-                  <input type="text" data-type="date" #if("$!colprop.size" != '') size="${colprop.size}" title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle', [$services.localization.render("${transprefix}${colname}")]))"#end />
+                  <input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="hidden" />
+                  <input type="text" data-type="date" #if("$!colprop.size" != '') size="${colprop.size}"
+                  title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle',
+                  [$services.localization.render("${transprefix}${colname}")]))"#end />
                 </td>
               #elseif("$!colprop.type" == 'hidden')
               #elseif($colname != '_actions' || $hasAdmin)
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
@@ -34,7 +34,8 @@
  * Displays a submit button, i.e. an <tt>input</tt> element with <tt>type=submit</tt>.
  * For styling purposes, the button is wrapped in a span element.
  * @param name The value of the <tt>name</tt> attribute
- * @param shortcut The keyboard shortcut for this button. The value will be displayed as the title of the element. The actual shortcut is attached elsewhere, in
+ * @param shortcut The keyboard shortcut for this button. The value will be displayed
+ *        as the title of the element. The actual shortcut is attached elsewhere, in
  *        the javascript code.
  * @param value The value of the <tt>value</tt> attribute, i.e. the text displayed on
  *        the button.
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -710,14 +710,14 @@
     #set ($dateProperty = "${safe_tableAlias}.value")
   #end
   #set ($timestamps = $filterValue.split('-'))
-  #if ($filterValue.contains('-') &amp;&amp; $timestamps.size() == 2)
+  #if ($timestamps.size() == 2)
     #set ($whereSql = "${whereSql} and $dateProperty >= ? and $dateProperty <= ?")
     #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[0]))))
     #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[1]))))
   #else
     ## Fall back to the older string matching
-    #set ($whereSql = "${whereSql} and upper($dateProperty) like upper(?)")
-    #set ($discard = $whereParams.add($filterValue))
+    #set ($whereSql = "${whereSql} and upper(str($dateProperty)) like upper(?)")
+    #set ($discard = $whereParams.add("%$filterValue%"))
   #end
 #end
 
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -609,6 +609,9 @@
           #set ($discard = $whereParams.add("%${locationFilterValue}%"))
           ## We add the parameter twice because we use it twice in the WHERE clause (it's not a named parameter).
           #set ($discard = $whereParams.add("%${locationFilterValue}%"))
+        #elseif ($colName == 'doc.date')
+          #livetable_getTableAlias($colname)
+          #livetable_filterDateProperty()
         #else
           #set ($safeColName = $colName.replaceAll('[^a-zA-Z0-9_.]', '').replace('_', '.'))
           #set ($whereSql = "${whereSql} and upper(str($safeColName)) like upper(?)")
@@ -701,9 +704,15 @@
  * NOTE: This macro uses variables defined in livetable_filterProperty . It was not meant to be used alone.
  *#
 #macro (livetable_filterDateProperty)
-  ## TODO: The date filter should be smarter than this..
-  #set ($whereSql = "${whereSql} and upper(str(${safe_tableAlias}.value)) like upper(?)")
-  #set ($discard = $whereParams.add("%$filterValue%"))
+  #if ($safe_tableAlias.startsWith('doc.'))
+    #set ($dateProperty = $safe_tableAlias)
+  #else
+    #set ($dateProperty = "${safe_tableAlias}.value")
+  #end
+  #set ($timestamps = $filterValue.split('-'))
+  #set ($whereSql = "${whereSql} and $dateProperty >= ? and $dateProperty <= ?")
+  #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[0]))))
+  #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[1]))))
 #end
 
 
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -710,9 +710,15 @@
     #set ($dateProperty = "${safe_tableAlias}.value")
   #end
   #set ($timestamps = $filterValue.split('-'))
-  #set ($whereSql = "${whereSql} and $dateProperty >= ? and $dateProperty <= ?")
-  #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[0]))))
-  #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[1]))))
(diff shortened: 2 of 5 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+        });
+
+        var updateInput = function(elem, ev, picker) {
+            if (ev.type == 'cancel')
+            {
+                input.val('');
+                hidden.val('');
+            } else {
+                if (picker.startDate.isSame(picker.endDate, 'day'))
+                    input.val(picker.startDate.format('DD/MM/YYYY'));
```
sdumitriu: This should use either the globally configured date format, or, if this is going to be used for regular xproperties, the format set for that xproperty. Nobody in USA will understand this hardcoded format, will they?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+                else
+                    input.val(picker.startDate.format('DD/MM/YYYY') + ' - ' + picker.endDate.format('DD/MM/YYYY'));
+                hidden.val(picker.startDate.format('x') + '-' + picker.endDate.format('x'));
```
sdumitriu: There's no code in the actual query to handle this kind of input, is there?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
sdumitriu: Are you sure `DD/MM/YYYY` works? I don't know what format options the daterangepicker library accepts, but in Java it's case sensitive, with `D` meaning "day in year" and `Y` meaning "week year", so instead of `31/12/2015` it will print `365/12/2016`.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -306,6 +306,13 @@
                 </td>
               #elseif(("$!colprop.type" == 'text' || "$!colprop.type" == 'number') && $isFilterable)
                 <td class="xwiki-livetable-display-header-filter"><input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="text" #if("$!colprop.size" != '') size="${colprop.size}" title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle', [$services.localization.render("${transprefix}${colname}")]))"#end /></td>
+              #elseif("$!colprop.type" == 'date')
+                #set($ok = $xwiki.jsfx.use('js/xwiki/table/livetabledate.js', true))
+                <td class="xwiki-livetable-display-header-filter">
+                  ## Hidden input for date fields allows us to send timestamps to the server while displaying pretty dates
+                  <input id="xwiki-livetable-${htmlLiveTableId}-filter-${velocityCount}" name="${colname}" type="hidden" aria-hidden="true" />
```
mflorea: I think ``aria-hidden="true"`` is redundant for hidden inputs, which are not hidden by style. They are inherently hidden. You can't make them visible.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
pbondoer: Moment.js formats and Java formats are different: https://momentjs.com/docs/#/displaying/format/

I haven't figured out a way to use the configured date format yet, we'd need some Javascript code that can do that. I know it's possible in the backend, but I haven't found a way to do it client-side.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -306,6 +306,13 @@
...
+                  <input type="text" data-type="date" #if("$!colprop.size" != '') size="${colprop.size}" title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle', [$services.localization.render("${transprefix}${colname}")]))"#end />
```
mflorea: For improved readability lines should not exceed 120 chars.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
pbondoer: Oh, you're right! I messed up, didn't add my changes to the commit for some reason...

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+        'daterangepicker': ['jquery', 'bootstrap', 'moment']
+    }
+});
+
+require(['jquery', 'moment', 'daterangepicker', 'xwiki-events-bridge'],
+function($, moment) {
+
+    // Load bootstrap-daterangepicker CSS only when needed
+    // This is a bit of a hack, but it's required until we have an API to
+    // include CSS from a webjar in Velocity.
```
mflorea: We actually have the API, see https://github.com/xwiki-contrib/application-ckeditor/blob/master/ui/src/main/resources/CKEditor/VelocityMacros.xml#L105 .

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+    $('head').append('<link rel="stylesheet" type="text/css" href="$services.webjars.url('bootstrap-daterangepicker', 'css/bootstrap-daterangepicker.css')">');
```
mflorea: A general remark: when you mix Velocity with JavaScript you need to make sure the result is still a valid JavaScript code. In the line above, as you can see from the syntax highlighting, the apostrophe opened in JavaScript is closed unintentionally in Velocity. Have you tried to build ``xwiki-platform-web``? My guess is that the build will fail because this file can't be minified.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+            autoUpdateInput: false,
+            timePicker: true,
+            timePicker24Hour: true,
+            ranges: {
+                '$escapetool.javascript($services.localization.render("daterange.today"))': [moment().startOf('day'), moment().endOf('day')],
+                '$escapetool.javascript($services.localization.render("daterange.yesterday"))': [moment().subtract(1, 'days').startOf('day'), moment().subtract(1, 'days').endOf('day')],
+                '$escapetool.javascript($services.localization.render("daterange.lastSevenDays"))': [moment().subtract(6, 'days').startOf('day'), moment().endOf('day')],
+                '$escapetool.javascript($services.localization.render("daterange.lastThirtyDays"))': [moment().subtract(29, 'days').startOf('day'), moment().endOf('day')],
+                '$escapetool.javascript($services.localization.render("daterange.thisMonth"))': [moment().startOf('month'), moment().endOf('month')],
+                '$escapetool.javascript($services.localization.render("daterange.lastMonth"))': [moment().subtract(1, 'month').startOf('month'), moment().subtract(1, 'month').endOf('month')]
```
mflorea: Again, lines are too long.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+            },
+            locale: {
+                cancelLabel: '$escapetool.javascript($services.localization.render("daterange.clear"))',
+                applyLabel: '$escapetool.javascript($services.localization.render("daterange.apply"))'
+            }
+        });
+
+        var updateInput = function(elem, ev, picker) {
```
mflorea: I prefer explicit variable names. E.g. ``event`` instead of ``ev``.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+            if (ev.type == 'cancel')
+            {
```
mflorea: I don't think we have a formal JavaScript code style, but you should at least try to be consistent across your code. I prefer to keep the ``{`` on the same line as the ``if``. If you put it on a new line then why doesn't this apply also to the following ``else``?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
mflorea: I agree with @sdumitriu . I would pass the date format from ``macros.vm`` in a ``data-dateFormat`` attribute of the filter text input.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
mflorea: @sdumitriu there should be some changes to the ``LivetableResultsMacros`` page. @pbondoer you didn't include those changes?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+                input.val('');
+                hidden.val('');
+            } else {
+                if (picker.startDate.isSame(picker.endDate, 'day'))
```
mflorea: Bad practice: you should **always** wrap the ``if`` body with ``{...}``, even when there is only one statement.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+                    input.val(picker.startDate.format('DD/MM/YYYY'));
+                else
```
mflorea: Same. The body of the ``else`` needs to be wrapped in ``{...}``. Don't rely on the formatting.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        input.on('apply.daterangepicker cancel.daterangepicker', function(ev, picker) {
+            updateInput(this, ev, picker);
+        });
+
+        input.on('hide.daterangepicker', function(event) {
+            // Overwrite at instance level the 'hide' function added by
+            // Prototype.js to the Element prototype. This removes the 'hide'
+            // function only for the event target.
```
mflorea: Line wrapping should be done at 120 chars.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
sdumitriu: Plus a conversion between the two formats, translate between [Java](http://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) and [Moment](https://momentjs.com/docs/#/displaying/format/).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
pbondoer: Yup, I agree. Just checked the specification, type="hidden" hides it from the accessibility tree in all browsers so it won't be shown no matter what.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
pbondoer: Oh thank you, I'll change it :+1:

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -701,9 +704,15 @@
...
-  #set ($discard = $whereParams.add("%$filterValue%"))
+  #if ($safe_tableAlias.startsWith('doc.'))
+    #set ($dateProperty = $safe_tableAlias)
+  #else
+    #set ($dateProperty = "${safe_tableAlias}.value")
+  #end
+  #set ($timestamps = $filterValue.split('-'))
+  #set ($whereSql = "${whereSql} and $dateProperty >= ? and $dateProperty <= ?")
+  #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[0]))))
+  #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[1]))))
```
mflorea: Can you express "before X" or "after X" with the date range picker? The code above assumes that you always have a closed range, where both ends are specified. I think it makes sense to support also the case where one end is missing. Also, it wouldn't hurt to fall back on the string matching if the filter value doesn't have the expected range format. This is more for bulletproofing. You probably get an exception right now if the filter value is "foo".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,93 @@
...
+  },
+  shim: {
+    'moment': ['jdateformatparser'],
+    'daterangepicker': ['jquery', 'bootstrap', 'moment']
+  }
+});
+
+require(['jquery', 'moment', 'daterangepicker'],
+  function($, moment) {
+    var dateFormat = moment().toMomentFormatString('$escapetool.javascript($xwiki.getXWikiPreference("dateformat", "yyyy/MM/dd HH:mm"))');
```
mflorea: I would rather read the date format from an attribute of the input field that is enhanced by the date range picker. This means calling ``getXWikiPreference()`` in ``macros.vm``. Moreover, Date xproperties can overwrite the date format specified in XWiki preferences and it's easier to get the date format from the xclass in ``macros.vm`` rather than in JavaScript.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,93 @@
...
+            hidden.val('');
+          } else {
+            if (picker.startDate.isSame(picker.endDate, 'day')) {
+              input.val(picker.startDate.format(dateFormat));
+            } else {
+              input.val(picker.startDate.format(dateFormat) + ' - ' + picker.endDate.format(dateFormat));
+            }
+            hidden.val(picker.startDate.format('x') + '-' + picker.endDate.format('x'));
+          }
+          if ('dispatchEvent' in element) {
```
mflorea: Why don't you use jQuery to trigger the event cross-browser?

pbondoer: My bad! A bit of an oversight there :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
pbondoer: So something like
```html
<input ... data-dateformat='$escapetool.html($xwiki.getXWikiPreference("dateformat", "yyyy/MM/dd HH:mm"))' ... >
```
in macros.vm instead?

mflorea: Yep

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
pbondoer: Technically, no, but the daterangepicker control I'm using right now doesn't support that - not sure how to go about this without adding to the control. A checkbox to signify open-ended?

As for the bulletproofing, just a simple `$filterValue.contains('-')`, `$timestamps.count == 2` and `$numbertoo.isNumber($timestamp[0])` should do!

mflorea: I found https://github.com/dangrossman/bootstrap-daterangepicker/issues/886 .

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
pbondoer: Hmm, actually just doing `$(element).trigger('change');` doesn't actually do anything... I'm not sure what's going on.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -701,9 +704,21 @@
...
+  #end
+  #set ($timestamps = $filterValue.split('-'))
+  #if ($filterValue.contains('-') &amp;&amp; $timestamps.size() == 2)
+    #set ($whereSql = "${whereSql} and $dateProperty >= ? and $dateProperty <= ?")
+    #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[0]))))
+    #set ($discard = $whereParams.add($datetool.toDate($numbertool.toNumber($timestamps[1]))))
+  #else
+    ## Fall back to the older string matching
+    #set ($whereSql = "${whereSql} and upper($dateProperty) like upper(?)")
+    #set ($discard = $whereParams.add($filterValue))
```
mflorea: These two lines look different than the lines you deleted. Any reason for this?

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -701,9 +704,21 @@
...
-  ## TODO: The date filter should be smarter than this..
-  #set ($whereSql = "${whereSql} and upper(str(${safe_tableAlias}.value)) like upper(?)")
-  #set ($discard = $whereParams.add("%$filterValue%"))
+  #if ($safe_tableAlias.startsWith('doc.'))
+    #set ($dateProperty = $safe_tableAlias)
+  #else
+    #set ($dateProperty = "${safe_tableAlias}.value")
+  #end
+  #set ($timestamps = $filterValue.split('-'))
+  #if ($filterValue.contains('-') &amp;&amp; $timestamps.size() == 2)
```
mflorea: The first condition is implied by the second so it's enough to use the second condition no?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
mflorea: You're sending an event from jQuery and the listener is in Prototype.js. See http://platform.xwiki.org/xwiki/bin/view/DevGuide/JavaScriptAPI#HBridgingcustomXWikieventsbetweenPrototypeandjQuery .

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
pbondoer: I forgot to add the percentages for the `like`. `$dateProperty` is because we want to support two cases.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
pbondoer: Doesn't the event-bridge only do `xwiki:*` events? Not only that, but `change` is a standard DOM event, and `trigger()` is supposed to send those as well, which prototype picks up? Or am I missing something?

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
mflorea: ``str(...)`` is also important.

pbondoer: Yes, thank you 😱

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
mflorea: Indeed, it's not related to the event bridge. Still, ``$(element).trigger('change')`` should work. You need to debug to see why the listener is not called.

pbondoer: After some debugging and searching I found this:
https://github.com/kangax/protolicious/blob/5b56fdafcd7d7662c9d648534225039b2e78e371/event.simulate.js

It's an incompatibility with Prototype. This code seems to do the same as what I originally wrote. Most browsers support `dispatchEvent` except for IE8 and really old versions of other browsers, for which there is `fireEvent`.

mflorea: I found http://stackoverflow.com/questions/5351442/jquery-change-not-firing-prototype-event-handlers . So I would trigger the event both ways:

    // Trigger Prototype.js listeners because the live table is written using Prototype.js
    element.fire('change');
    // Trigger jQuery listeners.
    $(element).change();

pbondoer: This doesn't work either :(
The accepted answer to that is using the same method as I am though, as we found the same thread.

mflorea: First, I think you need to add a comment to explain why you need to trigger the "change" event. I suppose it's because you want to update the live table rows. Now, if I look at https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js#L999 I see that the livetable is using ``Event.observe`` to register its listener, so I still don't understand why calling ``element.fire('change')`` or ``Event.fire(element, 'change')`` doesn't call call this listener. If you don't want to debug then there's still a better way to trigger the refresh of the live table, see https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js#L1008 (i.e. trigger the custom event). Since this is a custom event, unlike "change", you could even trigger it from jQuery, if you add ``xwiki-event-bridge`` as a dependency to your Require.js module. WDYT?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -0,0 +1,90 @@
...
+});
+
+require(['jquery', 'moment', 'daterangepicker', 'xwiki-events-bridge'],
+  function($, moment) {
+    var bindInputs = function(bind) {
+      $(bind).find('input[data-type="date"]').each(function(i, element) {
+        var input = $(element);
+        var hidden = input.prev('input[type="hidden"]');
+        var dateFormat = moment().toMomentFormatString(input.attr('data-dateformat'));
+        var livetable = input.closest('.xwiki-livetable');
```
mflorea: I think you already have a reference to the live table element in ``bind`` variable above. Depends whether you think ``bindInputs`` might be called on something other than a live table. Your choice.

PR #568, opened 2017-05-16
XWIKI-14138 : Allow proxying the versions metadata

krisss2121:
After many efforts and obstacles - finally!

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesinline.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesinline.vm
@@ -10,7 +10,7 @@
   #set($origdoc = $xwiki.getDocument($tdoc.documentReferenceWithLocale, $rev1))
 #end
 #if ($tdoc.version == $rev2)
-  #set($newdoc = $toc)
+  #set($newdoc = $tdoc)
 #else
   ## Search for the revision
   #set($newdoc = $xwiki.getDocument($tdoc.documentReferenceWithLocale, $rev2))


tmortagne:
There is a few code formatting issues, you should probably look at http://dev.xwiki.org/xwiki/bin/view/Community/JavaCodeStyle#HConfiguringyourIDEtousetheXWikicodestyle and let you IDE do a pass of formatting.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
```diff
@@ -707,6 +735,15 @@ public DocumentReference importExtension(String extensionId, ExtensionRepository
         return document.getDocumentReference();
     }
 
+    public boolean isVersionProxyingEnabled(XWikiDocument extensionDocument)
+    {
+        BaseObject extensionProxyObject = extensionDocument.getXObject(XWikiRepositoryModel.EXTENSIONPROXY_CLASSREFERENCE);
+        if(extensionProxyObject == null){
+            return false;
+        }
+        return "version".equals(getValue(extensionProxyObject, XWikiRepositoryModel.PROP_PROXY_PROXYLEVEL, (String) null));
```
tmortagne: Would be nice to put "version" in a field of XWikiRepositoryModel along with the other possible values.

Also something like "history" might be a better name since the last version is not proxied, WDYT ?

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
```diff
@@ -996,9 +1033,66 @@ private boolean updateExtensionVersionDependencies(XWikiDocument document, Exten
...
+        if(extensionProxyObject == null){
+            return null;
+        }
+        String repositoryId = getValue(extensionProxyObject, XWikiRepositoryModel.PROP_PROXY_REPOSITORYID, (String) null);
+
+        if(extensionId == null || repositoryId == null){
+            return null;
+        }
+
+        ExtensionRepository repository = this.extensionRepositoryManager.getRepository(repositoryId);
```
tmortagne: Not sure what you do all this, the id of the version is stored in the version object itself so you don't need to search for it.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
```diff
@@ -1102,4 +1198,38 @@ private boolean updateCollection(BaseObject extensionObject, String fieldName, C
...
+            } else {
+                return objects.get(objects.size() - 1);
+            }
+        }
+
+        BaseObject extensionVersionObject =
+                extensionDocument.getObject(XWikiRepositoryModel.EXTENSIONVERSION_CLASSNAME, "version", version, false);
+
+        if( extensionVersionObject == null && allowProxying && isVersionProxyingEnabled(extensionDocument) ){
+             //no ExtensionVersionClass object so we need to create such object temporarily
```
tmortagne: As I said in a previous comment I really don't like this hack.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-test/xwiki-platform-repository-test-utils/src/main/java/org/xwiki/repository/test/RepositoryTestUtils.java
```diff
@@ -21,12 +21,9 @@
 
 import java.io.File;
 import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
tmortagne: Please don't use wildcard imports.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-test/xwiki-platform-repository-test-utils/src/main/java/org/xwiki/repository/test/RepositoryTestUtils.java
```diff
@@ -319,4 +317,10 @@ public void waitUntilReady() throws Exception
         // Make sure Solr queue is empty
         this.solrUtils.waitEmpyQueue();
     }
+
+    public ObjectEditPage gotoExtensionObjectsEditPage(String extensionName) {
+        LocalDocumentReference extensionPageReference = getExtensionPageReference(extensionName);
+        testUtils.gotoPage(extensionPageReference, "edit", ImmutableMap.of("editor","object"));
```
tmortagne: No need to use Guava. Collections.singletonMap("editor", "object") should have the same effect.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
tmortagne: As I said in a previous comment I really don't like this hack.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
tmortagne: Not sure why you do all this, the id of the version is stored in the version object itself so you don't need to search for it in repositories again (which is very expensive).

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: totally ok for me :)

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-test/xwiki-platform-repository-test-utils/src/main/java/org/xwiki/repository/test/RepositoryTestUtils.java
krisss2121: ok
The same with static imports? (I mean - I should not use static imports?)

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-test/xwiki-platform-repository-test-utils/src/main/java/org/xwiki/repository/test/RepositoryTestUtils.java
krisss2121: ok

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: The current (latest version) object (ExtensionClass) nor related version and dependency objects (ExtensionVersionClass, ExtensionDependencyClass) nor proxy object (ExtensionProxyClass) don't store the info about the previous versions. In other words - to figure out what are the historic versions of given extension, it's required to ask remote repository - what I do in line 1053.

What I can optimize (and will be very useful optimization) is to check if requested version is current version. If so, all data is cached in backing xObjects of extensionDocument and no need to call remote repository.

Does it make sense, what I wrote?

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: I applied modification to previously discussed code, that I save document after adding version object to it.
The case is - it's the easiest way, making use of existing and working code.
The biggest minus is that - after requesting previous version of extension, it data will be kept in object behind page until the extension is not updated. Maybe it's not such a big drawback? Especially, taking into consideration, that if one of previous version is requested once, it's quite high probability it will be requested again. 
WDYT?

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-test/xwiki-platform-repository-test-utils/src/main/java/org/xwiki/repository/test/RepositoryTestUtils.java
tmortagne: static imports are fine. It's just that the rule is to not import more than needed.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
tmortagne: So resolveExtensionVersion method is called only if there is no object for this version in the  document and the history proxy is enabled ?

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
```diff
@@ -474,24 +488,27 @@ public void validateExtensions() throws QueryException, XWikiException
...
-    {
-        // Has a version
-        String extensionVersion = getValue(extensionVersionObject, XWikiRepositoryModel.PROP_VERSION_VERSION);
+    public ResourceReference getDownloadReference(XWikiDocument document, String extensionVersion) throws ResolveException {
 
-        // The download reference seems ok
-        String download = getValue(extensionVersionObject, XWikiRepositoryModel.PROP_VERSION_DOWNLOAD);
+        String downloadURL = null;
+
+        BaseObject extensionVersionObject = getExtensionVersionObject(document, extensionVersion, false);
```
tmortagne: Both getExtensionVersionObject and resolveExtensionDownloadURL are dealing with proxied versions from what I understand so you seems to have some duplicates here. As I indicated in the other comment I would really prefer that getExtensionVersionObject only deal with actually existing object.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: yes - I mean - after adding the optimization (described in my comment above) the behavior will be exactly like that

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: Ok - I reinvestigate it all again - to assure that accessing remote repositories happens only when critically necessary
;)

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
tmortagne: > Maybe it's not such a big drawback?

Having a document which is different from the cache and the database can cause all sorts of issues. Here is an example: you do a REST request which add an object in the XWikiDocument, if you then call import it will try to delete that object (since it's supposed to proxy this version) that does and will fail to save. Another issue if that you if someone modify the document for any reason (add some documentation) it will actually save the object in the database and the history won't make any sense ("hey I did not added this object !"), etc.

At the very least you should clone the document before adding an object it and put a FIXME comment to improve this hack with something cleaner in the future.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: I checked it all - and it already worked optimally. I wrote some javadoc comments for bigger clarity

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: The idea that I came up with is to - (reusing existing code) add version xObject to document, save document (like it's done so far), then delete this object and save document again. In this way we'll prevent such cases that - this version object is misused whilst manual edit of extension document, or some cache'ing problems. Of course, from the other side - the time for REST access of historic version of extension is getting longer - but I strongly hope it's acceptable. Proxying will regard only previous versions and typically less important extensions - so a bit of delay in such cases is sth we can deal with - I guess.
Let's call it temporary solution - when I get more intuition of how this document, xobject and caching things work, I will rethink it, ok?

krisss2121:
Sooooo much sorrry for this formatting isssue. I imported XWiki code style, and too eagerly formated whole document. If it'll be too big misery to cope with this huge formating change - let me know - I'll try to revert it somehow

tmortagne:
Which IDE did you use ? With Eclipse I don't get all those changes when formatting.

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
tmortagne: This is even worst. Saving the document takes time so big -1 for this solution. Please look at my suggestion:

> At the very least you should clone the document before adding an object it and put a FIXME comment to improve this hack with something cleaner in the future.

krisss2121:
I use IntelliJ 2016 - I write down to investigate later what might be wrong in this code style xml. For now I will format only new pieces of code

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
krisss2121: So: when theres is need to add extensionVersionObject to document - I do not add it to base doc, but i create a clone of id e.g.: "Extension.<extensionId>.Clone" and when saving it I add comment: "FIXME"
Right?
(sorry for delay in reaction to your comment - I'm still fighting with some projects on studies, but from the begging of June it should be ok ;) )

tmortagne:
Ok, indeed I think there is some differences between Eclipse and IntelliJ formatters and this code was formatted with Eclipse (since I'm the only one who worked on it).

xwiki-platform-core/xwiki-platform-repository/xwiki-platform-repository-server-api/src/main/java/org/xwiki/repository/internal/RepositoryManager.java
tmortagne: No, again you should not save any document because it's super slow. I'm only talking about extensionDocument.clone() here so that you don't break the XWikiDocument instance coming from the cache since that's the issue here.

krisss2121: Aaaaaa okey! And I add this object to this clone and do not save it?
Where I put comment "FIXME"?

tmortagne: Here, next to your comment explaining the hack you are doing (adding the object to the document because it does not exist).

krisss2121: you mean comment in source code?

tmortagne: Yes explain in Java that this code could need some refactoring to be cleaner.

PR #569, opened 2017-05-18
XWIKI-13520: Use bootstrap-select for multilists in livetable

pbondoer:
Use a nice multiselect picker and be able to filter multilists.

Screenshot:
Multiselect view

JIRA:
https://jira.xwiki.org/browse/XWIKI-13520

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -739,8 +739,11 @@
  * NOTE: This macro uses variables defined in livetable_filterProperty . It was not meant to be used alone.
  *#
 #macro (livetable_filterStringListProperty)
-  #set ($whereSql = "${whereSql} and concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) like upper(?)")
-  #set ($discard = $whereParams.add("%|$filterValue|%"))
+  #set ($filterValues = $request.getParameterValues($colname))
+  #foreach ($filterValue in $filterValues)
+    #set ($whereSql = "${whereSql} and concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) like upper(?)")
+    #set ($discard = $whereParams.add("%|$filterValue|%"))
+  #end
 #end
 
 
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -744,7 +744,7 @@
   #if ($joinOperator != 'AND' &amp;&amp; $joinOperator != 'OR')
     #set ($joinOperator = 'AND')
   #end
-  #set ($filterQuery = "concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) LIKE upper(?)")
+  #set ($filterQuery = " concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) LIKE upper(?) ")
   #set ($constraints = $stringtool.repeat($filterQuery, $joinOperator, $filterValues.size()))
   #set ($whereSql = "${whereSql} and ( $constraints )")
   #foreach ($filterValue in $filterValues)
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -740,8 +740,14 @@
  *#
 #macro (livetable_filterStringListProperty)
   #set ($filterValues = $request.getParameterValues($colname))
+  #set ($joinOperator = "$!{request.get(""${colname}${joinModeMarker}"").toUpperCase()}")
+  #if ($joinOperator != 'AND' &amp;&amp; $joinOperator != 'OR')
+    #set ($joinOperator = 'AND')
+  #end
+  #set ($filterQuery = "concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) LIKE upper(?)")
+  #set ($constraints = $stringtool.repeat($filterQuery, $joinOperator, $filterValues.size()))
+  #set ($whereSql = "${whereSql} and ( $constraints )")
   #foreach ($filterValue in $filterValues)
-    #set ($whereSql = "${whereSql} and concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) like upper(?)")
     #set ($discard = $whereParams.add("%|$filterValue|%"))
   #end
 #end
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.css
@@ -167,7 +167,6 @@ thead.xwiki-livetable-display-header td.xwiki-livetable-display-header-filter se
   color: $theme.textPrimaryColor;
   margin: 0;
   padding: 0 0 0 2px;
-  margin-top: 2px;
 }
 .xwiki-livetable-filter-active {
   background-color: $theme.highlightColor !important;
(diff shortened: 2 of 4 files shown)


vmassol:
nice :)

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -739,8 +739,11 @@
  * NOTE: This macro uses variables defined in livetable_filterProperty . It was not meant to be used alone.
  *#
 #macro (livetable_filterStringListProperty)
-  #set ($whereSql = "${whereSql} and concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) like upper(?)")
-  #set ($discard = $whereParams.add("%|$filterValue|%"))
+  #set ($filterValues = $request.getParameterValues($colname))
+  #foreach ($filterValue in $filterValues)
+    #set ($whereSql = "${whereSql} and concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) like upper(?)")
```
sdumitriu: The join mode should be configurable.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -739,8 +739,11 @@
  * NOTE: This macro uses variables defined in livetable_filterProperty . It was not meant to be used alone.
  *#
 #macro (livetable_filterStringListProperty)
```
sdumitriu: This could be applied to the other multiselect lists.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.css
```diff
@@ -167,6 +167,7 @@ thead.xwiki-livetable-display-header td.xwiki-livetable-display-header-filter se
   color: $theme.textPrimaryColor;
   margin: 0;
   padding: 0 0 0 2px;
+  margin-top: 2px;
```
sdumitriu: This change seems to affect a lot more than just this new select. Are you sure about this UI change?

pbondoer: bootstrap-select attaches itself in a `btn-group` div as a `btn` so bootstrap applies it's own styles on top of it, including a margin. Here's what it looks like:

Margins

It's probably better if we encapsulated all inputs in `btn-group`s but that's also more work, and doing so will mean we have to change a lot of styles... Another option is to just remove the margin altogether from the encapsulating div in the livetable, but this might break some bootstrap expectations. What do you think?

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
pbondoer: It's true we could simply add the script on a more global level and apply it to `select[multiple="multiple"]`, since bootstrap-select is only a layer on top of the existing controls and shouldn't break anything, but then I think it should probably be discussed.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
pbondoer: I think I understand what you mean but where is this configured in the UI side?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.css
mflorea: It's not clear for me why do you need this margin and why don't you apply it only to the bootstrap-select.

pbondoer: Because the margin on `input`s is `0` and doesn't match the bootstrap `btn` and `btn-group` styles.

mflorea: A print screen without the margin set would help visualize the problem better :)

pbondoer: It's in my first comment ;)

mflorea: If that is the screen shot without the margin then it looks fine. I don't think the margin is needed.

pbondoer: Alright, done, but I still think it looks misaligned and I don't think my change breaks anything (but I'm probably missing something).

mflorea:
@sdumitriu is there anything else that is missing from this PR in order to be merged? On my side I would prefer to enable the live search on the bootstrap select, which is useful when there are many items in the list, but it's not mandatory.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
sdumitriu: It can't be configured visually with the default column filters. However, it could be configured:

1. Through a hidden input, based on a parameter set in the livetable column options. This means that it can't be toggled by the user in the UI, but it can still be configured by the livetable author when the data semantics makes it more logical to filter on ALL or ANY of the values.
2. Something that very few people are aware of, the livetable can have an advanced filters zone, where more advanced filtering is possible. See for example [this page](https://playground.phenotips.org/AllData), expand "Advanced data filters", and look at the filters for "By clinical diagnosis" where the switch is possible.

sdumitriu: In other words, you don't have to add the UI that lets users switch, but add support right here for a configurable join mode, with the default set to "AND". Feel free to copy code from https://github.com/phenotips/phenotips/blob/phenotips-1.3/components/navigation/ui/src/main/resources/PhenoTips/LiveTableResults.xml#L562-L566

pbondoer:
@sdumitriu I added the join mode. Let me know if it's okay to merge? 😄

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -739,8 +739,17 @@
...
-  #set ($whereSql = "${whereSql} and concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) like upper(?)")
-  #set ($discard = $whereParams.add("%|$filterValue|%"))
+  #set ($filterValues = $request.getParameterValues($colname))
+  #set ($joinOperator = "$!{request.get(""${colname}${joinModeMarker}"").toUpperCase()}")
+  #if ($joinOperator != 'AND' &amp;&amp; $joinOperator != 'OR')
+    #set ($joinOperator = 'AND')
+  #end
+  #set ($filterQuery = "concat('|', concat(upper(${safe_tableAlias}.textValue), '|')) LIKE upper(?)")
+  #set ($constraints = $stringtool.repeat($filterQuery, $joinOperator, $filterValues.size()))
```
mflorea: There's no white space around the join operator. Isn't this a problem? Don't you get "ANDconcat(..."?

pbondoer: Fixed, I'm not sure how that got past...

mflorea:
@sdumitriu everything looks good for me now. I'd like to merge. Let me know if you think there is anything else to add. Thanks.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetablemulti.js
```diff
@@ -0,0 +1,19 @@
+require.config({
```
sdumitriu: Shouldn't this have a license header?

pbondoer: I'll add that now, thanks! I forget about those often...

PR #570, opened 2017-05-22
XWIKI-14119: Be able to add notifications types from a wiki page

aubincleme:
**This pull request should not be merged before the 9.4 release ! It is meant for 9.5RC1.**

This pull request allows a user to define a custom notification type directly from a wiki `XObject`.

**Major additions :**
- Add a new `RecordableEventDescriptorContainer` role in `xwiki-platform-eventstream` which allows dynamic registration of recordable event descriptors ;
- Add a new `PageNotificationEventListener` that listens to every wiki event and check for each event if it should trigger (or not) a notification event ;
- Add a new `XObject` called `PageNotificationEventDescriptorClass` that is used to define a new custom event type in wiki pages ;
- Improve `xwiki-platform-notifications` functionnal tests in order to test the new feature ;
- Add unit tests.


xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/RecordableEventDescriptorContainer.java
```diff
@@ -0,0 +1,56 @@
...
+package org.xwiki.eventstream;
+
+import java.util.List;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provide a container for {@link RecordableEventDescriptor} that can’t be registered through the
+ * component manager.
```
gdelhumeau: This makes me ask 2 questions:

1. Since this "container" does not contain **all** descriptors, we should think about an other name that reflect that. It could be "PageNotificationDescriptorContainer" or "PageNotificationDescriptorManager".
2. Since this role has been created for the sake of the implementation, it might be better to put it in the "internal" package. Otherwise it means extensions developers can use this API, and I don't think we want it.

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/internal/DefaultRecordableEventDescriptorContainer.java
```diff
@@ -0,0 +1,67 @@
...
+ * Default implementation of {@link RecordableEventDescriptorContainer}.
+ *
+ * @version $Id$
+ * @since 9.4RC1
+ */
+@Component
+@Singleton
+public class DefaultRecordableEventDescriptorContainer implements RecordableEventDescriptorContainer
+{
+    private List<RecordableEventDescriptor> recordableEventDescriptorList;
```
gdelhumeau: Use ecordableEventDescriptorList = new ArrayList<>(); and so you can remove the constructor.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
```diff
@@ -60,4 +61,14 @@
...
     void setStartDateForUser(DocumentReference userReference, Date startDate) throws NotificationException;
+
+    /**
+     * Return the notification events of a given document.
+     *
+     * @param documentReference the document reference
+     * @return the list of applied notifications events
+     * @throws NotificationException if an error occurs
+     */
+    List<PageNotificationEventDescriptor> getPageNotificationEventDescriptors(DocumentReference documentReference)
```
gdelhumeau: Why a list? A page can contains several descriptors? Technically, I know it is possible, but I'm not sure we should encourage this.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
```diff
@@ -0,0 +1,153 @@
...
+    {
+        try {
+            velocityManager.getCurrentVelocityContext().put(EVENT_BINDING_NAME, eventNotification);
+
+            PageNotificationEventDescriptor eventDescriptor =
+                    pageNotificationEventDescriptorContainer.getDescriptorByType(
+                            eventNotification.getEvents().get(0).getType());
+
+            // If we have no template defined, use the default one
+            if (eventDescriptor.getNotificationTemplate().isEmpty()) {
```
gdelhumeau: I would prefer isBlank()

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
```diff
@@ -0,0 +1,153 @@
...
+                return templateManager.execute("notification/default.vm");
```
gdelhumeau: You could also try if "notification/{eventName}.vm".

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private String executeTemplate(DocumentReference userReference, String templateContent)
+    {
+
+        try {
+            return this.authorExecutor.call(
+                    new VelocityTemplateStringEvaluator(this.contextProvider, templateContent),
+                    userReference);
+        } catch (Exception e) {
+            return "";
```
gdelhumeau: ??
At least use logger.warn() something.
You could also return the stack trace so user/developers can understand what is going on.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEvent.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public PageNotificationEventDescriptor getDescriptor()
+    {
+        return this.descriptor;
+    }
+
+    @Override
+    public boolean matches(Object otherEvent)
+    {
+        return otherEvent != null && otherEvent instanceof PageNotificationEvent;
```
gdelhumeau: Why don't you handle the eventName here?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
tmortagne: Same here, -1 for a list.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptorContainer.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@Role
+@Unstable
+public interface PageNotificationEventDescriptorContainer
+{
+    /**
+     * Update the object descriptorList.
+     * @param descriptorList the new list of descriptors to apply
+     */
+    void updateDescriptorList(List<PageNotificationEventDescriptor> descriptorList);
```
gdelhumeau: I don't understand what it is supposed to do

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,170 @@
...
+    @Inject
+    private PageNotificationEventDescriptorContainer pageNotificationEventDescriptorContainer;
+
+    @Inject
+    private Provider<XWikiContext> contextProvider;
+
+    @Inject
+    private AuthorExecutor authorExecutor;
+
+    private class VelocityTemplateBooleanEvaluator implements Callable<Boolean>
```
gdelhumeau: I have already read this code somewhere... :)

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,170 @@
...
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        boolean xObjectFound;
+
+        // Filter the event descriptors concerned by the event, then create the concerned events
+        for (PageNotificationEventDescriptor descriptor : pageNotificationEventDescriptorContainer.getDescriptorList())
+        {
+            // If the event is expected by our descriptor
+            if (descriptor.getEventTrigger().contains(event.getClass().getCanonicalName())) {
```
gdelhumeau: I think I would rely on the `matches` method instead of comparing the raw type like this.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
```diff
@@ -0,0 +1,153 @@
...
+            return new RawBlock(executeTemplate(eventDescriptor.getAuthorReference(),
+                    eventDescriptor.getNotificationTemplate()), Syntax.HTML_5_0);
+        } catch (Exception e) {
+            throw new NotificationException("Unable to render notification template.", e);
+        } finally {
+            velocityManager.getCurrentVelocityContext().remove(EVENT_BINDING_NAME);
+        }
+    }
+
+    private String executeTemplate(DocumentReference userReference, String templateContent)
```
tmortagne: You seems to be reinventing the wheel here. You can use TemplateManager#executeNoException(Template template) for custom templates.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Should also rename `getEventTrigger` to `getEventTriggers` or `getEventTriggerList`. Also need to use split() in order to return a `List<String>` of event triggers.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
```diff
@@ -0,0 +1,153 @@
...
+    private PageNotificationEventDescriptorContainer pageNotificationEventDescriptorContainer;
+
+    @Inject
+    private Provider<XWikiContext> contextProvider;
+
+    @Inject
+    private TemplateManager templateManager;
+
+    @Inject
+    private VelocityManager velocityManager;
```
tmortagne: Please don't use Velocity directly, you should always go trough template manager which among other things gives you much more option on how to implement the template.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Should also look at this : https://stackoverflow.com/questions/7495785/java-how-to-instantiate-a-class-from-string

gdelhumeau: Ok maybe not easy to create the proper Event from the canonical name of the class...

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,170 @@
...
+                XWikiDocument document = (XWikiDocument) source;
```
gdelhumeau: First verify that source is an instance of XWikiDocument. It may not. It depends on the event.

tmortagne: Yes in many events it won't be a document.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,170 @@
...
+        VelocityTemplateBooleanEvaluator(Provider<XWikiContext> contextProvider, String content)
+        {
+            this.contextProvider = contextProvider;
+            this.content = content;
+        }
+
+        @Override
+        public Boolean call() throws Exception
+        {
+            return contextProvider.get().getWiki().evaluateVelocity(content, "page-notification").contains("true");
```
gdelhumeau: 1. In my opinion, it's not the job of the `VelocityTemplateBooleanEvaluator` to also check that the result is `true`. The velocity template evaluator should be more generic and shared with the other use-case.
1. Also check if you can do this using a lambda (so you have less code duplication)
1. You rely to much on contains(). It would be safer to do result.trim().equals("true").

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-tests/src/test/it/org/xwiki/platform/notifications/test/ui/NotificationsTest.java
```diff
@@ -122,11 +145,96 @@ public void testNotifications() throws Exception
         // Delete the "Deletion test page" and test the notification
         getUtil().login(FIRST_USER_NAME, FIRST_USER_PASSWORD);
         getUtil().deletePage(getTestClassName(), "DTP");
-        
+
```
gdelhumeau: I would suggest to create a new test method for this.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/PageNotificationEventDescriptorClass.xml
```diff
@@ -0,0 +1,194 @@
...
+            <disabled>0</disabled>
+            <name>listenTo</name>
+            <number>5</number>
+            <picker>0</picker>
+            <prettyName>Listen to …</prettyName>
+            <size>64</size>
+            <unmodifiable>0</unmodifiable>
+            <validationMessage/>
+            <validationRegExp/>
+            <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
```
gdelhumeau: Since it's a list, use Textarea.

tmortagne: Why not static list ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
gdelhumeau: The problem is if the event's constructor has parameters...

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/PageNotificationEventDescriptorClass.xml
gdelhumeau: Because you don't know what are the different possible values. Maybe a static list with a display type: input.

gdelhumeau: Actually if you use a static list, it must use relational storage, multiple select, and display type: input. I don't know if the input displayer handle the other options.
In the watchlist, we use textarea to avoid the field limits (255 characters for a string).

gdelhumeau:
I have a general remark here: on the main wiki, you don't handle descriptors that comes from subwikis. See: https://jira.xwiki.org/browse/XWIKI-14275

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/PageNotificationEventDescriptorClass.xml
tmortagne: You are mixing things here. Static list officially support two use cases:
* select one or several values in a known list
* store lists

Storing a list in a texarea does not make sense, what does that is legacy stuff.

When you want to store a list then yes you usually use the input displayer and you select multiselect (which actually means "store several values"). Relational storage is optional but usually better.

gdelhumeau: If you don't use relational storage, you are limited by the size of the field that stores the values of the list (usually a string).
So ok, if you are sure that a static list using the input displayer will be stored in the database using the relational storage, then yes it's the correct approach.

tmortagne: > If you don't use relational storage, you are limited by the size of the field that stores the values of the list (usually a string).

No it's not a string. See ListClass#newProperty() implementation.

>  So ok, if you are sure that a static list using the input displayer will be stored in the database using the relational storage, then yes it's the correct approach.

It's not related to the input displayer but to multiselect option. If you plan to use the default displayer and not do your own thing you should also make sure to select the right separators for it to not conflict with characters that could be part of your values.

Here are a few examples:
* XWiki.ConfigurableClass_propertiesToShow
* many properties in ExtensionCode.Extension*Class classes
* Watchlist is using this too even if for some weird reason it's using addDBListField with no class or sql request which does not make much sense but have the same effect (everything is done at ListClass level anyway)

gdelhumeau: You're right, this works correctly.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Should we move `VelocityTemplateBooleanEvaluator` and `VelocityTemplateStringEvaluator` to a the public class ? If so, I’m not sure of where to put it. It seems to me that it’s more a generic class that shouldn’t be bound to `xwiki-platform-notifications`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptorContainer.java
aubincleme: This method is meant to bind every given descriptors to a `RecordableEventDescriptorContainer` (as the descriptors are not components, we can’t use `@Inject`) and also "un-bind" them from the `RecordableEventDescriptorContainer` when a new list is given ("un-binding" them make them recoverable by the garbage collector). 
For implementation, see : https://github.com/xwiki/xwiki-platform/pull/570/files/080eefd2bafa1a041273631d2aac22794e4ea8ad#diff-63bd246df6cd989520f8379694756f7eR51

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
aubincleme: This method is actually calling `VelocityTemplateStringEvaluator` in order to evaluate the template using the rights of the last notification author. This ensures that no privilege escalation could be done when rendering velocity templates.

tmortagne: And Template (actually TemplateContent) contains the author that need to be used to execute the template. Again you are reinventing the wheel. Velocity should never be manipulated directly in XWiki anymore. Either it's wiki content or its template content.

aubincleme: Ok, thanks !
I found different template implementations here https://github.com/xwiki/xwiki-platform/tree/master/xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template and here https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java

… but I can’t find a way to properly instantiate a `Template` in order to give him the proper author (the `Template` interface is very reduced : https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template/Template.java). So I’m thinking about writing a new implementation of `Template` that would allow me to instantiate a template by directly passing the template content and the template author to the constructor.

What do you think ? Am I still reinventing the wheel ?

tmortagne: https://github.com/xwiki/xwiki-platform/tree/master/xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template only contains API which are implemented by https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java.

> So I’m thinking about writing a new implementation of Template that would allow me to instantiate a template by directly passing the template content and the template author to the constructor.

Implementing Template is no much but sure your use case is generic enough to provide something like a StringTemplate helper somewhere in https://github.com/xwiki/xwiki-platform/tree/master/xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template yes.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-tests/src/test/it/org/xwiki/platform/notifications/test/ui/NotificationsTest.java
aubincleme: As said here : http://dev.xwiki.org/xwiki/bin/view/Community/Testing#HFunctionalTesting, it might be better to keep this into a single method, in order to limit performance issues (we talked about it with Vincent).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/PageNotificationEventDescriptorClass.xml
aubincleme: Fixed in e2bb30c

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
aubincleme: Fixed in 804fe32

aubincleme:
@gdelhumeau, @tmortagne Your comments have been fixed mostly in e2bb30c. I also added unit tests ; Jacoco instruction ratio has been raised from 0.54 to 0.65.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEvent.java
aubincleme: eventName is a value defined by the user in the notification XObject. If I use the eventName here, some comportments of the Event Stream extension aren’t respected anymore. 

For example : the component `PageNotificationRecordableEventConverter` isn’t called anymore in order to convert a `PageNotificationEvent` because the event manager extension uses `.matches()` in order to compare two events (in this case, the event we have to convert, and the result of the method `getSupportedEvents()` of `PageNotificationRecordableEventConverter`).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/CachedModelBridge.java
```diff
@@ -95,6 +96,13 @@ public void setStartDateForUser(DocumentReference userReference, Date startDate)
     }
 
     @Override
+    public PageNotificationEventDescriptor getPageNotificationEventDescriptor(DocumentReference documentReference)
+            throws NotificationException
+    {
+        return modelBridge.getPageNotificationEventDescriptor(documentReference);
```
gdelhumeau: So you don't cache it?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
```diff
@@ -62,6 +63,16 @@
     void setStartDateForUser(DocumentReference userReference, Date startDate) throws NotificationException;
 
     /**
+     * Return the notification event of a given document.
```
gdelhumeau: You mean descriptor?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Inject
+    private RecordableEventDescriptorContainer recordableEventDescriptorContainer;
+
+    @Override
+    public void updateDescriptorList(List<PageNotificationEventDescriptor> descriptorList)
+    {
+        // Remove the «old» descriptors from their RecordableEventDescriptorContainer …
+        for (PageNotificationEventDescriptor descriptor : this.descriptorList) {
+            descriptor.unRegister();
```
gdelhumeau: This looks very dangerous.
`unRegister` removes from the list the current descriptor, and you are doing a for each on the same list. You can easily end up with a `ConcurrentModificationException`.
See: https://stackoverflow.com/a/1196612.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        // … and register the new descriptors
+        for (PageNotificationEventDescriptor descriptor : descriptorList) {
+            descriptor.register(this.recordableEventDescriptorContainer);
```
gdelhumeau: Why is the descriptor which adds itself to the container? Why it is not the opposite ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        this.descriptorList = descriptorList;
+    }
+
+    @Override
+    public List<PageNotificationEventDescriptor> getDescriptorList()
```
gdelhumeau: I would prefer `getDescriptors` but it's personal :)

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEvent.java
gdelhumeau: I still don't understand. I think it's the role of PageNotificationRecordableEventConverter or the code which call it to handle this special use-case.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.xwiki.eventstream.RecordableEventDescriptor;
+import org.xwiki.eventstream.RecordableEventDescriptorContainer;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Represent a custom user-defined notification.
```
gdelhumeau: Well. This description does not make sense. Users will never create this kind of things unless they are developers. And developers can create java components instead, it's just a choice.

What is represented is a notification-type described by a wiki PAGE. It's a fact and it describes the concept better.

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/internal/DefaultRecordableEventDescriptorContainer.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.eventstream.RecordableEventDescriptor;
+import org.xwiki.eventstream.RecordableEventDescriptorContainer;
+
+/**
+ * Default implementation of {@link RecordableEventDescriptorContainer}.
+ *
+ * @version $Id$
+ * @since 9.4RC1
```
gdelhumeau: (minor: now it's 9.5RC1)

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Register the {@link PageNotificationEventDescriptor} to the {@link RecordableEventDescriptorContainer}.
+     *
+     * @param recordableEventDescriptorContainer Reference to the current {@link RecordableEventDescriptorContainer}
+     * component
+     */
+    public void register(RecordableEventDescriptorContainer recordableEventDescriptorContainer) {
+        this.recordableEventDescriptorContainer = recordableEventDescriptorContainer;
+        this.recordableEventDescriptorContainer.addRecordableEventDescriptor(this);
```
gdelhumeau: I don't understand why this descriptor register itself.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * @return the notification template
+     */
+    public String getNotificationTemplate()
+    {
+        return this.notificationTemplate;
+    }
+
+    /**
+     * @return the event canonical name
```
gdelhumeau: Which event? The events that generate new events?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public List<String> getEventTriggers()
+    {
+        return this.eventTriggers;
+    }
+
+    /**
+     * @return the event validation expression
```
gdelhumeau: What does it mean?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * @return the author of the XObject
+     */
+    public DocumentReference getAuthorReference()
+    {
+        return this.authorReference;
+    }
+
+    @Override
+    public String getEventType()
```
gdelhumeau: So it's the name the user can chose for its event?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEvent.java
gdelhumeau: Ok I see.... matches() is not supposed to work like equals()

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventUpdater.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+@Role
+@Unstable
+public interface PageNotificationEventUpdater
+{
+    /**
+     * Fetch every registered PageNotificationEventDescriptorClass XObjects in the wiki, then update the
+     * {@link PageNotificationEventDescriptorManager}.
+     */
+    void updateDescriptors();
```
gdelhumeau: Why PageNotificationEventDescriptorManager does not do it by itself?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,155 @@
...
+        for (PageNotificationEventDescriptor descriptor : pageNotificationEventDescriptorManager.getDescriptorList())
+        {
+            // If the event is expected by our descriptor
+            if (descriptor.getEventTriggers().contains(event.getClass().getCanonicalName())
+                    && checkXObjectCondition(descriptor, source)
+                    && this.evaluateVelocityTemplate(descriptor.getAuthorReference(),
+                    descriptor.getValidationExpression())) {
+                observationManager.notify(
+                        new PageNotificationEvent(descriptor),
+                        "org.xwiki.platform:xwiki-platform-notifications-api", source);
```
gdelhumeau: Why this second parameter?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,155 @@
...
+        if (StringUtils.isBlank(descriptor.getObjectType())) {
+            // If no XObject is specified in the Object Type field, we don’t need to check the source.
+            return true;
+        } else if (source instanceof XWikiDocument) {
+            XWikiDocument document = (XWikiDocument) source;
+            Map<DocumentReference, List<BaseObject>> documentXObjects = document.getXObjects();
+            /*  We can’t create a DocumentReference when only using descriptor.objectType, so we will have to
+                    iterate through the map */
+            for (DocumentReference documentReference : documentXObjects.keySet())  {
+                if (documentReference.toString().equals(descriptor.getObjectType())) {
```
gdelhumeau: It's not our best practice to use toString() directly, because the implementation can change.
Instead, you should resolve `getObjectType` to a document reference and compare the references.

You should actually use a Local Serializer, don't you think @tmortagne?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,155 @@
...
+            Template customTemplate = new StringTemplate(
+                    templateContent,
+                    Syntax.XWIKI_2_1,
+                    Syntax.PLAIN_1_0,
+                    userReference);
+
+            StringWriter writer = new StringWriter();
+            templateManager.render(customTemplate, writer);
+
+            return writer.toString().trim().contains("true");
```
gdelhumeau: contains or equals?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,155 @@
...
+        } catch (Exception e) {
+            return false;
```
gdelhumeau: You should log something.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-default/src/main/java/org/xwiki/notifications/internal/DefaultModelBridge.java
```diff
@@ -168,15 +188,55 @@ public void setStartDateForUser(DocumentReference userReference, Date startDate)
...
+        final DocumentReference pageNotificationEventDescriptorClass
+                = PAGE_NOTIFICATION_EVENT_DESCRIPTOR_CLASS.setWikiReference(documentReference.getWikiReference());
+
+        List<PageNotificationEventDescriptor> eventDescriptors = new ArrayList<>();
 
+        try {
+            XWikiDocument doc = xwiki.getDocument(documentReference, context);
+            BaseObject eventDescriptorObj = doc.getXObject(pageNotificationEventDescriptorClass);
+            if (eventDescriptorObj != null) {
+                if (eventDescriptorObj != null) {
```
gdelhumeau: `if (a) { if(a) {` ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/PageNotificationEventDescriptorClass.xml
gdelhumeau: So why am I still reading this?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
aubincleme: Yes

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
aubincleme: This is an expression (in XWiki syntax) that, when evaluated, should return `true` for the notification to be sent.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
gdelhumeau: I mean it's not clear on the javadoc

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
gdelhumeau: Same, update the javadoc (because I actually knows what it is :))

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: It’s `contains()` because the template evaluation can return elements such as `<p>`, `</p>` or `&nbsp`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/PageNotificationEventDescriptorClass.xml
aubincleme: @gdelhumeau No idea, it has been fixed in e2bb30c ; see https://github.com/xwiki/xwiki-platform/pull/570/commits/e2bb30c21864855224d201d9e99051c504bf48eb#diff-67b1ddc28dfb9b08d9fbe6e886d1a866R126

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/CachedModelBridge.java
aubincleme: No, it’s only called when adding / editing / deleting a `PageNotificationEventDescriptorClass` XObject, so I don’t think that it’s relevant enough.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventUpdater.java
aubincleme: I thought that separating the updater from the PageNotificationEventDescriptorManager would be a good idea because it would allow to better customize the way those descriptors are updated.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: Because it allows the descriptor to store a reference to the `RecordableEventDescriptorContainer` component. OFC it’s still possible to create a method in the `RecordableEventDescriptorContainer` in order to add the descriptor to its internal list and then, pass a reference of this container to the descriptor. WDYT ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
gdelhumeau: Why? You don't render to plain text?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: Ok, will fix with an Iterator

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
gdelhumeau: > Because it allows the descriptor to store a reference to the RecordableEventDescriptorContainer component

But why do you want that ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: That’s useless, indeed. Will fix that

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: Because we want to have a reference of `RecordableEventDescriptorContainer` when calling `unRegister()`.

gdelhumeau: But who call unRegister?

aubincleme: Well … it’s called just here : https://github.com/xwiki/xwiki-platform/pull/570/files/804fe324d80681cbf57ec7f3e61812a7f415ee93#diff-61eea26524abcfb18ebd691eeea92274R54 by the `DefaultPageNotificationEventDescriptorManager`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: Wait … that’s not the same list at all.

gdelhumeau: Indeed, I've misunderstood it.

aubincleme: The `unRegister()` method removes the descriptor from the `RecordableEventContainer`, not from the private list of `DefaultPageNotificationEventDescriptorManager`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
aubincleme: Yes, will fix that

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
aubincleme: Ok, will fix that

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
aubincleme: See https://github.com/xwiki/xwiki-platform/pull/570/#discussion_r118924060

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
gdelhumeau: What is stored by this private list?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/PageNotificationEventDescriptor.java
aubincleme: Ok, will fix the javadoc for this one.
It returns a list of event canonical names.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
gdelhumeau: BTW I would prefer an API such as container.unRegister(someDescriptor) instead of someDescriptor.unRegistrer().

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Actually, I can’t find a way to render this expression (in the form of a template) into plain text.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
aubincleme: Yes, will fix that

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: My problem with this kind of API is that we don’t know how the `RecordableEventDescriptorContainer` will be used in others implementations and therefore, it is possible that this `register()` / `unRegister()` become useless in the context of others extensions (as a reminder, the `RecordableEventDescriptorContainer` is part of `xwiki-platform-eventstream`)

gdelhumeau: But the container does not need to be public. It can't be an internal component?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
gdelhumeau: Multiple options:
* change the target syntax in the [rendering context](https://github.com/xwiki/xwiki-rendering/blob/73cb55c4131ac6d600c45481751f1462b4b524f4/xwiki-rendering-api/src/main/java/org/xwiki/rendering/internal/transformation/DefaultRenderingContext.java#L266-L266)
* use TemplateManager#execute() and then do the rendering manually like [RenderingScriptService](https://github.com/xwiki/xwiki-platform/blob/ed04ab37293ba194fe1006dbfe5e86820fb10534/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/script/RenderingScriptService.java#L159-L159) does.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/CachedModelBridge.java
gdelhumeau: Ok so add this explanation as a comment in the code, so nobody else will ask the question :)

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/internal/DefaultRecordableEventDescriptorContainer.java
```diff
@@ -0,0 +1,59 @@
...
+ * Default implementation of {@link RecordableEventDescriptorContainer}.
+ *
+ * @version $Id$
+ * @since 9.5RC1
+ */
+@Component
+@Singleton
+public class DefaultRecordableEventDescriptorContainer implements RecordableEventDescriptorContainer
+{
+    private List<RecordableEventDescriptor> recordableEventDescriptorList = new ArrayList<>();
```
tmortagne: ArrayList is not thread safe and this component is a singleton. You should probably use ConcurrentLinkedQueue instead for example (or deal yourself with the synchronization but usually a pain).

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/internal/DefaultRecordableEventDescriptorContainer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public void deleteRecordableEventDescriptor(RecordableEventDescriptor recordableEventDescriptor)
+    {
+        this.recordableEventDescriptorList.remove(recordableEventDescriptor);
+    }
+
+    @Override
+    public List<RecordableEventDescriptor> getAllRecordableEventDescriptor() {
+        return this.recordableEventDescriptorList;
```
tmortagne: @gdelhumeau seems to suggest that all this is supposed to be internal so It's less of an issue, otherwise returning directly a modifiable list here would be quite dangerous.

tmortagne:
Note: don't forget to add the right @since all all new public/protected methods/classes you introduced (even internal ones) as soon as you know which version it's going to end up in.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
tmortagne: There is something weird, your commit does not seems to be part of this pull request.

xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template/helpers/StringTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Allows the creation of a template directly by specifying its content as a string in the constructor.
+ *
+ * @version $Id$
+ * @since 9.5RC1
+ */
+public class StringTemplate implements Template
+{
+    private class StringTemplateContent implements TemplateContent
```
tmortagne: Hmm actually it's not really how I was seeing it. It should behave like others TemplateContent implementations and let the template itself choose options like the syntax, etc. Most of that is taken into account by DefaultTemplateContent. So instead of a public StringTemplate class you should probably provide a TemplateManager#createStringTemplate(String content, DocumentReference author) which will reuse DefaultTemplateContent behind the scene.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Fixed in d13bdf85aa

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: Fixed in d13bdf85aa

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/CachedModelBridge.java
aubincleme: Fixed in d13bdf85aa

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Fixed

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationEventDescriptorManager.java
aubincleme: Fixed in  	9126718, I merged the `PageNotificationEventDescriptorManager` with the `PageNotificationEventUpdater`

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/internal/DefaultRecordableEventDescriptorContainer.java
aubincleme: Fixed in eb81d90eb8

aubincleme:
@gdelhumeau @tmortagne Closing this pull request because of GitHub UI problems, I’ll open a new one for the same branch.

PR #572, opened 2017-05-23
XWIKI-12869- Pressing enter in the "new property" textbox should add the property

sarthak-sopho:


vmassol:
Thanks. However since this code was put in purpose we need to investigate a bit more what's the right thing to do.

sarthak-sopho:
Yeah Sure! No problem :)

sdumitriu:
If I remember correctly, the problem that was solved by that preview button was that when pressing enter in any of the inputs, not just the "new property name" input, the form was submitted. What made things worse is that before HTML added support for proper placeholders, the new property name input had a default value of `name` in it, so whenever someone pressed enter while editing the existing class properties, a new property named "name" was created, and all the existing unsaved changes were lost.

A workaround at the moment was to add that preview button before the new property form with its submit button, because that's how Enter is treated in a form: it's the same as pushing the first submit button, even if it's hidden.

The ideal behavior would be:
- pressing enter in the "new property name" input should create that property, via AJAX, just like clicking the "Add" button
- pressing enter in any other field would do nothing at all, not even preview

So, given that XWiki relies more and more on JavaScript, I would add a JS listener that captures keys on the new property name and new property type inputs, and when Enter is pressed, triggers the AJAX property addition. The hidden input should, IMHO, stay in place, since for users that have JS disabled it prevents accidental data loss.

PR #575, opened 2017-05-30
XWIKI-14119: Be able to add notifications types from a wiki page

aubincleme:
This pull request is the same as [this one](https://github.com/xwiki/xwiki-platform/pull/570). It has been recreated in order to fix GitHub UI bugs when viewing diffs.

---

This pull request allows a user to define a custom notification type directly from a wiki `XObject`.

Major additions :
- Add a new `RecordableEventDescriptorContainer` role in  `xwiki-platform-eventstream` which allows dynamic registration of recordable event descriptors ;
- Add a new `PageNotificationEventListener` that listens to every wiki event and check for each event if it should trigger (or not) a notification event ;
- Add a new `XObject` called `PageNotificationEventDescriptorClass` that is used to define a new custom event type in wiki pages ;
- Improve `xwiki-platform-notifications` functionnal tests in order to test the new feature ;
- Add unit tests.


aubincleme:
@tmortagne I fixed what you suggested [here](https://github.com/xwiki/xwiki-platform/pull/570#discussion_r118952169) in 75b193c.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
```diff
@@ -0,0 +1,178 @@
...
+            if (StringUtils.isBlank(templateContent)) {
+                return true;
+            }
+
+            Template customTemplate = templateManager.createStringTemplate(templateContent, userReference);
+            XDOM templateXDOM = templateManager.execute(customTemplate);
+
+            WikiPrinter printer = new DefaultWikiPrinter();
+            BlockRenderer renderer =
+                    this.componentManagerProvider.get().getInstance(BlockRenderer.class, "plain");
```
gdelhumeau: I know you have copied the code from the rendering script service. But I wonder why you need to get the BlockRenderer this way instead of using @Inject like any other component? @tmortagne?

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/RecordableEventDescriptorContainer.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provide a container for {@link RecordableEventDescriptor} that can’t be registered through the
+ * component manager.
+ *
+ * @version $Id$
+ * @since 9.4RC1
```
tmortagne: Wrong @since.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
tmortagne: There is no such code in rendering script service.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
```diff
@@ -62,6 +63,16 @@
     void setStartDateForUser(DocumentReference userReference, Date startDate) throws NotificationException;
 
     /**
+     * Return the notification event descriptor of a given document.
+     *
+     * @param documentReference the document reference
+     * @return the applied notification event descriptor
+     * @throws NotificationException if an error occurs
+     */
+    PageNotificationEventDescriptor getPageNotificationEventDescriptor(DocumentReference documentReference)
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+        if (event instanceof ApplicationReadyEvent) {
+            pageNotificationEventDescriptorManager.updateDescriptors();
+        } else if (event instanceof XObjectAddedEvent
+                || event instanceof XObjectDeletedEvent
+                || event instanceof XObjectUpdatedEvent) {
+            // Extract the BaseObjectReference to be able to inspect the XClassReference.
+            BaseObjectReference objectReference = getBaseObjectReference((XObjectEvent) event);
+            DocumentReference objectClassReference = objectReference.getXClassReference();
+
+            // Only interested in PageNotificationEventDescriptorClass XObjects
```
tmortagne: It's possible to directly match the class you want in the list of events. See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/event/CommentEventGeneratorListener.java#L62.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String EVENT_BINDING_NAME = "event";
+
+    @Inject
+    private PageNotificationEventDescriptorManager pageNotificationEventDescriptorManager;
+
+    @Inject
+    private TemplateManager templateManager;
+
+    @Inject
+    private VelocityManager velocityManager;
```
tmortagne: As I already said you should never manipulate Velocity directly. The right way to bind script variables is trough ScriptContext, not VelocityContext. So please use ScriptContextManager instead.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
vmassol: BTW regarding this comment I'm wondering what's our best practice: filter in onEvent() or filter in the getEvents() by returning the restrictive events. Seems to me the getEvents() solution would be the best practice since we can probably make sure we optimize the performance while the onEvent() has to be written properly to make sure it's well optimized. Also it's probably slightly easier to read (but I'm less sure about this one ;)).

Anyone thinks we should define a best practice and if yes which one?

xwiki-platform-core/xwiki-platform-eventstream/src/main/java/org/xwiki/eventstream/RecordableEventDescriptorContainer.java
aubincleme: Fixed in d6ad8a43fd

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/ModelBridge.java
aubincleme: Fixed in 188d8f9

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/page/DefaultPageNotificationDisplayer.java
aubincleme: Sorry, I forgot to remove the manager in this particular displayer ; fixed in 188d8f9.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-default/src/main/java/org/xwiki/notifications/internal/DefaultModelBridge.java
```diff
@@ -52,8 +54,6 @@
 @Singleton
 public class DefaultModelBridge implements ModelBridge
 {
-    private static final String EVENT_TYPE_FIELD = "eventType";
-
     private static final SpaceReference NOTIFICATION_CODE_SPACE = new SpaceReference("Code",
         new SpaceReference("Notifications",
             new SpaceReference("XWiki", new WikiReference("toChange"))
```
tmortagne: > new WikiReference("toChange")

seriously... (yes I know, not your fault)

Would be much cleaner to use LocalDocumentReference instead.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsTrayPage.java
```diff
@@ -181,6 +181,21 @@ public String getNotificationContent(int notificationNumber)
     }
 
     /**
+     * Get a notification.
+     *
+     * @param notificationNumber index of the notification in the list
+     * @return notification
+     */
+    public WebElement getNotificationElement(int notificationNumber)
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsUserProfilePage.java
```diff
@@ -134,9 +136,35 @@ public void setPageUpdated(boolean status)
     }
 
     /**
-     * Disable every notification parameters.
+     * Try to find a row corresponding to the notification parameter of the given type.
+     *
+     * @param eventType the event that has to be found
+     * @return the row corresponding to this element
+     */
+    public WebElement findNotificationParameterRow(String eventType)
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsUserProfilePage.java
```diff
@@ -134,9 +136,35 @@ public void setPageUpdated(boolean status)
...
+        return this.getDriver().findElement(
+                By.cssSelector("div#notificationsPane tr[data-eventtype='" + eventType + "']"));
+    }
+
+    /**
+     * Use the given row to enable a notification preference.
+     *
+     * @param notificationParameterRow the concerned notification preference row
+     */
+    public void enablePreference(WebElement notificationParameterRow)
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsUserProfilePage.java
```diff
@@ -134,9 +136,35 @@ public void setPageUpdated(boolean status)
...
+        }
+
+        this.waitForNotificationSuccessMessage(SAVED_NOTIFICATION_TEXT);
+    }
+
+    /**
+     * Disable every standard notification parameters.
      */
-    public void disableAllParameters()
+    public void disableAllStandardParameters()
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
```diff
@@ -894,4 +895,20 @@ public Template getTemplate(String templateName)
...
     }
+
+    /**
+     * Create a DefaultTemplate with the given content and the given author.
+     *
+     * @param content the template content
+     * @param author the template author
+     * @return the template
+     */
+    public Template createStringTemplate(String content, DocumentReference author)
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
```diff
@@ -894,4 +895,20 @@ public Template getTemplate(String templateName)
...
+    {
+        DefaultTemplate newTemplate = new DefaultTemplate(null);
```
tmortagne: It would be cleaner to provide a dedicated StringTemplate taking content and author and generating a proper StringResource (probably extending AbstractInputSourceResource with a StringInputSource containing the passed content).

I don't even understand how your code can work since newTemplate.content should be null.

xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template/TemplateManager.java
```diff
@@ -219,4 +220,13 @@
...
     Template getTemplate(String templateName);
+
+    /**
+     * Create a new template using a given content and a specific author.
+     *
+     * @param content the template content
+     * @param author the template author
+     * @return the template
+     */
+    Template createStringTemplate(String content, DocumentReference author);
```
tmortagne: Missing @ since.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
tmortagne: Most of the code in this class is dealing with testing what the object xclass is. Seems to me it would be far more readable to have what is in CommentEventGeneratorListener...

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
aubincleme: I don’t understand what you mean by generating a proper StringResource. How does using an AbstractInputSourceResource helps me to set a template content ?
Also, I thought that using a StringTemplate was useless ? https://github.com/xwiki/xwiki-platform/pull/570#discussion_r118952169

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
gdelhumeau: Yes we do: https://github.com/xwiki/xwiki-platform/blob/ed04ab37293ba194fe1006dbfe5e86820fb10534/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/script/RenderingScriptService.java#L162-L162

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
aubincleme: @tmortagne @vmassol As this listener extends `AbstractEventListener` (that implements `EventListener`), I don’t think that we should override the `getEvents()` method, as it is already defined in the `AbstractEventListener`. The list of events that should be listened is defined in the `super()` call in the constructor of the listener.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
tmortagne: > I don’t understand what you mean by generating a proper StringResource. How does using an AbstractInputSourceResource helps me to set a template content ?

If you look at how other templates are working you will see that the content come from the Resource. The Resource is then parsed to extract various properties (like the actual syntax used by the template), see http://extensions.xwiki.org/xwiki/bin/view/Extension/Template+Module#HContent for the currently supported properties.

What you are doing here is totally bypassing all this process to set what is supposed to be the final content after the parsing which is exactly what you where already doing in your previous public StringTemplate and that I did not liked.

Also are you sure you tested this code ? From what I see you should end up with a nullpoint  because "newTemplate.content" should be null (but it's possible I missed something).

>  Also, I thought that using a StringTemplate was useless ? #570 (comment)

I never said that. I said that having a public one not having all the features of AbtractTemplate was not very nice.

Here I'm suggesting you to have an internal StringTemplate which deal with the generation of the proper Resource to pass to AbstractTemplate. So just a constructor, that's all. I'm not talking about API but about code readability.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
tmortagne: You misread the code. We don't pass a known syntax here, this is a generic method to get a renderer from a script.

gdelhumeau: OK so now I know why Clement did it this way, because I pointed him to this code to have an inspiration.
So now it would be better to use @Inject instead.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
tmortagne: @aubincleme the point is not that CommentEventGeneratorListener getEvents (it's old that all), please look at how it defines the events.

xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template/TemplateManager.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsUserProfilePage.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsUserProfilePage.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsUserProfilePage.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsTrayPage.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Fixed in b284178149

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
tmortagne: I improve it, should be more clear now.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: @gdelhumeau I found out that injecting this component raises errors at application startup : 
`org.xwiki.component.manager.ComponentLookupException: Can't find descriptor for the component [role = [interface org.xwiki.rendering.renderer.BlockRenderer] hint = [default]]` ; so I think that I’m going to use the «old way».

vmassol: You must always understand problems :) Why is it causing problems at startup?

BTW and once you understand the reason, if you really need delayed injection we have Providers for that use case...

tmortagne: @aubincleme look at the role hint in the error message and look at the one you are using in getInstance. If I had to guess I would say you used

```
@Inject
BlockRenderer renderer;
```

instead of 
```
@Inject
@Named("plain")
BlockRenderer renderer;
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
aubincleme: Ok, thanks for your explanation @tmortagne.

So, I created a private class in the `InternalTemplateManager` : 
```
private class StringTemplate extends DefaultTemplate
{
    StringTemplate(String content) {
        super(new StringResource(content));
    }
}
```
The `StringResource` class is very basic, I pasted the implementation here : https://gist.github.com/aubincleme/cbafc6f2f2bdeca8fef227cb1f847ccf.

I think that I’m now correctly instanciating a `Template`, as a matter of fact, I can create my `StringTemplate` and assign it the correct author with two instructions : 
```
DefaultTemplate newTemplate = new StringTemplate(content);
newTemplate.content.setAuthorReference(author);
```

Unfortunately, I also found out that adding the new method `createStringTemplate` to the `TemplateManager` interface would make RevAPI scream during maven build (I forgot that `TemplateManager` was a public API). 

So, do we have a particular workflow in this case ? I’m not sure that adding the method directly to `DefaultTemplateManager` is the best idea, but I also dont’t think that it would be nice to directly access `InternalTemplateManager#createStringTemplate` directly in my extension code, as it relies a lot on the old core.

tmortagne: > Unfortunately, I also found out that adding the new method createStringTemplate to the TemplateManager interface would make RevAPI scream during maven build (I forgot that TemplateManager was a public API).

Yes this is always the case in interface. reapi is not really complaining because you are modifying an API but because older implementations of TemplateManager will not have implemented this new method.

Fortunately Java 8 introduced a solution for this use case. You can use default keyword in front of an interface method to provide a default implementation in case none is found in the actual class implementing the interface.

aubincleme: Ok, thanks (again) @tmortagne. Though I’ve read different public APIs using the `default` keyword, I did not thought about using it in this particular context …

As I can’t create a replacement template, I am going to throw a `UnsupportedOperationException`.

tmortagne: Should be fine yes. In practice there is not going to have alternative implementation of TemplateManager anytime soon I guess.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-default/src/main/java/org/xwiki/notifications/internal/DefaultModelBridge.java
aubincleme: Fixed in eeb98c40179fdc6135a16fe6c1967528607f7f24

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
aubincleme: Fixed in 75cad7deee99144734265f4dc4d997ed18fc1803

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
aubincleme: Thanks a lot for your explanations @tmortagne, I changed the implementation in 75cad7deee99144734265f4dc4d997ed18fc1803

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListenerUpdater.java
aubincleme: Thanks @tmortagne, fixed in 75cad7deee99144734265f4dc4d997ed18fc1803.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/page/events/PageNotificationEventListener.java
aubincleme: Thanks (again) @tmortagne ; fixed in 75cad7deee99144734265f4dc4d997ed18fc1803

PR #577, opened 2017-06-09
XWIKI-5146: Support LiveTable text filtering on DBListclass columns

pbondoer:
Adds DBList filtering in livetable.

Screenshot:
2017-06-09-143721_1920x1080_scrot

JIRA:
https://jira.xwiki.org/browse/XWIKI-5146


xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
```diff
@@ -0,0 +1,101 @@
...
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>false</hidden>
+  <content>{{velocity wiki="false"}}
+#set($reqClass = "$!request.get('class')")
+#set($reqProp = "$!request.get('prop')")
+#set($reqInput = "$!request.get('input')")
+#set($reqLimit = "$!request.get('limit')")
+#if ($reqLimit != '')
+  #set ($limit = $numbertool.integer($reqLimit))
```
mflorea: I see that ``$numbertool.integer`` returns a String. Is that what you want? I've used ``$mathtool.toInteger`` instead in other places. Also, if you want to be bulletproof then convert to integer whatever value is passed on the request and then fall back on a default value if necessary. See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/DocumentTreeMacros.xml#L158 .

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
```diff
@@ -0,0 +1,101 @@
...
+#else
+  ## The limit is hardcoded here, as we are only dealing with suggestions
+  #set ($limit = 25)
```
mflorea: Why not use the same default value as in the JavaScript code?

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
```diff
@@ -0,0 +1,101 @@
...
+      #foreach ($obj in $query.execute()[0].results)
+        #set ($res = {
+          'id': $obj.getFieldValue("property.${dbClass}.${dbId}_string"),
+          'name': $obj.getFieldValue("property.${dbClass}.${dbValue}_string")
+        })
+        #set($discard = $results.add($res))
+      #end
+    #else
+      ## Use HQL as there is a query present
+      #set ($sql = $sql + " AND UPPER(doc.title) LIKE UPPER(:input)")
```
mflorea: You assume that the display value is the page title, but that's not always true. What you know is that the rows selected by the query have two columns, the first is the id and the second is the display value. So you need to filter on the second column, or even both. Moreover, the HQL query might end with an ``order by`` clause or it may not even have the ``where`` clause, so you cannot simply append the ``AND`` constraint at the end.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
```diff
@@ -0,0 +1,101 @@
...
+  #set ($field = $classRef.get($reqProp))
+  #if ($field.classType == 'DBList')
+    #set ($sql = $!field.getProperty('sql').value)
+    #set ($results = [])
+    #if ($sql == '')
+      ## Use SOLR
+      #set ($dbId = "$!field.getProperty('idField').value")
+      #set ($dbValue = "$!field.getProperty('valueField').value")
+      #set ($dbClass = $services.model.serialize("$!field.getProperty('classname').value", 'solr'))
+      #set ($solr = "property.${dbClass}.${dbValue}:$reqInput")
```
mflorea: This can be easily broken if the user input has special characters. A better solution is to replace

    q = author:mflorea

with

    q = mflorea
    qf = author

So instead of putting the constraint in the query, you limit the "query fields" (where to search).

Also, the syntax "className.propertyName" is an implementation detail of the [``SolrFieldStringEntityReferenceSerializer``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/SolrFieldStringEntityReferenceSerializer.java). It doesn't make much sense to pass a String to ``serialize``. You should pass a class property reference (that you create with ``$services.model``) and just append the result to 'property.'

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
```diff
@@ -0,0 +1,101 @@
...
+      #set ($query = $services.query.createQuery($solr, 'solr'))
+      #set ($discard = $query.bindValue('fq', "type:DOCUMENT class:$dbClass"))
```
mflorea: Passing a list of filter queries instead of a string would help Solr improve the caching.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
pbondoer: I was afraid that would be the case... how can I fix this then?

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableDBListFilterService.xml
pbondoer: Something like this? Or an actual list object?

```
#set ($discard = $query.bindValue('fq', ['type:DOCUMENT', 'class:$dbClass']))
```

mflorea: Yep, exactly like this, a list of strings.

mflorea:
I did some refactoring on the Database List property in order to fix security issues ([XWIKI-14700](https://jira.xwiki.org/browse/XWIKI-14700)) and in the process I introduced a [REST resource](http://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#HClassresources) to provide suggestions for class properties (including Database List) and I integrated [selectize.js](https://selectize.github.io/selectize.js/) for an improved Database List displayer (when suggest input display type is used). I'm going to use the same displayer as live table filter. Unfortunately this makes this Pull Request obsolete. I'm sorry about this.

PR #578, opened 2017-06-12
XWIKI-14347: Add a way to register custom wiki components from XObjects

aubincleme:
Changes in interfaces : 
* Add WikiComponent#getEntityReference() method
* Add WikiComponentManager#unregisterComponents(EntityReference) method

Changes in listeners : 
* Refactor DefaultWikiComponentManagerEventListener listener
* Add DefaultWikiObjectComponentManagerEventListener listener

Changes in components : 
* Add WikiComponentManagerRegistrationHelper component
* Add WikiObjectComponentManagerRegistererProxy component
* Add WikiObjectComponentBuilder component
* Add WikiBaseObjectComponentBuilder component

Changes in tests : 
* Update tests for DefaultWikiComponentManagerEventListenerTest
* Add unit tests

### DIFF
--- a/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
+++ b/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
@@ -37,11 +37,12 @@ public interface WikiObjectComponentBuilder
 {
     /**
      * Get the class that should trigger {@link #buildComponents(ObjectReference)} when one XObject implementing
-     * this very class is added, updated or deleted in the wiki.
+     * this very class is added, referenceupdated or deleted in the wiki. The returned reference should be local if the
+     * builder is supposed to be used globally, or absolute if the builder is only used in a particular wiki.
      *
      * @return an {@link EntityReference} to the correct class.
      */
-    ObjectReference getClassReference();
+    EntityReference getClassReference();
 
     /**
      * Build the components that is linked to the given {@link ObjectReference}.
--- a/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
+++ b/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
@@ -37,7 +37,7 @@ public interface WikiObjectComponentBuilder
 {
     /**
      * Get the class that should trigger {@link #buildComponents(ObjectReference)} when one XObject implementing
-     * this very class is added, referenceupdated or deleted in the wiki. The returned reference should be local if the
+     * this very class is added, updated or deleted in the wiki. The returned reference should be local if the
      * builder is supposed to be used globally, or absolute if the builder is only used in a particular wiki.
      *
      * @return an {@link EntityReference} to the correct class.
--- a/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
+++ b/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.xwiki.component.annotation.Role;
 import org.xwiki.model.reference.EntityReference;
+import org.xwiki.model.reference.ObjectReference;
 
 /**
  * Allows to build one or more {@link WikiComponent} based on the XObjects contained in an EntityReference.
@@ -35,20 +36,20 @@ import org.xwiki.model.reference.EntityReference;
 public interface WikiObjectComponentBuilder
 {
     /**
-     * Get the class that should trigger {@link #buildComponents(EntityReference)} when one XObject implementing
+     * Get the class that should trigger {@link #buildComponents(ObjectReference)} when one XObject implementing
      * this very class is added, updated or deleted in the wiki.
      *
      * @return an {@link EntityReference} to the correct class.
      */
-    EntityReference getClassReference();
+    ObjectReference getClassReference();
 
     /**
-     * Build the components that is linked to the given {@link EntityReference}.
+     * Build the components that is linked to the given {@link ObjectReference}.
      *
      * @param reference the reference of the object that should be used to create the component.
      * @return the new component
-     * @throws WikiComponentException if the given {@link EntityReference} is incompatible with the current builder or
+     * @throws WikiComponentException if the given {@link ObjectReference} is incompatible with the current builder or
      * if the {@link WikiComponentBuilder} has not been able to instanciate the component.
      */
-    List<WikiComponent> buildComponents(EntityReference reference) throws WikiComponentException;
+    List<WikiComponent> buildComponents(ObjectReference reference) throws WikiComponentException;
 }
--- a/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
+++ b/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
@@ -1,0 +1,54 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.component.wiki;
+
+import java.util.List;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.model.reference.EntityReference;
+
+/**
+ * Allows to build one or more {@link WikiComponent} based on the XObjects contained in an EntityReference.
+ * The role hint of the component should match the XObject class name
+ *
+ * @version $Id$
+ * @since 9.5RC1
+ */
+@Role
+public interface WikiObjectComponentBuilder
+{
+    /**
+     * Get the class that should trigger {@link #buildComponents(EntityReference)} when one XObject implementing
+     * this very class is added, updated or deleted in the wiki.
+     *
+     * @return an {@link EntityReference} to the correct class.
+     */
+    EntityReference getClassReference();
+
+    /**
+     * Build the components that is linked to the given {@link EntityReference}.
+     *
+     * @param reference the reference of the object that should be used to create the component.
+     * @return the new component
+     * @throws WikiComponentException if the given {@link EntityReference} is incompatible with the current builder or
+     * if the {@link WikiComponentBuilder} has not been able to instanciate the component.
+     */
+    List<WikiComponent> buildComponents(EntityReference reference) throws WikiComponentException;
+}
--- a/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
+++ b/xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
@@ -88,8 +88,7 @@ public class DefaultWikiObjectComponentManagerEventListener extends AbstractEven
 
             // If the modified XObject can produce a WikiComponent
             if (this.wikiObjectComponentManagerEventListenerProxy.getWikiObjectsList()
-                    .contains(entityReferenceSerializer.serialize(
-                                    ((BaseObjectReference) xObject).getXClassReference()))) {
+                    .contains(((BaseObjectReference) xObject).getXClassReference())) {
                 if (event instanceof XObjectAddedEvent || event instanceof XObjectUpdatedEvent) {
                     this.wikiObjectComponentManagerEventListenerProxy
                             .registerObjectComponents((ObjectReference) xObject, (XWikiDocument) source);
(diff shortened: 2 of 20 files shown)


xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Build the components that is linked to the given {@link EntityReference}.
+     *
+     * @param reference the reference of the object that should be used to create the component.
+     * @return the new component
+     * @throws WikiComponentException if the given {@link EntityReference} is incompatible with the current builder or
+     * if the {@link WikiComponentBuilder} has not been able to instanciate the component.
+     */
+    List<WikiComponent> buildComponents(EntityReference reference) throws WikiComponentException;
```
tmortagne: I would pass an ObjectReference here since it's not supposed to work with any EntityReference, just objects.

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
```diff
@@ -0,0 +1,113 @@
...
+ * This {@link EventListener} is responsible for registering wiki components derived from XObjects using the same
+ * principle as {@link DefaultWikiComponentManagerEventListener}.
+ * 
+ * @version $Id$
+ * @since 9.5RC1
+ */
+@Component
+@Named("defaultWikiObjectComponentManagerEventListener")
+@Singleton
+public class DefaultWikiObjectComponentManagerEventListener implements EventListener
```
tmortagne: Would be a bit nicer to extend AbstractEventListener I think.

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/WikiComponentBridge.java
```diff
@@ -120,4 +122,13 @@
...
     boolean hasProgrammingRights(DocumentReference reference) throws WikiComponentException;
+
+    /**
+     * Get the XWikiDocument corresponding to the given Document Reference.
+     *
+     * @param reference the document reference
+     * @return the found XWikiDocument
+     * @throws WikiComponentException if an error occurs
+     */
+    XWikiDocument getDocument(DocumentReference reference) throws WikiComponentException;
```
tmortagne: The purpose of WikiComponentBridge is to hide oldcore API, this method is doing quite the opposite of that :)

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/WikiObjectComponentManagerRegistererProxy.java
```diff
@@ -0,0 +1,218 @@
...
+                List<String> results = query.execute();
+
+                DocumentReference sourceDocumentReference;
+                for (String result : results) {
+                    sourceDocumentReference = this.currentDocumentReferenceResolver.resolve(result);
+                    XWikiDocument document = this.wikiComponentBridge.getDocument(sourceDocumentReference);
+
+                    for (BaseObject xObject : document.getXObjects(xObjectLocalDocumentReference)) {
+                        this.registerObjectComponents(
+                                xObject.getReference(), this.wikiComponentBridge.getDocument(sourceDocumentReference));
```
tmortagne: You should pass the BaseObject instead of searching it again in the document.

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/WikiObjectComponentBuilder.java
aubincleme: Fixed in f6c73ae470
Thanks!

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
aubincleme: Indeed, that’s fixed in f6c73ae470

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/WikiComponentBridge.java
aubincleme: Indeed, it’s fixed in f6c73ae470

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/WikiObjectComponentManagerRegistererProxy.java
aubincleme: Ok, I simplified the method call in f6c73ae470
Thanks!

aubincleme:
@tmortagne : in f6c73ae470, `WikiComponentManagerRegistrationHelper` and `WikiObjectComponentManagerRegistererProxy` have been respectively renamed to `WikiComponentManagerEventListenerHelper` and `WikiObjectComponentManagerEventListenerProxy` in order to mark that those two components are (and should be) used only by event listeners.

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/WikiObjectComponentManagerEventListenerProxy.java
```diff
@@ -0,0 +1,214 @@
...
+    {
+        List<String> wikiObjectsList = new ArrayList<>();
+
+        try {
+            // Get a list of WikiObjectComponentBuilder
+            List<WikiObjectComponentBuilder> componentBuilders =
+                    this.componentManager.getInstanceList(WikiObjectComponentBuilder.class);
+
+            for (WikiObjectComponentBuilder componentBuilder : componentBuilders) {
+                wikiObjectsList.add(this.entityReferenceSerializer.serialize(componentBuilder.getClassReference()));
```
tmortagne: I don't understand why you go trough Strings. It would be safer to work with entity references IMO. Especially since you are actually manipulating entity reference everywhere you manipulate this list from what I could see producing a lot of useless resolve/serialize.

Also you should allow the same logic than MandatoryDocumentInitializer which is that WikiObjectComponentBuilder#getEntityReference() should be able to return either absolute reference (when it's only applied to main wiki for example but the extension itself is a core extension) or local reference (when it's applied to all wikis).

aubincleme: Ok, I’ll move to EntityReferences instead of strings in the wikiObjectsList.

Regarding `WikiObjectComponentBuilder#getEntityReference()`, I can absolutely change the definition of the API.

aubincleme: This is fixed in 6839eb1

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
```diff
@@ -105,4 +114,30 @@ public void onEvent(Event event, Object source, Object data)
...
+    {
+        // First, try to find a component that has a hint related to the DocumentReference of the XClass
+        try {
+            return this.componentManager.getInstance(WikiObjectComponentBuilder.class,
+                    this.entityReferenceSerializer.serialize(xClassReference));
+        } catch (ComponentLookupException e1) {
+            try {
+                // Try to find a component using the LocalDocumentReference of the XClass
+                return this.componentManager.getInstance(WikiObjectComponentBuilder.class,
+                        this.entityReferenceSerializer.serialize(xClassReference.getLocalDocumentReference()));
```
tmortagne: You will get better performance by using the "local" serializer, getLocalDocumentReference() create a new LocalDocumentReference when you call it.

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
```diff
@@ -105,4 +114,30 @@ public void onEvent(Event event, Object source, Object data)
...
+     * @param xClassReference the class reference used when searching for the component builder
+     * @return the first component builder that has been found, null if no builder is found
+     */
+    private WikiObjectComponentBuilder getAssociatedComponentBuilder(DocumentReference xClassReference)
+    {
+        // First, try to find a component that has a hint related to the DocumentReference of the XClass
+        try {
+            return this.componentManager.getInstance(WikiObjectComponentBuilder.class,
+                    this.entityReferenceSerializer.serialize(xClassReference));
+        } catch (ComponentLookupException e1) {
```
tmortagne: Catch is expensive and most component are going to be registered with a local reference so you should use hasInstance to test if the component is registered instead.

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
```diff
@@ -87,11 +94,13 @@ public void onEvent(Event event, Object source, Object data)
             EntityReference xObject = xObjectEvent.getReference();
 
             // If the modified XObject can produce a WikiComponent
-            if (this.wikiObjectComponentManagerEventListenerProxy.getWikiObjectsList()
-                    .contains(((BaseObjectReference) xObject).getXClassReference())) {
+            WikiObjectComponentBuilder componentBuilder =
+                    this.getAssociatedComponentBuilder(((BaseObjectReference) xObject).getXClassReference());
```
tmortagne: This cast might be a bit dangerous since there is nothing in the event specification about the reference being a BaseObjectReference, you might end up with an ObjectReference or even an EntityReference of type Object. I think I would get the actual object and then call getXClassReference() on it.

aubincleme: Do you mean that it would be better to get the BaseObject (or the BaseObjectReference) directly by using old core features (such as the XWikiContext) ?

I’m just asking, as this would mean moving this listener to the bridge, and having a major part of a feature implementation depending a lot on the old core might be a bit dangerous.

tmortagne: I mean that the event give you a XWikiDocument instance (with getOriginalDocument() for delete event) from which you can get the BaseObject and you can then call getXClassReference() on it. And since you have the BaseObject you can also directly pass it to wikiObjectComponentManagerEventListenerProxy.registerObjectComponents

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
aubincleme: Fixed in 75456b2a9c

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
aubincleme: Fixed in 75456b2a9c

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/DefaultWikiObjectComponentManagerEventListener.java
aubincleme: Ok, fixed in 75456b2a9c

aubincleme: Also added comments in 8e21119

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/DefaultWikiObjectComponentManagerEventListener.java
```diff
@@ -0,0 +1,164 @@
...
+    {
+        /* If we are dealing with an XObject related event, then it might be related to a
+         * WikiObjectComponentBuilder */
+        if (event instanceof XObjectEvent) {
+            XWikiDocument document = (XWikiDocument) source;
+            // If we are dealing with a XObjectDeletedEvent, we have to retrieve the original XObject document.
+            if (event instanceof XObjectDeletedEvent) {
+                document = document.getOriginalDocument();
+            }
+            BaseObjectReference baseObjectReference = document.getXObject().getReference();
```
tmortagne: Document.getXObject() return the first object of the document which is really not what you want (unless you are lucky).

You should pass the reference given by the event to getXObject().

aubincleme: Fixed in 4d6c4cf

xwiki-platform-core/xwiki-platform-component/xwiki-platform-component-wiki/src/main/java/org/xwiki/component/wiki/internal/bridge/DefaultWikiObjectComponentManagerEventListener.java
```diff
@@ -0,0 +1,165 @@
...
+                    return this.componentManager.getInstance(
+                            WikiObjectComponentBuilder.class, serializedXClassReference);
+                } else {
+                    return null;
+                }
+            }
+        } catch (ComponentLookupException e) {
+            /* As we test if a component is present in the component manager before fetching it, we shouldn't get any
+             * exception at this point. */
+            this.logger.warn(String.format(
```
tmortagne: If we end up here is means the component exist but failed to initialize which is a serious issue and should be logged with an error and the actual exception IMO.

aubincleme: Fixed in 6214fa9

tmortagne:
Great work @aubincleme !

PR #579, opened 2017-06-14
XWIKI-14348: Allow registration of custom events in wiki pages

aubincleme:
* Add TemplateManager#createStringTemplate()
* Moved xwiki-platform-eventstream to xwiki-platform-eventstream-api
* Add xwiki-platform-eventstream-default
* Add xwiki-platform-eventstream-ui
* Add UntypedRecordableEvent API
* Add UntypedRecordableEventDescriptor API
* Add WikiComponent implementation of UntypedRecordableEventDescriptor
* Update unit tests & add unit tests for
xwiki-platform-eventstream-default


xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEventDescriptor.java
```diff
@@ -0,0 +1,71 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * This interface represents the descriptor of an {@link UntypedRecordableEvent}. As an {@link UntypedRecordableEvent},
+ * UntypedRecordableEventDescriptors are registered
+ *
+ * @version $Id$
+ * @since 9.5RC1
+ */
+@Role
```
gdelhumeau: A role that extends an other role... so the component will implement which roles at the end? AFAIK, a component can only implement 1 role so you might have problem to instantiate it.

gdelhumeau: @tmortagne: WDYT?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEventDescriptor.java
```diff
@@ -0,0 +1,71 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.eventstream;
```
gdelhumeau: Should it be internal instead?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEvent.java
```diff
@@ -0,0 +1,43 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.eventstream;
```
gdelhumeau: should it be internal?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEvent.java
```diff
@@ -0,0 +1,43 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * This interface represents an event that is not identified by a specific class, but by its component hint. The event
+ * can then be retrieved through the ComponentManager.
+ *
+ * @version $Id$
+ * @since 9.5RC1
+ */
+@Role
```
aubincleme: TODO : Remove the role annotation : useless in this case

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Ensure that the given the source matches what the descriptor needs.
+     * If the source is an instance of XWikiDocument, will check if the document contains the XObject specified in
+     * the descriptor.
+     *
+     * @param descriptor the event descriptor
+     * @param source the event source
+     * @return
```
aubincleme: Add comment

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
```diff
@@ -0,0 +1,190 @@
...
+        } catch (ComponentLookupException e) {
+            logger.warn(
+                    String.format("Unable to retrieve UntypedRecordableEventDescriptor from the ComponentManager. "
+                            + "Error : %s", e.getMessage()));
+        }
+    }
+
+    /**
+     * Ensure that the given the source matches what the descriptor needs.
+     * If the source is an instance of XWikiDocument, will check if the document contains the XObject specified in
```
aubincleme: Fix `the given the source` to `the given source`

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/test/java/org/xwiki/eventstream/internal/UntypedEventListenerTest.java
```diff
@@ -0,0 +1,218 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.eventstream.internal;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import javax.jws.WebParam;
```
gdelhumeau: what is this?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/DefaultModelBridge.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Inject
+    private AuthorizationManager authorizationManager;
+
+    @Inject
+    private DocumentReferenceResolver documentReferenceResolver;
+
+    @Override
+    public Map<String, Object> getEventDescriptorProperties(BaseObject untypedEventObject)
+            throws EventStreamException
```
aubincleme: BaseObject depends on the oldcore, this method should be moved somewhere else

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/DefaultModelBridge.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public DocumentReference getAuthorReference(EntityReference entityReference) throws EventStreamException
+    {
+        return this.findDocument(entityReference).getAuthorReference();
+    }
+
+    @Override
+    public void checkRights(EntityReference entityReference, DocumentReference authorReference)
```
aubincleme: This method may be moved somewhere else as it does not depends on the oldcore

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/UntypedRecordableEventDescriptorWikiComponentBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public List<WikiComponent> buildComponents(BaseObject baseObject) throws WikiComponentException
+    {
+        try {
+            XWikiDocument parentDocument = baseObject.getOwnerDocument();
+            this.modelBridge.checkRights(parentDocument.getDocumentReference(), parentDocument.getAuthorReference());
+
+            // Get the parameters of the XObject we’re working
+            Map<String, Object> parameters = this.modelBridge.getEventDescriptorProperties(baseObject);
```
aubincleme: getEventDescriptorProperties() should be a private method of UntypedRecordableEventDescriptorWikiComponentBuilder.

xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template/TemplateManager.java
```diff
@@ -219,4 +220,20 @@
...
+     *
+     * @param content the template content
+     * @param author the template author
+     * @return the template
+     * @throws Exception if an error occurred during template instanciation
+     * @since 9.5RC1
+     */
+    default Template createStringTemplate(String content, DocumentReference author) throws Exception
+    {
+        throw new UnsupportedOperationException(
```
gdelhumeau: so it's an hidden breakage... what do others think?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEventDescriptor.java
tmortagne: It's going to implement both roles (as two different component instances in practice).

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEvent.java
tmortagne: Yes does not seems to make much sense for an event.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEvent.java
```diff
@@ -0,0 +1,43 @@
...
+@Unstable
+public interface UntypedRecordableEvent extends RecordableEvent
+{
+    /**
+     * Get the event type (the hint) of the current event.
+     *
+     * @return the event type
+     * @since 9.5RC1s
```
tmortagne: You already that this @ since at class level.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEventDescriptor.java
tmortagne: From what I understand this @Role is a mistake in this context. Nobody is supposed to request a UntypedRecordableEventDescriptor directly. It's just that some RecordableEventDescriptor are untyped.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
```diff
@@ -0,0 +1,190 @@
...
+    {
+        try {
+            // Get every UntypedEventDescriptor registered in the ComponentManager
+            List<UntypedRecordableEventDescriptor> descriptors =
+                    componentManager.getInstanceList(UntypedRecordableEventDescriptor.class);
+
+            // Filter the event descriptors concerned by the event, then create the concerned events
+            for (UntypedRecordableEventDescriptor descriptor : descriptors) {
+                // If the event is expected by our descriptor
+                if (descriptor.getEventTriggers().contains(event.getClass().getCanonicalName())
```
tmortagne: As we already discussed, this design does not make much sense to me. The event triggers should be the role hint(s) of the RecordableEventDescriptor and instead of requesting all descriptors all the time you should directly get the one associated to the event type (the class of the event unless it's a untyped event in which case it's the event#getEventType()).

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEvent.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/UntypedRecordableEventDescriptorWikiComponentBuilder.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/DefaultModelBridge.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/DefaultModelBridge.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/test/java/org/xwiki/eventstream/internal/UntypedEventListenerTest.java
aubincleme: An unused import ; fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEvent.java
aubincleme: Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEventDescriptor.java
aubincleme: Ok, I misunderstood the real utility of `@Role` annotations in interfaces : I thought that a class should implement an interface with a `@Role` in order to be registered in the component manager. Fixed in e1e2b4e.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
aubincleme: I don’t think that I get your point. 

Presently, the event type that is stored in an `UntypedRecordableEvent` acts as a sort of "unique identifier" ; it could be something like `BlogPostPublished`, or `ForumThreadAdded` … it is a way to identify the type of event we are dealing with without relying on its class name. 

The event type of an `UntypedRecordableEventDescriptor` is the same unique identifier. Therefore, with an `UntypedEvent`, we can get the corresponding descriptor with `componentManager.getInstance(UntypedRecordableEventDescriptor.class, event.getEventType())`.

With that in mind the event triggers represent a list of `Event` (not `RecordableEvent`) that should trigger the particular `UntypedEvent`. This list could be formed of `com.xpn.xwiki.internal.event.XObjectAddedEvent` or `org.xwiki.bridge.event.DocumentCreatedEvent` for example.

So, if we use the event triggers list as a hint for the `RecordableEventDescriptor`, we will have to join it, as hints are AFAIK only String-based, it will not be very clean and will involve some computation time in order to split the String and convert it back to a List every time an event is raised.

On the other side we do have to get all the descriptors all the time, as multiple descriptors can be triggered by the same event.

Sorry for the long post …

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/UntypedRecordableEventDescriptor.java
tmortagne: > I thought that a class should implement an interface with a @Role in order to be registered in the component manager

You perfectly understood but RecordableEventDescriptor already have @Role.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
tmortagne: > multiple descriptors can be triggered by the same event

That part I really don't understand, each kind of event is supposed to have only one descriptor.

aubincleme: Not really : base events (`XObjectAddedEvent`, …) are hardcoded and should trigger any UntypedRecordableEvent : for example, in a particular context, a DocumentUpdatedEvent could trigger two UntypedRecordableEvents.

tmortagne: OK it's really not clear the way you wrote the comments actually.

You are not searching for descriptors here, you are searching for configuration indicating how to generate some custom events. You decided to store those configuration in the descriptors but that's not really what descriptors are about in the public API. So indeed you can have several event generators associated to the same source event.

I'm wondering if it would not have been cleaner to separate the two concepts (on Java side at least). That would allow someone to provide some event generator configuration for a type of event which already have a descriptor.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/UntypedRecordableEventDescriptorWikiComponentBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+ * This implementation of {@link WikiBaseObjectComponentBuilder} allows to dynamically register an
+ * {@link DefaultUntypedRecordableEventDescriptor} against the Component Manager.
+ *
+ * @version $Id$
+ * @since 9.5RC1
+ */
+@Component
+@Named(UntypedRecordableEventDescriptorWikiComponentBuilder.BOUNDED_XOBJECT_CLASS)
+@Singleton
+public class UntypedRecordableEventDescriptorWikiComponentBuilder implements WikiBaseObjectComponentBuilder
```
tmortagne: Just a note, `UntypedRecordableEventDescriptorComponentBuilder` would be enough (no need for the `Wiki` part) I think (sale as `PanelWikiUIExtensionComponentBuilder` and `WikiUIExtensionComponentBuilder`).

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
gdelhumeau: Can we merge the PR like this, and consider this improvement afterwards? I think it would be great to have this work merged for 9.5RC1.

gdelhumeau: BTW I just have created https://jira.xwiki.org/browse/XWIKI-14360 to handle the improvement you mentioned in the call we had.

tmortagne: @gdelhumeau as I said I misunderstood what the code was about because of all the reference to descriptors. I checked DefaultUntypedRecordableEventDescriptor and it seems to follow hint based system we discussed.

gdelhumeau: You also mentioned:

> I'm wondering if it would not have been cleaner to separate the two concepts (on Java side at least). That would allow someone to provide some event generator configuration for a type of event which already have a descriptor.

and Clement was about to code it. However, it's gonna be too late for the next release this is why I ask if we can merge it like this.

tmortagne: That was just a comment. All that is internal anyway.

There is still a few things to fix @aubincleme is working on from what I understood then it's more up to the 9.5RC1 release manager than me since the release is on Monday and this PR is huge.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/UntypedRecordableEventDescriptorWikiComponentBuilder.java
aubincleme: Fixed in d67f267

aubincleme:
@tmortagne, @gdelhumeau `@since` annotations updated in e769a3c.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.platform</groupId>
+    <artifactId>xwiki-platform-eventstream</artifactId>
+    <version>9.5-SNAPSHOT</version>
```
gdelhumeau: 9.6-SNAPSHOT now...

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-tool-test-component</artifactId>
+      <version>${commons.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-activitystream-api</artifactId>
+      <version>9.5-SNAPSHOT</version>
```
gdelhumeau: You must use ${project.version}

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
```diff
@@ -0,0 +1,191 @@
...
+                        event, descriptor.getAuthorReference(), descriptor.getValidationExpression()))
+                {
+                    observationManager.notify(
+                            new DefaultUntypedRecordableEvent(descriptor.getEventType()),
+                            "org.xwiki.platform:xwiki-platform-eventstream-api",
+                            source);
+                }
+            }
+        } catch (ComponentLookupException e) {
+            logger.warn(
```
gdelhumeau: Usually, we pass the the exception itself as the last argument of logger.warn() so the stacktrace is displayed on the logs.

tmortagne: We don't display the stack trace when it's a warning. But we use ExceptionUtilse#getRootMessage since e.getMessage() is usually close to useless.

That being said this looks like a serious issue and should IMO be logged as an error with the complete stack trace instead.

aubincleme: Fixed in f4d5270

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/pom.xml
aubincleme: Fixed in f4d5270

tmortagne:
Applied, thanks for your patience @aubincleme !

PR #580, opened 2017-06-21
XWIKI-14376: Add support to DocumentReference in XWiki#getExternalURL()

aubincleme:
* Add XWiki#getExternalURL(DocumentReference, String, String, String,
XWikiContext)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -4675,6 +4675,28 @@ public class XWiki implements EventListener
     }
 
     /**
+     * Get the full URL of the given {@link DocumentReference}. This also includes the server name of the wiki.
+     *
+     * @param documentReference the document that should be resolved
+     * @param action the action of the URL
+     * @param querystring the URL parameters
+     * @param anchor the anchor of the document
+     * @param context the current XWikiContext
+     * @return the full URL of the given reference
+     *
+     * @since 9.6RC1
+     */
+    public String getExternalURL(DocumentReference documentReference, String action,
+            String querystring, String anchor, XWikiContext context)
+    {
+        URL url = context.getURLFactory().createExternalURL(
+                this.getLocalStringEntityReferenceSerializer().serialize(documentReference.getLastSpaceReference()),
+                documentReference.getName(), action, querystring, anchor,
+                documentReference.getWikiReference().getName(), context);
+        return url.toString();
+    }
+
+    /**
      * @since 7.2M1
      */
     public String getAttachmentURL(AttachmentReference attachmentReference, String action, String queryString,


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -4675,6 +4675,28 @@ public String getExternalURL(String fullname, String action, String querystring,
...
+     * @param context the current XWikiContext
+     * @return the full URL of the given reference
+     *
+     * @since 9.6RC1
+     */
+    public String getExternalURL(DocumentReference documentReference, String action,
+            String querystring, String anchor, XWikiContext context)
+    {
+        URL url = context.getURLFactory().createExternalURL(
+                this.getLocalStringEntityReferenceSerializer().serialize(documentReference.getLastSpaceReference()),
```
sdumitriu: Does `getLastSpaceReference()` play nice with nested spaces?

sdumitriu: I think it does, since it returns the full reference, including all its parents, right?

aubincleme: I guess so (at least, it works on my tests). Are you thinking about a particular use case where this wouldn’t work ?
Also, maybe @tmortagne has a more direct answer to your question.

tmortagne: getLastSpaceReference() does not return a new reference in practice, it's just a "pointer" to the last space element so yes you have all the parents.

Enygma2002:
Nice, @aubincleme. We were missing that.

Would have been even nicer to also refactor the other string-based methods to delegate to this new reference-based one, instead of having multiple implementations.

aubincleme:
@Enygma2002 good idea, I didn’t take the time to do it because I needed the feature for something else.
I’ll do the refactor at some point and I’ll keep you up to date :) !

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -4675,6 +4675,28 @@ public String getExternalURL(String fullname, String action, String querystring,
...
+     * @param documentReference the document that should be resolved
+     * @param action the action of the URL
+     * @param querystring the URL parameters
+     * @param anchor the anchor of the document
+     * @param context the current XWikiContext
+     * @return the full URL of the given reference
+     *
+     * @since 9.6RC1
+     */
+    public String getExternalURL(DocumentReference documentReference, String action,
```
mflorea: Why not make it more generic with ``getExternalURL(EntityReference, ...)``? We have ``getURL(EntityReference)``.

PR #587, opened 2017-06-27
XWIKI-14348: Allow registration of custom events in wiki pages

aubincleme:
* Add source object in the context of the validation template in UntypedEventListener
* Fix the type of renderer used in validation templates : the plain renderer could not display Velocity output.


xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-default/src/main/java/org/xwiki/eventstream/internal/UntypedEventListener.java
```diff
@@ -70,15 +70,20 @@
...
     @Inject
     private ObservationManager observationManager;
 
     @Inject
     private TemplateManager templateManager;
 
     @Inject
-    @Named("plain/1.0")
-    private BlockRenderer plainRenderer;
+    @Named("html/5.0")
```
gdelhumeau: Could explain quickly why?

aubincleme: When using the plain renderer, the velocity code is not rendered ; I think that it’s due to the XDOM  object returned by `templateManager#execute()` (in https://github.com/xwiki/xwiki-platform/pull/587/files#diff-0a971406e804cfdf9941914cbd70ec4cL169) has the property "HTML".

gdelhumeau: @tmortagne: Does it ring a bell?

tmortagne: This has nothing to do with velocity. But plain renderer skip RawBlock indeed (what you end up when you use the html macro for example).

gdelhumeau: Thanks

PR #590, opened 2017-07-04
XWIKI-14507: Allow actions in the event stream to be observed

aubincleme:
* Add EventStreamAddedEvent and EventStreamDeletedEvent
* Update the BridgeEventStream to trigger those events when needed
* Update BridgeEventStream in order to avoid loops between observation events and event stream events

### DIFF
--- a/xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
+++ b/xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
@@ -31,6 +31,7 @@ import org.xwiki.context.Execution;
 import org.xwiki.eventstream.Event;
 import org.xwiki.eventstream.EventGroup;
 import org.xwiki.eventstream.EventStream;
+import org.xwiki.eventstream.events.AbstractEventStreamEvent;
 import org.xwiki.eventstream.events.EventStreamAddedEvent;
 import org.xwiki.eventstream.events.EventStreamDeletedEvent;
 import org.xwiki.observation.ObservationManager;
@@ -53,6 +54,12 @@ import com.xpn.xwiki.plugin.activitystream.plugin.ActivityStreamPlugin;
 @Singleton
 public class BridgeEventStream implements EventStream
 {
+    /**
+     * Used to provide a key to a property in the current execution context that avoids stepping into a loop when
+     * triggering new events.
+     */
+    private static final String EVENT_CONTEXT_LOCK_PROPERTY = "eventContextProperty";
+
     /** Needed for accessing the current request context. */
     @Inject
     private Execution execution;
@@ -74,7 +81,7 @@ public class BridgeEventStream implements EventStream
             XWikiContext context = getXWikiContext();
             ActivityStreamPlugin plugin = getPlugin(context);
             plugin.getActivityStream().addActivityEvent(eventConverter.convertEventToActivity(e), context);
-            this.observationManager.notify(new EventStreamAddedEvent(e), e);
+            this.sendEventStreamEvent(new EventStreamAddedEvent(e));
         } catch (ActivityStreamException ex) {
             // Unlikely; nothing we can do
         }
@@ -87,9 +94,20 @@ public class BridgeEventStream implements EventStream
             XWikiContext context = getXWikiContext();
             ActivityStreamPlugin plugin = getPlugin(context);
             plugin.getActivityStream().deleteActivityEvent(eventConverter.convertEventToActivity(e), context);
-            this.observationManager.notify(new EventStreamDeletedEvent(e), e);
+            this.sendEventStreamEvent(new EventStreamDeletedEvent(e));
         } catch (ActivityStreamException ex) {
             // Unlikely; nothing we can do
+        }
+    }
+
+    private void sendEventStreamEvent(AbstractEventStreamEvent eventStreamEvent)
+    {
+        // In order to avoid infinite loop caused by observation events triggering event stream events that are
+        // themselves triggering new observation events …, we add a new property to the execution context. Therefore,
+        // an observation event declared out of an event stream event can only be triggered once.
+        if (!this.execution.getContext().hasProperty(EVENT_CONTEXT_LOCK_PROPERTY)) {
+            this.execution.getContext().newProperty(EVENT_CONTEXT_LOCK_PROPERTY).declare();
+            this.observationManager.notify(eventStreamEvent, eventStreamEvent.getEvent());
         }
     }
 
--- a/xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
+++ b/xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
@@ -81,7 +81,7 @@ public class BridgeEventStream implements EventStream
             XWikiContext context = getXWikiContext();
             ActivityStreamPlugin plugin = getPlugin(context);
             plugin.getActivityStream().addActivityEvent(eventConverter.convertEventToActivity(e), context);
-            this.sendEventStreamEvent(new EventStreamAddedEvent(e));
+            this.sendEventStreamEvent(new EventStreamAddedEvent(), e);
         } catch (ActivityStreamException ex) {
             // Unlikely; nothing we can do
         }
@@ -94,20 +94,20 @@ public class BridgeEventStream implements EventStream
             XWikiContext context = getXWikiContext();
             ActivityStreamPlugin plugin = getPlugin(context);
             plugin.getActivityStream().deleteActivityEvent(eventConverter.convertEventToActivity(e), context);
-            this.sendEventStreamEvent(new EventStreamDeletedEvent(e));
+            this.sendEventStreamEvent(new EventStreamDeletedEvent(), e);
         } catch (ActivityStreamException ex) {
             // Unlikely; nothing we can do
         }
     }
 
-    private void sendEventStreamEvent(AbstractEventStreamEvent eventStreamEvent)
+    private void sendEventStreamEvent(AbstractEventStreamEvent eventStreamEvent, Event event)
     {
         // In order to avoid infinite loop caused by observation events triggering event stream events that are
         // themselves triggering new observation events …, we add a new property to the execution context. Therefore,
         // an observation event declared out of an event stream event can only be triggered once.
         if (!this.execution.getContext().hasProperty(EVENT_LOOP_CONTEXT_LOCK_PROPERTY)) {
             this.execution.getContext().newProperty(EVENT_LOOP_CONTEXT_LOCK_PROPERTY).declare();
-            this.observationManager.notify(eventStreamEvent, eventStreamEvent.getEvent());
+            this.observationManager.notify(eventStreamEvent, event);
         }
     }
 
--- a/xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
+++ b/xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
@@ -31,6 +31,9 @@ import org.xwiki.context.Execution;
 import org.xwiki.eventstream.Event;
 import org.xwiki.eventstream.EventGroup;
 import org.xwiki.eventstream.EventStream;
+import org.xwiki.eventstream.events.EventStreamAddedEvent;
+import org.xwiki.eventstream.events.EventStreamDeletedEvent;
+import org.xwiki.observation.ObservationManager;
 import org.xwiki.query.Query;
 import org.xwiki.query.QueryException;
 import org.xwiki.query.QueryManager;
@@ -61,6 +64,9 @@ public class BridgeEventStream implements EventStream
     @Inject
     private EventConverter eventConverter;
 
+    @Inject
+    private ObservationManager observationManager;
+
     @Override
     public void addEvent(Event e)
     {
@@ -68,6 +74,7 @@ public class BridgeEventStream implements EventStream
             XWikiContext context = getXWikiContext();
             ActivityStreamPlugin plugin = getPlugin(context);
             plugin.getActivityStream().addActivityEvent(eventConverter.convertEventToActivity(e), context);
+            this.observationManager.notify(new EventStreamAddedEvent(e), e);
         } catch (ActivityStreamException ex) {
             // Unlikely; nothing we can do
         }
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
```diff
@@ -80,11 +94,23 @@ public void deleteEvent(Event e)
...
         }
     }
 
+    private void sendEventStreamEvent(AbstractEventStreamEvent eventStreamEvent)
+    {
+        // In order to avoid infinite loop caused by observation events triggering event stream events that are
+        // themselves triggering new observation events …, we add a new property to the execution context. Therefore,
+        // an observation event declared out of an event stream event can only be triggered once.
+        if (!this.execution.getContext().hasProperty(EVENT_CONTEXT_LOCK_PROPERTY)) {
+            this.execution.getContext().newProperty(EVENT_CONTEXT_LOCK_PROPERTY).declare();
```
aubincleme: @tmortagne: Can you confirm that the execution context has its properties flushed at the end of a request ?
If not, I’ll have to add something else here.

tmortagne: Yes it's not your job to take care of cleaning ExecutionContext.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/EventStreamAddedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public EventStreamAddedEvent(Event event)
+    {
+        super(event);
+    }
+
+    @Override
+    public boolean matches(Object o)
+    {
+        return (o instanceof EventStreamAddedEvent && ((EventStreamAddedEvent) o).getEvent().equals(this.event));
```
tmortagne: Sounds like something that should be moved to AbstractEventStreamEvent.

aubincleme: Fixed in 84ec8f3cc5

xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
```diff
@@ -50,6 +54,12 @@
 @Singleton
 public class BridgeEventStream implements EventStream
 {
+    /**
+     * Used to provide a key to a property in the current execution context that avoids stepping into a loop when
+     * triggering new events.
+     */
+    private static final String EVENT_CONTEXT_LOCK_PROPERTY = "eventContextProperty";
```
gdelhumeau: I think the name is not very descriptive :(

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
```diff
@@ -0,0 +1,56 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.eventstream.events;
+
+import org.xwiki.eventstream.Event;
+
+/**
+ * Groups every event regarding the {@link org.xwiki.eventstream.EventStream}.
```
gdelhumeau: I don't understand the description.

xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
tmortagne: The id should be a it more specific to be sure to not hit a conflict.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
aubincleme: Fixed in eb29721c8a

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
```diff
@@ -0,0 +1,62 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.eventstream.events;
+
+import org.xwiki.eventstream.Event;
+
+/**
+ * Abstract method for every event that is related to the {@link org.xwiki.eventstream.EventStream}.
```
gdelhumeau: Abstract **class** :)

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
```diff
@@ -0,0 +1,62 @@
...
+ * When an event that extends an {@link AbstractEventStreamEvent} is triggered, it should also be triggered
+ * with the corresponding event stream event as the source object.
```
gdelhumeau: Why? Since it is also given in the event itself?

aubincleme: Well … when registering an event against the ObservationManager, we have to provide at least one object with the event itself. So it’s more like a "guideline" like what is advised in https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentUpdatedEvent.java (for example).

tmortagne: > we have to provide at least one object

Not sure where you got that and it's certainly not from https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentUpdatedEvent.java which only indicate what you get with DocumentUpdatedEvent, no general guidelines in there.

aubincleme: Well then we should have been way more careful when designing the way the ObservationManager works, because if you take a look at a vast majority of the events defined [here](https://github.com/xwiki/xwiki-platform/tree/master/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event), you’ll see that they all come with javadoc such as : 
```
 * The event also send the following parameters:
 * </p>
 * <ul>
 * <li>source: the current {com.xpn.xwiki.doc.XWikiDocument} instance</li>
```

… which, combined with the fact that the [ObservationManager#notify()](https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/ObservationManager.java#L90) methods requires at least one object with the event and that [the observation manager isn’t that well documented](https://network.xwiki.com/xwiki/bin/view/DocXCS2En/ObservationModule), is **super confusing**.

---

Anyway … just wanted to say that it’s not obvious. Will be fixed soon.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
aubincleme: Fixed in 03d2311b2b

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
aubincleme: Fixed in 03d2311b2b

xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/eventstreambridge/BridgeEventStream.java
aubincleme: Fixed in 03d2311b2b

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
tmortagne: I really don't understand what is your issue here. The vast majority of event do have things to send that make sense, they don't repeat what they already have in the Event.

If you don't have anything special to send then don't. As simple as that.

gdelhumeau: Well... I was just asking :)

I'm not against this javadoc since you actually pass the event as source, it's even better because it means it is documented. I was wondering what other object we could pass as "source", and I see no candidate :(

gdelhumeau: @tmortagne: you mean we can do `om.notify(event, null);` ?

tmortagne: Sure, why not ? OM does not care about what you send, it's just passing it trough.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/EventStreamAddedEvent.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @since 9.6RC1
+ * @version $Id$
+ */
+public class EventStreamAddedEvent extends AbstractEventStreamEvent
+{
+    /**
+     * Constructs a new {@link EventStreamAddedEvent}.
+     */
+    public EventStreamAddedEvent()
```
gdelhumeau: Why do you need a constructor with no event?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public Event getEvent()
+    {
+        return this.event;
+    }
+
+    @Override
+    public boolean matches(Object o)
+    {
+        return (this.getClass().isInstance(o) && ((AbstractEventStreamEvent) o).getEvent().equals(this.event));
```
gdelhumeau: It means you cannot create a listener actually, because the listener will never match:
```
((AbstractEventStreamEvent) o).getEvent().equals(this.event))
```
You see what I mean?

gdelhumeau: This actually tends in favor of passing the event... as the source.

You are free to hate me.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/EventStreamAddedEvent.java
aubincleme: See https://github.com/aubincleme/xwiki-platform/blob/xwiki-14353/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/internal/email/LiveNotificationEmailListener.java#L106 for example

tmortagne: > Why do you need a constructor with no event?

For the listener ?

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
tmortagne: That's why we usually don't pass complex object in the Event, it's supposed to contain what's needed to match the event only and real data are passed as parameters.

aubincleme: Fixed in 8845ce0190

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/EventStreamAddedEvent.java
aubincleme: Fixed in 8845ce0190

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @since 9.6RC1
+ * @version $Id$
+ */
+public abstract class AbstractEventStreamEvent implements org.xwiki.observation.event.Event
+{
+    @Override
+    public boolean matches(Object o)
+    {
+        return (this.getClass().isInstance(o));
```
gdelhumeau: I would have used `instance of`. See: https://stackoverflow.com/a/8692245
But it's ok.

PR #593, opened 2017-07-07
XWIKI-14507: Allow actions in the event stream to be observed

aubincleme:
* Add new semaphore on EventStreamEvents


xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/impl/ActivityStreamImpl.java
```diff
@@ -884,8 +885,14 @@ public void onEvent(Event event, Object source, Object data)
             return;
         }
 
+        Execution executionContext = Utils.getComponent(Execution.class);
+
         // Take events into account only once in a cluster
-        if (!Utils.getComponent(RemoteObservationManagerContext.class).isRemoteState()) {
+        if (!Utils.getComponent(RemoteObservationManagerContext.class).isRemoteState()
+                && executionContext.getContext().hasProperty(
```
tmortagne: `!executionContext.getContext().hasProperty` ?

xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/internal/RecordableEventListener.java
```diff
@@ -86,8 +91,16 @@ public void onEvent(Event event, Object source, Object data)
         }
 
         try {
-            // Save the event in the event stream
-            eventStream.addEvent(convertEvent(event, source, data));
+            if (this.execution.getContext().hasProperty(AbstractEventStreamEvent.EVENT_LOOP_CONTEXT_LOCK_PROPERTY)) {
```
tmortagne: Same comment, looks like inverted if

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
```diff
@@ -27,6 +27,12 @@
  */
 public abstract class AbstractEventStreamEvent implements org.xwiki.observation.event.Event
 {
+    /**
+     * Used to provide a key to a property in the current execution context that avoids stepping into a loop when
+     * triggering new events.
+     */
+    public static final String EVENT_LOOP_CONTEXT_LOCK_PROPERTY = "eventLoopContextLockProperty";
```
tmortagne: I would keep this internal. By putting it here you are making it a public API (not an *.internal.* package).

xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/impl/ActivityStreamImpl.java
aubincleme: Yes … see the next line :) (have to respect the 120 chars rule)

aubincleme: Fixed in ad4eae2b1d.

xwiki-platform-core/xwiki-platform-activitystream/xwiki-platform-activitystream-api/src/main/java/com/xpn/xwiki/plugin/activitystream/internal/RecordableEventListener.java
aubincleme: Fixed in ad4eae2b1d.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/events/AbstractEventStreamEvent.java
aubincleme: I’m not sure that it’s a good idea to put the property in an internal package : this could be accessed by an external listener for any kind of computation.
Also, it would mean that we’re using internal properties of the Event Stream from the Activity Stream module.

tmortagne: bq.  Also, it would mean that we’re using internal properties of the Event Stream from the Activity Stream module.

This is not really an issue IMO. You are talking about two things located in the same module right now. But nobody else is supposed to used that property since you did what was required to protect against cycles already. Making an internal tool public is easy but the opposite is very hard so without any use case in mind you should start internal.

PR #594, opened 2017-07-12
XWIKI-14524: Korean PDF export broken

evalica:
- Replaced 'kp' with 'ko' since the 'kp' language code does not exist (according to https://www.loc.gov/standards/iso639-2/php/code_list.php) and 'ko' is the correct code for Korean language


evalica:
Waiting for 9.6-rc-1 to be released, so after I will push for 9.6 final

PR #608, opened 2017-08-17

pbondoer:
2017-08-17-231452_1920x1064_scrot

JIRA: http://jira.xwiki.org/browse/XWIKI-14162

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -6,6 +6,18 @@
   .bottombuttons {
     .buttons {
       margin: 10px 0;
+      
+      &.buttons-fixed {
+        z-index: 1030;
+        position: fixed;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        margin: 0;
+        padding: 1em;
+        background-color: white;
+        box-shadow: 0 0 1em rgba(0, 0, 0, 0.3);
+      }
     }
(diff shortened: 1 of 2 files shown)


tmortagne:
Very nice screenshot except that I don't understand what the Show tour is doing here (both in edit mode and in this location).

vmassol:
It's also weird that it's showing the GWT editor :)

vmassol:
@tmortagne show tour is always displayed at the bottom right. But now with visible save we'll need to do something about this. Indeed one idea is to not show it in edit mode.

tmortagne:
> show tour is always displayed at the bottom right

I tough is was under Cancel button but after testing a recent version I understand it's actually not really inside the page but a flowing bottom button. Anyway does not make any sense to have this in anything else than view mode IMO but yes nothing to do with this pull request. Sorry for the noise @pbondoer.

evalica:
What is displayed in the screenshot is misleading since this PR doesn't include also the 
editactions.vm. Also there are many issues that can occur when the buttons are displayed in inline mode, etc. 
So this PR is a good start, but in order for the feature to be usable, someone else needs to handle the further integration. So I would not want us to merge it, unless we have a new assignee.

pbondoer:
@evalica That would be because I didn't have to modify `editactions.vm` to get this result. This is more of a proof of concept that works for what I managed to figure out on my own, however I am not nearly as experienced as others so I'm not really aware of what might cause problems down the line :(

mflorea:
@tmortagne, @vmassol we can have tours for edit mode. A tour for the CKEditor would be nice for instance. Also, with the ability to have tours that span across multiple pages we will want to make tours for an entire workflow: create page wizard then edit mode then view mode. So we do need to take into account the "Show tour" button if we redesign the form action buttons.

vmassol:
@mflorea indeed, good point

evalica:
Closing this PR since it was integrated with https://github.com/xwiki/xwiki-platform/pull/738

PR #609, opened 2017-08-21
XWIKI-14608: The UI does not allow setting Marathi as language anymore.

gdelhumeau:
* Add ICU4j as a dependency to handle more locales

The dependency's weight is 11 MB, which is heavy for our use-case. We end up with 710 locales... maybe too much

### DIFF
--- a/xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-script/src/main/java/org/xwiki/localization/script/LocalizationScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-script/src/main/java/org/xwiki/localization/script/LocalizationScriptService.java
@@ -40,6 +40,8 @@ import org.xwiki.rendering.renderer.printer.DefaultWikiPrinter;
 import org.xwiki.rendering.syntax.Syntax;
 import org.xwiki.script.service.ScriptService;
 
+import com.ibm.icu.util.ULocale;
+
 /**
  * Provides Component-specific Scripting APIs.
  * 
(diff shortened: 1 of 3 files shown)


aubincleme:
@gdelhumeau : Do you know if we can get the same problem as reported in https://jira.xwiki.org/browse/XWIKI-14168 ?

gdelhumeau:
@aubincleme: It does not fix the problem you mention, and we can even consider it makes it worse.

See: https://jira.xwiki.org/browse/XWIKI-14608

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-script/src/main/java/org/xwiki/localization/script/LocalizationScriptService.java
```diff
@@ -267,4 +269,15 @@ public String render(String key, Syntax syntax, Collection<?> parameters, Locale
...
+
+    /**
+     * @return the list of all known locales
+     * @since 9.7RC1
+     * @since 8.4.6
+     * @since 9.6.1
+     */
+    public ULocale[] getKnownLocales()
+    {
+        return ULocale.getAvailableLocales();
```
gdelhumeau: @tmortagne told me there is a bridge provided by ICU to use plain old `Locale` instead of `ULocale`.

It miss the `@Unstable` tag too.

gdelhumeau: I don't understand your question

tmortagne: Using the ICU API somewhere only make sense if you use it everywhere. For the bridge, see http://userguide.icu-project.org/icu4j-locale-service-provider.

PR #611, opened 2017-08-31
XWIKI-14649: Delete UI announces success immediately regardless of the actual result (for big terminal pages)

lucaa:
Changed the condition for operation finished in delete.js, to check the actual state of the job and not its progress


lucaa:
Note that I only tested this on a 8.4.4 and for the deletion of a single terminal page.
I would say it doesn't break the other situations, but I didn't really test it.

PR #612, opened 2017-09-05
XWIKI-14663: missing class id in flamingo menu

Pbast:


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/shortcuts.vm
```diff
@@ -54,6 +54,9 @@
   #if ($docextralink.size() > 4)
     #set ($extraParams = $docextralink.get(4))
   #end
+  #if ($docextralink.size() > 5)
+    #set ($idHref = $docextralink.get(5))
```
mflorea: ``$idHref`` is not initialized (reset) if the condition is not met which means it will keep the value from the previous ``for`` loop. As a result you can have a menu entry that uses the ``$idHref`` from the previous menu entry.

Pbast: Fine, I fixed it

PR #613, opened 2017-09-08
XWIKI-14674: Viewing a deleted document as guest redirects to login

sdumitriu:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/DeletedDocument.java
```diff
@@ -151,8 +151,7 @@ public boolean canDelete()
         try {
             XWikiDocument doc = new XWikiDocument();
             doc.setFullName(getFullName(), this.context);
-            if (!hasAccessLevel(ADMIN_RIGHT, getFullName())
-                && !getXWikiContext().getWiki().getRightService().checkAccess("delete", doc, this.context)) {
+            if (!hasAccessLevel(ADMIN_RIGHT, getFullName()) && !hasAccessLevel("delete", getFullName())) {
```
tmortagne: Since you are at it you could remove the ADMIN right test since it imply DELETE anyway.

sdumitriu: Done.

PR #614, opened 2017-10-04
XWIKI-14763: Use the document's locale to find the title

piis3:
The search result's title field is empty because it's multi-lingual field.
This causes the title in solr search results to be null. So use the locale
field to look up the correct title field to populate the search result with.

Jira: https://jira.xwiki.org/browse/XWIKI-14763

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
@@ -136,7 +136,7 @@ public class SOLRSearchSource extends AbstractSearchSource
         // Boost
         // FIXME: take it from configuration
         query.bindValue("qf",
-            "title^10.0 name^10.0 doccontent^2.0 objcontent^0.4 filename^0.4 attcontent^0.4 doccontentraw^0.4 "
+            "title_^10.0 name^10.0 doccontent^2.0 objcontent^0.4 filename^0.4 attcontent^0.4 doccontentraw^0.4 "
                 + "author_display^0.08 creator_display^0.08 " + "comment^0.016 attauthor_display^0.016 space^0.016");
 
         // Order
@@ -161,7 +161,6 @@ public class SOLRSearchSource extends AbstractSearchSource
 
                 DocumentReference documentReference = this.solrDocumentReferenceResolver.resolve(document);
                 searchResult.setPageFullName(this.localEntityReferenceSerializer.serialize(documentReference));
-                searchResult.setTitle((String) document.get(FieldUtils.TITLE));
                 searchResult.setWiki(documentReference.getWikiReference().getName());
                 searchResult.setSpace(this.localEntityReferenceSerializer.serialize(documentReference.getParent()));
                 searchResult.setPageName(documentReference.getName());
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
```diff
@@ -136,7 +136,7 @@
         // Boost
         // FIXME: take it from configuration
         query.bindValue("qf",
-            "title^10.0 name^10.0 doccontent^2.0 objcontent^0.4 filename^0.4 attcontent^0.4 doccontentraw^0.4 "
+            "title_^10.0 name^10.0 doccontent^2.0 objcontent^0.4 filename^0.4 attcontent^0.4 doccontentraw^0.4 "
```
sdumitriu: Are you sure about this change?

mflorea: Doesn't look good to me either.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
```diff
@@ -181,20 +180,27 @@
                     searchResult.setAuthorName((String) document.get(FieldUtils.AUTHOR_DISPLAY));
                 }
 
-                Locale locale = LocaleUtils.toLocale((String) document.get(FieldUtils.DOCUMENT_LOCALE));
+                Locale docLocale = LocaleUtils.toLocale((String) document.get(FieldUtils.DOCUMENT_LOCALE));
+                Locale locale = LocaleUtils.toLocale((String) document.get(FieldUtils.LOCALE));
+
+                List<String> titles = (List<String>) document.get(FieldUtils.getFieldName(FieldUtils.TITLE, locale));
```
mflorea: You should use ``document.getFirstValue(...)`` instead of doing the tests yourself below.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
piis3: I wasn't totally sure about this, but there isn't actually a field in the schema called "title". Instead there's a field called "title_" which looks like it gets populated with all of the localized title fields. Here's the definition of that field in the solr schema:

https://github.com/xwiki/xwiki-platform/blob/d3ca366ca7ff9926d0ff25ecf89acaf697368d6a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/managed-schema#L170

Later in the schema there's a copyField from title_* to title_. So to boost a match against title, it seems like this is the way to go.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
piis3: Better, yes. Thanks.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java
mflorea: See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/solrconfig.xml#L766 . "title" is a shortcut for all "title_*" fields.

piis3: Ok, that makes sense. I will revert that line to the original then. Thanks for the explanation.

mflorea:
Thanks for your contribution!

PR #615, opened 2017-10-19
XWIKI-5059: Fix issue related to getURLContent()

aubincleme:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
```diff
@@ -2283,8 +2283,13 @@ public String getAdClientId()
      * @return Content of the specified URL
      * @throws IOException
      */
+    @Programming
     public String getURLContent(String surl, String username, String password) throws IOException
     {
+        if (!hasProgrammingRights()) {
+            return "";
```
vmassol: I think I would have returned null instead to show it's not expected (same for the other APIs below). WDYT?

vmassol: Q: are any of the methods below used by code used by the Standard Flavor? If so, that would mean more pages requiring PR.

vmassol: Note that if we apply this, we'll need to update doc on xwiki.org too to mention the need for PR (I'm thinking for example about http://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/Scripting/APIGuide/#HGettingexternalcontent ).

aubincleme: We can return `null` indeed. I'll check for the use of those methods on the Standard Flavor and get back to you.

vmassol: Replying to myself about "I think I would have returned null instead to show it's not expected (same for the other APIs below). WDYT?". I see now that you've kept the current behavior. Makes sense even though I'd say null could be better but I guess that would break backward compat.

vmassol: "We can return null indeed. I'll check for the use of those methods on the Standard Flavor and get back to you." No you were right, you kept the behavior of what was done in the try/catch, which is good I think.

aubincleme: Yep, it seems that the API is never used (`grep`'d it)

vmassol: How do we handle backward compat for existing scripts and extensions that were using these APIs? Just some notes in the RN? Is that enough?

vmassol: There's plenty that we can do (if we want). Some ideas:
* Warn in the DW about the change
* On a DW step, search using SOLR all pages where "getURLContent" is found and explain the change
* In getURLContent() return a string explaining the change instead of returning "" (but only do this if upgrading from a version where the behavior was different - need to access the info about the previous version from the perm dir somewhere).

They're all pretty complex though and we don't have the manpower to do any of them so I agree that realistically RN + update doc is only what we can afford.

aubincleme: > On a DW step, search using SOLR all pages where "getURLContent" is found and explain the change

This looks like a great idea indeed, maybe we should consider adding a new feature to the DW that checks for API calls in wiki pages that have changed since the last upgrade. I've created [XWIKI-14788](https://jira.xwiki.org/browse/XWIKI-14788) for that.

PR #616, opened 2017-11-18
XWIKI-14808: Replace OpenJS Keyboard JS with Keypress JS for handling keyboard shortcuts

aubincleme:
* Add (again) the Keypress JS library
* Fix regression that was overriding browser shortcuts
* Allow sequence shortcuts to be triggered without triggering combo shortcuts

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
@@ -381,24 +381,24 @@
   <script type="text/javascript">
   //<![CDATA[
     #if ($hasEdit && !$isReadOnly)
-      shortcut.add("$services.localization.render('core.shortcuts.view.edit')",function() { location.href=$('tmEdit').down('a').href; }, { 'type':'keypress', 'propagate':false, 'disable_in_input':true });
+      shortcut.add("$services.localization.render('core.shortcuts.view.edit')",function() { location.href=$('tmEdit').down('a').href; }, { 'disable_in_input':true });
       #if($isAdvancedUser)
-        shortcut.add("$services.localization.render('core.shortcuts.view.wiki')",function() { location.href=$('tmEditWiki').href; }, { 'type':'keypress', 'propagate':false, 'disable_in_input':true });
-        shortcut.add("$services.localization.render('core.shortcuts.view.wysiwyg')",function() { location.href=$('tmEditWysiwyg').href; }, { 'type':'keypress', 'propagate':false, 'disable_in_input':true });
-        shortcut.add("$services.localization.render('core.shortcuts.view.inline')",function() { location.href=$('tmEditInline').href; }, { 'type':'keypress', 'propagate':false, 'disable_in_input':true });
-        shortcut.add("$services.localization.render('core.shortcuts.view.rights')",function() { var editRights = $('tmEditRights'); location.href= editRights ? editRights.href : "$xwiki.getURL($spacePreferencesDocumentReference, 'admin', 'category=1')";}, { 'type':'keypress', 'propagate':false, 'disable_in_input':true });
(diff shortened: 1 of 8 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -943,235 +943,206 @@ function checkAdvancedContent(message) {
...
+     *
+     * @param shortcut_combination {string} the shortcut that should trigger the callback
+     * @param callback the function triggered by the shortcut
+     * @param opt optional associative array defining parameters for registering the shortcut
+     */
+    add: function(shortcut_combination, callback, opt) {
+        // Require Keypress JS to be fully loaded before registering the shortcut.
+        require(["$services.webjars.url('org.webjars.bower:Keypress', 'keypress.js')", 'jquery'], function(keypress) {
+
+            // If no options are defined, create a blank array
```
sdumitriu: Isn't it semantically more correct to use a map `{}` instead of an array `[]`?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
```diff
@@ -381,24 +381,24 @@
...
+        shortcut.add("$services.localization.render('core.shortcuts.view.wysiwyg')",function() { location.href=$('tmEditWysiwyg').href; }, { 'disable_in_input':true });
+        shortcut.add("$services.localization.render('core.shortcuts.view.inline')",function() { location.href=$('tmEditInline').href; }, { 'disable_in_input':true });
+        shortcut.add("$services.localization.render('core.shortcuts.view.rights')",function() { var editRights = $('tmEditRights'); location.href= editRights ? editRights.href : "$xwiki.getURL($spacePreferencesDocumentReference, 'admin', 'category=1')";}, { 'disable_in_input':true });
+        shortcut.add("$services.localization.render('core.shortcuts.view.objects')",function() { location.href=$('tmEditObject').href; }, { 'disable_in_input':true });
         #if($hasAdmin)
-          shortcut.add("$services.localization.render('core.shortcuts.view.class')",function() { location.href=$('tmEditClass').href; }, { 'type':'keypress', 'propagate':false, 'disable_in_input':true });
+          shortcut.add("$services.localization.render('core.shortcuts.view.class')",function() { location.href=$('tmEditClass').href; }, { 'disable_in_input':true });
```
Enygma2002: Did not follow the details, but what about backwards compatibility wrt. the option `propagate`? I am guessing that calls that explicitly set it to false will now behave differently and potentially cause problems, right?

aubincleme: The problem was raised when implementing the new library. In Keypress JS, there's no way to `'propagate': true` a keyboard shortcut.

But, it turns out that no shortcut on the platform is using the `'propagate': true` option, either by setting it explicitly to `false` or just because `propagate` is not defined and was set to `false` by default by OpenJS Keyboard Shortcut library (http://www.openjs.com/scripts/events/keyboard_shortcuts/#add_docs).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
aubincleme: Yep, will fix that, thanks :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
Enygma2002: That sounds good for the platform's code, but we can't say anything about contrib extensions (unless we check) or about other code 3rd party code that might have used the library. I guess it's fine at this point, as long as we make sure to include this information in the release notes. Thanks for the explanation.

PR #617, opened 2017-11-21
Added Solr sloppy phrase matching capability and config

benmegson:
I find this works well in my XWiki instance however I think I should set the "locale" rather than hard code the language to English. Wasn't sure how to achieve this so open to comment.

Also tried to boost by tag object independently of others but couldn't get this to work. Does the database schema need updating to allow this?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
@@ -40,6 +40,10 @@
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
   <content>{{velocity output="false"}}
+#set ($supportedLocales = $xwiki.getXWikiPreference('languages').split('\s*[|, ]\s*'))
+{{/velocity}}    
+    
+{{velocity output="false"}}
 #set ($__defaultSolrConfig = {
   'queryFields': {
     'DOCUMENT': 'title^10.0 name^10.0
@@ -51,14 +55,60 @@
     'OBJECT': 'objcontent',
     'OBJECT_PROPERTY': 'propertyvalue'
   },
-  'phraseFields': 'title_en^200.0 title__^200.0
-                   filename^45.0 
-                   doccontent_en^40.0 doccontent__^40.0
-                   attcontent_en^35.0 attcontent__^35.0 
-                   comment^25.0
-                   author_display^20.0 creator_display^20.0  attauthor_display^20.0'
-  ,
-  'phraseFieldSlop': '10',
+  'phraseFields': {
+    'DOCUMENT': "#expandField('title' '^400.0')
+                 #expandField('doccontent' '^80.0')
+                 filename^90.0 #expandField('attcontent' '^70.0')
+                 comment^50.0
+                 author_display^40.0 creator_display^40.0 attauthor_display^20.0",
+    'ATTACHMENT': "filename^90.0 #expandField('attcontent' '^70.0') attauthor_display^40.0",
+    'OBJECT': 'objcontent^10.0',
+    'OBJECT_PROPERTY': 'propertyvalue^10.0',
+    ''        : "#expandField('title' '^400.0')
+                 #expandField('doccontent' '^80.0')
+                 filename^90.0 #expandField('attcontent' '^70.0')
+                 author_display^40.0 creator_display^40.0
+                 comment^50.0 attauthor_display^20.0
+                 objcontent^10.0 propertyvalue^10.0"
+  },
+  'bigramPhraseFields': {
+    'DOCUMENT': "#expandField('title' '^200.0')
+                 #expandField('doccontent' '^40.0')
+                 filename^45.0 #expandField('attcontent' '^35.0')
+                 comment^25.0 
+                 author_display^20.0 creator_display^20.0 attauthor_display^10.0",
+    'ATTACHMENT': "filename^45.0 #expandField('attcontent' '^35.0') attauthor_display^20.0",
+    'OBJECT': 'objcontent^5.0',
+    'OBJECT_PROPERTY': 'propertyvalue^5.0',
+    ''        : "#expandField('title' '^200.0')
+                 #expandField('doccontent' '^40.0')
+                 filename^45.0 #expandField('attcontent' '^35.0')
+                 author_display^20.0 creator_display^20.0
+                 comment^25.0 attauthor_display^10.0
+                 objcontent^5.0 propertyvalue^5.0"
+  },
+  'trigramPhraseFields': {
+    'DOCUMENT': "#expandField('title' '^300.0')
+                 #expandField('doccontent' '^60.0')
+                 filename^67.5 #expandField('attcontent' '^52.5')
+                 comment^37.5 
+                 author_display^30.0 creator_display^30.0 attauthor_display^15.0",
+    'ATTACHMENT': "filename^67.5 #expandField('attcontent' '^52.5') attauthor_display^30.0",
+    'OBJECT': 'objcontent^7.5',
+    'OBJECT_PROPERTY': 'propertyvalue^7.5',
+    ''        : "#expandField('title' '^300.0')
+                 #expandField('doccontent' '^60.0')
+                 filename^67.5 #expandField('attcontent' '^52.5')
+                 author_display^30.0 creator_display^30.0
+                 comment^37.5 attauthor_display^15.0
+                 objcontent^7.5 propertyvalue^7.5"
+  },
+  'phraseFieldSlop': '15',
+  'bigramPhraseFieldSlop': '5',
+  'trigramPhraseFieldSlop': '10',
+  'boostQuery': ['property.XWiki.TagClass.tags_string','^1000.0'],
+  'tieBreaker': '0.1',
+  'minShouldMatch': '2',
   'sortFields': {
     'DOCUMENT': {
       'score': 'desc',
@@ -163,5 +213,13 @@
   #set ($discard = $__defaultSolrConfig.putAll($solrConfig))
 #end
 #set ($solrConfig = $__defaultSolrConfig)
+
+#macro (expandField $fieldName $fieldBoost)
+  ${fieldName}__${fieldBoost} ##
+  #foreach ($locale in $supportedLocales)
+    ${fieldName}_${locale}${fieldBoost} ##
+  #end
+#end
+    
 {{/velocity}}</content>
 </xwikidoc>
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
@@ -106,7 +106,6 @@
   'phraseFieldSlop': '15',
   'bigramPhraseFieldSlop': '5',
   'trigramPhraseFieldSlop': '10',
-  'boostQuery': ['property.XWiki.TagClass.tags_string','^1000.0'],
   'tieBreaker': '0.1',
   'minShouldMatch': '2',
   'sortFields': {
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
@@ -58,6 +58,7 @@
                    comment^25.0
                    author_display^20.0 creator_display^20.0  attauthor_display^20.0'
   ,
+  'phraseFieldSlop': '10',
   'sortFields': {
     'DOCUMENT': {
       'score': 'desc',
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
@@ -51,6 +51,13 @@
     'OBJECT': 'objcontent',
     'OBJECT_PROPERTY': 'propertyvalue'
   },
+  'phraseFields': 'title_en^200.0 title__^200.0
+                   filename^45.0 
+                   doccontent_en^40.0 doccontent__^40.0
+                   attcontent_en^35.0 attcontent__^35.0 
+                   comment^25.0
+                   author_display^20.0 creator_display^20.0  attauthor_display^20.0'
+  ,
   'sortFields': {
     'DOCUMENT': {
       'score': 'desc',
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
@@ -658,7 +658,7 @@ $services.localization.render('core.users.unknownUser')##
   #end
   #if ("$!phraseFieldsBoost" != '')
     #set ($discard = $query.bindValue('pf', $phraseFieldsBoost))
-    #set ($discard = $query.bindValue('ps', '10'))
+    #set ($discard = $query.bindValue('ps', $solrConfig.phraseFieldSlop))
   #end
 #end
 
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
@@ -617,6 +617,13 @@ $services.localization.render('core.users.unknownUser')##
   #set ($discard = $query.setLimit($rows))
   #set ($discard = $query.setOffset($start))
   #set ($discard = $query.bindValue('sort', "${sort} ${sortOrder}"))
+  #set ($discard = $query.bindValue('tie', $solrConfig.tieBreaker))
+  #set ($discard = $query.bindValue('mm', $solrConfig.minShouldMatch))
+  ## Little rudimentary but this should impart higher relevance to
+  ## documents with tags matching the search terms. Will only match single word tags at the moment
+  #set ($fieldName = $solrConfig.boostQuery[0])
+  #set ($fieldBoost = $solrConfig.boostQuery[1])
+  #set ($discard = $query.bindValue('bq', "${fieldName}:(${queryString})${fieldBoost}"))    
   #setQueryFields($query)
   #setPhraseFields($query)
   #setFacetFields($query)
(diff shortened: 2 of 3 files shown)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
```diff
@@ -51,6 +51,13 @@
     'OBJECT': 'objcontent',
     'OBJECT_PROPERTY': 'propertyvalue'
   },
+  'phraseFields': 'title_en^200.0 title__^200.0
```
mflorea: You should use just "title" which gets expanded automatically to "title_en", like for the query fields.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
```diff
@@ -51,6 +51,13 @@
...
+                   filename^45.0 
+                   doccontent_en^40.0 doccontent__^40.0
+                   attcontent_en^35.0 attcontent__^35.0
```
mflorea: Same, you should use just "doccontent" and "attcontent", and they will get expanded.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
```diff
@@ -51,6 +51,13 @@
     'OBJECT': 'objcontent',
     'OBJECT_PROPERTY': 'propertyvalue'
   },
+  'phraseFields': 'title_en^200.0 title__^200.0
+                   filename^45.0 
+                   doccontent_en^40.0 doccontent__^40.0
+                   attcontent_en^35.0 attcontent__^35.0 
+                   comment^25.0
+                   author_display^20.0 creator_display^20.0  attauthor_display^20.0'
+  ,
```
mflorea: I prefer to have the phrase fields set by result type, as we do with the query fields.

mflorea: Can't we use the value of the query fields to set the phrase fields? i.e. have by default query fields = phrase fields. This means we don't specify phrase fields in the default configuration but we check for it below so that users can have different phrase fields than query fields.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -645,6 +646,21 @@ $services.localization.render('core.users.unknownUser')##
...
+    ## If the value of the 'queryFields' parameter is a string then it means that the same query fields are used for
+    ## all result types.
+    #set ($phraseFieldsBoost = $solrConfig.phraseFields)
+  #else
+    ## There are different query fields for each result type.
+    #set ($phraseFieldsBoost = $solrConfig.phraseFields.get($type))
+  #end
+  #if ("$!phraseFieldsBoost" != '')
+    #set ($discard = $query.bindValue('pf', $phraseFieldsBoost))
+    #set ($discard = $query.bindValue('ps', '10'))
```
mflorea: We could make the phrase slop configurable also.

mflorea:
> Also tried to boost by tag object independently of others but couldn't get this to work.

See http://design.xwiki.org/xwiki/bin/view/Design/SolrSchema#HDocumentFields . Look for ``object.XWiki.TagClass_fr``.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: Same here. I couldn't get it to expand, hence why I hard coded it. If I didn't explicitly specify as I've done then the terms wouldn't appear at all in the query string as displayed in debug mode.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
benmegson: True. I'll implement.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: I couldn't get it to expand, hence why I hard coded it. If I didn't explicitly specify as I've done then the terms wouldn't appear at all in the query string as displayed in debug mode.

mflorea: The code that is responsible for expanding the field name is [this](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-plugin/src/main/java/org/xwiki/search/solr/internal/XWikiDismaxQParserPlugin.java#L253). Indeed, it adds the alias only for the ``qf`` (query fields) parameter. Would be great if you could check if the ``pf`` parameter supports aliases (e.g. specified with ``f.%s.pf``).

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: Marius, what do you mean by result type exactly? I tried to group as you did.

I did try query fields = phrase fields initially, but by gradual experimentation found that some fields were skewing my results the wrong way. For example *_raw would boost documents with a large number of links that weren't as relevant as the docs they pointed too. Maybe we could set the weighting to be similar, I wasn't sure if the qf score was summed with the pf score hence why I went for large numbers so that pf would always trump qf. Furthermore some of the shorter fields like author didn't really make sense to me as places to do phrase matching.

mflorea: OK, so indeed it makes more sense to have separate values (fields and weights) between ``qf`` and ``pf``. By "result type" I meant "document", "attachment", "object", "object property". So I meant having separate ``pf`` value when searching for documents than when searching for attachments. For instance attachment rows in the index don't have the ``doccontent`` field so it doesn't make sense to list it in the ``pf`` value when you look for attachments.

benmegson: > I meant "document", "attachment", "object", "object property"
Ah! I see... I did try this but had to merge into a single string for some reason. I'll try and figure it out and post back. Believe it had something to do with the fact that when searching across all result types the pf type parameters weren't being concatenated. 

On the subject of result types. I couldn't select combinations of result types from the facet menu, e.g. document and attachment. Is this behaviour by design?

mflorea: > Is this behaviour by design?

Yes, see this [comment](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml#L657). In order to have multiple selection we would need to use the same query fields for all result types. We don't do this because some fields are shared between results types and their importance (weight) depends on the result type. For instance the ``filename`` is very important when searching for attachments and less important when searching for documents. But if you search for both attachments and documents then the ``filename`` should have the weight close to the weight of the document name.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: So... Done some research on the web and in the Solr documentation and aliasing doesn't seem to be possible for `pf` fields. Unsure what the best way around this is.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: > By "result type" I meant "document", "attachment", "object", "object property". So I meant having separate pf value when searching for documents than when searching for attachments

@mflorea, if I convert this to be by result type, how do I get it to also work when searching across all entities. As mentioned previously,

> when searching across all result types the pf type parameters weren't being concatenated.

If I implement as suggested we would lose all `pf` bosting when unselecting the facets. I think there should be at least some basic `pf` bosting when doing a search across all result types.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
benmegson: done

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
mflorea: OK, thanks for the investigation. In this case the solution I see is to write a small Velocity macro in ``SolrSearchConfig`` that gets as input parameters the field name (e.g. "title") and the weight (e.g. 200) and outputs ``title_ab^200 title_cd^200 ... title_yz^200 title__^200`` by reading the list of supported languages (locales) from the XWiki preferences. You can then call this macro like this:

    'phraseFields': "#expandField('title' 200) filename^45 #expandField('doccontent' 40) ..."

WDYT?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
mflorea: We could map the phrase/query fields value for the "all result types" search to the empty key:

    'queryFields': {
        'DOCUMENT': 'title^10.0 name^10.0 ...',
        ...,
        '': '<the value to be used when there is no result type selected>'
    }

I think this will work because when there is no result type selected the ``$type`` variable (in ``SolrSearchMacros``) is null so writing this should work:

    #set ($boost = $solrConfig.queryFields.get("$!type"))

Using silent notation to get the empty string when ``$type`` is null. WDYT?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: Sounds good. I think I can manage that.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: Sounds like a good solution

benmegson:
Marius,
Been having a play around with boosting docs based on tag text over the weekend. Getting some good results but can't figure out the best way to compose the expanded multiterm search string. I presume `#set ($query = $services.query.createQuery($queryString, 'solr'))` does this normally. I've found I need to sum a score to the main score for this to work correctly, hence why I want to add a boost query parameter or a re-rank query as shown below.

Very rough code I'm adding to SolrSearchMacros for testing purposes:
```
 #set ($discard = $query.bindValue( 'rqq', "(property.XWiki.TagClass.tags_string:(${queryString}))" ))
 #set ($discard = $query.bindValue( 'rq', "{!rerank reRankQuery=$rqq reRankDocs=1000 reRankWeight=30}&rqq={}}" ))
```
```
#set ($discard = $query.bindValue('bq', "property.XWiki.TagClass.tags_string:(${queryString})^=20"))
```

How do I get the `$queryString` parameter to expand to something like `((term1) or (term2) or (term3) or ... )`?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
benmegson: Marius,
Can't figure out how to get the list of locales.
I've tried:
```
#set ($supportedLocales = $xwiki.getXWikiPreference('locales'))
#set ($supportedLocales = $services.localization.availableLocales)
#set ($supportedLocales = $xwiki.supportedLocales)
```
What am I missing?

mflorea:
This works for me:

    #set ($discard = $query.bindValue('bq', 'property.XWiki.TagClass.tags_string:("fake news" book)^20'))

The expanded query shown with ``&debug=true`` in the search URL looks like:

    +(<expanded query here>) (property.XWiki.TagClass.tags_string:fake news property.XWiki.TagClass.tags_string:book)^20.0

As you can see the match for the actual query is mandatory (specified by the + sign) while the match for the boost query is optional. If you know your tags don't contain spaces then you can simply pass the list of tokens, space separated like this:

    property.XWiki.TagClass.tags_string:(tag1 tag2 tag3)^20

. The ``OR`` operator is implicit here. So if ``$queryString`` is a string like "tag1 tag2 tag3" you can pass it directly

    property.XWiki.TagClass.tags_string:($queryString)^20

If the tags can contain spaces then you need to wrap your tags in quotes. I haven't played with ``rqq`` and ``rq`` but ``bq`` works for me. Here's an example of how the score is influenced:

    5.7536416 = sum of:
      5.7536416 = weight(property.XWiki.TagClass.tags_string:book in 0) [SchemaSimilarity], result of:
        5.7536416 = score(doc=0,freq=1.0 = termFreq=1.0
    ), product of:
          20.0 = boost
          0.2876821 = idf, computed as log(1 + (docCount - docFreq + 0.5) / (docFreq + 0.5)) from:
            1.0 = docFreq
            1.0 = docCount
          1.0 = tfNorm, computed as (freq * (k1 + 1)) / (freq + k1) from:
            1.0 = termFreq=1.0
            1.2 = parameter k1
            0.0 = parameter b (norms omitted for field)

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchConfig.xml
mflorea: This should work:

    #set ($supportedLocales = $xwiki.getXWikiPreference('languages').split('\s*[|, ]\s*'))

The preference name is "languages" and not "locales" because it's an old preference that we didn't rename in order to not break backward compatibility.

benmegson:
OK, thanks Marius.

I was actually hoping to be able to construct a proper query. I.e. be able to pump a subquery (formed from the original query string) into the bq parameter with its own `qf` and `pf`, `pf2`, `pf3` parameters.

This would allow us to find the most relevant tag that can be formed from the combination of input strings.

Something like:
```
$BQ_query.bindValue(q, "user query string")
$BQ_query.bindValue(qf, 'foo')
$BQ_query.bindValue(pf2, 'bar')
$query.bindValue('bq', "property.XWiki.TagClass.tags_string:(${BQ_query.expandedQueryString})^=20")
```

Do you think this could work??

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -617,7 +617,15 @@ $services.localization.render('core.users.unknownUser')##
   #set ($discard = $query.setLimit($rows))
   #set ($discard = $query.setOffset($start))
   #set ($discard = $query.bindValue('sort', "${sort} ${sortOrder}"))
+  #set ($discard = $query.bindValue('tie', $solrConfig.tieBreaker))
+  #set ($discard = $query.bindValue('mm', $solrConfig.minShouldMatch))
+  ## Little rudimentary but this should impart higher relevance to
+  ## documents with tags matching the search terms. Will only match single word tags at the moment
+  #set ($fieldName = $solrConfig.boostQuery[0])
+  #set ($fieldBoost = $solrConfig.boostQuery[1])
+  #set ($discard = $query.bindValue('bq', "${fieldName}:(${queryString})${fieldBoost}"))
```
mflorea: I think the boost query makes sense if it's different than the main query. E.g. the user looks for "books" and you want to boost the books from the "science-fiction" category (even though the user didn't ask for it). In your case, since the boost query string is the same as the main query string, I think it's enough to add the field with the boost to the query fields. Don't you think?

    'queryFields': {
        'DOCUMENT': 'title^10.0 name^10.0 ... property.XWiki.TagClass.tags^1000.0 ...'

which basically means: "if the search term matches a tag then increase the score". Isn't this what you want?

mflorea:
Regarding the boost query, there's no API on XWiki side (``$services.query.*`` is an XWiki API) to get the expanded query as you want. Besides that, I don't think that on Solr side the boost query supports parameters. AFAIK the boost query is a parameter itself of the main query and it's not an actual sub-query, meaning that for instance you can't specify the query fields parameter for the boost query.

So no, I don't think you can pass a fully fledged query as the value of the boost query parameter.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
benmegson: Well, I'll run you through what I would like to achieve, then it might make more sense. I wasn't convinced the `bq` parameter was the best way to go either. Hence why I was playing around with rerank.

When entering a multi term query such as `cocktail bars in New York` I'd like to check for documents with tags made up of the search terms and boost these to be the most relevant without changing the order of the results as determined by the main query. To do this I believe I need to sum a fixed boost value to the score rather than multiply the score by the boost. If more than one tag matches then multiple boost values would need summing.

In shorthand, this (sub)query would look like this:
```
q=cocktail bars in New York
tie=1.0
qf=property.XWiki.TagClass.tags^=1000
pf=property.XWiki.TagClass.tags^=1000
pf2=property.XWiki.TagClass.tags^=1000
pf3=property.XWiki.TagClass.tags^=1000
```
If tags existed for `cocktail bars` and `New York` the useful terms from the `bq` subquery that would contribute to highest scoring documents would be: `((cocktail bars)^=1000) ((New York)^=1000)`

All documents with both tags would appear as most relevant due to `1000 + 1000 + (main query document score)`. Then documents matching one tag, followed by documents with no matching tags. Note that the order of the documents within each of these "groups" (2 tags, 1 tag, no tags) is determined by the main query score.

I think I could achieve a simplified version of this type of query (without stopword removal) by composing the query using a velocity macro. But it woudn't be very elegant. Hence why I was wanting to make use of solr's natural capabilities.

The `1000` value is a little bit of a hack, but it doesn't really matter as long as it's a value larger than the highest scoring result. There might be a dynamic way of getting this value.

benmegson:
On a separate note, I've noticed quite a few documents in my XWiki instance seem to have no language set. This is messing up the stopwords filter. What causes a doc to have no language set?

mflorea:
Technical documents should have no language set because they contain code (e.g. pages that define the XWiki UI, like the search page, the live table, the administration, etc.). It doesn't make sense to set the language to EN for these pages because it would mean we could translate them in other language like FR. All content pages should have the language set. By default when you create a new wiki page it has the language set. So normally all the pages you create should have the language set.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
mflorea: I see. I don't know if you can achieve this with boost query (I haven't encountered this use case before). I guess you tried to use boost query with [local parameters](https://lucene.apache.org/solr/guide/6_6/local-parameters-in-queries.html) and it didn't work? Something like:

    $query.bindValue('bq', "{!dismax tie=1.0 qf=property.XWiki.TagClass.tags^1000 ...}$queryString")

And no, we don't have an API to serialize a Solr query to a string with local parameters like above.

benmegson:
That's what I thought, but it seems that inserting a velocity macro into a page causes it to become "technical", for example, all the docs with a table of contents in them don't have a language set. Bug?

mflorea:
I don't understand the relation between "velocity macro" and "table of contents". They are 2 different macros. Inserting a macro (either from the WYSIWYG editor or from the Wiki editor) into the page content doesn't affect the page language. Changing the page content doesn't affect the page language. So there must be something else. Is this a page created from a template or is it a plain (blank) wiki page?

benmegson:
You're right... It's something to do with templates. Documents created with an old template had the language set but there must be something in my new template that's causing the lack of language. Where do I need to go to check this? and how do I change the language of the current docs? I realise this is probably best discussed on the forum, would you like me to move it there?

mflorea:
Yes, we should discuss this on the forum.

benmegson:
Right... I'm happy with this now. @mflorea let me know what you think.

This should address: https://jira.xwiki.org/browse/XWIKI-14938

Regarding the tag boosting. I've managed to implement a custom solution that works well for me. I'll share it in another pull request if you think it would be beneficial to others.

I've started the language issue discussion on the forum: https://forum.xwiki.org/t/pages-language-not-set-when-created-from-tempate/2509

mflorea:
Looks good. I'll merge it today after we release 10.0.

mflorea:
Thanks for your contribution!

PR #618, opened 2017-11-29
XWIKI-9702: In User Profile, use term 'About' instead of 'Comment' for the free text field

ykarim:
As a GCI task, link: https://jira.xwiki.org/browse/XWIKI-9702


tmortagne:
Updating the translations was not required for this task (translation task are actually forbidden in GCI).

Note that any change made to the translations (so all langues other than English) made trough a pull request is going to be overwritten during next release since the reference is http://l10n.xwiki.org.

Also for example the French translation is not right for "About" so IMO you should just keep the English fix.

tmortagne:
By the way I can't find your GCI task instance anymore, not even in the abandoned tasks list.

evalica:
The task is https://codein.withgoogle.com/dashboard/tasks/6355213571063808/ but I don't think it has been correctly claimed. 

Regarding your pull request, the idea is that this needs to be manually changed only in English, since the other languages are dynamically updated during the release process.

vmassol:
I think we could improve the task desc since it wasn't clear to me too when I read it.

PR #622, opened 2017-12-02
Update Translations.pl.xml

USA2021:
contains 2nd and 3rd edit


Enygma2002:
This change is incorrectly overwriting a Polish translations file with English values.

PR #623, opened 2017-12-02
Update Translations.pl.xml

USA2021:
all edits done


Enygma2002:
This change is incorrectly overwriting a Polish translations file with English values.

PR #628, opened 2017-12-02
Update ApplicationResources.properties

USA2021:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -4892,7 +4892,7 @@ platform.appwithinminutes.appNameInvalidClassNameError=We can't extract a valid
 
 ## The "type" property has been removed and data migrated to the new "terminal" property.
 xe.templateprovider.templatetype=Template type
-xe.templateprovider.templatetype.info=Whether this template should be used for creating generic pages or is specific to space homepages
+xe.templateprovider.templatetype.info=Whether this template should be used for creating generic pages or is specified to spacing homepages
```
Enygma2002: This does not apply in the context where this translation is displayed. Spaces are a concept in XWiki, under which pages are stored.

USA2021: oh

USA2021: can you please tell me how many errors are left that I need to find?

tmortagne: > can you please tell me how many errors are left that I need to find?

The thing is we don't really know, we did not exactly left mistake on purpose for GCI students to find them. We just know there must be various errors and typo (because statistically there is or someone remembered seeing some typos when using XWiki but then forgot to fix them) :)

PR #630, opened 2017-12-04
 [codestyle] Improve readability

akshitdewan:
Removed inline conditional and merged if statements.


tmortagne:
Thanks for your effort !

Unfortunately we cannot accept this pull request:

* The first change does not really  improve readability (it actually make it worst since you did not followed the official XWiki codestyle for the code you added, I added more explicit indication about that in the task)
* You changed all white spaces with tabs in the second file. Using white space is clearly not a violation listed by Sonar
* "[codestyle] Improve readability" is just an example, it's much better to put something more precise that you can find in the Sonar explanation

Please consider looking at Blocker/Critical/Major violations (which contains a lot less false positives than minor issues) only and, as indicated on the issue, don't hesitate to ask on the chat when you are not sure.

akshitdewan:
Okay, I will work on some more!

PR #631, opened 2017-12-04
Update Translations.xml

akshitdewan:
Fixed grammatical errors - can not --> cannot,  which --> who, will starting --> will start


vmassol:
Thanks!

PR #633, opened 2017-12-05
XWIKI-9702 : In User Profile, use term 'About' instead of 'Comment' for the free text field

SrijanJha28:
Update ApplicationResources.properties by changing 'Comment' to 'About'.


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3269,7 +3269,7 @@ XWiki.XWikiRights_users=Users
 XWiki.XWikiUsers_active=Active
 XWiki.XWikiUsers_password=Password
 XWiki.XWikiUsers_email=Email
-XWiki.XWikiUsers_comment=Comment
+XWiki.XWikiUsers_about=About
```
vmassol: I think this will fail the User application since the key is being modified.

SrijanJha28: What should i do to pass it?

tmortagne: "XWiki.XWikiUsers_comment" is the key which is used by the code to retrieve the translation corresponding to the pretty name of the field `comment`. We don't want to change the field id (changing the id of an old field might break a lot of code), just the corresponding pretty name you can see in the UI.

evalica: As stated in the issue, only the value needs to be changed, not the key.

evalica:
Thanks for your contribution!

PR #634, opened 2017-12-05
XWIKI-14784 : Display the user avatar instead of just the name in the "Author" columns of the Livetables

SrijanJha28:
Update ApplicationResources.properties by replacing the name with the user avatar.


evalica:
This PR is not valid. Also the CI (Continuous Integration) build fails. 

You also committed a change that you've fixed in a previous PR (the About/Comment change).

The solution is not in editing the translations. You need to look at the livetable code inside the Index Application and change the Author, Deleted by, etc. columns to use the user displayer.

PR #636, opened 2017-12-06
[codestyle] Added hashcode method because class has equals

akshitdewan:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
```diff
@@ -2161,7 +2162,22 @@ public boolean equals(java.lang.Object arg0)
...
+            .append(this.initialDoc)
+            .append(this.doc)
+            .append(this.currentObj)
+            .append(this.currentMixedDocumentReferenceResolver)
+            .append(this.defaultEntityReferenceSerializer)
+            .append(this.localEntityReferenceSerializer)
+            .append(this.compactWikiEntityReferenceSerializer)
+            .append(this.documentRevisionProvider)
+            .append(this.authorizationManager)
+            .toHashCode();
```
vmassol: Adding some hashcode is always tricky so we would also need you to add some tests to prove that it works. Now I can see for sure that it's not correct :) since for example injected components have no reason to be class model data (they're are simply services used by the class).

tmortagne:
This class is a bit tricky, the equals() itself does not make much sense and I actually doubt it's really used... Better leave this one alone for now, there is other use cases like this one which are much easier.

In general when you implement hashCode you usually manipulate the same data than equals (since according to Java specifications they are supposed to be in sync).

PR #637, opened 2017-12-07
Add xwiki.authentication.useip property

akshitdewan:
XWIKI-14839 : xwiki.authentication.useip should be mentioned in xwiki.cfg

### DIFF
--- a/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
+++ b/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
@@ -370,7 +370,8 @@ xwiki.authentication.unauthorized_code=200
 #-# Must be set to "empty".
 # xwiki.authentication.createuser=empty
 
-#-# If set to true(the default value), cookies of users are blocked from being used except by the same IP address which got them.  
+#-# If set to true(the default value), cookies of users are blocked from being used except by the same IP address 
+#-# which got them.  
 # xwiki.authentication.useip=true
 
 #---------------------------------------
--- a/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
+++ b/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
@@ -370,8 +370,7 @@ xwiki.authentication.unauthorized_code=200
 #-# Must be set to "empty".
 # xwiki.authentication.createuser=empty
 
-#-# If set to true, you can block the cookies of users from being used except by the same IP address which got them. 
-#-# except by the same IP address which got them.
+#-# If set to true(the default value), cookies of users are blocked from being used except by the same IP address which got them.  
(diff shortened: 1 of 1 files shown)


xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
```diff
@@ -370,6 +370,10 @@ xwiki.authentication.unauthorized_code=200
 #-# Must be set to "empty".
 # xwiki.authentication.createuser=empty
 
+#-# If set to true, you can block the cookies of users from being used except by the same IP address which got them. 
+#-# except by the same IP address which got them.
+# xwiki.authentication.useip=true
```
vmassol: I don't think that the main reason for useip. It seems to me more a side-effect/consequence. Also you have duplicated content.

tmortagne: Actually Akshit is right, that's exactly what this property is for: for security reason we restrict the cookie to the same IP by default and it can be disabled with this property.

You should more explicitly indicate what is the default behavior since it feel like it's disabled by default when reading what you wrote. Also you repeat twice "except by the same IP address which got the".

vmassol:
BTW you shouldn't say "you can block". Does it mean that I can decide to not block it? How do I decide. Maybe you meant to say "If true the cookies from users are blocked..."?

Also I still don't understand how something named "useip" is used to block cookies. It would have been called "blockcookies" if that were the case. So I'm pretty sure that "useip" must have something to do with using the IP somwhere rather than being used solely to block cookies.

vmassol:
Maybe something like:

"We use a token to validate cookie information, containing hashed information about the other cookies and a secret parameter, optionally binding with the current IP of the user (so that the cookie cannot be reused on another computer) and controlled by this parameter. By default it's set to true."

tmortagne:
"useip" means "use the ip to encrypt the credentials in the cookie". It's a bad naming (use the technical detail instead of the goal).

> Does it mean that I can decide to not block it? 

If you put false in that property then the credential from the cookie will be used even if your client's IP changes.

vmassol:
> If you put false in that property then the credential from the cookie will be used even if your client's IP changes.

I wasn't commenting on that but on the words used by Akshit :) Again: "If set to true, you can block the cookies of users". This means that even if set to true you can also decide to NOT block the cookies. What I said:

> BTW you shouldn't say "you can block". Does it mean that I can decide to not block it? How do I decide. Maybe you meant to say "If true the cookies from users are blocked..."?

xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
vmassol: I think I'd prefer explaining with more details. I made a proposal in comment earlier, could you check it out?

I'm also not sure that the way it's formulated is clear enough. I don't really understand it personally (could be an issue with English).

Thanks, we're getting closer :)

tmortagne:

PR #638, opened 2017-12-07
[codestyle] added hashCode method

akshitdewan:
hashCode supplements equals method

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Object.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Object.java
@@ -19,6 +19,8 @@
  */
 package com.xpn.xwiki.api;
 
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
 import com.xpn.xwiki.XWikiContext;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Object.java
```diff
@@ -102,6 +104,15 @@ public boolean equals(java.lang.Object arg0)
...
     }
+    
+    @Override
+    public int hashCode()
+    {
+        return new HashCodeBuilder(3, 17)
+            .append(getXWikiContext())
+            .append(element)
+            .toHashCode();
+    }
```
vmassol: This doesn't look correct. For ex I wouldn't put the xwiki context in it.

Also please provide some test cases when you modify equals or hashcode to prove that the changes work. Thanks!

tmortagne: Akshit please forget all the classes in com.xpn.xwiki.api package and any other weird equals that includes test on the context...

tmortagne:
Thanks for your contribution but I have to close this as invalid. Please use it as a false positive.

PR #639, opened 2017-12-08
XWIKI-11038 : title in desc

akshitdewan:
"xwiki.authentication.unauthorized_code has a bad default value and is not taken into account"
not sure if this is an issue, could not understand what Thomas said, "Note that I just tested to fail the login and got 403 so additionally it's not really fully taken into account (I don't really see the point of making it configurable anyway)".

### DIFF
--- a/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
+++ b/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.cfg.vm
@@ -363,7 +363,7 @@ xwiki.authentication.logoutpage=(/|/[^/]+/|/wiki/[^/]+/)logout/*
 # xwiki.hidelogin=false
 
 #-# HTTP status code to sent when the authentication failed.
-xwiki.authentication.unauthorized_code=200
+xwiki.authentication.unauthorized_code=403
 
 #-# Used by some authenticators (like com.xpn.xwiki.user.impl.xwiki.AppServerTrustedAuthServiceImpl)
 #-# to indicate that the users should be created. In this kind of authenticator the user are not created by default.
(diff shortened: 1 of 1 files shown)


tmortagne:
> could not understand what Thomas said

This means you did not really tested it :)

What I said is that if you fail to login (use the wrong credentials for example, I don't remember what I used back then) you don't get what's indicated in xwiki.cfg. That being said this configuration does not make much sense IMO and at the very least it should be commented by default.

akshitdewan:
I'm not clear on what I need to do, could you elaborate? @tmortagne

tmortagne:
Several things:

* before fixing anything, try to reproduce this issue: use wrong credentials and see what error code the server return (you can get that with the dev tools provided by your browser)
* it does not make any sense to me that this property is set by default in xwiki.cfg, this property should be commented and only used in last resort

akshitdewan:
reproduced and got 403 code. I added one # for the comment, but I'm not sure if you mean #-#.

tmortagne:
> reproduced and got 403 code

If you got 403 despite the fact that it was configured to return 200 it means the property is not properly taken into account which should be fixed.

akshitdewan:
@tmortagne could you explain please?

tmortagne:
You have a configuration in xwiki.cfg which indicate that you should get error code 200 and you end up with 403. You don't think there is something wrong about it ? :)

akshitdewan:
Sorry, let me clarify. I understand that the configuration doesn't work but I don't know how I would go about fixing it. Could I get some guidance?

tmortagne:
You should search for "xwiki.authentication.unauthorized_code" in the codebase and see how it's used.

You can take a look at https://codein.withgoogle.com/dashboard/tasks/5662395449999360/ to learn how to debug Java code in Eclipse.

PR #640, opened 2017-12-11
[misc] xar format

RobertVlp:


tmortagne:
So is it open or closed ? :)

acotiuga:
There is also #641 which was merged, so it is closed :)

PR #646, opened 2017-12-12
Update Translations.xml

akshitdewan:
"Like" implies that there are many more options, thus no etc. ; colon was unnecessary and awkward, could have added if we removed the "like".
Should not pause after "The button" so no comma. The second comma was also unnecessary


tmortagne:

PR #648, opened 2017-12-12
Update Translations.xml

akshitdewan:
Mail is a non count noun - verb form is singular


xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-ui/src/main/resources/Mail/Translations.xml
```diff
@@ -67,7 +67,7 @@ Mail.SendMailConfigClass_properties.hint=The list of additional JavaMail propert
 Mail.SendMailConfigClass_sendWaitTime=Send Wait Time
 Mail.SendMailConfigClass_sendWaitTime.hint=The delay to wait between each mail being sent, in milliseconds (defaults to 8 seconds)
 Mail.SendMailConfigClass_discardSuccessStatuses=Discard Success Statuses?
-Mail.SendMailConfigClass_discardSuccessStatuses.hint=Whether mail statuses are kept or not for mail that have been sent successfully (defaults to true)
+Mail.SendMailConfigClass_discardSuccessStatuses.hint=Whether mail statuses are kept or not for mail that has been sent successfully (defaults to true)
```
tmortagne: Should the fix be to use "mails" instead of "mail" ? I feel like that was the intent here.

akshitdewan: I looked into that but mail does not have a plural form.

vmassol: Interesting. Found this: http://www.nytimes.com/2010/05/09/magazine/09FOB-onlanguage-t.html

I still think it would make more sense to use the plural. It looks really strange with singular version IMO. Would be great if some English-native dev or user could help us on this :)

akshitdewan: I think singular sounds more natural, but we can get a second opinion.

vmassol:
I’ve read it again and indeed the singular form looks ok to me too. So whatever the majority prefer :)

tmortagne:
Welcome !

tmortagne:

PR #649, opened 2017-12-12
Update Translations.xml

akshitdewan:
in which with second in is iffy


tmortagne:

PR #650, opened 2017-12-12
Update Translations.xml

akshitdewan:
What are lines 134 and 136 trying to say? 
Is 169 saying that the macro should not be used here for compatibility reasons or that it is here for compatibility reasons? Not sure which one but if it is the latter, they should be separate sentences.
I think 230 and 231 be consistent


tmortagne:

PR #652, opened 2017-12-15
[misc] xar format

RobertVlp:
Update the XML document format of 10 committed wiki pages


tmortagne:

PR #653, opened 2017-12-15
[Misc] xar format

Dagmmar:
Update the XML document format of 5 committed wiki pages


acotiuga:
Thanks for your contribution!

PR #659, opened 2017-12-20
formatted xml files

akshitdewan:
removed help file from pr altogether, as it always seems to download with very different code

PR #660, opened 2017-12-21
xml version updating

akshitdewan:


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/Treeview.xml
```diff
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
```
tmortagne: Please execute `mvn xar:format` twice.

tmortagne:

Enygma2002:
@tmortagne:There were some problems with this PR:
* 5 pages were already included in the previous PR https://github.com/xwiki/xwiki-platform/pull/659/files.
* there are changes on translation pages (other than the default language)
* there are changes on date fields

IMO, we should revert this PR as well and allow @akshitdewan to redo it.

akshitdewan:
@Enygma2002 my apologies, I undid those 5 pages here because I forgot to make a new branch but they seem to have reappeared somehow. Nonetheless, it shouldn't affect anything because they are both merged right? Also, should I not be changing any translation pages? the bulk of the next PR is also translation pages. Finally, what is the point of excluding the update dates since it was, in fact, updated on this new date?

PR #665, opened 2017-12-23
XWiki-12180 : ScriptMimeMessage#addPart should support BodyPart

akshitdewan:
I'm not completely sure if this is the way you wanted it, let me know what you were looking for.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -101,8 +101,8 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
-     *        of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime 
+     *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
      *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
      *        accepts a {@link BodyPart} object, in which case, the mime type is not used.
      */
@@ -117,13 +117,13 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
-     *        of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime 
+     *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
      *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
      *        accepts a {@link BodyPart} object, in which case, the mime type and parameters are not used.
      * @param parameters the list of extra parameters. This is used for example to pass alternate content for the mail
-     *        using the {@code alternate} key in the HTML Mime Body Part Factory. Mail headers can also be passed
-     *        using the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys
+     *        using the {@code alternate} key in the HTML Mime Body Part Factory. Mail headers can also be passed using 
+     *        the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys
      *        and values.
      */
     public BodyPart addPart(String mimeType, Object content, Map<String, Object> parameters)
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -101,10 +101,10 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime 
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
      *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
-     *        accepts a {@link BodyPart} object, in which case, the mime type is not used.
+     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     *        Also accepts a {@link BodyPart} object, in which case, the mime type is not used.
      */
     public BodyPart addPart(String mimeType, Object content)
     {
@@ -117,10 +117,10 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime 
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
      *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
-     *        accepts a {@link BodyPart} object, in which case, the mime type and parameters are not used.
+     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. 
+     *        Also accepts a {@link BodyPart} object, in which case, the mime type and parameters are not used.
      * @param parameters the list of extra parameters. This is used for example to pass alternate content for the mail
      *        using the {@code alternate} key in the HTML Mime Body Part Factory. Mail headers can also be passed using 
      *        the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -85,7 +85,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
      *
      * @param bodyPart a {@link BodyPart} object to include in the mail
-     * @since 10.0-SNAPSHOT
+     * @since 10.0RC1
      */
     @Unstable
     public void addPart(BodyPart bodyPart)
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -87,7 +87,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param bodyPart a {@link BodyPart} object to include in the mail
-     * @since 
+     * @since 10.0-SNAPSHOT
      */
     @Unstable
     public BodyPart addPart(BodyPart bodyPart)
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -86,7 +86,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail
+     * @param content the content to include in the mail, also directly accepts BodyPart objects
      */
     public BodyPart addPart(String mimeType, Object content)
     {
@@ -95,11 +95,11 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
 
     /**
      * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
-     * Directly supports BodyPart objects instead of the content as well.
+     *
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail
+     * @param content the content to include in the mail, also directly accepts BodyPart objects
      * @param parameters the list of extra parameters. This is used for example to pass alternate content for the mail
      *        using the {@code alternate} key in the HTML Mime Body Part Factory. Mail headers can also be passed using
      *        the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys and values
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -82,7 +82,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
     
     /**
      * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
-     * This method can be used for scripting languages other than Velocity
+     *
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param bodyPart a {@link BodyPart} object to include in the mail
@@ -101,8 +101,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
      *            of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
      *            {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
-     *            accepts a {@link BodyPart} object, in which case, the mime type is not used. A
-     *            {@link BodyPart} object can be passed for scripting languages other than Velocity.
+     *            accepts a {@link BodyPart} object, in which case, the mime type is not used.
      */
     public BodyPart addPart(String mimeType, Object content)
     {
@@ -118,8 +117,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
      *            of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
      *            {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
-     *            accepts a {@link BodyPart} object, in which case, the mime type and parameters are not used. A
-     *            {@link BodyPart} object can be passed for scripting languages other than Velocity.
+     *            accepts a {@link BodyPart} object, in which case, the mime type and parameters are not used. 
      * @param parameters the list of extra parameters. This is used for example to pass alternate content for the mail
      *            using the {@code alternate} key in the HTML Mime Body Part Factory. Mail headers can also be passed
      *            using the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -82,7 +82,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
 
     /**
      * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
-     * Directly supports BodyPart objects instead of the content as well.
+     *
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -88,7 +88,8 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @param mimeType the mime type of the content parameter
      * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
      *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also 
+     *        directly accepts a BodyPart object.
      */
     public BodyPart addPart(String mimeType, Object content)
     {
@@ -103,7 +104,8 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
      * @param mimeType the mime type of the content parameter
      * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
      *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also 
+     *        directly accepts a BodyPart object.
      * @param parameters the list of extra parameters. This is used for example to pass alternate content for the mail
      *        using the {@code alternate} key in the HTML Mime Body Part Factory. Mail headers can also be passed using
      *        the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys and values
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
@@ -79,7 +79,7 @@ public class ScriptMimeMessage extends ExtendedMimeMessage
         this.execution = execution;
         this.componentManager = componentManager;
     }
-
+    
     /**
      * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
      *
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -129,6 +129,28 @@ public BodyPart addPart(String mimeType, Object content, Map<String, Object> par
         return bodyPart;
     }
 
+    /**
+     * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
```
vmassol: i'd have added some explanation about why we need this method when it's not going to be possible to call it in Velocity (since you can't create a BodyPart in Velocity).

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -129,6 +129,28 @@ public BodyPart addPart(String mimeType, Object content, Map<String, Object> par
...
+     *
+     * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
+     *         part after it's been set (get/set some headers, etc)
+     * @param bodyPart the Mime Body Part object to be added
+     */
+    public BodyPart addPart(BodyPart bodyPart)
+    {
+        BodyPart temp = bodyPart;
```
vmassol: I prefer naming variables by what they represent. In this case for example: `returnedBodyPart` or `result`.

vmassol:
Sounds good globally even though I'd like to see an example in the javadoc of a Groovy/Java usage without this script service and with, to understand why this method is beneficial. This is a question more for @tmortagne I guess ;)

Thanks!

akshitdewan:
@vmassol does it look okay now?

vmassol:
Sure, it was almost ok already. The only thing missing is the use case and an example on how to use it in the javadoc since it's far from obvious why it's needed in the script service API (which are usually for velocity). It could also be mentioned in the javadoc if this cannot be used from velocity. That's why I'm asking for an example to compare how you'd use it in Java using the standard API vs using it in Groovy with the scripting API.  But this is more for @tmortagne I guess. Thx

akshitdewan:
@vmassol I did add that it cannot be used from velocity, but will wait for Thomas for more details(use cases, necessity etc.) since I'm not as informed about this.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -129,6 +129,28 @@ public BodyPart addPart(String mimeType, Object content, Map<String, Object> par
...
     }
 
+    /**
+     * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
+     * Although BodyParts cannot be created in Velocity, this method is useful when using Groovy
+     * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
+     *         part after it's been set (get/set some headers, etc)
+     * @param bodyPart the Mime Body Part object to be added
+     */
+    public BodyPart addPart(BodyPart bodyPart)
```
tmortagne: There is not reason to add a new method. You should check if the object is already a BodyPart in the other method instead.

akshitdewan: @tmortagne How do you check if it is a BodyPart using just the other parameters? Isn't a BodyPart parameter necessary?

tmortagne: The other method takes an Object and Object can be anything. You can test the type of the input using `instanceof` in Java.

akshitdewan: @tmortagne Oh sorry, I did not notice the other method 👍

akshitdewan:
@tmortagne is this better?

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -90,6 +90,17 @@
...
+        if (content instanceof BodyPart) {
+            try {
+                Multipart multipart = getMultipart();
+                multipart.addBodyPart((BodyPart) content);
+            } catch (Exception e) {
+                // Save the exception for reporting through the script services's getError() API
+                setError(e);
+                return null;
+            }
+            return (BodyPart) content;
```
vmassol: Could you please rewrite it so that there's only 1 single return please? Otherwise looks good (maybe missing some javadoc to explain what objects are supported). Thanks

vmassol: Actually you put the "if" in the wrong method. You should put in the most specific method (the one with the most parameters) since otherwise calling that other method will not handle BodyPart input...

akshitdewan: @vmassol should I just put an if and then call the less specific method to handle it?

vmassol: You need to move the "If" to the other `addPart` method...

vmassol: to this one: `public BodyPart addPart(String mimeType, Object content, Map<String, Object> parameters)`

akshitdewan: okay, will do

tmortagne: @akshitdewan as I indicated this is not about what do so when the content is already a BodyPart but what NOT to do when it's already a BodyPart. So the if here should be around the code that create a BodyPart when you don't already have one.

akshitdewan:
@tmortagne Is this what you were looking for?

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -82,7 +82,7 @@
 
     /**
      * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
-     *
+     * Directly supports BodyPart objects instead of the content as well.
```
vmassol: This should go in the doc for the "content" parameter.

vmassol: Otherwise, sounds good to me.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -86,7 +86,7 @@
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail
+     * @param content the content to include in the mail, also directly accepts BodyPart objects
```
vmassol: I've just improved the javadoc for the content parameter. Could you please update your javadoc too? thanks!

akshitdewan: where did you update it?

akshitdewan: ah I found it

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -86,7 +86,9 @@
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail, also directly accepts BodyPart objects
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
+     *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
+     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
```
vmassol: hmm that's weird that what I added appears as added by you... It's also missing your part about supporting BodyPart independently of the mimetype.

akshitdewan: Must be because I added it manually to my local fork. I’ll take care of it in a couple hours or so.. thanks!

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -88,7 +88,8 @@
      * @param mimeType the mime type of the content parameter
      * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
      *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also 
+     *        directly accepts a BodyPart object.
```
vmassol: --> "Also accepts a {@link BodyPart} object, in which case, the mime type is not used."

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -103,7 +104,8 @@ public BodyPart addPart(String mimeType, Object content)
...
```
vmassol: Same

akshitdewan:
@vmassol good now?

vmassol:
@tmortagne I really don't think this is the right approach. If you really want to allow supporting a BodyPart parameter then we need to add a new method: `public BodyPart addPart(BodyPart bodyPart)` since both mime type and parameters are not used.

But again I've been asking for use case examples since I don't really understand the need for this compared to the Java API. Especially since it won't be usable from velocity.

@akshitdewan: The javadoc is still missing two pieces of info: the parameters parameter is also not used when a body part is used and the fact that this is useful for scripting languages other than Velocity since it can't be used in Velocity. But first we need to agree with @tmortagne that this new API is really needed... Sorry about that...

akshitdewan:
@vmassol sure, I will add those things anyway and let thomas weigh in on if we need it. Thanks!

tmortagne:
> But again I've been asking for use case examples since I don't really understand the need for this compared to the Java API. Especially since it won't be usable from velocity.

All that is already in the Jira issue but again:

This is a not a Velocity tool but a scripting API, my use case was attaching a file to a mail (exported PDF) which was impossible with the provided API. Also you could have some BodyPart coming from another tool even in Velocity for any reason. It does not make any kind of sense to prevent the user to pass a body part if he already have one especially since the current API is very limited in the kind of parts it's supporting out of the box.

tmortagne:
> @tmortagne I really don't think this is the right approach. If you really want to allow supporting a BodyPart parameter then we need to add a new method: public BodyPart addPart(BodyPart bodyPart) since both mime type and parameters are not used.

Sure we can add a new method. But the other one should not crash if enough information is provided like it used to.

@akshitdewan add back the method you previous wrote and it should be all OK IMO.

tmortagne:
Actually another missing thing: @akshitdewan would be nice to add new tests in https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/script/ScriptMimeMessageTest.java to verify the code you added works as expected.

vmassol:
@tmortagne Actually I'm not sure it's a good idea to add a new `addPart(BodyPart)` since this would simply call the `addBodyPart` API. Maybe all that is needed is to be able to get the object on which to call addBodyPart, i.e. make `getMultipart()` public or something like that. This would allow to be able to call any of the javamail API. This is why I was asking for a code example, in order to see what would be required to make it simpler from scripting and to understand why it's not already possible using the java API (since the java api can be used from all scripting languages other than velocity).

vmassol:
> This is a not a Velocity tool but a scripting API, my use case was attaching a file to a mail (exported PDF) which was impossible with the provided API. Also you could have some BodyPart coming from another tool even in Velocity for any reason. It does not make any kind of sense to prevent the user to pass a body part if he already have one especially since the current API is very limited in the kind of parts it's supporting out of the box.

What would be nice would be to see the code for this in Java and in scripting.

akshitdewan:
@tmortagne @vmassol I'm not sure what should be done - should I add a new method?

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -79,16 +79,30 @@
         this.execution = execution;
         this.componentManager = componentManager;
     }
+    
+    /**
+     * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
+     * This method can be used for scripting languages other than Velocity
```
tmortagne: This line should be removed. Nothing prevent Velocity to use this method as long as some API provide any way to get a BodyPart object. It could even be used directly if a BodyPart Converter is implemented.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -79,16 +79,30 @@
...
+     * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
```
tmortagne: I don't think the return make much sense since this method always return the passed BodyPart.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -79,16 +79,30 @@
...
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
-     *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
+     *            of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
+     *            {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
+     *            accepts a {@link BodyPart} object, in which case, the mime type is not used. A
+     *            {@link BodyPart} object can be passed for scripting languages other than Velocity.
```
tmortagne: Again no need for this disclaimer about Velocity, you should remove it.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -101,28 +115,35 @@ public BodyPart addPart(String mimeType, Object content)
...
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
-     *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
+     *            of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
+     *            {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
+     *            accepts a {@link BodyPart} object, in which case, the mime type and parameters are not used. A
+     *            {@link BodyPart} object can be passed for scripting languages other than Velocity.
```
tmortagne: Again no need for this disclaimer about Velocity, you should remove it.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/script/ScriptMimeMessageTest.java
```diff
@@ -90,4 +91,14 @@ public void addHeader() throws Exception
...
+    
+    @Test
+    public void addPart() throws Exception
+    {
+        String header = "bodyPartTest";
+        MimeBodyPart bp = new MimeBodyPart();
+        bp.addHeader("BodyPart-Test", header);
+        this.scriptMessage.addPart(bp);
+        assertEquals(header, this.scriptMessage.getHeader("BodyPart-Test", null));
+    }
```
tmortagne: You should also add a test for the other method. I know right now this method use the other one but it might change and we need to make sure if someone refactor something the method keep supporting BodyPart.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
akshitdewan: @tmortagne Just to clarify, do you want to make the method void or modify the description of the return?

akshitdewan:
@tmortagne you said to remove the return for the least specific addPart method - did you mean make the method void or modify the description of the @return

tmortagne:
I mean `void`, there is no need for this method to return anything.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -79,16 +79,29 @@
...
     }
+    
+    /**
+     * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
+     *
+     * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
+     *         part after it's been set (get/set some headers, etc)
+     * @param bodyPart a {@link BodyPart} object to include in the mail
+     */
+    public BodyPart addPart(BodyPart bodyPart)
```
vmassol: missing a @since and an @Unstable annotations.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -79,16 +79,29 @@
...
      * @return the Mime Body Part object that was added. Returning it allows script to make modifications to that body
      *         part after it's been set (get/set some headers, etc)
      * @param mimeType the mime type of the content parameter
-     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime
-     *        type of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
-     *        {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc.
+     * @param content the content to include in the mail. The type depends on the mimetype. For example for a mime type
+     *            of {@code text/plain} or {@code text/html}, the content should be a String, for a mime type of
+     *            {@code xwiki/template} it should be a {@link org.xwiki.model.reference.DocumentReference}, etc. Also
+     *            accepts a {@link BodyPart} object, in which case, the mime type is not used.
```
vmassol: This shouldn't be part of your PR since this is something I committed!

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -101,28 +114,34 @@ public BodyPart addPart(String mimeType, Object content)
...
+     *            using the {@code headers} key with a {@code Map&lt;String, String&gt;} value containing header keys
+     *            and values.
      */
     public BodyPart addPart(String mimeType, Object content, Map<String, Object> parameters)
     {
         BodyPart bodyPart;
 
         try {
-            MimeBodyPartFactory factory = getBodyPartFactory(mimeType, content.getClass());
+            if (!(content instanceof BodyPart)) {
```
vmassol: I'm not sure I like this since it means accepting a null mimetype and a null parameters. This is not what this method was meant to be. So I'd remove all the changes you brought to this method.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/script/ScriptMimeMessageTest.java
```diff
@@ -90,4 +93,34 @@ public void addHeader() throws Exception
...
+    
+    @Test
+    public void addPartTwo() throws Exception
```
vmassol: Please rename the test method to something more meaningful, explaining what you're testing. For example: "addPartWhenNullMimeType". However, IMO we shouldn't support this, see above. I much prefer to throw an exception if the mimetype is null.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/script/ScriptMimeMessageTest.java
```diff
@@ -90,4 +93,34 @@ public void addHeader() throws Exception
...
+    {
+        String header = "bodyPartTest2";
+        MimeBodyPart bp = new MimeBodyPart();
+        bp.addHeader("BodyPart-Test2", header);
+        this.scriptMessage.addPart(null, bp);
+        assertEquals(header, this.scriptMessage.getHeader("BodyPart-Test2", null));
+    }
+
+    @Test
+    public void addPartThree() throws Exception
```
vmassol: Same here, we need some meaningful test method name. Thx!

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
akshitdewan: Hi @vmassol, I'm not sure what to do as I'm hearing different things from you and Thomas. He wants to add the functionality to this method to avoid running the extra code when the object is already a BodyPart and also to support calls from the other method while you think that it is unnecessary and it is not a good idea to accept null parameters.

vmassol: @akshitdewan You're perfectly right: @tmortagne and I need to agree first :) Sorry about that.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
akshitdewan: I just formatted it using ctrl shift f, that's why its showing as added by me

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
akshitdewan: @vmassol which version should go in the since tag?

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
vmassol: Indeed. Why did you do that? :) Best is to follow the same strategy as what was there before ;) Would be good if you could undo that. Thanks!

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
vmassol: `@since 10.0RC1`

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/script/ScriptMimeMessageTest.java
akshitdewan: Done!

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
akshitdewan: Sure!

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/ScriptMimeMessage.java
```diff
@@ -79,6 +80,18 @@
         this.execution = execution;
         this.componentManager = componentManager;
     }
+    
+    /**
+     * Add some content to the mail to be sent. Can be called several times to add different content type to the mail.
+     *
+     * @param bodyPart a {@link BodyPart} object to include in the mail
+     * @since 10.0-SNAPSHOT
```
tmortagne: It's 10.0RC1 actually.

tmortagne:
@akshitdewan please build the module before committing, you will see you have a Java error.

akshitdewan:
@tmortagne it is not showing any errors in the two files

tmortagne:
I definitely get an error with Maven and I confirm that your code is not right. Please make sure you are building the right module.

akshitdewan:
@tmortagne I built the mail send default with maven and got a build success. there are some warnings though - is that what you are referring to?

tmortagne:
No I'm not referring to a warning, again you have a very clear Java error which I would prefer that you find yourself by actually building the module instead of having to indicate it.

If you don't get an error you are definitely not building the right thing. You need to build from `xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default`.

akshitdewan:
@tmortagne https://kek.gg/i/5G25gg.png
This is what I get when I run mvn compile from the location you said - am I running the wrong mvn command?

tmortagne:
Are you sure what you have locally matches what's in this PR ? Make sure you don't have local modification or that you did not forget to push some commit.

akshitdewan:
@tmortagne should be all good now! I might've been on the wrong branch yesterday

tmortagne:
@akshitdewan your tests are failing when I build your branch

akshitdewan:
@tmortagne Hmm, do you know if they are failing because I wrote the test wrong or is it a fault of the methods?

akshitdewan:
Also, how do I see these failed tests on my dev environment? I still get a build success

akshitdewan:
I'm not sure what was going on with checking the headers in my previous test and if they are even supposed to be the same. I updated the tests to check if the BodyPart added to the multipart is the same as the one I created and those tests all passed. Let me know if the first iteration of the test that compared the headers should have been working because then there is something wrong with the method. Thanks!

tmortagne:

PR #666, opened 2017-12-23
removed use of close quietly

akshitdewan:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -1552,23 +1552,21 @@ public Date getResourceLastModificationDate(String name)
...
             }
         }
 
-        if (is == null) {
+        if (temp == null) {
             return FileUtils.readFileToByteArray(new File(name));
         }
 
-        try {
+        try (InputStream is = temp) {
```
tmortagne: The previous code was not great in the first place but I would move this whole block and merge it with the content of `if (getEngineContext() != null) {` which would be a lot cleaner.

akshitdewan:
@tmortagne is it okay to just return null if the contextengine is null?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -1552,24 +1552,16 @@ public Date getResourceLastModificationDate(String name)
...
         if (getEngineContext() != null) {
-
-            try {
-                is = getResourceAsStream(name);
+            try (InputStream is = getResourceAsStream(name)) {
+                if (is == null) {
+                    return FileUtils.readFileToByteArray(new File(name));
+                }
+                return IOUtils.toByteArray(is);
             } catch (Exception e) {
```
tmortagne: You should not catch all exceptions like this. This method is supposed to throw is there is an issue.

akshitdewan: @tmortagne What do you mean?

tmortagne: I mean remove this catch. The previous version was not catching anything.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -1552,24 +1552,16 @@ public Date getResourceLastModificationDate(String name)
...
-        if (is == null) {
-            return FileUtils.readFileToByteArray(new File(name));
-        }
-
-        try {
-            return IOUtils.toByteArray(is);
-        } finally {
-            IOUtils.closeQuietly(is);
-        }
+        return output;
```
akshitdewan: @tmortagne is it okay to return an empty byte array if the engine context is null?

akshitdewan: or maybe throw an exception?

tmortagne: Actually I mixed with the other pull request, there was indeed a catch before so you should put it back, sorry.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -1552,24 +1552,16 @@ public Date getResourceLastModificationDate(String name)
...
     public byte[] getResourceContentAsBytes(String name) throws IOException
     {
-        InputStream is = null;
+        byte output[] = new byte[0];
         if (getEngineContext() != null) {
-
-            try {
-                is = getResourceAsStream(name);
-            } catch (Exception e) {
+            try (InputStream is = getResourceAsStream(name)) {
```
tmortagne: Actually I mixed with the other pull request, there was indeed a catch before so you should put it back, sorry.

tmortagne: One difficulty is that readFileToByteArray itself is not supposed to be catched.

akshitdewan: Well, in order to use the try-with-resources block we will have to put it in the try block right? Maybe we could find the specific exception thrown by the input stream to avoid catching anything for readFileToByteArray?

tmortagne: readFileToByteArray does not nee the InputStream so no you don't need to put it in the try.

akshitdewan: @tmortagne okay, I removed it

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -1552,24 +1552,16 @@ public Date getResourceLastModificationDate(String name)
...
+                if (is != null) {
+                    return IOUtils.toByteArray(is);
+                }
+            } catch(Exception e){
             }
-        }
-
-        if (is == null) {
             return FileUtils.readFileToByteArray(new File(name));
```
tmortagne: Was not part of the `if (getEngineContext() != null) {` before.

tmortagne:

PR #669, opened 2017-12-30
Update the XML document format of 15 committed wiki pages

RobertVlp:


tmortagne:

PR #670, opened 2017-12-31
Update 10 xml file

Srijanjha:


tmortagne:
Could you execute mvn xar:format again, there was an issue in it (https://jira.xwiki.org/browse/XCOMMONS-1326).

Srijanjha:
Yes, I executed mvn xar: format

tmortagne:
I know but there was a bug in the XAR plugin when you did it and it's fixed now.

Srijanjha:
So should I redo this?

Srijanjha:
But it will yield the same result btw.

tmortagne:
> But it will yield the same result btw.

No, it should change the XML version in the declaration. Make sure you build with the latest version of the plugin (if you still don't see any change, try with -U to force SNAPSHOT updates).

tmortagne:
Please use the plugin, don't modify things by hand.

Srijanjha:
Sorry :)

tmortagne:
I meant what you did is not what you would get with the plugin so please use it or I can't validate it.

Srijanjha:
OK I'll try that again :)

PR #671, opened 2018-01-01
[codestyle] Remove this unused method parameter context.

fennecus-z:
Hi, I decided to submit this as my third issue fix rather. The issue is found here:

http://sonar.xwiki.org/issues/search#issues=3a465e46-2a37-401a-967e-190d045c7caf

Thanks.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -1042,9 +1042,8 @@ public class XWiki implements EventListener
 
     /**
      * @param content the content of the text area
-     * @param context see {@link XWikiContext}
      */
-    public static String getTextArea(String content, XWikiContext context)
+    public static String getTextArea(String content)
     {
         StringBuilder result = new StringBuilder();
 


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -1042,9 +1042,8 @@ public static String getServerWikiPage(String servername)
 
     /**
      * @param content the content of the text area
-     * @param context see {@link XWikiContext}
      */
-    public static String getTextArea(String content, XWikiContext context)
+    public static String getTextArea(String content)
```
vmassol: Actually you can't do that in this manner since you'd be breaking an API :)

You'll need to introduce new method, deprecate the old one (and move it to the legacy module). See http://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility

tmortagne:
This Sonar issue cannot really be fixed since this method is an API. I closed it.

vmassol:
@tmortagne It can be fixed, the way I mentioned it in comment above.

vmassol:
Now a better fix would probably be to move this static method elsewhere since it should not be in the XWiki class ideally. Remove the param would be a small improvement but not a big one.

PR #672, opened 2018-01-01
XWIKI-13640: Overflow issue in the UI of the Distribution Wizard

ApCoder123:
Resolves XWIKI-13640. Pushes text onto new line, removing need for any scrollbars. Tested on both extension manager and distribution wizard.


vmassol:
Thanks @ApCoder123 

Image before/after: https://codein.withgoogle.com/serve/6505846034923520/

@evalica I'm applying the commit based on the screenshot only. Please validate the fix and revert if not correct. Thanks!

evalica:
This is the correct fix. Thank you for your contribution.

PR #677, opened 2018-01-03
[codestyle] Replace the synchronized class Hashtable by an unsynchronized one such as HashMap.

fennecus-z:
Hey, sorry for the sloppy branch name, was feeling quite drained and slightly unwell when I was naming it, and accidentally created it with "hash" as part of the name instead of "hashable"

Here is the link to the issue I picked and my task, thanks.

http://sonar.xwiki.org/issues/search#issues=0ca6d3ee-41ab-4e81-b3e9-9abc94710409

https://codein.withgoogle.com/dashboard/task-instances/5241134932557824/

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWikiContext.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWikiContext.java
@@ -23,6 +23,7 @@ import java.lang.reflect.ParameterizedType;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWikiContext.java
```diff
@@ -68,7 +69,7 @@
  *
  * @version $Id$
  */
-public class XWikiContext extends Hashtable<Object, Object>
+public class XWikiContext extends HashMap<Object, Object>
```
tmortagne: XWikiContext being a very old public API and HashMap having different behavior than Hashtable regarding a few things this could break some use cases so I will close this Sonar issue as a false positive.

PR #680, opened 2018-01-05
update ApplicationResources.properties 2

1somya:


tmortagne:
This pull request includes changes that already been applied. Please make sure your fork is up to date (see https://help.github.com/articles/syncing-a-fork/).

1somya:
i updated my fork but still I don't find the changes I proposed in the updated file

tmortagne:
I'm not sure what you mean in your last comment but the diff definitely makes more sense than before :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2358,24 +2358,24 @@ xe.admin.passwordReset.step2.title=Reset your password
 xe.admin.passwordReset.step2.newPassword.label=New password
 xe.admin.passwordReset.step2.newPasswordVerification.label=Re-enter new passwod
 xe.admin.passwordReset.step2.submit=Save
-xe.admin.passwordReset.step2.success=The password has been successfully set.
+xe.admin.passwordReset.step2.success=The password has been successfully reset.
```
tmortagne: You changed the meaning of the message. This is displayed after you set a new password.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2358,24 +2358,24 @@ xe.admin.passwordReset.step2.title=Reset your password
...
 xe.admin.passwordReset.step2.login=Please login to continue \u00BB
 xe.admin.passwordReset.step2.backToStep1=Back to the password reset page \u00BB
 xe.admin.passwordReset.step2.error.emptyPassword=The password cannot be empty.
 xe.admin.passwordReset.step2.error.verificationMismatch=The two passwords do not match.
 xe.admin.passwordReset.step2.error.wrongParameters=Wrong parameters.
-xe.admin.passwordReset.step2.error.noProgrammingRights=This page requires programming rights to work, which currently isn't the case. Please notify an administrator of this problem and try again later.
+xe.admin.passwordReset.step2.error.noProgrammingRights=This page requires programming rights to work, which currently isn't the case. Please notify an administrator for this problem and try again later.
```
tmortagne: I don't understand why you changed it here, `of` sounded correct to me (but I might be wrong).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2358,24 +2358,24 @@ xe.admin.passwordReset.step2.title=Reset your password
...
 xe.admin.passwordReset.step2.versionComment.passwordReset=Password was reset
 xe.admin.passwordReset.step2.versionComment.changeValidationKey=Refreshed password reset token
 
 ### XWiki.Configurable - application configuration
 xe.admin.configurable.title=Custom configurable sections
 xe.admin.configurable.macros.title=Macros for custom configurable sections
 xe.admin.configurable.noPermissionThisApplication=You don't have permission to configure this application.
-xe.admin.configurable.applicationAuthorNoAdmin=This configuration cannot be displayed because it was last edited by [[{0}]] who doesn''t have permission to edit this page.
-xe.admin.configurable.cannotLockNoJavascript=This page cannot be locked for editing because Javascript is turned off, for page editing safety, please enable Javascript.
-xe.admin.configurable.configurationClassNonexistant=No class found by the name {0}, can''t display configuration.
```
tmortagne: Those double ' are not mistakes, the are part of the syntax (in the end you have only one).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
1somya: Actually the above would have been shown during creating new password so I thought it that way....

tmortagne: The step 1 is sending you a mail to change the password, then in the step 2 you set a new password.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
1somya: ok

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
1somya: I think that the user should be notifying the administrator for the concerned problem

1somya: in my opinion

aubincleme: As far as I know, the rule for business English is to use "notify of" (at least, I didn't found any references to "notify for" [here](https://dictionary.cambridge.org/dictionary/english/notify)) ; so we better keep the current translation :) .

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2240,7 +2240,7 @@ admin.colortheme.manage=Manage color themes\u00BB
 admin.customize=Customize
 admin.save=Save
 admin.defaultwikinotinstalled=Your wiki seems empty. You may want to import the default XWiki Enterprise wiki which contains a set of useful pages: user profiles, recent activity, administration pages and many more. This wiki is distributed as a XAR file, you can download it from {0}.
-admin.adminappnotinstalled=The administration application is not installed. Since XWiki Enterprise 1.5 the Administration is distributed as an application. You can download it from {0}.
+admin.adminappnotinstalled=The administration application is not installed. Since XWiki Enterprise 1.5, the Administration is distributed as an application. You can download it from {0}.
```
vmassol: Outside of scope for GCI:

Since we're now at XWiki 9+ I think we should change this message to: "The administration application is not installed. You should install it with the {0}Extension Manager{1}" and make "Extension Manager" point to the EM page. This means deprecating the previous key and creating a new one probably since the parameters would be different. WDYT?

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
vmassol: +1 to keep the current translation

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2373,7 +2373,7 @@ xe.admin.configurable.title=Custom configurable sections
 xe.admin.configurable.macros.title=Macros for custom configurable sections
 xe.admin.configurable.noPermissionThisApplication=You don't have permission to configure this application.
 xe.admin.configurable.applicationAuthorNoAdmin=This configuration cannot be displayed because it was last edited by [[{0}]] who doesn''t have permission to edit this page.
-xe.admin.configurable.cannotLockNoJavascript=This page cannot be locked for editing because Javascript is turned off, for page editing safety, please enable Javascript.
+xe.admin.configurable.cannotLockNoJavascript=This page cannot be locked for editing because Javascript is turned off.For page editing safety, please enable Javascript.
```
vmassol: You're missing a space before the "For". 

Also note that this text has already been changed by another student. FTR I had already commented that I didn't like to have long sentences but it was still committed apparently. Funnily the suggestion from this student is to put it back as it was previously ;)

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2519,7 +2519,7 @@ xe.invitation.doAction.decline.success=This invitation has successfully been dec
 
 xe.invitation.doUserActionOnMultipleMessages.notPossibleOnMultipleMessages=This action is not possible on multiple messages.
 xe.invitation.doUserActionOnMultipleMessages.confirmLabel=Confirm
-xe.invitation.doUserActionOnMultipleMessages.noMessagesFound=No messages were found by the ids given.
+xe.invitation.doUserActionOnMultipleMessages.noMessagesFound=No messages were found by the IDs given.
```
vmassol: I'd have said "No messages were found with the provided IDs."

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -4008,7 +4008,7 @@ xe.admin.passwordreset.nouser=The ~~{0}~~ user does not exist.
 xe.admin.passwordreset.ldapuser=The ~~{0}~~ user is an LDAP user. In that case the password has to be changed on the LDAP server.
 xe.admin.passwordreset.cannotreset=Cannot reset password: email address not provided in the user profile.
 xe.admin.passwordreset.emailsent=An e-mail was sent to <tt>{0}</tt>. Please follow the instructions in that e-mail to complete the password reset procedure.
-xe.admin.passwordreset.reseterror=An unknown problem occurred while sending the reset email.
+xe.admin.passwordreset.reseterror=An unknown problem occurred while sending the reset e-mail.
```
aubincleme: In my opinion, we should keep `email` (without hyphen) as we are using this word across all of our translations.
Moreover, it seems that `email` should be preferred against `e-mail` according to [this source](http://grammarist.com/style/e-mail-email/).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2240,7 +2240,7 @@ admin.colortheme.manage=Manage color themes\u00BB
 admin.customize=Customize
 admin.save=Save
 admin.defaultwikinotinstalled=Your wiki seems empty. You may want to import the default XWiki Enterprise wiki which contains a set of useful pages: user profiles, recent activity, administration pages and many more. This wiki is distributed as a XAR file, you can download it from {0}.
-admin.adminappnotinstalled=The administration application is not installed. Since XWiki Enterprise 1.5 the Administration is distributed as an application. You can download it from {0}.
+admin.adminappnotinstalled=The administration application is not installed. You should install it with the {0}Extension Manager{1}.
```
aubincleme: The comment made by @vmassol [here](https://github.com/xwiki/xwiki-platform/pull/680#discussion_r160031901) should not apply. We can't change the parameters of a translation key that easily.

tmortagne:
Thank for your contribution !

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2996,7 +2996,7 @@ xe.monitor.requests.size=Active requests size:
 xe.monitor.requests.page=Page:
 xe.monitor.thread=Thread:
 xe.monitor.requests.unfinished=Latest unfinished requests
-xe.monitor.requests.unfinished.description=These are requests that didn't reach "endRequest", but where cleaned-up by a reuse of threads. Only 32 requests max are kept in memory.
+xe.monitor.requests.unfinished.description=These are requests that didn't reach "endRequest", but where cleaned-up by a reuse of threads. Maximum 32 requests are kept in memory.
```
vmassol: I don't think this is correct. I think you need to write "A maximum of 32 requests are kept in memory.".

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -5156,7 +5156,7 @@ xe.blog.migration.inspace=in space
 xe.blog.migration.skipping=Skipping protected page
 xe.blog.migration.done=Done.
 xe.blog.migration.backtoblog=Back to the blog
-xe.blog.migration.pleaseconfirm=Please confirm that you want to migrate old articles to the new blog application:
+xe.blog.migration.pleaseconfirm=Please confirm if you want to migrate old articles to the new blog application:
```
vmassol: I'm not convinced that "if" is better than "that" here. I find it less good actually... Can someone confirm it's correct?

Enygma2002: I also prefer `that`, but neither ones are technically wrong. It just depends on the nuance you give it. Both practically end up in the same place, just that `that` asks for your intention, while `if` asks for your permission. Both ask you to confirm and I consider both acceptable.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
Enygma2002: I agree with you, but I also consider the currently proposed version an improvement from what already existed. It is a bit more concise, even if wrt. English correctness it's not the best (but then neither was the previous version).

vmassol: Yes changing an incorrect version with another incorrect one doesn't make us progress... ;)

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
vmassol: Then why accept the new one?

PR #685, opened 2018-01-11
XWiki-6022 : XWikiDocument toXML lists attachments in an inconsistent order

akshitdewan:
@tmortagne I worked on the general idea of what I think needs to be done but I wanted to check in before moving forward. I have not added the necessary javadoc comments and other formatting stuff - let me know if I am on the right track and I can then work on some of the finetuning aspects. Thanks!


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/DecoratedList.java
```diff
@@ -0,0 +1,84 @@
+package com.xpn.xwiki.doc;
+
+import java.util.ArrayList;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class DecoratedList extends ArrayList<XWikiAttachment>
```
tmortagne: I would call that XWikiAttachmentList to be more explicit and put it in an *.internal.* package to be sure it's not public API.

tmortagne:
The general logic looks OK. Make sure to keep all that internal.

akshitdewan:
I made the class internal, changed the name, and added Javadoc comments.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -492,9 +493,8 @@ public String toString()
      */
     private Map<DocumentReference, List<BaseObject>> xObjects = new TreeMap<DocumentReference, List<BaseObject>>();
 
-    // TODO: use a Map instead of a List to store attachment in XWikiDocument
     private final List<XWikiAttachment> attachmentList =
-        new AbstractNotifyOnUpdateList<XWikiAttachment>(new ArrayList<XWikiAttachment>())
+        new AbstractNotifyOnUpdateList<XWikiAttachment>(new XWikiAttachmentList())
```
tmortagne: Since you introduced a custom class it would be cleaner to also include the features of AbstractNotifyOnUpdateList in it and replace all this code.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -492,9 +493,7 @@ public String toString()
      */
     private Map<DocumentReference, List<BaseObject>> xObjects = new TreeMap<DocumentReference, List<BaseObject>>();
 
-    // TODO: use a Map instead of a List to store attachment in XWikiDocument
-    private final List<XWikiAttachment> attachmentList =
-        new AbstractNotifyOnUpdateList<XWikiAttachment>(new ArrayList<XWikiAttachment>())
+    private final List<XWikiAttachment> attachmentList = new XWikiAttachmentList()
```
tmortagne: You should make attachmentList be a XWikiAttachmentList, that would remove all the casts you are currently doing.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,216 @@
...
+
+import com.xpn.xwiki.doc.XWikiAttachment;
+
+/**
+ * AttachmentList that holds elements in order of filename.
+ * 
+ * @version $Id$
+ * @since 10.0RC1
+ */
+public abstract class XWikiAttachmentList extends ArrayList<XWikiAttachment>
```
tmortagne: Shouldn't you extend AbstractNotifyOnUpdateList instead of reimplementing it ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -492,9 +493,7 @@ public String toString()
...
         {
             @Override
             public void onUpdate()
```
tmortagne: I would have also moved those methods to XWikiAttachmentList (setMetaDataDirty is public so you can call it from another class). You could pass the XWikiDocument instance to the XWikiAttachmentList constructor for example.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4846,17 +4845,14 @@ public XWikiAttachment removeAttachment(XWikiAttachment attachment)
...
-                return attachment;
-            }
+        if (!list.contains(attachmentToRemove)) {
+            attachmentToRemove = null;
+        } else {
+            ((XWikiAttachmentList) (list)).remove(attachmentToRemove);
         }
-
-        return null;
+        this.attachmentsToRemove.add(new XWikiAttachmentToRemove(attachmentToRemove, toRecycleBin));
```
tmortagne: All this should be in the else otherwise you will end up adding a null XWikiAttachment to the list of attachments to remove.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,216 @@
...
+     * @param attachment the attachment to add to the list
+     * @return the attachment that was added to the list in order of filename
+     * @since 10.0RC1
+     */
+    public XWikiAttachment set(XWikiAttachment attachment)
+    {
+        map.put(attachment.getFilename(), attachment);
+        super.clear();
+        super.addAll(map.values());
+        return attachment;
```
tmortagne: This method should have the same behavior than XWikiDocument#setAttachment instead of make setAttachment workaround it. Also there is no point in returning what you get as parameter.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,216 @@
...
+    private Map<String, XWikiAttachment> map;
+
+    /**
+     * Initializes the map.
+     * 
+     * @since 10.0RC1
+     */
+    public XWikiAttachmentList()
+    {
+        map = new TreeMap<String, XWikiAttachment>();
```
tmortagne: I did a quick check and TreeMap is not thread safe so you will have issues with it (XWikiDocument can be manipulated from several threads at the same time). ConcurrentSkipListMap seems be here for that.

By the way you don't really need this XWikiAttachmentList constructor, you could directly initialize `map` field.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: Wouldn't it be more efficient to just create an AbstractNotifyOnUpdateList with a XWikiAttachmentList parameter like it was originally? I thought you were thinking of eliminating AbstractNotifyOnUpdateList - that's why I did it this way

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: now since I added the document parameter, seems like better practice to initialize it in the constructor

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
tmortagne: My point was to hide all this plumbing from XWikiDocument so that code in XWikiDocument is much simpler. It only deal with a XWikiAttachmentList which hide all the complexity.

akshitdewan: @tmortagne so are you saying I should extend it and then for method implementations call super and then modify the map as necessary?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: okay sure, I can move it and make it final

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: Now that I think about it, we can only extend one class and we are already extending ArrayList

tmortagne: You don't really have to extend ArrayList (you just need it to implement List interface) and you don't need XWikiAttachmentList to take a parameter. Creating and passing an Arraylist to AbstractNotifyOnUpdateList can be a XWikiAttachmentList decision.

akshitdewan: I extended AbstractNotifyOnUpdateList on this PR https://github.com/xwiki/xwiki-platform/pull/695, that should be better since it implements list already

akshitdewan:
Closing - all work is on https://github.com/xwiki/xwiki-platform/pull/695

PR #695, opened 2018-01-12
XWiki-6022 : XWikiDocument toXML lists attachments in an inconsistent order

akshitdewan:
I made a separate PR so I didn't lose any changes made on the other branch in case this is the wrong approach. Feel free to close the other PR if this is more along the lines of what should be done! I did have to change the visibility of the list in AbstractNotifyOnUpdateList to protected so I could access it and I'm not sure if that is okay to do. I had to do that because I can't call the addAll and clear methods without the notifier class calling onUpdate() and added() methods which should not be called in this scenario. @tmortagne I also fixed all the comments you left on the other PR here.


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4709,7 +4695,7 @@ public void setAttachmentList(List<XWikiAttachment> list)
         }
     }
 
-    public List<XWikiAttachment> getAttachmentList()
+    public XWikiAttachmentList getAttachmentList()
```
tmortagne: This actually breaks an API. XWikiAttachmentList must remain internal, what is important for outside world is that it impléments List<XWikiAttachment>. By the way I'm pretty sure this change produce a build error.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4846,17 +4832,14 @@ public XWikiAttachment removeAttachment(XWikiAttachment attachment)
...
         List<XWikiAttachment> list = getAttachmentList();
-        for (int i = 0; i < list.size(); i++) {
-            XWikiAttachment attachment = list.get(i);
-            if (attachmentToRemove.getFilename().equals(attachment.getFilename())) {
-                list.remove(i);
-                this.attachmentsToRemove.add(new XWikiAttachmentToRemove(attachment, toRecycleBin));
-                setMetaDataDirty(true);
-                return attachment;
-            }
+        if (!list.contains(attachmentToRemove)) {
```
tmortagne: You don't need this if. You could simply rely on list.remove return (which is true when it found something to remove).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -5442,22 +5424,10 @@ public XWikiAttachment setAttachment(XWikiAttachment attachment)
...
-                it.add(attachment);
-
-                return currentAttachment;
-            }
-        }
-
-        getAttachmentList().add(attachment);
-
-        return null;
+        XWikiAttachmentList list = getAttachmentList();
```
tmortagne: No need for this step, just use `this.attachmentList` directly.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4846,17 +4832,14 @@ public XWikiAttachment removeAttachment(XWikiAttachment attachment)
     public XWikiAttachment removeAttachment(XWikiAttachment attachmentToRemove, boolean toRecycleBin)
     {
         List<XWikiAttachment> list = getAttachmentList();
```
tmortagne: No need for this step, just use `this.attachmentList` directly.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,238 @@
...
+ * @version $Id$
+ * @param <E>
+ * @since 10.0RC1
+ */
+public class XWikiAttachmentList extends AbstractNotifyOnUpdateList<XWikiAttachment>
+{
+
+    private final Map<String, XWikiAttachment> map = new ConcurrentSkipListMap<String, XWikiAttachment>();
+
+    private XWikiDocument document;
```
tmortagne: Minor improvement: make it `final` to indicate to the compiler it's never supposed to change.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * {@inheritDoc}
+     * 
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean addAll(Collection<? extends XWikiAttachment> c)
+    {
+        onUpdate();
```
tmortagne: The list is not updated yet, should this be moved after the change ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,238 @@
...
+import com.xpn.xwiki.internal.AbstractNotifyOnUpdateList;
+
+/**
+ * AttachmentList that holds elements in order of filename.
+ * 
+ * @version $Id$
+ * @param <E>
+ * @since 10.0RC1
+ */
+public class XWikiAttachmentList extends AbstractNotifyOnUpdateList<XWikiAttachment>
```
tmortagne: Now that I walk your code more, because of the clear()/addAll() logic you actually seems to reimplement the whole  AbstractNotifyOnUpdateList write logic anyway, right ? So looks like there is no point in extending it after all, sorry for the mislead.

Another note: instead of clear/addAll, creating a new ArrayList would improve Thread safety.

So here is a proposal: extend org.apache.commons.collections4.list.AbstractListDecorator from which you override all the write methods to modify the Map instead and then re-create and change the decorated List.

Here is what I would have done from when starting from scratch: make XWikiDocument#attachmentList a simple ConcurrentSkipListMap and create a new XWikiAttachmentList-like each time getAttachmentList() is called (i.e. consider it like some retro compatibility instead of being the core the XWikiDocument attachment handling). But your version is OK also so it's just a "for your information" note :)

akshitdewan: @tmortagne why not just extend abstractnotifyonupdatelist? We get the same additional list methods and also have the onUpdate() and added(). Is there any added benefit to extending the listdecorator

akshitdewan: @tmortagne also just wondering, would you have returned an independent list from getAttachmentList() that just mirrors the map?  how would that handle new things being added to the list?

akshitdewan: I fixed all the minor changes that you asked, I'm not sure why the decorator is needed as we do reuse a few of the notifier methods and they automatically call update and added. I left it extending the notifier for now. The only limitation I see is that the list in the notifier can't be final since we want to set it to a new list instead of clear/addAll. I will be sleeping now, will check in in the morning! Let me know your thoughts!

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4845,18 +4831,14 @@ public XWikiAttachment removeAttachment(XWikiAttachment attachment)
...
-        List<XWikiAttachment> list = getAttachmentList();
-        for (int i = 0; i < list.size(); i++) {
-            XWikiAttachment attachment = list.get(i);
-            if (attachmentToRemove.getFilename().equals(attachment.getFilename())) {
-                list.remove(i);
-                this.attachmentsToRemove.add(new XWikiAttachmentToRemove(attachment, toRecycleBin));
-                setMetaDataDirty(true);
-                return attachment;
-            }
+        List<XWikiAttachment> list = this.attachmentList;
```
tmortagne: Why this temporary `list` variable ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4845,18 +4831,14 @@ public XWikiAttachment removeAttachment(XWikiAttachment attachment)
...
+        if (list.remove(attachmentToRemove)) {
+            this.attachmentsToRemove.add(new XWikiAttachmentToRemove(attachmentToRemove, toRecycleBin));
+            setMetaDataDirty(true);
```
tmortagne: Shouldn't this be done by list.remove() already ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -5442,22 +5423,10 @@ public XWikiAttachment setAttachment(XWikiAttachment attachment)
...
-                it.add(attachment);
-
-                return currentAttachment;
-            }
-        }
-
-        getAttachmentList().add(attachment);
-
-        return null;
+        XWikiAttachmentList list = this.attachmentList;
```
tmortagne: Same comment about useless temporary variable.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -5442,22 +5423,10 @@ public XWikiAttachment setAttachment(XWikiAttachment attachment)
...
+        XWikiAttachment currentAttachment = list.getByFilename(attachment.getFilename());
```
tmortagne: set() already take care of this so you could simply return the result of set().

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
tmortagne: > @tmortagne also just wondering, would you have returned an independent list from getAttachmentList() that just mirrors the map? how would that handle new things being added to the list?

As I said it would be something similar to XWikiAttachmentList but only for getAttachmentList() so it would handle new things added to the Map. What getAttachmentList() is a very bad practice to I would have deprecated it and considered all this as a retro compatibility thing and introduce a new method which return an unmodifiable list. Basically the idea would be to not "pollute" the whole XWikiDocument with this.

> @tmortagne why not just extend abstractnotifyonupdatelist? We get the same additional list methods and also have the onUpdate() and added(). Is there any added benefit to extending the listdecorator

I much preference extending a more or less standard org.apache.commons.collections4.list.AbstractListDecorator when all I need is a decorator. You don't really need onUpdate() here since you override all the write methods (you have to since you can't rely on abstractnotifyonupdatelist implementation because you need to recreate the list from the Map).

tmortagne:
You did a great work so far @akshitdewan ! Not much left.

I'm picky because XWikiDocument is the heart of XWiki ;)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: @tmortagne when you said "you don't need onUpdate()", did you mean that I don't need it at all or its a functionality of the notifier that can be easily rewritten?

akshitdewan:
@tmortagne Thanks! This has been a very interesting task compared to bug fix ones!

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: Also, listdecorator won't actually allow me to access the list right? So I will have to use clear/addAll?

tmortagne: > Also, listdecorator won't actually allow me to access the list right? So I will have to use clear/addAll?

You can access it but even so if you recreate it from the map values every time all you need is to call #setCollection with the new ArrayList.

akshitdewan: It also will require a lot of casting, since the methods don't override if the parameter is an XWikiAttachment while it has to be an Object

tmortagne: > @tmortagne when you said "you don't need onUpdate()", did you mean that I don't need it at all or its a functionality of the notifier that can be easily rewritten?

The only point of onUpdate() as defined by AbstractNotifyOnUpdateList is to be overwritten by an extending class to know when the list is modified but your are not writing a class to extend here. You don't really care about this generic feature, all you care about is updating the XWikiDocument medatada dirty flag. Also you are not "using" AbstractNotifyOnUpdateList onUpdate() feature since you are the one calling that method (since you reimplement all the writing methods). In short my point is that you don't really reuse from AbstractNotifyOnUpdateList more that you would get from AbstractListDecorator and I usually prefer to reuse Apache commons stuff than maintaining custom XWiki stuff.

tmortagne: > It also will require a lot of casting, since the methods don't override if the parameter is an XWikiAttachment while it has to be an Object

I don't understand what you mean here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -5441,23 +5421,7 @@ public void addAttachment(XWikiAttachment attachment)
     public XWikiAttachment setAttachment(XWikiAttachment attachment)
     {
         attachment.setDoc(this);
```
tmortagne: Should probably move this inside XWikiAttachmentList#set() to be sure all added attachments have the right document.

tmortagne: Actually you do have this feature in XWikiAttachmentList arelady but you do it after adding the attachment instead of before :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+    /**
+     * Adds attachment to the list in order of filename.
+     * 
+     * @param attachment XWikiAttachment to add to the list
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean add(XWikiAttachment attachment)
+    {
+        map.put(attachment.getFilename(), attachment);
```
tmortagne: Most of the method is a duplicate of #set() so you should reuse it instead.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * {@inheritDoc}
+     * 
+     * @since 10.0RC1
+     */
+    @Override
+    public void clear()
+    {
+        super.clear();
```
tmortagne: Same as the other write method, it's safer to create a new ArrayList that modify it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+     * Adds attachment to the list in order of filename.
+     * 
+     * @param attachment XWikiAttachment to add to the list
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean add(XWikiAttachment attachment)
+    {
+        map.put(attachment.getFilename(), attachment);
+        this.list = new ArrayList<XWikiAttachment>(map.values());
```
tmortagne: Since you need to both:
* recreate the List
* update document metadata dirty flag

for each write operation you should probably have both in the same method called by all write operations.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+            map.put(x.getFilename(), x);
+            added(x);
+        }
+        this.list = new ArrayList<XWikiAttachment>(map.values());
+        onUpdate();
+        return true;
+    }
+
+    /**
+     * {@inheritDoc}
```
tmortagne: You should probably copy/paste the explanation you have in add(int index, XWikiAttachment attachment) to be clear.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+    @Override
+    public boolean addAll(int index, Collection<? extends XWikiAttachment> c)
+    {
+        return addAll(c);
+    }
+
+    /**
+     * {@inheritDoc}
+     * 
+     * @since 10.0RC1
```
tmortagne: You don't really need this comment IMO (unless you add more stuff to it).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+     * 
+     * @param attachment the attachment to add to the list
+     * @return the attachment that was added to the list in order of filename
+     * @since 10.0RC1
+     */
+    public XWikiAttachment set(XWikiAttachment attachment)
+    {
+        XWikiAttachment ret = map.put(attachment.getFilename(), attachment);
+        this.list = new ArrayList<XWikiAttachment>(map.values());
+        added(ret);
```
tmortagne: You are modifying the wrong instance, the one you added is `attachment`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+    public XWikiAttachment getByFilename(String filename)
+    {
+        return map.get(filename);
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c)
+    {
+        for (XWikiAttachment x : (Collection<? extends XWikiAttachment>) c) {
+            if (this.list.contains(x))
```
tmortagne: Missing {} to follow XWiki codestyle.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+     * Removes XWikiAttachment.
+     * 
+     * @param attachment XWikiAttachment to remove.
+     * @return XWikiAttachment that was removed or null if not found
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean remove(Object attachment)
+    {
+        int index = list.indexOf(attachment);
```
tmortagne: It would be faster to use the Map for this: call Map#put() and check if the return is the same than the passed XWikiAttachment and if they are different update the list, notify the XWikiDocument, return true, etc.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c)
+    {
+        for (XWikiAttachment x : (Collection<? extends XWikiAttachment>) c) {
+            if (this.list.contains(x))
+                remove(x);
+        }
+        return true;
```
tmortagne: It's supposed to be true only if something actually changed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c)
+    {
+        for (XWikiAttachment x : this.list) {
+            if (!((Collection<? extends XWikiAttachment>) (c)).contains(x))
+                remove(x);
+        }
+        return true;
```
tmortagne: It's supposed to be true only if something actually changed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: I just left it there for the since tag - is thats necessary?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: @tmortagne The casting thing was a misunderstanding on my part, I apologize. You said that I can access the list, but I don't know how we could do that as it would be a private field, right? In order to reassign it to a new list, we can't just get it from the decorate() (gets the list) method.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
tmortagne: Not really, XWikiAttachmentList already have this since. You put a `@since` in the method when it's more recent than the class itself. So yes you can remove all the method `@since` you have.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
tmortagne: > In order to reassign it to a new list, we can't just get it from the decorate() (gets the list) method.

You missed the following: "all you need is to call #setCollection with the new ArrayList" :)

akshitdewan: I did see that, but I don’t see how I can create a “new ArrayList”

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: Okay!

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
tmortagne: I don't understand, you are doing it in updateMap().

akshitdewan: @tmortagne Oh I see! I thought you wanted me to create a new list and then call setCollection on it but I don't think I can access the list field.

tmortagne: In practice I don't think you need to access the list field for any other reason than reseting it.

akshitdewan: The setCollection method is a protected method part of AbstractCollectionDecorator which is extended by AbstractListDecorator and java does not allow super.super, how should I go about calling it?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+     */
+    @Override
+    public boolean addAll(Collection<? extends XWikiAttachment> c)
+    {
+        for (XWikiAttachment x : c) {
+            map.put(x.getFilename(), x);
+            added(x);
+        }
+        updateMap();
+        return true;
```
tmortagne: It's supposed to be true only of something changed, same for updateMap().

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+     * {@inheritDoc}
+     * 
+     * @since 10.0RC1
+     */
+    @Override
+    public XWikiAttachment remove(int index)
+    {
+        XWikiAttachment removedAttachment = map.remove(this.list.get(index).getFilename());
+        if (removedAttachment != null) {
+            removedAttachment = super.remove(index);
```
tmortagne: You should only manipulate the map and then reset the list if the map changed, not modify the list.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+     * @return true unless the attachment is not found
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean remove(Object attachment)
+    {
+        XWikiAttachment xwikiAttachment = (XWikiAttachment) attachment;
+        if(map.put(xwikiAttachment.getFilename(), xwikiAttachment) == null)
+            return false;
+        map.remove(attachment);
```
tmortagne: This won't work since map key is a String, not an XWikiAttachment.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+     * 
+     * @param attachment XWikiAttachment to remove.
+     * @return true unless the attachment is not found
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean remove(Object attachment)
+    {
+        XWikiAttachment xwikiAttachment = (XWikiAttachment) attachment;
+        if(map.put(xwikiAttachment.getFilename(), xwikiAttachment) == null)
```
tmortagne: I don't understand why you are calling put in a method supposed to remove something.

akshitdewan: @tmortagne you left a comment saying to call put in this method to check if the attachment exists in the list

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+     * Adds or replaces attachment with the same filename as the parameter.
+     * 
+     * @param attachment the attachment to add to the list
+     * @return the attachment that was added to the list in order of filename
+     * @since 10.0RC1
+     */
+    public XWikiAttachment set(XWikiAttachment attachment)
+    {
+        map.put(attachment.getFilename(), attachment);
+        added(attachment);
```
tmortagne: You should probably do all that only if what map.put return is different from what you passed (i.e. something actually changed).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+    {
+        return map.get(filename);
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c)
+    {
+        boolean changed = false;
+        for (XWikiAttachment x : (Collection<? extends XWikiAttachment>) c) {
+            if (this.list.contains(x)) {
```
tmortagne: You don't need this like since `remove()` gives you the same information already.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+        return changed;
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c)
+    {
+        boolean changed = false;
+        for (XWikiAttachment x : this.list) {
+            if (!((Collection<? extends XWikiAttachment>) (c)).contains(x)) {
```
tmortagne: You don't need this `if` since `remove()` gives you the same information already.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+                if(remove(x)) {
+                    changed = true;
+                }
+            }
+        }
+        return changed;
+    }
+
+    /** Called when the list is updated. The method will be called at least once, but may be called several times */
+    public void onUpdate()
```
tmortagne: You don't seems to be using this method anymore.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,250 @@
...
+    {
+        document.setMetaDataDirty(true);
+    }
+    
+    public void updateMap()
```
tmortagne: Should be private since it's only used internally. There is no reason to call it from outside.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: I don't know why this is still showing up, it disappeared here : https://github.com/akshitdewan/xwiki-platform/blob/list2/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.ListIterator;
+import java.util.List;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ConcurrentSkipListMap;
+
+import com.xpn.xwiki.doc.XWikiAttachment;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+import com.xpn.xwiki.internal.AbstractNotifyOnUpdateList;
```
tmortagne: Unused import.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Adds attachment to the list in order of filename.
+     * 
+     * @param attachment XWikiAttachment to add to the list
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean add(XWikiAttachment attachment)
+    {
+        return set(attachment) != null;
```
tmortagne: Comparing with null is not enough. You might replace an existing attachment with a different instance in which case you should return true.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param index index is ignored as list is reordered based on filename
+     * @param c Collection that contains XWikiAttachment objects
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean addAll(Collection<? extends XWikiAttachment> c)
+    {
+        boolean changed = false;
+        for (XWikiAttachment x : c) {
+            if (map.put(x.getFilename(), x) == null) {
```
tmortagne: You should compare with x, not null since there could already be a different attachment in the map.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,240 @@
...
+     * 
+     * @param attachment the attachment to add to the list
+     * @return the attachment that was previously matched to the same filename or null if no attachment was matched to
+     *         it
+     * @since 10.0RC1
+     */
+    public XWikiAttachment set(XWikiAttachment attachment)
+    {
+        XWikiAttachment put = map.put(attachment.getFilename(), attachment);
+        if (put == null || !put.equals(attachment)) {
```
tmortagne: I would use == instead of equals. XWikiAttachment don't implement equals and even if they were is might be super slow (they would have to compare sometime big files contents).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+     * Adds attachment to the list in order of filename.
+     * 
+     * @param attachment XWikiAttachment to add to the list
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean add(XWikiAttachment attachment)
+    {
+        XWikiAttachment set = set(attachment);
+        return !set.equals(attachment);
```
akshitdewan: is this return correct?

tmortagne: ==, not equals.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+     * @param c Collection that contains XWikiAttachment objects
+     * @since 10.0RC1
+     */
+    @Override
+    public boolean addAll(Collection<? extends XWikiAttachment> c)
+    {
+        boolean changed = false;
+        for (XWikiAttachment x : c) {
+            XWikiAttachment put = map.put(x.getFilename(), x);
+            if (!x.equals(put)) {
```
akshitdewan: @tmortagne Is this correct?

tmortagne: Don't use equals for attachment instances.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: okay

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+        return map.get(filename);
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c)
+    {
+        boolean changed = false;
+        XWikiAttachmentList list = (XWikiAttachmentList) this.decorated();
+        for (XWikiAttachment x : (Collection<? extends XWikiAttachment>) c) {
+            if (list.contains(x)) {
```
tmortagne: Useless if, remove does the job already.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+                if (remove(x)) {
+                    changed = true;
+                }
+            }
+        }
+        return changed;
```
tmortagne: Shouldn't you call updatedMap(); if true ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c)
+    {
+        boolean changed = false;
+        XWikiAttachmentList list = (XWikiAttachmentList) this.decorated();
+        for (XWikiAttachment x : (Collection<? extends XWikiAttachment>) c) {
+            if (list.contains(x)) {
+                if (remove(x)) {
```
tmortagne: Would be better for performances to manipulate directly the map and then at the end recreate the list instead of recreating for each element.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+        return changed;
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c)
+    {
+        boolean changed = false;
+        XWikiAttachmentList list = (XWikiAttachmentList) this.decorated();
+        for (XWikiAttachment x : list) {
+            if (remove(x)) {
```
tmortagne: Actually you are supposed to remove the elements which are NOT in the input list.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,243 @@
...
+            }
+        }
+        return changed;
+    }
+
+    /** Sets MetaDataDirty to true and resets the list with the values in the map */
+    private void updatedMap()
+    {
+        document.setMetaDataDirty(true);
+        this.setCollection(map.values());
```
tmortagne: This is not right. You need to pass a new new ArrayList to setCollection() since values() is not a List and any call to decorated() will crash.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,247 @@
...
+        return changed;
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c)
+    {
+        boolean changed = false;
+        Collection<XWikiAttachment> values = map.values();
+        for(XWikiAttachment x : values) {
+            if(!c.contains(x)) {
```
akshitdewan: @tmortagne is this correct now?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,247 @@
...
+        return changed;
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> c)
+    {
+        boolean changed = false;
+        Collection<XWikiAttachment> values = map.values();
+        for (XWikiAttachment x : values) {
+            if (!c.contains(x)) {
```
akshitdewan: @tmortagne is this correct now?

tmortagne: It will remove everything not in c so yes sounds about right :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,247 @@
...
+                map.remove(x.getFilename(), x);
```
tmortagne: Too quick, I doubt this builds :)

akshitdewan: @tmortagne looks fine on eclipse, do you mean a maven build?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
```diff
@@ -0,0 +1,247 @@
...
+    {
+        return map.get(filename);
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> c)
+    {
+        boolean changed = false;
+        for (XWikiAttachment x : (Collection<? extends XWikiAttachment>) c) {
+            if (map.remove(x.getFilename(), x)) {
```
tmortagne: Too quick :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
akshitdewan: @tmortagne do I have the logic correct : 
while iterator has next value
if next is equal to x
remove using iterator and break

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/AbstractNotifyOnUpdateList.java
```diff
@@ -36,7 +36,7 @@
 {
 
     /** The list to wrap. */
-    private final List<E> list;
+    protected List<E> list;
```
tmortagne: No need for this change anymore.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/XWikiAttachmentList.java
tmortagne: Sorry I just forgot about the Map#get with two parameters...

Still I don't see why you don't use the other remove actually.

tmortagne: > @tmortagne do I have the logic correct :
> while iterator has next value
> if next is equal to x
> remove using iterator and break

The logic is the same as what you have right now (don't see any break in tour current code). Simply instead of a foreach you should use an Iterator and call Iterator#remove() instead of Map#remove.

akshitdewan: @tmortagne you're right, it is unnecessary - I was making sure that the filename corresponds to the correct attachment but it will because we are iterating through the map.values()

akshitdewan:
@tmortagne could you look at the logic for the removeAll and retainAll now?

tmortagne:
> @tmortagne could you look at the logic for the removeAll and retainAll now?

You don't plan to use Iterator after all ?

akshitdewan:
@tmortagne I can use iterator - sorry, I misread what you said, I thought you were saying it wasn't necessary after all.

tmortagne:
Your choice. It will be a bit better for performances but it's not critical.

akshitdewan:
@tmortagne I may as well, let me work on that and then write up some tests

akshitdewan:
@tmortagne I'm not sure if the tests are running - I put one to compare "hi" and "hello" but it did not show up as a failure. It throws a nullpointer on the call to getAttachmentList(), any idea why? Also, the ones that I have written so far, are they okay or should they be done differently?

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentMockitoTest.java
```diff
@@ -1397,4 +1399,37 @@ public void tofromXMLDocument() throws XWikiException
 
         assertEquals(this.document, newDocument);
     }
+    
+    @Test
+    public void listAddAttachment() throws XWikiException{
+        XWikiAttachment attachment = new XWikiAttachment(this.document, "testAttachment");
+        attachmentList.add(attachment);
+        assertEquals(this.document.getAttachmentList().contains(attachment),true);
```
tmortagne: I would also check  what this.document.getAttachment("testAttachment") returns.

tmortagne: Should be assertTrue instead of assertEquals.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentMockitoTest.java
```diff
@@ -1397,4 +1399,37 @@ public void tofromXMLDocument() throws XWikiException
...
+        List<XWikiAttachment> list = new ArrayList<XWikiAttachment>();
+        list.add(attachment1);
+        list.add(attachment2);
+        list.add(attachment3);
+        assertEquals(this.document.getAttachmentList(),list);
+    }
+    
+    @Test
+    public void testTest() throws XWikiException{
+        assertEquals("hi","hello");
```
tmortagne: You might want to remove that test :)

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentMockitoTest.java
```diff
@@ -1397,4 +1399,37 @@ public void tofromXMLDocument() throws XWikiException
...
+    }
+    
+    @Test
+    public void listClear() throws XWikiException{
+        attachmentList.clear();
+        assertEquals(attachmentList,new ArrayList<XWikiAttachment>());
```
tmortagne: I would assertTrue attachmentList.isEmpty() instead since that's what you want to know (not that it's equals to some list).

tmortagne:
> @tmortagne I'm not sure if the tests are running

It should be very clear which one is executed if you run them in Eclipse: right click on the test class -> Run/Debug as -> Unit test.

@Test annotation and being public are the only criteria for a test method to be executed AFAIK.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentMockitoTest.java
akshitdewan: yup, I'm just experimenting with testing a bit - this test doesn't seem to fail so I'm confused

akshitdewan:
@tmortagne Those seem to be the main ones - which other ones should I include? Also, I still have not used the iterator so please do not merge until I do that. Thanks!

akshitdewan:
Should I just write tests for all of the methods?

tmortagne:
Applied. Thanks for this great contribution !

PR #697, opened 2018-01-13
Update 10 XML Files

ApCoder123:
I have run `mvn xar:format` on 10 files


acotiuga:
You can also make all these change in one commit, instead of doing 10 commits :)

PR #699, opened 2018-01-13
Updated XML format for some files

siddhpant:
Updated XML format for some files.

I have done the **work of ALL TASKS** in this one commit so as to save time.

Edited some extra files too because why not?


acotiuga:
Thanks for your contribution! I appreciate that you used git interactive staging and you skipped unnecessary changes.

PR #701, opened 2018-01-14
XWIKI-14436 : Livetable shows 'emptyvalue' for objects with no content

Srijanjha:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2814,7 +2814,7 @@ platform.index.doc.author=Last Author
 platform.index.doc.title=Title
 platform.index.doc.fullName=Page
 platform.index._actions=Actions
-platform.index.emptyvalue=
+platform.index.emptyvalue= -
```
Enygma2002: Unfortunately this is not the solution. Please see the updated description of the jira issue https://jira.xwiki.org/browse/XWIKI-14436, including the last comment I added.

PR #703, opened 2018-01-14
XWIKI-6732 : Introduce DocumentMovingEvent and DocumentMovedEvent events 

Srijanjha:
…movejob.java

### DIFF
--- a/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
+++ b/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
@@ -1,0 +1,121 @@
+/*
+  * See the NOTICE file distributed with this work for additional
+  * information regarding copyright ownership.
+  *
+  * This is free software; you can redistribute it and/or modify it
+  * under the terms of the GNU Lesser General Public License as
+  * published by the Free Software Foundation; either version 2.1 of
+  * the License, or (at your option) any later version.
+  *
+  * This software is distributed in the hope that it will be useful,
+  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  * Lesser General Public License for more details.
+  *
+  * You should have received a copy of the GNU Lesser General Public
+  * License along with this software; if not, write to the Free
+  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  */
+ package org.xwiki.bridge.event;
+ 
+ import org.xwiki.model.reference.DocumentReference;
+ import org.xwiki.observation.event.filter.EventFilter;
+ import org.xwiki.observation.event.EndEvent;
+ 
+ /**
+  * An event triggered after a document is moved.
+  * <p>
+  * The event also send the following parameters:
+  * </p>
+  * <ul>
+  * <li>source: the current {com.xpn.xwiki.doc.XWikiDocument} instance</li>
+  * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
+  * </ul>
+  * 
+  * @version $Id$
+  * @since 10.0RC1
+  */
+ @Unstable 
+ public class DocumentMovedEvent extends AbstractDocumentEvent implements EndEvent
+ {
+ +    /**
+      * The version identifier for this Serializable class. Increment only if the <i>serialized</i> form of the class
+      * changes.
+      */
+     private static final long serialVersionUID = 1L;
+     
+    /**
+      * Constructor initializing the event filter with an
+      * {@link org.xwiki.observation.event.filter.AlwaysMatchingEventFilter}, meaning that this event will match any other document move event.
+      */
+    
+     private String move;
+    
+    /**
+      * Matches all {@link DocumentMovedEvent} events.
+      */
+     public DocumentMovedEvent()
+     {
+         super();
+     }
+ 
+     /**
+      * Matches {@link DocumentMovedEvent} events that target the specified document.
+      * 
+      * @param documentReference the reference of the document to match
+      */
+     public DocumentMovedEvent(DocumentReference documentReference)
+     {
+         this(documentReference, null);
+     }
+ 
+     /**
+      * Constructor initializing the event filter with a {@link org.xwiki.observation.event.filter.FixedNameEventFilter},
+      *  meaning that this event will match only move events affecting the same document.
+      */
+      
+      /**
+      * Matches {@link DocumentMovedEvent} events that target the specified document and move. The move is
+      * matched only if it's not {@code null}.
+      * 
+      * @param documentReference the reference of the document to match
+      * @param move the move the document was moved to
+      */
+     public DocumentMovedEvent(DocumentReference documentReference, String move)
+     {
+         super(documentReference);
+         this.move = move;
+     }
+     
+    /**
+      * Constructor using a custom {@link EventFilter}.
+      * 
+      * @param eventFilter the filter to use for matching events
+      */
+     public DocumentMovedEvent(EventFilter eventFilter)
+     {
+         super(eventFilter);
+     }
+ 
+     /**
+      * @return the move the document was moved to
+      */
+     public String getMove()
+     {
+         return move;
+     }
+ 
+     @Override
+     public boolean matches(Object otherEvent)
+     {
+         boolean matches = super.matches(otherEvent);
+ 
+         if (matches) {
+             DocumentMovedEvent documentMovedEvent = (DocumentMovedEvent) otherEvent;
+             matches = move == null || move.equals(documentMovedEvent.getMove());
+         }
+ 
+         return matches;
+     }
+ }
--- a/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovingEvent.java
+++ b/xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovingEvent.java
@@ -1,0 +1,111 @@
+/*
+  * See the NOTICE file distributed with this work for additional
+  * information regarding copyright ownership.
+  *
+  * This is free software; you can redistribute it and/or modify it
+  * under the terms of the GNU Lesser General Public License as
+  * published by the Free Software Foundation; either version 2.1 of
+  * the License, or (at your option) any later version.
+  *
+  * This software is distributed in the hope that it will be useful,
+  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  * Lesser General Public License for more details.
+  *
+  * You should have received a copy of the GNU Lesser General Public
+  * License along with this software; if not, write to the Free
+  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  */
+ package org.xwiki.bridge.event;
+ 
+ import org.xwiki.model.reference.DocumentReference;
+ import org.xwiki.observation.event.filter.EventFilter;
+ import org.xwiki.observation.event.BeginEvent;
+ 
+ /**
+  * An event triggered after a document is moved.
+  * <p>
+  * The event also send the following parameters:
+  * </p>
+  * <ul>
+  * <li>source: the current {com.xpn.xwiki.doc.XWikiDocument} instance</li>
+  * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
+  * </ul>
+  * 
+  * @version $Id$
+  * @since 10.0RC1
+  */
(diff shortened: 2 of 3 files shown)


tmortagne:
You are not sending any of those events. See http://extensions.xwiki.org/xwiki/bin/view/Extension/Observation+Module+Local to get more details on how to send events.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/job/MoveJob.java
```diff
@@ -223,6 +226,19 @@ private void move(DocumentReference oldReference, DocumentReference newReference
             this.progressManager.popLevelProgress(this);
         }
     }
+    
+    private ObservationManager getObservationManager()
```
tmortagne: Why this method ?

Srijanjha: Is this not needed?

vmassol: You're injecting it already.... with `@Inject`.... so it's definitely not needed

tmortagne:
You care still not sending any event. ObservationManager cannot guess what you want to do, you need to use it now.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,121 @@
...
+ import org.xwiki.observation.event.filter.EventFilter;
+ import org.xwiki.observation.event.EndFoldEvent;
+ 
+ /**
+  * An event triggered after a document is moved.
+  * <p>
+  * The event also send the following parameters:
+  * </p>
+  * <ul>
+  * <li>source: the current {com.xpn.xwiki.doc.XWikiDocument} instance</li>
```
tmortagne: This documentation does not match the event sending code.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovingEvent.java
```diff
@@ -0,0 +1,111 @@
...
+ import org.xwiki.observation.event.filter.EventFilter;
+ import org.xwiki.observation.event.BeginFoldEvent;
+ 
+ /**
+  * An event triggered after a document is moved.
+  * <p>
+  * The event also send the following parameters:
+  * </p>
+  * <ul>
+  * <li>source: the current {com.xpn.xwiki.doc.XWikiDocument} instance</li>
```
tmortagne: This documentation does not match the event sending code.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/job/MoveJob.java
```diff
@@ -218,6 +222,8 @@ private void move(DocumentReference oldReference, DocumentReference newReference
             if (this.request.isDeleteSource() && this.request.isAutoRedirect()) {
                 this.modelBridge.createRedirect(oldReference, newReference);
             }
+            DocumentMovedEvent newEvent = new DocumentMovingEvent();
```
tmortagne: You are supposed to send two events and DocumentMovingEvent is supposed to be sent BEFORE the move as suggested by its name.

Also this event is useless without the references.

The move might fail and you need to make sure you always send an EndEvent to "close" the BeginEvent so you should send the DocumentMovedEvent in the finally and make sure DocumentMovedEvent indicate of it failed or not.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+     {
+         super();
+     }
+ 
+     /**
+      * Matches {@link DocumentMovedEvent} events that target the specified document.
+      * 
+      * @param documentReference the reference of the document to match
+      */
+     public DocumentMovedEvent(DocumentReference documentReference)
```
tmortagne: It's really not clear if this is the source or the target.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+ public class DocumentMovedEvent extends AbstractDocumentEvent implements EndFoldEvent
+ {
+ +    /**
+      * The version identifier for this Serializable class. Increment only if the <i>serialized</i> form of the class
+      * changes.
+      */
+     private static final long serialVersionUID = 1L;
+     
+    /**
+      * Constructor initializing the event filter with an
```
tmortagne: This javadoc does not make sense, seems to be associated with nothing.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+      * {@link org.xwiki.observation.event.filter.AlwaysMatchingEventFilter}, meaning that this event will match any other document move event.
+      */
+    
+     private String move;
```
tmortagne: You should use DocumentReference instead of String to refer to documents. Also "move" does not make it clear if it's related to source or target.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+ import org.xwiki.observation.event.EndFoldEvent;
+ 
+ /**
+  * An event triggered after a document is moved.
+  * 
+  * @version $Id$
+  * @since 10.0RC1
+  */
+ @Unstable 
+ public class DocumentMovedEvent extends AbstractDocumentEvent implements EndFoldEvent
```
tmortagne: You should not extend AbstractDocumentEvent IMO. Just makes the event way more complicated than it should be.

You should have an event with a source and a target and ways to match them.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+      */
+      
+      /**
+      * Matches {@link DocumentMovedEvent} events that target the specified document and move. The move is
+      * matched only if it's not {@code null}.
+      * 
+      * @param documentReference the reference of the document to match
+      * @param move the move the document was moved to
+      */
+     public DocumentMovedEvent(DocumentReference documentReference, String move)
```
tmortagne: Use DocumentReference to refer to documents, not String.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+         super(documentReference);
+         this.move = move;
+     }
+     
+    /**
+      * Constructor using a custom {@link EventFilter}.
+      * 
+      * @param eventFilter the filter to use for matching events
+      */
+     public DocumentMovedEvent(EventFilter eventFilter)
```
tmortagne: Forget about EventFilter and all other String related stuff. This is old and now bad design.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,114 @@
...
+     {
+         super(eventFilter);
+     }
+ 
+     /**
+      * @return the move the document was moved to
+      */
+     public String getMove()
```
tmortagne: No String.

tmortagne:
Does not build.

Srijanjha:
But it shows no error while building with maven

tmortagne:
Make sure you build the right module, for example it's impossible that DocumentMovedEvent build.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+  * </p>
+  * <ul>
+  * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
+  * </ul>
+  * 
+  * @version $Id$
+  * @since 10.0RC1
+  */
+ @Unstable 
+ public class DocumentMovedEvent implements EndFoldEvent
```
tmortagne: This event should indicate the source and the target document references. It does not really makes any sense right now.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovingEvent.java
```diff
@@ -0,0 +1,63 @@
...
+  * </p>
+  * <ul>
+  * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
+  * </ul>
+  * 
+  * @version $Id$
+  * @since 10.0RC1
+  */
+ @Unstable 
+ public class DocumentMovingEvent implements BeginFoldEvent
```
tmortagne: This event should indicate the source and the target document references. It does not really makes any sense right now.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+ import org.xwiki.model.reference.DocumentReference;
+ import org.xwiki.observation.event.EndFoldEvent;
+ 
+ /**
+  * An event triggered after a document is moved.
+  * <p>
+  * The event also send the following parameters:
+  * </p>
+  * <ul>
+  * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
```
tmortagne: There is no much point in sending the XWikiContext, especially since you don't have the XWikiContext when you send this event. You should also explicitly indicate what's in the `source` parameter (i.e. null here I guess).

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentMovedEvent.java
```diff
@@ -0,0 +1,62 @@
...
+  * </ul>
+  * 
+  * @version $Id$
+  * @since 10.0RC1
+  */ 
+ public class DocumentMovedEvent extends AbstractDocumentEvent implements EndFoldEvent
```
tmortagne: As I said in a previous comment, you should not extend AbstractDocumentEvent for new events IMO. It's quite old and deprecated now (String based) plus it bring more complexity than simplification here (you don't have a specific document but two).

slauriere: I'm interested in this pull request. I have two questions:
- What class should DocumentMovedEvent extend instead of AbstractDocumentEvent? (I had a look at the existing document events and I could not find one extending a different class so far). 
- Since this request seems to be stalled, can I ask you for a hint on how to proceed? I guess I should merge the existing pull request into an xwiki-platform fork, modify it, and create a new pull request out of it, is that right?

vmassol: @tmortagne Why not deprecate `AbstractDocumentEvent` then and recommend to use `AbstractEntityEvent` instead then? Except that ideally we need to get away from oldcore which was the reason for introducing `AbstractDocumentEvent` in the bridge module initially ;)

tmortagne: As I said documents events are based on AbstractDocumentEvent so it does not make much sense to deprecated it unless you plan to deprecated them too and introduce new events.

vmassol: I don't understand the reasoning: I don't see why deprecating `AbstractDocumentEvent` would mean deprecating classes extending it. Deprecating it is an indication that it shouldn't be used from now on.

PR #709, opened 2018-01-17
XWIKI-13496 : FileUpload plugin assumes only one file for a given field name

akshitdewan:


tmortagne:

PR #713, opened 2018-03-03
XWIKI-15091: AbstractWikiReferenceExtractor fails to extract a wiki reference …

aubincleme:
… if the XWiki Context isn't correctly initialized

Check if the XWikiContext is present in the ExecutionContext before trying to use the WikiDescriptorManager

### DIFF
--- a/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-schemes/xwiki-platform-url-scheme-standard/src/main/java/org/xwiki/url/internal/standard/AbstractWikiReferenceExtractor.java
+++ b/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-schemes/xwiki-platform-url-scheme-standard/src/main/java/org/xwiki/url/internal/standard/AbstractWikiReferenceExtractor.java
@@ -34,6 +34,8 @@ import org.xwiki.wiki.manager.WikiManagerException;
  */
 public abstract class AbstractWikiReferenceExtractor implements WikiReferenceExtractor
 {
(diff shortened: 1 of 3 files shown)


aubincleme:
@tmortagne ; @vmassol I'm more concerned about whether this issue (that is related to https://jira.xwiki.org/browse/XWIKI-14982) should be treated or not than I am about the fix itself ^^

aubincleme:
The point here is not to access the `XWikiContext` but to check if it has been initialized and is present in the current `ExecutionContext` (it's the case during the first request).

vmassol:
BTW I see that this PR is still open (and https://jira.xwiki.org/browse/XWIKI-15091 is also still open)... We should do something about it.

tmortagne:
> The point here is not to access the XWikiContext but to check if it has been initialized and is present in the current ExecutionContext (it's the case during the first request).

Actually there is a @ Named("reaonly") Provider<XWikiContext> for this use case (return the XWikiContext if it exist, null otherwise).

aubincleme:
> Actually there is a @ Named("readonly") Provider for this use case (return the XWikiContext if it exist, null otherwise).

But then, we would need to add a dependency to the oldcore on this module (`xwiki-platform-url-scheme-standard`) which is currently not the case. The solution using `@Named` could be indeed very nice, but it requires some refactoring.

tmortagne:
> we would need to add a dependency to the oldcore on this module

Sorry, did not looked at the details, was just reacting to the previous conversion :)

PR #714, opened 2018-03-12
XWIKI-15108: Refresh the default Color Theme for XWiki 10.x

evalica:


evalica:
The question is if we should have a generic xwiki-platform-flamingo-theme-default module or xwiki-platform-flamingo-theme-iceberg?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-themes/pom.xml
```diff
@@ -33,6 +33,7 @@
   <description>XWiki Platform - Flamingo - Themes - Parent POM</description>
   <modules>
     <module>xwiki-platform-flamingo-theme-bootswatch</module>
+    <module>xwiki-platform-flamingo-theme-default</module>
```
vmassol: Why are you creating a new module? Why not put it inside the existing `*-theme-ui` one?

evalica: How can I define a dependency towards org.webjars.bower.open-sans just for 1 CT in that *-theme-ui one?

vmassol: What does it change if it's needed only by one CT but is defined in the pom.xml for all CTs? 

We don't create a module every time there's a java class needing a JAR dependency :)

evalica: I found the module solution to be more clean, allowing uninstall, active count, documentation. 
Sure I can put them all in the same folder, still I don't see this as optimal.

vmassol: There's a bad practice in maven land to have too many modules. I've been there in a previous life and it's not nice. There are already several places where we have used too many modules in xwiki land. We should only add new modules when there's a need. I don't see the case for this ATM.

evalica: k - thank you for your feedback Vincent.

evalica:
Done in https://github.com/xwiki/xwiki-platform/commit/6d211c7000beb785171e7f03748e108c9a18da95

PR #715, opened 2018-03-27
syntax error

Pbast:
Hello a tiny syntax error fixed.


tmortagne:
Actually the right place to fix that is http://l10n.xwiki.org/xwiki/bin/view/Platform/PlatformXWikiAdminTranslationsxml_-836307046_administration-sectionTitle-page_fr. Translations located on github gets overwritten by what's coming from l10n.xwiki.org during each release.

PR #716, opened 2018-04-05
XWIKI-15161: XWikiHibernateStore ConcurrentModificationException

msladek:
https://jira.xwiki.org/browse/XWIKI-15161

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
@@ -47,6 +47,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.EntityMode;
 import org.hibernate.FlushMode;
+import org.hibernate.HibernateException;
 import org.hibernate.ObjectNotFoundException;
 import org.hibernate.Query;
 import org.hibernate.Session;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -1368,17 +1369,20 @@ private void loadXWikiCollectionInternal(BaseCollection object1, XWikiDocument d
...
+                        for (String prop : bclass.getCustomMappingPropertyList(context)) {
+                            if (map.get(prop) != null) {
+                                handledProps.add(prop);
                             }
                         }
                     }
-                } catch (Exception e) {
+                } catch (HibernateException e) {
+                    this.logger.error("Failed loading custom mapping for doc [{}], class [{}], nb [{}]", 
+                            object.getDocumentReference(), object.getXClassReference(), object.getNumber(), e);
```
vmassol: Question: shouldn't we throw an exception if custom mapping failed to load? Isn't that an important enough problem?

PS: I know this is not strictly related to your PR since what was there before was even worse, but it got me wondering.

msladek: Perhaps a `HibernateException` is thrown in case a class mapping is set to internal, but no actual hibernate mapping is given for the class? I'm not sure if that is the case.
If yes, a valid fallback would be to continue and load from `xwikiproperties`.
If no, I would expect the returned map to be empty and the exception should be wrapped and re-thrown in the catch block.

tmortagne:

PR #717, opened 2018-04-11
Stable 9.11.x

smallez:
the xwiki version installed is 9.11.3,after installed,i can edit,but when i want to upload file, it happened an error : xwiki/bin/get/Main/wiki%E4%BD%BF%E7%94%A8%E8%AF%B4%E6%98%8E/?sheet=CKEditor.FileUploader&outputSyntax=plain&form_token=cw0h6dVtu4Dku1cJsxUuVQ&responseType=json] ERROR o.x.v.i.DefaultVelocityEngine  - Exception in macro #handleUploadRequest called at 40:xwiki:CKEditor.FileUploader[line 2, column 1] 
can you help to see what's the cause?


mflorea:
This is not the right way/place to ask questions. Please see http://www.xwiki.org/xwiki/bin/view/Main/Support#HCommunitySupport . Use the forum instead.

PR #718, opened 2018-04-18
XWIK-15129: Add navigation (previous/next version) buttons in the changes view

atallahade:
See https://jira.xwiki.org/browse/XWIKI-15129


xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -276,6 +279,64 @@ public void testUnifiedDiffShowsInlineChanges()
...
+        assertTrue(getUtil().getDriver().hasElement(By.id("arrow-right-shift")));
+
+        changesPane.getArrowRightTo().click();
+        changesPane = new ChangesPane();
+        assertTrue(changesPane.getFromVersionSummary().startsWith("From version < 1.2 >"));
+        assertTrue(changesPane.getToVersionSummary().startsWith("To version < 6.5"));
+        assertTrue(getUtil().getDriver().hasElement(By.id("arrow-left-shift")));
+        assertTrue(!getUtil().getDriver().hasElement(By.id("arrow-right-shift")));
+
+        changesPane.getArrowLeftTo().click();
```
vmassol: This should go in the pageobject (PO) since it's the responsbility of the PO to perform the actions.

atallahade: So I should create a `click` method for each arrows instead of the `get`?

tmortagne: The general idea is that PO's job to hide technical details and it should expose hight level API, it should not even talk about arrows (might be something else later). For example I would not "create a click method for each arrows" but instead have method corresponding to what we want to do, things like: switchToPreviousOriginalVersion(), switchToNextVersion(), etc.

vmassol: Note that for me it's acceptable to use `clickXXX()` because that's an action you can do on the page represented by the PO (we do that in lots of places). What Thomas mentioned is an even higher level of abstraction that would require less maintenance should we replace the click by something else but I wouldn't consider that to be mandatory. It's also simpler to understand that `clickXXX()` represents a click in the UI so you know exactly what the API you use will do in your test.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,41 @@
...
+  content: '<';
+  color: @xwiki-page-content-bg;
+  float: left;
+}
+
+.changes-arrow-right::after {
+  content: '>';
+  float: right;
+  position: absolute;
+  right: 30px;
```
evalica: How does this behave on responsive viewports?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,41 @@
...
+  }
+}
+
+.changes-version {
+  white-space: nowrap;
+}
+
+.changes-arrow-left::before {
+  content: '<';
+  color: @xwiki-page-content-bg;
```
evalica: this code is duplicate with line 357

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,41 @@
...
+  float: left;
+}
+
+.changes-arrow-right::after {
+  content: '>';
+  float: right;
+  position: absolute;
```
evalica: Do you need float and position in the same time? 1 method overwrites the other

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,41 @@
...
+  right: 30px;
+  top: 0;
+}
+
+.changes-arrow-left::before, .changes-arrow-right::after {
+  color: @xwiki-page-content-bg;
+  font-size: 7em;
+  font-weight: 700;
```
evalica: can be grouped with line 328

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,34 @@
...
+
+.changes-arrow-left {
+  position: absolute;
+  top: 0;
+  left: 30px;
+}
+
+.changes-arrow-right {
+  position: absolute;
+  top: 0;
```
evalica: Ideally, you could have used percentages to calculate the middle of the container.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,34 @@
...
+  }
+}
+
+.changes-arrow-left, .changes-arrow-right {
+  font-size: 7em;
+}
+
+.changes-arrow-left {
+  position: absolute;
+  top: 0;
```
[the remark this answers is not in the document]
atallahade: Done. Thanks

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
atallahade: Done, using `transform: translateY(-50%);`

atallahade:
@vmassol, @tmortagne, @evalica Is there anything else you could say about this PR?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -324,6 +324,33 @@
   }
 }
 
+.changes-version {
+  white-space: nowrap;
+}
+
+.changes-arrow, .changes-arrow-big, .changes-arrow-left, .changes-arrow-right {
```
evalica: Do you know where the ".changes-arrow-big" is used?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -32,6 +58,12 @@
       #end
       </div> ## changes-info-content
     </div> ## changes-info-box-to
+    #if ("$!previousLink" != '')
+      <a href="$previousLink" id="arrow-left-shift" class="changes-arrow-left">&lt;</a>
+    #end
+    #if ("$!nextLink" != '')
+      <a href="$nextLink" id="arrow-right-shift" class="changes-arrow-right">&gt;</a>
```
evalica: On the CSS part I don't have anything else to add. 
On the HTML, do you really need all those IDs and classes: #arrow-right-shift, #arrow-left-shift, .changes-arrow-left, .changes-arrow-right, .changes-arrow ?

atallahade: I needed the classes for the css. I've added the ids to be able to easily locate the elements for the functional tests. I know it's a bit uggly so I can try to remove the ids and locate the element using XPath (it's currently used for some tests with our version of Selenium I think)

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -276,6 +279,64 @@ public void testUnifiedDiffShowsInlineChanges()
...
+    private void testVersionNavigation()
+    {
+        String queryString = "viewer=changes&rev1=1.2&rev2=6.4";
+        getUtil().gotoPage(getTestClassName(), testPage.getMetaDataValue("page"), "view", queryString);
+        ChangesPane changesPane = new ChangesPane();
+
+        assertTrue(changesPane.getFromVersionSummary().startsWith("From version < 1.2 >"));
+        assertTrue(changesPane.getToVersionSummary().startsWith("To version < 6.4 >"));
+
+        By arrowLeftShiftSelector = By.cssSelector("#changes-info-boxes > a.changes-arrow-left");
```
Enygma2002: This is still exposing/leaking implementation details of the UI into the test.

As an example: If the UI is slightly updated (i.e. class name changed) and you have 5 tests for that UI, you will need to update the UI + 5 tests. If, instead, you encapsulate the UI's implementation details inside the page object class, when the UI changes, you only update the page object and all the tests that are using it will not need to be changed.

You have added the methods to the `ChangesPage` page object in order to click on the links. If you need to check in your tests that you *can* click on those links, i.e. if they exist, then you can extend the page object even more and add methods like:
* hasPreviousFromVersion()
* hasNextFromVersion()
* hasPreviousToVersion()
* hasNextToVersion()
* hasPreviousChange()
* hasNextChange()

Also note when designing the page object that you want to expose the functionality of the UI and not its implementation. E.g., a better name for `clickArrowLeftShift` would be `clickPreviousChange` and instead of `clickArrowLeftFrom` would be `clickPreviousFromVersion`, in the sense that the `arrowLeft` is an implementation detail that might change at a future redesign, while the functionality of "previousChange"/"previousVersion" will continue to exist in the UI.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -276,6 +279,64 @@ public void testUnifiedDiffShowsInlineChanges()
...
+        By arrowRightShiftSelector = By.cssSelector("#changes-info-boxes > a.changes-arrow-right");
+
+        changesPane.clickArrowLeftShift();
+        changesPane = new ChangesPane();
+        assertTrue(changesPane.getFromVersionSummary().startsWith("From version 1.1 >"));
+        assertTrue(changesPane.getToVersionSummary().startsWith("To version < 6.3 >"));
```
Enygma2002: I'm afraid there has been a misunderstanding on the functionality of the left and right outer "arrows".

The idea is to move to the previous or to the next *change* and not to *shift* the currently defined "changes window". The "changes window" is controlled by the use of the left/right from and left/right to arrows, while the 1-by-1 change navigation is controlled by the outer-most arrows.

In this case:
* the previous change for the window "1.2 > 6.4" should be "1.1 > 1.2"
* the next change for the window "1.2 > 6.4" should be "6.4 > 6.5 (or 7.1, or whatever the next version is, compared to 6.4)"

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -276,6 +279,64 @@ public void testUnifiedDiffShowsInlineChanges()
...
+        assertTrue(!getUtil().getDriver().hasElementWithoutWaiting(arrowLeftShiftSelector));
+        assertTrue(getUtil().getDriver().hasElementWithoutWaiting(arrowRightShiftSelector));
+
+        changesPane.clickArrowRightShift();
+        changesPane = new ChangesPane();
+        assertTrue(changesPane.getFromVersionSummary().startsWith("From version < 1.2 >"));
+        assertTrue(changesPane.getToVersionSummary().startsWith("To version < 6.4 >"));
```
Enygma2002: As mentioned above, here it should be "1.2 > 1.3".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesinline.vm
```diff
@@ -64,5 +64,9 @@
     #end
   #end
 #end
+#set ($previousOrigdoc = $origdoc.previousVersion)
+#set ($previousNewdoc = $newdoc.previousVersion)
+#set ($nextOrigdoc = $origdoc.documentArchive.getNextVersion($origdoc.RCSVersion))
+#set ($nextNewdoc = $newdoc.documentArchive.getNextVersion($newdoc.RCSVersion))
```
Enygma2002: It might be a good idea to include the word `version` inside these variable names, otherwise their types can be misunderstood.

For example, `$previousOrigdoc` does not actually contain the previous original document object, but only a version (String or Version object -- it might be a good idea to cast the Version object to String, for consistency).

Note: A good idea at this point would be to observe that we have a `getPreviousVersion` method for a document but we are missing the corresponding `getNextVersion` method. It would be the perfect opportunity to extend the API and add the missing method (to be used when navigating through the document history, as we are doing here), instead of resorting to use the protected API and access the document archive.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -250,6 +251,8 @@ public void testAllChanges()
         assertDiff(ageDiff.getDiff("Pretty Name"), "+age");
         assertDiff(ageDiff.getDiff("Size"), "+30");
         assertDiff(ageDiff.getDiff("Number Type"), "+integer");
+
+        testVersionNavigation();
```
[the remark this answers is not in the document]
vmassol: No this not correct Enygma. I actually told Adel to do it this way since this is the way it should be. Functional tests need to have a single test method with a scenario. That's because they're slow to execute and you do not want to incur any setup penalty.

See http://dev.xwiki.org/xwiki/bin/view/Community/Testing#HBestpractices-2

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesinline.vm
atallahade: > A good idea at this point would be to observe that we have a `getPreviousVersion` method for a document but we are missing the corresponding `getNextVersion` method. It would be the perfect opportunity to extend the API and add the missing method

That's actually what I wanted to do but I've been told that a `getNextVersion` method wouldn't be generic enough. What do you think @vmassol?

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -276,6 +279,94 @@ public void testUnifiedDiffShowsInlineChanges()
...
+        assertTrue(changesPane.hasPreviousFromVersion());
+        assertTrue(changesPane.hasNextFromVersion());
+        assertTrue(changesPane.hasPreviousToVersion());
+        assertTrue(changesPane.hasNextToVersion());
+
+        changesPane.clickPreviousChange();
+        changesPane = new ChangesPane();
+        assertEquals(changesPane.getFromVersion(), "1.1");
+        assertEquals(changesPane.getToVersion(), "1.2");
+        assertTrue(!changesPane.hasPreviousChange());
```
Enygma2002: As a general note, it's clearer and recommended to use the dedicated `assertFalse`.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -276,6 +279,94 @@ public void testUnifiedDiffShowsInlineChanges()
...
+        changesPane = new ChangesPane();
+        assertEquals(changesPane.getFromVersion(), "1.3");
+        assertEquals(changesPane.getToVersion(), "1.3");
+        assertTrue(changesPane.hasPreviousChange());
+        assertTrue(changesPane.hasNextChange());
+        assertTrue(changesPane.hasPreviousFromVersion());
+        assertTrue(changesPane.hasNextFromVersion());
+        assertTrue(changesPane.hasPreviousToVersion());
+        assertTrue(changesPane.hasNextToVersion());
+
```
Enygma2002: It would have been interesting to continue and have From `1.4` To `1.3`, which should still be a supported case.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -30,6 +30,7 @@
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.openqa.selenium.By;
```
Enygma2002: This looks like it's not being used. Should probably be removed.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesinline.vm
```diff
@@ -64,5 +64,9 @@
     #end
   #end
 #end
+#set ($previousOrigdocVersion = $origdoc.previousVersion)
+#set ($previousNewdocVersion = $newdoc.previousVersion)
+#set ($nextOrigdocVersion = $origdoc.documentArchive.getNextVersion($origdoc.RCSVersion))
```
tmortagne: Did you check if this is working for extension version ?

tmortagne: I'm not really discussion the theory here. I just want to make sure @atallahade tested this use case and those lines don't crash or end up providing the 2.1 version of the current document as next version (which would not make any sense) when dealing with non-history documents since manipulating the archive of this kind of document never really been a use case before.

atallahade: @tmortagne I tried the feature with an extension. It seems that the extension version is not reachable when comparing two normal versions (made on the wiki). For instance we can have this kind of history:
```
1.1 <--> 1.2 <--> 2.1 <--> 2.2
```
An extension version will have this type of name `xar:org.xwiki.contrib:application-help-center/1.2` and will put itself before a version, like this:
```
                                 1.1 <--> 1.2 <--> 2.1 <--> 2.2
xar:org.xwiki.contrib:application-help-center/1.2 --^
```

If you compare `xar:org.xwiki.contrib:application-help-center/1.2` to `2.1`, you will not be able to access previous changes (no left arrows will be displayed). Other arrows work fine.

tmortagne: > An extension version will have this type of name xar:org.xwiki.contrib:application-help-center/1.2 and will put itself before a version, like this:

Not really sure what you mean here. The extension version is always displayed first in the history UI and no code produce an ordered list with the extension version in the middle of it (that would not make any sense) unless it's a new thing you added.

> If you compare xar:org.xwiki.contrib:application-help-center/1.2 to 2.1, you will not be able to access previous changes (no left arrows will be displayed). Other arrows work fine.

IMO there should be either no arrow at all (comparing with an extension version disable previous/next feature) or have previous and next only apply to the non extension version (i.e. compare the extension version with previous or next document version).

atallahade: Fixed by #719

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -2,6 +2,26 @@
 ##
 ## Preamble: which versions are being compared
 ##
+
+#if ("$!previousOrigdocVersion" != '')
+  #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${previousOrigdocVersion}&rev2=${rev2}")})
+  #set ($previousOrigdocLink = "<a class='changes-arrow' href='$changesLink'>&lt;</a>")
```
Enygma2002: Just realized that we are missing a `title` for these links, to describe the action when the user hovers the link with the mouse.

They could be generic for the from/to versions, so that we need only 2 translation keys, i.e. "Previous version" and "Next version".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -32,6 +54,12 @@
       #end
       </div> ## changes-info-content
     </div> ## changes-info-box-to
+    #if ("$!previousLink" != '')
+      <a href="$previousLink" class="changes-arrow-left">&lt;</a>
```
Enygma2002: In this case, a corresponding title translation key could be "Previous change".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -32,6 +54,12 @@
...
+    #end
+    #if ("$!nextLink" != '')
+      <a href="$nextLink" class="changes-arrow-right">&gt;</a>
```
Enygma2002: ...and here "Next change".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -2,6 +2,26 @@
...
 ##
+
+#if ("$!previousOrigdocVersion" != '')
+  #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${previousOrigdocVersion}&rev2=${rev2}")})
+  #set ($previousOrigdocLink = "<a class='changes-arrow' href='$changesLink'>&lt;</a>")
+  #set ($previousLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${previousOrigdocVersion}&rev2=${rev1}")})
+#end
+#if ("$!nextOrigdocVersion" != '')
+  #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${nextOrigdocVersion}&rev2=${rev2}")})
+  #set ($nextOrigdocLink = "<a class='changes-arrow' href='$changesLink'>&gt;</a>")
```
Enygma2002: ..."Next version"

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -2,6 +2,26 @@
...
+#end
+#if ("$!previousNewdocVersion" != '')
+  #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev1}&rev2=${previousNewdocVersion}")})
+  #set ($previousNewdocLink = "<a class='changes-arrow' href='$changesLink'>&lt;</a>")
```
Enygma2002: ..."Previous version"

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -2,6 +2,26 @@
...
+#end
+#if ("$!nextNewdocVersion" != '')
+  #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev1}&rev2=${nextNewdocVersion}")})
+  #set ($nextNewdocLink = "<a class='changes-arrow' href='$changesLink'>&gt;</a>")
```
Enygma2002: ..."Next version"

atallahade:
@Enygma2002

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesinline.vm
Enygma2002: @tmortagne, @atallahade Did you guys reach a conclusion?

AFAIU, right now, there would still be a changes arrow (left or right), next to the document version. So if you are comparing `xar:org.xwiki.contrib:application-help-center/1.2` to `2.1`, you would have a "nextChange" arrow, on the outer right side. Clicking it, would lead to comparing `2.1` to `2.2`.

I personally don't find that too bad, only that it would be lacking a bit of consistency. If you can go to the next change, why not be able to go to the previous change as well? What would that mean in this context? I would say that, if we can only stay relative to the only document version we have (i.e. `2.1`), previous change (i.e. left outer arrow) should lead to comparing `1.2` to `2.1` (i.e. "ignoring" the extension version completely).

Does this make sense?

atallahade: It's kind of the same problem as when the "From" version is higher than the "To" version (i.e. comparing `2.1` to `1.2`), where should the next change button lead to?
For the case of comparing an extension version, I think there should be just no changes arrow at all.
As for the arrows around the extension version, they will never be display as an extension doesn't have a previous/next version anymore.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -24,10 +24,18 @@
   #set ($nextNewdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&gt;</a>")
 #end
 #if ("$!previousOrigdocVersion" != '' && "$!previousNewdocVersion" != '')
-  #set ($previousLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${previousOrigdocVersion}&rev2=${rev1}")})
+  #if ($origdoc.RCSVersion.isLessThan($newdoc.RCSVersion))
```
Enygma2002: This should probably be `isLessOrEqualThan` instead, since you don't want to start considering the "reverse order" logic when comparing 2 equal versions.

Also, since this can be tricky to follow, it might be worth to document the 2 edge cases we are addressing here:
* extension version or end/beginning of history version
* reverse order navigation, preserving the reversed direction

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/changesdoc.vm
```diff
@@ -24,10 +24,18 @@
...
+    #set ($previousLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${previousOrigdocVersion}&rev2=${rev1}")})
+  #else
+    #set ($previousLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev2}&rev2=${previousNewdocVersion}")})
+  #end
 #end
 #if ("$!nextOrigdocVersion" != '' && "$!nextNewdocVersion" != '')
-  #set ($nextLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev2}&rev2=${nextNewdocVersion}")})
+  #if ($origdoc.RCSVersion.isLessThan($newdoc.RCSVersion))
```
Enygma2002: Same here.

atallahade:
@Enygma2002

Enygma2002:
Thanks for your contribution and patience, @atallahade!

PR #721, opened 2018-05-03
XWIKI-15234: HTML Meta information about XWiki documents are incorrec…

atallahade:
…t in the case of translations

* Use the version of the translated document for the data attribute
* Add the locale data attribute
* Use locale when possible to generate document rest urls

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
@@ -22,8 +22,7 @@
   #xwikiDataAttribute('rest-url', $services.rest.url($tdoc.documentReferenceWithLocale))
   #xwikiDataAttribute('form-token', "$!{services.csrf.token}")
   #xwikiDataAttribute('user-id', $!services.model.serialize($xcontext.userReference, 'default'))
-  #xwikiDataAttribute('locale', $tdoc.locale)
-  #xwikiDataAttribute('locale-real', $tdoc.realLocale)>
+  #xwikiDataAttribute('locale', $tdoc.locale)>
   <head>
     ## ---------------------------------------------------------------------------------------------------------------
     ## Ensure that the Content-Type meta directive is the first one in the header.


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/url/resources/DocumentRestURLGenerator.java
```diff
@@ -44,6 +46,13 @@ public URL getURL(DocumentReference reference) throws XWikiRestException
         // The idea is to use the UriBuilder of jax-rs to generate URLs that match the resources paths.
         // So it is consistent.
         try {
+            Locale locale = reference.getLocale();
+            // Get the URL of the translated document if possible
+            if (locale != null && !locale.toString().isEmpty()) {
+                return Utils.createURI(getBaseURI(), PageTranslationResource.class,
+                    reference.getWikiReference().getName(), getSpaceList(reference.getLastSpaceReference()),
+                    reference.getName(), reference.getLocale()).toURL();
+            }
```
vmassol: Small detail: could you introduce some variable so that the methods only has a single "return" statement? That makes it simpler to read and less error prone. Thx

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/url/resources/DocumentRestURLGenerator.java
```diff
@@ -44,6 +46,13 @@ public URL getURL(DocumentReference reference) throws XWikiRestException
         // The idea is to use the UriBuilder of jax-rs to generate URLs that match the resources paths.
         // So it is consistent.
         try {
+            Locale locale = reference.getLocale();
+            // Get the URL of the translated document if possible
```
vmassol: I would replace "if possible" by "if the passed reference defines a Locale". WDYT?

vmassol:
Very nice PR @atallahade TBH I'm impressed you were able to generate it on your own since you've joined the project so recently :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
```diff
@@ -18,10 +18,11 @@
...
   #xwikiDataAttribute('space', $doc.space)##deprecated, use 'reference' instead
   #xwikiDataAttribute('page', $doc.name)##deprecated, use 'reference' instead
-  #xwikiDataAttribute('version', $doc.version)
-  #xwikiDataAttribute('rest-url', $services.rest.url($doc.documentReference))
+  #xwikiDataAttribute('version', $tdoc.version)
+  #xwikiDataAttribute('rest-url', $services.rest.url($tdoc.documentReferenceWithLocale))
   #xwikiDataAttribute('form-token', "$!{services.csrf.token}")
-  #xwikiDataAttribute('user-id', $!services.model.serialize($xcontext.userReference, 'default'))>
+  #xwikiDataAttribute('user-id', $!services.model.serialize($xcontext.userReference, 'default'))
+  #xwikiDataAttribute('locale', $tdoc.realLocale)>
```
vmassol: I'm not an expert here but could you explain the usage of `realLocale` vs `locale`?

tmortagne: IMO you should use $tdoc.locale instead, otherwise it's very missleading. What you need most of the time technically is the locale of the document, if you feel it's really necessary you could add the locale of the content in a different metadata.

vmassol: For my knowledge what the difference between the locale of the document and the locale of the document's content?

tmortagne: The default document locale is always "" (Locale.ROOT as in the last locale you get when you go up in the parents of the other locales). You can use the "default locale" property to indicate if the content of this document has been written in a specific locale ("en" for example for translation document usually) but this is just an information, it has no meaning at API level.

atallahade: From the `getRealLocale` method of `com.xpn.xwiki.api.Document`:
> Gets the real locale of the document. The real locale is either the default locale field when the locale field is empty (when the document is the default document) or the locale field otherwise when the document is a translation document

I had some issues when using `$tdoc.locale` as I was getting an empty string when the document was in english (default). OTOH `$doc.realLocale` is used in `xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_language.vm` so I guess it's ok to use it.

tmortagne: > I had some issues when using $tdoc.locale as I was getting an empty string when the document was in english (default)

You are mixing things actually, english is not always the default and you can see the previous comment for details on the rest. Empty string IS the locale of the document from API point of view so if there is only one indicate it should be this one, the other one is an information you may or may not indicate in the UI.

vmassol: ok so using `locale` instead of `realLocale` allows knowing if the content is written in the default locale or not, right? (if we were using `realLocale` we wouldn't have this info)

atallahade: > Empty string IS the locale of the document from API point of view

Ok I'll do the requested change then :)

vmassol:
I know it's a lot to ask... but is your change to DocumentRestURLGenerator covered by an existing test?

atallahade:
@vmassol I guess it is as the method should be executed for most of the pages and we probably test some pages with different locales. I'll run a test coverage test to make sure of it.

vmassol:
Awesome, thanks Adel!

PR #728, opened 2018-05-11
XWIKI-14892: Displays changes in the URL

Remorax:
This PR fixes XWIKI-14892 and shows both revision numbers in the URL when comparing history versions.
The following changes have been made:
- Hidden inputs have been added for viewer params for showing history revisions
- Csrf param has been removed
- The method has been changed from post to get
- Hidden inputs have been added for showminor and viewer params for displaying minor edits
- The changes have been tested and work fine.
Please review ASAP :)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
@@ -49,10 +49,9 @@ $xwiki.ssfx.use('uicomponents/pagination/pagination.css', true)##
 ##
 ##
     #set ($formname = 'historyform')
-    <form id="$formname" action="$doc.getURL('view', "viewer=changes&amp;$docvariant")" method="post">
+    <form id="$formname" action="$doc.getURL('view', "viewer=changes&amp;$docvariant")" method="get">
       <div id="_history">
       ## CSRF prevention
-      <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />
       <input type="hidden" name="language" value="$!xcontext.locale" />
       <table class="table table-striped table-hover responsive-table">
         <caption class="sr-only">$services.localization.render('core.viewers.history.summary', [$escapetool.xml($doc.displayTitle), $versions.get($mathtool.sub($versions.size(), 1)), $versions.get(0)])</caption>
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
@@ -49,10 +49,11 @@ $xwiki.ssfx.use('uicomponents/pagination/pagination.css', true)##
 ##
 ##
     #set ($formname = 'historyform')
-    <form id="$formname" action="$doc.getURL('view', "viewer=changes&amp;$docvariant")" method="get">
+    <form id="$formname" action="$doc.getURL('view', "$docvariant")" method="get">
       <div id="_history">
       ## CSRF prevention
       <input type="hidden" name="language" value="$!xcontext.locale" />
+      <input type="hidden" name="viewer" value="changes" />
       <table class="table table-striped table-hover responsive-table">
         <caption class="sr-only">$services.localization.render('core.viewers.history.summary', [$escapetool.xml($doc.displayTitle), $versions.get($mathtool.sub($versions.size(), 1)), $versions.get(0)])</caption>
 ## Print the table header
@@ -157,10 +158,12 @@ $xwiki.ssfx.use('uicomponents/pagination/pagination.css', true)##
         #if ($xwiki.hasMinorEdit())
           #if ("$!request.showminor" != 'true')
             <input class="btn btn-default" type="submit" name="viewminorversions" value="$services.localization.render('core.viewers.history.showMinorEdits')"
-              onclick="document.forms.${formname}.action='$tdoc.getURL('view', "viewer=history&amp;showminor=true&amp;$docvariant")'; if (document.forms.${formname}.onsubmit) document.forms.${formname}.onsubmit();" />
+              onclick="document.forms.${formname}.action='$tdoc.getURL('view', "$docvariant")'; if (document.forms.${formname}.onsubmit) document.forms.${formname}.onsubmit();" />
           #else
+          <input type="hidden" name="showminor" value="true" />
+          <input type="hidden" name="viewer" value="history" />
             <input class="btn btn-default" type="submit" name="hideminorversions" value="$services.localization.render('core.viewers.history.hideMinorEdits')"
-              onclick="document.forms.${formname}.action='$tdoc.getURL('view', "viewer=history&amp;$docvariant")'; if (document.forms.${formname}.onsubmit) document.forms.${formname}.onsubmit();" />
+              onclick="document.forms.${formname}.action='$tdoc.getURL('view', "$docvariant")'; if (document.forms.${formname}.onsubmit) document.forms.${formname}.onsubmit();" />
           #end
         #end
       </div>


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
```diff
@@ -49,11 +49,11 @@ $xwiki.ssfx.use('uicomponents/pagination/pagination.css', true)##
 ##
 ##
     #set ($formname = 'historyform')
-    <form id="$formname" action="$doc.getURL('view', "viewer=changes&amp;$docvariant")" method="post">
+    <form id="$formname" action="$doc.getURL('view', "$docvariant")" method="get">
       <div id="_history">
       ## CSRF prevention
-      <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />
       <input type="hidden" name="language" value="$!xcontext.locale" />
+      <input type="hidden" name="viewer" value="changes" />
```
vmassol: For my knowledge, could you explain to me why moving `viewer=changes` from the URL to a form param makes it work better? (I can't figure out why). Thanks

Remorax: @vmassol it doesn't work when the method is changed to get and the viewer param is part of the URL in the action field, that's why I made a hidden input field for it the same way its' been done for other params like language, and it worked fine :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
```diff
@@ -158,10 +158,12 @@ $xwiki.ssfx.use('uicomponents/pagination/pagination.css', true)##
         #if ($xwiki.hasMinorEdit())
           #if ("$!request.showminor" != 'true')
             <input class="btn btn-default" type="submit" name="viewminorversions" value="$services.localization.render('core.viewers.history.showMinorEdits')"
-              onclick="document.forms.${formname}.action='$tdoc.getURL('view', "viewer=history&amp;showminor=true&amp;$docvariant")'; if (document.forms.${formname}.onsubmit) document.forms.${formname}.onsubmit();" />
+              onclick="document.forms.${formname}.action='$tdoc.getURL('view', "$docvariant")'; if (document.forms.${formname}.onsubmit) document.forms.${formname}.onsubmit();" />
```
vmassol: Could you explain why you removed "viewer=history"? And why you removed "showminor=1"? Maybe it's obvious but since I don't know this code, I'm curious to know why changing the computed URL in this way makes it work? Thx

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
vmassol: yes but my question is why? Could you explain why please?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
Remorax: @vmassol I did that for the same reason I added a hidden input for viewer param, when the request is changed to GET, with everything else being kept the same, providing params in the action field doesn't work and hidden inputs need to be added.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
```diff
@@ -158,10 +158,12 @@ $xwiki.ssfx.use('uicomponents/pagination/pagination.css', true)##
...
           #else
+          <input type="hidden" name="showminor" value="true" />
+          <input type="hidden" name="viewer" value="history" />
```
vmassol: bad indentation BTW

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
Remorax: As the accepted answer says (https://stackoverflow.com/a/1116026), this is part of how GET works and the params need to be included as hidden input fields but it works for POST if the params are included in the URL. I guess that's part of the specs.
And regarding the $docvariant variable, i guess that can be removed too, I removed that and it works for me :)

vmassol: Note that `$docvariant` is defined in `xwikivars.vm`. So we need to find a solution for it since it won't work and your changes introduce a regression.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
Remorax: Fixed that! let me know if there are any other errors, and I'll make those changes too and push them :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
vmassol: >  I removed that and it works for me :)

What did you test to know that works? Did you setting your wiki in multilingual and verify that the language stays passed. I highly doubt it!

Remorax: Sure, so should I remove $docvariant too? Or can this be merged and someone else can do that since this is outside the scope of the issue?

vmassol: Globally-speaking you need to be a bit more careful Vivek about what you're doing. It's very important that:
* You stop guessing and be sure about the changes your bring, i.e. you can explain each one you do. This means researching and being able to prove things too.
* You also need to very careful to test all the behaviors. XWiki is very rich and has a lof of use cases. We must not break existing behaviors. Most of the time we'll have tests failing to prove the breakage but not always.

Remorax: No I didn't do that, I mean I don't know whether $docvariant is relevant or not, I guess we would need the advice of the person who wrote this code, but as of now it works fine! And if the parameter has been added, I'm assuming there would be a purpose for it too, so its better not to remove it in this PR and if its actually unnecessary, maybe create a separate issue for handling redundant code?

Remorax: Sure @vmassol that's why I didn't remove $docvariant at first, I just added hidden input fields because it doesn't work to add these fields as part of action when the method is changed to GET, because that's part of their functionality

vmassol: > but as of now it works fine! 

From what I see, as of now you've broken some features so no I don't think you can say it works fine...

There's no need to check much: By moving from POST to GET you've removed `$docvariant` and thus broken use cases. As I said, check xwikivars.vm and you'll see how docvariant is set and thus you can deduce use cases where it's not empty to verify if it works (but I'm pretty sure it won't).

vmassol: > Sure @vmassol that's why I didn't remove $docvariant at first, I just added hidden input fields because it doesn't work to add these fields as part of action when the method is changed to GET, because that's part of their functionality

The problem is that you removed `viewer=changes` but not `$docvariant` and this shows you didn't understand what you're doing at all and why you had to remove `viewer=changes` ;)

Remorax: > The problem is that you removed viewer=changes but not $docvariant and this shows you didn't understand what you're doing at all and why you had to remove viewer=changes ;)

No @vmassol I removed viewer=changes but I also added it as a hidden input field and I did that because that's how GET works.
As for $docvariant, I don't think that needs to be removed, like I said it would have a purpose since the person who wrote the code included it.

Remorax: In $docvariant, the language params and the rev params are stored as a variable and they don't have to be passed as a key-value pair, and this is what I guess, you're right, I'm not sure about this. But apart from changing the languages, I have tested for the other use cases. Let me test for language too and I'll let you know what happens, is that fine?

vmassol: >  Let me test for language too and I'll let you know what happens, is that fine?

The use cases can be seen in xwikivars.vm as I said:

```
## Check language option
#set ($languageparams = '')
#if ($xwiki.isMultiLingual() && ("$!{tdoc.realLocale}" != ''))
  #set ($languageparams = "language=${tdoc.realLocale}")
#end
## Check revision option
#if ($request.rev)
  #set ($revparams = "rev=$!{escapetool.url(${request.rev})}")
#else
  #set ($revparams = '')
#end
#if ($revparams != '' && $languageparams != '')
  #set ($docvariant = "$languageparams&amp;$revparams")
#else
  #set ($docvariant = "${languageparams}${revparams}")
#end
```

So as you can see you need to test so that `$languageparams` is not empty and/or `$revparams` is not empty.

Remorax: I checked for $revparams being empty, in which case it just defaults to the top two revisions as it should. When $languageparams is empty it defaults to English.

Remorax: @vmassol I set $languageparams and $revparams to empty individually and both at the same time, and tested if the output I got was the same as earlier, when POST method was being used, and yes it's the same :)

Remorax: Please let me know if the changes are fine and it can be merged :)

vmassol: > I set $languageparams and $revparams to empty individually and both at the same time

This is what I said above: 
"So as you can see you need to test so that $languageparams is not empty and/or $revparams is not empty."

> Please let me know if the changes are fine and it can be merged :)

Obviously no :)

Remorax: Can you please explain what exactly you mean by "$languageparams is not empty and/or $revparams is not empty." because it seems I misunderstood you. 
This is what I understood: I need to test for the cases when $languageparams is not empty and/or $revparams is not empty. So to handle those cases, I took:
- $revparams empty and $languageparams is not empty
- $revparams is not empty and $languageparams is empty
- Both are not empty
And it worked fine for these test cases, did you mean something else?

vmassol:
@Remorax Don't forget to finish this PR! :) Thx

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm
Remorax: @vmassol re my previous reply can you please clear this up once?

vmassol: Remorax, I've already explained it several times... :( I don't know what more to say... I think someone else will need to help you since I've reached the end of my capabilities it seems.

Also you need to explain what you mean by "And it worked fine for these test cases" since obviously that doesn't mean much. For example does it mean you're getting the right language used in the UI?

vmassol: Note that to go further I would need to fix the issue myself (I already did a lot more than I was prepared to do) but the goal here is to have you fix the issue.

Remorax: Yes, I tested it for French and German, in the test cases listed above and it works perfectly.
IMO this can be merged, since I understood your initial concern and have tested it appropriately, and it works perfectly :)
I'm open to feedback from others also, in case they feel anything remains to be tested.

vmassol: > > IMO this can be merged, since I understood your initial concern and have tested it appropriately, and it works perfectly :)

It can't be merged since `$docvariant` is still there:

```
<form id="$formname" action="$doc.getURL('view', "$docvariant")" method="get">
```

And I'm still not 100% convinced that docvariant had no usage in the original code.

atallahade:
I think we can close this PR @vmassol as the issue is closed, right?

vmassol:
Right! Thanks for mentioning it.

vmassol:
And thanks a lot @Remorax for this PR even though we didn't fully applied it.

PR #729, opened 2018-05-14
XWIKI-15251: XWQL queries do not support "group by 1" or "order by 1"

atallahade:
https://jira.xwiki.org/browse/XWIKI-15251

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -643,15 +643,6 @@
                     "methodName": "getAllFilters",
                     "elementKind": "method",
                     "justification": "Young API"
-                  },
-                  {
-                    "code": "java.method.addedToInterface",
-                    "new": "method void org.xwiki.query.jpql.analysis.Analysis::caseAIntegerComparableItem(org.xwiki.query.jpql.node.AIntegerComparableItem)",
-                    "package": "org.xwiki.query.jpql.analysis",
-                    "classSimpleName": "Analysis",
-                    "methodName": "caseAIntegerComparableItem",
-                    "elementKind": "method",
-                    "justification": "Support for queries of type 'GROUP BY 1' and 'ORDER BY 1'"
                   }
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/pom.xml
```diff
@@ -637,6 +637,15 @@
...
                     "justification": "Young API"
+                  },
+                  {
+                    "code": "java.method.addedToInterface",
+                    "new": "method void org.xwiki.query.jpql.analysis.Analysis::caseAIntegerComparableItem(org.xwiki.query.jpql.node.AIntegerComparableItem)",
+                    "package": "org.xwiki.query.jpql.analysis",
+                    "classSimpleName": "Analysis",
+                    "methodName": "caseAIntegerComparableItem",
+                    "elementKind": "method",
+                    "justification": "Support for queries of type 'GROUP BY 1' and 'ORDER BY 1'"
```
vmassol: The justification is not correct. It's a justification explaining why backward compatibility should be broken. And we should not break it, unless in very rare cases.

vmassol: I think the rationale is that this doesn't break backward compatibility because generated code from the grammar is not considered public API. 

@All: BTW maybe we have a mistake and this code should be moved to internal package.

tmortagne: Yes I don't think this code was really meant to be public.

vmassol: Created https://jira.xwiki.org/browse/XWIKI-15265

atallahade: Ok then I'll remove these lines once we fix the issue that you've created.

vmassol: @atallahade I've just closed https://jira.xwiki.org/browse/XWIKI-15265 :)

vmassol:
Thanks @atallahade !

PR #738, opened 2018-05-31

atallahade:
Design page: http://design.xwiki.org/xwiki/bin/view/Proposal/IdeaVisibleSave
Jira: https://jira.xwiki.org/browse/XWIKI-14162

*Wiki editor:*
Wiki editor

*WYSIWYG editor:*
WYSIWYG editor

*Class/Objects editor:*
Class/Objects editor

*Inline form editor:*
Inline form editor

*Responsiveness: (we might want to do something for that)*
Responsiveness

*Fullscreen:*
image
image

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -6,6 +6,18 @@
   .bottombuttons {
     .buttons {
       margin: 10px 0;
+      
+      &.buttons-fixed {
+        z-index: 1030;
+        position: fixed;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        margin: 0;
+        padding: 1em;
+        background-color: white;
+        box-shadow: 0 0 1em rgba(0, 0, 0, 0.3);
+      }
     }
   }
 
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -2,25 +2,34 @@
 // Edit
 // --------------------------------------------------
 
-#editcolumn, #inline {
-  .bottombuttons {
-    .buttons {
-      margin: 10px 0;
-      
-      &.buttons-fixed {
-        z-index: 1030;
-        position: fixed;
-        bottom: 0;
-        left: 0;
-        right: 0;
-        margin: 0;
-        padding: 1em;
-        background-color: white;
-        box-shadow: 0 0 1em rgba(0, 0, 0, 0.3);
-      }
-    }
+
+.bottombuttons {
+  background-color: @xwiki-page-content-bg;
+  .buttons {
+    margin: 0;
+    padding: 1em;
   }
 
+  &.bottombuttons-fixed {
+    z-index: 100;
+    position: fixed;
+    bottom: 0;
+    background-color: @breadcrumb-bg;
+  }
+}
+
+.bottom-editor {
+  border: 1px solid @input-border;
+  border-top: 0;
+  .bottombuttons {
+    background-color: @xwiki-background-secondary-color;
+    &.bottombuttons-fixed {
+      border-top: 1px solid @input-border;
+    }
+  }
+}
+
+#editcolumn, #inline {
   .edit-meta-tools {
     .pull-right();
     margin: @breadcrumb-padding-vertical @breadcrumb-padding-horizontal 0 0;
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
@@ -233,11 +233,18 @@
 
 #xwikieditor .main {
   background-color: @xwiki-page-content-bg;
-  box-shadow: 0 1px 2px @xwiki-border-color;
+  box-shadow: 0 1px 2px @input-border;
 }
 
 #xwikieditcontent .main {
   background-color: @xwiki-page-content-bg;
+}
+
+#xwikieditcontent #content {
+  vertical-align: bottom;
+  border-radius: 0;
+  border-color: @input-border;
+  resize: vertical;
 }
 
 .main,
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
@@ -4,9 +4,11 @@
 
 // Left menu2 =================================================================
 .leftmenu2 {
-  background-color: #F0F0EE;
+  background-color: @xwiki-background-secondary-color;
   min-height: 30px;
   padding: 5px;
+  border: 1px solid @input-border;
+  border-bottom: 0;
   a {
     img {
       margin: 2px;
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.css
@@ -1,11 +1,8 @@
 #template('colorThemeInit.vm')
 .meta-versionSummary label {
-  color: $theme.textSecondaryColor;
-  font-size: .85em;
-  font-weight: 700;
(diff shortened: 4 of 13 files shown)


atallahade:
We might also want to change the save buttons from the preview page:
image

It wasn't on the proposal but I can still do it here, WDYT?

mflorea:
+1 to change the preview mode also. Instead of showing the action buttons both at the top and at the bottom of the preview page like now, I would display them only at the button, like in edit mode, with the "floating" bar.

evalica:
Regarding buttons, you are missing the btn-default styling for "Save" and "Preview". Also please leave "Cancel" as default too, don't make it red :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.css
```diff
@@ -25,5 +22,5 @@
 }
 
 #commentinput {
-  width: 99.7%; /* 100% - 2px border; FIXME when calc() becomes standard */
+  width: inherit;
```
evalica: is this needed? or it can be removed?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/widgets/fullScreen.css
```diff
@@ -67,6 +62,9 @@
 #body .fullScreenWrapper .btn {
   width: auto;
 }
+#body .fullScreenWrapper .buttons .buttonwrapper input {
```
evalica: can you reduce this specificity? lots of classes ... are they all needed?

evalica:
Regading Preview, I would put "Back to edit" also as default. We kind of have just 2 states for buttons: primary and secondary.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.css
atallahade: Yes it's needed, otherwise the width of the comment input of the inline form is changed.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/widgets/fullScreen.css
atallahade: I've copied that from the above line, but it can totally be reduced with
```css
.fullScreenWrapper .buttonwrapper {
```

atallahade:
@evalica, here is the new style from the changes:
image
image

evalica:
Would be nice if we could also fix the styling for the Autosave (remove the bold, etc.)
Regarding styling, the border is bigger than 1px in WYSIWYG mode. Also can we remove it from Preview mode? When it has an editor near it, it attaches to the editor. But in preview mode, it's discrepant from the breadcrumb styling that doesn't have borders.

atallahade:
@evalica:
> Regarding styling, the border is bigger than 1px in WYSIWYG mode. 

WDYM? Should the border always be at most 1px?

> Also can we remove it from Preview mode?

I though it would be a bit weird without borders. Here is what it would looks like:
image

evalica:
It's fine without the border, since it's consistent with the breadcrumb.

atallahade:
Here is what the last changes look like:
*Wiki editor:*


*WYSIWYG editor:*


*Class/Objects editor:*


*Inline form editor:*


Tell me when the PR is ready to be merged so I can squash the commits myself (that way we can keep the first commit made by Pierre).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -14,8 +14,15 @@
     z-index: 100;
     position: fixed;
     bottom: 0;
+    background: @breadcrumb-bg;
```
evalica: 'background-color' was more correct, but minimal change

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -14,8 +14,15 @@
...
+  }
+}
+
+.bottom-editor {
+  border: 1px solid @input-border;
+  border-top: none;
```
evalica: Valid :) still '0' was shorter.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
```diff
@@ -233,13 +233,20 @@
...
 
 #xwikieditcontent .main {
   background-color: @xwiki-page-content-bg;
 }
 
+#xwikieditcontent #content {
+  vertical-align: bottom;
+  border-radius: 0;
+  border-color: @input-border;
+  resize: vertical;
```
evalica: Not supported by IE, see https://caniuse.com/#feat=css-resize
Is this needed?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -4,9 +4,11 @@
 
 // Left menu2 =================================================================
 .leftmenu2 {
-  background-color: #F0F0EE;
+  background-color: @xwiki-background-secondary-color;
   min-height: 30px;
   padding: 5px;
+  border: 1px solid @input-border;
+  border-bottom: none;
```
evalica: '0' was shorter :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/autosave.css
```diff
@@ -20,4 +20,7 @@
 #body #autosaveControl label.frequency, #body #autosaveControl label.frequency span {
   text-transform: none;
   font-weight: normal;
-}
\ No newline at end of file
+}
+#autosaveControl label.autosave {
+  font-weight: normal;
```
evalica: 400 was shorter :) anyway, these shorter comments are more for fun :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
atallahade: Without it:

atallahade: If IE doesn't have a resize at all then it's not a problem.

atallahade:
@evalica the missing top border:

evalica:
For me this PR looks ok. Thanks @atallahade for making all the changes. It looks great! 
We should integrate it and see what bugs / tests we break :)

evalica:
Another question: did you update the Save&Continue translation? Since I see it change in the screenshots, but not sure it was committed.

evalica:
There is still one inconsistency between "Is minor edit" and "Autosave". One is capitalize and the other is not.

atallahade:
@evalica:
image

atallahade:
With space:
image

evalica:
Yeap, there are still some alignment and space issues with the checkboxes. The spacing is important in order to determine which checkbox corresponds to which label.

atallahade:
I've also change the "Is minor edit" to "Minor edit".

evalica:
Things that could still be improved: 
* Responsiveness: 
** Order: Buttons should be the last ones
** Elements display: we could maybe hide some functionality on small displays

atallahade:
@evalica: Thanks for your feedback
> Dashboard - Objects editor display Auto-save on the next line, even if it has enough space

Can you give me a screenshot? There is a "nowrap" on this element:
image
that may be why the auto-save appears on the next line earlier.

> Alignment: still some problems with the input, buttons, checkboxes, labels alignment

Can you be more specific? I see the input and buttons well aligned. I'm not sure about the checkboxes and labels though.

I'll make the changes for the responsiveness.

evalica:
Regarding alignment, something like this is expected
expected
expected_lines

atallahade:
I've also find out that some administration pages are using the `.bottombuttons` container which enables the feature.
image
We should maybe use something else than the `.bottombuttons` class so that we don't have the feature enabled on some pages we are not aware of.
For instance, this page has the sticky save enabled and is quite buggy (we would need to change the extension to fix the problem if we keep using the `.bottombuttons` class!):
random

WDYT?

evalica:
.buttombuttons is not really an XWiki API class, it's used in some places and not used in others. 
Indeed we could add this sticky buttons functionality in other places, I've mentioned the buttons from History, but ideal we should do it in a consistent way. 
Administration is a good candidate since it contains long pages with lots of fields. 
Now we either stick to the .bottombuttons and we add it to http://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/SpecialCSSClasses/ , look for and fix the places where it is used or use another class and control its usage.

atallahade:
I'd rather control its usage by adding a class `.stickybuttons`. I'll do the change, tell me if you think it's fine :).

evalica:
I guess adding a dedicated class that we can document is fine. Not sure we have any naming convention. @mflorea do you have any suggestions for the name?

PR #739, opened 2018-06-05
XWIKI-15321: XWQL "in element()" expression create duplicates results…

atallahade:
… with DBStringListProperty classes

Note that the non relational static lists are not working correctly as of now and their use should probably be deprecated at the expense of relational static lists.


vmassol:
Hi Adel, could we also have a unit test to make sure that XWIKI-15231 doesn't reproduce?

atallahade:
@vmassol Actually I've forgotten to update the unit tests. Let me do that.

PR #745, opened 2018-06-08

atallahade:
See: https://github.com/xwiki/xwiki-platform/pull/748


evalica:
The alignment is good. Still the spacing has not been fixed. The minot edit checkbox is still too close  to the summary input; and the 5 minutes input is not the same size as the other input.

PR #747, opened 2018-06-10
Update from Weblate.

xwikiorgci:


xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/resources/ApplicationResources_fr.properties
```diff
@@ -47,9 +47,9 @@
 #     old.key.name=Some translation
 ###############################################################################
 
-### Missing: extension.xar.log.install.document.begin=Installing document [{}]
-### Missing: extension.xar.log.install.document.success.end=Done installing document [{}]
-### Missing: extension.xar.log.install.document.failure.end=Failed to install document [{}]
+extension.xar.log.install.document.begin=Installation du document [{}] en cours
+extension.xar.log.install.document.success.end=Installation du document [{}] termin\u00E9e
+extension.xar.log.install.document.failure.end=\u00C9chec d''installation du document [{}]
```
vmassol: @tmortagne is it normal that we're not using `{0}`, `{1}`etc for translations with parameters? Also since this probably means we're not using MessageFormat, isn't it wrong to double the singe quote character? (@atallahade).

atallahade: I'll make sure not to double the single quotes if @tmortagne confirms that the MessageFormat is not used in this case.

tmortagne: Yes this is not a parametrized translation but a special log string so it won't go trough MessageFormat.

tmortagne: @vmassol I'm not the one deciding slf4j format :)

tmortagne: > @atallahade You need to match only {<digit>} and not {}I think. Thanks

no you should match {.+} since you have have more that digits in here

vmassol: >  I'm not the one deciding slf4j format :)

True but we're not discussing logging syntax here but translation syntax :) The fact that we trh slf4j syntax for translation is maybe not the best idea. How hard would it be to use our translation syntax (using MessageFormat) and just git the resulting translation to logging if we need to log it?

tmortagne: Again you did not understood: this string is a log message which is passed in Logger#info API.

atallahade: I'll make the change on my side to ignore `{}` when making double single quotes.

vmassol: I understood very well (see my last message). So the answer is that you believe it's ok/a good idea to have several syntax formats for translations. This means the translators need to be careful and not try to understand the parameter syntax and instead copy the format used in the EN version (since there's not way for them to know which syntax is used for a given translation component). We should document this on http://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HTranslationBestPractices and mention that there's no specific format for passing parameters and that it depends on the translation component.

PR #748, opened 2018-06-11
XWIKI-15342: Sticky save should only be displayed in some pages

atallahade:
*Responsiveness:*
Responsiveness
responsiveness2

*Alignment:*

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -2,34 +2,50 @@
 // Edit
 // --------------------------------------------------
 
-
-.bottombuttons {
+.sticky-buttons {
   background-color: @xwiki-page-content-bg;
   .buttons {
     margin: 0;
     padding: 1em;
   }
 
-  &.bottombuttons-fixed {
+  &.sticky-buttons-fixed {
     z-index: 100;
     position: fixed;
     bottom: 0;
     background-color: @breadcrumb-bg;
+    @media (max-width: @screen-md-min) {
+      #commentinput, #autosaveControl, .metamodifs {
+        display: none;
+      }
+    }
+  }
+
+  input {
+    vertical-align: middle;
+    position: relative;
+    margin: 0;
+    bottom: 1px;
   }
 }
 
 .bottom-editor {
-  border: 1px solid @input-border;
-  border-top: 0;
-  .bottombuttons {
+  .sticky-buttons {
+    border: 1px solid @input-border;
+    border-top: 0;
     background-color: @xwiki-background-secondary-color;
-    &.bottombuttons-fixed {
+    &.sticky-buttons-fixed {
       border-top: 1px solid @input-border;
     }
   }
 }
 
 #editcolumn, #inline {
+  :not(.sticky-buttons).bottombuttons {
+    .buttons {
+      margin: 10px 0;
+    }
+  }
   .edit-meta-tools {
     .pull-right();
     margin: @breadcrumb-padding-vertical @breadcrumb-padding-horizontal 0 0;
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editclass.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editclass.vm
@@ -157,7 +157,7 @@ $xwiki.jsfx.use('js/xwiki/editors/dataeditors.js', true)##
     </div></div>
   </div>
 </div> ## properties
-<div class="bottombuttons">#template("editactions.vm")</div>
+<div class="bottombuttons sticky-buttons">#template("editactions.vm")</div>
 </div> ## xwikieditcontent
 </form>
 #end
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editinline.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editinline.vm
@@ -106,7 +106,7 @@ $renderedContent
 ## Action buttons
 ## --------------
 #set($editactions = "bottom")
-<div class="bottombuttons">#template("editactions.vm")</div>
+<div class="bottombuttons sticky-buttons">#template("editactions.vm")</div>
 </div> ## form
 </form>
 #end ## can edit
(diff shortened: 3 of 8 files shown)


atallahade:
@evalica 
> The alignment is good. Still the spacing has not been fixed. The minot edit checkbox is still too close to the summary input; and the 5 minutes input is not the same size as the other input.

I'll do the change :).

atallahade:
*Spacing:*

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -2,34 +2,51 @@
 // Edit
 // --------------------------------------------------
 
-
-.bottombuttons {
+.sticky-buttons {
```
evalica: replacing this class might have some problems in the places that were using it.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -2,34 +2,51 @@
...
     bottom: 0;
     background-color: @breadcrumb-bg;
+    @media (max-width: @screen-sm-max) {
+      #commentinput, #autosaveControl, .metamodifs {
+        display: none;
+      }
+    }
+  }
+
+  input[type=checkbox] {
```
evalica: I would prefer to have the quotation marks, like [type="checkbox"], especially since the value is not numeric

evalica: Also this selector is very generic, hope it won't affect multiple places in a negative way. Would be good if it would mark in a positive way :) Still if it is so generic, I think it would be better suited in forms.less than in edit.less

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -2,34 +2,51 @@
...
+    vertical-align: middle;
+    position: relative;
+    margin: 0;
+    margin-left: 1em;
+    bottom: 1px;
```
evalica: this value sounds strange

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -2,34 +2,51 @@
...
   }
 }
 
 .bottom-editor {
-  border: 1px solid @input-border;
-  border-top: 0;
-  .bottombuttons {
+  .sticky-buttons {
+    border: 1px solid @input-border;
```
evalica: You could have used "@border-width"

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
atallahade: This css rule wasn't used before.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
atallahade: Sure!

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
atallahade: I would rather remove it but since you wanted to have the checkbox aligned with the input, there wasn't many other choices from what I saw. What do you think we can do about that?

evalica: I don't know how it looks exactly, but one thing to check would be to see the value of margin-bottom or padding-bottom. Maybe somewhere is 6px or another value that can be reused. Setting a bottom especially of this size sounds strange. 
If you really need bottom and we don't have a param for 1px, you can have it like this, but just check if it's really needed.

atallahade: @evalica It looks like this by default:
image

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
atallahade: > Also this selector is very generic, hope it won't affect multiple places in a negative way. 

It is only used inside the `.sticky-buttons` elements.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
atallahade: Most of the changed have been introduced by #738.
This code makes sure to keep the previous behavior outside the `.sticky-buttons` element:

```css
#editcolumn, #inline {
  :not(.sticky-buttons).bottombuttons {
    .buttons {
      margin: 10px 0;
    }
  }
...
```

evalica:
So what's still left for this PR? the move of input[checkbox] to forms.less?

atallahade:
@evalica Acutally I was going to use the `checkbox-inline` class and remove the `input[checkbox]` property.

evalica:
k

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -22,11 +22,7 @@
   }
 
   input[type="checkbox"] {
-    vertical-align: middle;
-    position: relative;
-    margin: 0;
     margin-left: 1em;
```
evalica: Not sure why you left this property and didn't move everything.

atallahade: @evalica:
This property is used to have a gap between the checkboxes and other inputs.
If I also move this property to be used gloabally, every checkboxes will then be affected. If you have a checkbox alone on a line, it will seem to be indented.
Maybe a better solution could be not to put a margin to the left of the checkbox but a margin to the right of the labels and inputs next to it, but we have to decide if it should be specific to the "save bar" or to every inputs.

evalica: k

PR #749, opened 2018-06-14
XWIKI-15283: Vertical scroll bar randomly appears

atallahade:
See https://jira.xwiki.org/browse/XWIKI-15283

*Before:*
image

*After:*
image

Note that the section titles (h1 and h2) now appear with the h1 height.


evalica:
The scrollbar appears because of the negative margin. Also this reproduce only on Chrome, not on Firefox.
Aligning the margin and height values fixes the issue.

PR #750, opened 2018-06-15
Fix failing tests

atallahade:


atallahade:
Don't merge this PR yet. I may have found an other solution at the UI level.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -376,19 +376,22 @@ require(['jquery'], function ($) {
   var $container = $('.sticky-buttons');
   if ($container.length == 0) return;
   // Use the placeholder to get the initial container position
-  var $placeholder = $('<span></span>');
+  var $placeholder = $('<div></div>');
+  $placeholder.height(0);
```
vmassol: Would be nice to have some explanations as comments about this! :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -376,19 +376,22 @@ require(['jquery'], function ($) {
...
+    $placeholder.height($container.height());
+    var position = $placeholder.offset().top + $placeholder.height();
 
     if (!isFullScreen && $(window).height() + $(window).scrollTop() < position) {
       $container.addClass('sticky-buttons-fixed');
       $container.innerWidth($container.parent().width());
     } else {
       $container.removeClass('sticky-buttons-fixed');
       $container.innerWidth('');
+      $placeholder.height(0);
```
vmassol: Same

vmassol: Does this fix some problems? Do we need a jira?

atallahade: With this change you won't have anymore "jumps" when the save bar switch from sticky to not sticky as shown on the video below. It also fix the issue when the functional tests try to click on the button just after it moved.

*Before:*
sticky-02

*After:*
sticky-01

vmassol: Do we have a jira for 10.5 about the sticky save? You could commit in it. If not you'll need a new jira.

It would be nice to add some comments in the code to explain why this is needed. Why DIV instead of SPAN, why height(0), etc. I'm worried that in 1 year or more nobody will remember what those are doing and more importantly why they're there. Thx Adel

atallahade: I've introduce the `span` at first so it shouldn't surprise anyone to see a `div` instead. The reason I did this is to make a gap with the placeholder with the sticky save bar height to place the save bar here once it reaches its original position like you could see on the video (when it reaches the bottom).

vmassol: > I've introduce the span at first so it shouldn't surprise anyone to see a div instead.

ok didn't realize you were modifying recent code you added. I thought it was some old code being modified :)

vmassol: Still comments would be most welcome

vmassol: Since you've explained it already, you just need to copy paste it in the right place ;)

PR #754, opened 2018-06-19
XWIKI-15346: Use the compact DBList suggest input displayer for Users and Groups properties

atallahade:
Remove the ReleaseEntryEditPage class as the User picker isn't used here anymore.


vmassol:
Thanks @atallahade Could you change the commit messages to link those to the jira issue that caused the changes?

atallahade:
@vmassol I've tag the PR as WIP as there are other places to make changes.

atallahade:
There is still one problem to solve to make the tests pass successfully. While editing an AWM structure, a reloaded user field has its `data-propertyname` wrong as shown on this video:

awm-autocomplete

@mflorea maybe you can help, the code that modify the `data-propertyname` can be found [here](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml#L384).

PR #755, opened 2018-06-21
[Misc] Create a component to get debug information

atallahade:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/debug/DebugConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+    private ConfigurationSource properties;
+
+    @Inject
+    private Container container;
+
+    /**
+     * @return true if resources should be minified when possible
+     * @since 10.6RC1
+     */
+    public boolean isMinify() {
```
vmassol: hmm I'm wondering if this should go in CoreConfiguration instead (which is about the config for oldcore).

vmassol: However seems that `debug.minify` was not set a core configuration property which is maybe an error we did. @tmortagne any take on that?

tmortagne: It's not really oldcore configuration. The helper to access it is currently internal but I think it will go in a debug module one day.

PR #758, opened 2018-06-26
XWIKI-15389: Improve the icon API for REST API usage

atallahade:
* Add a render custom method in the icon API and icon script service.
* Add a method to retreive the icon set used for an icon.
* Add a method to pull icon resources in the icon API and icon script
  service.
* Write related tests.

See https://jira.xwiki.org/browse/XWIKI-15389

### DIFF
--- a/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconRenderer.java
+++ b/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconRenderer.java
@@ -41,20 +41,11 @@ public interface IconRenderer
     String render(String iconName, IconSet iconSet) throws IconException;
 
     /**
-     * Generate the HTML code to display an icon.
-     *
-     * @param iconName name of the icon to render
-     * @param iconSet icon set that contains the icon to display
-     * @return the HTML code that displays the icon or an empty string if the icon does not exist
-     * @throws IconException if problems occur
-     */
-    String renderHTML(String iconName, IconSet iconSet) throws IconException;
-
-    /**
      * Render an icon with the specified renderer.
      *
      * <p>The renderer contains velocity code which should use the <b>$icon</b> variable.
-     * For instance: <pre>String renderer = "fa fa-$icon";</pre>
+     * For instance:
+     * <pre>String renderer = "fa fa-$icon";</pre>
      *
      * @param iconName name of the icon to render
      * @param iconSet icon set that contains the icon to display
@@ -63,10 +54,20 @@ public interface IconRenderer
      * @throws IconException if problems occur
      * @since 10.6RC1
      */
-    default String renderIcon(String iconName, IconSet iconSet, String renderer) throws IconException
+    default String render(String iconName, IconSet iconSet, String renderer) throws IconException
     {
         throw new UnsupportedOperationException();
     }
+
+    /**
+     * Generate the HTML code to display an icon.
+     *
+     * @param iconName name of the icon to render
+     * @param iconSet icon set that contains the icon to display
+     * @return the HTML code that displays the icon or an empty string if the icon does not exist
+     * @throws IconException if problems occur
+     */
+    String renderHTML(String iconName, IconSet iconSet) throws IconException;
 
     /**
      * Pull the necessary resources to use the specified icon set.
--- a/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconRenderer.java
+++ b/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconRenderer.java
@@ -24,14 +24,15 @@ import org.xwiki.component.annotation.Role;
 /**
  * Component to render an icon (either with Wiki Syntax or HTML).
  *
- * @since 6.2M1
  * @version $Id$
+ * @since 6.2M1
  */
 @Role
 public interface IconRenderer
 {
     /**
      * Generate the wiki syntax to display an icon.
+     *
      * @param iconName name of the icon to render
      * @param iconSet icon set that contains the icon to display
      * @return the wiki syntax that displays the icon or an empty string if the icon does not exist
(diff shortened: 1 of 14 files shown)


mflorea:
I find the new ``renderCustom`` method very confusing. I thought you wanted to add some API to get icon meta data (like URL or CSS class name) which removes the need for a new render method. Moreover, the new ``getIconSet`` methods on the ``IconManager`` interface don't make sense to me.

atallahade:
@mflorea:
When rendering an icon from the `IconManager`, you have no way to know which icon set is being used. Having a `getIconSet` can be useful to obtain the type of an icon (image or class).

I initially created a `getMetadata` method but the issue is that you would have to render both the icon for the wiki syntax and the html syntax which is not very efficient if you only need one. You would also have to remove them if it had to be written in a REST API response.

The `renderCustom` gives the user the choice of implementation. For a CSS icon, you'll get for instance `fa fa-page` and for an image icon you'll have a URL.

mflorea:
> Having a getIconSet can be useful to obtain the type of an icon (image or class).

Yes, the icon type is an attribute of the icon set (unless we want to have icon sets with mixed icon types, which is possible but I doubt we need it), but we already have a way to retrieve an icon set: ``IconSetManager#getIconSet(iconSetName)``. What doesn't make sense to me is having a method to retrieve an icon set given an icon name.

IMO having ``IconManager#getMetaData(iconName)`` is enough since the icon set name should be part of the meta data. So having the icon meta data you can get the icon set name and then get the icon set to see its type (font or image).

> I initially created a getMetadata method but the issue is that you would have to render both the icon for the wiki syntax and the html syntax which is not very efficient if you only need one.

I don't understand why you have to render the icon as wiki or HTML when retrieving the meta data. Isn't it enough to add some new properties to the icon set, e.g.:

xwiki.iconset.icon.url=$xwiki.getSkinFile("icons/silk/${icon}.png")
xwiki.iconset.icon.cssClassName=fa fa-$icon

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconSet.java
```diff
@@ -171,13 +182,54 @@ public String getRenderHTML()
...
     public void setRenderHTML(String renderHTML)
     {
         this.renderHTML = renderHTML;
     }
 
+    /**
+     * @return the icon url
+     * @since 10.6RC1
+     */
+    public String getUrl()
```
mflorea: ``iconSet.getURL()`` makes me think it returns the **icon set URL** not the **icon URL**. I would use a more explicit name, such as ``iconSet.getIconURL()``.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconSet.java
```diff
@@ -171,13 +182,54 @@ public String getRenderHTML()
...
+    public void setUrl(String url)
+    {
+        this.url = url;
+    }
+
+    /**
+     * @return the icon css class
+     * @since 10.6RC1
+     */
+    public String getCssClass()
```
mflorea: Same comment as for ``getURL()``.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-default/src/main/java/org/xwiki/icon/internal/DefaultIconManager.java
```diff
@@ -91,6 +93,33 @@ public String renderHTML(String iconName, String iconSetName, boolean fallback)
         return iconRenderer.renderHTML(iconName, iconSet);
     }
 
+    @Override
+    public Map<String, Object> getMetaData(String iconName) throws IconException
+    {
+        return this.getMetaData(iconName, getIconSet(iconName).getName());
```
mflorea: This means ``getIconSet`` will be called twice for the most common use case. I would add a private ``getMetaData(String iconName, IconSet iconSet)`` and call it from the rest of the ``getMetaData()`` methods.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconRenderer.java
```diff
@@ -41,10 +42,41 @@
...
+     * For instance: <pre>String renderer = "fa fa-$icon";</pre>
+     *
+     * @param iconName name of the icon to render
+     * @param iconSet icon set that contains the icon to display
+     * @param renderer velocity code to render
+     * @return rendered icon using the renderer through velocity
+     * @throws IconException if problems occur
+     * @since 10.6RC1
+     */
+    default String renderIcon(String iconName, IconSet iconSet, String renderer) throws IconException
```
mflorea: I don't like that much the name of this method but I don't have a better name ATM. Also, I feel that we're exposing some internal method, but I guess we have to.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-script/src/main/java/org/xwiki/icon/IconManagerScriptService.java
```diff
@@ -154,8 +164,43 @@ public String renderHTML(String iconName, String iconSetName, boolean fallback)
...
+     *
+     * @param iconSetName name of the icon set
+     * @since 10.6RC1
+     */
+    public void use(String iconSetName)
+    {
+        try {
+            IconSet iconSet = iconSetManager.getIconSet(iconSetName);
+            if (iconSet == null) {
+                iconSet = iconSetManager.getDefaultIconSet();
```
mflorea: I don't think we should do a fallback here.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-api/src/main/java/org/xwiki/icon/IconRenderer.java
atallahade: @mflorea I agree that we should be careful when making internal methods public. I just wanted to avoid making a copy-paste of the method inside the `getMetaData` method. Moreover we already have the `render` and `renderHTML` methods that behave the same.

We can also name this method just `render(String iconName, IconSet iconSet, String renderer)` WDYT? That way we will have `render(String iconName, IconSet iconSet)` and `renderHTML(String iconName, IconSet iconSet)` just calling `render(String iconName, IconSet iconSet, String renderer)`.

PR #761, opened 2018-06-28
XWIKI-15393: Cannot style specific Panels because of limited HTML cla…

atallahade:
…ss attribute generated

See https://jira.xwiki.org/browse/XWIKI-15393

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
@@ -768,10 +768,7 @@ $fn
 #macro(panelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = '')
-#if ($paneldoc)
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
-#end
+#set ($specialClassAttribute = $util.convertToAlphaNumeric($!paneldoc))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
 <div class="panel expanded $!specialClassAttribute">
 <h1 class="xwikipaneltitle">$title</h1>
@@ -786,10 +783,7 @@ $fn
 #macro(panelhiddenheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = '')
-#if ($paneldoc && $paneldoc != '')
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
-#end
+#set ($specialClassAttribute = $util.convertToAlphaNumeric($!paneldoc))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
 <div class="panel expanded $!specialClassAttribute">
 <h1 class="xwikipaneltitle hidden">$title</h1>
@@ -804,10 +798,7 @@ $fn
 #macro(largepanelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = '')
-#if ($paneldoc && $paneldoc != '')
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
-#end
+#set ($specialClassAttribute = $util.convertToAlphaNumeric($!paneldoc))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
 <div class="large panel $expanded $!specialClassAttribute">
 <h1 class="xwikipaneltitle">$title</h1>
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
@@ -768,9 +768,9 @@ $fn
 #macro(panelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = $stringtool.convertToAlphaNumeric("$!paneldoc.documentReference"))
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.documentReference"))
 ## We keep this class for compatibility purpose but the previous one should be used instead.
-#set ($deprecatedClassAttribute = $stringtool.convertToAlphaNumeric("$!paneldoc.name"))
+#set ($deprecatedClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.name"))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
 <div class="panel expanded $!specialClassAttribute $!deprecatedClassAttribute">
 <h1 class="xwikipaneltitle">$title</h1>
@@ -785,9 +785,9 @@ $fn
 #macro(panelhiddenheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = $stringtool.convertToAlphaNumeric("$!paneldoc.documentReference"))
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.documentReference"))
 ## We keep this class for compatibility purpose but the previous one should be used instead.
-#set ($deprecatedClassAttribute = $stringtool.convertToAlphaNumeric("$!paneldoc.name"))
+#set ($deprecatedClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.name"))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
 <div class="panel expanded $!specialClassAttribute $!deprecatedClassAttribute">
 <h1 class="xwikipaneltitle hidden">$title</h1>
@@ -802,9 +802,9 @@ $fn
 #macro(largepanelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = $stringtool.convertToAlphaNumeric("$!paneldoc.documentReference"))
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.documentReference"))
 ## We keep this class for compatibility purpose but the previous one should be used instead.
-#set ($deprecatedClassAttribute = $stringtool.convertToAlphaNumeric("$!paneldoc.name"))
+#set ($deprecatedClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.name"))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
 <div class="large panel $expanded $!specialClassAttribute $!deprecatedClassAttribute">
 <h1 class="xwikipaneltitle">$title</h1>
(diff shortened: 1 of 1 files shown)


atallahade:
We can also keep the previous class to ensure the compatibility doesn't break:

image

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -768,10 +768,7 @@ $fn
 #macro(panelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = '')
-#if ($paneldoc)
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
-#end
+#set ($specialClassAttribute = $util.convertToAlphaNumeric($!paneldoc))
```
vmassol: * if paneldoc is not defined you'll get warnings in the console, why did you remove the "if"?
* you should use stringutils to convert to alpha and not the depredated old $util
* $paneldoc is a Document object. It's not a good idea to rely on the toString() method which can be changed at any time. You should get the documentReference and serialize it.
* you should keep backward compat too

atallahade: > if paneldoc is not defined you'll get warnings in the console, why did you remove the "if"?

Doesn't using `$!paneldoc` instead of `$!paneldoc` make it safe?
http://velocity.apache.org/engine/devel/user-guide.html#quiet-reference-notation

mflorea: > Doesn't using $!paneldoc instead of $paneldoc make it safe?

Only for printing AFAIK, i.e. when you use ``$!paneldoc`` inside a string. Here you're passing the value to a Java method so I don't think it applies, unless you do:

    $util.convertToAlphaNumeric("$!paneldoc"))

atallahade: @mflorea  You are right, null is passed to the method when no value was found. What you suggest should be just fine.

vmassol: Note: Regarding convertToAlphaNumeric if there's no simple api that exists in one of the velocity tool we provide, we could introduce our own StringTool that extends StringUtils for ex.

vmassol: Cool thanks @mflorea for catching the `$!paneldoc` issue.

vmassol: It's actually `org.xwiki.text.StringUtils`.

tmortagne: Sure it's name it's not "StringTool" but it exactly follow your definition.

tmortagne: Sure. I was just answering to your suggestion to introduce something that already exist :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -768,16 +768,15 @@ $fn
 #macro(panelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = '')
-#if ($paneldoc)
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
-#end
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.documentReference"))
```
vmassol: I'm wondering if we should strip any trailing `WebHome` or not. In the future page API we won't have `WebHome` in the page reference and in the future the direction is not display `WebHome` so since this is the equivalent of an API maybe we shouldn't have `WebHome` in the class name?

vmassol: @mflorea @evalica @tmortagne WDYT?

atallahade: I agree with stripping the `WebHome`.

vmassol: For the serialization you should use instead: `$!services.model.serialize($paneldoc.documentReference)`

evalica: is there a way to verify if the page is terminal or not? if it's terminal use the old / backward compatible class created with name and if it's not terminal, use the documentReference?

Anyway, in CSS will still be hard to style documentReference since they will contain dots '.' and those will need escaping. Hope the reference will not contain ':'

vmassol: > is there a way to verify if the page is terminal or not? if it's terminal use the old / backward compatible class created with name and if it's not terminal, use the documentReference?

Except that the old name is not good enough since you can have several panel pages with the same name in different spaces.

> Anyway, in CSS will still be hard to style documentReference since they will contain dots '.' and those will need escaping. Hope the reference will not contain ':'

We only keep alpha numeric chars.

atallahade: The `$stringtool.toAlphaNumeric` method removes the dots. To be more precise it will remove accents and keep only letters and digits.

vmassol: Actually we should not serialize the wiki name IMO. So we should use a `compactwiki` serializer, i.e. `$!services.model.serialize($paneldoc.documentReference, 'compactWiki')`

mflorea: +1 to use the last space **name** if the panel page is non-terminal, and the page **name** if the panel page is terminal. I know that technically we can have panels with the same (page) name at different locations but this is an edge case and I prefer not to have the panel location (full path) in the class name so that moving the panel doesn't require to update the CSS.

+0.5 to serialize (using compact wiki) the last space **reference** if the panel page is non-terminal, and the page **reference** if the panel page is terminal.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -768,16 +768,21 @@ $fn
...
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
+#set ($defaultDocumentName = $services.model.getEntityReference('DOCUMENT', 'default').name)
+#if ($paneldoc.documentReference.name == $defaultDocumentName)
+  #set ($serializedDocument = $!services.model.serialize($paneldoc.documentReference.parent, 'compactwiki'))
+#else
+  #set ($serializedDocument = $!services.model.serialize($paneldoc.documentReference, 'compactwiki'))
 #end
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!serializedDocument"))
+## We keep this class for compatibility purpose but the previous one should be used instead.
+#set ($deprecatedClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.name"))
```
mflorea: I would move the above lines in a macro to avoid the duplication. Also, if ``toAlphaNumeric`` handles ``null`` nicely (as it should, by returning null) then you probably don't need the silent notation (as long as you use the silent notation below).

atallahade: Sure it can be refactored.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -768,16 +768,21 @@ $fn
 #macro(panelheader $title)
 ## Note: We pass the Panel name as an HTML class attribute so that it's possible to style
 ## the Panels selectively using CSS.
-#set ($specialClassAttribute = '')
-#if ($paneldoc)
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
+#set ($defaultDocumentName = $services.model.getEntityReference('DOCUMENT', 'default').name)
+#if ($paneldoc.documentReference.name == $defaultDocumentName)
```
vmassol: Missing some doc here to explain why we do that. 

For example: "We want to avoid showing the "WebHome" suffix in the generated class since we're working to remove it from the UI and from the APIs. In addition it generates class names longer than they are needed. Thus for nested pages we strip it".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -768,16 +768,21 @@ $fn
...
+  #set ($serializedDocument = $!services.model.serialize($paneldoc.documentReference.parent, 'compactwiki'))
+#else
+  #set ($serializedDocument = $!services.model.serialize($paneldoc.documentReference, 'compactwiki'))
```
vmassol: note that the bangs are not really needed here but ok.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
vmassol: Maybe also explain we don't want to display the wiki prefix to have a shorter generated class name

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -768,16 +768,21 @@ $fn
...
 #end
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!serializedDocument"))
```
vmassol: Maybe explain that CSS class name can only contain alphanumeric characters and provide and example, e.g. "xwiki:A.B.WebHome" will generate "AB".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -786,16 +791,21 @@ $fn
...
-#if ($paneldoc && $paneldoc != '')
-  #set ($specialClassAttribute = $util.convertToAlphaNumeric($paneldoc.name))
+#set ($defaultDocumentName = $services.model.getEntityReference('DOCUMENT', 'default').name)
+#if ($paneldoc.documentReference.name == $defaultDocumentName)
+  #set ($serializedDocument = $!services.model.serialize($paneldoc.documentReference.parent, 'compactwiki'))
+#else
+  #set ($serializedDocument = $!services.model.serialize($paneldoc.documentReference, 'compactwiki'))
 #end
+#set ($specialClassAttribute = $stringtool.toAlphaNumeric("$!serializedDocument"))
+## We keep this class for compatibility purpose but the previous one should be used instead.
```
vmassol: This loos like a duplicate of what's above. I'd suggest introducing an internal velocity macro for that.

vmassol: Or even a `#define` since the duplication seems to go deep even in the generated wiki syntax.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -804,16 +814,21 @@ $fn
...
+#set ($deprecatedClassAttribute = $stringtool.toAlphaNumeric("$!paneldoc.name"))
 #if ($paneldoc.syntax.toIdString() == 'xwiki/1.0')
-<div class="large panel $expanded $!specialClassAttribute">
+<div class="large panel $expanded $!specialClassAttribute $!deprecatedClassAttribute">
 <h1 class="xwikipaneltitle">$title</h1>
 <div class="xwikipanelcontents">
 #else
-(% class="large panel expanded $!specialClassAttribute" %)(((
+(% class="large panel expanded $!specialClassAttribute $!deprecatedClassAttribute" %)(((
```
vmassol: Duplication again here.

vmassol:
Globally looks good. Missing some explanations in comments + remove duplication and we should be good. Thanks Adel.

atallahade:
@vmassol, @mflorea: The changes have been made.

PR #767, opened 2018-07-05
The fields from "Job name" and "Job cron expression" have not the same size[XWIKI-11277]

CostiChD:


xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -65,7 +65,7 @@ $job.getProperty('script').value
 #end
 
 #if($xcontext.action=="edit")
-{{info}}$services.localization.render('xe.scheduler.job.scriptexplanation'){{/info}}
+{{info}}$services.localization.render('xe.scheduler.job.scriptexplanation')){{/info}}
```
acotiuga: Here's an extra `)`.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -20,19 +20,19 @@
...
-<xwikidoc version="1.1">
+<xwikidoc version="1.3" reference="XWiki.SchedulerJobSheet" locale="">
   <web>XWiki</web>
   <name>SchedulerJobSheet</name>
   <language/>
   <defaultLanguage/>
   <translation>0</translation>
   <creator>xwiki:XWiki.Admin</creator>
-  <creationDate>1252454400000</creationDate>
+  <creationDate>1252450800000</creationDate>
```
evalica: Also I would prefer we skip the dates changes

evalica:
You still need to add the <label> tags in order for the styling to apply. The labels should be capitalized. 
Careful on what values you select for the "for" attribute. We should not break HTML validation.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -40,49 +40,49 @@
...
+  #set ($job = $doc.getObject("XWiki.SchedulerJobClass"))
+  #macro (displayJobProperty $obj $propName)
+    #set ($class = $obj.xWikiClass)
+    #set ($prettyName = $class.get($propName).translatedPrettyName)
+    ; #if ($xcontext.action == 'edit')
+      {{html}}&lt;label for="${class.getName()}_${obj.number}_${propName}"&gt;$prettyName&lt;/label&gt;{{/html}}
+    #else
+      $prettyName
+    #end
+    : $doc.display($propName, $obj)
```
vmassol: maybe a `#define` could be enough? See http://velocity.apache.org/engine/1.7/user-guide.html#define

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -20,67 +20,69 @@
...
   <content>{{velocity}}
-#if($doc.getObject('XWiki.SchedulerJobClass'))
-#set($job = $doc.getObject("XWiki.SchedulerJobClass"))
-**$services.localization.render('xe.scheduler.job.name')**
+#if ($doc.getObject('XWiki.SchedulerJobClass'))
+  #set ($job = $doc.getObject("XWiki.SchedulerJobClass"))
+  #macro (displayJobProperty $obj $propName)
+    #set ($class = $obj.xWikiClass)
+    #set ($prettyName = $class.get($propName).translatedPrettyName)
+    ; #if ($xcontext.action == 'edit')
```
evalica: Not sure why you are putting `xform` only on edit mode. For example, if you look at an User Profile, the labels styling is present also in View mode. 
So I would remove the `#if / #else`.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -20,67 +20,69 @@
...
-| **0 15 10 ? * MON-FRI** | Fire at 10:15am every Monday, Tuesday, Wednesday, Thursday and Friday
-| ** 0 0/5 14 * * ?** | Fire every 5 minutes starting at 2pm and ending at 2:55pm, every day
-| ** 0 15 10 ? * 6L** | Fire at 10:15am on the last Friday of every month
-{{/info}}
-
-#elseif($xcontext.action=="view")
-[[$services.localization.render('xe.scheduler.job.backtolist')&gt;&gt;Scheduler.WebHome]]
-#end
+  #elseif ($xcontext.action == 'view')
+    [[$services.localization.render('xe.scheduler.job.backtolist')&gt;&gt;Scheduler.WebHome]]
```
evalica: Not related necessarily to this PR, but I consider the back link "Back to the job list" to be deprecated and inconsistent with the other applications we have across XWiki.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
Enygma2002: @vmassol Not sure what you mean by that. Are you suggesting to use a #define instead of a #macro? Could you please elaborate?

vmassol: > Are you suggesting to use a #define instead of a #macro? 

Yes that was the idea. You'll need to decide what's best (macro or define). I haven't really looked in details, was just doing a suggestion since from the outside it seemed it was about computing something to display.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,63 @@
...
+#if ($doc.getObject('XWiki.SchedulerJobClass'))
+  #set ($job = $doc.getObject("XWiki.SchedulerJobClass"))
+  {{html wiki='true'}}
+  &lt;div class = 'xform'&gt;
+    &lt;dl&gt;
+      #if ($job)
+        #displayJobProperty($job 'jobName')
+        #displayJobProperty($job 'jobDescription')
+        #displayJobProperty($job 'cron')
+      #end
```
evalica: Should this #if close here?

evalica:
We should wait for release and merge for 10.6.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
CostiChD: Actually, I just realised that we don't need it at all because it is already checked here:
`#if ($doc.getObject('XWiki.SchedulerJobClass'))
   #set ($job = $doc.getObject("XWiki.SchedulerJobClass"))`

acotiuga: Good catch! Could you update the PR?

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,61 @@
...
+            $services.localization.render('xe.scheduler.job.scriptexplanation')
+          &lt;/span&gt;
+      &lt;/dt&gt;
+      &lt;dd&gt;
+        #if ($xcontext.action == 'edit')
+          $!doc.display('script')
+        #else
+          ((({{code language='groovy'}}$!job.getValue('script'){{/code}})))
+        #end
+      &lt;/dd&gt;
```
Enygma2002: I do not like the code duplication done above. Instead, you should handle the special property `script` inside the `displayJobProperty` macro.

You should try to avoid duplication as much as possible, specially when it's about handling implementation details in more than one places.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,61 @@
...
+          &lt;label for="${class.getName()}_0_script"&gt;$prettyName&lt;/label&gt;
+          &lt;span class="xHint"&gt;
+            $services.localization.render('xe.scheduler.job.scriptexplanation')
+          &lt;/span&gt;
+      &lt;/dt&gt;
+      &lt;dd&gt;
+        #if ($xcontext.action == 'edit')
+          $!doc.display('script')
+        #else
+          ((({{code language='groovy'}}$!job.getValue('script'){{/code}})))
```
Enygma2002: It might look nicer here if instead of `(((`,`)))` you add an empty line before and after the `code` macro.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,58 @@
...
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content>{{velocity}}
-#if($doc.getObject('XWiki.SchedulerJobClass'))
-#set($job = $doc.getObject("XWiki.SchedulerJobClass"))
-**$services.localization.render('xe.scheduler.job.name')**
+  <content>{{velocity output='false'}}
+#macro (displayJobProperty $obj $propName)
+  #set ($class = $obj.xWikiClass)
+  #set ($prettyName = $class.get($propName).translatedPrettyName)
```
evalica: It's good that we are using the translatedPrettyName, but this means the old translations key are not used anymore, like xe.scheduler.job.name, xe.scheduler.job.description, xe.scheduler.job.expression, xe.scheduler.job.script. 

If they are not used anymore (needs to be tested: do a grep inside platform and see if they are used somewhere else) we could deprecate them.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,58 @@
...
+      &lt;span class='xHint'&gt;
+        $services.localization.render('xe.scheduler.job.scriptexplanation')
+      &lt;/span&gt;
+    #end
+    &lt;/dt&gt;
+    &lt;dd&gt;
+    #if ($propName == 'script' &amp;&amp; $xcontext.action == 'view')
 
-$!doc.display('jobName')
+      {{code language='groovy'}}$!job.getValue($propname){{/code}}
```
evalica: There was a verification if the script value was empty. Now this is gone and an empty box/code macro is displayed.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,58 @@
...
   <hidden>true</hidden>
-  <content>{{velocity}}
-#if($doc.getObject('XWiki.SchedulerJobClass'))
-#set($job = $doc.getObject("XWiki.SchedulerJobClass"))
-**$services.localization.render('xe.scheduler.job.name')**
+  <content>{{velocity output='false'}}
+#macro (displayJobProperty $obj $propName)
+  #set ($class = $obj.xWikiClass)
+  #set ($prettyName = $class.get($propName).translatedPrettyName)
+  &lt;dt&gt;
```
Enygma2002: It does seem that you have manually added `&lt;` and `&gt;` in order to escape `<` and `>`. This also means that you have worked on the content of the page with a text editor or directly pasted into the XML file of the page, which is something you must avoid in the future. The problem is that, while you have manually fixed the `<` character, there can be other cases inside the document's content that need escaping and it is not good for you to do them manually. Instead, you should always work on the wiki page, in the browser, save and export the page before committing it to git. This way, any needed escapes will be performed by XWiki and you will rest assured that you're not missing anything.

A good example here is the fact that the `>` character does not need escaping (`&gt;`). As a result, once someone will export this page and try to commit, they will see that XWiki did not escape the `>` character and their commit will generate changes beyond the expected ones.

Thus, to be sure, you need to do a final export of this page from your wiki and commit the changes it will produce (i.e. at least the unescaping of `&gt;` back to `>`).

atallahade: Would be great to have the `mvn xar:format` doing the same export as XWiki.

Enygma2002: Actually, I don't see how it could do that, since XWiki is the only one that has the content in its original version and could be able of doing the proper exporting. `xar:format` gets as input an xml document on which is hard to tell if a character should be escaped or not, depending on the author's intention and if XWiki already escaped it or not.

It would seem that `xar:format` is already doing some escaping, but in an old fashion, not synchronized with what XWiki is doing. So XWiki is not escaping `>` but `xar:format` is... which is wrong.

CostiChD: I actually pasted some rows into the XML file of the page(my mistake), but that happened 4 commits ago. Since then I used only the file exported from the wiki.

xwiki-platform-core/xwiki-platform-scheduler/xwiki-platform-scheduler-ui/src/main/resources/XWiki/SchedulerJobSheet.xml
```diff
@@ -39,48 +39,58 @@
...
 
-#elseif($xcontext.action=="view")
-[[$services.localization.render('xe.scheduler.job.backtolist')&gt;&gt;Scheduler.WebHome]]
-#end
+    |= Cron expression |= Meaning
+    | **0 15 10 * * ?** | Fire at 10:15am every day
+    | **0 15 10 ? * MON-FRI** | Fire at 10:15am every Monday, Tuesday, Wednesday, Thursday and Friday
+    | ** 0 0/5 14 * * ?** | Fire every 5 minutes starting at 2pm and ending at 2:55pm, every day
+    | ** 0 15 10 ? * 6L** | Fire at 10:15am on the last Friday of every month
+    {{/info}}
```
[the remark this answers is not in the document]
CostiChD: Should I create an issue which contains both(link as xhint + xform adapted to ue xhints) ?

CostiChD:
I created a new issue for the untranslated info box here: https://jira.xwiki.org/browse/XWIKI-15446?filter=-2 
Also, another issue could be to adapt the xform to use as xhints the property hints defined in the class.

Enygma2002:
Thanks for your contribution and patience! :)

PR #768, opened 2018-07-05
XWIKI-15406: Parsed LESS code in SSX objects always produces the same…

atallahade:
… CSS, even with 'no caching' set

* Disable LESS caching for SSX when the cache policy is set to "no caching".

See https://jira.xwiki.org/browse/XWIKI-15406.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/web/sx/SxDocumentSource.java
+++ b/xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/web/sx/SxDocumentSource.java
@@ -138,7 +138,8 @@ public class SxDocumentSource implements SxSource
                     LESSResourceReference lessResourceReference =
                         lessResourceReferenceFactory.createReferenceForXObjectProperty(objectPropertyReference);
                     try {
-                        sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1), false);
+                        // We don't use cache here because it should be managed by the browser using cache policy.
+                        sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1), true);
                     } catch (LESSCompilerException e) {
                         // Set the error message in a CSS comment to help the developer understand why its SSX is not
                         // working (it will work only if the CSS minifier is not used).
--- a/xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/web/sx/SxDocumentSource.java
+++ b/xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/web/sx/SxDocumentSource.java
@@ -139,7 +139,7 @@ public class SxDocumentSource implements SxSource
                         lessResourceReferenceFactory.createReferenceForXObjectProperty(objectPropertyReference);
                     try {
                         sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1),
-                            CachePolicy.FORBID.equals(getCachePolicy()));
+                            CachePolicy.FORBID == getCachePolicy());
                     } catch (LESSCompilerException e) {
                         // Set the error message in a CSS comment to help the developer understand why its SSX is not
                         // working (it will work only if the CSS minifier is not used).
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/web/sx/SxDocumentSource.java
```diff
@@ -138,7 +138,8 @@ public String getContent()
                     LESSResourceReference lessResourceReference =
                         lessResourceReferenceFactory.createReferenceForXObjectProperty(objectPropertyReference);
                     try {
-                        sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1), false);
+                        // We don't use cache here because it should be managed by the browser using cache policy.
+                        sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1), true);
```
gdelhumeau: So you never use the LESS cache? Note that the LESS compiler is very CPU-consuming. If you never cache, LESS compiler will be used for each individual request. IMO, you will just make the wiki crash all the time.

See: https://jira.xwiki.org/browse/XWIKI-15406?focusedCommentId=99267&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-99267

gdelhumeau: > The LESS compiler will not always be used if you enable the SSX cache as the browser will 
> not make the request (if it has previously cached the page).

If thousands of new users (with empty cache) come in the same time, you will have trouble. What I would do is to disable the LESS cache when the developer of the SSX set "no cache" at the browser policy, but still use it otherwise. This way, it is the responsibility of the developer to set the correct cache value.

> Also why would the wiki crash if the LESS cache is disable?

Having too much load on a server is a good way to make it crash.

>  I guess it depends on how many requests are being made.

Definitively.
> What I can do is to disable velocity compiled code being cached and instead compile the 
> velocity code each time the request is made instead of the whole LESS + velocity.

I don't think you can do that because velocity is executed before LESS.

atallahade: > What I would do is to disable the LESS cache when the developer of the SSX set "no cache" at the browser policy, but still use it otherwise

Done, indeed you suggestion is better as LESS cache will be used at least when the cache policy is not set to "no caching".

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/web/sx/SxDocumentSource.java
```diff
@@ -138,7 +138,8 @@ public String getContent()
                     LESSResourceReference lessResourceReference =
                         lessResourceReferenceFactory.createReferenceForXObjectProperty(objectPropertyReference);
                     try {
-                        sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1), false);
+                        sxContent = lessCompiler.compile(lessResourceReference, true, (parse == 1),
+                            CachePolicy.FORBID.equals(getCachePolicy()));
```
tmortagne: It's an enum so you can use `==`.

tmortagne: That too :)

atallahade: > BTW, should it be CachePolicy.FORBID != getCachePolicy() ?

Nope, the last argument is:
> * @param force force the computation, even if the output is already in the cache (not recommended)

So we want to force the computation only when `CachePolicy.FORBID == getCachePolicy()`

tmortagne: Right, I looked at the wrong diff and tough `false` was your previous version :)

So indeed it should be false unless we don't want to cache

PR #769, opened 2018-07-05
XWIKI-15381: Simple users don't have right to remove an annotation (t…

atallahade:
…hey created) from the Comments section

* Fix wrong comparison between user comment author and current user.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
@@ -141,6 +141,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
   #if ("$!commentSelection" != "")
     #set($isAnnotation = true)
   #end
+  #set ($isUserComment = $services.model.resolveDocument($comment.author, 'current') == $xcontext.userReference)
   <div id="xwikicomment_${comment.number}" class="xwikicomment  #if($comment.getProperty('author').value == $doc.creator) commentByCreator#end#if($isAnnotation) annotation#end">
     <div class="commentavatar">#if("$!comment.replyto" == '')#largeUserAvatar($comment.author)#{else}#mediumUserAvatar($comment.author)#end</div>
     <div class="commentheader">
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -141,6 +141,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
   #if ("$!commentSelection" != "")
     #set($isAnnotation = true)
   #end
+  #set ($isUserComment = $services.model.resolveDocument($comment.author, 'current') == $xcontext.userReference)
```
mflorea: I think the 'user' (hint) resolver would be more appropriate in this case.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -153,12 +154,12 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
...
-        #if($services.model.resolveDocument($comment.author, 'current') == $xcontext.userReference || $hasAdmin)
+        #if($hasAdmin || $isUserComment)
           <a class="edit btn btn-default btn-xs" rel="nofollow" href="$doc.getURL('view', "viewer=comments&amp;number=${comment.number}&amp;xredirect=$doc.getURL('view')")" title="$services.localization.render('core.viewers.comments.edit')">$services.icon.renderHTML('pencil')</a>
         #end
       #end
       <a class="permalink btn btn-default btn-xs" rel="nofollow" href="$doc.getURL('view', 'viewer=comments')#xwikicomment_${comment.number}" title="$services.localization.render('core.viewers.comments.permalink')">$services.icon.renderHTML('link')</a>
-      #if ($hasAdmin || (!$isGuest && $hasEdit && $comment.author == $xcontext.user))
+      #if ($hasAdmin || ($hasEdit && $isUserComment))
```
mflorea: You dropped the ``!$isGuest`` condition, probably thinking that if a guest user can add a comment/annotation then he should be able to remove it, but then a guest user will be able to delete the comments/annotations of other guests.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
atallahade: I didn't change the existing code but if you think the `user` hint is better then we can try to use it.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
atallahade: I though a guest was a non-logged user. In this case he wouldn't have rights to make annotations and the `$isUserComment` would always be `false`.

atallahade: Well I guess it also depends on the rights that have been setup.

mflorea: AFAIK you can give edit right to guest users. For instance @vmassol allows guest comments on [his blog](http://massol.myxwiki.org/xwiki/bin/view/Blog/).

atallahade: Tried it and the `$isUserComment` is false because the `$comment.author` is empty and $xcontext.userReference`  is null.

mflorea: Which means you're not resolving the comment author correctly. If the comment was added by a guest then the resolved user reference should be null (same as ``$xcontext.userReference``).

atallahade: Right, this behavior has been here since a while (for the remove button on the annotation) but I guess we can fix it now.

PR #771, opened 2018-07-09
XWIKI-15434: Create a displayer for the Page field that supports autocomplete

atallahade:
## Changes
* DefaultPageQueryBuilder, ImplicitlyAllowedValuesPageQueryBuilder
* Add support to nested maps serialization in MapAdapter.
* Improve AbstractDocumentListClassPropertyValuesProvider
  * Add icon and hint metadata to the property values.
  * Make the getIcon method return a map with icon related information.
  * Replace the document name with the parent space name when the
   document is WebHome in the getLabel method.
* Create a Page class property values provider.
* Display icons in the selectize when an icon metadata is passed.
* Create a displayer for the Page class.
* Add related tests
* Increase jacoco threshold

## Jira
https://jira.xwiki.org/browse/XWIKI-15434

## Example
autocomplete-07

## Could be enhanced
* Search pages with the shown label name (e.g. "Sandbox Test Page 1").
* Support for the `wiki:` prefix / search through sub-wikis.
* Show also the name of the document in the hint?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/MapAdapter.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/MapAdapter.java
@@ -29,7 +29,7 @@ import javax.xml.bind.annotation.adapters.XmlAdapter;
  * output format is XML and as a JSON object when the output format is JSON. This adapter allows us to use
  * {@link java.util.Map} in the REST resource that retrieves the class property values, while the XML serialization will
  * be done using the schema-generated {@link Map}.
- * 
+ *
  * @version $Id$
  * @since 9.8RC1
  */
@@ -42,7 +42,15 @@ public class MapAdapter extends XmlAdapter<Map, java.util.Map<String, java.lang.
             return null;
         } else {
             Map output = new Map();
-            input.forEach((key, value) -> output.getEntries().add(new MapEntry().withKey(key).withValue(value)));
+            for (String key : input.keySet()) {
+                java.lang.Object value = input.get(key);
+                if (value instanceof java.util.Map) {
+                    java.util.Map<String, java.lang.Object> nestedMap = new HashMap<>();
+                    ((java.util.Map<?, ?>) value).forEach((k, v) -> nestedMap.put(k.toString(), v));
+                    value = marshal(nestedMap);
+                }
+                output.getEntries().add(new MapEntry().withKey(key).withValue(value));
+            }
             return output;
         }
     }
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/MapAdapter.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/MapAdapter.java
@@ -42,8 +42,9 @@ public class MapAdapter extends XmlAdapter<Map, java.util.Map<String, java.lang.
             return null;
         } else {
             Map output = new Map();
-            for (String key : input.keySet()) {
-                java.lang.Object value = input.get(key);
+            for (java.util.Map.Entry<String, java.lang.Object> entry : input.entrySet()) {
+                String key = entry.getKey();
+                java.lang.Object value = entry.getValue();
                 if (value instanceof java.util.Map) {
                     java.util.Map<String, java.lang.Object> nestedMap = new HashMap<>();
                     ((java.util.Map<?, ?>) value).forEach((k, v) -> nestedMap.put(k.toString(), v));
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/AbstractDocumentListClassPropertyValuesProvider.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/AbstractDocumentListClassPropertyValuesProvider.java
@@ -19,6 +19,8 @@
  */
 package org.xwiki.rest.internal.resources.classes;
 
+import java.util.Map;
+
 import javax.inject.Inject;
 import javax.inject.Named;
 
@@ -32,13 +34,14 @@ import org.xwiki.rest.model.jaxb.PropertyValue;
 import org.xwiki.rest.model.jaxb.PropertyValues;
 import org.xwiki.rest.resources.classes.ClassPropertyValuesProvider;
 
+import com.xpn.xwiki.XWiki;
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.objects.classes.ListClass;
 
 /**
  * Base class for {@link ClassPropertyValuesProvider} implementations that work with list properties whose values are
  * document references.
- * 
+ *
  * @param <T> the property type
  * @version $Id$
  * @since 10.6RC1
@@ -57,6 +60,17 @@ public abstract class AbstractDocumentListClassPropertyValuesProvider<T extends 
     @Inject
     @Named("viewable")
     private QueryFilter viewableFilter;
+
+    /**
+     * A map with an icon css class / url or with {@link org.xwiki.icon.IconManager#getMetaData icon metadata}.
+     *
+     * <p>The entries of this map are meant to be added to the metadata of the
+     * {@link #getValueFromQueryResult values of the query result}.
+     *
+     * @param documentReference document reference used to generate the icon
+     * @return a map with icon related information
+     */
+    protected abstract Map<String, Object> getIcon(DocumentReference documentReference);
 
     @Override
     protected PropertyValues getUsedValues(T propertyDefinition, int limit, String filter) throws QueryException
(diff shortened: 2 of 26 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/PageClassPropertyValuesProvider.java
```diff
@@ -0,0 +1,157 @@
...
+    protected PropertyValues getAllowedValues(PageClass pageClass, int limit, String filter) throws Exception
+    {
+        // Execute the query with the rights of the class last author because the query may not be safe.
+        return this.authorExecutor.call(() -> getValues(
+            this.allowedValuesQueryBuilder.build(pageClass), limit, filter, pageClass
+        ), pageClass.getOwnerDocument().getAuthorReference());
+    }
+
+    @Override
+    protected PropertyValue getValueFromQueryResult(Object result, PageClass propertyDefinition)
```
mflorea: ``AbstractDocumentListClassPropertyValuesProvider.java`` is already calling ``getHint()`` so I don't think you need to overwrite this method at all.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/UsersClassPropertyValuesProvider.java
```diff
@@ -80,23 +83,27 @@ protected PropertyValues getAllowedValues(UsersClass propertyDefinition, int lim
...
+                icon.put(META_DATA_ICON, xcontext.getWiki().getURL(avatarAttachment.getReference(), "download",
+                    "width=30&height=30&keepAspectRatio=true", null, xcontext));
             }
         } catch (XWikiException e) {
             this.logger.warn(
                 "Failed to read the avatar of user [{}]. Root cause is [{}]. Using the default avatar instead.",
                 userReference.getName(), ExceptionUtils.getRootCauseMessage(e));
         }
+        if (!icon.containsKey(META_DATA_ICON)) {
+            icon.put(META_DATA_ICON, xcontext.getWiki().getSkinFile("icons/xwiki/noavatar.png", true, xcontext));
```
mflorea: We could also try to use the icon theme here, for instance the ``user`` icon.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/AbstractDocumentListClassPropertyValuesProvider.java
```diff
@@ -58,6 +61,17 @@
...
+    /**
+     * A map with an icon css class / url or with {@link org.xwiki.icon.IconManager#getMetaData icon metadata}.
+     *
+     * <p>The entries of this map are meant to be added to the metadata of the
+     * {@link #getValueFromQueryResult values of the query result}.
+     *
+     * @param documentReference document reference used to generate the icon
+     * @return a map with icon related information
+     */
+    protected abstract Map<String, Object> getIcon(DocumentReference documentReference);
```
mflorea: I'm not convinced we should support both types of icon representations. It complicates the code, especially ``xwiki.selectize.js``. I think I prefer to return only ``META_DATA_ICON`` having a map as value. The user and group providers can fill the icon meta data map with the avatar URL and the ``IMAGE`` icon type. WDYT?

mflorea:
The issue mentioned in the PR title (XWIKI-8613) is not right I think. We're not adding the suggest to macro parameter yet (it is the goal but I don't think we're be able to do it for 10.6). We need an issue that is focused on improving the Page xclass property.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/PageClassPropertyValuesProvider.java
atallahade: Indeed, I forgot to remove this code. It's great that we don't need this anymore as it's a bit complex.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/MapAdapter.java
```diff
@@ -42,7 +42,15 @@ public Map marshal(java.util.Map<String, java.lang.Object> input) throws Excepti
             return null;
         } else {
             Map output = new Map();
-            input.forEach((key, value) -> output.getEntries().add(new MapEntry().withKey(key).withValue(value)));
+            for (String key : input.keySet()) {
```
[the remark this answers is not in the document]
atallahade: @mflorea Something like that?
```java
for (java.util.Map.Entry<String,java.lang.Object> entry: input.entrySet()) {
  String key = entry.getKey();
  java.lang.Object value = entry.getValue();
  ...
```
I wanted to use the foreach method with a lambda at first but I would have to handle the exception.
```java
input.forEach((key, value) -> { ... });
```

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/AbstractDocumentListClassPropertyValuesProvider.java
atallahade: If I understand you want to move the old icon key (`META_DATA_ICON`) into the new icon metadata map (`META_DATA_ICON_META_DATA`) so that we only keep the map? It will make `xwiki.selectize.js` simpler indeed but we have to make sure that there are no other files to change.

mflorea:
Regarding what could be enhanced:

> Search pages with the shown label name (e.g. "Sandbox Test Page 1")

Indeed, but note that the label is determined by the Page property configuration. If the class author uses an explicit query or sets the "Value Field" then he controls what the label is that label will be part of the query. The problem is just that if he leaves the "Value Field" empty (which happens by default) then the ``ImplicitlyAllowedValuesPageQueryBuilder#build()`` fall-back (indirectly) on the ``doc.fullName`` (i.e. if the value field is not specified then the id field is used as value field). So the fix is easy: set the value field to ``doc.title`` in ``ImplicitlyAllowedValuesPageQueryBuilder#build()``, similar to what we do for the id field.

> Show also the name of the document in the hint?

No, but this made me realized that I overlooked something. The page name is technical. Everywhere in the UI we show the page title. This is fine because the page title is going to be the same as the page name in 90% of the cases (when simple users create pages they specify the page title and that title is used as page name). The rest of 10% is code / technical pages that use Velocity in the page title (mostly for localization). Now, the thing I overlooked. In ``PageClassPropertyValuesProvider#getHint()`` you use the entity name, which is technical. We need to do the same as the breadcrumb, i.e. show the (rendered) title of each page in the path. So the suggestion should look like:

    Title of the page
    Title of parent 1 / Title of parent 2 / ... / Title of parent N

For this I would add a ``getLabel(EntityReference)`` method in ``PageClassPropertyValuesProvider`` and use it both in ``getLabel(DocumentReference, String)`` and in ``getHint()``. Note that we also need to handle the case where the path in the hint is very long. Some ellipsis with CSS, preferably in the center, would be nice.

> Support for the wiki: prefix / search through sub-wikis.

I would keep things simple for now. Note that there is a wiki id and a wiki pretty name and we show the wiki pretty name in the UI so simple users won't be aware of the wiki id. Moreover, the ``wiki:`` prefix is technical and thus targeted for advanced users. Last but not least, an xclass can be used only in the wiki where it is defined: xclasses are local. So if we're talking about the Page xproperty picker then I think it's ok to keep it local for now. If we'll add support for cross wiki suggestions in the future we'll probably take into account the current wiki's user scope, like the user picker does, so that we don't show suggestions of main wiki pages on a subwiki that is closed to local users. Anyway, let's start simple.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/PageClassPropertyValuesProvider.java
```diff
@@ -93,24 +93,31 @@ protected PropertyValues getAllowedValues(PageClass pageClass, int limit, String
...
-                label = super.getLabel(documentReference, currentLabel);
+        String label;
+        try {
+            XWikiContext xcontext = this.xcontextProvider.get();
+            XWikiDocument document = xcontext.getWiki().getDocument(entityReference, xcontext);
+            label = document.getRenderedTitle(Syntax.PLAIN_1_0, xcontext);
+        } catch (XWikiException e) {
+            this.logger.error("Error while loading the document [{}]. Root cause is [{}]", entityReference,
+                ExceptionUtils.getRootCause(e));
+            if (entityReference instanceof DocumentReference) {
```
mflorea: I think the common practice is, in general, to check

    entityReference.type == EntityType.DOCUMENT

and then use

    new DocumentReference(entityReference)

but what you did will work fine in this context.

PR #772, opened 2018-07-09
[Misc] Avoid making the buttons sticky when they are not displayed

atallahade:


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -384,11 +384,12 @@ require(['jquery'], function ($) {
 
   $(window).on("scroll resize load click xwiki:dom:refresh", function() {
     var isFullScreen = $('.fullScreenWrapper').length > 0
+    var isVisible = $container.css('display') != 'none';
```
mflorea: Why not https://api.jquery.com/visible-selector/ ?

    var isVisible = $container.is(':visible');

mflorea:
Can you give an example where this is currently happening? (i.e. when the action buttons are hidden)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
atallahade: It's even better, thanks.

atallahade:
@mflorea 

When you create / edit an AWM:
sticky-bug

I'm only scrolling down on the video.

PR #775, opened 2018-07-13
XWIKI-11283 : Missing "Actions" column name on the History table

CamiAndrei:
* add the "Action" title for the column name
* simplified the condition on the title display because the administration rights imply the edit rights


mflorea:
Thanks for your contribution.

PR #779, opened 2018-07-17
XWIKI-8587: Add a Document field type that has a picker associated which suggests document names

atallahade:
## Changes

* Add a Page picker on the AWM.
* Change the icons of the AWM field palette.
* Add a new mapping for the Font Awesome header icon.
* Use suggest on Page live table columns.
* Customize the page displayer for view mode.

## Jira

https://jira.xwiki.org/browse/XWIKI-8587

## Screenshots

image

image

image

## Note

Should other modes of the page displayer be changed like the view mode?


xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
```diff
@@ -114,13 +114,18 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
           #set($formFieldDoc = $xwiki.getDocument($formField))
           #set($discard = $formFieldDocs.add($formFieldDoc))
           #set($formFieldIcon = $formFieldDoc.getObject($formFieldClassName).getProperty('icon').value)
-          #if($formFieldIcon.contains('/'))
-            #set($formFieldIconURL = $xwiki.getSkinFile($formFieldIcon))
-          #else
-            #set($formFieldIconURL = $formFieldDoc.getAttachmentURL($formFieldIcon))
+          ## Fallbacks on the default icon set if Silk is not available.
+          #set($formFieldIconRendered = $services.icon.renderHTML($formFieldIcon, 'Silk'))
```
evalica: Why did you forced it on Silk?

atallahade: Because it was nicer with the other icons. Here is how it looks with Font Awesome:

evalica: If you really want them aligned, you can give some min-width from CSS.

atallahade: There is actually the css class `fa-fw` that gives a constant width to every icons and align them at the center. The problem is that we probably shouldn't use it for all the icon sets. Maybe giving just a `min-width` can help but how much would it be? Is it going to work nicely with other icon sets?

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/Pages.xml
```diff
@@ -0,0 +1,169 @@
...
+      <disabled>0</disabled>
+      <displayType>input</displayType>
+      <hint/>
+      <idField/>
+      <multiSelect>0</multiSelect>
+      <name>page</name>
+      <number>1</number>
+      <picker>1</picker>
+      <prettyName>Page</prettyName>
+      <relationalStorage>0</relationalStorage>
```
mflorea: Relational storage is recommended when multiple selection is on, and since we don't expose this meta property the users won't be able to set it when enabling multiple selection. Let's keep relation storage enabled to be ready for multiple selection. It shouldn't have any effect on single selection.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
```diff
@@ -1 +1,23 @@
-#template('displayer_dblist.vm')
+#if ($type == 'edit' || $type == 'search')
+  #if ($field.getProperty('displayType').value == 'input' && $field.getProperty('picker').value == 1)
+    #xpropertySuggestInputDisplayer($field $prefix $name $value)
+  #else
+    $doc.displayEdit($field, $prefix, $object)
+  #end
+#elseif ($type == 'view')
+  #set ($value = $doc.displayView($field, $prefix, $object))
```
mflorea: You already have ``$value``. I don't understand why you need to call ``displayView``.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
```diff
@@ -1 +1,23 @@
...
+  #set ($pageDoc = $xwiki.getDocument($value))
+  #if ("$!pageDoc" == "" || $pageDoc.isNew())
+    $value
```
mflorea: You don't need to get the document to obtain the URL. The value is a document reference and you can get the URL with that so you can show a link to the missing document (using the proper CSS class to show the question mark, see what happens when you create a link to a missing wiki page from wiki syntax).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
```diff
@@ -1 +1,23 @@
...
+  #else
+    <a href="$pageDoc.URL">$pageDoc.getRenderedTitle('plain/1.0')</a>
+  #end
```
mflorea: If you want to cover all cases, it should be:

* no link when the doc is null, which means the current user doesn't have the right to access it
* a link when doc is not null
  * a link to a "missing" doc when the document is new (with proper CSS)
  * a normal link otherwise

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
atallahade: What do you mean I already have `$value`? What I wanted was to use `displayView` only if the method didn't return a page reference (e;g. Main.WebHome). In the other case I wanted to render the title of the page reference.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
atallahade: I also forgot to take care of the case where there are multiple pages.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
atallahade: @mflorea Seems like the `$value` variable is either a String or a List.

mflorea: Yes, you need to handle multiple selection. See the user displayer.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
```diff
@@ -1 +1,40 @@
...
+#end
+
+#macro(displayPageReference $pageReference)
+  #set ($pageDoc = $xwiki.getDocument($pageReference))
+  #if ("$!pageDoc" == "")
+    $pageReference
+  #else
+    #if ($pageDoc.isNew())
+      <span class="wikicreatelink">
+        <a href="$pageDoc.getURL('create')">$pageDoc.getRenderedTitle('plain/1.0')</a>
```
mflorea: ``$pageDoc.plainTitle`` is shorter. In any case, you need to ``$escapetool.xml(...)`` as the plain title can have XML / HTML special characters.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
```diff
@@ -1 +1,40 @@
...
+  #else
+    $doc.displayEdit($field, $prefix, $object)
+  #end
+#elseif ($type == 'view')
+  #set ($pages = [])
+  #foreach($pageReference in $value)
+    #set ($discard = $pages.add("#displayPageReference($pageReference)"))
+  #end
+  #if ($pages.size() > 0)
+    $stringtool.join($pages, '-')
```
mflorea: I'm not sure about this separator, we don't use it elsewhere. Why not use a comma ``,``?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
```diff
@@ -1 +1,40 @@
...
+    #xpropertySuggestInputDisplayer($field $prefix $name $value)
+  #else
+    $doc.displayEdit($field, $prefix, $object)
+  #end
+#elseif ($type == 'view')
+  #set ($pages = [])
+  #foreach($pageReference in $value)
+    #set ($discard = $pages.add("#displayPageReference($pageReference)"))
+  #end
+  #if ($pages.size() > 0)
```
mflorea: I would not mix the code that determines the values with the code that displays the values (the call to ``displayPageReference``). When determining the values there are 3 cases, see the ``suggestInput`` macro in [macros.vm](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm#L2544):

1. ``$value`` is an array (because relational storage is enabled)
1. ``$value`` is not an array and multiple selection is enabled (because relational storage is disabled)
1. ``$value`` is not an array and single selection is enabled

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_page.vm
atallahade: @mflorea Pages ended up separated with a comma and spaces like that: `Page1 , Page2 , Page3`. I'll fix the issue.

mflorea:
Regarding multiple selection without relational storage, indeed, the value returned is an array. It is a string for Users and Groups properties because they use ``LargeStringProperty`` to store the value.

PR #780, opened 2018-07-18
XWIKI-15434: Create a displayer for the Page field that supports autocomplete

atallahade:
Fix failing functional test.


xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/appwithinminutes/UserClassFieldTest.java
```diff
@@ -136,11 +136,10 @@ private void assertUserSuggestion(SuggestionElement user, String name, String...
         }
 
         if (image != null) {
-            Assert.assertTrue(user.getIcon().contains("/" + image));
+            Assert.assertTrue(user.getIcon().contains(image));
         } else {
             try {
-                user.getIcon();
-                Assert.fail();
+                Assert.assertEquals("", user.getIcon());
```
mflorea: Why did you change this? The point was that if ``image`` is null then the test wants to check that there is no icon, in which case ``getIcon()`` should throw an exception (element not found). ``xwiki.selectize.js`` does

    output.find('.xwiki-selectize-option-icon').remove();

This issue is that it does this only if ``typeof icon === 'object'``. You need to fix that instead.

PR #785, opened 2018-07-20
XWIKI-15454: Wrong suggest input size in object edition

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15454

### Changes

* Use the xform and skin css rules for the editor inputs.

### Screenshots
*Before:*
image-2018-07-20-14-40-03-958

*After:*
image

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
@@ -18,14 +18,6 @@
   margin: 0;
   position: static;
 }
-/* Basic form elements styling */
-#xwikiobjects input, #xwikiobjects textarea,
-#xwikiobjects select, #xwikiobjects .selectize-input,
-#xwikiclassproperties input, #xwikiclassproperties textarea,
-#xwikiclassproperties select, #xwikiclassproperties .selectize-input {
-  border: 1px solid $theme.borderColor;
-  width: 90%;
-}
 div#xwikiobjects input[type="checkbox"], div#xwikiclassproperties input[type="checkbox"],
 div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio"] {
   width: auto;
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
```diff
@@ -20,7 +20,9 @@
 }
 /* Basic form elements styling */
 #xwikiobjects input, #xwikiobjects textarea,
-#xwikiclassproperties input, #xwikiclassproperties textarea {
+#xwikiobjects select, #xwikiobjects .selectize-input,
+#xwikiclassproperties input, #xwikiclassproperties textarea,
+#xwikiclassproperties select, #xwikiclassproperties .selectize-input {
   border: 1px solid $theme.borderColor;
   width: 90%;
```
evalica: The correct solution here is to use .xform

atallahade: @evalica You mean replacing `#xwikiclassproperties` and `#xwikiobjects` with `.xform`?

mflorea: I don't think ``dataeditors.css`` should know about ``selectize.js`` classes.

atallahade: Makes sense, I'll move the related code in `selectize.css`.

evalica: in 10.7 we apparently have the xform class added to form#update. So instead of adding more selectors, I would just remove the border and the width declarations and rely on .xform+Skin styling.
Before
before

After
after


for http://localhost:8080/xwiki/bin/edit/Help/Applications/Movies/Meet%20John%20Doe?editor=object&force=1 with 
```
#xwikiobjects input, #xwikiobjects textarea, #xwikiclassproperties input, #xwikiclassproperties textarea {
    /* border: 1px solid #E8E8E8; */
    /* width: 90%; */
}
```
in dataeditors.css

evalica: With suggest objects type, After: 
suggest

atallahade:
@mflorea We were still discussing the issue. I've pushed a new commit to add the `xform` css class to class editor inputs. Here is what it looks like:
*On the class editor:*
image

*The object editor: (nothing changed here)*
image

PR #799, opened 2018-08-02
XWIKI-15435: CTRL+V in the Goto Modal (CTRL+G) doesn't work

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15435

### Changes

* Remove the META+V shortcut.


aubincleme:
Looks good to me as META+V is also missleading in this context. Generally, we shouldn't change the behavior of shortcuts that are commonly used by browsers to do a certain task.

PR #800, opened 2018-08-02
XWIKI-15474: Toggles aren't saved in the UI when changing the page of the Filters

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15474

### Changes

* Cache livetable row elements.

### Result

*Before:*
fix-livetable-cache-before

*After:*
fix-livetable-cache-after

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
@@ -127,7 +127,6 @@ XWiki.widgets.LiveTable = Class.create({
     this.handler = handler || function(){};
     this.totalRows = -1;
     this.fetchedRows = new Array();
-    this.renderedRows = new Array();
     this.getUrl = url;
     this.sendReqNo = 0;
     this.recvReqNo = 0;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -303,7 +304,10 @@ XWiki.widgets.LiveTable = Class.create({
 
     for (var i = off; i <= f; i++) {
       if (this.fetchedRows[i]) {
-        var elem = this.handler(this.fetchedRows[i], i, this);
+        if (this.renderedRows[i] === undefined) {
+          this.renderedRows[i] = this.handler(this.fetchedRows[i], i, this);
+        }
+        var elem = this.renderedRows[i];
```
Enygma2002: The idea of caching the DOM rendering produced by the `handler()` method.

However, I think it's not good to cache the replies in a different array that is not synthconized with the data `fetchedRows`. If you follow `fetchedRows`, you will see that it gets updated in various situtations and, if `renderedRows` is not updated as well, each time, you will most likely end up showing stale renderings in the UI that no longer correspond to the new data inside `fetchedRows`.

What I propose instead is that you set some special `this.fetchedRows[i].livetableRenderedRow` property (hoping no LivetableResults implementation actually defines a property named `livetableRenderedRow` in its JSON results rows) where you can store the rendering of individual `fetchedRows`. This way, whenever a data `fetchedRow` is removed or added, its rendering will stay close to it and be added/removed as well.

The only downside I can see in this proposed approach would be that if some `xwiki:livetable:newrow` event handlers or even the row renderer callback would want to modify the cached data row that it receives as parameter, its modification would not be reflected in the UI, unless maybe if it explicitly clears the cached rendering (by clearing `row.livetableRenderedRow` and maybe recalling `displayRows`). Anyway, I'm not sure this is supported right now either, so, IMO, it's safe to assume that the rendering done by the `this.handler` (i.e. javascript callback) method is cacheable. If you want to be extra safe, you could add a livetable option that controls if rendering caching is performed or not (enabled by default), so as to not block anyone that might be doing weird client side hacks. I personally don't think it's worth the complication, since we don't support that by design.

Enygma2002: Was also thinking about javascript event listeners that might be registered on the row renderings that get cached. You'd need to check and make sure that the listeners are still attached and working once you start adding and removing rendered rows multiple times on the DOM.

atallahade: AFAIK javascript event listeners bind to an element are not automatically removed by the gc until the element itself is removed. In our case, removing a row from the DOM doesn't make the gc remove it as long as it is still referenced in our cache.
I'll make sure this is the case.

atallahade:
@Enygma2002 I had to disable triggering new row events when the rendered row was taken from the cache as it was creating issue (duplication of events attached to the row): https://github.com/xwiki/xwiki-platform/pull/800/files#diff-0be0c0cd10986487e48f31e0197eeb36R317

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -303,18 +303,22 @@ XWiki.widgets.LiveTable = Class.create({
...
+          var memo = {
+            "data": this.fetchedRows[i],
+            "row":elem,
+            "table":this,
+            "tableId":this.domNodeName
+          };
+          // 1. Named event (for code interested by that table only)
+          document.fire("xwiki:livetable:" + this.domNodeName + ":newrow", memo);
+          // 2. Generic event (for code potentially interested in any livetable)
+          document.fire("xwiki:livetable:newrow", memo);
```
Enygma2002: Note that this will slightly change the meaning of the `newrow` event, in the way that it will no longer trigger on each **display** of a livetable row, but on the actual **rendering** of a new livetable row, which is done only once for the set of cached data.

In practice, I believe this change will have no negative impact, as the general usecase for this event is to augment the rendering of a livetable row and, since we will cache each rendered row, it will be enough (and better performance-wise) to do it once and reuse it on each display.

PR #803, opened 2018-08-06
XWIKI-15452: Use auto-suggestion on xproperties that can support it

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15452
http://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference

### Changes

* Introduce a generic page picker.
* Use the Page / User property on these fields:
  * XWiki.SearchSuggestSourceClass
    * Service
  * Global Administration: Export
    * Author
  * XWiki.DeletedDocuments
    * Livetable: page type
  * Class editor (Lists)
    * XWiki Class Name
  * XWiki.XWikiUsers
    * Skin
  * XWiki.XWikiServerClass
    * Home page
  * WikiManager.AdminWikiDescriptorSheet (Global Administration: Descriptor)
    * Home page
  * XWiki.XWikiPreferences
    * Parent Space
    * Skin
    * Right/Left panels
    * Internationalization Document Bundles
  * XWiki.XWikiComments
    * Author (also fix how comment authors are being displayed)
    * Target
  * XWiki.WatchListJobClass
    * Template
  * XWiki.WatchListClass
    * Documents
    * Users
  * XWiki.Notifications.Code.NotificationFilterPreferenceClass
    * Users
    * Pages
  * XWiki.TemplateProviderClass
    * Template
  * XWiki.RedirectClass
    * Location
  * XWiki.Registration
    * Default redirect
  * XWiki.FeedEntryClass
    * Author
  * XWiki.ConfigurableClass
    * Configuration class
  * XWiki.DocumentSheetBinding
    * Sheet
  * XWiki.ClassSheetBinding
    * Sheet
  * AppWithinMinutes.MetadataClass
    * Data space name
  * AppWithinMinutes.LiveTableClass
    * Class
    * Data space
  * WikiManager.WikiCandidateMemberClass
    * User Name
    * Reviewer
  * PanelsCode.NavigationConfigurationClass
    * Inclusions
    * Exclusions
  * Invitation.WebHome
    * Email Class
    * Email Container
  * Invitation.InvitationMailClass
    * Sending User

### Notes

This PR is meant to add an autocomplete support to most of the inputs listed in [this design page](http://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference).

Here is a list of tasks, from the design page, that have not (or partially) been taken care of:
1. [Template: Docextra](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HTemplate:Docextra)
    Out of scope.
2. [XWiki.XWikiUsers](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:XWiki.XWikiUsers)
    We might need another picker to suggest attachments on the Avatar field.
3. [App: Page Index](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HApp:PageIndex)
    *Deleted Pages*: the search input of the *Page* column doesn't suggest deleted documents.
    *Deleted Attachments*: suggest attachments problem.
4. [XWiki.XWikiPreferences](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:XWiki.XWikiPreferences2BAdministration:Look26Feel-Themes2CPanels)
    The page picker hasn't been implemented for the *Skin* field of the Administration page (see the screenshot of the design page).
5. [XWiki.XWikiGroups](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:XWiki.XWikiGroups)
    The Member property can take either a User or a Group entity.
6. No suggest for *Wiki* and *Space* properties for:
    * [XWiki.WatchListClass](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:XWiki.WatchListClass)
    * [XWiki.Notifications.Code.NotificationFilterPreferenceClass](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:XWiki.Notifications.Code.NotificationFilterPreferenceClass)
    * [XWiki.TemplateProviderClass](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:XWiki.TemplateProviderClass)
7. [TourCode.TourClass + TourCode.StepClass](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HClass:TourCode.TourClass2BTourCode.StepClass2BApp:Tour)
    Need to make the change directly in the [tour application](https://github.com/xwiki-contrib/application-tour).
8. [Create, Copy, Rename - Advanced](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HCreate2CCopy2CRename-Advanced)
    Need to recreate or change the location picker.
9. [Inputs inside WYSIWYG macros](https://design.xwiki.org/xwiki/bin/view/Proposal/AutocompleteOnReference#HWYIWYGMacros)
    Requires a different work from what was done here. Another PR will be made.

Some page properties might need a custom SQL query (to apply a filter):
* *Skin* properties? (e.g. XWiki.XWikiPreferences > Skin)
* *Space* properties? (e.g. XWiki.XWikiPreferences > Parent Space)

The [json/xml response](https://github.com/xwiki/xwiki-platform/blob/a196061711ff91fc9f7e2f038e464709a3b55b43/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pagesuggest.vm) generated by the generic page picker has been implemented in the same way as [the one for users and groups](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/uorgsuggest.vm). Using the REST API on [`wikis/query`](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#H2Fwikis2Fquery3Fq3D7Bquery7D26wikis3DwikiList5B26distinct3D7Btrue2Cfalse7D5D5B26order3D7Basc2Cdesc7D5D5B26start3Dn5D5B26number3Dn5D5B26prettyNames3D7Btrue2Cfalse7D5D) would have create inconsistencies between the results of the generic page picker and the ones of the page picker used for property values ([`/wikis/{wikiName}/classes/{className}/properties/{property}/values`](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#H2Fwikis2F7BwikiName7D2Fclasses2F7BclassName7D2Fproperties2F7Bproperty7D2Fvalues)). I think it would have been better to have a common / generic API instead of duplicating code.

Another point is that an input from the class editor now support the autocomplete. A problem will occur if a user tries, for instance, to create a Page property inside an empty class. The related input will be shown like this:
image
instead of
image

The problem is that the javascript file (`suggestPages.js`) is not fetched by the AJAX call once the created property is displayed. It would be nice to use the [same mechanism](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.js#L159-L176) as the object editor.

Edit: Also can't properly look for the Main.WebHome page on the page picker.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/LeftPanelsUIExtensionManager.java
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/LeftPanelsUIExtensionManager.java
@@ -19,8 +19,11 @@
  */
 package org.xwiki.panels.internal;
 
+import java.util.List;
+
 import javax.inject.Named;
 import javax.inject.Singleton;
+
 import org.xwiki.component.annotation.Component;
 
 /**
@@ -35,7 +38,7 @@ import org.xwiki.component.annotation.Component;
 public class LeftPanelsUIExtensionManager extends AbstractPanelsUIExtensionManager
 {
     @Override
-    protected String getConfiguration()
+    protected List<String> getConfiguration()
     {
         return this.configurationSource.getProperty("leftPanels");
     }
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ConfigurableClass.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ConfigurableClass.xml
@@ -485,17 +485,29 @@ $doc.dropPermissions()##
       <classType>com.xpn.xwiki.objects.classes.TextAreaClass</classType>
     </codeToExecute>
     <configurationClass>
+      <cache>0</cache>
+      <classname/>
       <customDisplay/>
       <disabled>0</disabled>
+      <displayType>input</displayType>
+      <hint/>
+      <idField/>
+      <multiSelect>0</multiSelect>
       <name>configurationClass</name>
       <number>3</number>
-      <picker>0</picker>
+      <picker>1</picker>
       <prettyName>configurationClass</prettyName>
       <size>30</size>
+      <relationalStorage>0</relationalStorage>
+      <separator> </separator>
+      <separators/>
+      <sort>none</sort>
+      <sql/>
       <unmodifiable>0</unmodifiable>
       <validationMessage/>
       <validationRegExp/>
-      <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      <valueField/>
+      <classType>com.xpn.xwiki.objects.classes.PageClass</classType>
     </configurationClass>
     <configureGlobally>
       <customDisplay/>
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
@@ -907,16 +907,29 @@
     <nameField/>
     <validationScript/>
     <defaultRedirect>
+      <cache>0</cache>
+      <classname/>
+      <customDisplay/>
       <disabled>0</disabled>
+      <displayType>input</displayType>
+      <hint/>
+      <idField/>
+      <multiSelect>0</multiSelect>
       <name>defaultRedirect</name>
       <number>7</number>
-      <picker>0</picker>
+      <picker>1</picker>
       <prettyName>Redirect here after registration</prettyName>
       <size>30</size>
+      <relationalStorage>0</relationalStorage>
+      <separator> </separator>
+      <separators/>
+      <sort>none</sort>
+      <sql/>
       <unmodifiable>0</unmodifiable>
       <validationMessage/>
       <validationRegExp/>
-      <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      <valueField/>
+      <classType>com.xpn.xwiki.objects.classes.PageClass</classType>
     </defaultRedirect>
     <heading>
       <disabled>0</disabled>
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderClass.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/TemplateProviderClass.xml
@@ -172,17 +172,29 @@
       <classType>com.xpn.xwiki.objects.classes.DBListClass</classType>
     </spaces>
     <template>
+      <cache>0</cache>
+      <classname/>
       <customDisplay/>
       <disabled>0</disabled>
+      <displayType>input</displayType>
+      <hint/>
+      <idField/>
+      <multiSelect>0</multiSelect>
       <name>template</name>
       <number>2</number>
-      <picker>0</picker>
+      <picker>1</picker>
       <prettyName>template</prettyName>
       <size>30</size>
+      <relationalStorage>0</relationalStorage>
+      <separator> </separator>
+      <separators/>
+      <sort>none</sort>
+      <sql/>
       <unmodifiable>0</unmodifiable>
       <validationMessage/>
       <validationRegExp/>
-      <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      <valueField/>
+      <classType>com.xpn.xwiki.objects.classes.PageClass</classType>
     </template>
     <terminal>
       <customDisplay/>
--- a/xwiki-platform-core/xwiki-platform-annotations/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/internal/AnnotationClassDocumentInitializer.java
+++ b/xwiki-platform-core/xwiki-platform-annotations/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/internal/AnnotationClassDocumentInitializer.java
@@ -81,7 +81,7 @@ public class AnnotationClassDocumentInitializer extends AbstractMandatoryClassIn
             ContentType.PURE_TEXT);
         xclass.addTextAreaField(Annotation.ORIGINAL_SELECTION_FIELD, "Original Selection", 40, 5,
             ContentType.PURE_TEXT);
-        xclass.addTextField(Annotation.TARGET_FIELD, "Target", 30);
+        xclass.addPageField(Annotation.TARGET_FIELD, "Target", 30);
         xclass.addTextField(Annotation.STATE_FIELD, "State", 30);
     }
 
--- a/xwiki-platform-core/xwiki-platform-annotations/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/AnnotationClass.xml
+++ b/xwiki-platform-core/xwiki-platform-annotations/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/AnnotationClass.xml
@@ -166,16 +166,25 @@
       <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
     </state>
     <target>
+      <cache>0</cache>
+      <classname/>
+      <customDisplay/>
       <disabled>0</disabled>
       <name>target</name>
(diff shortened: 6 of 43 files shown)


xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/LeftPanelsUIExtensionManager.java
```diff
@@ -35,7 +38,7 @@
 public class LeftPanelsUIExtensionManager extends AbstractPanelsUIExtensionManager
 {
     @Override
-    protected String getConfiguration()
+    protected List<String> getConfiguration()
```
tmortagne: If you have to do it it means you introduced a breaking change on XWikiPreference side. I guess it means that any code which manipulate/request panels configuration is going to be broken.

atallahade: Could you give me an example? The class is internal and the method protected so I didn't thought there could be a breaking change outside the package.

tmortagne: What I said is that you had to modify this method (which is not a breaking change) BECAUSE you introduce a breaking change: the list of plugins is not stored the same way anymore.

atallahade: I've change the type of the `XWiki.XWikiPreferences > leftPanels` field, from String to Page, but the way it is stored doesn't change. In this case the relational storage is set to false which mean a [`StringListProperty`](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java#L414) will be created instead of a [`StringProperty`](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java#L83) but AFAIK they are stored in the same way in the database (i.e. as a string where values are separated by a comma or something else). I could have avoid using a `List<String>` and just convert the list to a string [here](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/RightPanelsUIExtensionManager.java#L41).
Though, I'm not very aware of what could be broken (other than internal code) after changing `String` fields to `Page` fields. Maybe you can give me some examples?

tmortagne: StringListProperty yes but the separator changed. StringListProperty uses `|` as separator while leftPanels separator used to be based on `,`.

atallahade: I can probably force the separator to be `,` instead of `|` but I need to know how I can make sure no breaking changes have been made.

tmortagne: Well in theory this change should break all panels in XWiki Standard since they are listed with a `,`: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml#L2739

tmortagne: Anyway even if you fix the separator (which might not be as easy as it sounds) there is still the breakage illustrated by your getConfiguration change: getValue() now return a List while any code which manipulate this field expect a String.

atallahade: It seems that the separator used is a space ` `, but still my panels are not broken and they are well listed:
image
I can set the `separators` of all multi-select Page fields to `|, ` if you think that can help.

tmortagne: > It seems that the separator used is a space

Definitely not a space in https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml#L2739 and https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java#L99.

atallahade: > getValue() now return a List while any code which manipulate this field expect a String.

That's true, the value bound to the property will be a `List` instead of a `String`. There might be other cases like that so I can see at least 2 solutions:
1. Put back the type of the field to `StringClass`.
2. Keep the `PageClass` but override the [`newProperty`](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java#L407-L420) method to create a `StringProperty` in case we are using a multi-select field, without relational storage and when some new value is set to true (to use a String instead of a List). This solution is too complex I think.

Maybe you have other ideas?

> Definitely not a space

Sorry, that was the default separator for the page field, which isn't used by the string field.

tmortagne: So I did some debugging and the reason why upgrade "works" is mostly because of  things:
* XWikiPreference is very special and is not really based on standard properties tables (it's custom mapped on a dedicated table) which means it's going trough BaseClass#fromValueMap instead of loading each object property based on the table
* the hack you added in https://github.com/xwiki/xwiki-platform/pull/803/files#diff-4bbd11fef5480d7e7db21e44ee739043R851. This change is actually wrong because you are mixing UI things (fromString which is supposed to deal with html form values, so UI stuff) and storage thing (fromValue) which don't always work the same way. For example by default in ListClass the UI syntax and storage syntax is not exactly the same (not the same separators plus escaping support on storage side).

Putting aside retro compatibility, instead of calling fromString in fromValue the right way to support StringListProperty in a custom mapping (other than rewritting all this mess) would be to overwrite setValue in StringListProperty (to not crash and burn when getting a String) IMO.

I agree that given the javadoc available on fromString and fromValue/fromValueMap it was not exactly obvious for you :)

atallahade: Ok thanks, I'll do the change.

tmortagne: Actually I see that a getCustomMappingValue (in BaseProperty) exist so maybe a better fix would be to introduce setCustomMappingValue.

atallahade:
Closing this PR as I've been splitting it into several ones.

PR #808, opened 2018-08-09
XWIKI-15486: Broken less code when saving FlamingoThemesCode.ThemeClass

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15486

### Change

* Set the editor of lessCode from FlamingoThemesCode.ThemeClass to Text.


tmortagne:
Actually what need to be set is not the editor but the content type and the editor will follow.

PR #819, opened 2018-08-24
XWIKI-15568: Create a generic page picker

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15568

### Change

* Introduce a generic page picker.
* Use the Search resources REST API for the generic page picker.

### Note

This change was taken from #803.

The [json/xml response](https://github.com/xwiki/xwiki-platform/blob/a196061711ff91fc9f7e2f038e464709a3b55b43/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pagesuggest.vm) generated by the generic page picker has been implemented in the same way as [the one for users and groups](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/uorgsuggest.vm). Using the REST API on [`wikis/query`](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#H2Fwikis2Fquery3Fq3D7Bquery7D26wikis3DwikiList5B26distinct3D7Btrue2Cfalse7D5D5B26order3D7Basc2Cdesc7D5D5B26start3Dn5D5B26number3Dn5D5B26prettyNames3D7Btrue2Cfalse7D5D) would have create inconsistencies between the results of the generic page picker and the ones of the page picker used for property values ([`/wikis/{wikiName}/classes/{className}/properties/{property}/values`](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#H2Fwikis2F7BwikiName7D2Fclasses2F7BclassName7D2Fproperties2F7Bproperty7D2Fvalues)). I think it would have been better to have a common / generic API instead of duplicating code.

**Edit**:
The page picker now uses the Search resources REST API (Solr) to perform the request.


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -2322,6 +2328,31 @@ $string##
...
+#**
+* Pulls all the JavaScript and CSS resources needed by the page picker.
+*#
+#macro (pagePicker_import)
+  #picker_import
+  #set ($discard = $xwiki.jsfx.use('uicomponents/suggest/suggestPages.js',
+    {'forceSkinAction': true, 'language': $xcontext.locale}))
+#end
+
+#macro (pagePicker_input $multiSelect $parameters)
```
mflorea: I wouldn't keep ``$multiSelect`` as separate parameter. I know the ``userPicker`` macro has this but:

* when I added the ``userPicker`` macro 3 years ago I was thinking that ``$parameters`` are just attributes to pass to the picker input.
* with the new user picker it's not just attributes but I kept the "signature" of the macro to not break existing calls.

Thus I would remove the ``pagePicker_input`` macro and simply call

    #suggestInput($parameters)

in ``pagePicker``, which has a single parameter ``$parameters``. WDYT?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,66 @@
...
+    return {
+      create: true,
+      load: function(text, callback) {
+        loadPages(text).done(callback).fail(callback);
+      }
+    }
+  };
+
+  var loadPages = function(text) {
+    // ${request.contextPath}/rest/wikis/query/
```
mflorea: The [old page "picker"](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/SpecialCSSClasses/#HB2.Behaviorclassnames) that you can obtain using the ``.suggestDocuments`` CSS class is using (see [xwiki.js](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js#L1498)) the following REST URL

    ${request.contextPath}/rest/wikis/__wiki__/search?scope=name&number=10

(which uses database search rather than Solr) You can check for instance the "Go To Page" (Ctrl + G) feature. So I would rather use the same service to get the suggestions, improving it of course if we need to.

The **generic** page picker and the Page **property** picker have different suggestions in may cases:

* when an explicit query is used
* when the implicit query restricts the pages using the class name field
* the generic picker doesn't have "used values"

so I think it's fine that they don't use the same service to get the suggestions and thus that there may be differences in the provided suggestions.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/pagesuggest.vm
```diff
@@ -0,0 +1,85 @@
+#macro (suggestPages)
```
mflorea: I wouldn't add this new template unless there are good reasons that prevent us from using the existing REST API. See my comment above.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
atallahade: Yes it makes more sense to have the `multiple` parameter directly in the `$parameters` map.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
atallahade: We can try to use the existing REST API to fetch suggested documents but I'm afraid that the *generic page picker* won't have the same behavior as the *page property picker* when the latter is used with its default behavior (no explicit query or other restrictions). I'll do some tests.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,90 @@
...
+        }).fail(callback);
+      }
+    }
+  };
+
+  var loadPages = function(text) {
+    // We need to escape backslashes for Solr.
+    text = text.replace(/\\/g, "\\\\");
+    var response = $.getJSON("${request.contextPath}/rest/wikis/query/", {
+      'q': "title:*" + text + "* or fullname:*" + text + "*",
```
mflorea: Computing the Solr query from the user input can lead to unexpected results. For instance if the text is "one two" then your query is ``title:*one two*`` which means:

* match the titles that **end** with "one"
* match **any field** that starts with "two"

You could use [grouping](https://lucene.apache.org/solr/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-GroupingClauseswithinaField) but you'd get ``title:(*one two*)`` which means:

* match the titles that **end** with "one"
* match the titles that **start** with "two"

And this can go on. The bottom line is: avoid computing the Solr query from the user input. The safest is to pass the user input as Solr query  without any modification and to use the query parameters to restrict / tune the search. For instance instead of doing:

    q=title:text or fullname:text

you could write:

    q=text
    qf=title fullname

Which means: look for "text" in the specified query fields (title and fullname). I don't know if the ``/rest/wikis/query/`` REST resource supports specifying the query fields parameter, but it probably should.

Now, even with this you still have the problem with the wildcards. Note that the live search (search suggest) from the top right corder of the page doesn't use wildcards and neither the tree finder, but there have been complains about this. The real fix for this is to index more information, e.g. using [NGrams](http://blog.florian-hopf.de/2014/03/prefix-and-suffix-matches-in-solr.html). In other words, with Solr, if there's a search problem you don't fix the query, you fix the index (most of the time).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,90 @@
...
+    return {
+      create: true,
+      load: function(text, callback) {
+        loadPages(text).done(function(data) {
+          var pages = [];
+          data.searchResults.forEach(function (element) {
+            var label = element.title;
+            var queryString = "";
+            // Separate spaces with " / " and unescape characters. E.g. A\.B.C\\D => A.B / C\D
+            var hint = element.space.replace(/([^\\])\./g, "$1 / ").replace(/\\(.)/g, "$1");
```
mflorea: This is error-prone and this code shouldn't be aware or care about the reference syntax. Moreover, [the ``space`` field is deprecated](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/managed-schema#L133). If it weren't deprecated you should have resolved the space reference (as you do below with the document reference). But since it is, you could use the ``spaces`` field which is an array of space names, making it even easier to compute the hint. But, if you do this, the hint will still be made of space names not titles. The Page property picker computes the hint by rendering the title of each parent page. Same does the breadcrumb displayed on top of each page. We should display the "space" titles here also. One option is to make an AJAX request to ``hierarchy_reference.vm`` for each search result (suggestion), but it may be too costly..

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,90 @@
...
+          callback(pages);
+        }).fail(callback);
+      }
+    }
+  };
+
+  var loadPages = function(text) {
+    // We need to escape backslashes for Solr.
+    text = text.replace(/\\/g, "\\\\");
+    var response = $.getJSON("${request.contextPath}/rest/wikis/query/", {
```
mflorea: I guess the REST resource that uses database search is not good enough?

atallahade: @mflorea I thought we wanted to use the Solr search, didn't we? You want me to try using [/wikis/{wikiName}/search](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#H2Fwikis2F7BwikiName7D2Fsearch3Fq3D7Bkeywords7D5B5B26scope3D7Bname2Ccontent2Ctitle2Cobjects7D...5D26start3Dn5D5B26number3Dn5D5B26orderField3Dfield26order3D7Basc2Cdesc7D5D5Bdistinct3D7Btrue2Cfalse7D5D5B26prettyNames3D7Btrue2Cfalse7D5D)?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
atallahade: > The real fix for this is to index more information, e.g. using NGrams. In other words, with Solr, if there's a search problem you don't fix the query, you fix the index (most of the time).

Yes I thought about that but can we change the Solr configuration?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -2322,6 +2328,28 @@ $string##
...
+#**
+* Pulls all the JavaScript and CSS resources needed by the page picker.
+*#
+#macro (pagePicker_import)
+  #picker_import
+  #set ($discard = $xwiki.jsfx.use('uicomponents/suggest/suggestPages.js',
+    {'forceSkinAction': true, 'language': $xcontext.locale}))
+#end
+
+#macro (pagePicker_input $parameters)
```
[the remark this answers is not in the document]
atallahade: Mostly to be more consistent with the `userPicker` and so that we can add other instructions in the macro but it's not very useful for now indeed. I'll remove it.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
mflorea: We can of course but we need to be aware of the consequences. Indexing prefixes/suffixes can grow the size of the index and can increase the index time. The search time shouldn't be affected normally. In any case, this is something to discuss on the mailing list and we need to see if it fits in the roadmap.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
mflorea: In a previous comment I suggested you to try to use the same REST resource as the "Go To Page" (Ctrl + G) feature. Which is:

    ${request.contextPath}/rest/wikis/__wiki__/search?scope=name&number=10

I don't know if it's possible or easy, but I would investigate it at least.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
mflorea: > I am not able to find the ``spaces`` field

Sorry, I thought the search response matches the data we have in the [Solr index](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-data/src/main/resources/xwiki/conf/managed-schema#L133). I was referring to the ``space`` and ``spaces`` fields from the Solr index, but the response is actually the serialization of a list of [``SearchResult``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/resources/xwiki.rest.model.xsd#L540), which is part of the REST model and generic (used by all query types, [``SOLRSearchSource``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-rest/src/main/java/org/xwiki/search/solr/internal/rest/SOLRSearchSource.java#L160) being just one of the implementations).

>  add a ``spaces`` field in the ``/wikis/{wikiName}/search`` endpoint

We could but we don't need it, since you can use the [JavaScript API](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/JavaScriptAPI/#HWorkwithEntityReferences) to resolve the value of the ``space`` field.

atallahade: > We could but we don't need it, since you can use the JavaScript API to resolve the value of the `space` field.

Sure but it doesn't give you the title of the spaces. Should we just display the spaces (or document) name for now? As you said resolving the space title with AJAX calls is very inefficient.

mflorea: I see. I wouldn't put the space titles in the ``spaces`` field anyway. I would use a different field, say ``breadcrumb`` or ``hierarchy`` or ``path``. I'd keep ``spaces`` for the space names. I'm OK to add this to the ``SearchResult`` model.

atallahade:
@mflorea I've updated the PR. Note that I do not use the `url` attribute in the hierarchy because the url is absolute and the hierarchy is only used in one place (`searchPages` method).

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java
```diff
@@ -282,6 +290,23 @@
                     pageLink.setRel(Relations.PAGE);
                     searchResult.getLinks().add(pageLink);
 
+                    Hierarchy hierarchy = new Hierarchy();
+                    for (EntityReference entityReference : doc.getDocumentReference().getReversedReferenceChain()) {
+                        HierarchyItem hierarchyItem = new HierarchyItem();
+                        Document document = xwikiApi.getDocument(entityReference);
```
mflorea: I would call this only for space and document entities below.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java
```diff
@@ -282,6 +290,23 @@
...
+                        if (entityReference instanceof SpaceReference || entityReference instanceof DocumentReference) {
```
mflorea: We usually check the entity type:

    entityReference.getType() == EntityType.SPACE

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java
```diff
@@ -282,6 +290,23 @@
...
+                            hierarchyItem.setLabel(document.getPlainTitle());
+                        } else {
+                            hierarchyItem.setLabel(entityReference.getName());
+                        }
+                        hierarchyItem.setName(entityReference.getName());
+                        hierarchyItem.setType(entityReference.getType().getLowerCase());
+                        hierarchyItem.setEntityReference(entityReferenceSerializer.serialize(entityReference));
+                        hierarchyItem.setUrl(document.getURL());
```
mflorea: This is not right for attachments for instance. Better use:

    xwikiApi.getURL(entityReference)

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/resources/xwiki.rest.model.xsd
```diff
@@ -618,4 +619,20 @@
       <element name="element" type="string" minOccurs="0" maxOccurs="unbounded"></element>
     </sequence>
   </complexType>
+
+  <complexType name="HierarchyItem">
+    <sequence>
+      <element name="label" type="string"></element>
+      <element name="name" type="string"></element>
+      <element name="type" type="string"></element>
+      <element name="entityReference" type="string"></element>
```
mflorea: ``entityReference`` is not consistent with ``name`` and ``type``. Between ``entityName / entityType / entityReference`` and ``name / type / reference`` I would choose the later. But since you added ``name`` I would drop ``reference / entityReference`` for now. We can add it later if it's needed.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,81 @@
...
+            });
+          });
+          callback(pages);
+        }).fail(callback);
+      }
+    }
+  };
+
+  var loadPages = function(text) {
+    var url = XWiki.Document.WikiSearchURLStub.replace("__wiki__", XWiki.Document.currentWiki);
```
mflorea: You can use [``XWiki.Document.getRestSearchURL()``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js#L1397).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,81 @@
...
+    return $.getJSON(url, 'q=' + text + '&scope=name&scope=title&number=10&media=json');
```
mflorea: Either pass the query string as an object and let jQuery handle the URL encoding or do the URL encoding yourself (for ``text``). I prefer the first option. It also makes the query string more easy to read.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,81 @@
...
+          var pages = [];
+          data.searchResults.forEach(function (element) {
+            var hierarchy = element.hierarchy.elements;
+            var label = hierarchy.pop();
+            if (element.pageName === 'WebHome') {
+              label = hierarchy.pop();
+            }
+            var hint = hierarchy.join(' / ');
+            var doc = new XWiki.Document(XWiki.Model.resolve(element.id, XWiki.EntityType.DOCUMENT))
+            var url = doc.getURL("view");
```
mflorea: I think you can skip the "view" parameter (it's the default action).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,81 @@
...
+      load: function(text, callback) {
+        loadPages(text).done(function(data) {
+          var pages = [];
+          data.searchResults.forEach(function (element) {
+            var hierarchy = element.hierarchy.elements;
+            var label = hierarchy.pop();
+            if (element.pageName === 'WebHome') {
+              label = hierarchy.pop();
+            }
+            var hint = hierarchy.join(' / ');
```
mflorea: It's hard to understand the above code without knowing how the JSON data looks like. You posted the XML version.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
atallahade: That's what I did before but for some reason I was not able to pass multiple scope. I tried something like:
```javascript
{
  'scope': ['name', 'title']
}
```
The two scope are added into the query string but contains the `[` and `]` for some reason.

atallahade:
@mflorea Example of a json response:
```json
{  
   "links":[  

   ],
   "searchResults":[  
      {  
         "links":[  
            {  
               "href":"http://localhost:8080/xwiki/rest/wikis/xwiki/spaces/Main/pages/WebHome/translations/fr",
               "rel":"http://www.xwiki.org/rel/page",
               "type":null,
               "hrefLang":null
            }
         ],
         "type":"page",
         "id":"xwiki:Main.WebHome",
         "pageFullName":"Main.WebHome",
         "title":"Home",
         "wiki":"xwiki",
         "space":"Main",
         "pageName":"WebHome",
         "modified":1536259252000,
         "author":"XWiki.superadmin",
         "authorName":null,
         "version":"1.1",
         "language":"fr",
         "className":null,
         "objectNumber":null,
         "filename":null,
         "score":null,
         "object":null,
         "hierarchy":{  
            "items":[  
               {  
                  "label":"xwiki",
                  "name":"xwiki",
                  "type":"wiki",
                  "entityReference":"xwiki",
                  "url":"http://localhost:8080/xwiki/bin/view/Main/"
               },
               {  
                  "label":"Home",
                  "name":"Main",
                  "type":"space",
                  "entityReference":"xwiki:Main",
                  "url":"http://localhost:8080/xwiki/bin/view/Main/"
               },
               {  
                  "label":"Home",
                  "name":"WebHome",
                  "type":"document",
                  "entityReference":"xwiki:Main.WebHome",
                  "url":"http://localhost:8080/xwiki/bin/view/Main/"
               }
            ]
         }
      },
      {  
         "links":[  
            {  
               "href":"http://localhost:8080/xwiki/rest/wikis/xwiki/spaces/Sandbox/spaces/AAAA/spaces/BBBB/spaces/CCCC&%5C../pages/WebHome",
               "rel":"http://www.xwiki.org/rel/page",
               "type":null,
               "hrefLang":null
            }
         ],
         "type":"page",
         "id":"xwiki:Sandbox.AAAA.BBBB.CCCC&\\\\\\.\\..WebHome",
         "pageFullName":"Sandbox.AAAA.BBBB.CCCC&\\\\\\.\\..WebHome",
         "title":"CCCC&amp;\\.. Title",
         "wiki":"xwiki",
         "space":"Sandbox.AAAA.BBBB.CCCC&\\\\\\.\\.",
         "pageName":"WebHome",
         "modified":1536332347000,
         "author":"XWiki.Admin",
         "authorName":null,
         "version":"5.1",
         "language":null,
         "className":null,
         "objectNumber":null,
         "filename":null,
         "score":null,
         "object":null,
         "hierarchy":{  
            "items":[  
               {  
                  "label":"xwiki",
                  "name":"xwiki",
                  "type":"wiki",
                  "entityReference":"xwiki",
                  "url":"http://localhost:8080/xwiki/bin/view/Main/"
               },
               {  
                  "label":"Sandbox",
                  "name":"Sandbox",
                  "type":"space",
                  "entityReference":"xwiki:Sandbox",
                  "url":"http://localhost:8080/xwiki/bin/view/Sandbox/"
               },
               {  
                  "label":"AAAA Title",
                  "name":"AAAA",
                  "type":"space",
                  "entityReference":"xwiki:Sandbox.AAAA",
                  "url":"http://localhost:8080/xwiki/bin/view/Sandbox/AAAA/"
               },
               {  
                  "label":"BBBB Title",
                  "name":"BBBB",
                  "type":"space",
                  "entityReference":"xwiki:Sandbox.AAAA.BBBB",
                  "url":"http://localhost:8080/xwiki/bin/view/Sandbox/AAAA/BBBB/"
               },
               {  
                  "label":"CCCC&\\.. Title",
                  "name":"CCCC&\\..",
                  "type":"space",
                  "entityReference":"xwiki:Sandbox.AAAA.BBBB.CCCC&\\\\\\.\\.",
                  "url":"http://localhost:8080/xwiki/bin/view/Sandbox/AAAA/BBBB/CCCC%26%5C../"
               },
               {  
                  "label":"CCCC&\\.. Title",
                  "name":"WebHome",
                  "type":"document",
                  "entityReference":"xwiki:Sandbox.AAAA.BBBB.CCCC&\\\\\\.\\..WebHome",
                  "url":"http://localhost:8080/xwiki/bin/view/Sandbox/AAAA/BBBB/CCCC%26%5C../"
               }
            ]
         }
      }
   ],
   "template":"http://localhost:8080/xwiki/rest/wikis/xwiki/search?q={keywords}(&scope={content|name|title|spaces|objects})*(&number={number})(&start={start})(&orderField={fieldname}(&order={asc|desc}))(&prettyNames={false|true})"
}
```

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
mflorea: Yes, look for ``traditional`` on https://api.jquery.com/jQuery.ajax/ . Since ``getJSON`` doesn't support advanced configuration options then you need to use https://api.jquery.com/jQuery.param/ with the second parameter set to ``true``.

atallahade:
@mflorea The PR has been updated. I was wondering if it wouldn't be cleaner to directly fetch the document instead of the fields here:
https://github.com/xwiki/xwiki-platform/blob/b4095dbabf797c0cbb176b3625e29ce1a5b27d20/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java#L165-L173

Unless cache is involved, I don't think it will have a great impact on performances as we are fetching the document anyway, when the user has view rights on it:
https://github.com/xwiki/xwiki-platform/blob/b4095dbabf797c0cbb176b3625e29ce1a5b27d20/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java#L240-L253

This change would also allow us to get the document with the right locale.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
```diff
@@ -0,0 +1,89 @@
...
+      create: true,
+      load: function(text, callback) {
+        loadPages(text).done(function(data) {
+          var pages = [];
+          data.searchResults.forEach(function (element) {
+            var hierarchy = element.hierarchy.items;
+            var label = hierarchy.pop().label;
+            if (element.pageName === 'WebHome') {
+              label = hierarchy.pop().label;
+            }
```
mflorea: That's fine, but I would probably take the label from ``element.title`` and use the `hierarchy`` just to compute the hint.

atallahade: `element.title` contains the display title whereas the `hierarchy` has the plain title for spaces and documents and the name of the entity for other types.
https://github.com/xwiki/xwiki-platform/blob/b4095dbabf797c0cbb176b3625e29ce1a5b27d20/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java#L254
https://github.com/xwiki/xwiki-platform/blob/b4095dbabf797c0cbb176b3625e29ce1a5b27d20/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java#L293-L298

mflorea:
> I do not use the url attribute in the hierarchy because the url is absolute

That's OK.

> The hierarchy is only used in one place (searchPages method)

I don't see the problem.

> I wanted to use constants for "WebHome" and "space" in the JS file but didn't found any, is there something to do about that?

There's nothing in JavaScript for ``WebHome``. You can of course put the value of

    $services.model.getEntityReference('DOCUMENT', 'default').name

in a JavaScript icon like you do with the page icon. For ``space`` I think it's OK.

> As you can also see below, some characters are not escaped (e.g. & could be &amp;), is that something necessary?

We don't do the XML serialization ourselves. If it's a bug, it's not in our code.

mflorea:
>  directly fetch the document

What exactly do you mean by this?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/suggestPages.js
mflorea: OK.

atallahade:
>> directly fetch the document

> What exactly do you mean by this?

`select distinct doc` instead of `select distinct doc.fullName, doc.space, doc.name, doc.language`

atallahade:
@mflorea As you can see on the JSON response that I've pasted above, the `xwiki:Main.WebHome` page was fetched when I typed "CC" in the picker even though the title of the page is "Home". It happened because, in fact, "Accueil" is the French title of the document.

We could select `doc.title` by adding it to this query
https://github.com/xwiki/xwiki-platform/blob/b4095dbabf797c0cbb176b3625e29ce1a5b27d20/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java#L165-L173
so that we use the title of the translated document instead of the one from the root locale.
My idea was to directly select `doc` instead of all these fields to make things cleaner.
The only possible downside I see about doing that would be a decrease on the performances but I'm not sure such a change would cause that.

atallahade:
@mflorea How do you think we should handle languages? There are at least two things to consider:
1. Which locales should we do the search with?
2. What do we do in case the document is found twice with different locales?

What I found about doing is to make two DB queries instead of one. The first query will do the search with the user's locale so that we can show documents in the user's language. The second query will be executed only if the first one returned less results than the specified limit and will do the search with the root locale. The documents found in the second query will be added only if they haven't been found in the user's language before and without exceeding the limit.
The problem with this solution is that I don't thinks we can change that code to make it happen:
https://github.com/xwiki/xwiki-platform/blob/b4095dbabf797c0cbb176b3625e29ce1a5b27d20/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java#L165-L173

atallahade:
Thanks for the clear answer @mflorea, I'll work on that.

atallahade:
@mflorea I've made the changes based on your proposal:
> For a current locale such as `fr_CA` ideally we should look with priority in `fr_CA`, `fr`, default locale, other locales.

I just thought that searching pages with "other locales" might not be the best thing to do, WDYT? IMO, we should only look for pages based on the user locale + the default one (especially in case the user is using the page full name).

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I'm wondering if it's maybe possible to achieve what I've done with a single query using the `CASE` expression. The idea would be to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)

I'll investigate.

atallahade:
I've improved the change by making a single query using the `CASE` expression. The idea is to sort the results in this order:
1. The locale (e.g. `fr_CA`)
2. The language (e.g. `fr`)
3. The default locale (`""`)
4. Other locales

Then we just have to take the result one by one and add only to our list of result the first document we encounter (to return the "best" translated document).

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseSearchResult.java
```diff
@@ -212,84 +238,133 @@
 
             // Add ordering
             f.format(") order by %s", orderClause);
+            String queryString = f.toString();
 
-            String query = f.toString();
+            Locale userLocale = contextProvider.get().getLocale();
```
mflorea: You should use [``LocalizationContext#getCurrentLocale()``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-api/src/main/java/org/xwiki/localization/LocalizationContext.java#L38) instead, if that's the only reason you access the XWiki context.

atallahade: Thanks I was looking for something like this.

PR #821, opened 2018-08-28
XWIKI-15522: Space missing between icon and text when using Search Suggest

CostiChD:


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/search/searchSuggest.js
```diff
@@ -161,7 +161,7 @@ var XWiki = (function (XWiki) {
         sources : this.sources,
         insertBeforeSuggestions : new Element("div", {'class' : 'results'}).update( allResults ),
         displayValue: true,
-        displayValueText: "$escapetool.javascript($services.localization.render('platform.search.suggestResultLocatedIn'))",
+        displayValueText: "$escapetool.javascript("$services.localization.render('platform.search.suggestResultLocatedIn') ")",
```
mflorea: There are two problems with this change:
* we can't minify ``searchSuggest.js`` at build time anymore because the JavaScript code is invalid without the Velocity evaluation.
* the spacing should be done from CSS

CostiChD: Working on it.

PR #825, opened 2018-08-29
XWIKI-15452: Use auto-suggestion on xproperties that can support it

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15452

### Changes

* Add methods to create page fields.
* Use a Page field on the Skin property of the XWikiUsers class.

### Note

We might need another picker to suggest attachments on the Avatar field.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
@@ -988,6 +988,92 @@ public class BaseClass extends BaseCollection<DocumentReference> implements Clas
         return false;
     }
 
+    /**
+     * Adds a page field to the class.
+     * <p>
+     * The input has no multiselect by default.
+     *
+     * @param fieldName the field name
+     * @param fieldPrettyName the shown name
+     * @param size size of the input
+     * @return true if the field has been added
+     * @since 10.8RC1
+     */
+    public boolean addPageField(String fieldName, String fieldPrettyName, int size)
+    {
+        return addPageField(fieldName, fieldPrettyName, size, false);
+    }
+
+    /**
+     * Adds a page field to the class.
+     *
+     * @param fieldName the field name
+     * @param fieldPrettyName the shown name
+     * @param size size of the input
+     * @param multiSelect specifies if there can be several values
+     * @return true if the field has been added
+     * @since 10.8RC1
+     */
+    public boolean addPageField(String fieldName, String fieldPrettyName, int size, boolean multiSelect)
+    {
+        return addPageField(fieldName, fieldPrettyName, size, multiSelect, multiSelect, "");
+    }
+
+    /**
+     * Adds a page field to the class.
+     *
+     * @param fieldName the field name
+     * @param fieldPrettyName the shown name
+     * @param size size of the input
+     * @param multiSelect specifies if there can be several values
+     * @param relationalStorage specifies how the values are being stored
+     * @param sql the optional sql query used for the autosuggest
+     * @return true if the field has been added
+     * @since 10.8RC1
+     */
+    public boolean addPageField(String fieldName, String fieldPrettyName, int size, boolean multiSelect,
+        boolean relationalStorage, String sql)
+    {
+        return addPageField(fieldName, fieldPrettyName, size, multiSelect, relationalStorage, sql,
+            ListClass.DISPLAYTYPE_INPUT, true);
+    }
+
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
```diff
@@ -988,6 +988,92 @@ public boolean addDBTreeListField(String fieldName, String fieldPrettyName, int
...
+    }
+
+    /**
+     * Adds a page field to the class.
+     *
+     * @param fieldName the field name
+     * @param fieldPrettyName the shown name
+     * @param size size of the input
+     * @param multiSelect specifies if there can be several values
+     * @param relationalStorage specifies how the values are being stored
```
vmassol: I would explain what happens when "true" is passed instead since a reader will not understand what "true" or "false" means.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
```diff
@@ -988,6 +988,92 @@ public boolean addDBTreeListField(String fieldName, String fieldPrettyName, int
...
+     * @param sql the optional sql query used for the autosuggest
```
vmassol: Suggestion: use `autoSuggestSQL`  as the parameter name instead of 'sql'. Is it really SQL or HQL?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
```diff
@@ -988,6 +988,92 @@ public boolean addDBTreeListField(String fieldName, String fieldPrettyName, int
...
+     * @param displayType the display type
```
vmassol: Would be good to explain the possible values or at least provide an example.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
mflorea: It's HQL and  ``autoSuggestSQL`` is not good, and neither the parameter description (I missed that). The HQL is used to get the list of allowed values. The meta property name is ``sql`` though so using something else can confuse the developers that know the Database List meta properties.

atallahade: I'll let `sql` as a variable name because that's the name of the property but change the comment to; `the optional HQL query used for the autosuggest`.

vmassol: I don't understand the relationship between an xproperty name and a java method property name, especially since you don't document that! ;)

If you were saying "this value correspond to the value of the "sql" xproperty for the XXX xclass metadata" then I'd agree more, even though I don't think we need to have the same naming rules for both.

vmassol: Said differently how is the user supposed to know that this parameter has the same value as the xproperty metdata for Database List?

Isn't this an implementation detail BTW?

vmassol: So for me this is an implementation detail and `sql` is not a great name since 1) it's not SQL, 2) it's not obvious what the query is about without reading the javadoc

atallahade: I agree that it's a bit implicit.

vmassol: Now this is really a small detail.... More generally we have tons of javadoc, especially in oldcore that has no javadoc at all... It's just that it would be nice that we fix that, at least by making sure that the added methods provides the best possible javadoc.

atallahade: Better spend time now trying to make the best possible javadoc. It probably take less time in the future.

PR #833, opened 2018-08-30
XWIKI-15493: stop_xwiki.sh does not wait long enough for clean exit

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15493

### Changes

* Add an optional argument to the stop_xwiki.sh script to define the number of second to wait for the XWiki lock before exiting.
* Set the default wait time to 30sec.


vmassol:
Sounds good to me. Thx

PR #836, opened 2018-09-04
TFD: Add mutation threshold for some modules of platform-core.

surli:
Oldcore has been skipped from mutation testing for now because it takes too much time.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-oldcore/pom.xml
@@ -624,6 +624,14 @@
   </dependencies>
   <build>
     <plugins>
+      <!-- Skipped for now as it really take very long time -->
+      <plugin>
+        <groupId>org.pitest</groupId>
+        <artifactId>pitest-maven</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
       <!-- Publish a Core Test JAR -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-python/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-python/pom.xml
@@ -34,6 +34,9 @@
     <xwiki.jacoco.instructionRatio>1.00</xwiki.jacoco.instructionRatio>
     <!-- Name to display by the Extension Manager -->
     <xwiki.extension.name>Python Macro</xwiki.extension.name>
+
+    <!-- Error when launching pit-test: no mutation found -->
+    <xwiki.pitest.skip>true</xwiki.pitest.skip>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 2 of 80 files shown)


xwiki-platform-core/xwiki-platform-oldcore/pom.xml
```diff
@@ -624,6 +624,14 @@
   </dependencies>
   <build>
     <plugins>
+      <!-- Skipped for now as it really take very long time -->
```
vmassol: Do you know how long it takes or you stopped it before the end? :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-python/pom.xml
```diff
@@ -60,4 +63,16 @@
...
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <!-- Skip pit/descartes for now as it's failing on this module, see: https://github.com/STAMP-project/pitest-descartes/issues/72 -->
+      <plugin>
+      <groupId>org.pitest</groupId>
+      <artifactId>pitest-maven</artifactId>
+      <configuration>
+        <skip>true</skip>
```
vmassol: Do we need this even with `<xwiki.pitest.skip>true</xwiki.pitest.skip>` above?

Also shouldn't this be in the quality profile only?

vmassol:
Thanks @surli Now we just need to add a CI job. I can do it, unless you want to do it.

xwiki-platform-core/xwiki-platform-oldcore/pom.xml
surli: No, actually I indeed stopped it before the end.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-python/pom.xml
surli: My bad, I forgot to remove the property. The property is not working as we override it by passing the argument in command line when executing pitest.

surli: > Also shouldn't this be in the quality profile only?

Not sure: actually we should always avoid launching pit/descartes on this module as it's failing, no matter if it's because we are in quality profile or for another reason.

surli:
> we just need to add a CI job. I can do it, unless you want to do it.

Actually it might be interesting for me to do it, even if I take much more time than you ;)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-python/pom.xml
vmassol: Except that right now, all the pitest/descartes definitions/configurations are only present in the quality profile in the top level pom.

vmassol:
> Actually it might be interesting for me to do it, even if I take much more time than you ;)

Perfect. I agree. I can help ofc.

PR #839, opened 2018-09-06
XWIKI-15543: Create a proper copyright file for debian distribution

surli:


tmortagne:
Well this is going to work for one package the but the license is still missing in all the others. Also you should remove the wrong one I inserted in control.

xwiki-platform-distribution/xwiki-platform-distribution-debian/xwiki-platform-distribution-debian-common/pom.xml
```diff
@@ -51,7 +51,36 @@
...
+        <filtering>true</filtering>
+        <targetPath>${project.build.directory}/deb</targetPath>
+      </resource>
+    </resources>
     <plugins>
+      <plugin>
+        <!-- Get the current year -->
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <version>1.12</version>
```
tmortagne: This plugins version is already defined in xwiki-commons root pom (like most plugins we use).

xwiki-platform-distribution/xwiki-platform-distribution-debian/xwiki-platform-distribution-debian-common/pom.xml
```diff
@@ -51,6 +51,16 @@
     </dependency>
   </dependencies>
   <build>
+    <resources>
```
tmortagne: Have you tried to put that in the parent pom instead of duplicating it ?

tmortagne:
Would be much easier to maintain if the `copyright` template was in some resource module and then get it using the xwiki-commons-tool-remote-resource-plugin plugin like we do for NOTICE/LICENSE files.

surli:
@tmortagne I now use the `xwiki-commons-tool-remote-resource-plugin` however I did not manage to use the value `${current.year}` defined during the validate of the parent pom for the copyright current year. Any idea to fix that?

tmortagne:
Do you see build-helper-maven-plugin executed before xwiki-commons-tool-remote-resource-plugin in the Maven log ?

xwiki-platform-distribution/xwiki-platform-distribution-debian/pom.xml
```diff
@@ -69,6 +69,47 @@
...
+            <artifactId>xwiki-commons-tool-remote-resource-plugin</artifactId>
+            <executions>
+                <execution>
+                    <id>xwiki-platform-tool-configuration-resources</id>
+                    <phase>process-resources</phase>
+
+                    <configuration>
+                        <resourceBundles>
+                            <!-- Using ${platform.version} on purpose here. An XWiki distribution with a different version might inherit this. -->
+                            <resourceBundle>org.xwiki.platform:xwiki-platform-tool-configuration-resources:${platform.version}</resourceBundle>
```
tmortagne: Why xwiki-platform-tool-configuration-resources ?

surli: It's defined in the parent pom but overridden if I specify a new resourceBundle here, so I need to specify them both. This one is used for generating the hibernate.cfg and other config files.

surli: it looks pretty obvious now you tell me... thanks!

surli: Finally it's not working: if I rename this one, only the parent one is executed, and this one remains overridden. Apparently it's not linked to the name: tried to rename them xwiki-1 and xwiki-2 and only "xwiki-1" appears in my console.

surli:
> Do you see build-helper-maven-plugin executed before xwiki-commons-tool-remote-resource-plugin in the Maven log ?

this one's fixed.

xwiki-platform-distribution/xwiki-platform-distribution-debian/pom.xml
```diff
@@ -69,6 +69,47 @@
     </extensions>
     <pluginManagement>
       <plugins>
+        <plugin>
```
tmortagne: I think you should put that in `<build>` instead of `<pluginManagement>` since you want to execute it for all children (right ?), it would avoid having to do it in each of those children.

xwiki-platform-distribution/xwiki-platform-distribution-debian/pom.xml
```diff
@@ -69,6 +69,47 @@
...
+                  </goals>
+                  <phase>validate</phase>
+                  <configuration>
+                      <name>current.year</name>
+                      <pattern>yyyy</pattern>
+                  </configuration>
+              </execution>
+          </executions>
+        </plugin>
+        <plugin>
```
tmortagne: Same comment as above.

surli: Done.

xwiki-platform-distribution/xwiki-platform-distribution-debian/xwiki-platform-distribution-debian-common/pom.xml
```diff
@@ -130,6 +130,17 @@
                 <prefix>/usr/lib/xwiki/META-INF</prefix>
               </mapper>
             </data>
+            <!-- Copyright file -->
+            <data>
```
tmortagne: Hmm I think it might work to define this one in the parent and then use `combine.children="append"` in each child `<dataSet>`.

tmortagne:
Thanks you for your contribution (and your patience) !

surli:
Thanks for you help!

PR #840, opened 2018-09-06
XWIKI-13101: Removed deprecated "Panels.SpaceDocs", "Panels.Spaces", "Main.Spaces" and "Main.SpaceIndex"

evalica:
See https://jira.xwiki.org/browse/XWIKI-13101


evalica:
Currently I don't plan to create a contrib extension for these pages. There are some mails threads related to how we could deprecate these panels / macros (see http://markmail.org/message/jkqz3k67n6xb5kfh) and my impression was that people were ok if we delete the pages after a time. Currently I've provided just a XAR for the deleted pages, see https://jira.xwiki.org/secure/attachment/35871/Legacy.xar

I didn't deleted the Index/Translations page since there are other pages in the app, like the Page Index that could move / add some translations there.

vmassol:
@evalica If you have a XAR then it wouldn't be that much to commit the sources in a contrib repo and thus make it an extension.

tmortagne:
> my impression was that people were ok if we delete the pages after a time

Well my impression reading those mail is that there is more people in favor of moving the pages to an extension :) As Vincent said it's very easy.

vmassol:
> Q1: Should the extension be in a git repo? or just a XAR in exo?

Better in a git repo

> Q2: Should the pages be installable individually or as a group?

I think group is good enough.

tmortagne:
> Q1: Should the extension be in a git repo? or just a XAR in exo?

Standard contrib repository but without any jira since we don't really plan to maintain it.

> Q2: Should the pages be installable individually or as a group?

Just put all of them in one spaces related extension.

PR #844, opened 2018-09-10
XWIKI-15572: Use auto-suggestion on the WatchListClass class

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15572

### Changes

* Use a Page field on the Page list property of the WatchListClass class.
* Use a User field on the User list property of the WatchListClass class.

### Note

This change was taken from #803.

Is there something special to do when using multiselect xproperties that I've missed? The change seems to work well but can have side effects I'm not aware of.


tmortagne:
It's not only a migration issue, it also break all the requests which manipulate those fields.

atallahade:
@tmortagne There are other places where we need multiselect pages / users fields, I wanted to do all the necessary it in here first so I can reproduce the same process on other xclasses.

atallahade:
Won't fix.

PR #849, opened 2018-09-11
misc: Fix pseudo-tested method in xwiki-platform-resource-default

surli:
* Refactor one test to use JUnit 5
 * Add a more exhaustive test on MainResourceReferenceHandlerManager
 * Extract a method and add a test for DefaultEntityResourceActionLister


xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-default/src/main/java/org/xwiki/resource/internal/entity/DefaultEntityResourceActionLister.java
```diff
@@ -71,30 +71,36 @@
     @Named("context")
     private Provider<ComponentManager> contextComponentManagerProvider;
 
-    @Override
-    public void initialize() throws InitializationException
+    protected SAXBuilder createSAXBuilder()
```
vmassol: I'm surprised that checkstyle doesn't complain about the missing javadoc

PR #855, opened 2018-09-14
XWIKI-15444: Fix export UI

surli:
## Scope of the PR

This PR aims at fixing issue XWIKI-15444, specifically 3 different issues have been listed:
 1. The export is not working when "More pages" pagination node is selected,
 2. The export is not working when lot of pages are selected,
 3. The export does not take into account the hidden pages

This PR targets all issues.

## Design

As for issues 1 and 2, the main change of the PR is to support in the Export API excluded pages, and to relate them to an included pages. It is then possible to build complex requests by properly ordering the argument: 

```
pages[0] = Foo.% // everything under Foo will be exported...
excludes[0] = Foo.Bar&Foo.Baz // ... except Bar and Baz pages

// the 3 following pages are included individually without excludes
pages[1] = Bar.WebHome
excludes[1] = 

pages[2] = Bar.Baz
excludes[2] = 

pages[3] = Bar.Foo
excludes[3] = 

pages[4] = Baz.% // everything under Baz is exported
excludes[4] = Baz.Bar.% // except all children of Baz.Bar including WebHome
```

Note that the excludes parameter is optional to respect backward compatibility of the API, however whenever it's used, it should always be accompanied by a pages argument, e.g.: 

```
// old API, still ok
pages[0] = Bar.%
pages[1] = Bar.Baz
pages[2] = Bar.Foo

// new API: OK
pages[0] = Bar.%
excludes[0] = Bar.Bar

pages[1] = Bar.Baz
excludes[1] = 

pages[2] = Bar.Foo
excludes[2] = 

// new API OK
pages[0] = Bar.%
excludes[0] = Bar.Bar

pages[1] = Bar.Baz
pages[2] = Bar.Foo

// new API NOT OK -> will throw an exception
pages[0] = Bar.%
excludes[0] = Bar.Bar
excludes[1] = Bar.Baz
``` 

Then the javascript associated to the tree modal as been rewritten to create proper hidden inputs on the fly to satisfy this new API given what has been selected or not (including choice made for pagination nodes).

Finally, regarding issue 3, the document tree is now displaying hidden files no matter what are the user preferences.


xwiki-platform-core/xwiki-platform-oldcore/pom.xml
```diff
@@ -399,6 +399,21 @@
...
+    <!-- ExportAction -->
+
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-tree-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-index-tree-api</artifactId>
```
tmortagne: You cannot add this dependency since it depends on oldcore. You will need to mock the implementation you need in it.

surli: Even when it's a test dependency?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -268,8 +337,26 @@ private String exportXAR(XWikiContext context) throws XWikiException, IOExceptio
...
         String name = request.get("name");
-        String[] pages = request.getParameterValues("pages");
-        boolean all = ArrayUtils.isEmpty(pages);
+        String[] checkedPages = request.getParameterValues("pages");
+        String[] uncheckedPages = null;
+        String checkedPagesRequestValue = request.get("checked-pages");
+        String uncheckedPagesRequestValue = request.get("unchecked-pages");
+        String otherPagesRequestValue = request.get("other-pages");
+
+        boolean all = ArrayUtils.isEmpty(checkedPages) && StringUtils.isEmpty(otherPagesRequestValue) && StringUtils.isEmpty(checkedPagesRequestValue);
```
atallahade: This line is exceeding [120 characters](https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml#L214).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -268,8 +337,26 @@ private String exportXAR(XWikiContext context) throws XWikiException, IOExceptio
...
+
+        boolean otherPages = false;
+        if (!StringUtils.isEmpty(otherPagesRequestValue)) {
```
atallahade: Same comment as above.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -114,7 +119,26 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
 
-        Collection<DocumentReference> pageList = resolvePagesToExport(request.getParameterValues("pages"), context);
+        String[] checkedPages = request.getParameterValues("pages");
+        String[] uncheckedPages = null;
+        String checkedPagesRequestValue = request.get("checked-pages");
+        String uncheckedPagesRequestValue = request.get("unchecked-pages");
+        String otherPagesRequestValue = request.get("other-pages");
+
+        boolean otherPages = false;
+        if (!StringUtils.isEmpty(otherPagesRequestValue)) {
```
atallahade: From the javadoc `Boolean valueOf(String s)`:
> Returns a Boolean with a value represented by the specified string. The Boolean returned represents a true value if the string argument is not null and is equal, ignoring case, to the string "true".

So you might not need this line.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+          $(treeReference.get_json(tree, {
+              flat: true
+          })).each( function () {
+            var node = treeReference.get_node(this.id);
+
+            // the node is checked
+            if (treeReference.is_checked(node)) {
+                // if the node is pagination then we should set the otherPages attribute
+                if (node.data.type == "pagination") {
+                    otherPages.attr('value', 'true');
```
mflorea: The recommended way to set the value is through `val()``.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+                                pageReference.parent);
+                            checkedPages.push(XWiki.Model.serialize(webPreferencesReference));
+                        }
+                    }
+                }
+            // the node is unchecked
+            } else {
+                // if the node is pagination then we should set the otherPages attribute
+                if (node.data.type == "pagination") {
+                    otherPages.attr('value', 'false');
```
mflorea: Same as above. Any reason for not using ``val()``?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+
+            // the node is checked
+            if (treeReference.is_checked(node)) {
+                // if the node is pagination then we should set the otherPages attribute
+                if (node.data.type == "pagination") {
+                    otherPages.attr('value', 'true');
+                } else {
+                    var pageReference = XWiki.Model.resolve(node.data.id, XWiki.EntityType.DOCUMENT);
+                    checkedPages.push(XWiki.Model.serialize(pageReference));
+                    // In case the page coudl have children
```
mflorea: Typo.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+                }
+            // the node is unchecked
+            } else {
+                // if the node is pagination then we should set the otherPages attribute
+                if (node.data.type == "pagination") {
+                    otherPages.attr('value', 'false');
+                } else {
+                    var pageReference = XWiki.Model.resolve(node.data.id, XWiki.EntityType.DOCUMENT);
+                    uncheckedPages.push(XWiki.Model.serialize(pageReference));
+                    // In case the page coudl have children
```
mflorea: Typo.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+                    if (pageReference.getName() == 'WebHome') {
+                        // The node could be checked but not loaded
```
mflorea: > // The node could be **un**checked but not loaded

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+                            var spaceReference = new XWiki.EntityReference('%', XWiki.EntityType.DOCUMENT,
+                                pageReference.parent);
+                            uncheckedPages.push(XWiki.Model.serialize(spaceReference));
+                        }
+                    }
+                }
             }
+          });
+
+          var aggregatePageNames = function (arrayOfNames) {
```
mflorea: You can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join instead.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -176,36 +180,90 @@ require([
...
+
+                  if (i < arrayOfNames.length - 1) {
+                      finalValue += "&";
+                  }
+              }
+              return finalValue;
+          };
+
+          checkedPagesInput.attr('value', aggregatePageNames(checkedPages));
+          uncheckedPagesInput.attr('value', aggregatePageNames(uncheckedPages));
```
mflorea: Again, better use ``val()``. What happens if a page name contains ``&``. Any character is valid within the page name.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
mflorea: Yes, you can assign directly ``Boolean.valueOf()``.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -114,7 +119,26 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
+            otherPages = Boolean.valueOf(otherPagesRequestValue);
+        }
+
+        if (!StringUtils.isEmpty(uncheckedPagesRequestValue)) {
+            uncheckedPages = uncheckedPagesRequestValue.split("&");
```
mflorea: You make the assumption that page names can't contain ``&``, but they can.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -114,7 +119,26 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
+        }
+
+        if (!StringUtils.isEmpty(checkedPagesRequestValue)) {
+            checkedPages = checkedPagesRequestValue.split("&");
```
mflorea: Same here, ``&`` is valid and not that uncommon in page names.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -179,7 +202,7 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
             }
 
             DocumentReferenceResolver<String> resolver =
-                Utils.getComponent(DocumentReferenceResolver.TYPE_STRING, "current");
+                    Utils.getComponent(DocumentReferenceResolver.TYPE_STRING, "current");
```
mflorea: I assume this is the result of the automatic formatter using the XWiki code style. It strange that it uses 8 spaces to indent.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -197,23 +220,69 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
             } finally {
                 context.setWikiId(database);
             }
         }
-
         return pageList;
     }
 
+    /**
+     * Resolve the set of pages to export based on choices made in a paginated tree
```
mflorea: The XWiki code style leaves an empty line here. Did you configure your IDE to use the XWiki code style?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -197,23 +220,69 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
+     */
+    private Collection<DocumentReference> resolvePagesToExport(boolean otherPages, String[] checkedPages, String[] uncheckedPages, XWikiContext context)
+        throws XWikiException
+    {
+        if (otherPages) {
+            // if other pages is selected, we select all but the unchecked pages
+            XWikiDocument doc = context.getDoc();
+
+            // use the document tree to get all the pages
+            Tree nestedpages = Utils.getComponent(Tree.class, "nestedPages");
```
mflorea: Camel case makes the variable names easier to read.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -197,23 +220,69 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
+            List<String> children = nestedpages.getChildren(nodeId, 0, childCount);
+
+            // result as obtained as nodeId with "document:" as prefix
+            List<String> formattedReferences = new ArrayList<>();
+            for (String child : children) {
+                int index = "document:".length();
+                formattedReferences.add(child.substring(index));
+            }
+
+            // we need to add the current document also
```
mflorea: Why?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -197,23 +220,69 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
...
+    {
+        if (otherPages) {
+            // if other pages is selected, we select all but the unchecked pages
+            XWikiDocument doc = context.getDoc();
+
+            // use the document tree to get all the pages
+            Tree nestedpages = Utils.getComponent(Tree.class, "nestedPages");
+            String nodeId = "document:" + doc.getDocumentReference().toString();
+            int childCount = nestedpages.getChildCount(nodeId);
+            List<String> children = nestedpages.getChildren(nodeId, 0, childCount);
```
mflorea: This gives you the direct child nodes. How do you include all the descendant child nodes? (I mean all the sub-tree).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -268,8 +337,26 @@ private String exportXAR(XWikiContext context) throws XWikiException, IOExceptio
...
+            otherPages = Boolean.valueOf(otherPagesRequestValue);
+        }
+
+        if (!StringUtils.isEmpty(uncheckedPagesRequestValue)) {
+            uncheckedPages = uncheckedPagesRequestValue.split("&");
+        }
+
+        if (!StringUtils.isEmpty(checkedPagesRequestValue)) {
+            checkedPages = checkedPagesRequestValue.split("&");
+        }
```
mflorea: The above lines are duplicated in the ``render()`` method.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
surli: I suppose I don't right now: I did not check this scenario yet.
So 1. do you know if there's any way to get the descendant recursively from the tree API? 
   2. according to you, what is the minimum scenario to need an export with subtrees: defining a subwiki?

surli:
Thanks @mflorea @atallahade for the review: I'll check my checkstyle config in the IDE, and try to add the corner-cases you mentioned in my test (page with `&` in the name, subtrees).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
mflorea: I think the reason is that currently you can't select "other pages" without selecting the parent page.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
mflorea: Yes, that should work.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -60,40 +60,46 @@
...
-            ## When nested page also export WebPreferences which contain among other things the right of this entity
-            #set($exportURL = "$exportURL&amp;pages=${escapetool.url($services.model.serialize($doc.documentReference.parent,'default'))}.WebPreferences")
-          #end
-          <a href="$doc.getURL('export', "$exportURL&amp;$!docvariant")" class="btn btn-primary" rel="nofollow">$services.localization.render('core.menu.export.xar')</a>
-        #end
-        #foreach ($exportExtension in $exportExtensions)
-          $services.rendering.render($exportExtension.execute(), 'html/5.0')
-        #end
+          <form method="post" id="export-form">
+              <input type="hidden" id="checked-pages" name="checked-pages" />
```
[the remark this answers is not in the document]
surli: Just checked and I'm using the XWiki profile in the IDE but apparently it did not set the right indent for HTML, CSS, JS & Velocity.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
surli: OK I'll check what happens with nested pages then.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -114,7 +119,26 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
             name = context.getDoc().getFullName();
         }
 
-        Collection<DocumentReference> pageList = resolvePagesToExport(request.getParameterValues("pages"), context);
+        String[] checkedPages = request.getParameterValues("pages");
+        String[] uncheckedPages = null;
+        String checkedPagesRequestValue = request.get("checked-pages");
```
tmortagne: If this data is supposed to be a list why invent your own syntax instead of using standard HTTP one and simply use getParameterValues like we do for pages ?

surli: Because here the values are coming from a POST request: it would involve to create as many input field as I have pages checked/unchecked. 
Here it seems cleaner to me to only have 2 input fields and to parse their value: then I need to do it manually.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
surli: Indeed there is a problem with nested pages, but I cannot process them as done in the javascript: in the JS we know which nodes are not loaded, so we can mark them with a joker to mark the subtree to be exported.
Here basically I need to process the nodes that are hidden: everyone of them might contain a subtree: so I would need to add the joker to everyone of them.

surli:
@mflorea this is ready for review again: I still have one test to add to check subtress. I already checked it manually so from my point of view the design won't change now.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -94,6 +111,43 @@ public String render(XWikiContext context) throws XWikiException
...
+            this.name = request.get("name");
+            if (StringUtils.isBlank(name)) {
+                this.name = context.getDoc().getFullName();
+            }
+
+            String[] pagesParam = request.getParameterValues("pages");
+            String checkedPagesParam = request.get("checked-pages");
+
+            if (!StringUtils.isEmpty(checkedPagesParam)) {
+                this.checkedPages = checkedPagesParam.split("&");
```
tmortagne: As @mflorea pointed out in a previous comment this code seems to be impossible to have a page name containing `&` in that list. If you insist on inventing your own syntax you will need escaping support in this syntax because in XWiki there is no forbidden characters in names.

surli: The names of pages are encoded in the exporter.js, see: https://github.com/xwiki/xwiki-platform/pull/855/files#diff-732756cd30b04d597a99718c625c573aR246

So the "&" which might exist in the filename are already encoded.
Now as I answered in your previous comment I cannot use the `getParamValues` as it would involve create as many input fields as I have checked pages: if you have any other advice to not use this syntax, I take it.

tmortagne: The code would be much easier to understand if the decode was done here. So that checkedPages actually contain a list of pages :)

surli: Indeed. I'll fix that.

surli:
@tmortagne as we discussed the design yesterday, could you review this one? Thanks.

surli:
@tmortagne I need to manage the hidden pages but you can already review it.

surli:
Hidden pages are now managed. Please read the description of the PR: it has been edited to explain the changes.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -70,6 +79,10 @@
  */
 public class ExportAction extends XWikiAction
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExportAction.class);
```
tmortagne: Does not seems to be used.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -70,6 +79,10 @@
...
+    private static final String TREE_DOCUMENT_PREFIX = "document:";
```
tmortagne: Same.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -136,52 +225,100 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
         return null;
     }
 
-    private Collection<DocumentReference> resolvePagesToExport(String[] pages, XWikiContext context)
+    private String extractWikiName(String pattern, XWikiContext context)
+    {
+        String wikiName;
+        if (pattern.contains(":")) {
```
tmortagne: You should do the if on `index` instead of parsing the String twice (only substring if index > 0).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -195,11 +332,11 @@ private String exportHTML(XWikiContext context) throws XWikiException, IOExcepti
                     List<String> docsNames = dbQuery.setWiki(wikiName).bindValues(params).execute();
 
                     for (String docName : docsNames) {
-                        String pageReference = wikiName + XWikiDocument.DB_SPACE_SEP + docName;
-                        if (context.getWiki().getRightService().hasAccessLevel(
-                            "view", context.getUser(), pageReference, context))
-                        {
-                            pageList.add(resolver.resolve(pageReference));
+                        String pageReferenceStr = wikiName + XWikiDocument.DB_SPACE_SEP + docName;
```
tmortagne: Better pass a WikiReference as second parameter of the resolver (should always avoid directly manipulate reference syntax).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -4,70 +4,82 @@
...
+     aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+        <h4 class="modal-title" id="exportModalLabel">$services.localization.render('core.menu.export')</h4>
+      </div>
+      <div class="modal-body">
+        <div class="panel-group" id="exportModelAccordion" role="tablist" aria-multiselectable="true">
+          #set($exportExtensions = $services.uix.getExtensions('org.xwiki.plaftorm.menu.export.buttons'))
```
mflorea: The HTML was already nicely indented. I don't understand why you removed the indentation.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -4,70 +4,82 @@
...
+         class="btn btn-primary" rel="nofollow">$services.localization.render('core.menu.export.pdf')</a>
+      #if("$!{services.officemanager.serverState}" == 'Connected')
+        <a href="$doc.getURL('export', "format=odt&amp;$!escapetool.html($request.getQueryString())")"
+           class="btn btn-primary" rel="nofollow">$services.localization.render('core.menu.export.odt')</a>
+        <a href="$doc.getURL('export', "format=rtf&amp;$!escapetool.html($request.getQueryString())")"
+           class="btn btn-primary" rel="nofollow">$services.localization.render('core.menu.export.rtf')</a>
+      #end
+      #foreach ($exportExtension in $exportExtensions)
+        #if ($exportExtension.parameters.category == 'office')
+          $services.rendering.render($exportExtension.execute(), 'html/5.0')
```
mflorea: Same here. Couldn't you fix the long lines without affecting the indentation of the other lines? It also makes the diff harder to read.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -4,70 +4,82 @@
...
   <div class="panel panel-default">
-    <div class="panel-heading" role="tab" id="exportModalHeadingOther">
-      <h4 class="panel-title">
-        <a class="collapsed" role="button" data-toggle="collapse" data-parent="#exportModelAccordion" href="#exportModelOtherCollapse" aria-expanded="false" aria-controls="exportModelOtherCollapse">
-          $escapetool.xml($services.localization.render('core.exporter.headings.otherFormats'))
-        </a>
-      </h4>
-    </div>
-    <div id="exportModelOtherCollapse" class="panel-collapse collapse" role="tabpanel" aria-labelledby="exportModalHeadingOther">
-      <div class="panel-body">
```
mflorea: Same. I think the previous indentation was good.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -4,70 +4,82 @@
...
           #set ($discard = $xwiki.jsfx.use('uicomponents/exporter/exporter.js', {'forceSkinAction': true}))
           <div class="row">
             <div class="col-xs-6">
               <p>$escapetool.xml($services.localization.render('core.exporter.selectPages'))</p>
             </div>
             <div class="col-xs-6">
-              <p class="text-right"><a href="#" class="tree_select_all">$escapetool.xml($services.localization.render('core.exporter.selectAll'))</a> / <a href="#" class="tree_select_none">$escapetool.xml($services.localization.render('core.exporter.selectNone'))</a></p>
+              <p class="text-right"><a href="#" class="tree_select_all">$escapetool.xml($services.localization.render(
+                'core.exporter.selectAll'))</a> / <a href="#"
+                                                     class="tree_select_none">$escapetool.xml($services.localization.render(
```
mflorea: Strange indentation.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -78,23 +90,23 @@
...
               'showAttachments': false,
               'showTranslations': false,
-              'readOnly': false
+              'readOnly': false,
+              'filterHiddenDocuments': false
             })
           </div>
         #end
-        <a href="$doc.getURL('export', "format=html&amp;$!escapetool.html($request.getQueryString())")" class="btn btn-primary" rel="nofollow">$services.localization.render('core.menu.export.html')</a>
+        <a href="$doc.getURL('export', "format=html")" class="btn btn-primary"
```
mflorea: You dropped the query string from the current URL. Where did it go? You don't need it at all?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
```diff
@@ -78,23 +90,23 @@
...
+           rel="nofollow">$services.localization.render('core.menu.export.html')</a>
         #if($hasAdmin && $services.xar.isXARExportAvailable())
-          #set($exportURL = "format=xar&amp;name=$escapetool.url($doc.fullName)&amp;pages=$escapetool.url($services.model.serialize($doc.documentReference,'default'))")
-          #if ($doc.name == 'WebHome')
-            ## When nested page also export WebPreferences which contain among other things the right of this entity
-            #set($exportURL = "$exportURL&amp;pages=${escapetool.url($services.model.serialize($doc.documentReference.parent,'default'))}.WebPreferences")
-          #end
-          <a href="$doc.getURL('export', "$exportURL&amp;$!docvariant")" class="btn btn-primary" rel="nofollow">$services.localization.render('core.menu.export.xar')</a>
+          #set($exportURL = "format=xar&amp;name=$escapetool.url($doc.fullName)")
```
mflorea: You dropped the ``$!docvariant``. Was it not needed?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
surli: Regarding all indentation problem: once I fixed my IDE I used the "format code" which completely mess up the diff. My bad.
I'll use the first version of the file and apply back my changes to fix that.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
surli: No it's now completely computed through the JS. But we might keep it, as the URL is rewritten in the JS, so I can push it back.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
surli: That's something I actually did not understood in the file: did not find any usage or declaration of this neither in the .vm or in the ExportAction API. If you have any clue I take it.

mflorea: * ``view.vm`` includes ``startpage.vm`` includes ``xwikivars.vm``
* ``view.vm`` includes ``contentview.vm`` includes ``contentheader.vm`` includes ``menus_content.vm`` includes ``export_modal.vm``

``$docvariant`` defined in ``xwikivars.vm`` is available in ``export_modal.vm``.

surli: Thanks! But then it's strange because it seems that XAR export in ExportAction does not handle revision or language and from what I see in `xwikivars` docvariant is used only for those two parameters.

surli: So should we keep it?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/export_modal.vm
surli: @mflorea here I just added the form around the div but apparently git has some trouble to process a good diff there...

mflorea: You could try

    <form id="export-form" method="post" class="panel-body">

i.e. convert the panel body DIV into a form instead of wrapping it. This way you don't need to change the indentation.

surli: Hm. Didn't thought about that. I'll try.

surli: Indeed it fixes the diff.

surli:
For me this one is finished @tmortagne @mflorea

tmortagne:
Fine with me (but did not really looked at client side).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-tests/src/test/it/org/xwiki/flamingo/test/ui/XARExportTest.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private void setupPages() throws Exception
+    {
+        getUtil().loginAsSuperAdmin();
+
+        // Create a space Foo
+        getUtil().createPage("Foo", "WebHome", "Foo", "Foo");
+
+        // Create 100 pages under that space (will be used for the test with lot of selected pages)
```
atallahade: Do you need that many pages? Each page being saved takes some time AFAIK, unless you can do it in batch or in a database session.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-tests/src/test/it/org/xwiki/flamingo/test/ui/XARExportTest.java
```diff
@@ -0,0 +1,249 @@
...
+       Don't change anything in the tree of export.
+     */
+    private void exportXARAll() throws Exception
+    {
+        getUtil().loginAsSuperAdmin();
+        ViewPage viewPage = getUtil().gotoPage("Foo", "WebHome");
+        viewPage.clickMoreActionsSubMenuEntry("tmExport");
+
+        ExportModal exportModal = new ExportModal();
+        //assertTrue(exportModal.isDisplayed());
```
atallahade: Leftover.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -70,6 +75,117 @@
...
+     * Manage the arguments of the export and provides them in an object to
+     * simplify their usage in the different methods.
+     *
+     * @since 10.9RC1
+     */
+    private class ExportArguments {
+        /**
+         * All patterns that have to be included in the request
+         *
+         * Warning: order of the arguments matters, it's linked with excludedPages
```
atallahade: In that case you can maybe use [`org.apache.commons.lang3.tuple.Pair`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html) or a HashMap?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -70,6 +75,117 @@
...
+            String[] pages = request.getParameterValues("pages");
+
+            String[] excludes = request.getParameterValues("excludes");
+
+            this.includedPages = Arrays.asList(pages);
+            this.excludedPages = this.extractArguments(excludes, context);
+
+            if (!this.excludedPages.isEmpty() && this.includedPages.size() < this.excludedPages.size()) {
+                throw new XWikiException(XWikiException.MODULE_XWIKI_APP, XWikiException.ERROR_XWIKI_APP_EXPORT,
+                    "You cannot more excludes argument than pages argument. See the export documentation.");
```
atallahade: Typo:
> You cannot **have** more excludes arguments than pages arguments

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -94,6 +210,8 @@ public String render(XWikiContext context) throws XWikiException
         return defaultPage;
     }
 
+
+
```
atallahade: Code style: you should remove this extra line.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/ExportActionTest.java
```diff
@@ -122,12 +142,109 @@ public void exportFullSpaceUsingWildcardsAsXAR() throws Exception
...
+        XWikiRequest request = mock(XWikiRequest.class);
+        when(request.get("format")).thenReturn("xar");
+        context.setRequest(request);
+
+        // Set other request parameters
+        when(request.get("name")).thenReturn("myexport");
+        when(request.getCharacterEncoding()).thenReturn("UTF-8");
+        // Export all pages in the "Space" space
+        when(request.getParameterValues("pages")).thenReturn(new String[] {"Space.%25", "Foo.%25", "Bar.Baz.%25"});
+        when(request.getParameterValues("excludes")).thenReturn(new String[] { "Space.Page1&Space.Page2", "", "Bar.Baz.WebHome" });
```
atallahade: Code style: line is too long.

surli:
@atallahade 
> The way excludes and pages arguments are handle is a bit complicated IMO. Wouldn't it be simpler to "untie" the two arguments (and give more freedom to the way they can be used), i.e. having a list of "excludes" independent of the "pages" list.

Looking for a counterexample but I cannot find any right now, so I guess it should work. WDYT @tmortagne?

tmortagne:
@surli as we discussed applying all the excludes to all the pages is a problem when the excludes are patterns. What @atallahade propose (which is what you did yourself in a previous version) only work if you forbid any pattern other than something ending with %.

atallahade:
@tmortagne Could you give an example? I'm not sure I understand why having an exclude pattern ending with `%` might cause problems.

tmortagne:
>  I'm not sure I understand why having an exclude pattern ending with % might cause problems

That's because I said the opposite :)

Imagine that you export two spaces but in one of them you want to exclude `%.Toto` (but only in that space). It's not a problem for patterns where the % (if any) is only at the end.

atallahade:
@tmortagne 
> Imagine that you export two spaces but in one of them you want to exclude %.Toto (but only in that space). It's not a problem for patterns where the % (if any) is only at the end.

If you want to exclude `%.Toto` in a space, why not write `Space.%.Toto`?

tmortagne:
> If you want to exclude %.Toto in a space, why not write Space.%.Toto?

This is not enough, you also need to exclude `Space.Toto`.

You are the one talking about simplicity but I find much simpler to tie excludes to pages because that's actually what you want to do 90% of the time (exclude stuff from a specific `pages` element) and you don't have to wonder how to craft something which is not going to mess with other `pages` elements that don't you care about.

atallahade:
> you don't have to wonder how to craft something which is not going to mess with other pages elements that don't you care about.

Good point, it might actually be a good trade off to restrict the excludes pattern to spaces over than having more freedom but also having to deal with tricky cases.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-tests/src/test/it/org/xwiki/flamingo/test/ui/XARExportTest.java
surli: I'm using the REST API here so it does not take so much time (the test were taking around 20 seconds, once launched).
I need lot of pages to check the behaviour with a long list of pages selected, and I also want to check the "more pages" node, but indeed 100 is quite a lot. I'll create 50.

surli:
@mflorea could you review the JS of this PR?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-tests/src/test/it/org/xwiki/flamingo/test/ui/XARExportTest.java
atallahade: It seems that you haven't updated the comment (and it's great that the pages creation is fast enough!).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -152,60 +154,188 @@ require([
...
+         * @param arrayOfPages: a list of list of pages to includes
+         * @param arrayOfExcludes: a list of list of pages to excludes relatively to their included pages
          */
-         $('#exportModal #exportModelOtherCollapse a.btn').click(function(event) {
-          var button = $(this);
-          var url = button.data('url').clone();
-          var checkedNodes = treeReference.get_checked(true);
-          if (url.params === undefined) {
-            url.params = {'pages': []};
+        var processLevel = function(rootNode, arrayOfPages, arrayOfExcludes) {
```
mflorea: If I understood correctly, the items from ``arrayOfExcludes`` are mapped to the items from ``arrayOfPages`` so it would have been nice to use a map data type to make this more explicit. Unfortunately [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) is not supported on older version of Internet Explorer and the plain JavaScript object has some reserved keys that theoretically could collide with the page references. In practice the page reference should contain at least one dot so there should be no collision. The only issues could be the page order, but I don't think it matters (Map preserves the key insertion order but we can't use it yet..)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -152,60 +154,188 @@ require([
...
+                excludedPages.push(subPageJoker);
+                nodeWithChildren.push(node);
               }
             }
+          });
+
+          // if we don't have a pagination node, or we have one and it's checked:
+          // we manage the export with exporting the space and excluding stuff that needs to be
+          if (!isPaginationExisting || (isPaginationExisting && isPaginationChecked)) {
+            includedPages = [ pageJoker ];
```
mflorea: What's the point of setting ``includedPages`` here? You don't use it after this.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
```diff
@@ -152,60 +154,188 @@ require([
...
+
         /**
-         * Update the URL of the export buttons according to the tree 
+         * This function is called recursively to process each parent node in order to built properly the list of list
+         * of pages to includes (i.e. pages parameter of export) and pages to excludes (i.e. excludes parameter of
+         * export).
+         * FTR the parameters are ordered as the first element of pages is linked to the first element of excludes.
+         *
+         * @param rootNode: the parent node to process
+         * @param arrayOfPages: a list of list of pages to includes
```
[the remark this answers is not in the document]
surli: Indeed it changes after a second round. I'll fix the doc.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
surli: OK I'll refactor to use a POJO then.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/exporter/exporter.js
surli: indeed. Old stuff I guess. Will remove it.

surli:
OK I changed both the java and javascript side to use a Map instead of two lists for included/excluded pages, as suggested by @atallahade and @mflorea. 
So for me all suggestions have been taken into account and fixed.

surli:
> We'll need to test it properly.

I added some functional test to check the behaviour as well as a new unit test. Now, I also tested it manually. 
Do you have another kind of test in mind?

mflorea:
> Do you have another kind of test in mind?

The QA team, but after we merge it :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -70,6 +75,93 @@
...
+            this.name = request.get("name");
+            if (StringUtils.isBlank(name)) {
+                this.name = context.getDoc().getFullName();
+            }
+
+            String[] pages = request.getParameterValues("pages");
+            String[] excludes = request.getParameterValues("excludes");
+
+            this.exportPages = new HashMap<>();
```
tmortagne: Just a detail: I would use a LinkedHashMap here to keep the same order than what you got in input.

surli: Fixed.

PR #863, opened 2018-09-19
XWIKI-15605: Auto-suggest control doesn't display correctly hidden pages

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15605

### Changes

* Add an optional parameter to the ClassPropertyValuesResource API to get the exact match value.
* Add a getValue method to the ClassPropertyValuesProvider to get the exact match value.
* Use the exact match parameter when fetching selected items in the property suggest picker.
* Use the exact match query for selected values in users or groups suggest picker.

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -398,6 +398,19 @@
                     "old": "parameter void org.xwiki.notifications.filters.watch.WatchedUserReference::<init>(java.lang.String, org.xwiki.notifications.filters.internal.user.EventUserFilterPreferencesGetter, ===org.xwiki.notifications.filters.NotificationFilterManager===)",
                     "new": "parameter void org.xwiki.notifications.filters.watch.WatchedUserReference::<init>(java.lang.String, org.xwiki.notifications.filters.internal.user.EventUserFilterPreferencesGetter, ===org.xwiki.notifications.filters.NotificationFilterPreferenceManager===)",
                     "justification": "Young API"
+                  },
+                  {
+                    "code": "java.method.numberOfParametersChanged",
+                    "old": "method org.xwiki.rest.model.jaxb.PropertyValues org.xwiki.rest.resources.classes.ClassPropertyValuesResource::getClassPropertyValues(java.lang.String, java.lang.String, java.lang.String, java.lang.Integer, java.util.List<java.lang.String>) throws org.xwiki.rest.XWikiRestException",
+                    "new": "method org.xwiki.rest.model.jaxb.PropertyValues org.xwiki.rest.resources.classes.ClassPropertyValuesResource::getClassPropertyValues(java.lang.String, java.lang.String, java.lang.String, java.lang.Integer, java.util.List<java.lang.String>, java.lang.Boolean) throws org.xwiki.rest.XWikiRestException",
+                    "package": "org.xwiki.rest.resources.classes",
+                    "classQualifiedName": "org.xwiki.rest.resources.classes.ClassPropertyValuesResource",
+                    "classSimpleName": "ClassPropertyValuesResource",
+                    "methodName": "getClassPropertyValues",
+                    "oldArchive": "org.xwiki.platform:xwiki-platform-rest-api:jar:10.7",
+                    "newArchive": "org.xwiki.platform:xwiki-platform-rest-api:jar:10.9-SNAPSHOT",
+                    "elementKind": "method",
+                    "justification": "Add an optional parameter to a REST API end point"
                   }
                 ]
               }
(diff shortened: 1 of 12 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/AbstractClassPropertyValuesProvider.java
```diff
@@ -92,6 +98,31 @@ public PropertyValues getValues(ClassPropertyReference propertyReference, int li
...
+        DocumentReference documentReference = documentReferenceResolver.resolve(filter, EntityType.DOCUMENT);
+        if (documentReference != null) {
+            PropertyValue value =
+                this.getValueFromQueryResult(documentReference, getPropertyDefinition(propertyReference));
+            if (value != null) {
+                value.setValue(filter);
+                propertyValues.getPropertyValues().add(value);
+            }
+        }
+        return propertyValues;
```
mflorea: I would probably simply return:

    getValueFromQueryResult(rawValue, getPropertyDefinition(propertyReference))

and let derived classes overwrite this method if they want to resolve the raw value before calling ``getValueFromQueryResult``.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/ClassPropertyValuesResourceImpl.java
```diff
@@ -84,8 +84,13 @@ public PropertyValues getClassPropertyValues(String wikiName, String className,
         propertyLink.setHref(propertyURI.toString());
         propertyLink.setRel(Relations.PROPERTY);
 
-        PropertyValues propertyValues =
-            this.propertyValuesProvider.getValues(classPropertyReference, limit, filterParameters.toArray());
+        PropertyValues propertyValues;
+        if (isExactMatch) {
+            propertyValues = this.propertyValuesProvider.getValue(classPropertyReference, filterParameters.toArray());
```
mflorea: Here we could call ``getValue`` for each raw value in ``filterParameters`` and collect the results in ``PropertyValues``. This would allow the client side to make a single request to resolve the selected values rather than making multiple requests, one per selected value.

atallahade:
Thanks for the review @mflorea, I've made the changes.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-api/src/main/java/org/xwiki/rest/resources/classes/ClassPropertyValuesProvider.java
```diff
@@ -46,4 +46,18 @@
...
+    /**
+     * Provides the document specified by the filter as a property value.
+     *
+     * @param propertyReference the property to provide the value for
+     * @param rawValue raw value used to resolve the property value
+     * @return the property value based on the raw value
+     * @throws XWikiRestException if retrieving the property value fails
+     * @since 10.9RC1
+     */
+    default PropertyValues getValue(ClassPropertyReference propertyReference, Object rawValue)
```
mflorea: Why not return ``PropertyValue`` (singular) because the method is supposed to return a single value? The only advantage I see for returning ``PropertyValues`` (plural) is to avoid NPE (you return an empty collection instead of null). But on the other hand it complicates the code a bit because you need to work with a collection. WDYT?

atallahade: I think I did that because the related REST API needed a collection. It is indeed more logical to have the method return a single value or null and let the caller wrap it into a collection.

xwiki-platform-core/pom.xml
```diff
@@ -398,6 +398,19 @@
                     "old": "parameter void org.xwiki.notifications.filters.watch.WatchedUserReference::<init>(java.lang.String, org.xwiki.notifications.filters.internal.user.EventUserFilterPreferencesGetter, ===org.xwiki.notifications.filters.NotificationFilterManager===)",
                     "new": "parameter void org.xwiki.notifications.filters.watch.WatchedUserReference::<init>(java.lang.String, org.xwiki.notifications.filters.internal.user.EventUserFilterPreferencesGetter, ===org.xwiki.notifications.filters.NotificationFilterPreferenceManager===)",
                     "justification": "Young API"
+                  },
+                  {
```
tmortagne: You might want to remove that and build to get a more lightweight version from the new revapi version error. In any case you should definitely remove newArchive which is going to fail the release.

atallahade: Ok I'll fix that, thanks!

PR #864, opened 2018-09-19
XWIKI-15483: Add missing dependency to macro display

surli:


surli:
ping @vmassol

vmassol:
Notes:
* Missing a lot of other macros too: include, velocity, etc
* MenuMacro.xml should not draw unnecessary dependencies such as "toc" and "code" macros

PR #867, opened 2018-09-25
[Misc] Fix partially tested method in rendering-configuration-default

surli:


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-configuration/xwiki-platform-rendering-configuration-default/src/test/java/org/xwiki/rendering/internal/configuration/DefaultExtendedRenderingConfigurationTest.java
```diff
@@ -98,6 +98,15 @@ public void isImageDimensionsIncludedInImageURL() throws Exception
         assertFalse(this.defaultExtendedRenderingConfiguration.isImageDimensionsIncludedInImageURL());
     }
 
+    @Test
+    public void isImageDimensionsIncludedInImageURL2() throws Exception
```
vmassol: What does the "2" means? Could we be more explicit in the method name to differentiate it from the other test ?

vmassol: Maybe it should go int he existing test.

surli: Actually this is exactly the same test as above but reversing the values: I'm checking that a mock returning true, returns true... 
I'd be glad to have suggestion if you have a more meaningful name in this case

vmassol: Solution: merge it with the previous one. They're testing the same thing.

vmassol: >  I'm checking that a mock returning true, returns true...

I think it's a bit more complex than that, and that there are intermediary steps no?

vmassol: Just checked and indeed it's more complex than that. It's the "default" configuration source that is mocked.

It also checks that configuration source is called with the right parameter.

surli: OK so I'll merge it with the previous one.

PR #873, opened 2018-09-27
XWIKI-14314: Jobs waiting conflict when leaving the page during action

surli:
* Add a confirm dialog to prevent users leaving a pending job in
delete/rename/move

For the record, we cannot specify the text in the confirm dialog as stated in https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Notes

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.js
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.js
@@ -12,6 +12,12 @@ require.config({
   }
 });
 require(['jquery', 'xwiki-meta', 'tree'], function($, xm) {
+  // prevent the user from leaving the delete
+  $(window).on("beforeunload", function(e) {
+    e.preventDefault();
+    e.returnValue = "";
+  });
+
   $(document).ready(function() {
     var progressBar = $('#delete-progress-bar');
     var jobId = progressBar.data('job-id');
@@ -108,6 +114,10 @@ require(['jquery', 'xwiki-meta', 'tree'], function($, xm) {
         // Called when the user click on "delete"
         $('.btConfirmDelete').click(function(event){
           event.preventDefault();
+
+          // we don't ask again the user to confirm he's leaving the page
+          $(window).off("beforeunload");
+
           /*jshint camelcase:false */
 	        // Get the selection
           var selectedNodes = $('.deleteTree').jstree().get_selected(true);
@@ -139,6 +149,9 @@ require(['jquery', 'xwiki-meta', 'tree'], function($, xm) {
         // Called when the user click on "cancel"
         $('.btCancelDelete').click( function(event) {
           event.preventDefault();
+
+          // we don't ask again the user to confirm he's leaving the page
+          $(window).off("beforeunload");
           $('.btConfirmDelete').prop('disabled', 'disabled');
           var notif = new XWiki.widgets.Notification(
             "$escapetool.javascript($services.localization.render('core.delete.warningExtensions.canceling'))",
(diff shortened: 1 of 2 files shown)


surli:
It's ready for review @atallahade @vmassol @mflorea @tmortagne

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.js
```diff
@@ -12,6 +12,12 @@ require.config({
   }
 });
 require(['jquery', 'xwiki-meta', 'tree'], function($, xm) {
+  // prevent the user from leaving the delete
+  $(window).on("beforeunload", function(e) {
+    e.preventDefault();
+    e.returnValue = "";
+  });
```
mflorea: I don't like this. ``delete.js`` could be loaded after the delete job is finished (e.g. if the delete job is quick) or before the delete job starts (and you can't be certain that the job will start). Moreover, if the question is not asked (because the delete job is run in non-interactive mode) the ``beforeunload`` listener is not removed.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/job/job.js
```diff
@@ -32,6 +32,12 @@ require.config({
 
 require(['jquery', 'xwiki-meta', 'JobRunner'], function($, xm, JobRunner) {
   'use strict';
+  // prevent the user from leaving the job
+  $(window).on("beforeunload", function(e) {
+    e.preventDefault();
+    e.returnValue = "";
+  });
```
mflorea: Same as above. ``job.js`` can be loaded anytime before, during or after the job is finished and there might be no question to be asked or the job could be run in non-interactive mode.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.js
surli: Do you have a suggestion to put it somewhere else then? I think I looked on `ExtensionBreakingQuestion.js` and it wasn't the proper location, but I can look again.

mflorea: I wouldn't put it anywhere because I don't think the leave confirmation is a solution. As @evalica commented on the JIRA issue, you don't have to leave the page to have problems. You can just start a new job on a new browser tab or have another user start another job. I would rather let the user navigate away and use the notification menu to indicate that there is a job waiting for an answer.

surli: I completely agree with you: for me this PR was really a quickfix that could help some users in such situation. I actually did it because I had the issue and wasn't understood what happened: for me the confirm dialog would have helped.

Now what you suggest should be done, but I don't know how much it will take and I don't think the solutions are conflicting.

mflorea: The issue is that it's not easy to show the confirmation dialog only when it's needed. If we show it all the time (even if the job didn't start yet, or the job is already finished or the job doesn't have any question, etc.) it becomes annoying.

surli: ok, then indeed it's not a good quickfix if it could become annoying for users. I'll close the PR.

PR #875, opened 2018-09-27
XWIKI-15547: When getting a 401 after going to a tab it reloads the page

surli:


surli:
This is also ready for review @atallahade @vmassol @mflorea @tmortagne

PR #876, opened 2018-09-27
XWIKI-14647: Allow filtering by title in all Page Index Livetables using titles

ClemensRobbenhaar:
Also contains a fix for XWIKI-15458 Filtering by title in the "Children" view is ineffective

I have not fiddled much with core templates lately, so maybe please someone review this.


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -287,6 +287,9 @@
   <div id="${htmlLiveTableId}-inaccessible-docs" class="hidden">
     #info($services.localization.render('rightsmanager.documentrequireviewrights'))
   </div>
+  <div id="${htmlLiveTableId}-computed-title-docs" class="hidden">
+    #info("(&#xb9;)&#160;$services.localization.render('platform.livetable.doc_title_computed.filter_sort_hint')")
```
mflorea: Since you're generating HTML I think it would be more clear to use ``<sup>1</sup>`` and ``&nbsp;`` (for non-breaking space) rather than these cryptic HTML code entities.

And regarding the translation key, I haven't seen that many keys using underscore. We're using camel case (or dot) most of the time I think.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DocumentsMacro.xml
```diff
@@ -177,6 +177,7 @@ Generates:
 })
 #set($options = { 
   'translationPrefix' : 'platform.index.',
+  'selectedColumn' : 'doc.location',
```
mflorea: The list of columns is configurable so I hope there's no problem if ``doc.location`` is not in the list (for instance when you want to display documents from the same location).

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -340,6 +340,9 @@
     #set($discard = $row.put('doc_author_url', $xwiki.getURL($translatedDoc.author)))
     #set($discard = $row.put('doc_date', $xwiki.formatDate($translatedDoc.date)))
     #set($discard = $row.put('doc_title', $translatedDoc.plainTitle))
+    #if($translatedDoc.getTitle() != $translatedDoc.getPlainTitle())
+	  #set($discard = $row.put('doc_title_computed', true))
```
mflorea: I'm wondering if we should send the raw title back (``doc_title_raw``) instead of the boolean flag (``doc_title_computed``). I would make more sense data-wise but it would also make the JSON a little bigger.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/getdocuments.vm
```diff
@@ -153,6 +154,9 @@ $response.setContentType("application/json")
     #set ($discard = $row.put('doc_author_url', $xwiki.getURL($translatedDoc.author)))
     #set ($discard = $row.put('doc_date', $xwiki.formatDate($translatedDoc.date)))
     #set ($discard = $row.put('doc_title', $translatedDoc.plainTitle))
+    #if ($translatedDoc.getTitle() != $translatedDoc.getPlainTitle())
+      #set ($discard = $row.put('doc_title_computed', true))
```
mflorea: Same as above.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -1452,6 +1452,9 @@ $Msz MB##
   <div id="${htmlLiveTableId}-inaccessible-docs" class="hidden">
     #info($services.localization.render('rightsmanager.documentrequireviewrights'))
   </div>
+  <div id="${htmlLiveTableId}-computed-title-docs" class="hidden">
+    #info("(&#xb9;)&#160;$services.localization.render('platform.livetable.doc_title_computed.filter_sort_hint')")
```
mflorea: Same as above.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -1767,6 +1773,9 @@ function (row, i, table) {
         if (fieldName === 'doc_name' && !row['doc_viewable']) {
           text += '*';
         }
+        if (fieldName === 'doc_title' && row['doc_title_computed']) {
+          text += '\u00b9';
```
mflorea: I think it would be nicer to add a CSS class on the ``container`` (table cell). This would allow us to apply whatever styling we want, including adding the superscript 1 using the ``:after`` selector. This is important especially since we could have titles that end with ``\u00b9``.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
ClemensRobbenhaar: I see about the translation key; will fix it this evening.
I am not sure about the marker; I can use the ``<sup>1</sup>`` as well, but then I should use it in the display of the entries as well. (The other marker telling about non-viewable pages currently uses a plain ``*``)

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
ClemensRobbenhaar: Also sending the ``doc_title_raw`` means the logic about if the title is computed moves into the JavaScript code. Maybe it is some old habit of mine, but I prefer to have the logic on the server side code whenever possible. (Probably pointless, as lots of logic end up in JavaScript today anyway.)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
ClemensRobbenhaar: If I add a CSS class to the container, then maybe I should try to add the same class to the explanatory message so they do not end up differerent symbols.
I will give it a try to see if that is possible without needing awkward code.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
atallahade: @mflorea there are many translation keys with underscore but maybe this is not the best practice. In fact [translations keys from old core](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties) are not very consistent (mix of camel case, underscore, dots and lower case).

tmortagne: Yeah many of the oldcore translation are much older that any kind of best practice :)

See https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HTranslationPropertyNaming.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DocumentsMacro.xml
ClemensRobbenhaar: I just added the ``selectedColumn``for backwards compatibility reasons (maybe users got used to having the page sorted by location initially ...). I tested and read the code, and this does not cause any errors when the ``doc.location`` column is not present; however in that case no column is selected by default, not even the first one.
I guess I will change the code so it adds the ``selectedColumn`` setting here only if ``doc.location`` is in the list of columns to be displayed.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
ClemensRobbenhaar: I went the rout to use CSS to create that symbol, so it will remain a "cryptic code entity" but this time in the CSS (to avoid encoding issues).

ClemensRobbenhaar:
I hope I addressed all concerns with my last two commits

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
ClemensRobbenhaar: Ah, ok, now I understand the use case.
The JS code should not get more complicated by this change either, so I will give it a try.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/getdocuments.vm
```diff
@@ -154,8 +154,9 @@ $response.setContentType("application/json")
     #set ($discard = $row.put('doc_author_url', $xwiki.getURL($translatedDoc.author)))
     #set ($discard = $row.put('doc_date', $xwiki.formatDate($translatedDoc.date)))
     #set ($discard = $row.put('doc_title', $translatedDoc.plainTitle))
-    #if ($translatedDoc.getTitle() != $translatedDoc.getPlainTitle())
-      #set ($discard = $row.put('doc_title_computed', true))
+    #set ($rawTitle = $translatedDoc.title)
+    #if ($rawTitle != $translatedDoc.plainTitle)
```
mflorea: Same as above.

PR #877, opened 2018-10-02
[Misc] Specify the VNC container to use in order to avoid segfault

surli:
@vmassol so VNC is now working with this change on my machine. Now we can also wait a new release of testcontainer.


vmassol:
Thanks!

PR #880, opened 2018-10-04
XWIKI-15601: Global menus are not displayed in subwiki

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15601

## Fix proposed 

When possible retrieve directly the extension reference, instead of building it based on its ID. It also allows to keep the retrocompatibility with existing menu.

Another solution would have been to put an absolute name in the menu (i.e. with wiki name), but then it would break current behaviour: users currently don't have to put an absolute name in Panels config, unless they are in a subwiki.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
@@ -20,7 +20,9 @@
 package org.xwiki.panels.internal;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -30,7 +32,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.xwiki.component.manager.ComponentLookupException;
 import org.xwiki.component.manager.ComponentManager;
+import org.xwiki.component.wiki.WikiComponent;
 import org.xwiki.configuration.ConfigurationSource;
+import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.DocumentReferenceResolver;
 import org.xwiki.model.reference.EntityReferenceSerializer;
 import org.xwiki.uiextension.UIExtension;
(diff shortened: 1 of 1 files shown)


tmortagne:
I don't know AbstractPanelsUIExtensionManager very well but the name suggest that it manipulate panels ? Or anyone can inject any uix as panel ?

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
```diff
@@ -109,7 +110,18 @@
                     for (UIExtension extension : allExtensions) {
                         // Resolve and serialize the extension id to make it absolute before we can compare it with
                         // the serialized panel reference.
-                        String extensionId = serializer.serialize(resolver.resolve(extension.getId()));
```
tmortagne: Not your fault but this code is pretty bad. Going trough serialized references does not make much sense, panelSerializedReferences should be a list of DocumentReference and `extensionId.equals(panelSerializedReference)` should be replaced by an equals between two references.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
```diff
@@ -109,7 +110,18 @@
...
+
+                        String extensionId;
+
+                        // in this case we can directly retrieve the document reference which gives also the wiki name
+                        // allows to fix XWIKI-14557
+                        if (extension instanceof WikiUIExtension) {
```
tmortagne: Would be better to manipulate WikiComponent instead of WikiUIExtension since both PanelWikiUIExtension and WikiUIExtension implement it.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
tmortagne: Actually actually panelSerializedReferences should even be replaced with a Set of DocumentReference so that the following code can just use #contains instead of testing each extension with each configured panel.

surli:
> I don't know AbstractPanelsUIExtensionManager very well but the name suggest that it manipulate panels only. Anyone can inject any uix as panel ?

I get both instances of WikiUIExtension and PanelWikiUIExtension, so I imagine we could have other kind of extension too.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
surli: I will look for some refactoring.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
surli: Hm indeed. I'll use it.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
```diff
@@ -95,24 +99,30 @@
...
-                        if (extensionId.equals(panelSerializedReference)) {
-                            panels.add(extension);
-                        }
+                // TODO: This is not performant and will not scale well when the number of UIExtension instances
+                // increase in the wiki
+                for (UIExtension extension : allExtensions) {
+                    DocumentReference extensionId;
+
+                    // in this case we can directly retrieve the document reference which gives also the wiki name
+                    // allows to fix XWIKI-14557
```
vmassol: it's XWIKI-15601. 

However, in general we try to not reference jira issue too much as we consider jira to be transient. So instead we explain the problem we're fixing.

vmassol: It seems that you improved performance by reusing the doc ref that is already stored in the WikiComponent instance.

vmassol: Now that doesn't explain what was the problem with XWIKI-15601

vmassol: You need to check that https://jira.xwiki.org/browse/XWIKI-15388 is still working

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
```diff
@@ -95,24 +99,35 @@
 
         // Verify that there's a panel configuration property defined, and if not don't return any panel extension.
         if (!StringUtils.isEmpty(panelConfigurationString)) {
-            List<String> panelSerializedReferences = new ArrayList<String>();
+            Set<DocumentReference> panelSerializedReferences = new HashSet<>();
```
tmortagne: It's not really "serialized references" anymore ;)

PR #882, opened 2018-10-05
XWIKI-15601: Order was no longer preserved in panels

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15601

## Solution
  * Fix by maintaining a map with the panel order and create a
well-ordered list.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
@@ -20,9 +20,9 @@
 package org.xwiki.panels.internal;
 
 import java.util.ArrayList;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.List;
-import java.util.Set;
+import java.util.Map;
 
 import javax.inject.Inject;
 import javax.inject.Named;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-api/src/main/java/org/xwiki/panels/internal/AbstractPanelsUIExtensionManager.java
```diff
@@ -126,8 +130,12 @@
...
                     }
 
-                    if (panelSerializedReferences.contains(extensionId)) {
-                        panels.add(extension);
+                    if (panelReferenceWithPosition.containsKey(extensionId)) {
+                        int position = panelReferenceWithPosition.get(extensionId);
+                        if (position > panels.size()) {
+                            position = panels.size();
+                        }
+                        panels.add(position, extension);
```
vmassol: This looks complex. Can't you achieve the same with a LinkedSet instead of a HashSet?

vmassol: Also if I remember we were iterating on the panelSerializedReferences before, need to check

surli: No because I don't iterate on the panels element but on the UIExtensions we get, so I don't know in which order I'm building the new list

vmassol: Yes we had this which was guaranteing the order:

```
for (String panelSerializedReference : panelSerializedReferences) {
```

vmassol: So why not put that back?

surli: Yes I changed that to avoid iterating multiple times on the UIExtension, as the list might be big.

vmassol: indeed, you improved the perfs, that's good!

vmassol: Another idea: why not use `indexOf()` instead of `contains()` so that you get both the index + whether it exists, to replace: `if (panelSerializedReferences.contains(extensionId)) {` ?

It's slightly less efficient though than your solution, at the expense of less memory.

surli: I personally find my solution more clear, as it emphasizes the fact that you won't iterate on the panel references. But it's really more a matter of style there.

vmassol: My preference goes to conciseness in general. But I'm fine to keep your solution.

PR #885, opened 2018-10-11
XWIKI-15653: Prevent users from deleting/moving/renaming/copying pages containing XClass

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15653

## Solution

This PR use the Question API to ask question to the user during the job, in the same way that the mechanism to prevent breaking extension does. 

Then the mechanism is mostly inspired on BreakingExtension. However there is some differences:
  1. if the user is a simple user, we only display a red box with information and a cancel button. The job is cancelled after one minute.
  2. If the user is an advanced user, we display a tree with the list of pages containing an XClass, and the list of concerned objects, even if those cannot be modified yet. 

## Job done

 * Create a new EventListener on DeletingPage event and an associated
question
  * Create the dedicated template for the new question
  * Refactor delete.vm to rely on job.js for managing questions
  * Delete old js files related to the deletion as they are now useless
  * Add some unit tests
  * Fix a bug in ExtensionBreakingExtension.js related to the name of
parameters

## Tests

Unit tests has been added.


surli:
@tmortagne @mflorea I still need to add tests to finish the PR but I'd like a first review on this one if you don't mind. Especially since I'd need some help to fix the primary issue I'm talking about in the description of the PR.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassBreakingQuestion.java
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * The map of entities concerned by the refactoring, where the entity reference if the key.
+     */
+    private Map<EntityReference, EntitySelection> concernedEntities;
+
+    /**
+     * The map of objects that might be impacted by a deletion.
+     * The key is represented by the local name of the page containing the XClass.
+     */
+    private Map<String, Set<String>> impactedObjects;
```
tmortagne: I'm really not a fan of this, keep in mind that DeleteJob allow deleting documents from different wikis in the same job. If you want to avoid problem always manipulated references.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,161 @@
...
+        DocumentReference documentReference = (DocumentReference) entitySelection.getEntityReference();
+        String query = "select distinct doc.fullName from XWikiDocument doc, BaseObject obj where "
+            + "doc.fullName=obj.name and obj.className=:className";
+
+        String serializedReference = serializer.serialize(documentReference);
+        String className = serializedReference.substring(serializedReference.indexOf(":") + 1);
+        try {
+            List<String> results = this.queryManager.createQuery(query, Query.HQL)
+                .bindValue("className", className)
+                .setWiki(documentReference.getWikiReference().getName()).setLimit(10)
```
tmortagne: I don't understand the limit, you are going to have missing documents, don't you ? Either you put everything in the question or you use limit(1) just to make sure there is something and you let the question UI gather the rest.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
```diff
@@ -0,0 +1,163 @@
...
+  #foreach ($page in $pages)
+    #set ($pageReference = $page.entityReference)
+    #if ($velocityCount > $offset + $limit)
+      #set ($howMany = $pages.size() - $velocityCount + 1)
+      #paginationNode($data, $howMany, $velocityCount)
+      #break
+    #end
+    #if ($velocityCount > $offset)
+      #set ($pageDoc = $xwiki.getDocument($pageReference))
+      #set ($title = $pageDoc.getRenderedTitle('html/5.0'))
```
tmortagne: getDisplayTitle() will find the target syntax automatically (which might be another variant of html).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
```diff
@@ -0,0 +1,163 @@
...
+  #set ($pages = $question.impactedObjects.get($request.id))
+  #foreach ($pageReference in $pages)
+    #if ($velocityCount > $offset + $limit)
+      #set ($howMany = $pages.size() - $velocityCount + 1)
+      #paginationNode($data, $howMany, $velocityCount)
+      #break
+    #end
+    #if ($velocityCount > $offset)
+      #set ($pageDoc = $xwiki.getDocument($pageReference))
+      #set ($title = $pageDoc.getRenderedTitle('html/5.0'))
```
tmortagne: Same comment as above.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
```diff
@@ -0,0 +1,163 @@
...
+    #displayObjects($question, $data)
+  #end
+#end
+
+#**
+ * Add XClass nodes in the collection.
+ * @param $question the question beeing asked
+ * @param $data the collection where the nodes will be added
+ *#
+#macro(displayXClass $question $data)
```
tmortagne: There seems to be a lot of duplication between the various display* macros.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/job/question/XClassBreakingQuestion.js
```diff
@@ -0,0 +1,140 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+/*!
+#set ($jsExtension = '.min.js')
+#if (!$services.debug.minify)
+  #set ($jsExtension = '.js')
+#end
+*/
```
atallahade: Not sure if this is supposed to be in a JS comment.

surli: I get this code from ExtensionBreakingQuestion: I think the comments are on purpose to avoid lint checking errors. But I might be wrong.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,161 @@
...
+
+            if (results.isEmpty()) {
+                question.markAsFreePage(entitySelection);
+            } else {
+                for (String documentObjectName : results) {
+                    question.markImpactedObject(entitySelection, serializedReference, documentObjectName);
+                }
+            }
+        } catch (QueryException e) {
+            e.printStackTrace();
```
surli: I forgot to manage the exception case: should I cancel the job with the error?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
tmortagne: > it won't retrieve the objects of the other wikis

Not a problem. It's an official limitation of the classes that it can only be used in the same wiki.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/job/question/XClassBreakingQuestion.js
tmortagne: > I think the comments are on purpose to avoid lint checking errors

Yep. Took me several tries to make it right for ExtensionBreakingQuestion :)

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
tmortagne: If there is any issue with that request it means deleting a class become impossible. Might be safer to just log an error, at worst it's easy to revert that delete.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
surli: Actually the Javadoc says that `getDisplayTitle()` is deprecated and that I should use `getRenderedTitle()` instead, see: https://github.com/xwiki/xwiki-platform/blob/6fd9baa341e370206744f1577641316177e3d2e4/xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/doc/XWikiDocumentCompatibilityAspect.aj#L162-L172

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/job/question/XClassBreakingQuestion.js
mflorea: It's actually to prevent the comment from being removed when the JavaScript is minified.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
```diff
@@ -0,0 +1,174 @@
...
+    #if ($velocityCount > $offset + $limit)
+      #set ($howMany = $collection.size() - $velocityCount + 1)
+      #paginationNode($data, $howMany, $velocityCount)
+      #break
+    #end
+    #if ($velocityCount > $offset)
+      #set ($pageReference = '')
+      #getReference($type, $element, $pageReference)
+      #set ($pageDoc = $xwiki.getDocument($pageReference))
+      #set ($title = $pageDoc.getDisplayTitle())
```
mflorea: ``plainTitle`` is better.

surli:
So I think this one is ready for review @tmortagne @mflorea 
I'm sorry it's pretty big, since I had to do a bit of refactoring to be able to use directly the same question mechanism in both move/rename and delete UI.

You can have a look on the description: I mentionned my different actions on this PR. I didn't created a functional test for lack of time, I tested it locally on my own instance: maybe we should run part of the CI on this branch before merging to check the functional tests of flamingo? 

Regarding the problems of asking both question to the user when a page contains an XClass and is part of an extension, I solved it by checking directly in the concerned listener if the event has been canceled.

tmortagne:
>  I'm sorry it's pretty big, since I had to do a bit of refactoring to be able to use directly the same question mechanism in both move/rename and delete UI.

It's OK, it's great that someone finally did it.

surli:
Also note that this PR does not cover the "copy" case of the issue: I'll do it in another PR, as it won't use the same mechanism.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/delete/DocumentsDeletingListener.java
```diff
@@ -81,6 +81,12 @@ public DocumentsDeletingListener()
     @Override
     public void onEvent(Event event, Object source, Object data)
     {
+        CancelableEvent cancelableEvent = (CancelableEvent) event;
+        if (cancelableEvent.isCanceled()) {
+            logger.warn("Skipping " + this.getName() + " as the event is already cancelled.");
```
tmortagne:
```suggestion
            logger.warn("Skipping [{}] as the event is already cancelled.", this.getName());
```

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassBreakingQuestion.java
```diff
@@ -0,0 +1,191 @@
...
+    {
+        for (EntitySelection entitySelection : concernedEntities.values()) {
+            entitySelection.setSelected(false);
+        }
+    }
+
+    /**
+     * @return the list of entities containing used XClass.
+     */
+    public Set<EntitySelection> getXclassPages()
```
tmortagne: getXClassPages ?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @version $Id$
+ * @since 10.9RC1
+ */
+@Component
+@Singleton
+@Named("XClassDeletingListener")
+public class XClassDeletingListener extends AbstractEventListener
+{
+    private static final List<Event> EVENTS = Arrays.asList(new DocumentsDeletingEvent());
```
tmortagne: You don't really need this static. Usually I directly pass the events to the abstract constructor vararg.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,172 @@
...
+
+    @Inject
+    private Logger logger;
+
+    @Inject
+    private QueryManager queryManager;
+
+    @Inject
+    @Named("local")
+    private EntityReferenceSerializer<String> serializer;
```
tmortagne: Details: naming it `localSerializer` would make following code easier to read.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,172 @@
...
+                this.logger.warn("Confirm question has been interrupted.");
+                cancelableEvent.cancel("Question has been interrupted.");
+            }
+        }
+    }
+
+    private void checkIfDeleteIsAllowed(EntitySelection entitySelection, XClassBreakingQuestion question)
+    {
+        DocumentReference documentReference = (DocumentReference) entitySelection.getEntityReference();
+        String query = "select distinct doc.fullName from XWikiDocument doc, BaseObject obj where "
```
tmortagne: select distinct obj.name from BaseObject obj where obj.className=:className

would be faster.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,172 @@
...
+            List<String> results = this.queryManager.createQuery(query, Query.HQL)
+                .bindValue("className", className)
+                .setWiki(documentReference.getWikiReference().getName())
+                .<String>execute();
+
+            if (results.isEmpty()) {
+                question.markAsFreePage(entitySelection);
+            } else {
+                for (String documentObjectName : results) {
+                    EntityReference documentObjectReference = this.resolver.resolve(documentObjectName,
```
tmortagne: You should pass `documentReference` as 3rd parameter to resolve the reference based on the class instead of the current context which may have nothing to do with the class (it's always main wiki in the context pf the refactoring jobs AFAIK). Also that way you can use the default resolver.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,172 @@
...
+            } catch (InterruptedException e) {
+                this.logger.warn("Confirm question has been interrupted.");
+                cancelableEvent.cancel("Question has been interrupted.");
+            }
+        }
+    }
+
+    private void checkIfDeleteIsAllowed(EntitySelection entitySelection, XClassBreakingQuestion question)
+    {
+        DocumentReference documentReference = (DocumentReference) entitySelection.getEntityReference();
```
tmortagne: Naming it `classReference` might make the code a bit more readable.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/test/java/org/xwiki/refactoring/job/XClassDeletingListenerTest.java
```diff
@@ -0,0 +1,312 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * Validate the behaviour of {@link XClassDeletingListener}
+ *
+ * @version $Id$
+ * @since 10.9RC1
+ */
+@ComponentTest
+@ComponentList({
```
tmortagne: You could use `@ReferenceComponentList` instead if you don't want to deal with "what do I need to make reference serializers/resolvers work".

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/delete/DocumentsDeletingListener.java
surli: Ahah nice new Github feature but I'll apply it with the other changes ;)

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassBreakingQuestion.java
surli: ok

surli: I agree

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
tmortagne: It could technically but that would mean the database is not in a great shape. At worst if there is ghosts it just mean you will get a warning when deleting a xclass which is kind of not really used.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/delete/DocumentsDeletingListener.java
vmassol: If it's normal than cancelled event stop propagation I wouldn't use a `warn` level but more a `debug` one.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
surli: ok, thanks for the tip

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
surli: I was afraid it could lead to confusion because I'm manipulating a DocumentReference here, but apparently the concept of ClassReference does not exist, right? So indeed it might be better.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/test/java/org/xwiki/refactoring/job/XClassDeletingListenerTest.java
surli: It might simplify indeed.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
tmortagne: It's a good reflex to always try to find a reference to pass before relying on "current" resolver. It avoid mistakes like this.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/delete/DocumentsDeletingListener.java
surli: > If it's normal that cancelled events stop propagation

I made a proposal in that way, but It's not been accepted yet and Marius made some very good points against IMO. Now a warn might be a bit strong indeed. info could be appropriate here.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
tmortagne: Yes we did not introduce a ClassReference because it was always equal to the DocumentReference basically.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
surli: >  At worst if there is ghosts it just mean you will get a warning 

IMO the worst would be to get stucked when trying to delete an unused XClass because of a broken UI because of some remaining ghosts...

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
surli: ok. I'll change the name then.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/delete/DocumentsDeletingListener.java
vmassol: Not info, it's more like debug if it's normal. Warn means it's not normal (meaning not normal that propagation is cancelled, I thought this is what we wanted and thus was normal). 

See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
tmortagne: Why broken ? The UI will just tell you that the class is used and you conform you still want to delete it. But again it would mean someone purposely deleted an entry of the the document table without going trough the API. IMO the very small risk of useless warning worth the potential important performance boost in big databases (plus this optimization is often used).

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/test/java/org/xwiki/refactoring/job/XClassDeletingListenerTest.java
surli: Apparently `@ReferenceComponentList` is only visible platform-oldcore.

tmortagne: Indeed, I did not looked checked the module you were in. Now we could introduce another one with less resolvers/serializer in the model module.

surli: No real need for that PR though: it's only 2 components in the list.

surli:
@tmortagne @mflorea @vmassol could you review/merge this one? I'd like it to be in the 10.9RC1. Thanks

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/test/java/org/xwiki/refactoring/job/XClassDeletingListenerTest.java
tmortagne: Yes of course, was just a suggestion (which lost much of it's value with ReferenceComponentList not being available :)).

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassBreakingQuestion.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Represents a question asked to the user when he tries to refactor some pages containing used XClass.
+ *
+ * @version $Id$
+ * @since 10.9RC1
+ */
+public class XClassBreakingQuestion
+{
+    /**
+     * The map of entities concerned by the refactoring, where the entity reference if the key.
```
atallahade: Typo: `if the key` -> `is the key`

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/job/XClassDeletingListener.java
```diff
@@ -0,0 +1,167 @@
...
+    private EntityReferenceSerializer<String> localSerializer;
+
+    @Inject
+    private EntityReferenceResolver<String> resolver;
+
+    @Inject
+    private DocumentAccessBridge documentAccessBridge;
+
+    /**
+     * Construct a DocumentsDeletingListener.
```
atallahade: Not very important but you can write `{@link DocumentsDeletingListener}`

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/resources/ApplicationResources.properties
```diff
@@ -0,0 +1,82 @@
...
+job.question.XClassBreakingQuestion.confirm=Do you wish to continue?
+job.question.XClassBreakingQuestion.tree.selectAll=select all
+job.question.XClassBreakingQuestion.tree.selectNone=none
+job.question.XClassBreakingQuestion.tree.paginationNode={0} more....
+job.question.XClassBreakingQuestion.tree.freePages=Pages that do not contain any used XClass
+job.question.XClassBreakingQuestion.forbiddenConcernedPages=The following pages are incriminated in the forbidden action as they contain structured data types.
+
+job.question.XClassBreakingQuestion.refactoring/delete.title=You are about to delete pages that contain used XClass.
+job.question.XClassBreakingQuestion.refactoring/delete.explanation=If you delete these pages, some extensions might not work anymore.
+job.question.XClassBreakingQuestion.refactoring/delete.help=The recommended way of removing an XClass is to...
```
atallahade: @tmortagne Could you maybe help for this translation key?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/resources/ApplicationResources.properties
```diff
@@ -0,0 +1,82 @@
...
+job.question.XClassBreakingQuestion.refactoring/delete.tree.title=Pages to remove
+
+job.question.XClassBreakingQuestion.refactoring/delete.forbiddenTitle=You cannot proceed on this deletion.
+job.question.XClassBreakingQuestion.refactoring/delete.forbiddenExplanation=The pages you selected contain structured data types. Deleting these pages might break the code that uses those data types. Ask an advanced user to proceed on this refactoring.
+
+
+job.question.XClassBreakingQuestion.refactoring/rename.title=You are about to rename pages that contain used XClass.
+job.question.XClassBreakingQuestion.refactoring/rename.explanation=If you rename these pages, some extensions might not work anymore.
+job.question.XClassBreakingQuestion.refactoring/rename.help=
```
atallahade: To be done, same as above.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/resources/ApplicationResources.properties
```diff
@@ -0,0 +1,82 @@
...
+job.question.XClassBreakingQuestion.refactoring/rename.tree.title=Pages to rename
+
+job.question.XClassBreakingQuestion.refactoring/rename.forbiddenTitle=You cannot proceed on this renaming.
+job.question.XClassBreakingQuestion.refactoring/rename.forbiddenExplanation=The pages you selected contain structured data types. Renaming these pages might break the code that uses those data types. Ask an advanced user to proceed on this refactoring.
+
+
+job.question.XClassBreakingQuestion.refactoring/move.title=You are about to move pages that contain used XClass.
+job.question.XClassBreakingQuestion.refactoring/move.explanation=If you move these pages, some extensions might not work anymore.
+job.question.XClassBreakingQuestion.refactoring/move.help=
```
atallahade: Same as above.

xwiki-platform-core/xwiki-platform-web/overlays/org.xwiki.platform.xwiki-platform-index-tree-war-9.11.8-SNAPSHOT.info
```diff
@@ -0,0 +1,3 @@
+1539692341000
```
atallahade: I don't think you should include that file.

xwiki-platform-core/xwiki-platform-web/overlays/org.xwiki.platform.xwiki-platform-tree-war-9.11.8-SNAPSHOT.info
```diff
@@ -0,0 +1,3 @@
+1539692341000
```
atallahade: Same as above.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/job/question/XClassBreakingQuestion.js
```diff
@@ -0,0 +1,145 @@
...
+        // Register data callback
+        questionForm.data('job-answer-properties-extra', getAnswerProperties);
+
+        /**
+         * Represent the selected pages & extensions the user can chose to delete
+         */
+        var answerProperties = {
+            // Pages that have been manually marked by the user to be removed
+            selectedDocuments: [],
+            // Either or not all pages that don't belong to any extension should be removed (even those the user haven't
```
atallahade: I'm not sure I understand this comment.

xwiki-platform-core/xwiki-platform-web/overlays/org.xwiki.platform.xwiki-platform-index-tree-war-9.11.8-SNAPSHOT.info
surli: Indeed. Shouldn't it be in the gitignore @vmassol ?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/resources/ApplicationResources.properties
tmortagne: Not sure what is your question exactly.

surli: What's the recommended way of removing an XClass, for helping users? :) I did not manage to explain it in the help message.

xwiki-platform-core/xwiki-platform-web/overlays/org.xwiki.platform.xwiki-platform-index-tree-war-9.11.8-SNAPSHOT.info
tmortagne: Indeed. I don't even understand what is this file, never seen this.

surli: I saw the folder "overlays" several times, deleting it each time, except this one or I went a bit fast with the git add.

vmassol: Indeed. Now I've removed it and rebuilt platofrm-web and it didn't reappear. I thought it was the war plugin workDirectory but the default value seems ok: https://maven.apache.org/plugins/maven-war-plugin/war-mojo.html#workDirectory

Maybe it appears only when run in a reactor.

vmassol: I tried `mvn -pl org.xwiki.platform:xwiki-platform-web clean install` from the root and no overlays dir either

surli: Apparently the version is 9.11.8. It's certainly related to the `mvn install` done on an old branch.

atallahade: I sometimes have the folder too, I think we should simply do as @surli said and add a rule into the .gitignore file.

tmortagne: .gitignore would certainly not do any harm, at worst it's just useless

surli:
I removed the wrong files from my PR and apply some changes after Adel's review. So it's ready for me.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/resources/ApplicationResources.properties
```diff
@@ -0,0 +1,82 @@
...
+job.question.XClassBreakingQuestion.confirm=Do you wish to continue?
+job.question.XClassBreakingQuestion.tree.selectAll=select all
+job.question.XClassBreakingQuestion.tree.selectNone=none
+job.question.XClassBreakingQuestion.tree.paginationNode={0} more....
+job.question.XClassBreakingQuestion.tree.freePages=Pages that do not contain any used XClass
+job.question.XClassBreakingQuestion.forbiddenConcernedPages=The following pages are incriminated in the forbidden action as they contain structured data types.
+
+job.question.XClassBreakingQuestion.refactoring/delete.title=You are about to delete pages that contain used XClass.
+job.question.XClassBreakingQuestion.refactoring/delete.explanation=If you delete these pages, some extensions might not work anymore.
+job.question.XClassBreakingQuestion.refactoring/delete.help=The recommended way of refactoring an XClass is to first delete or edit the objects that are using it.
```
tmortagne: Sound weird written like this, it's not exactly a way. Something like "Its recommended to remove objects that are referring to an XClasses deleting it" ?

surli:
@mflorea @tmortagne this one is ready again for review. I added the limit in the query, but I also fixed some bugs:
  * in Java listener checking the canceledEvent wasn't enough apparently: when you cancel a job with the cancel button, it's the job itself which is canceled, the event cancel state remains false. It's true only if the cancel occurs because of a timeout
  * I refactored a bit the listener classes as I reached some complexity metrics with the new checks for cancelable event/job. 
  * in Java XClassBreakingQuestion I now check if the reference contains a locale, before putting it in a map of impacted objects
  * in JS, the checkboxes are also hidden after the paginated nodes are displayed

tmortagne:
> in Java listener checking the canceledEvent wasn't enough apparently: when you cancel a job with the cancel button, it's the job itself which is canceled, the event cancel state remains false. It's true only if the cancel occurs because of a timeout

We could improve the listeners to check the job status after the answer and set the event cancel state. It was indeed not needed when there was only one question so we did not tough about it.

surli:
> We could improve the listeners to check the job status after the answer and set the event cancel state.

Done. Then I could simplify the check before starting the listeners.

surli:
@mflorea wait before merging. I just catch a failing test :o

surli:
Fix of the functional test pushed.

mflorea:
Thanks for your contribution. For the future it would be good to split your work in multiple PR to make it easier to review the code and to catch problems.

surli:
> For the future it would be good to split your work in multiple PR to make it easier to review the code and to catch problems.

Yeah sorry for that I didn't expect this PR to become that big. :-/

surli:
@atallahade you wanted me to remind you to push the new properties file for Weblate ;)

PR #886, opened 2018-10-11
XWIKI-15719: .notifications-toggles inputs are missing labels

atallahade:
## Jira

https://jira.xwiki.org/browse/XWIKI-15719

## Change

* Add missing titles to inputs.

## Note

This approach was used: https://www.w3.org/WAI/tutorials/forms/labels/#using-the-title-attribute

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
@@ -799,7 +799,12 @@ require(['jquery', 'xwiki-meta', 'bootstrap', 'bootstrap-switch'], function ($, 
       </scope>
     </class>
     <property>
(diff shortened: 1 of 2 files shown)


evalica:
The issue is not about adding titles (those are good too), but adding **labels**, see 
* https://webaim.org/techniques/forms/controls#checkbox 
* https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/VerticalForms/ 
and we have some examples in Administration for checkboxes, etc.

evalica:
Since we don't want the labels be displayed for the graphical end-user, but we want them for accessibility reasons, we need to add classes like `.sr-only` or `.hidden`, see https://getbootstrap.com/docs/3.3/css/#helper-classes-show-hide

atallahade:
@evalica I've used labels instead of title as I don't think we need both.

evalica:
A. Spacing between the controls gets a bit different comparing to the old version.
B. Since there is repetitive code, maybe we could have a macro or have a for to remove duplication?

Besides this, looking good.

atallahade:
@evalica:
> A. Spacing between the controls gets a bit different comparing to the old version.

Yes it's more clear that way. I've added comments at the end of the lines to avoid new lines between the inputs.

> B. Since there is repetitive code, maybe we could have a macro or have a for to remove duplication?

I'll define a macro to make the code clean.

PR #887, opened 2018-10-11
XWIKI-15457: XAR Export of a page with a bang (!) in the name fails

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15457

## Solution

Add an explicit escape "!" to the clause. 

## Limitation

When testing the produced archive was named "Foo%21.xar" but it contained "Foo!.WebHome". Not sure if I should fix also the archive name or leave it as it is.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
@@ -295,7 +295,7 @@ public class ExportAction extends XWikiAction
                 }
 
                 where.append("doc.fullName like ?");
-                params.add(new DefaultQueryParameter(null).like(includePage));
+                params.add(getQueryParameter(includePage));
 
                 // if they exist we process the excludedPages associated with that include
                 if (!excludedPages.isEmpty()) {
@@ -316,7 +316,7 @@ public class ExportAction extends XWikiAction
                         }
 
                         where.append(" and doc.fullName not like ?");
-                        params.add(new DefaultQueryParameter(null).like(excludePage));
+                        params.add(getQueryParameter(excludePage));
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -361,6 +361,12 @@ private String extractWikiName(String documentName, XWikiContext context)
         return pageList;
     }
 
+    private QueryParameter getQueryParameter(String like)
+    {
+        String likeString = like.replaceAll("([!])", "!$1");
```
vmassol: Sounds very specific and missing information about the reason of the problem. Why is `!` not working? Because of the FS. Then you'll to reuse a FS path serializer (which already exists).

vmassol: ok so I've checked the jira issue and it's not related to the FS serialization.

We need to verify if the problem is not in the query module since it looks strange that we would need to replace the "!" char here.

surli:
ok to merge it then @vmassol ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -295,7 +295,7 @@ private String extractWikiName(String documentName, XWikiContext context)
                 }
 
                 where.append("doc.fullName like ?");
-                params.add(new DefaultQueryParameter(null).like(includePage));
+                params.add(this.getQueryParameter(includePage));
```
vmassol: Note: we don't usually use "this." for method calls.

vmassol:
@surli yes I agree but it still missing a comment to explain why you're escaping "!". Have you created the new issue?

PR #888, opened 2018-10-11
XWIKI-15724: ClassCastException when specifying the escape character and using like

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15724

## Solution

The expression should always be visited in `EscapeLikeParametersQuery` in order to be converted later.

## Test

Unit tests added in EscapeLikeParametersFilterTest

### DIFF
--- a/xwiki-platform-core/xwiki-platform-query/xwiki-platform-query-manager/src/main/java/org/xwiki/query/internal/EscapeLikeParametersQuery.java
+++ b/xwiki-platform-core/xwiki-platform-query/xwiki-platform-query-manager/src/main/java/org/xwiki/query/internal/EscapeLikeParametersQuery.java
@@ -161,8 +161,8 @@ public class EscapeLikeParametersQuery extends WrappingQuery
         {
             if (expr.getEscape() == null) {
                 expr.setEscape("!");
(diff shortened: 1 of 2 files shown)


vmassol:
Thanks. Would be good to augment the existing tests to test for this use case.

surli:
@vmassol 
> Would be good to augment the existing tests to test for this use case.

Finally I created a new dedicated unit test class as I didn't find an existing one. I tried to cover different cases. Without my fix 3 tests are failing, all green now. So for me it's ready.

vmassol:
We have EscapeLikeParametersFilterTest

vmassol:
I looked quickly and based on the test method name I think you may have recreated some tests that exist in EscapeLikeParametersFilterTest

surli:
:man_facepalming: ok I didn't checked as I assumed it was a different component.

surli:
I'll pick the ones that were failing and put them in it then, is that ok?

vmassol:
Sounds good

surli:
OK just picked up the 3 tests that were failing and reshape them to follow the style of EscapeLikeParametersFilterTest. So ready again for me.

PR #890, opened 2018-10-11
XWIKI-15718 #headerglobalsearchinput is missing the label

atallahade:
## Issue

https://jira.xwiki.org/browse/XWIKI-15718

## Change

* Add missing label to an input for accessibility purpose.


atallahade:
@evalica Indeed...

evalica:
Thanks for the fix Adel. I've merged the PR myself :) hoping that it will go into the BFD report.

PR #891, opened 2018-10-12
XWIKI-15726: Headings hierarchy is broken when using Bootstrap .modal-title and .panel-title

evalica:
- Adding styling for .modal-title for the font-size property
- Replace headings usage for .modal-title and .panel-title in order to comply with WCAG rules


tmortagne:
Seems much larger than export model.

evalica:
Yeap, I also fixed the other places where we had Headings for .modal-title and .panel-title (Export was using both). Please see individual commits that I've made.

tmortagne:
I'm fine with that, I just mean the you should probably change the issue title :)

evalica:
I've changed the title, we should make sure when merging to Squash and change the commit title. Thanks.

evalica:
Thank you for the review.

PR #895, opened 2018-10-16
Update from Weblate.

xwikiorgci:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources_fr.properties
```diff
@@ -2897,6 +2897,7 @@ platform.livetable.loading=Chargement...
 platform.livetable.tagsHelp=Cliquez sur un ou plusieurs tags afin de filtrer la liste
 platform.livetable.tagsHelpCancel=et cliquez \u00E0 nouveau sur un tag pour le d\u00E9s\u00E9lectionner
 platform.livetable.environmentCannotLoadTableMessage=L'environnement emp\u00EAche le tableau de charger ses donn\u00E9es.
+platform.livetable.docTitleComputedHint=Le titre de certaines pages est calcul\u00E9, filtrer et trier sur le titre de ces pages ne fonctionnera pas.
```
atallahade: Not sure what this mean.

gdelhumeau: Livetable generates a SQL query with some restrictions (WHERE a = b) and ordering (ORDER BY c). Here, we are explaining that we cannot filter or order on the title of the pages, because the title is rendered with Velocity, so we just cannot use standard SQL queries on it (you will order on the velocity code instead of the title output)

atallahade: What I meant is that the **French** translation was not understandable. I proposed another one here:
https://github.com/xwiki/xwiki-platform/pull/895/commits/85237acace2836d97f56f2734e407f0c9772d814#diff-14a83b37fa30ccf9ab30899ffa136b0aR68

gdelhumeau: +1

PR #898, opened 2018-10-18
XWIKI-14334: Sometimes HTML exporting fails and redirects to xwiki/pages/xwiki/Main/Search.html

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-14334

## Proposed solution

  * Stub the request response when rendering page for html, to avoid the
response to be closed.


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/export/html/HtmlPackager.java
```diff
@@ -351,6 +352,8 @@ private XWikiContext initializeContexts(ExecutionContextManager ecm, Execution e
         // Use the filesystem URL format for all code using the url module to generate URLs (webjars, etc).
         Utils.getComponent(URLContextManager.class).setURLFormatId("filesystem");
 
+        // We don't want the rendering of pages to close the current request
```
vmassol: The message should actually be that we don't want rendering of pages to output anything in the current servlet response since we'll send the zip in that response.

surli:
@vmassol I just edited the comment with your suggestion. It's ready for me.

PR #901, opened 2018-10-23
XWIKI-15444: Export UI does not export child pages if there are many (and tree has pagination)

surli:
* Fix XARExportTest flickering test

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/sking/test/po/ExportModal.java
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/sking/test/po/ExportModal.java
@@ -23,8 +23,6 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.xwiki.test.ui.po.BaseModal;
 
-import static org.xwiki.flamingo.sking.test.po.OtherFormatPane.CONTAINER_TREE_CLASS;
-
 /**
  * Represents the Export modal.
  *
@@ -42,7 +40,6 @@ public class ExportModal extends BaseModal
     public ExportModal()
     {
         super(By.id(EXPORT_MODAL_ID));
-        getDriver().waitUntilElementIsVisible(By.id(EXPORT_MODAL_ID));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


surli:
@mflorea this one's ready for review and should fix the XARExportTest which is failing on the CI.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/sking/test/po/ExportModal.java
```diff
@@ -49,8 +51,15 @@ public ExportModal()
      */
     public OtherFormatPane openOtherFormatPane()
     {
-        getDriver().findElementByLinkText(OTHER_FORMAT_PANE_LINK).click();
-        getDriver().waitUntilElementIsVisible(By.cssSelector("#exportModelOtherCollapse .panel-body"));
+        WebElement linkPane = getDriver().findElementByLinkText(OTHER_FORMAT_PANE_LINK);
+        linkPane.click();
+
+        // sometimes it looks like the first click does not work.
```
mflorea: This looks like a hack. Would be nice to understand why the first click doesn't work. Maybe the link is moving (e.g. if there is some animation). AFAIK the click is done in 2 steps:

* it first gets the rectangular area that fits the button / link and computes the center point
* sends a click event in the center point

This kind of simulates the user behavior:

* you locate the button visually on the page and move the mouse over it
* you click

The issue is that the button / link may be moved between the steps, so you end up clicking on something else.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/sking/test/po/ExportModal.java
```diff
@@ -49,8 +51,15 @@ public ExportModal()
...
+        if (linkPane.getAttribute("class").contains("collapsed")) {
+            linkPane.click();
+        }
+        By paneContent = By.className(CONTAINER_TREE_CLASS);
+        getDriver().waitUntilElementIsVisible(paneContent);
```
mflorea: AFAICS the pane content is displayed right away but there is some animation and some opacity transition. If you check the class it changes from:

    collapse -> collapsing -> collapse in

So you should probably wait until the pane content has class "collapse in".

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/sking/test/po/ExportModal.java
surli: It's absolutely a hack: I really don't understand why sometimes the click don't work the first time. Might be related to the fading of the modal (even though it should be managed in the modal PO). Might be the accordion, but don't see why.
It's the only solution I found to make the test passing 10 times on 10.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/sking/test/po/ExportModal.java
surli: I can change that, but anyway the bug is not related to that: I got a 20 sec timeout and when looking on the screenshot of the failing test, you can see that the pane is not displayed at all, as if the click hasn't been done.

PR #902, opened 2018-10-24
Recover username and pasword for global users from subwikis

ppantiru:
[XWIKI-15682: Global users can't recover their password from a subwiki](https://jira.xwiki.org/browse/XWIKI-15682)
[XWIKI-15681: Global users can't recover their login from a subwiki](https://jira.xwiki.org/browse/XWIKI-15681)

- pull request for both issues since they are related and should behave similarly
- fix aims to mimic the behaviour of the login action (first local user then global, if local does not exist)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ForgotUsername.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ForgotUsername.xml
@@ -49,6 +49,10 @@
 
 #else
   #set($results = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]).execute())
+  ## If local user does not exist check global user
+  #if($results.size() == 0)
+    #set($results = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]).setWiki('xwiki').execute())
+  #end
   #if($results.size() == 0)
     {{translation key="xe.admin.forgotUsername.error.noAccount"/}}
 
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ForgotUsername.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ForgotUsername.xml
@@ -48,10 +48,11 @@
 {{/html}}
 
 #else
-  #set($results = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]).execute())
+  #set($query = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]))
+  #set($results = $query.execute())
   ## If local user does not exist check global user
-  #if($results.size() == 0)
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ForgotUsername.xml
```diff
@@ -49,6 +49,10 @@
 
 #else
   #set($results = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]).execute())
+  ## If local user does not exist check global user
+  #if($results.size() == 0)
+    #set($results = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]).setWiki('xwiki').execute())
```
Enygma2002: 2 issues to avoid:
* Code duplication of the query that is created in the line above. You should reuse the query generated in the above line and execute it again, but for a different (main wiki)
* Hardcoding things (like the main wiki's name, `xwiki`). Instead, we should obtain it from `$xcontext.mainWikiName`

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
```diff
@@ -73,6 +73,15 @@ u = user account sent in the form
   ## Check if the user exists and has a valid email address configured in his profile
   #set ($userObj = '')
   #set ($userObj = $userDoc.getObject('XWiki.XWikiUsers'))
+  ## If local user does not exist check global user
+  #if (!$userObj)
+    #if ($userName.indexOf('.') != -1)
+      #set ($userDoc = $xwiki.getDocumentAsAuthor("xwiki:${userName}"))
+    #else
+      #set ($userDoc = $xwiki.getDocumentAsAuthor("xwiki:XWiki.${userName}"))
```
Enygma2002: Same note about hardcoding wiki names, both on this line and the one above.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ForgotUsername.xml
```diff
@@ -49,6 +49,10 @@
 
 #else
   #set($results = $services.query.hql(", BaseObject obj, StringProperty prop where obj.name = doc.fullName and obj.className = 'XWiki.XWikiUsers' and prop.id.id = obj.id and prop.id.name = 'email' and LOWER(prop.value) = ?").bindValues([$email.toLowerCase()]).execute())
+  ## If local user does not exist check global user
+  #if($results.size() == 0)
```
Enygma2002: The case where the current wiki is already the main wiki is missing. We should avoid running the query twice, for the main wiki.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
```diff
@@ -73,6 +73,15 @@ u = user account sent in the form
   ## Check if the user exists and has a valid email address configured in his profile
   #set ($userObj = '')
   #set ($userObj = $userDoc.getObject('XWiki.XWikiUsers'))
+  ## If local user does not exist check global user
+  #if (!$userObj)
```
Enygma2002: Same comment about checking if we are not already on the main wiki, to avoid unnecessary work.

ppantiru:
> Would also be nice to get some more opinions on whether this approach exposes a security problem or not. The idea would be that a main wiki admin would want to disable the "Recover Password/Username" possibility for the main wiki (global) users by denying view rights on the recovery pages, but using this approach, they would still be able to do it from subwikis.

The security issue seems to be a different matter, since the recover password is already possible by using the prefixed full name of the user, and for the username recovery you already need to know the email. This does not seem to impact much the security, but does preserve the consistency with the login.

P.S. made new commits with the requested changes.
P.P.S. Ty for taking the time to review.

Enygma2002:
Thanks for the contribution!

P.S.: Next time, please use separate PRs for separate issues. It's not enough that the 2 behaviors are similar and they should behave consistently to put them in the same PR. They are still 2 separate features/implementations and they have their own code and issue. The only time it's justified to include 2 issues in the same commit/PR is when making a code change fixes/affects 2 features/issues at the same time (i.e. it's common code). The current PR is not in this case, as it affects 2 separate pages, each belonging to their own feature.

PR #905, opened 2018-10-26
XWIKI-15653: Prevent users from deleting/moving/renaming pages containing XClass

surli:
* Fix checkstyle issue

### DIFF
--- a/xwiki-platform-core/xwiki-platform-job/xwiki-platform-job-handler/src/main/java/org/xwiki/job/handler/internal/question/QuestionJobResourceReferenceHandler.java
+++ b/xwiki-platform-core/xwiki-platform-job/xwiki-platform-job-handler/src/main/java/org/xwiki/job/handler/internal/question/QuestionJobResourceReferenceHandler.java
@@ -165,7 +165,8 @@ public class QuestionJobResourceReferenceHandler extends AbstractTemplateJobReso
         try {
             this.beans.populate(question, extractParameters(request, question));
         } catch (PropertyException e) {
-            throw new ResourceReferenceHandlerException("Failed to populate question object for job with id " + jobId, e);
+            throw new ResourceReferenceHandlerException(
+                String.format("Failed to populate question object for job with id [%s]", jobId), e);
         }
 
         // Cancel if supported


xwiki-platform-core/xwiki-platform-job/xwiki-platform-job-handler/src/main/java/org/xwiki/job/handler/internal/question/QuestionJobResourceReferenceHandler.java
```diff
@@ -165,7 +165,7 @@ private void answer(HttpServletRequest request, Job job, List<String> jobId, Obj
         try {
             this.beans.populate(question, extractParameters(request, question));
         } catch (PropertyException e) {
-            throw new ResourceReferenceHandlerException("Failed to populate question object for job with id " + jobId, e);
+            throw new ResourceReferenceHandlerException("Failed to populate question object for job with id " + jobId);
```
vmassol: this is not good, you've removed the information from the nested exception! :)

vmassol: BTW our best practice is to surround with `[...]`, see https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices

surli: Actually I added the information from the nested exception in #885. So I just restore the old behaviour. But if you think it's needed for the future I'll put it back.

vmassol: Yes we need it, we always need to forward the nested exception. Thanks!

surli: I used a String.format then, tell me if it's ok.

vmassol: That's perfect (I didn't dare tell you to use a String.format ;)).

PR #908, opened 2018-10-27
XWIKI-15771: Schema update failing using MSSQL as columns must use ag…

reighnman:
…g functions or be included in the group by.


tmortagne:
Should be fixed by https://jira.xwiki.org/browse/XWIKI-15835. But would be great to confirm it when 10.10RC1 is out.

PR #912, opened 2018-10-30
XWIKI-15703: Missing submit input for .extension-history-source-upload form

surli:
* Add an hidden submit button

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extensionHistory.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extensionHistory.vm
@@ -138,7 +138,7 @@
             <input type="hidden" name="xredirect" value="$xredirect" />
           </div>
           <input type="file" name="filepath" />
-          <input type="submit" name="submit" class="hidden" />
+          <input type="submit" value="$services.localization.render('extension.history.sources.import')" class="btn btn-default" />
         </form>
       </div>
     </div>
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extensionHistory.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extensionHistory.vm
@@ -138,6 +138,7 @@
             <input type="hidden" name="xredirect" value="$xredirect" />
           </div>
           <input type="file" name="filepath" />
+          <input type="submit" name="submit" class="hidden" />
         </form>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


surli:
@evalica if the fix looks ok to you I'll apply the same kind of fix for the similar issues

evalica:
from an HTML point of view it looks fine, but I don't know how the submit button will affect the JS or the form behavior. @mflorea wdyt?

surli:
> You could have added it inside the div.hidden at line 135

Yeah I hesitated as it was hidden fields inside an hidden div, I imagine it could have been used to get all the appropriate hidden fields in some JS. But I might be wrong.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extensionHistory.vm
```diff
@@ -138,6 +138,7 @@
             <input type="hidden" name="xredirect" value="$xredirect" />
           </div>
           <input type="file" name="filepath" />
+          <input type="submit" name="submit" class="hidden" />
```
mflorea: This will make the WCAG test pass but it doesn't mean we satisfy the WCAG rule. If we want to satisfy the rule then we should do something similar to what is done on the Attachments tab:

* we generate a visible submit button from HTML
* we hide the submit button from JavaScript

The effect is that:

* when JavaScript is enabled the file is uploaded automatically without the need to click on the submit button (which is hidden)
* when JavaScript is disabled you need to click on the submit button to upload the file

For this we need to:

* have a proper label (``value``) on the submit button (even if it's hidden)
* no need for a ``name`` since we don't need to have the button name submitted
* hide the button from JavaScript when the upload form is enhanced.

surli: OK for going in this direction. Just a side question: isn't the name of an input mandatory for a WCAG rule?

mflorea: I'm not aware of such a rule. What I know is that the ``name`` is needed if you want to push some data to the server. For instance, if you have multiple submit buttons and you want to know on the server side which one was clicked then you need to use the ``name`` attribute so that the button label (``value``) is sent to the server. Otherwise, if you have a single submit button then you don't need the button label on the server (most of the time).

surli:
Tested with and without JS activated and it seems to work properly.

PR #913, opened 2018-10-30
XWIKI-15704: Missing submit input for #usersandgroupstable form

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15704

## Solution 

 * Add an hidden input submit

### DIFF
--- a/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/AdminWikisRightsSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/AdminWikisRightsSheet.xml
@@ -39,8 +39,7 @@
   <content>{{velocity}}
 ### Sheet used to generically display the XWikiPreferences object fields in the administration sheets.
 {{html}}
-  &lt;form method="post" action="$xwiki.getURL($currentDoc, 'saveandcontinue')" class="xform"&gt;
-    &lt;input type="submit" name="submit" class="hidden" /&gt;
+  &lt;form method="post" class="xform"&gt;
     ############################################################################################
     ## RIGHTS
     ############################################################################################
--- a/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/AdminWikisRightsSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/AdminWikisRightsSheet.xml
@@ -40,6 +40,7 @@
 ### Sheet used to generically display the XWikiPreferences object fields in the administration sheets.
 {{html}}
   &lt;form method="post" action="$xwiki.getURL($currentDoc, 'saveandcontinue')" class="xform"&gt;
+    &lt;input type="submit" name="submit" class="hidden" /&gt;
     ############################################################################################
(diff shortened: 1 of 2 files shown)


evalica:
The HTML part looks ok.

xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/AdminWikisRightsSheet.xml
```diff
@@ -40,6 +40,7 @@
 ### Sheet used to generically display the XWikiPreferences object fields in the administration sheets.
 {{html}}
   &lt;form method="post" action="$xwiki.getURL($currentDoc, 'saveandcontinue')" class="xform"&gt;
+    &lt;input type="submit" name="submit" class="hidden" /&gt;
```
mflorea: Same comment as on https://github.com/xwiki/xwiki-platform/pull/912/files#r229593000 . Moreover, Bootstrap has the ``sr-only`` CSS class that we could use, if we want the screen readers to see the submit button.

surli: ok, thanks for the tip

surli:
So finally here I used only a button with class "sr-only sr-only-focusable": if the JS is deactivated the page is not usable anyway (livetable not displayed at all).

Now I tested it and it works properly: the button is only displayed when going through the form using tab to focus on the different elements.
I had to change the form action, as it was throwing an error when validating the form with the submit button. Now it works fine.

xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/AdminWikisRightsSheet.xml
```diff
@@ -39,7 +39,7 @@
   <content>{{velocity}}
 ### Sheet used to generically display the XWikiPreferences object fields in the administration sheets.
 {{html}}
-  &lt;form method="post" action="$xwiki.getURL($currentDoc, 'saveandcontinue')" class="xform"&gt;
+  &lt;form method="post" class="xform"&gt;
```
mflorea: I'm not sure about this change. Can you give more details about the error you got? It looks as if you're changing the place where the form is submitted and I'm surprised it still works.

mflorea: > CSRF validation failed when saving

It looks like the form is missing the CSRF token. Does it work if you add the hidden input with the form token? (the JavaScript code is probably adding it dynamically when submitting the form).

surli: Indeed. I should have looked better the stacktrace :)
I'll add the hidden input then and put back the form URL. Do you think I should remove the dynamic add of the token in the JS then?

mflorea: > Do you think I should remove the dynamic add of the token in the JS then?

Sure, if it's not too complex.

surli:
This one should now be ready.

PR #918, opened 2018-11-01
Xwiki 15648

acotiuga:


mflorea:
I don't see how the proposed change fixes the problem. The user profile categories **are not loaded asynchronously**. That is the actual issue. The category is put in the query string from JavaScript (using the History API) to be able to bookmark a category and thus to be able to specify the category that should be opened by default. But all the categories are loaded from the start.

tmortagne:
Yes the result is that when you switch to the wikis category it will be empty so it actually cause a huge regression.

PR #922, opened 2018-11-02
XWIKI-15576: Page filter on the livetable

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15576

### Changes

* Add a page filter type on the livetable.
* Use the page filter type for the Deleted Documents table.

### Notes

Some limitations:
* The input will not suggest deleted pages
* If the `multiple` attribute is set, only the first element will be used to make the filter

### Results

livetable

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
@@ -427,6 +427,17 @@
       class="suggest-propertyValues" multiple="multiple" size="1"
       data-className="$!escapetool.xml($xproperty.className)" data-propertyName="$!escapetool.xml($xproperty.name)">
     </select>
+  #elseif ($filterType == 'page')
+    #set ($discard = $xwiki.linkx.use($services.webjars.url('selectize.js', 'css/selectize.bootstrap3.css'),
+      {'type': 'text/css', 'rel': 'stylesheet'}))
+    #set ($discard = $xwiki.ssfx.use('uicomponents/suggest/xwiki.selectize.css', true))
+    #set ($discard = $xwiki.jsfx.use('uicomponents/suggest/suggestPages.js',
+      {'forceSkinAction' : true, 'language' : $xcontext.locale}))
+    <select id="xwiki-livetable-${htmlLiveTableId}-filter-$velocityCount" name="$!escapetool.xml($column)"
+      #if ("$!columnProperties.size" != '')size="$!escapetool.xml($columnProperties.size)"#end
+      #if ("$!columnProperties.multiple" != '')multiple="$!escapetool.xml($columnProperties.multiple)"#end
+      class="suggest-pages">
+    </select>
   #end
 #end
 
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -43,7 +43,7 @@ $xwiki.jsx.use('XWiki.DeletedDocuments')##
 ##
 #set($columns = ["ddoc.fullName", "ddoc.title", "ddoc.date", "ddoc.deleter", 'ddoc.batchId', 'actions'])
 #set($columnsProperties = {
-    'ddoc.fullName'  : { 'type' : 'text', 'size' : 10 },
+    'ddoc.fullName'  : { 'type' : 'page', 'size' : 10 },
     'ddoc.title'     : { 'type' : 'text', 'filterable' : false, 'sortable' : false },
     'ddoc.date'      : { 'type' : 'date'},
     'ddoc.deleter'   : { 'type' : 'text', 'size' : 10 },
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
@@ -1603,6 +1603,17 @@ $Msz MB##
       class="suggest-propertyValues" multiple="multiple" size="1"
       data-className="$!escapetool.xml($xproperty.className)" data-propertyName="$!escapetool.xml($xproperty.name)">
     </select>
+  #elseif ($filterType == 'page')
+    #set ($discard = $xwiki.linkx.use($services.webjars.url('selectize.js', 'css/selectize.bootstrap3.css'),
+      {'type': 'text/css', 'rel': 'stylesheet'}))
+    #set ($discard = $xwiki.ssfx.use('uicomponents/suggest/xwiki.selectize.css', true))
+    #set ($discard = $xwiki.jsfx.use('uicomponents/suggest/suggestPages.js',
+      {'forceSkinAction' : true, 'language' : $xcontext.locale}))
+    <select id="xwiki-livetable-${htmlLiveTableId}-filter-$velocityCount" name="$!escapetool.xml($column)"
+      #if ("$!columnProperties.size" != '')size="$!escapetool.xml($columnProperties.size)"#end
+      #if ("$!columnProperties.multiple" != '')multiple="$!escapetool.xml($columnProperties.multiple)"#end
+      class="suggest-pages">
+    </select>
   #end
 #end
 


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -427,6 +427,17 @@
...
     </select>
+  #elseif ($filterType == 'page')
+    #set ($discard = $xwiki.linkx.use($services.webjars.url('selectize.js', 'css/selectize.bootstrap3.css'),
+      {'type': 'text/css', 'rel': 'stylesheet'}))
+    #set ($discard = $xwiki.ssfx.use('uicomponents/suggest/xwiki.selectize.css', true))
+    #set ($discard = $xwiki.jsfx.use('uicomponents/suggest/suggestPages.js',
+      {'forceSkinAction' : true, 'language' : $xcontext.locale}))
+    <select id="xwiki-livetable-${htmlLiveTableId}-filter-$velocityCount" name="$!escapetool.xml($column)"
+      #if ("$!columnProperties.size" != '')size="$!escapetool.xml($columnProperties.size)"#end
+      #if ("$!columnProperties.multiple" != '')multiple="$!escapetool.xml($columnProperties.multiple)"#end
```
mflorea: The ``multiple`` property is not listed on https://extensions.xwiki.org/xwiki/bin/view/Extension/Livetable+Macro#HParameter24columnsProperties so we didn't need it before. I suppose it could be useful, although I think the users will want to be able to select multiple values, most of the time. The question is what value do you set? Because the [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#Attributes) says:

> When multiple is specified, most browsers will show a scrolling list box instead of a single line dropdown.

So it's enough to be specified, no matter the value. Thus if someone sets ``$columnProperties.multiple`` to ``false`` it will still be specified.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -427,6 +427,17 @@
...
       data-className="$!escapetool.xml($xproperty.className)" data-propertyName="$!escapetool.xml($xproperty.name)">
     </select>
+  #elseif ($filterType == 'page')
+    #set ($discard = $xwiki.linkx.use($services.webjars.url('selectize.js', 'css/selectize.bootstrap3.css'),
+      {'type': 'text/css', 'rel': 'stylesheet'}))
+    #set ($discard = $xwiki.ssfx.use('uicomponents/suggest/xwiki.selectize.css', true))
+    #set ($discard = $xwiki.jsfx.use('uicomponents/suggest/suggestPages.js',
+      {'forceSkinAction' : true, 'language' : $xcontext.locale}))
+    <select id="xwiki-livetable-${htmlLiveTableId}-filter-$velocityCount" name="$!escapetool.xml($column)"
+      #if ("$!columnProperties.size" != '')size="$!escapetool.xml($columnProperties.size)"#end
```
mflorea: Does it matter the size? The ``select`` element is hidden AFAIK. But even if it were visible, having a multi-line select in the table header doesn't look nice (doesn't align with the rest of the filters).

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -43,7 +43,7 @@ $xwiki.jsx.use('XWiki.DeletedDocuments')##
 ##
 #set($columns = ["ddoc.fullName", "ddoc.title", "ddoc.date", "ddoc.deleter", 'ddoc.batchId', 'actions'])
 #set($columnsProperties = {
-    'ddoc.fullName'  : { 'type' : 'text', 'size' : 10 },
+    'ddoc.fullName'  : { 'type' : 'page', 'size' : 10 },
```
mflorea: I'm not sure the new page filter is better (in this case) than the previous text filter. The suggest-based live table filter works best IMO when:

* the user is interested mainly in exact match (you can still do partial matching, but the main use case is exact matching)
* the value you select from the suggest picker appears on multiple rows, i.e. the value is shared by multiple entities listed in the live table, i.e. this value is not the main information from the live table

For instance, if the live table displays the users, it doesn't make sense to have a suggest-based filter for the first column (user name) because the result of the filtering will be zero or one row. On the other hand, if the list table displays pages, then it makes sense to have a suggest-based filter for the page author.

In the case of the page filter, it would make sense if the live table displays entities that have a relation n-to-1 with pages. A live table that displays objects (e.g. comments) or attachments for instance (but even there it's debatable if you want partial matching of exact matching).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
atallahade: I thought about that too and I was thinking that it should be the user responsibility to set the value he wants. If, as you said, someone sets `$columnProperties.multiple` to `false`, he should see right away that "multiple" is in fact enabled. What I mean is that I think it is as much error prone as just setting the `multiple` property in the html tag.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
atallahade: > Does it matter the size?

It really doesn't matter, I thought the property was some kind expected to be set as it was for other selectors.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
mflorea: @evalica WDYT? I don't think it brings much value to use the page autocomplete / suggest on the Deleted Pages / Attachments index. I think the free text filtering is more suited there. Do you agree?

@atallahade I agree that without a real use case it doesn't make sense to implement the live table page picker (it won't be tested and thus it could easily get broken without anyone noticing it).

evalica: First of all I don't think it's nice that Main.AllDocs 'Index' tab is not consistent with 'Deleted Pages'. Index is focused on page titles (as it should, since it's an application targeted towards normal users), while 'Deleted Pages' shows page names. 

I consider that Page filter is a 'component' that should be used for technical users, where it's vital that we specify the 'page name' (usually when referencing class names, etc.) and we cannot use 'page title' instead. I don't believe "Deleted Pages" to be this case. 

Also 'page filter' should be used in cases we don't have filtering present in other ways. In the livetable case, we are duplicating the functionality. The user will see first the result in the 'page filter' popover, then the result will be applied / refreshed in the livetable. For me, this is like an interaction that instead of taking 2 seconds, it's doubling, taking it 4 seconds to receive the same results. 

I don't think adding the 'page filter' in the 'Deleted Pages' will bring any improvements to the normal user. Also I think we should use this filter just for tech / development / advanced user use cases. 

What I would want to see on the 'Deleted Pages' is: 
* Instead of 'Page' and 'Title' columns, to have 'Title' and 'Location' columns instead (consistent with Index tab)
* That the user could filter the 'Title' column, which is not possible now
* We should preserve the 'text' filter on the livetable

We could test the following use case: the user accidentally deleted the test pages from Sandbox. He goes to the Recycle Bin and wants to restore the pages. He knows the pages have in their title something that contains the "test page" words. Help him find those pages :)

evalica: 'Deleted Attachments' should also be made consistent with 'Attachments': have 'Location' column instead of 'Page' (name.

atallahade: I guess the best thing to do is to create dedicated jira issues for what you are describing. I'll close this PR and mark the current issue as "won't do".

evalica: Created https://jira.xwiki.org/browse/XWIKI-15813

PR #926, opened 2018-11-05
XWIKI-15794: The default exported wiki XAR package name is 'XWiki.XWikiPreferences' instead of 'backup'

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15794

## Solution

  * Fix the name of the xar package when no name is given.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
@@ -81,6 +81,15 @@ public class ExportAction extends XWikiAction
     private static String PAGE_SEPARATOR = "&";
 
     /**
+     * Define the different format supported by the export.
+     */
+    private enum EXPORT_FORMAT {
+        XAR,
+        HTML,
+        OTHER
+    }
+
+    /**
      * Manage the arguments of the export and provides them in an object to simplify their usage in the different
      * methods.
      *
@@ -104,7 +113,7 @@ public class ExportAction extends XWikiAction
          */
         private String description;
 
-        ExportArguments(XWikiContext context, String format) throws XWikiException
+        ExportArguments(XWikiContext context, EXPORT_FORMAT format) throws XWikiException
         {
             XWikiRequest request = context.getRequest();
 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -104,20 +104,23 @@
...
             this.name = request.get("name");
-            if (StringUtils.isBlank(name)) {
-                this.name = context.getDoc().getFullName();
-            }
 
             String[] pages = request.getParameterValues("pages");
             String[] excludes = request.getParameterValues("excludes");
 
+            if (StringUtils.isBlank(name)) {
+                if (!format.equals("xar")) {
```
mflorea: You could have used a single ``if`` (with 2 conditions).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -104,20 +104,23 @@
          */
         private String description;
 
-        ExportArguments(XWikiContext context) throws XWikiException
+        ExportArguments(XWikiContext context, String format) throws XWikiException
```
mflorea: An alternative would be to pass the default name (that you can overwrite from the request) instead of the format, but it depends if you think the format might be useful in ``ExportArguments`` in the future for something else than determining the default name.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
surli: indeed

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
surli: Yeah I think the format would be more useful to process the arguments differently in the future. I actually hesitated in creating an enum for the different formats instead of relying on a string... Could be cleaner.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ExportAction.java
```diff
@@ -80,6 +80,15 @@
      */
     private static String PAGE_SEPARATOR = "&";
 
+    /**
+     * Define the different format supported by the export.
+     */
+    private enum EXPORT_FORMAT {
```
mflorea: I think we normally use camel case for the enum type name and upper case for the enum values.

PR #928, opened 2018-11-06
XWIKI-15787: Add a "Delete all" button on the recycle bin

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15787

## PR Scope

This PR only contains the backend of the solution for XWIKI-15787. 

## Job done

  * xwiki-refactoring:
     * New job created for permanent deletion with associated request
     * Abstract concepts for handling of deleted documents for both jobs for permanent deletion and restore of documents
     * Bridge support permanent deletion with different inputs (no input = all, list of IDs or batch ID)
  
 * xwiki-oldcore:
     * RecycleBinStore now supports queries for deleted documents
     * DeleteAction now handle deletion of all documents when a specific parameters is passed, and it creates and executes the appropriate job


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
```diff
@@ -144,6 +152,41 @@ protected boolean delete(XWikiContext context) throws XWikiException
         }
     }
 
+    private boolean deleteAllFromRecycleBin(XWikiContext context) throws XWikiException
+    {
+        RefactoringScriptService refactoring =
+            (RefactoringScriptService) Utils.getComponent(ScriptService.class, "refactoring");
+        PermanentlyDeleteRequest deleteRequest = refactoring.createPermanentlyDeleteRequest(Collections.emptyList());
```
tmortagne: I don't see how this can work for a subwiki since you never indicate the wiki for which you want to empty the trash bin (and I doubt your goal is to empty the trash bin of all wiki, that's definitely not what the job is doing).

tmortagne: Indeed I could not see it, and I just found initializeContext on job side. So it should be OK.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
```diff
@@ -661,4 +662,44 @@ private WikiReference getCurrentWikiReference()
...
     }
+
+    /**
+     * Creates a request to permanently delete a specified batch of deleted documents from the recycle bin.
+     *
+     * @param batchId the ID of the batch of deleted documents to permanently delete
+     * @return the permanently delete request
+     * @since 10.10RC1
+     */
+    public PermanentlyDeleteRequest createPermanentlyDeleteRequest(String batchId)
```
tmortagne: It's missing a method to actually use that PermanentlyDeleteRequest.

surli: The request is actually used by calling a `jobExecutor.execute(type, request)`. But maybe you mean that I should add a method in the RefactoringScriptService to call the executor, like what is done for copy/move/etc? I didn't do it because it wasn't done for restore. I might add it for both of them, if you think there's a value for that.

tmortagne: > The request is actually used by calling a jobExecutor.execute(type, request)

You are talking about a single use case here.

The problem here is that what  createPermanentlyDeleteRequest returns cannot be used from (Velocity) script point of view so this method does not make any sense in a script service. It's also not consistent with the other jobs, and yes this was a mistake in restore case too.

surli: ok. Do you think I should make it private and only expose the method I'll create which will also run the request? And same question for restore?

tmortagne: In theory restore and permanentlyDelete should expose the same kind of API than the other jobs:
* method(s) to create request from some values
* method to run the job with the passed request
* method(s) to rune the job with passed values

surli: I made some changes in that way. As usual the PR is growing, maybe it worth it to make a separate PR for the refactorings of xwiki-platform-refactoring, WDYT?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
```diff
@@ -661,4 +670,116 @@ private WikiReference getCurrentWikiReference()
...
+
+        return request;
+    }
+
+    /**
+     * Schedules an asynchronous job for permanently deleting entities.
+     *
+     * @param request the request to run.
+     * @return the job that has been scheduled and that can be used to monitor the progress of the operation,
+     *  {@code null} in case of failure
```
tmortagne: Don't forget the `@since`.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
tmortagne: > the refactorings of xwiki-platform-refactoring

Not sure what you are referring to, the methods you added to the script services looks OK.

surli: Added.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
```diff
@@ -131,7 +135,11 @@ protected boolean delete(XWikiContext context) throws XWikiException
         XWikiDocument doc = context.getDoc();
 
         String sindex = request.getParameter(RECYCLED_DOCUMENT_ID_PARAM);
-        if (sindex != null && xwiki.hasRecycleBin(context)) {
+        String emptyBin = request.getParameter(EMPTY_RECYCLE_BIN);
+
+        if ("1".equals(emptyBin)) {
```
atallahade: Is there a more generic way to do that? `emptybin=true` should also work right?

surli: Yeah it's actually what's used in https://github.com/xwiki/xwiki-platform/pull/928/files/d6dee0437738b231f426a43d00913860d00c9e7f#diff-70855079f42aabba98e87755fa9e71ecR65 I can change that.

surli:
Rebased the PR after #930

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/job/AbstractDeletedDocumentsJob.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private JobGroupPath groupPath;
+
+    @Override
+    protected AbstractJobStatus<AbstractDeletedDocumentsRequest> createNewStatus(
+        AbstractDeletedDocumentsRequest request)
+    {
+        AbstractJobStatus<AbstractDeletedDocumentsRequest> status = super.createNewStatus(request);
+
+        status.setCancelable(true);
```
tmortagne: Seems to me that would make more sense in AbstractDeletedDocumentsRequest constructor for example.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/UserOrientedRequest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * Define a request that can be processed by a user.
+ *
+ * @version $Id$
+ * @since 10.10RC1
+ */
+@Unstable
+public interface UserOrientedRequest extends Request
```
tmortagne: Does this really need to be public ? I would keep that as an internal refactoring interface.

If a public interface it's really not in the right place since there is nothing related to refactoring in it.

surli: I cannot put it as package protected as I need it in RefactoringScriptService, so yeah I think it's really a public one. 

> If a public interface it's really not in the right place since there is nothing related to refactoring in it.

Any suggestion where I should put it then?

tmortagne: Sorry I was not very clear: I was not talking about the `public` keywork but the fact that this should be in an `*.internal.*` package IMO.

surli: Oh, right. Then I can refactor to put it in `internal/job`.

tmortagne: Hmm actually you seems to expose it in other non internal classes so what I suggested is wrong. So I guess the best would be to move it in a more generic place so that it can be reused by other modules (for example in EM jobs requests since it's the exact same need). Could be a new xwiki-platform-job-security or a less specific xwiki-platform-job-api (I prefer this one I think) for example.

surli: So do you suggest I only move my new interface in this new module, or I also move some stuff related to the requests?

tmortagne: Any generic tool/api related to job and security checking.

surli: About this refactoring, I'd need to keep the same packages for backward compatibility reason, especially for the classes that are not inside "internal" package, shouldn't I?

tmortagne: I'm not really asking any refactoring, just putting in the right place the new generic public APIs you add.

xwiki-platform-core/xwiki-platform-job/xwiki-platform-job-api/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-job</artifactId>
+      <version>${commons.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-wiki-api</artifactId>
```
tmortagne: I don't understand why you need this dependency, you probably meant xwiki-platform-model.

surli: I took it back from xwiki-refactoring-api. Apparently it was used to get the DocumentReference. So yeah xwiki-platform-model should be sufficient.

surli:
@tmortagne just added the changes we discussed. All tests working on my side.

surli:
Tested back after the rebase and it's working properly. For me it's ready. I'll push the frontend on another PR to simplify.

PR #929, opened 2018-11-06
[Misc] Ignore overlays files

atallahade:
We have to exclude this kind of folders, so it would be nice to make the rule more generic. We probably shouldn't exclude all overlays folders as it would bring a limitation but we should exclude the ones that are generated in war modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ nbproject
 .clover
 *~
 **/overlays/*.info
+xwiki-platform-distribution/xwiki-platform-distribution-war/overlays/


tmortagne:
I never saw this folder on my side (and I just tried again and could not reproduce it with `mvn clean install -Pquality`). Could it be something produced by your IDE ?

This filter does not do any harm (I don't see why we would even have a folder like this) but I would like to know what is so bad during the build that it put temporary files outside of target/ (and should ideally be fixed).

mflorea:
I never seen this either.

atallahade:
> Could it be something produced by your IDE ?

It actually seems like that the folder is created automatically by my IDE (IntelliJ IDEA), so it's not related to the maven build.

If everyone agrees, we can do as @vmassol proposed and exclude the "overlays" folders. I didn't liked this solution at first because it's a bit too generic but I don't see any better way.

vmassol:
ok so that explains why I had seen this folder (I use intellij idea) but couldn't reproduce using the maven build. Maybe it got fixed in some versions of IntelliJ.

Found https://intellij-support.jetbrains.com/hc/en-us/community/posts/206975205-Where-does-the-overlays-directory-come-from-

PR #930, opened 2018-11-07
XWIKI-15819: Refactor platform-refactoring to take document author into account.

surli:


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultModelBridge.java
```diff
@@ -439,7 +439,9 @@ public boolean restoreDeletedDocument(long deletedDocumentId, boolean checkConte
                 // Default for now is to skip and log as error to restore over existing documents.
                 logger.error("Document [{}] with ID [{}] can not be restored. Document already exists",
                     deletedDocument.getFullName(), deletedDocumentId);
-            } else if (checkContextUser && !canRestoreDeletedDocument(deletedDocumentId, context.getUserReference())) {
+            } else if (checkContextUser && (!canRestoreDeletedDocument(deletedDocumentId, context.getUserReference())
```
tmortagne: Would probably be better to have two different message depending if the one who does not have the right is the user or the author. Otherwise you end up with "WTF I am admin !".

surli: Right.

surli:
ping @tmortagne it's fixed

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultModelBridge.java
```diff
@@ -440,6 +440,10 @@ public boolean restoreDeletedDocument(long deletedDocumentId, boolean checkConte
                 logger.error("Document [{}] with ID [{}] can not be restored. Document already exists",
                     deletedDocument.getFullName(), deletedDocumentId);
             } else if (checkContextUser && !canRestoreDeletedDocument(deletedDocumentId, context.getUserReference())) {
+                logger.error("The author of this script not allowed to restore document [{}] with ID [{}]",
```
tmortagne: I think it would be even better to be even more explicit and indicate the user reference in the message (like `The author ({}) if this...`).

surli: Done.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/DocumentAccessBridge.java
```diff
@@ -814,4 +814,14 @@ default void pushDocumentInContext(Map<String, Object> backupObjects, DocumentMo
      */
     @Deprecated
     String getCurrentWiki();
+
+    /**
+     * @return the author of the current document.
+     * @since 10.10RC1
```
tmortagne: I think it would be safer to port this in 10.8 and 9.11.x branches too so the corresponding `@since` tags should be added for them.

surli: done.

PR #943, opened 2018-11-12
XWIKI-15852: NPE on the propertyValues REST API end point when a document cannot be viewed

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15852

### Change

* Add checks for NPE.

### Note

The fix is not very sexy, would be great if someone could make suggestions.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/PageClassPropertyValuesProvider.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/PageClassPropertyValuesProvider.java
@@ -95,19 +95,26 @@ public class PageClassPropertyValuesProvider extends AbstractDocumentListClassPr
 
     private String getLabel(EntityReference entityReference)
     {
-        String label;
+        String label = null;
         try {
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/PageClassPropertyValuesProvider.java
```diff
@@ -95,19 +95,26 @@ protected PropertyValues getAllowedValues(PageClass pageClass, int limit, String
...
     {
-        String label;
+        String label = null;
         try {
             XWikiContext xcontext = this.xcontextProvider.get();
-            XWikiDocument document;
-            document = xcontext.getWiki().getDocument(entityReference, xcontext).getTranslatedDocument(xcontext);
-            label = document.getRenderedTitle(Syntax.PLAIN_1_0, xcontext);
+            XWikiDocument document = xcontext.getWiki().getDocument(entityReference, xcontext);
+            if (document != null) {
```
tmortagne: In the discussion we had I tough you were using the public `com.xpn.xwiki.api.XWiki` API like most REST services do but with `com.xpn.xwiki.XWiki#getDocument` it's indeed impossible to end up with null since there is no right check. Either you get a XWikiDocument or it throw an exception.

PR #944, opened 2018-11-13
XWIKI-13282: Allow wiki macros to indicate the type of parameters

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-13282

### Changes

* Add a parameter type to the Wiki Macro class.
* Create a parameter type class.
* Add known parameter types values.

### Notes

I've created a constant list of `WikiMacroParameterType` to avoid duplication but it might not be the best place to initialize it. The problem is that I need it for two different modules (`xwiki-platform-rendering-wikimacro-api` and `xwiki-platform-rendering-wikimacro-store`).
Also I could have used a map instead of a list of `WikiMacroParameterType` to avoid the loop when calling `WikiMacroParameterDescriptor#getType` but it was less readable and the current way is still acceptable in term of performances.


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-api/src/main/java/org/xwiki/rendering/macro/wikibridge/WikiMacroParameterDescriptor.java
```diff
@@ -60,6 +79,11 @@
      */
     private Object defaultValue;
 
+    /**
+     * Type of the parameter.
+     */
+    private String type;
```
tmortagne: This should be a Type, it's not WikiMacroParameterDescriptor job to do String -> Type, it should be done when doing xobject -> WikiMacroParameterDescriptor. Also forget about that very limited mapping, the value in the xobject should be the actual serialized Type ("jav.util.List<Integer>"), you can use org.xwiki.component.util.ReflectionUtils to parse it.

Also your implementation of getType() is wrong, looks as DefaultPropertyDescriptor#getPropertyClass() (except that the default would be String instead of null probably).

atallahade: Thanks @tmortagne, I agree with what you say. I just don't understand why the value of the xobject is `java.util.List<Integer>`, isn't it `String`? Also if I don't use a mapping, how can I provide the possible values in the Document initializer?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-api/src/main/java/org/xwiki/rendering/macro/wikibridge/WikiMacroParameterDescriptor.java
```diff
@@ -152,16 +134,23 @@ public String getDescription()
     @Deprecated
     public Class< ? > getType()
     {
-        return String.class;
+        Class<?> clazz;
```
tmortagne: You could use ReflectionUtils#getTypeClass(Type), I had forgotten about this method.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/WikiMacroParameterClassDocumentInitializer.java
```diff
@@ -55,5 +55,6 @@ protected void createClass(BaseClass xclass)
         xclass.addTextAreaField(PARAMETER_DESCRIPTION_PROPERTY, "Parameter description", 40, 5);
         xclass.addBooleanField(PARAMETER_MANDATORY_PROPERTY, "Parameter mandatory", "yesno");
         xclass.addTextField(PARAMETER_DEFAULT_VALUE_PROPERTY, "Parameter default value", 30);
+        xclass.addTextField(PARAMETER_TYPE_PROPERTY, "Parameter type", 30);
```
tmortagne: 30 might be a bit short for quite a few types.

atallahade:
@tmortagne: This is how it looks like when editing the `WikiMacroParameterClass` object:
image

tmortagne:
OK so I guess the size is not taken into account.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-api/src/main/java/org/xwiki/rendering/macro/wikibridge/WikiMacroParameterDescriptor.java
```diff
@@ -134,16 +134,11 @@ public String getDescription()
     @Deprecated
     public Class< ? > getType()
     {
-        Class<?> clazz;
-        if (this.parameterType instanceof Class) {
-            clazz = (Class) this.parameterType;
-        } else if (this.parameterType instanceof ParameterizedType) {
-            clazz = (Class) ((ParameterizedType) this.parameterType).getRawType();
+        if (this.parameterType == null) {
```
tmortagne: I would change a bit the logic and always call ReflectionUtils.getTypeClass but check the return for null and return String in that case.

PR #947, opened 2018-11-15
XWIKI-15880: Create a button to delete all from the recycle bin

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15880

## Solution

  * Add the button inside the livetable
  * Wait for user confirmation
  * Refresh the livetable once it's done

## Limitations

  * I only supported the button for the livetable, not for the "noscript" part of the view
  * For now the confirmation message displayed is the same as for single document deletion

## Preview

peek 26-11-2018 15-27

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -356,12 +356,18 @@ XWiki.index.trash.documents.confirmDeleteAll = function(event) {
   } else {
     new XWiki.widgets.ConfirmedAjaxRequest(
       /* Ajax request URL */
-      this.href = "$xwiki.getURL($doc.fullName, 'delete', '&amp;confirm=1&amp;emptybin=true')" + "&amp;form_token=$!{services.csrf.getToken()}" + (Prototype.Browser.Opera ? "" : "&amp;ajax=1"),
+      this.href = XWiki.currentDocument.getURL('delete'),
       /* Ajax request parameters */
       {
+        parameters : {
+          'confirm': 1,
+          'emptybin': 'true',
+          'form_token': "$!{services.csrf.getToken()}",
+          'ajax': 1
+        },
         onCreate : function() {
           // Disable the button, to avoid a cascade of clicks from impatient users
-          this.addClassName('disabled');;
+          this.addClassName('disabled');
         }.bind(this),
         onSuccess : function() {
           // Update the livetable
@@ -390,11 +396,11 @@ document.observe('xwiki:dom:loaded', function() {
 
 // we need to wait the complete display of the livetable to get the number of rows
 document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
-  var actionAllContainer = $('documentsTrash').
-      getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
-
   // avoid to get multiple times the button
   if (!$('trashAllDeleted')) {
+    var actionAllContainer = $('documentsTrash').
+    getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
+
     var trashAll = new Element('a', {
       'id': 'trashAllDeleted',
       'class': 'btn btn-danger',
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -389,10 +389,22 @@ XWiki.index.trash.documents.confirmDeleteAll = function(event) {
 document.observe('xwiki:dom:loaded', function() {
   if ($('documentsTrash')) {
     $('documentsTrash').up('.hidden').removeClassName('hidden');
-    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:nth-child(6)')[0];
+  }
+});
 
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:displayComplete', function() {
+  if ($('documentsTrash')) {
+    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
     var trashAllDisabled = (trashDocumentsLivetable.totalRows > 0) ? false : true;
+
+    // avoid to get multiple times the button
+    if ($('trashAllDeleted')) {
+      $('trashAllDeleted').remove();
+    }
+
     var trashAll = new Element('a', {
+        'id': 'trashAllDeleted',
         'class':'btn btn-danger',
         'href':'#',
         'disabled': trashAllDisabled
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -350,7 +350,7 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
  */
 XWiki.index.trash.documents.confirmDeleteAll = function(event) {
   event.stop();
-  if (this.disabled) {
+  if (this.hasClassName('disabled')) {
     // Do nothing if the button was already clicked and it's waiting for a response from the server.
     return;
   } else {
@@ -361,16 +361,12 @@ XWiki.index.trash.documents.confirmDeleteAll = function(event) {
       {
         onCreate : function() {
           // Disable the button, to avoid a cascade of clicks from impatient users
-          this.disabled = true;
+          this.addClassName('disabled');;
         }.bind(this),
         onSuccess : function() {
           // Update the livetable
           trashDocumentsLivetable.refresh();
-        }.bindAsEventListener(this),
-        onComplete : function() {
-          // In the end: re-enable the button
-          this.disabled = false;
-        }.bind(this)
+        }.bindAsEventListener(this)
       },
       /* Interaction parameters */
       {
@@ -394,29 +390,21 @@ document.observe('xwiki:dom:loaded', function() {
 
 // we need to wait the complete display of the livetable to get the number of rows
 document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
-  if ($('documentsTrash')) {
-    var actionAllContainer = $('documentsTrash').
-        getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
+  var actionAllContainer = $('documentsTrash').
+      getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
 
-    if (trashDocumentsLivetable.totalRows > 0) {
-      var trashAllClass = 'btn btn-danger';
-    } else {
-      var trashAllClass = 'btn btn-danger disabled';
-    }
-
-    // avoid to get multiple times the button
-    if ($('trashAllDeleted')) {
-      $('trashAllDeleted').writeAttribute('class', trashAllClass);
-    } else {
-      var trashAll = new Element('a', {
-          'id': 'trashAllDeleted',
-          'class': trashAllClass,
-          'href':'#'
-          }).update('$services.localization.render('index.deletedDocuments.deleteAll')');
-      actionAllContainer.insert(trashAll);
-      trashAll.observe('click', XWiki.index.trash.documents.confirmDeleteAll.bindAsEventListener(trashAll));
-    }
+  // avoid to get multiple times the button
+  if (!$('trashAllDeleted')) {
+    var trashAll = new Element('a', {
+      'id': 'trashAllDeleted',
+      'class': 'btn btn-danger',
+      'href':'#'
+    }).update("$services.localization.render('index.deletedDocuments.deleteAll')");
+    actionAllContainer.insert(trashAll);
+    trashAll.observe('click', XWiki.index.trash.documents.confirmDeleteAll.bindAsEventListener(trashAll));
   }
+
+  $('trashAllDeleted').toggleClassName('disabled', trashDocumentsLivetable.totalRows &lt;= 0);
 });
       </code>
     </property>
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -356,12 +356,12 @@ XWiki.index.trash.documents.confirmDeleteAll = function(event) {
   } else {
     new XWiki.widgets.ConfirmedAjaxRequest(
       /* Ajax request URL */
-      this.href = XWiki.currentDocument.getURL('delete'),
+      XWiki.currentDocument.getURL('delete'),
       /* Ajax request parameters */
       {
         parameters : {
           'confirm': 1,
-          'emptybin': 'true',
+          'emptybin': true,
           'form_token': "$!{services.csrf.getToken()}",
           'ajax': 1
         },
@@ -399,7 +399,7 @@ document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
   // avoid to get multiple times the button
   if (!$('trashAllDeleted')) {
     var actionAllContainer = $('documentsTrash').
-    getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
+      getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
 
     var trashAll = new Element('a', {
       'id': 'trashAllDeleted',
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -393,25 +393,28 @@ document.observe('xwiki:dom:loaded', function() {
 });
 
 // we need to wait the complete display of the livetable to get the number of rows
-document.observe('xwiki:livetable:displayComplete', function() {
+document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
   if ($('documentsTrash')) {
-    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
-    var trashAllDisabled = (trashDocumentsLivetable.totalRows > 0) ? false : true;
+    var actionAllContainer = $('documentsTrash').
+        getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
+
+    if (trashDocumentsLivetable.totalRows > 0) {
+      var trashAllClass = 'btn btn-danger';
+    } else {
+      var trashAllClass = 'btn btn-danger disabled';
+    }
 
     // avoid to get multiple times the button
     if ($('trashAllDeleted')) {
-      $('trashAllDeleted').remove();
-    }
-
-    var trashAll = new Element('a', {
-        'id': 'trashAllDeleted',
-        'class':'btn btn-danger',
-        'href':'#',
-        'disabled': trashAllDisabled
-        }).update('Delete all');
-    actionAllContainer.insert(trashAll);
-    if (!trashAllDisabled) {
-        trashAll.observe('click', XWiki.index.trash.documents.confirmDeleteAll.bindAsEventListener(trashAll));
+      $('trashAllDeleted').writeAttribute('class', trashAllClass);
+    } else {
+      var trashAll = new Element('a', {
+          'id': 'trashAllDeleted',
+          'class': trashAllClass,
+          'href':'#'
+          }).update('$services.localization.render('index.deletedDocuments.deleteAll')');
+      actionAllContainer.insert(trashAll);
+      trashAll.observe('click', XWiki.index.trash.documents.confirmDeleteAll.bindAsEventListener(trashAll));
     }
   }
 });
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -53,7 +53,8 @@ $xwiki.jsx.use('XWiki.DeletedDocuments')##
 #set($options = {
     'url'               : "$xwiki.getURL('XWiki.DeletedDocumentsJSON', 'view', 'list=1&amp;xpage=plain&amp;outputSyntax=plain')",
     'callback'          : 'XWiki.index.trash.documents.displayEntry',
-    'translationPrefix' : 'platform.index.trashDocuments.'
+    'translationPrefix' : 'platform.index.trashDocuments.',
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,15 +342,68 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+  }
+}
 /*
  * The livetable is hidden when javascript is disabled, so that an alternative &lt;noscript&gt; table is displayed.
  * Re-display it when the document loads and Javascript is available.
  */
 document.observe('xwiki:dom:loaded', function() {
   if ($('documentsTrash')) {
     $('documentsTrash').up('.hidden').removeClassName('hidden');
+    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:nth-child(6)')[0];
```
atallahade: Are you sure about this css selector? The element should always be the 6th child if we do that.

surli: Yeah that bothers me too, but I don't know another way how to select this specific cell of the livetable. It needs to be the last column of the livetable. Any other idea?

atallahade:
```suggestion
    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
```

atallahade: Maybe another better way would be to change the livetable to add a special class to the Action column. I don't know if it can easily be done though.

surli: Don't know either. I applied your suggestion for now.

surli:
@mflorea this one's ready if you want to review it

surli:
@mflorea would you have some time to review this one before the release?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,46 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
   }
 }
+
+/**
+ * Event listener, called when clicking the "delete all" button.
+ *
+ * @param event the link activation event
+ */
+XWiki.index.trash.documents.confirmDeleteAll = function(event) {
+  event.stop();
```
mflorea: We're not supposed to use Prototype.js anymore. In this case it is quite easy to write the event listener and to register the event using Require.js and jQuery, which is the recommended approach.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,46 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+  if (this.disabled) {
+    // Do nothing if the button was already clicked and it's waiting for a response from the server.
+    return;
+  } else {
+    new XWiki.widgets.ConfirmedAjaxRequest(
```
mflorea: We started replacing the Prototype.js based confirmation box with the Bootstrap modal. Would be nice to use the Bootstrap modal in new code.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,32 @@ document.observe('xwiki:dom:loaded', function() {
   if ($('documentsTrash')) {
     $('documentsTrash').up('.hidden').removeClassName('hidden');
   }
-});</code>
+});
+
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:displayComplete', function() {
+  if ($('documentsTrash')) {
```
mflorea: See https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/JavaScriptAPI/#HLivetableevents28livetable.js29 and https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js#L324 . You can listen to ``displayComplete`` event of a specific live table.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,32 @@ document.observe('xwiki:dom:loaded', function() {
...
+    if ($('trashAllDeleted')) {
+      $('trashAllDeleted').remove();
+    }
+
+    var trashAll = new Element('a', {
+        'id': 'trashAllDeleted',
+        'class':'btn btn-danger',
+        'href':'#',
+        'disabled': trashAllDisabled
+        }).update('Delete all');
```
mflorea: Needs a translation key.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,32 @@ document.observe('xwiki:dom:loaded', function() {
...
+
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:displayComplete', function() {
+  if ($('documentsTrash')) {
+    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
+    var trashAllDisabled = (trashDocumentsLivetable.totalRows > 0) ? false : true;
+
+    // avoid to get multiple times the button
+    if ($('trashAllDeleted')) {
+      $('trashAllDeleted').remove();
```
mflorea: Why don't you reuse the existing button? Why do you need to recreate it if it exists already?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,32 @@ document.observe('xwiki:dom:loaded', function() {
...
+    }
+
+    var trashAll = new Element('a', {
+        'id': 'trashAllDeleted',
+        'class':'btn btn-danger',
+        'href':'#',
+        'disabled': trashAllDisabled
```
mflorea: Shouldn't the button (link) appear as disabled? I would use a CSS class instead. See https://getbootstrap.com/docs/3.3/css/#anchor-element-1 .

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,32 @@ document.observe('xwiki:dom:loaded', function() {
...
+        }).update('Delete all');
+    actionAllContainer.insert(trashAll);
+    if (!trashAllDisabled) {
```
mflorea: It feels a bit strange to add the listener only when the button is enabled. Normally you don't need to check anything if you use an real button because disabled buttons don't trigger the click event. But since you're using a link (which is OK) there's no native way to disable the link so you need some JavaScript. What I normally do it check the link state from the event listener, which you already do above, so this ``IF`` shouldn't be needed.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,32 @@ document.observe('xwiki:dom:loaded', function() {
   if ($('documentsTrash')) {
     $('documentsTrash').up('.hidden').removeClassName('hidden');
   }
-});</code>
+});
+
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:displayComplete', function() {
+  if ($('documentsTrash')) {
+    var actionAllContainer = $('documentsTrash').getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
```
mflorea: Line too long. Same as in Java, we should use a 120 char limit to make the code more readable.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: Because I need to update it if it already exists. So it's easier to delete it and reuse the code to create it, instead of doing a specific code for updating it.

mflorea: So you need to update only the disabled state?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
mflorea: Both work, but I think the CSS class is more easy to understand in this particular case because you're using a link. At least I didn't know why it works with the attribute until I checked the Bootstrap styles. While when you see the CSS class you assume there is some CSS to make it look disabled. Just my opinion.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: As discussed, I'll do that in another PR.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: I'll change that in another PR.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: fixed.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,35 @@ document.observe('xwiki:dom:loaded', function() {
   if ($('documentsTrash')) {
     $('documentsTrash').up('.hidden').removeClassName('hidden');
   }
-});</code>
+});
+
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
+  if ($('documentsTrash')) {
```
mflorea: Is this still needed?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,35 @@ document.observe('xwiki:dom:loaded', function() {
...
+      $('trashAllDeleted').writeAttribute('class', trashAllClass);
+    } else {
+      var trashAll = new Element('a', {
+          'id': 'trashAllDeleted',
+          'class': trashAllClass,
+          'href':'#'
+          }).update('$services.localization.render('index.deletedDocuments.deleteAll')');
+      actionAllContainer.insert(trashAll);
+      trashAll.observe('click', XWiki.index.trash.documents.confirmDeleteAll.bindAsEventListener(trashAll));
+    }
```
mflorea: I find the above code a bit overly complicated. I would do this instead:

* get the button
* if the button doesn't exist then create it (enabled)
* update the button state (enabled / disabled) with:

        trashAll.toggleClassName('disabled', trashDocumentsLivetable.totalRows > 0);

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +390,35 @@ document.observe('xwiki:dom:loaded', function() {
...
+
+    // avoid to get multiple times the button
+    if ($('trashAllDeleted')) {
+      $('trashAllDeleted').writeAttribute('class', trashAllClass);
+    } else {
+      var trashAll = new Element('a', {
+          'id': 'trashAllDeleted',
+          'class': trashAllClass,
+          'href':'#'
+          }).update('$services.localization.render('index.deletedDocuments.deleteAll')');
```
mflorea: You don't get any warning in the server log that the JSX can't be minified? Might be because the JSX is evaluated (Velocity) first and then minified, but that's not the case with JavaScript code coming from the WAR. It's always good to ensure the JavaScript remains valid when you mix it with Velocity.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,46 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
 }
+
+/**
+ * Event listener, called when clicking the "delete all" button.
+ *
+ * @param event the link activation event
+ */
+XWiki.index.trash.documents.confirmDeleteAll = function(event) {
+  event.stop();
+  if (this.disabled) {
```
mflorea: Shouldn't this be updated to use the CSS class instead?

surli: Not sure it's necessary: this check is related to the `this.disabled = true` in the `onCreate`, so it still works. Apparently bootstrap deactivate the link when the `disabled` class is used, so we did not event enter this function.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: Hmmmm I suppose that if the livetable is not loaded then the event cannot be fired. So I guess not.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: Hm my bad I should have used double quotes. But I did not see any warning.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: I'll refactor.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
mflorea: Yes, but the delete button should appear as disabled while the delete operation is in progress. I'm talking about this flow:

* the button is enabled
* the user clicks on it
* this method is called
* you make an AJAX request
* the button should visually appear as disabled until we receive a response

surli: Indeed. I'll check

surli:
@mflorea it's ready again

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,42 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+ */
+XWiki.index.trash.documents.confirmDeleteAll = function(event) {
+  event.stop();
+  if (this.hasClassName('disabled')) {
+    // Do nothing if the button was already clicked and it's waiting for a response from the server.
+    return;
+  } else {
+    new XWiki.widgets.ConfirmedAjaxRequest(
+      /* Ajax request URL */
+      this.href = "$xwiki.getURL($doc.fullName, 'delete', '&amp;confirm=1&amp;emptybin=true')" + "&amp;form_token=$!{services.csrf.getToken()}" + (Prototype.Browser.Opera ? "" : "&amp;ajax=1"),
```
mflorea: I don't understand this syntax. You do an assignment and the result of evaluating the assignment is passed as the first parameter to the ``ConfirmedAjaxRequest`` constructor? ``this`` must be the link. Why do you need to set the link ``href`` property?

Note that if you need to get the URL of the current document in JavaScript you don't need Velocity. You can do

    XWiki.currentDocument.getURL('delete')

Also, the line is too long and not that easy to read. I would pass the request parameters using

    ``parameters: {}``

below. See http://api.prototypejs.org/ajax/ .

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,42 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+      /* Ajax request parameters */
+      {
+        onCreate : function() {
+          // Disable the button, to avoid a cascade of clicks from impatient users
+          this.addClassName('disabled');;
```
mflorea: Double ``;``

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +386,27 @@ document.observe('xwiki:dom:loaded', function() {
   if ($('documentsTrash')) {
     $('documentsTrash').up('.hidden').removeClassName('hidden');
   }
-});</code>
+});
+
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
+  var actionAllContainer = $('documentsTrash').
```
mflorea: I would move this inside the ``IF`` since it is used only there.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: > Why do you need to set the link href property?

Because right now when I build the link, the href property only contains `#`, which is wrong: I should put the right URL using:
`XWiki.currentDocument.getURL('delete')`. I'll change that.

Then I'll still need to add the confirm parameter at that place.

mflorea: > Because right now when I build the link, the href property only contains #, which is wrong

Why is it wrong? Why does it matter the value of the link ``href`` attribute? You're stopping the click event and you're doing an AJAX request. The link is not followed.

surli: > Why is it wrong?

I was thinking about the case when javascript is not available, but anyway the button won't be available in JS is deactivated so indeed it doesn't change anything.

surli:
Last changes applied @mflorea

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,48 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+ */
+XWiki.index.trash.documents.confirmDeleteAll = function(event) {
+  event.stop();
+  if (this.hasClassName('disabled')) {
+    // Do nothing if the button was already clicked and it's waiting for a response from the server.
+    return;
+  } else {
+    new XWiki.widgets.ConfirmedAjaxRequest(
+      /* Ajax request URL */
+      this.href = XWiki.currentDocument.getURL('delete'),
```
mflorea: You're still setting the link ``href``. I thought we agreed that it's useless.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -341,6 +342,48 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+      /* Ajax request parameters */
+      {
+        parameters : {
+          'confirm': 1,
+          'emptybin': 'true',
```
mflorea: Any reason to use a string rather than directly the boolean value?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -349,7 +392,27 @@ document.observe('xwiki:dom:loaded', function() {
...
   }
-});</code>
+});
+
+// we need to wait the complete display of the livetable to get the number of rows
+document.observe('xwiki:livetable:documentsTrash:displayComplete', function() {
+  // avoid to get multiple times the button
+  if (!$('trashAllDeleted')) {
+    var actionAllContainer = $('documentsTrash').
+    getElementsBySelector('td.xwiki-livetable-display-header-filter:last-child')[0];
```
mflorea: Bad indentation.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: I forgot to remove the first part of the line actually.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: None. I can use a boolean.

surli:
Fixed.

surli:
> Thanks again for your patience.

np, thanks for your reviews :)

PR #948, opened 2018-11-15
XWIKI-15864: Deep copy is not working properly

surli:
* Fix the check of authors when copying


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/job/MoveJob.java
```diff
@@ -155,11 +155,33 @@ protected void process(final SpaceReference source, final SpaceReference destina
             public void visit(DocumentReference oldChildReference)
             {
                 DocumentReference newChildReference = oldChildReference.replaceParent(source, destination);
-                maybeMove(oldChildReference, newChildReference);
+                if (MoveJob.this.hasAccess(Right.VIEW, newChildReference)) {
```
tmortagne: I don't understand what this check is doing here. maybeMove is already taking check of checking rights.

surli:
for me this one's ready

PR #949, opened 2018-11-15
XWIKI-15640: Allow executing UI extensions and panels asynchronously

atallahade:
@tmortagne Tried to fix the build, not sure this was the right way to do it.


tmortagne:
I fixed it in https://github.com/xwiki/xwiki-platform/commit/58f6cf4df39143734cd1e28a3381ec9c44bdd596#diff-d918d78308521375009fa3987b2bda01 actually.

PR #950, opened 2018-11-15
[Misc] Scroll to the WebElement in all find methods of XWikiWebDriver

surli:


xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiWebDriver.java
```diff
@@ -633,16 +633,19 @@ public String getTitle()
         return this.wrappedDriver.findElements(by);
     }
 
+    // Make sure the element is visible by scrolling it into view. Otherwise it's possible for example  that the
+    // visible floating save bar would hide the element.
```
vmassol: I would have used a proper javadoc here but it's a detail, I'll apply the PR.

vmassol:
Sounds good, thanks!

PR #952, opened 2018-11-16
XWIKI-15827: Add support for binding a specific input to a macro parameter type

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15827

### Changes

* Create a maven module for HTML Displayers.
* Create a Default HTML Displayer using velocity templates for the render.
* Add related tests.
* Create a script service to use HTML Displayers.
* Set the mutation threshold to 100.
* Set the jacoco threshold to 1.00.


xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/AbstractHTMLDisplayer.java
```diff
@@ -0,0 +1,89 @@
...
+import org.xwiki.template.TemplateManager;
+
+/**
+ * Abstract class to ease the implementation of {@code HTMLDisplayer} using templates.
+ *
+ * @param <T> the type of the {@code HTMLDisplayer}
+ * @version $Id$
+ * @since 10.10RC1
+ */
+public abstract class AbstractHTMLDisplayer<T> implements HTMLDisplayer<T>
```
tmortagne: If it's specific to template based implementation the name of the class should be more specific.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DocumentReferenceHTMLDisplayer.java
```diff
@@ -0,0 +1,42 @@
...
+ * @since 10.10RC1
+ */
+@Component
+@Singleton
+public class DocumentReferenceHTMLDisplayer extends AbstractHTMLDisplayer<DocumentReference>
+{
+    @Override
+    public String getTemplateName()
+    {
+        return "html_displayer_document_reference.vm";
```
tmortagne: The logic of the name might be moved in AbstractHTMLDisplayer in most cases probably: something like `html_displayer/<short class name>.vm` (to discuss) for example (my second comment being that it would be much nicer to use path instead of long file names). And only override getTemplateName() for types not happy with the default logic.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/AbstractHTMLDisplayer.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public String display(T value, Map<String, String> parameters, String mode) throws HTMLDisplayerException
+    {
+        try {
+            ScriptContext scriptContext = scriptContextManager.getCurrentScriptContext();
+            scriptContext.setAttribute("value", value, ScriptContext.ENGINE_SCOPE);
+            scriptContext.setAttribute("parameters", parameters, ScriptContext.ENGINE_SCOPE);
+            scriptContext.setAttribute("mode", mode, ScriptContext.ENGINE_SCOPE);
+
+            return templateManager.render(getTemplateName());
```
tmortagne: I'm wondering if instead of a fixed name we should not do something like trying several possibilities:

* try `html_display/<type>/<mode>.vm`
* then `html_display/<type>.vm`
* then `html_display/<mode>.vm`
* then `html_display/default.vm`

(maybe less or more, just listed everything I could think of quickly)

and only allow as an optional override the extending class to decide the name based on its own criteria.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/test/java/org/xwiki/rendering/display/html/internal/DisplayMacroTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for {@link org.xwiki.displayer.HTMLDisplayer}.
+ *
+ * @version $Id$
+ */
+@ComponentTest
+public class DisplayMacroTest
```
mflorea: Why "Macro"?

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/script/HTMLDisplayerScriptService.java
```diff
@@ -0,0 +1,82 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * Exposes {@link org.xwiki.displayer.HTMLDisplayer}s to scripts.
+ *
+ * @version $Id$
+ * @since 10.10RC1
+ */
+@Unstable
+@Named("htmldisplayer")
```
mflorea: There is already ``$services.display`` so why not have ``$services.display.html``?

tmortagne: +1

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
```diff
@@ -0,0 +1,124 @@
...
+     * The following names will be use in this priority order to find an existing template:
+     * <ul>
+     *   <li>html_displayer/[type]/[mode].vm
+     *   <li>html_displayer/[type].vm
+     *   <li>html_displayer/[mode].vm
+     *   <li>html_displayer/default.vm
+     * </ul>
+     * @return the template name used to make the rendering
+     */
+    protected String getTemplateName(Object value, String mode) {
```
tmortagne: I think it would make more sense in term of performance to have a `Template getTemplate(Object value, String mode)` instead of redoing the template search from the name twice.

atallahade: The idea was to override the method to change the logic on how the name is computed. I guess returning the `Template` instead of its name won't hurt.

tmortagne: Also does it really need to be protected ? DefaultTemplateHTMLDisplayer seems to be used more as a component than a super class.

tmortagne: Actually it's not used yet, I trough I had seen it used in another class.

atallahade: If you want to override it then yes. It's true that DefaultTemplateHTMLDisplayer is not meant to be overriden *anymore* (it was an abstract class at first). I think you are right, I'll make the method `private` and make it return a `Template`.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
```diff
@@ -0,0 +1,124 @@
...
+        String type = value.getClass().getSimpleName().toLowerCase();
```
tmortagne: This won't work well if the value is null which is not really something exotic.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
```diff
@@ -0,0 +1,124 @@
...
+    public String display(Object value, Map parameters) throws HTMLDisplayerException
+    {
+        return display(value, parameters, "view");
+    }
+
+    @Override
+    public String display(Object value, Map parameters, String mode) throws HTMLDisplayerException
+    {
+        try {
+            ScriptContext scriptContext = scriptContextManager.getCurrentScriptContext();
```
tmortagne: Would be cleaner to get rid of those binding in a finally since they only make sense for the template execution.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
atallahade: Right, I'll write a check.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
atallahade: I'm not sure I understand. If we remove the bindings, how can we produce a content based on the given parameters?

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
tmortagne: You might want to separate `AbstractTemplateHTMLDisplayer` which can use the Type indicated by its extending component from `DefaultHTMLDisplayer` which extend `AbstractTemplateHTMLDisplayer` with no explicit Type.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
tmortagne: I think you missed "in a finally". As I said those binding don't make any sense after we executed the template.

atallahade: Ok thanks

atallahade: I'll also add a prefix to the name of this bindings to avoid shadowing.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
atallahade: @tmortagne I'm not sure to understand. Could you give me an example?

tmortagne: I mean that instead of using default HTMLDisplayer as a component other HTML displayers should extend AbstractTemplateHTMLDisplayer so that it knows what is the Type even when the passed value null.

The alternative is to change all the signatures in HTMLDisplayer to also take the Type (and in which case probably get rid of them in HTMLDisplayerManager). I guess another benefit is that it makes multi Type displayers life easier.

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/internal/DefaultTemplateHTMLDisplayer.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * {@inheritDoc}
+     *
+     * Displays the value with the 'view' mode.
+     */
+    @Override
+    public String display(Type type, Object value) throws HTMLDisplayerException
+    {
+        return display(type, value, new HashMap<>());
```
tmortagne: Collections.emptyMap() would be better for performances (not actually created a new object).

PR #954, opened 2018-11-18
Update from Weblate.

xwikiorgci:


vmassol:
@ldmpub This looks like a duplicate of https://github.com/xwiki/xwiki-platform/pull/953 no?

atallahade:
@vmassol It is a duplicate yes but probably coming from a misconfiguration on l10n/weblate side. I'll investigate.

atallahade:
I think I found the issue, it should be fixed now. Closing the PR even though we could wait for new changes to come (which would produce a rebase and thus removing the duplicates commits).

vmassol:
Thanks @atallahade for checking and fixing!

PR #958, opened 2018-11-20
XWIKI-15878: RSS macro should use AbstractBoxMacro instead of BoxMacro

surli:
* Refactor RssMacro to use AbstractBoxMacro instead of BoxMacro

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
@@ -42,7 +42,9 @@ import org.xwiki.rendering.listener.reference.ResourceType;
 import org.xwiki.rendering.macro.AbstractMacro;
 import org.xwiki.rendering.macro.Macro;
 import org.xwiki.rendering.macro.MacroExecutionException;
+import org.xwiki.rendering.macro.box.AbstractBoxMacro;
 import org.xwiki.rendering.macro.box.BoxMacroParameters;
+import org.xwiki.rendering.macro.descriptor.DefaultContentDescriptor;
 import org.xwiki.rendering.macro.rss.RssMacroParameters;
 import org.xwiki.rendering.parser.ParseException;
 import org.xwiki.rendering.parser.Parser;
@@ -61,7 +63,7 @@ import com.sun.syndication.feed.synd.SyndFeed;
 @Component
 @Named("rss")
 @Singleton
-public class RssMacro extends AbstractMacro<RssMacroParameters>
+public class RssMacro extends AbstractBoxMacro<RssMacroParameters>
 {
     /**
      * The name of the CSS class attribute.
@@ -110,7 +112,7 @@ public class RssMacro extends AbstractMacro<RssMacroParameters>
      */
     public RssMacro()
     {
-        super("RSS", DESCRIPTION, RssMacroParameters.class);
+        super("RSS", DESCRIPTION, new DefaultContentDescriptor(DESCRIPTION, false), RssMacroParameters.class);
         setDefaultCategory(DEFAULT_CATEGORY_CONTENT);
     }
 
@@ -128,21 +130,16 @@ public class RssMacro extends AbstractMacro<RssMacroParameters>
         SyndFeed feed = this.romeFeedFactory.createFeed(parameters);
 
         if (parameters.isDecoration()) {
-            BoxMacroParameters boxParameters = new BoxMacroParameters();
             boolean hasImage = parameters.isImage() && (feed.getImage() != null);
-            boxParameters.setCssClass(FEED_CLASS_VALUE);
+            parameters.setCssClass(FEED_CLASS_VALUE);
 
-            if (StringUtils.isNotEmpty(parameters.getWidth())) {
-                boxParameters.setWidth(parameters.getWidth());
-            }
-
-            boxParameters.setBlockTitle(generateBoxTitle("rsschanneltitle", feed));
+            parameters.setBlockTitle(generateBoxTitle("rsschanneltitle", feed));
 
             if (hasImage) {
-                boxParameters.setImage(new ResourceReference(feed.getImage().getUrl(), ResourceType.URL));
+                parameters.setImage(new ResourceReference(feed.getImage().getUrl(), ResourceType.URL));
             }
 
-            result = this.boxMacro.execute(boxParameters, content == null ? StringUtils.EMPTY : content, context);
+            result = super.execute(parameters, content == null ? StringUtils.EMPTY : content, context);
         } else {
             result = Arrays.<Block>asList(new GroupBlock(Collections.singletonMap(CLASS_ATTRIBUTE, FEED_CLASS_VALUE)));
         }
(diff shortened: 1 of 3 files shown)


surli:
ping @vmassol you might want to review this one, since I need to ignore a revapi warning.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -152,6 +149,16 @@ public boolean supportsInlineMode()
         return result;
     }
 
+    @Override
+    protected List<Block> parseContent(RssMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+        if (content == null) {
+            content = StringUtils.EMPTY;
```
tmortagne: I doubt checkstyle like this much. Also I'm not sure what you feel the need to do this since parse() starts with a StringUtils.isEmpty check.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -128,21 +130,16 @@ public boolean supportsInlineMode()
         SyndFeed feed = this.romeFeedFactory.createFeed(parameters);
 
         if (parameters.isDecoration()) {
-            BoxMacroParameters boxParameters = new BoxMacroParameters();
             boolean hasImage = parameters.isImage() && (feed.getImage() != null);
-            boxParameters.setCssClass(FEED_CLASS_VALUE);
+            parameters.setCssClass(FEED_CLASS_VALUE);
```
tmortagne: Same comment I gave for AbstractMessageMacro. In general you should always avoid modifying what you get as input unless you are officially supposed to do it in the design of the API.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
surli: about checkstyle it was ok when running it with a `Pquality` so I guess it's ok. I did this because it was checked and changed in the original execute() method. Now I need it since I don't manage the same way null and empty content in execute.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
surli: I tried to override getClassProperty but I obtained failing test cases, so I keep things simple as I mostly wanted to fix the tests.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -128,21 +130,16 @@ public boolean supportsInlineMode()
...
 
-            if (StringUtils.isNotEmpty(parameters.getWidth())) {
-                boxParameters.setWidth(parameters.getWidth());
-            }
-
-            boxParameters.setBlockTitle(generateBoxTitle("rsschanneltitle", feed));
+            parameters.setBlockTitle(generateBoxTitle("rsschanneltitle", feed));
```
tmortagne: Instead of modifying `parameters` I would introduce something like a `AbstractBoxMacro#getTitle(P parameters, String content, MacroTransformationContext context)` and override it in RssMacro.

surli: So you suggest a new refactoring in AbstractBoxMacro then? We're a little short in time, no?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
tmortagne: > I don't manage the same way null and empty content in execute

I don't understand what you mean. This code is making any null 'content` an empty String so you definitely do the same thing for empty string and null. Also my point is that this check is not needed because parse already deal with that.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
tmortagne: I suggest to move one like in a dedicated protected method, not exactly a big refactoring.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
surli: Was talking about `AbstractBoxMacro#execute`. Here I just kept the check that was already done in `execute(parameters, content == null ? StringUtils.EMPTY : content, context);` as I didn't want to take any risk for a bugfix.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
surli: Sure. But I need to create the dedicated PR in xwiki-rendering, to be reviewed and merged and then this one to be fixed. And the release needs to wait in the meantime...

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
tmortagne: Actually parseContent will never been called with null content because:
* you already check for null content in execute
* AbstractBoxMacro never call parseContent with null content

So I really don't understand why you need this if.

PR #959, opened 2018-11-21
XWIKI-15882: RssMacro should rely on AbstractBoxMacro helper methods

surli:


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/pom.xml
```diff
@@ -57,23 +57,4 @@
       <version>${project.version}</version>
     </dependency>
   </dependencies>
-  <build>
```
tmortagne: Always nice to see that :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,22 +121,49 @@ public boolean supportsInlineMode()
...
+    protected List<? extends Block> getBlockTitle(RssMacroParameters parameters, String content,
+        MacroTransformationContext context)
+    {
+        return generateBoxTitle("rsschanneltitle", getFeed());
+    }
+
+    @Override
+    protected String getClassProperty()
+    {
+        return super.getClassProperty() + " " + FEED_CLASS_VALUE;
```
tmortagne: I would replace `" "` with `' '` since there is only one character.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,22 +121,49 @@ public boolean supportsInlineMode()
...
+        } else {
+            return super.getImageReference(parameters, content, context);
+        }
+    }
+
+    @Override
+    protected List<? extends Block> getBlockTitle(RssMacroParameters parameters, String content,
+        MacroTransformationContext context)
+    {
+        return generateBoxTitle("rsschanneltitle", getFeed());
```
tmortagne: What do you think about doing that only if block title is not set in `parameters` ? I'm thinking that we now expose this parameter and it does not cost much to do something with it instead of completely ignoring it.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,22 +121,49 @@ public boolean supportsInlineMode()
...
+    private SyndFeed getFeed()
+    {
+        return (SyndFeed) this.execution.getContext().getProperty(FEED_PROPERTY);
+    }
+
+    @Override
+    protected ResourceReference getImageReference(RssMacroParameters parameters, String content,
+        MacroTransformationContext context)
+    {
+        if (parameters.isDecoration() && parameters.isImage()) {
```
tmortagne: Same comment than for block title, it does not cost much to actually check if this parameter is already set or not.

tmortagne: Actually forget that, it's not really possible to set box image parameter since it conflict with rss image parameter.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,22 +121,49 @@ public boolean supportsInlineMode()
...
+        return super.getClassProperty() + " " + FEED_CLASS_VALUE;
+    }
+
     @Override
     public List<Block> execute(RssMacroParameters parameters, String content, MacroTransformationContext context)
         throws MacroExecutionException
     {
         List<Block> result;
         SyndFeed feed = this.romeFeedFactory.createFeed(parameters);
+        this.setFeed(feed);
```
tmortagne: Would be nice to explain a bit what we are doing here since it might not be obvious.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
surli: so basically if the super method returns null I return the generateBlockTitle, right?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,22 +121,49 @@ public boolean supportsInlineMode()
...
     }
 
+    private void setFeed(SyndFeed feed)
+    {
+        this.execution.getContext().setProperty(FEED_PROPERTY, feed);
+    }
+
+    private SyndFeed getFeed()
+    {
+        return (SyndFeed) this.execution.getContext().getProperty(FEED_PROPERTY);
```
tmortagne: We need to get rid of the property from the context at the end of the macro execution.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -69,6 +70,8 @@
 
     private static final String FEED_CLASS_VALUE = "rssfeed";
 
+    private static final String FEED_PROPERTY = "feed";
```
tmortagne: "feed" have a big chance of colliding with another property from the ExecutionContext, would be safer to put something more explicit like "RSSMacro.feed".

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
tmortagne: I was thinking about directly checking `parameters` actually to be safe but I guess it's OK too yes.

surli:
@tmortagne be careful to merge https://github.com/xwiki/xwiki-rendering/pull/163 before this one.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,29 +121,83 @@ public boolean supportsInlineMode()
...
+    private void removeContextFeed()
+    {
+        this.execution.getContext().removeProperty(FEED_PROPERTY);
+    }
+
+    @Override
+    protected ResourceReference getImageReference(RssMacroParameters parameters, String content,
+        MacroTransformationContext context)
+    {
+        if (parameters.isDecoration() && parameters.isImage()) {
```
tmortagne: From what I understand you don't need `parameters.isDecoration()` here because if it's false getImageReference won't be called.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-rss/src/main/java/org/xwiki/rendering/internal/macro/rss/RssMacro.java
```diff
@@ -112,29 +121,83 @@ public boolean supportsInlineMode()
...
+            return new ResourceReference(getFeed().getImage().getUrl(), ResourceType.URL);
```
tmortagne: The macro used to check `feed.getImage() != null` but this seems to be vanished.

PR #960, opened 2018-11-21
XWIKI-15666: Validation check broken for 2 consecutive line breaks on translation pages

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15666

## Solution

  * Ignore translations pages for checking rule Rpd3s4

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/CustomDutchWebGuidelinesValidator.java
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/CustomDutchWebGuidelinesValidator.java
@@ -76,6 +76,20 @@ public class CustomDutchWebGuidelinesValidator extends HTML5DutchWebGuidelinesVa
     }
 
     /**
+     * @return true if it's a translation document.
+     */
+    private boolean isTranslationDocument()
+    {
+        if (target instanceof DocumentReferenceTarget) {
+            DocumentReferenceTarget documentReferenceTarget = (DocumentReferenceTarget) target;
+
+            return documentReferenceTarget.getDocumentReference().getName().contains("Translations");
+        } else {
+            return false;
+        }
+    }
+
+    /**
      * Use the p (paragraph) element to indicate paragraphs. Do not use the br (linebreak) element to separate
      * paragraphs.
      */
@@ -83,7 +97,7 @@ public class CustomDutchWebGuidelinesValidator extends HTML5DutchWebGuidelinesVa
     public void validateRpd3s4()
     {
         if (!isPage("XWiki", "XWikiSyntax") && !isPage("XWiki", "XWikiSyntaxParagraphs")
-            && !isPage("XWiki", "XWikiSyntaxGeneralRemarks")) {
+            && !isPage("XWiki", "XWikiSyntaxGeneralRemarks") && !isTranslationDocument()) {
             super.validateRpd3s4();
         }
     }
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/CustomDutchWebGuidelinesValidator.java
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/CustomDutchWebGuidelinesValidator.java
@@ -96,6 +96,9 @@ public class CustomDutchWebGuidelinesValidator extends HTML5DutchWebGuidelinesVa
     @Override
     public void validateRpd3s4()
     {
+        // Exclude translation documents as they are using plain syntax and line breaking are then translated with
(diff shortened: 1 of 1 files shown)


surli:
FTR when running the validation with this change I fix around 140 validation issues.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/CustomDutchWebGuidelinesValidator.java
```diff
@@ -83,7 +97,7 @@ private boolean isPage(String space, String page)
     public void validateRpd3s4()
     {
         if (!isPage("XWiki", "XWikiSyntax") && !isPage("XWiki", "XWikiSyntaxParagraphs")
-            && !isPage("XWiki", "XWikiSyntaxGeneralRemarks")) {
+            && !isPage("XWiki", "XWikiSyntaxGeneralRemarks") && !isTranslationDocument()) {
```
vmassol: Could you explain with a comment why we need to exclude translation documents? Thx

vmassol: Is it because we think that translation documents are technical documents and don't need to be accessible? If so we might as well exclude technical documents, no? But I'm not sure it's right to exclude translations or technical documets from being accessible?

surli: It's because translations documents are using the plain syntax: the linebreaks are then transformed to `<br>`. 
From what I understood in the issue comments, it's expected.
So as they are not meant to be accessed by the users, I thought it would be safe to only exclude them.

vmassol: > So as they are not meant to be accessed by the users, I thought it would be safe to only exclude them.

I don't know where this was discussed. So you're saying that we don't want people with accessibility issues to be able to develop on XWiki (i.e. add translations, modify them). It's a pretty important statement to me and I would prefer that we VOTE on this so that we all agree. WDYT?

vmassol:
@atallahade I would have preferred that you didn't merge this when someone (me in this case ;)) didn't agree with it... Anyway it's done. Let's discuss now about whether it's correct or not and we can rollback if not.

atallahade:
@vmassol I didn't merged it :)

vmassol:
Ah yes approving is not merging, my bad :)

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/CustomDutchWebGuidelinesValidator.java
surli: > I don't know where this was discussed

In the comments of the linked issue. The first proposal was to completely remove this validation since its purpose is not really clear. To quote @evalica: 
> I've run the HTML content through ckeditor-4/accessibility-checker/ and wave.webaim.org and I got no error for these consecutive <br> elements. Also checked the WCAG 2.1 (https://www.w3.org/TR/WCAG/) and didn't found anything specific about this.

> So you're saying that we don't want people with accessibility issues to be able to develop on XWiki (i.e. add translations, modify them).

So basically, no I'm not saying that. I'm saying this specific rule is failing on all our translations pages and is not necessary pertinent even for people with disabilities. 

> I would prefer that we VOTE on this so that we all agree

Agreed on that :)

vmassol: If we don't believe that this test is good for WCAG (and I'm not saying I agree, I'd need to think about it/discuss it) then it should be removed, and not put an exception for translations.

atallahade: I believe we should keep this test as it forces us to use good practices (i.e. not using consecutive `<br>` elements). In this case, the HTML is *generated* from translations documents and that's why maybe we should make an exception here.

tmortagne: An alternative would be to introduce some dedicate parser syntax for translations to generate something nicer when viewing the page instead of the current plain/1.0. We never did it because it's totally useless from translation point of view but it's not very hard to do either.

surli:
Closing as following the discussion in https://markmail.org/message/okihfnpkavhnbkib we would more skip WCAG checks on technical pages.

PR #961, opened 2018-11-22
 XWIKI-15883: Selected page versions range cannot be deleted anymore from History tab

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15883

## Fix

  * Add a missing "s" to identify properly the input

## Test

Add new integration-test to identify properly the behaviour of delete range version

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
@@ -66,11 +66,11 @@ public class CompareVersionsTest extends AbstractTest
      */
     private ViewPage testPage;
 
+    private final static String pageName = "PageWithManyVersions";
+
     @Before
     public void setUp() throws Exception
     {
-        String pageName = "PageWithManyVersions";
-
         // Check if the test page exists.
         testPage = getUtil().gotoPage(getTestClassName(), pageName);
         if (testPage.exists()) {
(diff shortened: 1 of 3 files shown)


surli:
:warning: if I'm sure about the fix (tested manually) I wasn't able to test properly my changes on the integration-test since flavor-test-ui seems broken.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CompareVersionsTest.java
```diff
@@ -162,11 +162,47 @@ public static void tearDownClass() throws Exception
         getDriver().navigate().refresh();
     }
 
+    @Test
+    public void testCompareVersionScenario()
```
tmortagne: It's not a huge fan of huge single test which validate very different stuff.

surli: I didn't see another way to ensure the test for deleting version was executing last.
+ apparently it's a recommended practice for functional test:
> Write scenarios (i.e. write only a few test methods, even only one if you can so that you can have a single fixture) instead of a lot of individual tests as you'd do for unit tests.

(source: https://dev.xwiki.org/xwiki/bin/view/Community/Testing#HFunctionalTesting)

vmassol: I'm a big fan for the reasons we describe in our best practices at https://dev.xwiki.org/xwiki/bin/view/Community/Testing#HBestpractices-1

> Since functional tests take a long time to execute (XWiki to start, Browser to start, navigation, etc) it's important to write tests that execute as fast as possible. Here are some tips:
> Write scenarios (i.e. write only a few test methods, even only one if you can so that you can have a single fixture) instead of a lot of individual tests as you'd do for unit tests.

vmassol: It's also much closer to mimick what the user does when using XWiki. They never stop and recreate an XWiki instance. They just do a single scenario. So this allows to be closer to the reality and can allow to find issues that are caused by the sequential execution of stuff.

surli:
@tmortagne @vmassol the test is now passing.

PR #963, opened 2018-11-22
XWIKI-15882: RssMacro should rely on AbstractBoxMacro helper methods

surli:
* Fix conditions for RssMacro#getImageReference


surli:
@tmortagne last changes for XWIKI-15882

PR #965, opened 2018-11-22
XWIKI-15843: Missing use of translation key in tags.js

surli:
* Add missing translation keys and use them in tags.js


surli:
@atallahade another one

PR #970, opened 2018-11-26
XWIKI-15870: Warning message when copying a page

surli:
## Issue
https://jira.xwiki.org/browse/XWIKI-15870

## Fix
  * Allow a job to skip the check for page deletion


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/job/AbstractEntityJobWithChecks.java
```diff
@@ -50,6 +50,29 @@
      */
     protected Map<EntityReference, EntitySelection> concernedEntities = new HashMap<>();
 
+    /**
+     * Specify if the job needs to check that the entities can be deleted.
+     */
+    private boolean skipChecks;
```
tmortagne: Seems to me it would make more sense to override isSkipChecks() in MoveJob instead of setting this field (to not duplicate an information already stored in the request) but both are valid. WDYT ?

surli: In that case I only define the method `isSkipChecks` without field and setter: it's easier to maintain. I'm ok with that.

PR #971, opened 2018-11-27
[Misc] Add the missing mutation threshold for rendering-async-default module

surli:


tmortagne:
Indeed I forgot about that.

PR #972, opened 2018-11-27
XWIKI-15912: Improper "rename log" string is displayed when user is performing a page copy

surli:
* Create a dedicated CopyJob

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
@@ -19,8 +19,6 @@
  */
 package org.xwiki.refactoring.job;
 
-import org.xwiki.model.reference.EntityReference;
-
 /**
  * A job request that can be used to move a collection of entities to a specified destination. This request can also be
  * used to rename an entity.
@@ -28,17 +26,12 @@ import org.xwiki.model.reference.EntityReference;
  * @version $Id$
  * @since 7.2M1
  */
-public class MoveRequest extends EntityRequest
+public class MoveRequest extends AbstractCopyOrMoveRequest
 {
     /**
      * Serialization identifier.
      */
     private static final long serialVersionUID = 2L;
-
-    /**
-     * @see #getDestination()
-     */
-    private static final String PROPERTY_DESTINATION = "destination";
 
     /**
      * @see #isDeleteSource()
@@ -64,24 +57,6 @@ public class MoveRequest extends EntityRequest
      * @see #isUpdateParentField()
      */
     private static final String PROPERTY_UPDATE_PARENT_FIELD = "updateParentField";
-
-    /**
-     * @return the destination entity, where to move the entities specified by {@link #getEntityReferences()}
-     */
-    public EntityReference getDestination()
-    {
-        return getProperty(PROPERTY_DESTINATION);
-    }
-
-    /**
-     * Sets the destination entity, where to move the entities specified by {@link #getEntityReferences()}.
-     * 
-     * @param destination the destination entity
-     */
-    public void setDestination(EntityReference destination)
-    {
-        setProperty(PROPERTY_DESTINATION, destination);
-    }
 
     /**
      * @return {@code true} if the source entities specified by {@link #getEntityReferences()} should be deleted,
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
@@ -35,18 +35,11 @@ public class MoveRequest extends AbstractCopyOrMoveRequest
 
     /**
      * @see #isDeleteSource()
+     *
+     * @deprecated Since 10.11 the MoveRequest does not handle copy anymore. So it should always delete the source.
      */
+    @Deprecated
     private static final String PROPERTY_DELETE_SOURCE = "deleteSource";
-
-    /**
-     * @see #isUpdateLinks()
-     */
-    private static final String PROPERTY_UPDATE_LINKS = "updateLinks";
-
-    /**
-     * @see #isUpdateLinksOnFarm()
-     */
-    private static final String PROPERTY_UPDATE_LINKS_ON_FARM = "updateLinksOnFarm";
 
     /**
      * @see #isAutoRedirect()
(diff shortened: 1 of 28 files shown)


tmortagne:
Copy and Rename jobs should be more separated in the API. It was a mistake from the start to expose in the API that copy is "a rename where you don't delete the source" since copy+delete is supposed to be a (bad) implementation detail of the rename job. Since you are introducing a new copy job let's do it right this time.

Among other things `copy(MoveRequest request)` should be deprecated and a CopyRequest (which is not linked to MoveRequest) introduced and used by the new copy job.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
```diff
@@ -19,27 +19,20 @@
...
-
 /**
  * A job request that can be used to move a collection of entities to a specified destination. This request can also be
  * used to rename an entity.
  * 
  * @version $Id$
  * @since 7.2M1
  */
-public class MoveRequest extends EntityRequest
+public class MoveRequest extends AbstractCopyOrMoveRequest
```
tmortagne: Seems to me deleteSource related APIs should be deprecated since it does not make much sense for a move now.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
```diff
@@ -102,181 +103,21 @@
...
-     * @param type the type of refactoring
-     * @return an id for a job to perform the specified type of refactoring
-     */
-    private List<String> generateJobId(String type)
+    @Override
+    public void initialize() throws InitializationException
     {
-        String suffix = new Date().getTime() + "-" + ThreadLocalRandom.current().nextInt(100, 1000);
-        return getJobId(type, suffix);
+        this.requestFactory = new RequestFactory(documentAccessBridge, modelContext);
```
tmortagne: Would be cleaner to make RequestFactory and inject it.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
surli: Rename will delete source

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
```diff
@@ -410,7 +251,22 @@ public Job convertToTerminalDocument(DocumentReference documentReference)
...
+    /**
+     * Schedules an asynchronous job to perform the given copy request.
+     *
+     * @param request the copy request to perform
+     * @return the job that has been scheduled and that can be used to monitor the progress of the operation,
+     *         {@code null} in case of failure
+     * @deprecated Use {@link #copy(CopyRequest)} instead.
+     */
+    @Deprecated
     public Job copy(MoveRequest request)
```
tmortagne: Shouldn't this be moved to legacy too ?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
surli: Actually we discussed it with @vmassol and it seemed more interesting to keep it like that: we don't really need it as a component since we only need inside RefactoringScriptService. And it only needs two external components for now.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
surli: yep. Sorry it's still a WIP I forgot to update the title.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/MoveRequest.java
tmortagne: Yes but point was more that there is no real use case where we don't want to remove the source anymore.

surli: Yeah you're right.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
tmortagne: > we only need inside RefactoringScriptService

It does not really matter IMO.

Another argument is that RefactoringScriptService does not use those components so it will reduce the ClassFanOutComplexity.

surli:
@tmortagne I still need to fix some maven configuration in the legacywar and to execute the functional tests related to move/copy/rename, but else I'd say the refactoring is complete. So ready for review.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
surli: I'm still against since it would mean adding a new interface that we don't really need here. So I'd stay with this for now.

surli: Oh. Then I indeed don't really see a reason why not using a component. I'll change that.

tmortagne: Of course it also need to be added in components.txt ;)

surli:
@tmortagne I fixed the dependency issues with legacy and executed the associated functional tests. Everything's ok so for me it's ready to review/merge.

ldmpub:
Thanks guys for this work ... it sounds strange for me too that rename and copy jobs were not separated until now ;-).

PR #975, opened 2018-11-29
XWIKI-15899: Global users cannot recover their password from a subwiki due to macro errors

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15899

### Change

* Generate the password recovery URL by serializing the user document.
* Avoid NPE.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
@@ -106,7 +106,7 @@ u = user account sent in the form
       #set ($discard = $verifObj.set('verification', $verifStr))
       #set ($discard = $userDoc.saveAsAuthor($services.localization.render('xe.admin.passwordReset.versionComment'), true))
       ## Compose the verification URL
-      #set ($passwordResetURL = $xwiki.getDocument('XWiki.ResetPasswordComplete').getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
+      #set ($passwordResetURL = $xwiki.getDocument("${userDoc.wiki}:XWiki.ResetPasswordComplete").getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
       ## Send the email
       #set ($from = $services.mailsender.configuration.fromAddress)
       #if ("$!from" == '')
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
@@ -106,7 +106,8 @@ u = user account sent in the form
       #set ($discard = $verifObj.set('verification', $verifStr))
       #set ($discard = $userDoc.saveAsAuthor($services.localization.render('xe.admin.passwordReset.versionComment'), true))
       ## Compose the verification URL
-      #set ($passwordResetURL = $xwiki.getDocument("${userDoc.wiki}:XWiki.ResetPasswordComplete").getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
+      #set ($userDocRef = $escapetool.url($services.model.serialize($userDoc.documentReference, 'default')))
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/ResetPassword.xml
```diff
@@ -106,7 +106,7 @@ u = user account sent in the form
       #set ($discard = $verifObj.set('verification', $verifStr))
       #set ($discard = $userDoc.saveAsAuthor($services.localization.render('xe.admin.passwordReset.versionComment'), true))
       ## Compose the verification URL
-      #set ($passwordResetURL = $xwiki.getDocument('XWiki.ResetPasswordComplete').getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
+      #set ($passwordResetURL = $xwiki.getDocument("${userDoc.wiki}:XWiki.ResetPasswordComplete").getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
```
atallahade: I'm assuming the change is safe and the document will always be found.
If someone see an edge case, I can add a fallback when the document is not found to do what was previously done, i.e.:
```suggestion
      #set ($passwordResetURL = $xwiki.getDocument("${userDoc.wiki}:XWiki.ResetPasswordComplete").getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
      #if ($!passwordResetURL == '')
        #set ($passwordResetURL = $xwiki.getDocument("XWiki.ResetPasswordComplete").getExternalURL('view', "u=${userName}&amp;v=${verifStr}"))
      #end
```

surli: but then you'd get back the same error as in the jira issue, won't you? Wouldn't it be enough to just always call 'xwiki:XWiki.ResetPasswordComplete' since, anyway you'll need to change the password in the main wiki.

atallahade: Not really.
There are 2 cases, let's assume your main wiki is `xwiki` and the subwiki `sub`:
1. You are requesting a password change (in `sub`) for `XWiki.Toto` which is an `sub` user
2. You are requesting a password change (in `sub`) for `XWiki.Admin` which is an `xwiki` user

In 1, you need `XWiki.ResetPasswordComplete` (the page which deals with password change) to be in the same wiki as `XWiki.Toto`, which is `sub` so the previous code was working in this case.
In 2, you need `XWiki.ResetPasswordComplete` to use `XWiki.Admin` which is in `xwiki`. The fix I did changed to generated URL to use the `XWiki.ResetPasswordComplete` of the wiki where the user is.

Now another way to fix the issue would be to make `XWiki.ResetPasswordComplete` of any wiki, fetch the user in the right wiki.

atallahade: > the user of the subwiki would still need to change his password in the main wiki?

AFAIK, If the user is global then he can login in both the main and the subwiki, so changing the password in the main wiki does make sense.

> > Now another way to fix the issue would be to make XWiki.ResetPasswordComplete of any wiki, fetch the user in the right wiki.
>
> would it change something for the user?

I need see first if it's possible to do so without any problem but it shouldn't change anything for the user no.

atallahade: Actually what I can do is
```suggestion
      #set ($passwordResetURL = $xwiki.getDocument("XWiki.ResetPasswordComplete").getExternalURL('view', "u=${userDoc.wiki}:${userDoc.fullName}&v=${verifStr}"))
```

vmassol: @atallahade One remark: you should never construct manually a string representation of a reference when there are variable parts. This will leads to bugs. You should use one of the existing model APIs. For ex, imagine that in`${userDoc.wiki}` there's a ":" character. It wouldn't get escaped.

atallahade: @vmassol what if I use `$escapetool.xml`?

atallahade: ok thanks

PR #977, opened 2018-11-30
XWIKI-15825: Use feature and grouping parameters to the Include macro descriptor

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15825

### Changes

* Apply the PropertyGroup and PropertyFeature to some include macro
  parameters.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -74,7 +74,7 @@ public class IncludeMacroParameters
      * @since 3.4M1
      */
     @PropertyDescription("the reference of the resource to display")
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -72,6 +74,8 @@
      * @since 3.4M1
      */
     @PropertyDescription("the reference of the resource to display")
+    @PropertyGroup({"stringReference"})
```
tmortagne: You don't need the {} when there is only one.

atallahade: Good to know, I thought this feature had to be implemented in the annotation itself.

evalica:
This issue is supposed to be fixed in the next release. Can someone else approve the changes? Thanks.

atallahade:
I can merge it without any problem I think. I just wanted to show @vmassol and @mflorea if this was actually what we wanted to do.

PR #982, opened 2018-12-04
XWIKI-15936: 'Skin' page picker is missing on instances created through Distribution Wizard

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15936

### Changes

* Fix the skin property in XWikiPreferences.xml for the DW.
* Fix PageClass properties in several pages.
* Fix the documentBundles property.


tmortagne:
Note that outdated xobject class definition does not have much impact in practice since the import is based on the existing class. The problem with XWikiPreference class is that is was breaking the existing class until next restart.

atallahade:
@tmortagne Could consider removing the outdated xobject class then?

tmortagne:
XWiki Standard does not need it but removing it might cause issue with other stuff reading those files. It's retro compatibility basically like other stuff in the XAR files.

PR #984, opened 2018-12-05
[Misc] Fix pseudo-tested methods in refactoring-api

surli:
The mutation score is now reaching 97% again.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
@@ -142,8 +142,6 @@ public class RequestFactoryTest
         assertTrue(moveRequest.isAutoRedirect());
         assertFalse(moveRequest.isInteractive());
         assertTrue(moveRequest.isCheckRights());
-
-        // check added for fixing pseudo-tested method
         moveRequest.setUpdateParentField(false);
         assertFalse(moveRequest.isUpdateParentField());
     }
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
@@ -30,6 +30,7 @@ import org.xwiki.bridge.DocumentAccessBridge;
 import org.xwiki.model.ModelContext;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.EntityReference;
+import org.xwiki.model.reference.WikiReference;
 import org.xwiki.refactoring.job.CopyRequest;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
```diff
@@ -135,6 +142,10 @@ public void createMoveRequest() throws Exception
         assertTrue(moveRequest.isAutoRedirect());
         assertFalse(moveRequest.isInteractive());
         assertTrue(moveRequest.isCheckRights());
+
+        // check added for fixing pseudo-tested method
```
vmassol: I don't like too much this comment because it feels like we're adding this only for Descartes while we should only add tests because it adds value.

surli: so you propose I just delete the comment?

vmassol: I guess, yes, since what it does seems obvious (even if not very interesting...)

surli:
@vmassol comment removed, it's ready to be merged.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
```diff
@@ -58,14 +59,16 @@
     @MockComponent
     private ModelContext modelContext;
 
-    private DocumentReference userReference = new DocumentReference("wiki", "Users", "Carol");
+    private DocumentReference userReference = new DocumentReference("wiki1", "Users", "Carol");
```
atallahade: Was changing `wiki` to `wiki1` needed?

surli: I wanted to ensure it was really this one which was used for the current wiki reference. "wiki" is the standard name for the main wiki, so it could have been another reference.

vmassol: The standard name fo the main wiki is `xwiki`, not `wiki`(that's why we use `wiki` in the tests actually)

tmortagne: Note sure what you mean. XWiki#DEFAULT_MAIN_WIKI is `xwiki` but RequestFactoryTest does not seems to manipulate any API that would go trough this. I don't where where `wiki` could come from.

surli: My bad I could keep using wiki then.

PR #986, opened 2018-12-06
XWIKI-15912: Improper "rename log" string is displayed when user is performing a page copy

surli:
* Fix copy job

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
@@ -271,7 +271,7 @@ public class RefactoringScriptService implements ScriptService
      */
     public Job copyAs(CopyRequest request)
     {
-        return execute(RefactoringJobs.COPY, request);
+        return execute(RefactoringJobs.COPY_AS, request);
     }
 
     /**
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
@@ -271,7 +271,7 @@ public class RefactoringScriptService implements ScriptService
      */
     public Job copyAs(CopyRequest request)
     {
-        return execute(RefactoringJobs.COPY_AS, request);
+        return execute(RefactoringJobs.COPY, request);
     }
 
     /**
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RequestFactory.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RequestFactory.java
@@ -35,6 +35,7 @@ import org.xwiki.component.annotation.Component;
 import org.xwiki.job.api.AbstractCheckRightsRequest;
 import org.xwiki.model.EntityType;
 import org.xwiki.model.ModelContext;
+import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.EntityReference;
 import org.xwiki.model.reference.WikiReference;
 import org.xwiki.refactoring.job.CopyRequest;
@@ -153,7 +154,17 @@ public class RequestFactory
     {
         CopyRequest request = new CopyRequest();
         this.initEntityRequest(request, RefactoringJobs.COPY, sources);
-        request.setDestination(destination);
+
+        EntityReference dest;
+
+        // in case of a simple copy request, we only want it to be placed under the given destination document.
+        if (destination.getType().equals(EntityType.DOCUMENT)) {
+            DocumentReference documentReference = (DocumentReference) destination;
+            dest = documentReference.getLastSpaceReference();
+        } else {
+            dest = destination;
+        }
+        request.setDestination(dest);
         return request;
     }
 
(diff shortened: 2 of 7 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RefactoringScriptService.java
```diff
@@ -271,7 +271,7 @@ public Job copy(EntityReference source, EntityReference destination)
      */
     public Job copyAs(CopyRequest request)
     {
-        return execute(RefactoringJobs.COPY_AS, request);
+        return execute(RefactoringJobs.COPY, request);
```
tmortagne: If that's the right fix then the method name is very wrong.

surli: We don't have a specific job to distinguish between copy and copyAs. So we have to use the same component.

Now this entire concept of "copy" VS "copyAs" seems to add useless complexity, even the doc does not clearly specify the distinction:


> Copy a Space
>```
>#set ($source = $services.model.resolveSpace('Path.To.Source'))
>#set ($destination = $services.model.resolveSpace('Path.To.New.Parent'))
>$services.refactoring.copy($source, $destination).join()
>```
>Copy a Space As
>```
>#set ($source = $services.model.resolveSpace('Path.To.Source'))
>#set ($destination = $services.model.resolveSpace('Path.To.New.Name'))
>$services.refactoring.copyAs($source, $destination).join()
>```

So we might as well simplify the API by pushing to legacy copyAs.

tmortagne: > We don't have a specific job to distinguish between copy and copyAs.

So this means that the RefactoringJobs.COPY_AS does not make much sense in the first place since its javadoc is false.

surli: > So this means that the RefactoringJobs.COPY_AS does not make much sense in the first place since its javadoc is false.

As it was the case before for RefactoringJobs.COPY :) It was only used for the request, never for the job itself.

surli: ping @mflorea WDYT to put the copyAs to legacy since the distinction does not look really meaningful now?

mflorea: > even the doc does not clearly specify the distinction

The distinction is there and it's clear to me:

* for the **copy** operation you need to specify the **parent** (the location where to copy). The page name doesn't change.
* for the **copy as** operation you need to specify the **new reference**. The page name can change.

For me it makes a lot of sense to have both **copy** and **copyAs** at the level of the script service API, even though both use the same (copy) job under the hood, because **the input is different**.

surli: OK to follow this distinction, but it wasn't clear at all in the code: copyAs was previously calling a RenameJob without deletion instead of a MoveJob for copy, but besides that the destination was set the same way. 
And right now, after the refactoring a RenameJob *is* a CopyJob without deletion: which might mean my CopyJob is actually already a *CopyAsJob*.

So, in order to simplify, I propose to add a new boolean in CopyRequest:
```
keepPageName()
```

if this boolean is true, the CopyJob will work as a Copy, else it will work as a CopyAs. Then calling the `createCopyRequest` or `createCopyAsRequest` will always create a `CopyRequest` but with the proper boolean set. WDYT?

surli: Actually I even don't need it. I just changed the destination in case of a copy in the request factory. Tell me what do you think, I did not test it.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/script/RequestFactory.java
```diff
@@ -153,7 +154,17 @@ public CopyRequest createCopyRequest(Collection<EntityReference> sources, Entity
     {
         CopyRequest request = new CopyRequest();
         this.initEntityRequest(request, RefactoringJobs.COPY, sources);
-        request.setDestination(destination);
+
+        EntityReference dest;
+
+        // in case of a simple copy request, we only want it to be placed under the given destination document.
+        if (destination.getType().equals(EntityType.DOCUMENT)) {
+            DocumentReference documentReference = (DocumentReference) destination;
```
atallahade: I'm not sure you could do this cast. I think we usually write:
```suggestion
            DocumentReference documentReference = new DocumentReference(destination);
```

surli: thanks for your feedback!

surli:
FTR I just checked and the behaviour of **copy** is indeed broken compared to its behaviour before the refactoring I did in #972. The proposed fix here is not working. Still investigating on it.

surli:
So I manage to fix the bug by defining a specific job for CopyAs, since its behaviour is different than for Copy.
I fixed the tests for Copy and added a new test for CopyAs. I also checked by executing the scripts given in the documentation for both requests and it works properly.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/script/RequestFactoryTest.java
```diff
@@ -198,7 +198,7 @@ public void createCopyRequest() throws Exception
             new DocumentReference("code", Arrays.asList("Model", "Entity"), "WebHome");
         CopyRequest copyRequest = requestFactory.createCopyRequest(source, destination);
         assertEquals(Arrays.asList(source), copyRequest.getEntityReferences());
-        assertEquals(destination, copyRequest.getDestination());
+        assertEquals(destination.getLastSpaceReference(), copyRequest.getDestination());
```
tmortagne: `destination.getParent()` return the same thing but is more readable in this context I think

PR #989, opened 2018-12-07
XWIKI-15944: Add ability to pass test resources to Docker-based tests

surli:
* Fix places of the test resources: they should be mounted in browser
container, not in servlet container

### DIFF
--- a/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/AbstractContainerExecutor.java
+++ b/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/AbstractContainerExecutor.java
@@ -60,7 +60,7 @@ public abstract class AbstractContainerExecutor
      * @param targetDirectory the target where to mount the directory in the container.
      * @since 10.11RC1
      */
-    protected static void mountFromHostToContainer(GenericContainer container, String sourceDirectory,
+    protected void mountFromHostToContainer(GenericContainer container, String sourceDirectory,
         String targetDirectory)
     {
         // File mounting is awfully slow on Mac OSX. For example starting Tomcat with XWiki mounted takes
--- a/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/AbstractContainerExecutor.java
+++ b/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/AbstractContainerExecutor.java
@@ -23,6 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.utility.MountableFile;
 
 /**
  * Common code for container execution.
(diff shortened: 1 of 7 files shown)


surli:
@vmassol could you review this one please?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/TestConfiguration.java
```diff
@@ -459,4 +461,13 @@ public String getOutputDirectory()
         }
         return outputDirectory;
     }
+
+    /**
+     * @return the path of the directory containing the test resources on the container.
```
vmassol: I guess you mean "in the container", right?

vmassol: wait, what container? :) This missing.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/XWikiDockerExtension.java
```diff
@@ -84,6 +84,8 @@
 
     private static final String SUPERADMIN = "superadmin";
 
+    private static final String TEST_RESOURCES_HOST_PATH = "target/test-classes";
```
vmassol: Some comment to explain what path this is would be nice.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/XWikiDockerExtension.java
```diff
@@ -247,7 +249,8 @@ private BrowserWebDriverContainer startBrowser(TestConfiguration testConfigurati
             .withDesiredCapabilities(testConfiguration.getBrowser().getCapabilities())
             .withNetwork(Network.SHARED)
             .withNetworkAliases("vnchost")
-            .withRecordingMode(BrowserWebDriverContainer.VncRecordingMode.SKIP, null);
+            .withRecordingMode(BrowserWebDriverContainer.VncRecordingMode.SKIP, null)
+            .withFileSystemBind(TEST_RESOURCES_HOST_PATH, testConfiguration.getTestResourcesPath());
```
vmassol: This is not going to be performant, but it depends on the volume of data. In any case I'd recommend to do what we do for the servlet container.

vmassol: In any case this is missing some documentation too. See the doc you removed below from the Servlet container for example ;)

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/TestConfiguration.java
vmassol: I don't think this is the right place. You should put this in the Browser enum probably (similar to what we do for the other enums - See what you removed from the ServletEngine enum).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/AbstractContainerExecutor.java
```diff
@@ -51,4 +52,25 @@ protected void start(GenericContainer container, TestConfiguration testConfigura
...
     }
+
+    /**
+     * Utility method to mount a directory from the host to the container with some optimization for different OS.
+     * @param container the container on which to mount the directory.
+     * @param sourceDirectory the path of the directory on the host to mount in the container.
+     * @param targetDirectory the target where to mount the directory in the container.
+     * @since 10.11RC1
+     */
+    protected static void mountFromHostToContainer(GenericContainer container, String sourceDirectory,
```
vmassol: I don't like static too much (causes all sort of issues, including for testing when it sets some states).

PR #990, opened 2018-12-10
XWIKI-15956: Suggest Fails To Display Current Value Correctly On Edit

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15956

### Changes

* Implement the getValue method for DBList classes.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/DBListClassPropertyValuesProvider.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/DBListClassPropertyValuesProvider.java
@@ -24,7 +24,10 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.xwiki.component.annotation.Component;
+import org.xwiki.model.reference.ClassPropertyReference;
 import org.xwiki.query.QueryBuilder;
+import org.xwiki.rest.XWikiRestException;
+import org.xwiki.rest.model.jaxb.PropertyValue;
 import org.xwiki.rest.model.jaxb.PropertyValues;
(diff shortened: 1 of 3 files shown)


vmassol:
Thanks @atallahade  that was fast :) Is there some "easy" way to have some tests for this (be them unit or functional)?

atallahade:
@vmassol I'll add some unit tests.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/DBListClassPropertyValuesProvider.java
```diff
@@ -61,4 +64,26 @@ protected PropertyValues getAllowedValues(DBListClass dbListClass, int limit, St
...
+                                dbListClass.getDocumentReference());
+            } catch (Exception e) {
+                throw new XWikiRestException(e);
+            }
+        }
+        if (propertyValue == null) {
+            propertyValue = super.getValue(propertyReference, rawValue);
+        }
+
+        return propertyValue;
```
mflorea: I find this overly complicated. Why not simply call

    getValues(propertyReference, 1, rawValue.toString())

and return the first value from the returned ``PropertyValues``?

atallahade: @mflorea The query in `#getValues` uses this text filter: `text like '%value%'`, hence the first value returned by `#getValues` might not be the one that we are looking for. What we want is `text = 'value'`. (or `text like 'value'`)

mflorea: @atallahade right, it looks good then.

PR #991, opened 2018-12-10
XWIKI-15921: Refactor DeletedDocuments to get rid of prototypejs

surli:
* Add a modal and start refactoring the delete all code

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
@@ -61,6 +61,32 @@ $xwiki.jsx.use('XWiki.DeletedDocuments')##
 ##
 {{html wiki="true" clean="false"}}
 &lt;div class="hidden"&gt;
+    &lt;div class="modal" id="deleteDocumentConfirmModal" tabindex="-1" role="dialog"
+    aria-labelledby="deleteDocumentConfirmModal-label" data-liveTable="#documentsTrash"&gt;
+      &lt;div class="modal-dialog" role="document"&gt;
+        &lt;div class="modal-content"&gt;
+          &lt;div class="modal-header"&gt;
+            &lt;button type="button" class="close" data-dismiss="modal" aria-label="Close"&gt;
+            &lt;span aria-hidden="true"&gt;&amp;times;&lt;/span&gt;
+            &lt;/button&gt;
+            &lt;div class="modal-title" id="deleteDocumentConfirmModal-label"&gt;
+              $services.localization.render('platform.index.trashDocumentsActionsDeleteTooltip')
+            &lt;/div&gt;
+          &lt;/div&gt;
+          &lt;div class="modal-body"&gt;
+            &lt;p&gt;$escapetool.xml($services.localization.render('core.recyclebin.completelyDeleteConfirm'))&lt;/p&gt;
+          &lt;/div&gt;
+          &lt;div class="modal-footer"&gt;
+            &lt;button type="button" class="btn btn-default" data-dismiss="modal"&gt;
+              $escapetool.xml($services.localization.render('cancel'))
+            &lt;/button&gt;
+            &lt;button type="button" class="btn btn-danger" data-dismiss="modal"&gt;
+              $escapetool.xml($services.localization.render('delete'))
+            &lt;/button&gt;
+          &lt;/div&gt;
+        &lt;/div&gt;
+      &lt;/div&gt;
+    &lt;/div&gt;
 #livetable('documentsTrash' $columns $columnsProperties $options)
 &lt;/div&gt;
 {{/html}}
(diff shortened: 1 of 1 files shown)


surli:
:warning: this PR is not ready yet since we have a race condition on `xwiki:livetable:documentsTrash:displayComplete`: the event might be called before the handler is attached to it.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -343,76 +369,60 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
-document.observe('xwiki:dom:loaded', function() {
-  if ($('documentsTrash')) {
-    $('documentsTrash').up('.hidden').removeClassName('hidden');
-  }
+require(['jquery', 'xwiki-meta', 'xwiki-events-bridge'], function($, xm) {
+  //
+  // Delete All
+  //
+  var deleteDocumentConfirmModal = $('#deleteDocumentConfirmModal');
+  deleteDocumentConfirmModal.on('show.bs.modal').find('.btn-danger').click(function() {
```
mflorea: This looks strange:
* ``on()`` expects an event handler function
* you don't need to wait for the modal to be shown to register the click listener
* if you register the click listener each time the modal is shown then it will be called multiple times

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -343,76 +369,60 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+      ));
+      trashDocumentsLivetable.refresh();
+    }).fail(function() {
+      notification.replace(new XWiki.widgets.Notification(
+        $jsontool.serialize($services.localization.render('platform.index.trashDocumentsDeleteFailed')),
+        'error'
+      ));
+    });
+  });
+  $(document).on('xwiki:livetable:documentsTrash:displayComplete', function(event, data) {
```
mflorea: You don't use the ``event`` and ``data`` parameters below so you can move the code in a function that is called:
* initially, if ``trashDocumentsLivetable`` is defined and if ``$('#documentsTrash td.xwiki-livetable-display-header-filter:last-child')`` is not empty (that is if the live table was loaded before this code)
* on ``xwiki:livetable:documentsTrash:displayComplete`` (if the live table is not yet loaded and for the future reloads of the live table)

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -343,76 +369,60 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
-    var trashAll = new Element('a', {
-      'id': 'trashAllDeleted',
-      'class': 'btn btn-danger',
-      'href':'#'
-    }).update("$services.localization.render('index.deletedDocuments.deleteAll')");
-    actionAllContainer.insert(trashAll);
-    trashAll.observe('click', XWiki.index.trash.documents.confirmDeleteAll.bindAsEventListener(trashAll));
+/*
+* The livetable is hidden when javascript is disabled, so that an alternative &lt;noscript&gt; table is displayed.
+* Re-display it when the document loads and Javascript is available.
```
mflorea: I don't understand. AFAIK when you enable / disable the JavaScript code the browser reloads the entire page so I don't see why you need to do something special about it.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: OK so basically I extract the function, I attach it as a handler for the event, and I also call it at the end of the JS to be evaluated once in case the livetable is already loaded.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
surli: That looks indeed weird. I'll change that.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
mflorea: OK.. I haven't seen this done in any other places. Handling no-JavaScript case is definitely not a priority. If we follow this then whenever we use a live table we need to write some JavaScript code to un-hide it, and we use the live table in lots of places. If we really want to support no-JavaScript it should be handled in ``livetable.js`` (but I don't think it is worth). Anyway, it's not that important.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -61,6 +61,32 @@ $xwiki.jsx.use('XWiki.DeletedDocuments')##
 ##
 {{html wiki="true" clean="false"}}
 &lt;div class="hidden"&gt;
+    &lt;div class="modal" id="deleteDocumentConfirmModal" tabindex="-1" role="dialog"
+    aria-labelledby="deleteDocumentConfirmModal-label" data-liveTable="#documentsTrash"&gt;
```
mflorea: Is the ``data-liveTable`` attribute needed?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -343,76 +369,65 @@ XWiki.index.trash.documents.confirmDelete = function(event, table, rowIdx) {
...
+});
 
-  $('trashAllDeleted').toggleClassName('disabled', trashDocumentsLivetable.totalRows &lt;= 0);
+/*
+* The livetable is hidden when javascript is disabled, so that an alternative &lt;noscript&gt; table is displayed.
+* Re-display it when the document loads and Javascript is available.
+*/
+document.observe('xwiki:dom:loaded', function () {
+  if ($('documentsTrash')) {
+      $('documentsTrash').up('.hidden').removeClassName('hidden');
```
mflorea: I think we can move this inside the Require.js block above (converting the code to jQuery).

surli: Actually I tried that, and it wasn't working at all with a `$(document).on('wiki:dom:loaded', ...)`. When I looked on the doc and elsewhere in the wiki it seems this one is always called with a `document.observe`...

surli:
fixed @mflorea

PR #993, opened 2018-12-13
XWIKI-15666: Validation check broken for 2 consecutive line breaks on translation pages

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15666

## Solution

As discussed in https://markmail.org/message/okihfnpkavhnbkib the proposed solution for now is to skip technical pages from WCAG validator.
This has been done by checking a boolean `skipTechnicalPages` in the validator classes: if set to true, then the pages are parsed to get the value of the `hidden` attribute.

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/framework/AbstractValidationTest.java
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/framework/AbstractValidationTest.java
@@ -20,12 +20,18 @@
 package org.xwiki.test.webstandards.framework;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Field;
 import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.regex.Pattern;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
 
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.httpclient.Credentials;
@@ -36,6 +42,8 @@ import org.apache.commons.httpclient.auth.AuthScope;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.commons.httpclient.params.HttpClientParams;
 import org.apache.commons.lang3.StringUtils;
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
 import org.xwiki.model.internal.reference.DefaultStringEntityReferenceSerializer;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.SpaceReference;
@@ -43,7 +51,6 @@ import org.xwiki.model.reference.WikiReference;
 import org.xwiki.validator.Validator;
 import org.xwiki.xar.XarEntry;
 import org.xwiki.xar.XarPackage;
-import org.xwiki.xar.internal.type.DefaultXarEntryType;
 
 import junit.framework.Test;
 import junit.framework.TestCase;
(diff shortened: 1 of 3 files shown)


xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/framework/AbstractValidationTest.java
```diff
@@ -217,6 +218,11 @@ protected static void addXarFiles(Class< ? extends AbstractValidationTest> valid
         }
     }
 
+    private static boolean isNotTechnicalPage(XarEntry entry)
+    {
+        return DefaultXarEntryType.HINT.equals(entry.getEntryType());
```
tmortagne: This is a very different logic. Default type is used to protect pages which are not supposed to be modified by users which is very different from pages user are not supposed to see.

surli:
This is a first proposal for technical page skipping: I'm not sure about this solution for two reasons:
  1. it relies on the XarEntryType to only keep the "default" pages: from what I understood by reading the javadoc, the other types seems quite close to the technical page we want to avoid, but I'm not sure it's 100% accurate. 

  2. here it skips technical pages in all validators, since this filter is statically called. I might add a static flag to allow or not technical pages skipping in the right validator, but it looks not really good.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/framework/AbstractValidationTest.java
surli: OK. I wasn't quite sure about that :)

xwiki-platform-core/xwiki-platform-xar/xwiki-platform-xar-model/src/main/java/org/xwiki/xar/XarPackage.java
```diff
@@ -231,6 +237,35 @@ public void read(ZipFile zipFile) throws IOException, XarException
...
+     * {@link XarFile#getInputStream(LocalDocumentReference)}, else a FileInputStream is created from the file
+     * corresponding to the document reference if it exists.
+     *
+     * @param localDocumentReference the reference for which to return the InputStream.
+     * @return an InputStream corresponding to the given reference.
+     * @throws IOException in case the entry or the file cannot be found.
+     * @throws IOException in case of errors when creating the {@link XarFile}.
+     * @since 10.11RC1
+     */
+    public InputStream getInputStream(LocalDocumentReference localDocumentReference) throws IOException, XarException
```
tmortagne: I really don't like having this kind of API in XarPackage, it's not supposed to be its job.

If you really need a XarFile equivalent for folders then you should introduce one IMO.

surli: > If you really need a XarFile equivalent for folders then you should introduce one IMO.

Actually it's linked to the process of xwiki-platform-distribution-flavor-test-webstandards which I don't fully understand: during the preparation before the test, pages are exported in a folder, but without packaging it into a XAR, and then the `XarPackage.getEntries()` is used to read them.

So maybe we should just package the XAR?

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/framework/AbstractValidationTest.java
```diff
@@ -211,24 +219,75 @@ protected static void addXarFiles(Class< ? extends AbstractValidationTest> valid
...
+        for (DocumentReference documentReference : readXarContents(path, patternFilter, skipTechnicalPages)) {
             suite.addTest(validationTest.getConstructor(Target.class, HttpClient.class, Validator.class, String.class)
                 .newInstance(new DocumentReferenceTarget(documentReference), client, validator, "Admin:admin"));
         }
     }
 
+    private static boolean isTechnicalPage(String directoryPath, XarEntry entry, DocumentBuilder documentBuilder)
+        throws Exception
+    {
+        InputStream inputStream = new FileInputStream(new File(directoryPath, entry.getEntryName()));
```
tmortagne: You don't see to close this stream anywhere.

surli:
fixed.

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/framework/AbstractValidationTest.java
```diff
@@ -211,24 +219,76 @@ protected static void addXarFiles(Class< ? extends AbstractValidationTest> valid
...
+    {
+        try (InputStream inputStream = new FileInputStream(new File(directoryPath, entry.getEntryName()))) {
+            Document parsedDocument = documentBuilder.parse(inputStream);
+
+            NodeList elements = parsedDocument.getElementsByTagName("hidden");
+
+            boolean isHiddenPage = false;
+            if (elements.getLength() == 1) {
+                    isHiddenPage = "true".equals(elements.item(0).getTextContent());
+                }
```
tmortagne: Surprising indentation :)

tmortagne:
Looks OK but as reported by @mflorea in https://markmail.org/message/okihfnpkavhnbkib we might want to still enable some of the hidden pages. Now that you parse the document would be nice to also add at least a class white list. Link input is probably enough for other use cases.

surli:
> Link input is probably enough for other use cases.

not sure what you mean here. But ok I'll add the whitelist to check against className of the document.

tmortagne:
I mean that sure we don't check hidden documents by default but among those hidden documents we would still like to validate documents containing an object with a class which is part of a configured white list (on example being `Panels.PanelClass` because panels are visible stuff we want to validate but they are only visible when selected).

surli:
Just to be sure: if there is at least one object using a class whitelisted, then the whole document should be checked, right?

PR #994, opened 2018-12-13
XWIKI-15961: A page appears more then once in the "My Recent Modifications" when it has more then one translation

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15961

### Change

* Specify the document language when duplicates are shown.

### Result

image

### Note

Maybe a better solution would be to always show the document title using the user language (as it was already the case) and keep the language hint when duplicates are shown.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
@@ -186,8 +186,8 @@
 #end
 #if ($recentDocs.size() &gt; 0 || $showEmptyPanels)
   #panelheader($services.localization.render('xe.panels.modifications.my'))
-  #foreach ($doc in $recentDocs)
-    #set ($recentDoc = $xwiki.getDocument($doc[0]).getTranslatedDocument($doc[1]))
+  #foreach ($recentDocResult in $recentDocs)
+    #set ($recentDoc = $xwiki.getDocument($recentDocResult[0]).getTranslatedDocument($recentDocResult[1]))
     #if ($recentDoc)
       #set ($docLocale = $recentDoc.locale)
       #if ("$docLocale" == '')
@@ -199,7 +199,7 @@
         #set ($docUrl = $recentDoc.getURL())
       #end
       ## We use HTML here because we don't have a tool to escape wiki syntax in document title.
-      * {{html}}&lt;a href="${docUrl}"&gt;$escapetool.xml($recentDoc.plainTitle)#if ($docNamesCounter[$doc[0]] &gt; 1) &lt;em&gt;($docLocale)&lt;/em&gt;#end&lt;/a&gt;{{/html}}
+      * {{html}}&lt;a href="${docUrl}"&gt;$escapetool.xml($recentDoc.plainTitle)#if ($docNamesCounter[$recentDocResult[0]] &gt; 1) &lt;em&gt;($docLocale)&lt;/em&gt;#end&lt;/a&gt;{{/html}}
     #end
   #end
   #panelfooter()
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
@@ -173,7 +173,7 @@
     <property>
       <content>{{velocity}}
 #set ($void = $services.async.useEntity("wiki:$xcontext.database"))
-#set ($query = $services.query.xwql('select doc.fullName, doc.language from XWikiDocument doc where doc.author = :author order by doc.date desc'))
+#set ($query = $services.query.xwql('select doc.fullName, doc.language from Document doc where doc.author = :author order by doc.date desc'))
 #set ($recentDocs = $query.addFilter('hidden').addFilter('unique').bindValue('author', $xcontext.user).setLimit(5).execute())
 ## We count the occurence of a document to specify the document language when there are duplicates
 #set ($docNamesCounter = {})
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
```diff
@@ -173,15 +173,36 @@
     <property>
       <content>{{velocity}}
 #set ($void = $services.async.useEntity("wiki:$xcontext.database"))
-#set ($query = $services.query.xwql('where doc.author = :author order by doc.date desc'))
+#set ($query = $services.query.hql('select doc.fullName, doc.language from XWikiDocument doc where doc.author = :author order by doc.date desc'))
```
tmortagne: You can do a select with xwql too. It's just that it's `Document` instead of `XWikiDocument`.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
```diff
@@ -173,15 +173,36 @@
...
+      #if ($recentDoc)
+        #set ($docLanguage = $recentDoc.language)
+        #if ("$docLanguage" == '')
+          #set ($docLanguage = $recentDoc.defaultLanguage)
+        #end
+        ## We use HTML here because we don't have a tool to escape wiki syntax in document title.
+        * {{html}}&lt;a href="$recentDoc.getURL('view', "language=${docLanguage}")"&gt;$escapetool.xml($recentDoc.plainTitle) &lt;em&gt;($docLanguage)&lt;/em&gt;&lt;/a&gt;{{/html}}
+      #end
+    #else
+      #set ($recentDoc = $xwiki.getDocument($doc[0]).getTranslatedDocument())
```
tmortagne: I don't agree with this logic. Having only one entry does not mean that entry is the one corresponding to the context locale.

atallahade: I thought it might be a good idea to keep the actual behavior and only change it when we have duplicates.
I guess we can decide to always use the document with the contextual language.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
```diff
@@ -173,15 +173,33 @@
...
-    #set ($recentDoc = $xwiki.getDocument($docName).getTranslatedDocument())
+  #foreach ($doc in $recentDocs)
+    #set ($recentDoc = $xwiki.getDocument($doc[0]).getTranslatedDocument($doc[1]))
     #if ($recentDoc)
+      #set ($docLocale = $recentDoc.locale)
+      #if ("$docLocale" == '')
+        #set ($docLocale = $recentDoc.defaultLocale)
+      #end
+      #if ("$docLocale" != '' &amp;&amp; $docLocale != $xcontext.locale)
+        #set ($docUrl = $recentDoc.getURL('view', "language=${docLocale}"))
```
atallahade: I'm not sure if we should use something else than the `Locale.toString` method.

tmortagne: toString() will display the locale id which is what we are doing in the edit panel, on the other hand in the drawer we are displaying the locale display name (see https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_language.vm#L34)

atallahade: Ok I can keep this then.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
```diff
@@ -173,15 +173,33 @@
...
+    #set ($docNamesCounter[$doc[0]] = $docNamesCounter[$doc[0]] + 1)
+  #else
+    #set ($docNamesCounter[$doc[0]] = 1)
+  #end
+#end
 #if ($recentDocs.size() &gt; 0 || $showEmptyPanels)
   #panelheader($services.localization.render('xe.panels.modifications.my'))
-  #foreach ($docName in $recentDocs)
-    #set ($recentDoc = $xwiki.getDocument($docName).getTranslatedDocument())
+  #foreach ($doc in $recentDocs)
```
tmortagne: using $doc in a script is usually not a great idea since it override the standard $doc which is supposed to be the current document

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/MyRecentModifications.xml
```diff
@@ -173,15 +173,33 @@
     <property>
       <content>{{velocity}}
 #set ($void = $services.async.useEntity("wiki:$xcontext.database"))
-#set ($query = $services.query.xwql('where doc.author = :author order by doc.date desc'))
+#set ($query = $services.query.xwql('select doc.fullName, doc.language from XWikiDocument doc where doc.author = :author order by doc.date desc'))
```
tmortagne: You sure this works ? AFAIK you have you are supposed to use `Document` with xwql. Maybe it just go trough as is because the converter does not know it, but would be safer to use the standard from.

atallahade: Both are working yes.

PR #996, opened 2018-12-14
XWIKI-15980: Setting a main wiki or a filesystem skin is much harder with the new picker

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15980

### Change

* Introduce a freeText parameter in the ListClass which indicates that the property can take a simple string as a value.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/meta/ListMetaClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/meta/ListMetaClass.java
@@ -91,6 +91,13 @@ public class ListMetaClass extends PropertyMetaClass
         sortClass.setPrettyName("Sort");
         sortClass.setValues("none|id|value");
         safeput(sortClass.getName(), sortClass);
+
+
+        StaticListClass freeTextClass = new StaticListClass(this);
+        freeTextClass.setName("freeText");
+        freeTextClass.setPrettyName("Free text");
+        freeTextClass.setValues("forbidden|discouraged|allowed");
+        safeput(freeTextClass.getName(), freeTextClass);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/meta/ListMetaClass.java
```diff
@@ -91,6 +91,13 @@ private void addPresentationMetaProperties()
         sortClass.setPrettyName("Sort");
         sortClass.setValues("none|id|value");
         safeput(sortClass.getName(), sortClass);
+
+
+        StaticListClass freeTextClass = new StaticListClass(this);
+        freeTextClass.setName("freeText");
+        freeTextClass.setPrettyName("Free text");
+        freeTextClass.setValues("forbidden|discouraged|allowed");
```
tmortagne: I guess this means that the default for a new class (or when editing an old class) is `forbidden`. Problem is from what I understand this is different from null behavior which seems to be more `discouraged`.

atallahade: AFAICS The default is empty ("------" is selected in the object) which has the same effect as "discouraged". I'll move "discouraged" as the first element, just in case.

tmortagne: > AFAICS The default is empty

OK.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/AbstractDocumentListClassPropertyValuesProvider.java
```diff
@@ -89,11 +92,19 @@ public PropertyValue getValue(ClassPropertyReference propertyReference, Object r
...
+        if (xcontext.getWiki().exists(documentReference, xcontext)) {
+            propertyValue = super.getValue(propertyReference, documentReference);
         }
 
+        if (propertyValue == null) {
+            propertyValue = new PropertyValue();
+        }
+
+        propertyValue.setValue(reference);
+        propertyValue.getMetaData().put("freeText", propertyDefinition.getFreeText());
```
mflorea: Why do you need this meta data? Where is it used? "freeText" is about the property itself (whether it supports free text or not) not about one value of a property (i.e. the value you're returning here).

atallahade: Thanks for checking, this metadata is useless, I forgot to remove this change.

PR #1000, opened 2018-12-17
XWIKI-15979: Global Menu is displayed (with error) in subwiki for a local user

surli:
* Apply the fix to left panels too


atallahade:
#1000 :tada:

PR #1006, opened 2018-12-19
XWIKI-15991: Wrong locale used in multilingual wiki

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15991

### Changes

* Take into account the xwiki.language.forceSupported xwiki parameter.
* Move the logic of XWiki#getLanguagePreference to XWiki#getLocalePreference.

### Note

To actually see the benefits of the improvement, the `xwiki.language.forceSupported` parameter should be set to `1`.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -50,7 +50,6 @@ import java.util.Set;
 import java.util.TimeZone;
 import java.util.Vector;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 import java.util.zip.ZipOutputStream;
 
 import javax.annotation.Priority;
@@ -2817,8 +2816,7 @@ public class XWiki implements EventListener
         // in use.
         try {
             String language = Util.normalizeLanguage(context.getRequest().getParameter("language"));
-            if (!StringUtils.isEmpty(language)) {
-                locale = LocaleUtils.toLocale(language);
+            if (StringUtils.isNotEmpty(language)) {
                 if ("default".equals(language)) {
                     // forgetting language cookie
                     Cookie cookie = new Cookie("language", "");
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2756,20 +2757,28 @@ public String getUserPreference(String prefname, boolean useCookie, XWikiContext
...
-     * @return the locale to use
-     * @since 8.0M1
+     * @param locale the locale to use
+     * @param context the context
+     * @param availableLocales the accepted locales. Used only if {@see forceSupported} is true
+     * @param forceSupported determine if the {@see locale} should be checked against the {@see availableLocales}
+     * @return the locale that has been set or null
      */
-    public Locale getLocalePreference(XWikiContext context)
+    private Locale setLocale(Locale locale, XWikiContext context, Set<Locale> availableLocales, boolean forceSupported)
```
tmortagne: If the passe Locale is completely different from the allowed locales you will end up with a null Locale. Shouldn't it fallback on default Locale instead ? Also you should should stop at Locale.ROOT, not null.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2780,51 +2789,55 @@ public Locale getLocalePreference(XWikiContext context)
...
         // for a language parameter. If the language value is "default" use the default language
         // from the XWiki preferences settings. Otherwise set a cookie to remember the language
         // in use.
         try {
-            language = Util.normalizeLanguage(context.getRequest().getParameter("language"));
-            if ((language != null) && (!language.equals(""))) {
-                if (language.equals("default")) {
+            String language = Util.normalizeLanguage(context.getRequest().getParameter("language"));
+            if (!StringUtils.isEmpty(language)) {
+                locale = LocaleUtils.toLocale(language);
```
tmortagne: Should this line be moved in the `else` ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2833,24 +2846,29 @@ public String getLanguagePreference(XWikiContext context)
         // from a cookie.
         try {
             // First we get the language from the cookie
-            language = Util.normalizeLanguage(getUserPreferenceFromCookie("language", context));
-            if ((language != null) && (!language.equals(""))) {
-                context.setLanguage(language);
-                return language;
+            String language = Util.normalizeLanguage(getUserPreferenceFromCookie("language", context));
+            if (!StringUtils.isEmpty(language)) {
```
tmortagne: There is a `isNotEmpty`.

tmortagne:
Given the size and complexity of this method I think we need a unit test.

PR #1008, opened 2018-12-20
[Misc] Legacy must be build also when distribution profile is activated since distribution needs legacy dependencies

surli:


tmortagne:
> distribution needs legacy dependencies

This is not true actually. As you can see in the distribution WAR it properly take into account if legacy profile is enable or not (in which case it does not include any legacy module).

Looks like someone added legacy modules in the flavor without properly setting up a profile. Maybe that's what misled you.

surli:
> Looks like someone added legacy modules in the flavor without properly setting up a profile.

We currently have a legacy dependency in platform-distribution-flavor-common which is only marked as optional, and which is not bundled inside a profile (see: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-common/pom.xml#L148). There is a comment added at the top of the pom.xml, with the following lines:

>  
<xwiki.enforcer.no-legacy-dependencies.skip>true</xwiki.enforcer.no-legacy-dependencies.skip>

So apparently we assume that we can have legacy dependencies in that module. That explained my proposed change.

tmortagne:
> So apparently we assume that we can have legacy dependencies in that module. That explained my proposed change.

My point was that legacy dependencies are usually supposed to be in a profile, not that there should not be any legacy dependencies.

surli:
ping @tmortagne @vmassol this one's ready.

PR #1012, opened 2019-01-02
XWIKI-8286: Upgrade to Hibernate 5.1.8

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-8286

### Changes

* Upgrade to Hibernate 4.0.1.Final.
* Use the new hibernate package names.
* Replace the 'session.connection()' with either 'session.doWork()' method or by directly creating an sql query.
* Adapt the DBCPConnectionProvider using C3P0ConnectionProvider as a model.

### Notes

Not everything could have been properly tested, especially migration classes.

The migration was made using:
* [The migration guide (3.X to 4.0)](https://developer.jboss.org/wiki/HibernateCoreMigrationGuide40)
* [The hibernate 4.0 javadoc](http://docs.jboss.org/hibernate/orm/4.0/javadocs/allclasses-noframe.html)
* [The source code](https://github.com/hibernate/hibernate-orm)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -1246,13 +1245,12 @@ public void saveXWikiCollection(BaseCollection object, XWikiContext inputxcontex
                 // save object using the custom mapping
                 Map<String, Object> objmap = object.getCustomMappingMap();
                 handledProps = bclass.getCustomMappingPropertyList(context);
-                Session dynamicSession = session.getSession(EntityMode.MAP);
```
atallahade: This is not valid anymore and I'm not sure what will be the consequence of the removal.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -2845,20 +2841,7 @@ private SessionFactory injectCustomMappingsInSessionFactory(XWikiDocument doc, X
             return getSessionFactory();
         }
 
-        Configuration config = getConfiguration();
-        SessionFactoryImpl sfactory = (SessionFactoryImpl) config.buildSessionFactory();
-        Settings settings = sfactory.getSettings();
-        ConnectionProvider provider = ((SessionFactoryImpl) getSessionFactory()).getSettings().getConnectionProvider();
```
atallahade: The `connectionProvider` is not part of the settings anymore so setting it doesn't seem to be needed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
atallahade: See https://hibernate.atlassian.net/browse/HHH-6330

PR #1013, opened 2019-01-04
XWIKI-15199: Uncaught exception when previewing deleted section

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15199

## Solution 

 * Add a supplementary condition when previewing a deleted section.

## Test

tested locally.


surli:
@vmassol @tmortagne this one's ready for review.

vmassol:
Sounds good. Only issue is that this will reduce the global TPC of the xwiki-platform-core module a bit since I don't think we have any test (unit or functional) that will hit this line.

vmassol:
Should I merge on the 10.11.x branch? and 9.11.x one?

vmassol:
I'll merge

surli:
I don't know if it worthes it for 9.11.x but sure for 10.11.x

PR #1014, opened 2019-01-08
XWIKI-16016: Tree issues - deselection of pages acts like the only selection when Exporting as HTML or XAR

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16016

## Solution

  * Create a custom LikeExpression to remove the redundant "not" when
needed
  * Add a test to ensure the behaviour

This is really a hack to wait before JSQLParser is released. Tested locally it seems running.


surli:
Closed since JSqlParser has been released. See #1018

PR #1015, opened 2019-01-08

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-6073

## Proposal
  * xwiki.xwiki#getSkinFile now can take a version to create the url
  * api.xwiki#getSkinFile now use the current wiki version to create the
URLs

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -33,6 +33,10 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.FileTime;
 import java.text.DateFormatSymbols;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -43,6 +47,7 @@ import java.util.Date;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -80,6 +85,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.message.BasicNameValuePair;
 import org.apache.velocity.VelocityContext;
 import org.hibernate.HibernateException;
 import org.slf4j.Logger;
@@ -2433,8 +2441,11 @@ public class XWiki implements EventListener
             }
 
             // Look for a resource file
-            if (resourceExists("/resources/" + filename)) {
-                URL url = urlf.createResourceURL(filename, forceSkinAction, version, context);
+            String resourceFilePath = "/resources/" + filename;
+            if (resourceExists(resourceFilePath)) {
+                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
+                url = new URIBuilder(url.toURI()).addParameters(getResourceURLParameters(resourceFilePath)).build()
+                    .toURL();
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
@@ -2453,23 +2464,34 @@ public class XWiki implements EventListener
         return urlf.getURL(url, context);
     }
 
-    public String getSkinFile(String filename, String skin, XWikiContext context)
+    private List<NameValuePair> getResourceURLParameters(String resourceFilePath)
     {
-        return getSkinFile(filename, skin, context, null);
+        List<NameValuePair> parameters = new LinkedList<>();
+        parameters.add(new BasicNameValuePair("xwikiVersion", this.getVersion()));
+
+        try {
+            URL resource = getResource(resourceFilePath);
+            Path resourcePath = Paths.get(resource.toURI());
+            long size = Files.size(resourcePath);
+
+            if (size != 0) {
+                parameters.add(new BasicNameValuePair("size", String.valueOf(size)));
+            }
+
+            FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
+            parameters.add(new BasicNameValuePair("filedate", String.valueOf(lastModifiedTime.toMillis())));
+        } catch (Exception e) {
+            LOGGER.debug("Error when trying to access properties of resource file [{}]", resourceFilePath, e);
+        }
+        return parameters;
     }
 
-    public String getSkinFile(String filename, String skin, XWikiContext context, String version)
+    public String getSkinFile(String filename, String skin, XWikiContext context)
     {
-        return getSkinFile(filename, skin, false, context, version);
+        return getSkinFile(filename, skin, context);
     }
 
     public String getSkinFile(String filename, String skinId, boolean forceSkinAction, XWikiContext context)
-    {
-        return getSkinFile(filename, skinId, forceSkinAction, context, null);
-    }
-
-    public String getSkinFile(String filename, String skinId, boolean forceSkinAction, XWikiContext context,
-        String version)
     {
         try {
             Skin skin = getInternalSkinManager().getSkin(skinId);
@@ -2484,7 +2506,7 @@ public class XWiki implements EventListener
             if (resourceExists("/resources/" + filename)) {
                 XWikiURLFactory urlf = context.getURLFactory();
 
-                URL url = urlf.createResourceURL(filename, forceSkinAction, version, context);
+                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -2413,6 +2413,13 @@ public class XWiki implements EventListener
         Map<String, String> queryParameters)
     {
         XWikiURLFactory urlf = context.getURLFactory();
+        Map<String,String> queryParametersMap;
+
+        if (queryParameters == null) {
+            queryParametersMap = new HashMap<>();
+        } else {
+            queryParametersMap = new HashMap<>(queryParameters);
+        }
 
         try {
             // Try in the specified skin
@@ -2420,8 +2427,8 @@ public class XWiki implements EventListener
             if (skin != null) {
                 Resource<?> resource = skin.getResource(filename);
                 if (resource != null) {
-                    queryParameters.putAll(getResourceURLParameters(resource));
-                    String resourceUrl = resource.getURL(forceSkinAction, queryParameters);
+                    queryParametersMap.putAll(getResourceURLParameters(resource));
+                    String resourceUrl = resource.getURL(forceSkinAction, queryParametersMap);
                     return resourceUrl;
                 }
             } else {
@@ -2430,8 +2437,8 @@ public class XWiki implements EventListener
                 if (parentSkin != null) {
                     Resource<?> resource = parentSkin.getResource(filename);
                     if (resource != null) {
-                        queryParameters.putAll(getResourceURLParameters(resource));
-                        String resourceUrl = resource.getURL(forceSkinAction, queryParameters);
+                        queryParametersMap.putAll(getResourceURLParameters(resource));
+                        String resourceUrl = resource.getURL(forceSkinAction, queryParametersMap);
                         return resourceUrl;
                     }
                 }
@@ -2440,9 +2447,9 @@ public class XWiki implements EventListener
             // Look for a resource file
             String resourceFilePath = "/resources/" + filename;
             if (resourceExists(resourceFilePath)) {
-                queryParameters.putAll(getResourceURLParameters(resourceFilePath));
+                queryParametersMap.putAll(getResourceURLParameters(resourceFilePath));
                 URL url = urlf.createResourceURL(filename, forceSkinAction, context,
-                    queryParameters);
+                    queryParametersMap);
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -2409,6 +2409,16 @@ public class XWiki implements EventListener
         return getSkinFile(filename, false, context);
     }
 
+    /**
+     * Build and return a skin file url based on the given parameters.
+     * @param filename the file name of the skin file wanted
+     * @param forceSkinAction if true force the usage of directory /skins/ in the URL
+     * @param context current context for the request
+     * @param queryParameters parameters to add to the URL
+     * @return a resource URL for the asked filename
+     * @since 11.0RC1
+     */
+    @Unstable
     public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context,
         Map<String, String> queryParameters)
     {
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -2396,10 +2396,20 @@ public class XWiki implements EventListener
 
     public String getSkinFile(String filename, XWikiContext context)
     {
-        return getSkinFile(filename, false, context);
+        return getSkinFile(filename, context, null);
+    }
+
+    public String getSkinFile(String filename, XWikiContext context, String version)
+    {
+        return getSkinFile(filename, false, version, context);
     }
 
     public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context)
+    {
+        return getSkinFile(filename, forceSkinAction, null, context);
+    }
+
+    public String getSkinFile(String filename, boolean forceSkinAction, String version, XWikiContext context)
     {
         XWikiURLFactory urlf = context.getURLFactory();
 
@@ -2424,7 +2434,7 @@ public class XWiki implements EventListener
 
             // Look for a resource file
             if (resourceExists("/resources/" + filename)) {
-                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
+                URL url = urlf.createResourceURL(filename, forceSkinAction, version, context);
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
@@ -2445,10 +2455,21 @@ public class XWiki implements EventListener
 
     public String getSkinFile(String filename, String skin, XWikiContext context)
     {
-        return getSkinFile(filename, skin, false, context);
+        return getSkinFile(filename, skin, context, null);
+    }
+
+    public String getSkinFile(String filename, String skin, XWikiContext context, String version)
+    {
+        return getSkinFile(filename, skin, false, context, version);
     }
 
     public String getSkinFile(String filename, String skinId, boolean forceSkinAction, XWikiContext context)
+    {
+        return getSkinFile(filename, skinId, forceSkinAction, context, null);
+    }
+
+    public String getSkinFile(String filename, String skinId, boolean forceSkinAction, XWikiContext context,
+        String version)
     {
         try {
             Skin skin = getInternalSkinManager().getSkin(skinId);
@@ -2463,7 +2484,7 @@ public class XWiki implements EventListener
             if (resourceExists("/resources/" + filename)) {
                 XWikiURLFactory urlf = context.getURLFactory();
 
-                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
+                URL url = urlf.createResourceURL(filename, forceSkinAction, version, context);
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -47,7 +47,7 @@ import java.util.Date;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -85,9 +85,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
-import org.apache.http.NameValuePair;
-import org.apache.http.client.utils.URIBuilder;
-import org.apache.http.message.BasicNameValuePair;
 import org.apache.velocity.VelocityContext;
 import org.hibernate.HibernateException;
 import org.slf4j.Logger;
@@ -2407,7 +2404,8 @@ public class XWiki implements EventListener
         return getSkinFile(filename, false, context);
     }
 
-    public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context)
+    public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context,
+        Map<String, String> queryParameters)
     {
         XWikiURLFactory urlf = context.getURLFactory();
 
@@ -2417,7 +2415,9 @@ public class XWiki implements EventListener
             if (skin != null) {
                 Resource<?> resource = skin.getResource(filename);
                 if (resource != null) {
-                    return resource.getURL(forceSkinAction);
+                    queryParameters.putAll(getResourceURLParameters(resource));
+                    String resourceUrl = resource.getURL(forceSkinAction, queryParameters);
+                    return resourceUrl;
                 }
             } else {
                 // Try in the current parent skin
@@ -2425,7 +2425,9 @@ public class XWiki implements EventListener
                 if (parentSkin != null) {
                     Resource<?> resource = parentSkin.getResource(filename);
                     if (resource != null) {
-                        return resource.getURL(forceSkinAction);
+                        queryParameters.putAll(getResourceURLParameters(resource));
+                        String resourceUrl = resource.getURL(forceSkinAction, queryParameters);
+                        return resourceUrl;
                     }
                 }
             }
@@ -2433,9 +2435,9 @@ public class XWiki implements EventListener
             // Look for a resource file
             String resourceFilePath = "/resources/" + filename;
             if (resourceExists(resourceFilePath)) {
-                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
-                url = new URIBuilder(url.toURI()).addParameters(getResourceURLParameters(resourceFilePath)).build()
-                    .toURL();
+                queryParameters.putAll(getResourceURLParameters(resourceFilePath));
+                URL url = urlf.createResourceURL(filename, forceSkinAction, context,
+                    queryParameters);
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
@@ -2446,32 +2448,54 @@ public class XWiki implements EventListener
 
         // If all else fails, use the default base skin, even if the URLs could be invalid.
         URL url;
+
         if (forceSkinAction) {
             url = urlf.createSkinURL(filename, "skins", getDefaultBaseSkin(context), context);
         } else {
             url = urlf.createSkinURL(filename, getDefaultBaseSkin(context), context);
         }
-        return urlf.getURL(url, context);
+        queryParameters.putAll(getResourceURLParameters(url));
+        return urlf.getURL(url, context, queryParameters);
     }
 
-    private List<NameValuePair> getResourceURLParameters(String resourceFilePath)
+    public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context)
     {
-        List<NameValuePair> parameters = new LinkedList<>();
-        parameters.add(new BasicNameValuePair("xwikiVersion", this.getVersion()));
+        return getSkinFile(filename, forceSkinAction, context, new LinkedHashMap<>());
+    }
+
+    private Map<String, String> getResourceURLParameters(Resource<?> resource)
+    {
+        try {
+            URL resourceUrl = getResource(resource.getPath());
+            return getResourceURLParameters(resourceUrl);
+        } catch (MalformedURLException e) {
+            LOGGER.debug("Error while getting URL for resource filepath [{}]", resource.getPath(), e);
+            return Collections.singletonMap("xwikiVersion", getVersion());
+        }
+    }
+
+    private Map<String, String> getResourceURLParameters(String resourceFilePath)
+    {
+        try {
+            URL resourceUrl = getResource(resourceFilePath);
+            return getResourceURLParameters(resourceUrl);
+        } catch (MalformedURLException e) {
+            LOGGER.debug("Error while getting URL for resource filepath [{}]", resourceFilePath, e);
+            return Collections.singletonMap("xwikiVersion", getVersion());
+        }
+    }
+
+    private Map<String, String> getResourceURLParameters(URL resourceUrl)
+    {
+        Map<String, String> parameters = new LinkedHashMap<>();
 
         try {
-            URL resource = getResource(resourceFilePath);
-            Path resourcePath = Paths.get(resource.toURI());
-            long size = Files.size(resourcePath);
-
-            if (size != 0) {
-                parameters.add(new BasicNameValuePair("size", String.valueOf(size)));
-            }
-
+            Path resourcePath = Paths.get(resourceUrl.toURI());
             FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
-            parameters.add(new BasicNameValuePair("filedate", String.valueOf(lastModifiedTime.toMillis())));
+            parameters.put("filedate", String.valueOf(lastModifiedTime.toMillis()));
         } catch (Exception e) {
-            LOGGER.debug("Error when trying to access properties of resource file [{}]", resourceFilePath, e);
+            LOGGER.debug("Error when trying to access properties of resource URL [{}]", resourceUrl, e);
+            parameters.put("xwikiVersion", getVersion());
         }
         return parameters;
     }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -2404,20 +2404,10 @@ public class XWiki implements EventListener
 
     public String getSkinFile(String filename, XWikiContext context)
     {
-        return getSkinFile(filename, context, null);
-    }
-
-    public String getSkinFile(String filename, XWikiContext context, String version)
-    {
-        return getSkinFile(filename, false, version, context);
+        return getSkinFile(filename, false, context);
     }
 
     public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context)
-    {
-        return getSkinFile(filename, forceSkinAction, null, context);
-    }
-
-    public String getSkinFile(String filename, boolean forceSkinAction, String version, XWikiContext context)
     {
         XWikiURLFactory urlf = context.getURLFactory();
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -280,6 +280,11 @@ public class XWiki implements EventListener
     private static final String NO_VALUE = "---";
 
     /**
+     * Query parameters used to determine if a resource is in the browser cache.
+     */
+    private static final String CACHE_VERSION = "cache-version";
+
+    /**
      * List of top level space names that can be used in the fake context document created when accessing a resource
      * with the 'skin' action.
      */
@@ -2470,7 +2475,7 @@ public class XWiki implements EventListener
             return getResourceURLParameters(resourceUrl);
         } catch (MalformedURLException e) {
             LOGGER.debug("Error while getting URL for resource filepath [{}]", resource.getPath(), e);
-            return Collections.singletonMap("xwikiVersion", getVersion());
+            return Collections.singletonMap(CACHE_VERSION, getVersion());
         }
     }
 
@@ -2481,7 +2486,7 @@ public class XWiki implements EventListener
             return getResourceURLParameters(resourceUrl);
         } catch (MalformedURLException e) {
             LOGGER.debug("Error while getting URL for resource filepath [{}]", resourceFilePath, e);
-            return Collections.singletonMap("xwikiVersion", getVersion());
+            return Collections.singletonMap(CACHE_VERSION, getVersion());
         }
     }
 
@@ -2492,10 +2497,10 @@ public class XWiki implements EventListener
         try {
             Path resourcePath = Paths.get(resourceUrl.toURI());
             FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
-            parameters.put("filedate", String.valueOf(lastModifiedTime.toMillis()));
+            parameters.put(CACHE_VERSION, String.valueOf(lastModifiedTime.toMillis()));
         } catch (Exception e) {
             LOGGER.debug("Error when trying to access properties of resource URL [{}]", resourceUrl, e);
-            parameters.put("xwikiVersion", getVersion());
+            parameters.put(CACHE_VERSION, getVersion());
         }
         return parameters;
     }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -85,6 +85,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.http.client.utils.URIBuilder;
 import org.apache.velocity.VelocityContext;
 import org.hibernate.HibernateException;
 import org.slf4j.Logger;
@@ -2437,9 +2438,7 @@ public class XWiki implements EventListener
             if (skin != null) {
                 Resource<?> resource = skin.getResource(filename);
                 if (resource != null) {
-                    queryParametersMap.putAll(getResourceURLParameters(resource));
-                    String resourceUrl = resource.getURL(forceSkinAction, queryParametersMap);
-                    return resourceUrl;
+                    return createURLForResource(resource, forceSkinAction, queryParametersMap, context, urlf);
                 }
             } else {
                 // Try in the current parent skin
@@ -2447,9 +2446,7 @@ public class XWiki implements EventListener
                 if (parentSkin != null) {
                     Resource<?> resource = parentSkin.getResource(filename);
                     if (resource != null) {
-                        queryParametersMap.putAll(getResourceURLParameters(resource));
-                        String resourceUrl = resource.getURL(forceSkinAction, queryParametersMap);
-                        return resourceUrl;
+                        return createURLForResource(resource, forceSkinAction, queryParametersMap, context, urlf);
                     }
                 }
             }
@@ -2457,9 +2454,9 @@ public class XWiki implements EventListener
             // Look for a resource file
             String resourceFilePath = "/resources/" + filename;
             if (resourceExists(resourceFilePath)) {
-                queryParametersMap.putAll(getResourceURLParameters(resourceFilePath));
-                URL url = urlf.createResourceURL(filename, forceSkinAction, context,
-                    queryParametersMap);
+                queryParametersMap.putAll(getResourceURLCacheParameters(resourceFilePath));
+                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
+                url = urlf.addQueryParameters(url, queryParameters);
                 return urlf.getURL(url, context);
             }
         } catch (Exception e) {
@@ -2476,8 +2473,19 @@ public class XWiki implements EventListener
         } else {
             url = urlf.createSkinURL(filename, getDefaultBaseSkin(context), context);
         }
-        queryParameters.putAll(getResourceURLParameters(url));
-        return urlf.getURL(url, context, queryParameters);
+        queryParameters.putAll(getResourceURLCacheParameters(url));
+        url = urlf.addQueryParameters(url, queryParameters);
+        return urlf.getURL(url, context);
+    }
+
+    private String createURLForResource(Resource<?> resource, boolean forceSkinAction,
+        Map<String, String> queryParametersMap, XWikiContext context, XWikiURLFactory urlf) throws Exception
+    {
+        queryParametersMap.putAll(getResourceURLCacheParameters(resource));
+        String resourceUrlString = resource.getURL(forceSkinAction);
+        URL resourceUrl = new URIBuilder(resourceUrlString).build().toURL();
+        resourceUrl = urlf.addQueryParameters(resourceUrl, queryParametersMap);
+        return urlf.getURL(resourceUrl, context);
     }
 
     public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context)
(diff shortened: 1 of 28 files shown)


surli:
This PR contains a first proposal for fixing the cache problem. It adds automatically the current xwiki version when calling `xwiki.getSkinFile()` to the produced URL. 
I only tested by starting locally a wiki: everything seems to load fine.

tmortagne:
Have to tried to get the file date ? Might be more accurate, would also cover FS skinw which don't have anything to do with XWiki Standard and local customization.

surli:
> Have to tried to get the file date ? 

No I did not tried. I can check.

vmassol:
Getting the file date requires some additional I/O operations so it less good for perf IMO. Also makes it more fragile since it's less stable (and can vary). It would indeed allow covering other use cases such as FS skin. But at the cost of more I/O,. Not sure what's the best.

surli:
About that I just talked with @lucaa and she brings the idea about having a specific XWiki object with the value of the file version which could be updated/regenerated by an admin.
This covers several use cases and have interesting properties:
  1. we can create this object at first XWiki launch, with a generated hash from the XWiki version, so we can guarantee that the cache will be erased for that version
  2. whenever an admin modifies a config or a file he'll get the ability to generate a new hash version, which will again invalid all caches and allow for all users to enable the changes

So I propose to go directly in that direction, to avoid changing the API multiple times about this version. WDYT?

tmortagne:
So it's all or nothing, right ? If you want the browser to reload one resource (the one containing the file upload size for example) you have to make it reload everything ? I'm not such a huge fan of this logic.

About the file upload size, IMO this is not such a great example since it's just very bad design. As suggested in the past already this value have nothing to do in a js file and should be gathered dynamically when needed instead.

surli:
> So it's all or nothing, right ? If you want the browser to reload one resource (the one containing the file upload size for example) you have to make it reload everything ? I'm not such a huge fan of this logic.

yes it's all or nothing, but it leaves the ability for admin to control users' cache. IMO it could help solving lot of problems when debugging a client, and reloading everything is not that costly compared to ask everyone to clear the browser cache.

tmortagne:
> compared to ask everyone to clear the browser cache

It's not really the point...

surli:
> > compared to ask everyone to clear the browser cache

> It's not really the point...

It's a facet of the problem.
The main idea is that the FS files are cached by the browser and we currently don't have control on it. In this PR I already allow to load them based on an URL built using the XWiki version, so it solves a part of the problem: the user won't be stuck anymore when changing from one version to another.

Now another part of the problem is that if a change is made without an impact on the version, the file won't be reloaded: for that one I propose to go further and to allow admins to somehow control the "version" part of the generated URLs.

tmortagne:
> About the file upload size, IMO this is not such a great example since it's just very bad design

Actually this was fixed already in https://jira.xwiki.org/browse/XWIKI-14476.

> It's a facet of the problem.

What I mean is that my point was not to compare what you propose to doing nothing.

tmortagne:
> for that one I propose to go further and to allow admins to somehow control the "version" part of the generated URLs

Well what you actually proposed is to abandon any idea of automation (except for upgrade use case) and for the admin user to explicitly ask to update the hash every time a modification is made to force the browser to empty its cache.

Giving the opportunity to admins to have a way to force browser cache cleanup for all resources sure is useful but nothing prevent from combining several sources of "version" (a global counter explicitly incremented by an admin, the war version for FS, the file date for FS, maybe other ideas, the page version for JSX, etc.) in the URL to make it unique.

atallahade:
I don't think it's a bad idea to let the Admins have a way to force a global cache reload. They would use that in case others methods that we are using to reload the cache are not working in some rare cases.

tmortagne:
> I don't think it's a bad idea to let the Admins have a way to force a global cache reload. They would use that in case others methods that we are using to reload the cache are not working in some rare cases.

I don't think it's a bad idea either, what I don't like is making is the only way.

surli:
> Actually this was fixed already in jira.xwiki.org/browse/XWIKI-14476.

OK this was a bad example

> nothing prevent from combining several sources of "version" (a global counter explicitly incremented by an admin, the war version for FS, the file date for FS, maybe other ideas, the page version for JSX, etc.) in the URL to make it unique.

I'm afraid about the perf of generated such unique URL, especially if we need to get access to the files: basically we completely make the browser cache useless. 
Here the proposed solution is overkill, but it's simple: it's used when needed and won't impact much the perf. 

> you actually proposed is to abandon any idea of automation
> what I don't like is making is the only way.

there is a misunderstanding here: I just want to provide control to the user if needed, it can be combined with other solutions.

tmortagne:
> I'm afraid about the perf of generated such unique URL, especially if we need to get access to the files: basically we completely make the browser cache useless.

I don't think you understood what I indicated. I never proposed to generate a unique URL every time you call getSkinFile but to combine several source of versioning for the URL to be unique from those versions point of view.

tmortagne:
> API multiple times about this version.

I don't think we need to change any API, at least not the API you change in that PR. The currently existing getSkinFile should add a version information. Otherwise it means you don't really fix anything and all the code using getSkinFile has to be modified to pass a version (plus this code should not the one dealing with resource versionning).

tmortagne:
> it looks really expensive to retrieve all the information you mentioned for all resources at each load.

I listed ideas of versioning sources but you obviously don't get all of them for all URLs, it depends on the type of resource and we already get some of them for wiki resources like skin extensions. Re the cost of getting the file date have you tested that cost or if it's even possible to do it in a generic way ? That's all I asked so far. Note that we already access the file in practice to check that it exist (and if not fallback on parent skin, etc.).

IMO the important things is that when we generate the URL of a resource we have a system which can combine different versioning sources which was really not the direction that your proposal seemed to take (since you were combining different sources in the global hash itself). For example instead of maintaining a global hash generated based on the version of the was we could just simply store a basic counter the admin can increment and combine it with the war version (getting that version cost nothing) for FS skin files (and ideally other information when possible and fast enough). The admin switch feature also need to be added to other resources like skin extensions which already have a different source of versioning (would not really make sense to have this feature only for FS resources).

surli:
> Re the cost of getting the file date have you tested that cost or if it's even possible to do it in a generic way ?

No I did not tested it yet. I will check. 

> instead of maintaining a global hash generated based on the version of the war we could simply store a basic counter the admin can increment and combine it with the war version (getting that version cost nothing) for FS skin files

I agree on that. 

Another question that just occurred is where to put a common API for both old-core `getSkinFile` and `skinx#use`?
I don't want to write twice the mechanism which generate the URL properties: should I create a new extension used both in old core and skinx?

tmortagne:
> Another question that just occurred is where to put a common API for both old-core getSkinFile and skinx#use?

Sounds like something for xwiki-platform-url-api ? On a side note we are supposed to move everything related to URL generation in xwiki-platform-url/* at some point but we did not had time yet.

surli:
@tmortagne so I improve the API for getSkinFile to use three different parameters:
  * XWiki version
  * date file
  * file size

I will see to use the same parameters in JSX resources.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2423,8 +2431,11 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
             }
```
Enygma2002: Aren't you missing the above 2 cases where a file can come from a skin document (in which case, you want the document version appended to the URL) or from a parent skin (i.e. baseskin, a filesystem folder -- in which case you want to append the same file size and/or last modification time parameters to the URL)?

Enygma2002:
I don't really follow how adding the XWiki war file version (i.e. contents of `/WEB-INF/version.properties`) is necessarily a good thing? Coupled with the file size and last modified date, it's actually redundant (i.e. already covered by changes in size and modification date) and even harmful, since not files are guaranteed to be changed on an XWiki version upgrade (very few ones do change, in fact) and a full cache invalidation of all your clients will be forced for no good reason.

The only place where the version approach would try to make sense would be if you want a way to manually invalidate everything, which would be fine. However, due to the negative side-effect of the forced invalidation caused by an upgrade, I would advise against it.

Note, that any manual force mechanism would also imply a performance tax, because each resource would have to check the value of the force mechanism for any change (which might imply reading a file or reading an object from a page). OK, reading the value from a page would mitigate the negative impact, since the page would become one of the permanent residents in the document cache. I am personally not convinced we need such a mechanism and would prefer that we invest more time in identifying and automating the cache invalidations as much as possible.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractDocumentSkinExtensionPlugin.java
```diff
@@ -454,9 +467,10 @@ private String getLanguageQueryString(XWikiContext context)
     protected String getDocumentSkinExtensionURL(DocumentReference documentReference, String documentName,
             String pluginName, XWikiContext context)
     {
-        String queryString = String.format("%s&amp;%s%s",
+        String queryString = String.format("%s&amp;%s&amp;%s%s",
                 getLanguageQueryString(context),
                 getDocumentVersionQueryString(documentReference, context),
+                getWikiVersionQueryString(context),
```
tmortagne: I don't see the point for a wiki page, if the wiki page did not changed then why could we need to invalidate the cache ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2423,8 +2431,11 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
...
 
             // Look for a resource file
-            if (resourceExists("/resources/" + filename)) {
+            String resourceFilePath = "/resources/" + filename;
+            if (resourceExists(resourceFilePath)) {
                 URL url = urlf.createResourceURL(filename, forceSkinAction, context);
+                url = new URIBuilder(url.toURI()).addParameters(getResourceURLParameters(resourceFilePath)).build()
```
tmortagne: Would probably be better for performances and more consistent to add a createResourceURL taking a `String querystring` like it's the case for others XWikiURLFactory methods.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
tmortagne: Yes only /resources/ use case seems to be supported right now but version support would make more sense in Resource#getURL implementations than here IMO.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractDocumentSkinExtensionPlugin.java
surli: Here it's really for a dev usecase: specifically if I launch on my own machine XWiki 10.8.2 and then I launch XWiki 10.11, the doc version won't change so it will try to get it from the cache.
I know this shouldn't happen in prod since most of the case it would be the result of a migration and the doc version should change.
We can still imagine the usecase of an user willing to upgrade its version by installing a fresh XWiki, so here it might be useful for this specific case.

surli:
> I don't really follow how adding the XWiki war file version (i.e. contents of /WEB-INF/version.properties) is necessarily a good thing?

I'm still experimenting here, so thanks for your feedback :)

> Coupled with the file size and last modified date, it's actually redundant (i.e. already covered by changes in size and modification date) and even harmful

Agreed. I will change that to not put the XWiki version when I have the file size / last modified date. Now do you agree that XWiki version can be useful if I cannot access the file, so that I can use it as a fallback? 

> we invest more time in identifying and automating the cache invalidations as much as possible.

That's actually what @tmortagne proposed too, and that's the direction I'm currently investigating right now. So if you have any other proposal to improve such mechanism, I take it :)

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractDocumentSkinExtensionPlugin.java
tmortagne: Hmm maybe a better criteria for this use case is the instance id then.

tmortagne:
> so that I can use it as a fallback?

sounds good to me

surli:
> . I will change that to not put the XWiki version when I have the file size / last modified date.

Actually do we really need both file size + last modified date: I think only the file size is needed, since last modified date will change after an upgrade, even if the file didn't change. So I'd say the file size is enough, WDYT?

tmortagne:
> So I'd say the file size is enough, WDYT?

You can have changes which don't have any impact on the size so date is safer IMO.

Enygma2002:
> Now do you agree that XWiki version can be useful if I cannot access the file, so that I can use it as a fallback?

Sure, if actually checking the files is not an option, we're not left with much choice. Of course, this is only applicable for /resources/ or base/parent skin files, coming straight from the FS (i.e. from the WAR to which the XWiki version corresponds). Resources coming from the current skin (document) should always use the document's version.

Enygma2002:
A new suggestion I'd have is to avoid adding too many parameters to the resource's URL. IMO, we should use just one parameter, say "cache-buster=67dsad5a6s75ds675dsa67", that is built by joining and/or hashing multiple pieces of information (like file size+last data + instance ID + etc.).

tmortagne:
I don't know if having several parameters is really an issue but would certainly be a good idea to make more explicit in the parameter name that it's a cache related thing and it's actually useless on server side (like a `cache-` as suggested by @Enygma2002).

surli:
> would certainly be a good idea to make more explicit in the parameter name that it's a cache related thing

ok I'll change the query parameters name then. For now I only add one parameter so I don't see the need to computing a hash for it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
surli: Now all cases are supported.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2409,22 +2420,29 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
             if (skin != null) {
                 Resource<?> resource = skin.getResource(filename);
                 if (resource != null) {
-                    return resource.getURL(forceSkinAction);
+                    queryParameters.putAll(getResourceURLParameters(resource));
```
mflorea: You're modifying an input parameter which means this method has side effects that are probably not expected from a "getter". Moreover, the ``queryParameters`` map cold be unmodifiable.

surli: > You're modifying an input parameter which means this method has side effects that are probably not expected from a "getter"

Agreed but here the method is in fact about creating a skinFile URL, so from the beginning it's not a "real" getter IMO. Might be even renamed "createSkinFileURL". 

> Moreover, the queryParameters map cold be unmodifiable.

Right I should create a new map.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2435,17 +2453,61 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
...
+    }
+
+    private Map<String, String> getResourceURLParameters(URL resourceUrl)
+    {
+        Map<String, String> parameters = new LinkedHashMap<>();
+
+        try {
+            Path resourcePath = Paths.get(resourceUrl.toURI());
+            FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
+            parameters.put(CACHE_VERSION, String.valueOf(lastModifiedTime.toMillis()));
```
mflorea: For resources that have an associated last modification date I think the best practice is to use https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since .

surli: I don't think it applies on the current workflow. If I understand well `If-Modified-Since` it's a request header so it's when the browser needs to load a resource that it should fill this header. Now here we only send an URL to the browser to make the request. I don't see how/where we could indicate such a date to put in the right header.

mflorea: My suggestion was to fix the code that serves the resource (e.g. Skin action) instead of the code that builds the resource URL. The code that serves the resource can set and read HTTP headers.

surli: OK but then it's the `Last-Modified` header which should be changed: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified as a response header.

mflorea: Yes, the WebJars resource handler does something like this:
* sets the Last-Modified header https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java#L216
* and then checks the If-Modified-Since header https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java#L114

Maybe something like this can be applied to skin resources.

surli: interesting, I'll check that

surli: So I indeed check what you did, but as far as I can see it's difficult to apply right now since we don't use a ResourceReferenceHandler for the bin (see the comment in EntityResourceReferenceHandler https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-default/src/main/java/org/xwiki/resource/internal/entity/EntityResourceReferenceHandler.java#L55-L58) 
I might be wrong here but it seems to me that we should first fix that in order to properly apply the headers.
Since this looks like a big refactoring I propose that we apply first the easy solution already implemented with the query parameters.

surli:
so it's ready for review @tmortagne @mflorea @Enygma2002

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
mflorea: I wasn't suggesting you to use exactly the code from ``AbstractServletResourceReferenceHandler``. I linked it just as an example. My point was actually to use the **cache-related HTTP headers** instead of the query string. I find it wrong to put the date in the query string if there are dedicated HTTP headers to handle this.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2435,17 +2470,61 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
...
```
vmassol: Personally I would have preferred using the XWiki version for performance reasons. Note that if you change some files, you could have also changed the XWiki version for example with a prefix "10.8.3-patch1". It's logical since it means "patching" xwiki. Right now I think you need to restart XWiki for the version change to take effect but either that could be acceptable or we could have imagine a way to set/override the XWiki version in the XWiki Admin UI.

I don't know how much perfs are going to be impacted but this means a ton of IO more for large wikis. And this is is unnecessary on the very large majority of cases (you don't patch xwiki every second!).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2435,17 +2470,61 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
...
+        return getSkinFile(filename, forceSkinAction, context, new LinkedHashMap<>());
+    }
+
+    private Map<String, String> getResourceURLParameters(Resource<?> resource)
+    {
+        try {
+            URL resourceUrl = getResource(resource.getPath());
+            return getResourceURLParameters(resourceUrl);
+        } catch (MalformedURLException e) {
+            LOGGER.debug("Error while getting URL for resource filepath [{}]", resource.getPath(), e);
```
vmassol: I would use an error since apparently this is not expected to happen according to the debug message.

vmassol: s/Error while getting URL for resource filepath [{}]/Error while getting URL for resource path [{}]/ ?

vmassol: I'm wondering if it's good to fallback on the version if we have an error. Or if we do then we should log a warning then maybe.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractWikiResource.java
```diff
@@ -91,4 +94,21 @@ public String getURL(boolean forceSkinAction) throws Exception
...
+    /**
+     * Create URL for the given document with some query parameters.
+     * @param document the document wanted for the URL
+     * @param queryParameters some parameters to add in the URL
+     * @return a URL for the given document with some parameters
+     * @throws Exception if the URL cannot be built.
+     * @since 11.0RC1
+     */
+    @Unstable
+    public abstract String getURL(XWikiDocument document, Map<String, String> queryParameters) throws Exception;
```
vmassol: Looks weird to pass query parameters to get them back after the call. @tmortagne could you review this API to verify the design is good?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -490,6 +499,20 @@ public URL createResourceURL(String filename, boolean forceSkinAction, XWikiCont
...
         path.append("resources");
         addFileName(path, filename, false, context);
+        if (queryParameters != null && !queryParameters.isEmpty()) {
+            path.append("?");
+            Iterator<Map.Entry<String, String>> entryIterator = queryParameters.entrySet().iterator();
+            while (entryIterator.hasNext()) {
+                Map.Entry<String, String> entry = entryIterator.next();
+                path.append(entry.getKey());
+                path.append("=");
+                path.append(entry.getValue());
```
vmassol: I hope that the passed queryParameters are already encoded :)

tmortagne: I should definitely be encoded here and not in the Map.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractWikiResource.java
tmortagne: Not sure what you mean by "get them back after the call". What you get back is a URL.

From what I understand the point of this API is to provide custom parameters to add to the generated URL. You have the same thing in XWikiURLFactory.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractWikiResource.java
```diff
@@ -91,4 +94,21 @@ public String getURL(boolean forceSkinAction) throws Exception
     }
 
     public abstract String getURL(XWikiDocument document) throws Exception;
+
+    @Override
+    public String getURL(boolean forceSkinAction, Map<String, String> queryParameters) throws Exception
+    {
+        return getURL(getDocument());
```
tmortagne: Shouldn't this be `getURL(getDocument(), queryParameters);` ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AttachmentWikiResource.java
```diff
@@ -56,4 +59,20 @@ public String getURL(XWikiDocument document) throws Exception
...
+    @Override
+    public String getURL(XWikiDocument document, Map<String, String> queryParameters) throws Exception
+    {
+        StringBuilder queryString = new StringBuilder();
+
+        for (Map.Entry<String, String> parameters : queryParameters.entrySet()) {
+            queryString.append(parameters.getKey())
+                .append("=")
+                .append(parameters.getValue())
+                .append("&");
```
tmortagne: Adding the `&` all the time and then removing it in the final String is really not nice.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AttachmentWikiResource.java
```diff
@@ -56,4 +59,20 @@ public String getURL(XWikiDocument document) throws Exception
     {
         return document.getAttachmentURL(this.reference.getName(), "skin", this.xcontextProvider.get());
     }
+
+    @Override
+    public String getURL(XWikiDocument document, Map<String, String> queryParameters) throws Exception
+    {
+        StringBuilder queryString = new StringBuilder();
+
+        for (Map.Entry<String, String> parameters : queryParameters.entrySet()) {
```
tmortagne: I'm sure this query string generation can go (or even already exist) in some common utility class.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -652,6 +675,12 @@ public URL createAttachmentRevisionURL(String filename, String spaces, String na
      */
     @Override
     public String getURL(URL url, XWikiContext context)
+    {
+        return getURL(url, context, null);
+    }
+
+    @Override
+    public String getURL(URL url, XWikiContext context, Map<String, String> queryParameters)
```
tmortagne: The point of XWikiServletURLFactory#getURL is not to generate an URL but to move from absolute to relative URL (if applicable). It's really not it's job to add a query string the way it's currently defined.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-api/src/main/java/org/xwiki/skin/Resource.java
```diff
@@ -65,4 +68,18 @@
...
+     * Create a URL for the resource.
+     *
+     * @param forceSkinAction true if a dynamic skin URL should be forced
+     * @param queryParameters add the following parameters to the URL
+     * @return a URL to the resource
+     * @throws Exception when failing to create a URL for the resource
+     * @since 11.0RC1
+     */
+    @Unstable
+    default String getURL(boolean forceSkinAction, Map<String, String> queryParameters) throws Exception {
```
tmortagne: It's not very clear to me why this `queryParameters` is needed. You are not really supposed to know which parameter to add when you call getURL. The whole idea is that only the Resource implement knows what makes this specific resource unique (document version for an object, attachment version for an attachment, date or whatever for a file on disk, etc.) and will automatically add it to the generated URL.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiURLFactory.java
```diff
@@ -169,6 +185,8 @@ URL createAttachmentRevisionURL(String filename, String spaces, String name, Str
      */
     String getURL(URL url, XWikiContext context);
 
+    String getURL(URL url, XWikiContext context, Map<String, String> queryParameters);
```
tmortagne: The point of XWikiServletURLFactory#getURL is not to generate an URL but to move from absolute to relative URL (if applicable). It's really not its job to add a query string the way it's currently defined.

surli:
@tmortagne @vmassol this one is ready again for review. I simplified the changes a lot by not modifying the API of `Resource` and by adding a new utility method in XWikiURLFactory to add query parameters to a given URL.

The only thing I did not change for now is related to the comment of @vmassol with using the date of the file as parameter or only the version. IMO it would be safer for perf reason to only rely on XWiki version for now, even if we know it wouldn't be enough for some cases. WDYT?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2435,17 +2467,72 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
...
             url = urlf.createSkinURL(filename, "skins", getDefaultBaseSkin(context), context);
         } else {
             url = urlf.createSkinURL(filename, getDefaultBaseSkin(context), context);
         }
+        queryParametersMap.putAll(getResourceURLCacheParameters(url));
+        url = urlf.addQueryParameters(url, queryParametersMap);
         return urlf.getURL(url, context);
     }
 
+    private String createURLForResource(Resource<?> resource, boolean forceSkinAction,
```
tmortagne: As I keep saying, adding the parameters to the URL is Resource implementation's job because that's the only place where you really know what to add depending on the resource type. For example here you try to access the File as if all Resource were files on drive which is not true.

surli: I tried to answer to that comment by adding a query parameter when building the URL in EnvironmentSkinResource and by only returning the resource URL in getSkinFile when we already have a resource.
However when we only retrieve a path in getSkinFile I continue to try to retrieve the path to fill the query parameter, I did not change that.

I'm not sure it's what you had in mind but I've the feeling it would require far more refactoring to make it another way. I let you check.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
```diff
@@ -49,6 +55,13 @@ public SkinEnvironmentResource(String path, String resourceName, ResourceReposit
     public String getURL(boolean forceSkinAction) throws Exception
     {
         XWikiContext xcontext = this.xcontextProvider.get();
+
+        URL resourceUrl = this.xcontextProvider.get().getEngineContext().getResource(this.getPath());
+        Map<String, String> parameters = new LinkedHashMap<>();
+        Path resourcePath = Paths.get(resourceUrl.toURI());
```
tmortagne: You should probably bulletproof a bit this line in case this resource cannot be converted to a local file for some reason (like not expended WAR or weird URL scheme like in WebSphere).

surli: ok I'll try/catch them.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
```diff
@@ -59,6 +72,7 @@ public String getURL(boolean forceSkinAction) throws Exception
             url = urlf.createSkinURL(this.resourceName, getRepository().getId(), xcontext);
         }
 
+        url = urlf.addQueryParameters(url, parameters);
```
tmortagne: I still think this have nothing to do with XWikiURLFactory. Sounds more like something for a generic URL related tool on xwiki-commons side (did not saw anything in Apache commons but maybe there is).

tmortagne:
Would be nice to add something in ObjectPropertyWikiResource too (with the object document owner version in this case).

There is also AttachmentWikiResource (with the attachment version).

surli:
> There is also AttachmentWikiResource (with the attachment version).

There's already an issue for this one: https://jira.xwiki.org/browse/XWIKI-15450

I planned to do it afterwards. Maybe we could create a dedicated issue for ObjectPropertyWikiResource then.

tmortagne:
> There's already an issue for this one: https://jira.xwiki.org/browse/XWIKI-15450

Sure no need to do anything at AttachmentWikiResource level if it's done at XWikiURLFactory#createAttachmentURL.

> I planned to do it afterwards. Maybe we could create a dedicated issue for ObjectPropertyWikiResource then.

XWIKI-6073 is not specifically about filesystem in my mind but as you want. As long as it's taken into account in the end :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
tmortagne: Should probably also fallback on the XWiki version if there is an issue.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
tmortagne: > I'm not sure it's what you had in mind

Looks good :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
tmortagne: Same comments for the duplicated code located in XWiki.java (which should probably me moved in some common place, ultimately what's in XWiki.java right now should probably be refactored into a something extending AbstractEnvironmentResource but that's for later).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
surli: Actually I'm already using an Apache commons (https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/utils/URIBuilder.html) but I still need to encapsulate it in a method since I need to call the builder, encode the parameters etc. 

I checked in xwiki-commons I don't really where to put that in the existing modules. Were you thinking about creating a new module specifically for that? Seems a bit overkill.

tmortagne: My main point is that it does not make any sense as a XWikiURLFactory method and should go in some URL related tool instead.

Ideally it should probably go in some xwiki-commons-lang and rename xwiki-commons-text). In the meantime I guess the easiest and safest in term of technical debt is to just add some URLUtils in an internal package of oldcore or xwiki-platform-url-api for example.

surli: So, I agree with you that the utility method shouldn't be added in XWikiURLFactory. Now I really don't want to create a specific class URLUtils just to add a new method we only use twice. Moreover the purpose of XWikiURLFactory is to built dedicated URL for different stuff. 

Finally I propose to add some new methods in XWikiURLFactory by adding the queryParameters parameter to some existing method such as `createSkinURL` and `createResourceURL`, so when it creates those URLs it will be able to also put the needed query parameters. It avoid to create an URL and to modify it afterwards. WDYT?

tmortagne: > Finally I propose to add some new methods in XWikiURLFactory by adding the queryParameters parameter to some existing method such as `createSkinURL` and `createResourceURL`, so when it creates those URLs it will be able to also put the needed query parameters. It avoid to create an URL and to modify it afterwards. WDYT?

I'm fine with that, I remember suggesting it in a previous comment.

surli:
@tmortagne it's ready again for review. I tried to put some javadoc on the old and new methods in XWikiURLFactory but it's not always easy there :) 

I handled the `ObjectPropertyWikiResource` already since it was pretty straight forward. For the attachment, I have to work a bit more on it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2399,9 +2409,27 @@ public String getSkinFile(String filename, XWikiContext context)
...
+     * @param queryParameters parameters to add to the URL
+     * @return a resource URL for the asked filename
+     * @since 11.1RC1
+     */
+    @Unstable
+    public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext context,
+        Map<String, String> queryParameters)
     {
         XWikiURLFactory urlf = context.getURLFactory();
+        Map<String,String> queryParametersMap;
```
tmortagne: Does not seems to be used.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/ObjectPropertyWikiResource.java
```diff
@@ -59,10 +62,13 @@ public String getURL(XWikiDocument document) throws Exception
     {
         XWikiContext xcontext = this.xcontextProvider.get();
 
+        Map<String, String> parameters = new LinkedHashMap<>();
+        parameters.put(XWiki.CACHE_VERSION, document.getVersion());
```
tmortagne: The version seemed obvious to me but I'm now wondering if the document date (or maybe both version and date) would not be safer. Thinking about use cases where the document is replaced with its history (from a backup for example) in which case you can end up with the same version associated to totally different content.

surli: If I follow you, you're basically saying that using the docVersion in the JSX/SSX is not necessarily enough either?
I'd prefer to have a consistent behaviour, and to modify it everywhere if we feel the need.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
surli: Good catch! I indeed forgot to fill it, and to use it. I'll fix that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2423,8 +2444,11 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
             }
 
             // Look for a resource file
-            if (resourceExists("/resources/" + filename)) {
-                URL url = urlf.createResourceURL(filename, forceSkinAction, context);
+            String resourceFilePath = "/resources/" + filename;
+            if (resourceExists(resourceFilePath)) {
+                Map<String,String> queryParametersMap = getResourceURLCacheParameters(resourceFilePath);
+                queryParametersMap.putAll(queryParameters);
```
surli: So I updated the code to put the given queryParameters, in the only place I could use them and it actually feels wrong.
If you look at the stylesheets.vm I pushed in this same PR, we have the following line:
```
$urlParametersMap.put("colorTheme", $services.model.serialize($themeDoc.getDocumentReference(), "default")))
$xwiki.getSkinFile($stylesheet, true, ${urlParametersMap})
```

It means that basically those queryParameters will be discarded if we find the resource file in the current skin (previous lines of the same function) which doesn't feel right. I'd say that for this one I also need to add a method:

```
Resource#getURL(boolean, Map<String,String>)
```

to be able to ask a resource URL with some given parameters.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/ObjectPropertyWikiResource.java
tmortagne: >  I'd prefer to have a consistent behaviour, and to modify it everywhere if we feel the need.

I totally agree without.

Note that I'm not against the current version, I'm just wondering if we should also include (or replace with) the date or not.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiURLFactory.java
```diff
@@ -97,14 +100,99 @@ URL createURL(String spaces, String name, String action, String querystring, Str
...
+     * Create an URL for the given filename in the given skin directory.
+     * @param filename the file to reach.
+     * @param skin the skin where the file should be loaded
+     * @param context current context
+     * @param queryParameters the parameters to put at the end of the URL
+     * @return a URL to load the given file with the given query parameters.
+     * @since 11.1RC1
+     */
+    @Unstable
+    URL createSkinURL(String filename, String skin, XWikiContext context, Map<String, String> queryParameters);
```
tmortagne: I just realized that Map<String, String> is actually too restrictive because it make impossible to have a parameter with several values. You could change for Map<String, Object> for example and check in the implementation if it's a Iterable or a table for example.

surli: Do you have a usecase in mind? I was thinking that in such cases, the user should create his own string from the table/list values, before submitting the query parameter.

surli: Hmmm. Right. And since I'm using Apache commons to create the URL, I should be able to rely on it for processing the parameters. So it should be easy.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
surli: After discussion we decided to remove the queryParameters argument in getSkinFile.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
```diff
@@ -45,18 +56,41 @@ public SkinEnvironmentResource(String path, String resourceName, ResourceReposit
         this.xcontextProvider = xcontextProvider;
     }
 
+    private URLConfiguration getURLConfiguration() {
```
surli: I'm actually really not sure about this one in terms of perf: it means that for each resource, we will need to load an URLConfiguration component, is it ok? @tmortagne

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -444,23 +446,53 @@ public URL createExternalURL(String spaces, String name, String action, String q
...
+                    for (String paramValueElement : (String[]) paramValue) {
+                        uriBuilder.addParameter(key, paramValueElement);
+                    }
+                } else if (paramValue instanceof Collection) {
+                    for (Object paramValueElement : (Collection) paramValue) {
+                        uriBuilder.addParameter(key, paramValueElement.toString());
+                    }
+                } else {
+                    uriBuilder.addParameter(key, paramValue.toString());
+                }
```
surli: So finally URIBuilder does not provide a simple API for that: I only handle collection, since I'm not sure how we should handle it for a map.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
tmortagne: Why don't you inject it ?

surli: Wouldn't it be the same in terms of perf?

tmortagne: Sorry I read too fast and did not noticed where you where. You could do the same as what we do for the XWikiContext provider: pass the instance to the constructor.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -444,23 +446,53 @@ public URL createExternalURL(String spaces, String name, String action, String q
...
+    {
+        try {
+            URIBuilder uriBuilder = new URIBuilder(serverUrl.toURI()).setPath(path);
+            for (Map.Entry<String, Object> queryParameter : queryParameters.entrySet()) {
+                String key = queryParameter.getKey();
+                Object paramValue = queryParameter.getValue();
+
+                if (paramValue instanceof String) {
+                    uriBuilder.addParameter(key, (String) paramValue);
+                } else if (paramValue instanceof String[]) {
```
tmortagne: I would have supported any array but it's good enough with String[].

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/SkinEnvironmentResource.java
surli: > You could do the same as what we do for the XWikiContext provider: pass the instance to the constructor.

I'll check that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
surli: > supported any array

Indeed. I'll improve it.

surli:
so I guess it's ready @tmortagne

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -444,23 +447,97 @@ public URL createExternalURL(String spaces, String name, String action, String q
...
+                    stringBuilder.append("&");
+                }
+                index++;
+            }
+        } else {
+            appendQueryParameter(key, paramValue.toString(), stringBuilder);
+        }
+    }
+
+    private URL buildURL(URL serverUrl, String path, Map<String, Object> queryParameters, XWikiContext context)
```
surli: @tmortagne I had to entirely rewrite this one since we finally cannot rely on URIBuilder: it encodes directly the part of the URL, and that's something we also do in normalizeURL, so basically we encode it twice... I spot it with the failing test in XWikiServletFactoryURLTest that I updated.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/XWikiServletURLFactoryTest.java
```diff
@@ -544,6 +545,23 @@ public void createResourceURL()
...
+        queryParametersMap.put("aListWithStringToEncode", Arrays.asList("épervier", "androïde"));
+        queryParametersMap.put("aCustomObject", new Object() {
+            @Override
+            public String toString() {
+                return "foo";
+            }
+        });
+        url = this.urlFactory.createResourceURL("o;ne/t?w&o/t=hr#e e", false, this.oldcore.getXWikiContext(),
+            queryParametersMap);
+        assertEquals("http://127.0.0.1/xwiki/resources/o;ne/t%3Fw&o/t=hr%23e%20e?cache-version=11.1-SNAPSHOT%23"
```
surli: @tmortagne so I added this new assert to assess the behaviour of the URLBuilder with query parameters, please ensure that's indeed what we expect.

surli:
@tmortagne it's ready again. I had to add a new test for DefaultURLConfiguration to avoid coverage decrease. I wasn't really inspired to put more tests in the module...

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -444,23 +447,97 @@ public URL createExternalURL(String spaces, String name, String action, String q
         return this.createURL(spaces, name, action, querystring, anchor, xwikidb, context);
     }
 
+    private void appendQueryParameter(String key, Object paramValue, StringBuilder stringBuilder)
+        throws UnsupportedEncodingException
+    {
+        if (paramValue instanceof String) {
+            stringBuilder.append(URLEncoder.encode(key, "UTF-8"));
+            stringBuilder.append("=");
```
tmortagne:
```suggestion
            stringBuilder.append('=');
```

Better for performances.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2443,9 +2474,40 @@ public String getSkinFile(String filename, boolean forceSkinAction, XWikiContext
...
+            parameters.put(CACHE_VERSION, getVersion());
+        }
+
+        return parameters;
+    }
+
     public String getSkinFile(String filename, String skin, XWikiContext context)
     {
-        return getSkinFile(filename, skin, false, context);
+        return getSkinFile(filename, skin, context);
```
tmortagne: Is this change on purpose ? Looks like an infinite loop.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
surli: Right.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
surli: I don't think so. Too many changes and rollbacks on this one :)

PR #1016, opened 2019-01-09
XWIKI-16022: Limit the generic page picker search to a space

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-16022

### Change

* Add a way to limit the page picker search under a certain space.

### Usage

```
{{velocity}}
{{html}}
  #set ($param = {'data-suggest-space': 'XWiki'})
  #pagePicker($param)
{{/html}}
{{/velocity}}
```

dm4lxw


mflorea:
Note that from the user perspective XWiki has **nested pages** so an end user will rather ask for "limit the generic page picker search to the **children / descendants of a page**" (i.e. to a sub-tree of the page hierarchy). What you did is good for **nested spaces**, but if someone will want to use it for **nested pages** then there's a small issue: the parent page (the home page of the specified space, ``WebHome``) will be included in the suggestions. This is also going to be visible on the list of suggestions where the path (hint) of the space home page will be shorter.

PR #1019, opened 2019-01-10
XWIKI-16024: Selecting an option from the drop-down that lists available Skins navigates to the page

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-16024

### Changes

* Preload selectize suggestions earlier.
* Prevent opening the links of the selectize elements.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
@@ -116,7 +116,7 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
       dropdown.addClass('active');
     },
     persist: false,
-    preload: 'focus',
+    preload: true,
     render: {
       item: renderItem,
       option: renderOption,
@@ -135,6 +135,12 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
         var wrapper = this.$wrapper;
         wrapper.addClass(this.settings.loadingClass);
       }
+    },
+    onItemAdd: function(value, item) {
+      item.find('a.xwiki-selectize-option-label').click(function(event) {
+        // Clicking on the label should select the option not follow the link.
+        event.preventDefault();
+      });
     }
   };
 
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
@@ -62,10 +62,7 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
     }
     var url = option && option.url;
     if (typeof url === 'string') {
-      var anchor = $('<a class="xwiki-selectize-option-label" />').attr('href', url).click(function(event) {
-        // Clicking on the label should select the option not follow the link.
-        event.preventDefault();
-      });
+      var anchor = $('<a class="xwiki-selectize-option-label" />').attr('href', url);
       output.find('.xwiki-selectize-option-label').replaceWith(anchor);
     }
     var label = (option && typeof option === 'object') ? (option.label || option.value) : option;
@@ -116,7 +113,7 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
       dropdown.addClass('active');
     },
     persist: false,
-    preload: true,
+    preload: 'focus',
     render: {
       item: renderItem,
       option: renderOption,
@@ -136,8 +133,8 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
         wrapper.addClass(this.settings.loadingClass);
       }
     },
-    onItemAdd: function(value, item) {
-      item.find('a.xwiki-selectize-option-label').click(function(event) {
+    onInitialize: function() {
+      this.${escapetool.d}control.on('click', 'a.xwiki-selectize-option-label', function(event) {
         // Clicking on the label should select the option not follow the link.
         event.preventDefault();
       });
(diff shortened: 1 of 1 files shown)


mflorea:
@atallahade I would like to understand why the link is not followed all the time. @evalica said

>  **at a point** you will be redirected to the S1 or S2 page

This tells me that there is code that **already prevents the default behavior** when the link is clicked, otherwise the page picker and the user picker would be unusable since both use links on the suggestion label. So there must be some edge case that is not covered by the existing code and I think we should understand why before trying to fix the problem.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
```diff
@@ -116,7 +116,7 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
       dropdown.addClass('active');
     },
     persist: false,
-    preload: 'focus',
+    preload: true,
```
mflorea: Why is this change needed?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
```diff
@@ -135,6 +135,12 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
         var wrapper = this.$wrapper;
         wrapper.addClass(this.settings.loadingClass);
       }
+    },
+    onItemAdd: function(value, item) {
+      item.find('a.xwiki-selectize-option-label').click(function(event) {
+        // Clicking on the label should select the option not follow the link.
+        event.preventDefault();
+      });
```
mflorea: You don't need to add the event listener on each suggestion, each time we add a suggestion. You can add the event listener **once** on a parent element that is always there and use the 'selector' to filter where the event comes from (since the event is propagated upwards), see https://api.jquery.com/on/#on-events-selector-data-handler .

atallahade:
@mflorea To reproduce the issue, you should select an element `A` from the suggestion list, select another element and select back the element `A`. Clicking the element from the input will trigger a redirection.

The code responsible for preventing the redirection in the first place is:
https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js#L64-L70

I don't know exactly why it stop working after doing the above though, I'll investigate.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
mflorea: That's a separate thing IMO, not related to this issue. And there's no right / perfect choice. Preloading is nice for the user but it's not nice for the server. You don't know if the user will use the picker so you might load the default suggestions for nothing. On the other hand, if we load the default suggestions on focus, as we do now, we don't put the server at work for nothing.

Note that I think @evalica actually "complained" about the fact that you see one suggestion in the list (**the selected value**) and then, after 1 sec, the rest of the suggestions. But that's OK IMO because the selected value you already have it, while the list of suggestions need to be fetched.

In any case, if @evalica really wants to have the suggestions preloaded then she should open a separate issue, which would be an improvement.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
atallahade: @mflorea I did as you suggested which is a way better way of doing what we need and solves the issue.

atallahade:
@mflorea The issue is cause by:
https://github.com/selectize/selectize.js/blob/8c76832f95bae664fb1b533c41df414fc6192ebd/src/selectize.js#L1828
```js
clear: function(silent) {
  var self = this;
  //[...]
  self.$control.children(':not(input)').remove();
  //[...]
}
```
This function is called when we add an element (in a non multi selectize input).
According to the [documentation](https://api.jquery.com/remove/), all event handlers are removed from the element.

This following code is called only once by selectize and then cached (which mean the event handler is not added once removed and the element is not recreated).
https://github.com/xwiki/xwiki-platform/blob/b5056fede07fabf6224250ed4b8181a37257cbd8/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js#L64-L70

The cache takes place at https://github.com/selectize/selectize.js/blob/8c76832f95bae664fb1b533c41df414fc6192ebd/src/selectize.js#L2156 and https://github.com/selectize/selectize.js/blob/8c76832f95bae664fb1b533c41df414fc6192ebd/src/selectize.js#L2130.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
```diff
@@ -135,6 +132,12 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
         var wrapper = this.$wrapper;
         wrapper.addClass(this.settings.loadingClass);
       }
+    },
+    onInitialize: function() {
+      this.${escapetool.d}control.on('click', 'a.xwiki-selectize-option-label', function(event) {
```
mflorea: Are you sure this line can be minified? ``${escapetool.d}`` doesn't look like valid JavaScript. This file is minified at build time and then later evaluated for Velocity at runtime so the Velocity code needs to appear as JavaScript code. Assigning ``this.$control`` to a variable before adding the event might be an easier fix.

mflorea:
@atallahade thanks for the investigation. Makes sense.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
atallahade: Are you suggesting the following?
```suggestion
      var control = this.${escapetool.d}control;
      control.on('click', 'a.xwiki-selectize-option-label', function(event) {
```

atallahade: Seems like the right way to do it, I'll run some test and apply your suggestion.

PR #1021, opened 2019-01-14
XWIKI-15031: Move to sbraconnier's JODConverter

surli:
* Fix revapi exception in legacy-office-importer


tmortagne:
So it's OK to break this API ?

tmortagne:
> AFAIC it's not a signature breakage at least ;) (doesn't need recompilation)

The signature does not really matter here. Since the return type is completely different any code which use it will produce a ClassCastException at runtime and you cannot build if you don't modify your code. So it's definitely both a binary and source breakage.

The only thing I can think of would be to have the exposed `org.artofsolving.jodconverter.*` classes on our side calling the wrapping `org.jodconverter.` ones but it's going way too far down the rapid hole from what I have seen and I doubt it's used enough to justify this.

PR #1025, opened 2019-01-14
XWIKI-15182: Wrong action in XWikiContext for extensions contributing a new entity action URL handler

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15182

## Solution
  * Remove remaining TODOS from a closed issue

@vmassol could you review this one? it's unclear when reading your comments on the issue if this should be indeed removed after the issue 15182 is closed or if we should wait another refactoring?


vmassol:
@surli I don't understand. This PR references XWIKI-15182, but XWIKI-15182 is already closed (was closed in 10.3). So we cannot commit something against an issue that was already closed and released. If you mean that there are leftover comments that should be removed then it could be a "[Misc] .... (related to XWIKI-15182: ...)"

Checking the comments.

vmassol:
Just noticed that you didn't change just comments. So this PR will create a blocker bug since it puts back what XWIKI-15182 removed AFAICS :)

PR #1026, opened 2019-01-14
[Misc] Remove dependency between platform-resource-servlet and platform-resource-default

surli:
* Don't use concrete classes SkinResourceReference and
ResourceResourceReference in RoutingFilter but rely on a new API from
ResourceType.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
+++ b/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
@@ -35,9 +35,6 @@ public class ResourceType
      */
     private String id;
 
-    /**
-     * True if this type represents a static resource.
-     */
     private boolean isStatic;
 
     /**
@@ -99,7 +96,7 @@ public class ResourceType
     }
 
     /**
-     * @return true if this type represents a static resource
+     * @return true if this type represents a static resource (i.e: a ResourcesResource or a SkinResource only)
      */
     public boolean isStatic()
     {
--- a/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
+++ b/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
@@ -21,6 +21,7 @@ package org.xwiki.resource;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.xwiki.stability.Unstable;
 
 /**
  * Represents the type of Resource (eg Entity, Temporary, etc).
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
```diff
@@ -35,12 +35,27 @@
      */
     private String id;
 
+    /**
+     * True if this type represents a static resource.
```
vmassol: This should explain what is a static resource as otherwise it's useless :) (and it would be good to know what is a static resource here)

vmassol: Actually you should define that in `isStatic()` and link to it.

vmassol:
Sounds good otherwise!

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
```diff
@@ -82,4 +94,12 @@ public boolean equals(Object object)
             .append(getId(), rhs.getId())
             .isEquals();
     }
+
+    /**
+     * @return true if this type represents a static resource (i.e: a ResourcesResource or a SkinResource only)
```
vmassol: That's not the best definition  IMO :) A static resource contains static content and must be served by the Servlet Container directly, without going through XWiki's processing and handling.

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
```diff
@@ -35,12 +35,24 @@
...
     public ResourceType(String id)
+    {
+        this(id, false);
+    }
+
+    /**
+     * @param id see {@link #getId()}
+     * @param isStatic see {@link #isStatic()}
+     */
+    public ResourceType(String id, boolean isStatic)
```
tmortagne: Missing `@since`.

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
```diff
@@ -82,4 +94,12 @@ public boolean equals(Object object)
             .append(getId(), rhs.getId())
             .isEquals();
     }
+
+    /**
+     * @return true if this type represents a static resource (i.e: a ResourcesResource or a SkinResource only)
+     */
+    public boolean isStatic()
```
tmortagne: Missing `@since`.

vmassol: And `@Unstable`

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-api/src/main/java/org/xwiki/resource/ResourceType.java
vmassol: And `@Unstable`

surli:
I applied the change so it's ready again for review.

vmassol:
Thanks Simon!

PR #1034, opened 2019-01-17
XWIKI-16044: className vs classname inconsistency in "objectremove" and "objectadd" actions

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-16044

### Changes

* Use 'classname' instead of 'className' as this is the one use
  everywhere else.
* Add support for both 'className' and 'classname' (for backward
  compatibility).

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ObjectAddForm.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ObjectAddForm.java
@@ -47,7 +47,12 @@ public class ObjectAddForm extends XWikiForm
     @Override
     public void readRequest()
     {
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ObjectAddForm.java
```diff
@@ -47,7 +47,12 @@ public void setClassName(String className)
     @Override
     public void readRequest()
     {
-        setClassName(getRequest().getParameter("classname"));
+        // We need to support both 'classname' and 'className' since an inconsistency has been introduced.
```
vmassol: I'd rephrase that a bit saying that we keep "className" for backward compatibility and that we're updating to "classname" to be consistent with the parameter name used for Class Name for other actions.

PR #1036, opened 2019-01-18
XWIKI-16044: className vs classname inconsistency in "objectremove" and "objectadd" actions

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-16044

### Changes

* Use 'classname' instead of 'className' as this is the one use everywhere else.


tmortagne:
It sure is inconsistent but this commit is an API breakage. At the very least you should make sure `className` keep working.

PR #1039, opened 2019-01-21
XWIKI-16058: Cannot "Select all" the pages when deleting AWM apps

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16058

## Solution

  * Remove the checkbox of "objects" instead of just hiding them.
  * Remove the links on XClass

peek 01-02-2019 12-26

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
@@ -124,6 +124,10 @@ $response.setContentType("application/json")
       'children': true,
       'state': {
         'selected': $selected
+      },
+      'a_attr': {
+        'href': $pageDoc.getURL('view'),
+        'class': 'jstree-no-links'
       }
     }))
   #elseif ($type == 'freePages')
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
@@ -124,9 +124,6 @@ $response.setContentType("application/json")
       'children': true,
       'state': {
         'selected': $selected
-      },
-      'data': {
-        'link': $pageDoc.getURL('view')
       }
     }))
   #elseif ($type == 'freePages')
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
@@ -126,8 +126,7 @@ $response.setContentType("application/json")
         'selected': $selected
       },
       'a_attr': {
-        'href': $pageDoc.getURL('view'),
-        'class': 'jstree-no-links'
+        'href': $pageDoc.getURL('view')
       }
     }))
   #elseif ($type == 'freePages')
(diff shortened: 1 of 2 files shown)


surli:
@mflorea would you review this one? for me it's ready

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/job/question/XClassBreakingQuestion.data.vm
```diff
@@ -124,9 +124,6 @@ $response.setContentType("application/json")
       'children': true,
       'state': {
         'selected': $selected
-      },
-      'data': {
-        'link': $pageDoc.getURL('view')
```
mflorea: Having a link on the tree node is still useful even if not by clicking on it but by opening the link on a new browser tab. I would do it using the [``a_attr``](https://www.jstree.com/docs/json/) property though.

surli: I think the problem would be the same since the link is around the checkbox too: so selecting the element would always open the link, which is not expected by the user...

mflorea: The example from https://www.jstree.com/plugins/ (look for the checkbox plugin) has nodes with links. The browser shows the URL when you hover the node and you have the option to open the link in a new browser tab when opening the contextual menu (Right click).

Also, I remember the [tree widget](https://extensions.xwiki.org/xwiki/bin/view/Extension/Tree%20Widget) has a special ``jstree-no-links`` CSS class.

surli: So a link with `jstree-no-links` would only work when right clicking on it?

mflorea: That's the behavior of the location picker tree from the Create Page form. Clicking on a node selects the node but doesn't follow the link, and you can open the link in a new browser tab. The checkbox plugin is not used there though.

surli: Hm ok. I'll check if you think it can be helpful, but frankly as the link is a bit "hidden" because of the checkbox, I'm not sure it worths it to keep it here.
I assume that in most cases, the user would open a new tab a go manually to the appropriate page if needed.

surli:
So @mflorea I put back the links, but really the UI is far from being obvious:
peek 23-01-2019 15-11

as you can see the mouse remain exactly the same on a node with a link and another one "without a link". I use "" here because in fact on the other nodes you can also open them in a new tab (i.e.: there's a link) but it redirects just to the current page... 
So maybe the problem is more about the default value for the nodes, which shouldn't have any associated link. WDYT?

mflorea:
Sure, the node should have a link only if it's entity has a link (e.g. page or attachment).

surli:
so, something to improve on tree widget side? or it's already supposed to be like that?

surli:
>  You need to put the jstree-no-links class on the tree element (div) not on the node.

ok but then we cannot distinguish any of the link :thinking: 
capture d ecran_2019-01-23_15-53-44

So on the contrary I should put it only on the node that are not a link, or am I missing something?

surli:
@mflorea finally I put the class on the whole tree, since when I tested to put the class on the node without link:
  1. I still get the folder "Pages that do not contain any used XClass" as a link
  2. when selecting a node I was redirected to its link in the same window

However, as I mentioned in my previous comment, now we cannot distinguish the nodes with or without link (see the screenshot of the previous comment). So I'm still in favor of removing those links.

mflorea:
These nodes don't have to **appear** as links because they are not meant to be used as links. You don't want to leave the current page when you click on one of the nodes. In other words, this tree is not meant for navigation.

On the other hand, it's **very useful** to be able to open the page in a separate browser tab if you want to find more about it.

Compare the tree from the Navigation Panel with the tree from the Create/Copy/Rename page UI. This tree should behave similar to the later.

surli:
> This tree should behave similar to the later.

So it's the case with my last changes.

surli:
@vmassol @tmortagne would you merge?

PR #1047, opened 2019-01-28
[Misc] Improve tests in office-importer.

surli:
This should fix pitest job.


xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
```diff
@@ -52,12 +52,12 @@
     /**
      * Error message used to indicate that office server administration is restricted for main wiki.
      */
-    private static final String ERROR_FORBIDDEN = "Office server administration is forbidden for sub-wikis.";
+    protected static final String ERROR_FORBIDDEN = "Office server administration is forbidden for sub-wikis.";
```
vmassol: I'm not a big fan of changing visibility just for a test. This makes it a SPI... At minium you should use package level visibility instead.

But IMO you shouldn't even need this and instead copy the string.

surli: OK. I hesitated to copy the string, I'll do that.

PR #1049, opened 2019-01-30
XWIKI-15982: dependency issue: profile integration-tests required for xwiki-platform-test-upgrade

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15982

## Solution
  * test-upgrade and test-ui to the right profile

### DIFF
--- a/xwiki-platform-core/xwiki-platform-test/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-test/pom.xml
@@ -42,7 +42,14 @@
     <module>xwiki-platform-test-integration</module>
     <module>xwiki-platform-test-oldcore</module>
     <module>xwiki-platform-test-page</module>
-    <module>xwiki-platform-test-ui</module>
-    <module>xwiki-platform-test-upgrade</module>
   </modules>
+  <profiles>
+    <profile>
+      <id>integration-tests</id>
+      <modules>
+        <module>xwiki-platform-test-ui</module>
+        <module>xwiki-platform-test-upgrade</module>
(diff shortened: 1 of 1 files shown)


surli:
This is a first proposal actually AFAIU @vmassol suggested to move test-upgrade to the extension test module. According to the dependencies that could be done, now I don't really know the impact or if it's semantically correct.

xwiki-platform-core/xwiki-platform-test/pom.xml
```diff
@@ -42,7 +42,14 @@
...
     <module>xwiki-platform-test-page</module>
-    <module>xwiki-platform-test-ui</module>
-    <module>xwiki-platform-test-upgrade</module>
   </modules>
+  <profiles>
+    <profile>
+      <id>integration-tests</id>
+      <modules>
+        <module>xwiki-platform-test-ui</module>
+        <module>xwiki-platform-test-upgrade</module>
```
vmassol: If we do this, we should also move test-docker in the `integration-tests` profile IMO.

vmassol: The suggested change will prevent building these modules when not using the integration tests profiles, so a bit more build perf (always good to have though :)).

Now the root problem is not this one. The root problem is that the upgrade test should not depend on something in extension IMO. It could be fixed in another issue though.

surli: I also moved platform-test-docker. I let you create the other issue.

PR #1050, opened 2019-01-30
XWIKI-13362: Display the page reference for a page, for users to be able to copy-paste it

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-13362

## Solution

peek 05-02-2019 17-38

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
@@ -1,0 +1,53 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+require(['jquery', 'xwiki-meta'], function($, xm) {
+
+  $(document).on('click', '#button-inout', function() {
+    var button = $('#button-inout');
+    var referenceValue= $('#reference-value');
+    var localReference = xm.documentReference.relativeTo(xm.documentReference.getRoot());
+    var globalReference = xm.documentReference;
+
+    if (button.hasClass('btn-default')) {
+      referenceValue.text(globalReference);
+      button.removeClass('btn-default');
+      button.attr('title',"$services.localization.render('core.viewers.information.pageReference.localButton')");
+    } else {
+      referenceValue.text(localReference);
+      button.addClass('btn-default');
+      button.attr('title',"$services.localization.render('core.viewers.information.pageReference.globalButton')");
+    }
+  });
+
+  var copyToClipboard = function(element) {
+    var $temp = $("<input>");
+    $("body").append($temp);
+    var value = $(element).text();
+    $temp.val(value).select();
+    document.execCommand("copy");
+    $temp.remove();
+    return value;
+  };
+
+  $(document).on('click', '#button-paste', function() {
+    copyToClipboard('#reference-value');
+    alert("$services.localization.render('core.viewers.information.pageReference.copied')");
+  });
+});
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
@@ -33,7 +33,6 @@ require(['jquery', 'xwiki-meta'], function($, xm) {
       referenceValue.text(globalReference);
       button.addClass('btn-info');
       button.attr('title',"$services.localization.render('core.viewers.information.pageReference.localButton')");
-
     }
   });
 
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
@@ -25,14 +25,15 @@ require(['jquery', 'xwiki-meta'], function($, xm) {
     var localReference = xm.documentReference.relativeTo(xm.documentReference.getRoot());
     var globalReference = xm.documentReference;
 
-    if (button.hasClass('btn-default')) {
-      referenceValue.text(globalReference);
-      button.removeClass('btn-default');
-      button.attr('title',"$services.localization.render('core.viewers.information.pageReference.localButton')");
-    } else {
+    if (button.hasClass('btn-info')) {
       referenceValue.text(localReference);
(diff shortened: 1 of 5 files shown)


surli:
@mflorea @evalica could you review this one please?

vmassol:
Just checked your animated GIF and it looks good. I would just change one thing: I would not open a dialog to validate the copy in the clipboard. Instead I would use the notification we have at the bottom of the screen with a transient message in green that the reference has been copied in the clipboard. I feel it's nicer for the user flow. WDYT?

surli:
> Instead I would use the notification we have at the bottom of the screen with a transient message in green that the reference has been copied in the clipboard. I feel it's nicer for the user flow. WDYT?

yep. I wasn't sure about the UI for this. I'll change.

vmassol:
Nice :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
```diff
@@ -20,6 +21,19 @@
         #end
       </dd>
     </dl>
+    <dl>
```
mflorea: Do you really need to open a new definition list instead of adding the ``dt``/``dd`` to the previous list?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
```diff
@@ -20,6 +21,19 @@
...
+      <dt>$services.localization.render('core.viewers.information.pageReference')</dt>
+      <dd id="information-reference">
+        <span class="noitems">$services.localization.render('core.viewers.information.pageReference.tips')</span>
```
mflorea: We put the hint in the ``dt`` when [creating forms](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/VerticalForms/#H4.Hint), but I guess it's OK in the ``dd`` also (this is not an actual form).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
```diff
@@ -20,6 +21,19 @@
...
+        <div class="input-group">
+          <div class="input-group-addon btn btn-default" id="button-inout" title="$services.localization.render('core.viewers.information.pageReference.globalButton')">
+            <span class="fa fa-globe"></span>
+          </div>
+          <span class="form-control" id="reference-value">$doc.getDocumentReference().getLocaleDocumentReference()</span>
```
mflorea: How is a very long reference displayed?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
surli: I guess I'm not. I'll remove it.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
surli: I followed [the snippet](https://jira.xwiki.org/browse/XWIKI-13362?focusedCommentId=101580&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-101580) proposed by @evalica for this one and it looks good. I don't think it's breaking a WCAG rule either, but let me know if you think I should change.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
surli: Right. I need to check that.

vmassol: Note that I've proposed a solution that could solve this in the jira issue. It's to have UIXP and thus have a tabular layout to support that. See https://jira.xwiki.org/browse/XWIKI-13362?focusedCommentId=92183&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-92183

surli: > have a tabular layout

you mean displaying backlinks and included pages in other tabs, right?

vmassol: Nope :) In the comment I linked, there's a screenshot (from jira I think) showing a tabular approach, i.e. one under another and taking the full width.

Now we could still implement it generically (UIXP) and have a grid layout, even if it's harder to implement and may fit less well depending on the UIXs that implement the UIXP.

surli: I don't really see how it would fix this issue: if the ref is long enough it would just make an overflow in the table cell. 

Moreover, on another topic I'm not convinced about having a cell with all backlinks and another one with all included pages. I've the feeling it would create an ugly table really long to scroll.

surli:
@mflorea I finally had to rewrite the template and the CSS as I didn't manage to user directly the overflow with the given bootstrap classes because it used a `display: table`.
But with the help of @evalica I managed to obtain the result shown in the description. It's not perfect but I think it's closed enough to the original result and it now supports the overflow.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/viewers/information.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (button.hasClass('btn-info')) {
+      referenceValue.text(localReference);
+      button.removeClass('btn-info');
+      button.attr('title',"$services.localization.render('core.viewers.information.pageReference.globalButton')");
+    } else {
+      referenceValue.text(globalReference);
+      button.addClass('btn-info');
+      button.attr('title',"$services.localization.render('core.viewers.information.pageReference.localButton')");
+
```
mflorea: Extra empty line.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/informationinline.vm
```diff
@@ -1,40 +1,55 @@
...
+          #end
+        </dd>
+        <dt>$services.localization.render('core.viewers.information.pageReference')</dt>
+        <dd id="information-reference">
+          <span class="noitems">$services.localization.render('core.viewers.information.pageReference.tips')</span>
+          <div class="reference-group">
+            <div class="btn btn-default" id="button-inout" title="$services.localization.render('core.viewers.information.pageReference.globalButton')">
+              $services.icon.renderHTML('world')
+            </div>
+            <span class="form-control" id="reference-value">$doc.getDocumentReference().getLocaleDocumentReference()</span>
```
mflorea: The reference needs to be XML encoded. Also, relying on ``toString()`` to get the string reference is a bit risky.

surli: > relying on `toString()` to get the string reference is a bit risky.

what do you suggest? using a string resolver?

mflorea: My suggestion is to serialize the reference. BTW, ``getLocaleDocumentReference`` [is deprecated](https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-11.0/xwiki-platform-core/xwiki-platform-model/src/main/java/org/xwiki/model/reference/DocumentReference.java#L351). You should use ``getLocalDocumentReference`` (which makes more sense anyway).

```suggestion
            <span class="form-control" id="reference-value">$escapetool.xml($services.model.serialize($doc.getDocumentReference().getLocalDocumentReference()))</span>
```

surli: Hm indeed I meant serializer, not resolver. My bad. Thanks for the clarification :) Interesting to see how easy it is in velocity to use deprecated stuff...

tmortagne: > Interesting to see how easy it is in velocity to use deprecated stuff

But you get a warning in the log when you do that.

PR #1051, opened 2019-01-31
XWIKI-15641: Display information to user when a job is waiting another to finish/timeout

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15641

## Solution

  * Add a simple info box when a job is waiting to start

screenshot_2019-01-31 home - xwiki

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/resources/ApplicationResources.properties
@@ -66,3 +66,5 @@ job.question.OverwriteQuestion.lastModified=Last modified: {0}
 job.question.OverwriteQuestion.applyToAll=Apply this action to all the following pages.
 job.question.OverwriteQuestion.replace=Replace
 job.question.OverwriteQuestion.skip=Skip
+
+job.waiting=This job is waiting to start, please be patient: another job might be running on the same pages.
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/resources/ApplicationResources.properties
```diff
@@ -65,4 +65,6 @@ job.question.OverwriteQuestion.lastAuthor=Last author: {0}
 job.question.OverwriteQuestion.lastModified=Last modified: {0}
 job.question.OverwriteQuestion.applyToAll=Apply this action to all the following pages.
 job.question.OverwriteQuestion.replace=Replace
-job.question.OverwriteQuestion.skip=Skip
\ No newline at end of file
+job.question.OverwriteQuestion.skip=Skip
+
+job.waiting=This job is waiting to start, please be patient: another job might be running on the same pages.
```
tmortagne: Should probably choose a different work since WAITING is a job state indicating that the job has been started but is currently waiting for an input (generally a question).

surli: You mean for the translation key right? job.waitForStarting? job.noneState.display?

surli: That works for me. I'll use it.

tmortagne:
Would be even nicer to indicate the blocker job but it's a good first step.

surli:
> Would be even nicer to indicate the blocker job but it's a good first step.

Would you create another improvment issue for that? With some details on what info about the blocker job you'd like?

tmortagne:
> what info about the blocker job you'd like?

Just what it is would already be nice :) But before that I think we need a generic job UI in the admin so that we can link to it, otherwise all we can do is print the job id basically (at least in the generic way, specific uses case like extension can do specific stuff).

PR #1052, opened 2019-02-04
XWIKI-16093: Add extension point to make the HTML head pluggable

slauriere:
- Modify XWikiPreferences "meta" property to add new extension point.

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
@@ -2815,6 +2815,7 @@ Your account on ${wikiname} has been successfully activated. You can now login a
 &lt;meta name="keywords" content="$escapetool.xml($keywords.replaceAll('[|,]', ' ').trim())" /&gt;
 &lt;meta name="rating" content="General" /&gt;
 &lt;meta name="author" content="$escapetool.xml($!xwiki.getUserName($doc.getAuthor(), false))" /&gt;
+## Render extensions adding content to the page head, if any.
 #foreach($uix in $services.uix.getExtensions("org.xwiki.platform.head", {'sortByParameter' : 'order'}))
   $services.rendering.render($uix.execute(), 'xhtml/1.0')
 #end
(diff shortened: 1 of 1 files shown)


xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
```diff
@@ -2815,6 +2815,9 @@ Your account on ${wikiname} has been successfully activated. You can now login a
 &lt;meta name="keywords" content="$escapetool.xml($keywords.replaceAll('[|,]', ' ').trim())" /&gt;
 &lt;meta name="rating" content="General" /&gt;
 &lt;meta name="author" content="$escapetool.xml($!xwiki.getUserName($doc.getAuthor(), false))" /&gt;
+#foreach($uix in $services.uix.getExtensions("org.xwiki.platform.head", {'sortByParameter' : 'order'}))
```
vmassol: Since users will see this in the Admin UI, I believe it's very important to have some comments to explain what this is about.

I was hesitating between putting this in the Admin UI field, or in the template file for the header. The advantage of putting it in the Admin UI field, is that it's easier to control if the user doesn't want it vs having to disable UIXs. OTOH, it would be more consistent with other UIXP to have it in the template.

WDYT?

xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
```diff
@@ -2815,6 +2815,10 @@ Your account on ${wikiname} has been successfully activated. You can now login a
 &lt;meta name="keywords" content="$escapetool.xml($keywords.replaceAll('[|,]', ' ').trim())" /&gt;
 &lt;meta name="rating" content="General" /&gt;
 &lt;meta name="author" content="$escapetool.xml($!xwiki.getUserName($doc.getAuthor(), false))" /&gt;
+## Render extensions adding content to the page head, if any.
```
vmassol: Suggestion: 

"Add HTML tags inside the page's <HEAD> element, provided by UI Extensions implementing the "org.xwiki.platform.head" Extension Point"

slauriere: Thanks :-)

vmassol:
I'm applying it and we can iterate to improve or move it in a template later on if we think it's better.

Note that we will have a problem with users having customized their HTTP Meta Info since that will probably lead to conflict in the DW. If we move it to a template we won't have this issue.

slauriere:
Great, thanks @vmassol

PR #1055, opened 2019-02-07
XWIKI-6732: Introduce DocumentRenamingEvent and DocumentRenamedEvent events

surli:
* Unregister new listener components in some module tests to avoid
missing dependencies.


mflorea:
Thanks @surli!

PR #1056, opened 2019-02-07
 XWIKI-6732: Introduce DocumentRenamingEvent and DocumentRenamedEvent events

surli:
* Unregister new listener components in some module tests to avoid
missing dependencies.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/test/java/org/xwiki/extension/xar/internal/handler/WikiCopiedEventListenerTest.java
+++ b/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/test/java/org/xwiki/extension/xar/internal/handler/WikiCopiedEventListenerTest.java
@@ -35,6 +35,7 @@ import org.xwiki.extension.repository.internal.local.DefaultLocalExtensionReposi
 import org.xwiki.extension.test.EmptyExtension;
 import org.xwiki.extension.test.MockitoRepositoryUtilsRule;
 import org.xwiki.extension.version.internal.DefaultVersionConstraint;
+import org.xwiki.observation.EventListener;
 import org.xwiki.observation.ObservationManager;
 import org.xwiki.security.authorization.AuthorizationManager;
 import org.xwiki.security.authorization.ContextualAuthorizationManager;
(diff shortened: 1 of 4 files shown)


surli:
@tmortagne another batch of tests fixing

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/test/java/org/xwiki/extension/xar/internal/handler/WikiCopiedEventListenerTest.java
```diff
@@ -86,6 +87,15 @@ public void setUp() throws Exception
 
         this.localExtensionDependency1 = this.localExtensionRepository.storeExtension(extensionDependency);
         this.localExtension1 = this.localExtensionRepository.storeExtension(extension);
+
+        this.repositoryUtil.getComponentManager().unregisterComponent(EventListener.class,
```
tmortagne: I feel that registering a mock of ModelBridge (unless one already exist) in MockitoOldcore like we do for other components would be better and easier for new tests than copy/pasting this everywhere.

surli: I agree that it'll be more elegant, I'm working on that. However it won't be enough: here for example, we're not using MockitoOldCore. Now I noticed that I have less mock component to register, than listener to unregister, so I'll go toward this solution

tmortagne: > we're not using MockitoOldCore

Sure but in many of the affected test we do and it's even possible some of those not using MockitoOldCore should.

surli: > t's even possible some of those not using MockitoOldCore should (or should not use `@AllComponent`).

yep, but I don't necessarily plan to refactor them ;)

surli:
Closed in favor of https://github.com/xwiki/xwiki-platform/pull/1059

PR #1057, opened 2019-02-07
XWIKI-11447: Error message when trying to load annotations on search page

acotiuga:
* block the redirects as they are leading to an unexpected result for the annotations UI


lucaa:
@tmortagne IIRC, DefaultAnnotationService#getAnnotatedRenderedContent() is supposed to be equivalent to $doc.getRenderedContent(), which can do redirects. So I'd say that if there is any redirect, it should do it.
I think it's ok for it to be done on the REST request. In any case this strategy can't actually work properly for getting the annotated content, we're doing it to allow the REST service to return properly a JSON response to the caller saying there are 0 annotations on that document (because the number of annotations and the annotated content are in the same answer). But the annotated content cannot be correct if we interfere with the execution of the script by stopping redirects.

A more complete solution would be to copy all request parameters from the current view and send them to the REST service so that they are used when rendering the content of the page. This would allow the rendered content to be more accurate (as it would be done in the same conditions as the browser) and in the case of the Search page, it would prevent redirect - since redirect depends on the received request params. However, I think the complete solution is out of scope of this issue, we'd need to create another one.

tmortagne:
OK thanks @lucaa .

PR #1058, opened 2019-02-08
XWIKI-16115: Translation keys for solr facet is broken on backend

CamiAndrei:


tmortagne:

PR #1061, opened 2019-02-09
XWIKI-16120: some object properties are not part of the public api

ClemensRobbenhaar:
as discussed on the mailing list

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
@@ -102,6 +102,8 @@ import org.xwiki.rest.resources.spaces.SpaceSearchResource;
 import org.xwiki.rest.resources.spaces.SpacesResource;
 import org.xwiki.rest.resources.wikis.WikiSearchQueryResource;
 import org.xwiki.rest.resources.wikis.WikiSearchResource;
+import org.xwiki.security.authorization.ContextualAuthorizationManager;
+import org.xwiki.security.authorization.Right;
 
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
@@ -136,6 +138,9 @@ public class ModelFactory
 
     @Inject
     private JAXBConverter jaxbConverter;
+
+    @Inject
+    private ContextualAuthorizationManager authorizationManager;
 
     @Inject
     private Logger logger;
@@ -273,9 +278,7 @@ public class ModelFactory
     public void toObject(com.xpn.xwiki.api.Object xwikiObject, org.xwiki.rest.model.jaxb.Object restObject)
     {
         for (Property restProperty : restObject.getProperties()) {
-            if (isPublicType(restProperty, true)) {
-                xwikiObject.set(restProperty.getName(), restProperty.getValue());
-            }
+            xwikiObject.set(restProperty.getName(), restProperty.getValue());
         }
     }
 
@@ -333,11 +336,12 @@ public class ModelFactory
 
             property.setName(propertyClass.getName());
             property.setType(propertyClass.getClassType());
-            if (isPublicType(property, false)) {
+            if (isPublicType(property)) {
                 try {
                     property.setValue(serializePropertyValue(xwikiObject.get(propertyClass.getName())));
                 } catch (XWikiException e) {
                     // Should never happen
+                    logger.warn("unexpected exception when accessing property ["+propertyClass.getName()+"]", e);
                 }
             }
 
@@ -1077,13 +1081,12 @@ public class ModelFactory
     /**
      * Check if the given property should be exposed via REST.
      * @param restProperty the property to be read/written
-     * @param forWrite true if we want to write to the XObject property
-     * @return true if the property is considered public for read or write
+     * @return true if the property is considered public
      */
-    private boolean isPublicType(Property restProperty, boolean forWrite)
+    private boolean isPublicType(Property restProperty)
     {
         if (PASSWORD_TYPE.equals(restProperty.getType())) {
-            return false;
+            return authorizationManager.hasAccess(Right.ADMIN);
         }
 
         return true;
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
@@ -123,6 +123,8 @@ import com.xpn.xwiki.objects.classes.ListClass;
 @Singleton
 public class ModelFactory
 {
+    private static final String PASSWORD_TYPE = "Password";
+
     private final ObjectFactory objectFactory;
 
     @Inject
@@ -267,7 +269,9 @@ public class ModelFactory
     public void toObject(com.xpn.xwiki.api.Object xwikiObject, org.xwiki.rest.model.jaxb.Object restObject)
     {
         for (Property restProperty : restObject.getProperties()) {
-            xwikiObject.set(restProperty.getName(), restProperty.getValue());
+            if (!PASSWORD_TYPE.equals(restProperty.getType())) {
+                xwikiObject.set(restProperty.getName(), restProperty.getValue());
+            }
         }
     }
 
@@ -325,10 +329,12 @@ public class ModelFactory
 
             property.setName(propertyClass.getName());
             property.setType(propertyClass.getClassType());
-            try {
-                property.setValue(serializePropertyValue(xwikiObject.get(propertyClass.getName())));
-            } catch (XWikiException e) {
-                // Should never happen
+            if (!PASSWORD_TYPE.equals(property.getType())) {
+                try {
+                    property.setValue(serializePropertyValue(xwikiObject.get(propertyClass.getName())));
+                } catch (XWikiException e) {
+                    // Should never happen
+                }
             }
 
             String propertyUri;
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
@@ -37,6 +37,7 @@ import javax.xml.parsers.ParserConfigurationException;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.Logger;
 import org.xwiki.component.annotation.Component;
 import org.xwiki.job.DefaultRequest;
 import org.xwiki.job.Request;
@@ -46,6 +47,7 @@ import org.xwiki.logging.event.LogEvent;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.EntityReferenceSerializer;
 import org.xwiki.model.reference.ObjectReference;
+import org.xwiki.model.reference.SpaceReference;
 import org.xwiki.rest.Relations;
 import org.xwiki.rest.XWikiRestException;
 import org.xwiki.rest.model.jaxb.Attachment;
@@ -125,6 +127,8 @@ public class ModelFactory
 {
     private static final String PASSWORD_TYPE = "Password";
 
+    private static final String EMAIL_TYPE = "Email";
+
     private final ObjectFactory objectFactory;
 
     @Inject
@@ -135,6 +139,9 @@ public class ModelFactory
 
     @Inject
     private JAXBConverter jaxbConverter;
+
+    @Inject
+    private Logger logger;
 
     public ModelFactory()
     {
@@ -269,7 +276,7 @@ public class ModelFactory
     public void toObject(com.xpn.xwiki.api.Object xwikiObject, org.xwiki.rest.model.jaxb.Object restObject)
     {
         for (Property restProperty : restObject.getProperties()) {
-            if (!PASSWORD_TYPE.equals(restProperty.getType())) {
+            if (isTypePublic(restProperty, true)) {
                 xwikiObject.set(restProperty.getName(), restProperty.getValue());
             }
         }
@@ -329,7 +336,7 @@ public class ModelFactory
 
             property.setName(propertyClass.getName());
             property.setType(propertyClass.getClassType());
-            if (!PASSWORD_TYPE.equals(property.getType())) {
+            if (isTypePublic(property, false)) {
                 try {
                     property.setValue(serializePropertyValue(xwikiObject.get(propertyClass.getName())));
                 } catch (XWikiException e) {
@@ -1068,5 +1075,41 @@ public class ModelFactory
         }
 
         return log;
+    }
+
+    /**
+     * check if the given property should be exposed via REST
+     * @param restProperty the property to be read/written
+     * @param forWrite true iff we want to write to the XObject property
+     * @return true if the property is considered public for read or write
+     */
+    private boolean isTypePublic(Property restProperty, boolean forWrite)
+    {
+        if (PASSWORD_TYPE.equals(restProperty.getType())) {
+            return false;
+        }
+
+        if (!forWrite && EMAIL_TYPE.equals(restProperty.getType())) {
+            boolean obfuscate = true;
+            XWikiContext context = xcontextProvider.get();
+            try {
+                int configValue = -1;
+                XWikiDocument mailConfigDoc = context.getWiki().getDocument(new DocumentReference("MailConfig", new SpaceReference(context.getWikiId(), "Mail")), context);
+                BaseObject mailConfig = mailConfigDoc.getXObject(new DocumentReference("GeneralMailConfigClass", new SpaceReference(context.getWikiId(), "Mail")));
+                if (mailConfig != null) {
+                    configValue = mailConfig.getIntValue("obfuscate");
+                }
+                if (configValue < 0) {
+                    // we handle backward compatibility by also looking in the XWikiPreferences xobject
+                    configValue = context.getWiki().getXWikiPreferenceAsInt("obfuscateEmailAddresses", 1, context);
+                }
+                obfuscate = (configValue == 1);
+            } catch (XWikiException e) {
+                logger.warn("could not determine if emails should be obfuscated", e);
+            }
+            return ! obfuscate;
+        }
+
+        return true;
     }
 }
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
@@ -47,7 +47,6 @@ import org.xwiki.logging.event.LogEvent;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.EntityReferenceSerializer;
 import org.xwiki.model.reference.ObjectReference;
-import org.xwiki.model.reference.WikiReference;
 import org.xwiki.rest.Relations;
 import org.xwiki.rest.XWikiRestException;
 import org.xwiki.rest.model.jaxb.Attachment;
@@ -1087,8 +1086,7 @@ public class ModelFactory
     private boolean hasAccess(Property restProperty)
     {
         if (PASSWORD_TYPE.equals(restProperty.getType())) {
-            return authorizationManager.hasAccess(Right.ADMIN,
-                new WikiReference(xcontextProvider.get().getWikiId()));
+            return authorizationManager.hasAccess(Right.ADMIN, xcontextProvider.get().getWikiReference());
         }
 
         return true;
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -267,7 +269,9 @@ public ObjectSummary toRestObjectSummary(URI baseUri, Document doc, BaseObject x
     public void toObject(com.xpn.xwiki.api.Object xwikiObject, org.xwiki.rest.model.jaxb.Object restObject)
     {
         for (Property restProperty : restObject.getProperties()) {
-            xwikiObject.set(restProperty.getName(), restProperty.getValue());
+            if (!PASSWORD_TYPE.equals(restProperty.getType())) {
```
ClemensRobbenhaar: here I do not set the value in a PUT request. not sure if this is a good idea

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1063,4 +1076,40 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
...
+     * @param forWrite true iff we want to write to the XObject property
+     * @return true if the property is considered public for read or write
+     */
+    private boolean isTypePublic(Property restProperty, boolean forWrite)
+    {
+        if (PASSWORD_TYPE.equals(restProperty.getType())) {
+            return false;
+        }
+
+        if (!forWrite && EMAIL_TYPE.equals(restProperty.getType())) {
```
ClemensRobbenhaar: I have to admit that this part looks a bit like overkill; if this is not wanted I can roll back the last commit or remove this code from the PR.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1063,4 +1076,40 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
 
         return log;
     }
+
+    /**
+     * check if the given property should be exposed via REST
```
vmassol: Cosmetic: uppercase + dot at the end (will fail checkstyle otherwise).

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1063,4 +1076,40 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
...
+     * @param restProperty the property to be read/written
+     * @param forWrite true iff we want to write to the XObject property
```
vmassol: s/iff/if

vmassol: I'm not sure I understand the `forWrite` parameter.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1063,4 +1076,40 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
...
+     * @return true if the property is considered public for read or write
+     */
+    private boolean isTypePublic(Property restProperty, boolean forWrite)
```
vmassol: s/isTypePublic/isPublicType ?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1063,4 +1076,40 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
...
+                int configValue = -1;
+                XWikiDocument mailConfigDoc = context.getWiki().getDocument(new DocumentReference("MailConfig", new SpaceReference(context.getWikiId(), "Mail")), context);
+                BaseObject mailConfig = mailConfigDoc.getXObject(new DocumentReference("GeneralMailConfigClass", new SpaceReference(context.getWikiId(), "Mail")));
+                if (mailConfig != null) {
+                    configValue = mailConfig.getIntValue("obfuscate");
+                }
+                if (configValue < 0) {
+                    // we handle backward compatibility by also looking in the XWikiPreferences xobject
+                    configValue = context.getWiki().getXWikiPreferenceAsInt("obfuscateEmailAddresses", 1, context);
+                }
```
vmassol: I think this should go through some Java API and not directly get the value from XObjects. We have a ConfigurationSource for that :)

vmassol: Actually I'm not sure we have but we need to add one to the mail module if not. See for ex SendMailConfigClassDocumentConfigurationSource

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
vmassol: I thought we were already obfuscating the email in the REST API. This is not the case? If not, then yes it should be done. Would be better to have it in a different JIRA issue and different PR though.

vmassol:
Thanks a lot @ClemensRobbenhaar for working on this quickly :)

One thing we need too is a unit tests for it. I can probably help there too.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
ClemensRobbenhaar: If seems we have no config source for this. I copied the code from ``/templates/displayer_email.vm``, where it is done the same way after looking at the ``MailSenderConfigurationSource.java`` and did not find a fitting method there. Unfortunately the config property name is different in the ``Mail.MailConfig`` class and the ``XWiki.XWikiProperties``, so I am not sure if the SendMailConfigClassDocumentConfigurationSource helps. Well, it helps a bit by avoiding to look up the locations for the properties

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
ClemensRobbenhaar: yep, I wondered, too, if the other way round would be better. will change this.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
ClemensRobbenhaar: ``forWrite`` : do we want to read(false) or write(write) the property. Used e.g. to still allow to write the email via PUT (I can think of use cases for that). I first named it ``isWrite`` but was not sure if ``forWrite`` is better. Maybe use ``writeAccess`` instead?
the ``iff`` is a shortcut for ``if and only if`` ... ok, seems to be an unknown abbreviation.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
ClemensRobbenhaar: I have seen the email in plain when testing.
The corresponding issue would be XWIKI-8331 then.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
ClemensRobbenhaar: I thought I ran checkstyle, but now I see the class is in the exclude list; that is why I did not notice.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
ClemensRobbenhaar: Arf, it does not, the SendMailConfigClassDocumentConfigurationSource only looks up SendMailConfigClass properties, not GeneralMailConfigClass properties. 
This gets a bit out of hand. I better scrap mail handling for now and send a separate PR later.

ClemensRobbenhaar:
Help with the unit tests is very appreciated. As the tests passed without modification I was surprised and checked the test, but I could not figure out how they worked at a glance ...

ClemensRobbenhaar:
Now added unit test in ``src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java``

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.net.URI;
+import java.util.List;
+
+import javax.inject.Provider;
+
+@ComponentTest
+public class ModelFactoryTest
+{
+    protected XWikiContext xcontext = mock(XWikiContext.class);
```
vmassol: You should be able to write:

```
@Mock
private XWikiContext xcontext;
```

Why protected?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    @InjectComponentManager
+    protected ComponentManager componentManager;
```
vmassol: why not private?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @MockComponent
+    protected Provider<XWikiContext> xcontextProvider;
+
+    @MockComponent
+    SpaceReferenceResolver<String> resolver;
```
vmassol: missing private?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    @InjectMockComponents
+    ModelFactory modelFactory;
+    
+    @Test
+    @SuppressWarnings("deprecation")
+    void testToRestObject() throws Exception
```
vmassol: I'd suggest to remove the `test` prefix, according to https://dev.xwiki.org/xwiki/bin/view/Community/Testing/JavaUnitTesting/#HBestpractices

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+        Document doc = mock(Document.class);
+        SpaceReference spaceRef = new SpaceReference("Space", new WikiReference("wiki"));
+        when(doc.getPrefixedFullName()).thenReturn("wiki:Space.Page");
+        when(doc.getWiki()).thenReturn("wiki");
+        when(doc.getSpace()).thenReturn("Space");
+        when(doc.getName()).thenReturn("Page");
+        when(doc.getDocumentReference()).thenReturn(new DocumentReference("Page", spaceRef));
+        when(resolver.resolve(eq("Space"), any(String.class))).thenReturn(spaceRef);
+        
+        com.xpn.xwiki.web.Utils.setComponentManager(componentManager);
```
vmassol: Actually all this seems to indicate you should use `@OldcoreTest` instead of `@ComponentTest`.  That could allow you to reduce the setup code.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+        // that is as good as not showing the value
+        boolean stringFieldFound = false;
+        for (Property prop : properties) {
+            if ("String".equals(prop.getType())) {
+                assertEquals("abcd", prop.getValue(), "should see proper value for StringField");
+                stringFieldFound = true;
+            }
+            if ("Password".equals(prop.getType())) {
+                assertNull(prop.getValue(), "should see not value for PasswordField");
+            }
```
vmassol: So we want allow Password types but they must have null values? Or should there be no Password property? I'm asking since it's not clear from the test what is the contract we want.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+        assertTrue(stringFieldFound, "should have StringField in the result");
```
vmassol: Thanks @ClemensRobbenhaar for writing a test. Does it mean we don't have one that already calls REST and check the result?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
vmassol: From What I see at PageResourceTest, ideally we should add a new test method there like `testPUTGETWithPasswordXProperty` or modify the existing `testPUTGETWithObject` but instead of using the Tag XObject, use a XWikiUsers xobjects for example (which contains a password field).

You test verifies that we don't show the password which is cool but it doesn't test that we can set it. Doing it the way it's done in PageResourceTest would validate both aspects (PUT and GET).

Note to self: we need to convert PageResourceTest to a JUnit5 test.

@ClemensRobbenhaar Don't mistake me. I'm very happy that you wrote the fix + the test. For me that's good enough to close the issue.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
vmassol: From What I see at PageResourceTest, ideally we should add a new test method there like `testPUTGETWithPasswordXProperty` or modify the existing `testPUTGETWithObject` but instead of using the Tag XObject, use a XWikiUsers xobjects for example (which contains a password field).

You test verifies that we don't show the password which is cool but it doesn't test that we can set it. Doing it the way it's done in PageResourceTest would validate both aspects (PUT and GET).

Note to self: we need to convert PageResourceTest to a JUnit5 test.

@ClemensRobbenhaar Don't mistake me. I'm very happy that you wrote the fix + the test. For me that's good enough to close the issue.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
ClemensRobbenhaar: both copy & paste from the ``AbstractListClassPropertyValuesProviderTest``. fixing ...

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
ClemensRobbenhaar: ok

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
ClemensRobbenhaar: It checks that if we have a Password at all, then its value will not be exposed.
The current implementation just skips the ``<value>`` element from the output; i.e. REST Clients still can see that a user has a password, but nit which value it has.
This is not an intentional decision on my side, it was just the quickest way to fix things over the weekend.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
ClemensRobbenhaar: I agree that the test is not perfect and there is much to be improved. Especially I know the "set password" is not tested. Today I wondered if there is a use case for it; i.e. creating users with a password via REST. If the setter for the password accepts the unencrypted password and encrypts it, this might even be working, and it is probably not a security issue, as access rights are checked by the REST API. So maybe I broke something here instead if fixing it.

I did not notice that ``PageResourceTest`` would be the place to extend an existing test, or sure I would have been gone that way instead.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
ClemensRobbenhaar: ok, trying this ....

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/ModelFactoryTest.java
ClemensRobbenhaar: Ah, ``PageResourceTest`` is a HTTP-Test. I did not look there ... these integration tests are quite a different thing to write and run, at least for me.
The unit test has been improved according to your comments.
 
If password fields should still be writeable, I have to undo a change in the implementation (remove the check in ``ModelFactory#toObject``)

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1063,4 +1073,19 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
...
         return log;
     }
+
+    /**
+     * Check if the given property should be exposed via REST.
+     * @param restProperty the property to be read/written
+     * @param forWrite true if we want to write to the XObject property
+     * @return true if the property is considered public for read or write
+     */
+    private boolean isPublicType(Property restProperty, boolean forWrite)
```
tmortagne: I'm not a big fan of making impossible to access the whole model anymore, you might want to use REST to migrate data from one instance to another. At the very least user with PR should be allowed.

ClemensRobbenhaar: Yes, I understand. Actually to handle issue it might even be enough to only check for edit rights, or something like that. Update coming soon, I will go for Admin rights in the next attempt, and wait how the discussion sorts out.

ClemensRobbenhaar:
Ah, I thought any Admin would do. And yes, method name is misleading now.

vmassol:
@ClemensRobbenhaar Thanks, I'll apply this PR now. I've also created https://jira.xwiki.org/browse/XWIKI-16138 if you want to move the fix you had for that part too.

ClemensRobbenhaar:
Ok, will look into handle the other part in a better way without hard coding the page references in the REST code.

PR #1062, opened 2019-02-11

surli:
* Make sure that testss using MockitoOldCore loads a URLConfiguration
component

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/test/MockitoOldcore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/test/MockitoOldcore.java
@@ -66,6 +66,7 @@ import org.xwiki.security.authorization.ContextualAuthorizationManager;
 import org.xwiki.test.annotation.AllComponents;
 import org.xwiki.test.internal.MockConfigurationSource;
 import org.xwiki.test.mockito.MockitoComponentManager;
+import org.xwiki.url.URLConfiguration;
 import org.xwiki.wiki.descriptor.WikiDescriptorManager;
 
 import com.xpn.xwiki.CoreConfiguration;
@@ -379,6 +380,11 @@ public class MockitoOldcore
 
         if (!this.componentManager.hasComponent(LinkRefactoring.class)) {
             this.componentManager.registerMockComponent(LinkRefactoring.class);
+        }
+
+        // Make sure to a have an URLConfiguration component.
+        if (!this.componentManager.hasComponent(URLConfiguration.class)) {
+            this.componentManager.registerMockComponent(URLConfiguration.class);
         }
 
         // XWiki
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/test/MockitoOldcore.java
```diff
@@ -381,6 +382,11 @@ public void before(Class<?> testClass) throws Exception
             this.componentManager.registerMockComponent(LinkRefactoring.class);
         }
 
+        // Make sure to a have an URLConfiguration component.
+        if (!this.componentManager.hasComponent(URLConfiguration.class)) {
+            this.componentManager.registerMockComponent(URLConfiguration.class);
```
vmassol: I'm curious to know what's the benefit of this. Usually we register mocks when we can provide default behaviors or when the mock doesn't require any specific behavior and just needs to exist to make code work. Here, I believe it's not going to help.

What I would do if I were you is to add some mocking on URLConfiguration so that `getURLFormatId` returns the standard format.

WDYT?

surli: Here the problem is not related to getURLFormatId but with the new method I added in XWIKI-6073 to check if a the modification date of a resource should be checked on the disk. 
I spotted the bug in https://ci.xwiki.org/job/XWiki/job/xwiki-platform/job/master/1455/testReport/org.xwiki.activitystream/WebRssTest/Platform_Builds___quality___Build_for_Quality___webRssDisplay/: the produced HTML contains a macro error because it couldn't load the URLConfiguration component. So I imagine that we might obtain the same error in any XAR unit test. 

So for me it's quite enough to only load a mock, especially since the newly introduced method has a default behaviour in its interface.

vmassol: You don't understand. What we put in MockitoOldcore is to be generic as much as possible, i.e. to provide *default* behaviors so that tests don't need to mock *common* paths. Tests will need to register their own mocks if they want a different behavior. Right now what you've done is not enough to make it work fully. But adding the behavior is a simple one more line that adds the value. Otherwise you could simply add the mock to the test case.

surli: Hm indeed I don't understand.

> adding the behavior is a simple one more line that adds the value.

you were talking about `getURLFormatId` behaviour: I don't see the point on adding a behaviour on that one for now, since I don't have any issue with its behaviour. It's basically never used in my problematic tests.

I have behaviour issue with `useResourceLastModificationDate`, but it's only a boolean value, and now the component is registered I do get its value in my tests, because it relies on its default one defined in the interface.

So for me I'm completely align with what you say:
> to be generic as much as possible, i.e. to provide _default_ behaviors

vmassol: For a URLConfiguration mock to work it needs to provide a return value for calls to `getURLFormatId`. Just make it return `standard`.

vmassol:
@tmortagne why did you merge? I was on it (ok I didn't assign it to me), and if I didn't merge it's because I considered it wasn't good enough in the current state. Please see all my comments... If you don't agree with my comments then please reply ;)

vmassol:
My point is really simple though, it's a one liner and it seems so obvious to me that I don't see the problem at all.

tmortagne:
> My point is really simple though, it's a one liner and it seems so obvious to me that I don't see the problem at all.

If you are talking about `Just make it return `standard`.`, @surli did that so it seems it was ok.

vmassol:
Indeed I missed that, sorry, and thanks.

PR #1064, opened 2019-02-11

surli:
* Fix EditRightsPane to not rely anymore on a full resource URL

### DIFF
--- a/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
+++ b/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
@@ -20,10 +20,8 @@
 package org.xwiki.test.ui.po;
 
 import org.openqa.selenium.By;
-import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.FindBy;
-import org.openqa.selenium.support.ui.ExpectedCondition;
 
 /**
  * Class that holds the types of Rights and the types of States of a check-box
--- a/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
+++ b/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
@@ -20,8 +20,10 @@
 package org.xwiki.test.ui.po;
 
 import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.FindBy;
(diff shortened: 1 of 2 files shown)


surli:
@vmassol would you check this one? I'm not really fond of relying on the URL to check this, but not sure we have another choice there.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
```diff
@@ -71,7 +73,9 @@ State getNextState()
         static State getButtonState(WebElement button)
         {
             for (State s : values()) {
-                if ((button.getAttribute("src").endsWith(s.imageURL))) {
+                // starting with 11.1RC1 the resource URLs can now contain a query parameter to avoid cache issue
+                // so we cannot rely to a full URL anymore
+                if ((button.getAttribute("src").contains(s.imageURL))) {
```
vmassol: I would put a more generic message, that we use contains() because the URL may contain query string parameters (e.g.  starting with 11.1RC1 the resource URLs can now contain a query parameter to avoid cache issue) and we don't care about that to identify the state.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
```diff
@@ -127,8 +131,12 @@ public void clickGuestRight(Right right)
...
             // Selenium 2.0a7 it returns a *full* URL even though the DOM has a relative URL as in:
             // <img src="/xwiki/resources/js/xwiki/usersandgroups/img/allow.png">
-            getDriver().waitUntilElementEndsWithAttributeValue(buttonLocator, "src",
-                currentState.getNextState().imageURL);
+            getDriver().waitUntilCondition(webDriver -> {
+                // starting with 11.1RC1 the resource URLs can now contain a query parameter to avoid cache issue
+                // so we cannot rely to a full URL anymore
+                return webDriver.findElement(buttonLocator).getAttribute("src")
+                    .contains(currentState.getNextState().imageURL);
+            });
```
vmassol: Why not introduce a waitUntilElementContainsAttributeValue?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
surli: OK I'll improve it.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/EditRightsPane.java
surli: I hesitated, we have already lot of APIs I think: do you think it might be useful elsewhere too?

PR #1065, opened 2019-02-11
XWIKI-16118: Search facets confuse page name with page title

CamiAndrei:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3479,7 +3479,7 @@ platform.search.suggestSources=Sources
 platform.search.suggestSources.hint=Search suggest results are aggregated from multiple sources. The sources are grouped by the search engine they use. Each source is configured to match a specific thing (e.g. the page name). Only the sources that are active and that use the current search engine contribute results to the search suggest.
 platform.search.suggestAddNewSource=Add a new source
 platform.search.suggestNewSourceName=New Source
-platform.search.suggestSourceDocumentName=Page names
+platform.search.suggestSourceDocumentName=Page titles
```
surli: IMO it would require a new translation key:
  1. it would only change the wording in english
  2. it's ambiguous here since the key is about "DocumentName"

vmassol: A new translation key is cleaner but it has an important cost. It also means we loose all existing translation unless someone migrates all translations of the old key to the new key. So all in all, I think I'd keep the same key.

surli: > we loose all existing translation

actually apparently the translations need to be changed too I guess. To move from "name" to "title". So "loosing" them, would trigger the creation of new ones.

vmassol: It's very different:
* If you keep the key: you just need to change the EN version. All other versions are still ok. Not perfect but ok.
* If you change the key: 100% of translations (except EN) are wrong (not translated)!

Big difference :) Unless you do the copy I was mentioning.

surli: Since it's really a minor issue (just saw the ticket) I'd say we can keep it like this :)

vmassol: We used to have copy tools in the past AFAIR (@tmortagne / @atallahade  ?). Have we lost this when we moved to weblate? Would be useful for this case.

I don't see anything on https://dev.xwiki.org/xwiki/bin/view/Community/Weblate%20Configuration/

surli: There's a script to migrate keys if I remember well.

surli: https://github.com/xwiki/xwiki-dev-tools/blob/master/weblate-scripts/migrate_keys.py

vmassol: Cool would be nice if we could document it and how to use it.

atallahade: *Short explanation:*
You give the paths to the source and destination translation files (of the base translation, e.g. `ApplicationResources.properties` and not `ApplicationResources_fr.properties`) and the path of a text file with a list of keys (one key per line) and the script will copy the specified keys from source to destination (and that for every available languages).
*Example:*
```bash
$ migrate_keys.py \
xwiki-platform-core/xwiki-platform-annotations/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Translations.xml \
xwiki-platform-core/xwiki-platform-application/xwiki-platform-application-ui/src/main/resources/Applications/Translations.xml \
keys.txt
```
with keys.txt:
```
annotations.menu.loading
annotations.menu.loaderror
annotations.tab.info.noannotations
```

surli: Thanks for the explanations @atallahade :) 
@CamiAndrei would you improve your PR to rename the key? There is three actions to do:
  1. Rename the key `platform.search.suggestSourceDocumentName` to `platform.search.suggestSourceDocumentTitle` in all ApplicationResources*.properties of the same module. @atallahade suggests to use:
```
$ sed 's/platform.search.suggestSourceDocumentName/platform.search.suggestSourceDocumentTitle/' ApplicationResources*.properties
```

  2. Add the old key `platform.search.suggestSourceDocumentName` in the deprecated section of ApplicationResources.properties, in order to keep the backward compatibility (see: https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HTranslationPropertyDeprecations)

  3. Replace the old key where it's used. From what I see in xwiki-platform it's only used here: `xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfig.xml` 

If you don't feel to do it I'll handle the changes, tell me :)

CamiAndrei: @surli Thank you for your explanations. I will handle the changes.

surli:
Thanks @CamiAndrei :+1:

PR #1066, opened 2019-02-14
XWIKI-8346: Broken cross-wiki links on page rename

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-8346

## Solution

  * Check two schemes when executing load backlinks request: with and
without the wiki id.

## Limitation

This fix itself is not enough for fixing completely XWIKI-8346: the `updateLinksOnFarm` needs to be set to true in the move/rename request too. Which is by default not the case.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
@@ -2064,8 +2064,8 @@ public class XWikiHibernateStore extends XWikiHibernateBaseStore implements XWik
 
             // the select clause is compulsory to reach the fullName i.e. the page pointed
             Query query = session
-                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink"
-                    + " or backlink.id.link = :backlinkwithwiki");
+                .createQuery("select backlink.fullName from XWikiLink as backlink where "
+                    + "backlink.id.link in (:backlink, :backlinkwithwiki)");
             query.setString("backlink", this.localEntityReferenceSerializer.serialize(documentReference));
             query.setString("backlinkwithwiki", this.defaultEntityReferenceSerializer.serialize(documentReference));
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
@@ -2064,8 +2064,10 @@ public class XWikiHibernateStore extends XWikiHibernateBaseStore implements XWik
 
             // the select clause is compulsory to reach the fullName i.e. the page pointed
             Query query = session
-                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink");
+                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink"
+                    + " or backlink.id.link = :backlinkwithwiki");
             query.setString("backlink", this.localEntityReferenceSerializer.serialize(documentReference));
+            query.setString("backlinkwithwiki", this.defaultEntityReferenceSerializer.serialize(documentReference));
 
             @SuppressWarnings("unchecked")
             List<String> backlinkNames = query.list();
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
@@ -2063,44 +2063,34 @@ public class XWikiHibernateStore extends XWikiHibernateBaseStore implements XWik
             Session session = getSession(context);
 
             // the select clause is compulsory to reach the fullName i.e. the page pointed
-            // here we check both the backlink serialized with AND without the wiki part
-            // the reason is in case of a farm of wiki, if the link points to a page from another wiki
-            // its reference is saved with the wiki part, so using only the local serialization won't be enough in that
-            // case. This should be changed once the refactoring to support backlinks properly has been done.
-            // See: XWIKI-16192
             Query query = session
-                .createQuery("select backlink.fullName, backlink.link from XWikiLink as backlink where "
-                    + "backlink.id.link in (:backlink, :backlinkwithwiki)");
-            query.setString("backlink", this.localEntityReferenceSerializer.serialize(documentReference));
-            query.setString("backlinkwithwiki", this.defaultEntityReferenceSerializer.serialize(documentReference));
-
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -2064,8 +2064,10 @@ private void releaseAllLocksForCurrentUser(final XWikiContext ctx)
 
             // the select clause is compulsory to reach the fullName i.e. the page pointed
             Query query = session
-                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink");
+                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink"
```
tmortagne: I'm wondering if using `backlink.id.link in (:backlink, :backlinkwithwiki)` wouldn't be better for performances than two different conditions.

surli: I can change if you think it can be better. I don't know much in term of perf there.
Now you think this change is safe and won't break anything?

tmortagne: Looks good to me if you are sure the reference from others wiki end up in that table (I vaguely remember it was not the case at some point).

tmortagne:
Actually I don't really understand what is the point of this code.

Here is how the whole backlink system should be working IMO:
* when a page is saved we get all the links and saved **in linked target page wiki** a backlink entry (if that backlink is from another wiki then it's saved with the wiki part)
* when searching for the backlink of a page we you get all the backlink entries associated to that page in its own wiki (so with a local reference) and get get local and absolute references

surli:
the scenario is the following:
the backlink is from another wiki, so saved with the wiki part.
As you said, then when moving the page in its own wiki, currently it is checked only for the local reference, so without the wiki part: we cannot resolve the backlinks from the other wiki. So we cannot update it.

tmortagne:
When I read you code it feels like it was written expecting the backlink entry to be saved in the source wiki instead of the link target wiki. This need to be fixed otherwise you have to request all wikis when you search for backlinks which does not make much sense.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -2063,9 +2063,16 @@ private void releaseAllLocksForCurrentUser(final XWikiContext ctx)
...
             // the select clause is compulsory to reach the fullName i.e. the page pointed
+            // here we check both the backlink serialized with AND without the wiki part
+            // the reason is in case of a farm of wiki, if the link points to a page from another wiki
+            // its reference is saved with the wiki part, so using only the local serialization won't be enough in that
+            // case. This should be changed once the refactoring to support backlinks properly has been done.
+            // See: XWIKI-16192
             Query query = session
-                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink");
+                .createQuery("select backlink.fullName from XWikiLink as backlink where "
+                    + "backlink.id.link in (:backlink, :backlinkwithwiki)");
```
tmortagne: This logic is wrong. You will end up with links having nothing to do with the document you are interrested in (documents with the same fullname than the document in a different wiki).

Instead you should do a different search depending if you are requesting the document's wiki or not.

surli:
@tmortagne I fixed the bad behaviour you spotted by comparing the given wiki references. I added two unit tests for enforcing the new behaviour of loadBackLinks and I tested it by doing the following scenario:
  - 3 wikis: xwiki, subwiki and other
  - in subwiki, pages A and B, link A -> B
  - in other, pages A and B, link A -> B
  - in xwiki, page Foo, link Foo -> subwiki:B

I tried to rename subwiki:B, the links on Foo and subwiki:A have been changed, nothing has been modified in other wiki. 

I spotted a small issue in DefaultLinkRefactoring when it's about renaming a link from another wiki, so I fixed it too. 

Tell me if there's still some cases that I might have missed.

tmortagne:
You should check the wiki before, not after. It's not so great in term of performance to execute a request which return too many results and then ignore some of the result.

surli:
> You should check the wiki before

@tmortagne it's ready again. I did the same check as before. I added some more checks on the unit test too: it explains why I removed the endTransaction in the code: it's actually already called in the finally.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -2064,19 +2064,33 @@ private void releaseAllLocksForCurrentUser(final XWikiContext ctx)
 
             // the select clause is compulsory to reach the fullName i.e. the page pointed
             Query query = session
-                .createQuery("select backlink.fullName from XWikiLink as backlink where backlink.id.link = :backlink");
-            query.setString("backlink", this.localEntityReferenceSerializer.serialize(documentReference));
+                .createQuery("select backlink.fullName from XWikiLink as backlink where "
+                    + "backlink.id.link = :backlink");
+
+            WikiReference wikiReference = documentReference.getWikiReference();
+            boolean sameWikiContext = (wikiReference == null || wikiReference.equals(context.getWikiReference()));
```
tmortagne: wikiReference cannot be null here, it's mandatory in a DocumentReference.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -2064,19 +2064,33 @@ private void releaseAllLocksForCurrentUser(final XWikiContext ctx)
...
+
+            String serializedBacklink;
+
+            // if we are in the same wiki context, we should only get the local reference
+            // but if we are not, then we have to check the full reference, containing the wiki part since
+            // it's how the link are recorded.
+            // This should be changed once the refactoring to support backlinks properly has been done.
+            // See: XWIKI-16192
+            if (sameWikiContext) {
```
tmortagne: You are actually reimplementing the "compactwiki" serializer here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
surli: I'm always afraid with the possible NPE :) I'll remove it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
surli: ahah, right.

surli:
@tmortagne so it was a long way for something simple in the end...

PR #1067, opened 2019-02-15
XWIKI-11062: Make "Search Suggest" responsive on mobile

slauriere:
## Issue

* [XWIKI-11062](https://jira.xwiki.org/browse/XWIKI-11062)

## Solution

- Make the quick search input field responsive
- Make the quick search suggestions responsive
- For windows narrower than small tablets (768px): show the notification and avatar buttons only when the quick search widget is inactive

## Comments

- This highlights the possible need of an XWiki JavaScript API that checks the window current width range (small, medium, large, ...), considering that screens may not be measured uniformally, in case such an API does not exist already.
- There is still the need to add smoother show / hide transitions to the notification and avatar buttons, this will be the object of another issue if this pull request gets accepted.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/action-menus.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/action-menus.less
@@ -84,7 +84,10 @@
     border: 0;
     min-width: 0;
     transition: width 0.3s, padding 0.3s;
-    width: 200px; // we need to set a width in pixels otherwise the transition is not applied nicely
+    width: 115px; // we need to set a width in pixels otherwise the transition is not applied nicely
+    @media (min-width: @screen-sm-min) {
+      width: 200px;
+    }
   }
 
   // When the global search is close
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
@@ -118,12 +118,20 @@
     </property>
     <property>
       <code>require(['jquery', 'bootstrap'], function($) {
+
+  // Returns true if the window inner width is lower than 768.
+  function isSmallScreen() {
+    var windowWidth = window.innerWidth;
+    return windowWidth &lt; 768;
+  }
   $(document).ready(function() {
 
     // Some variables used in the next 2 functions
     var globalSearch = $('#globalsearch');
     var globalSearchInput = globalSearch.find('input');
     var globalSearchButton = globalSearch.find('button');
+    var notificationButton = $('#tmNotifications');
+    var avatar = $('li.navbar-avatar');
 
     // Open the global search when the user click on the global search button
     globalSearchButton.click(function(event) {
@@ -131,6 +139,11 @@
         return true;
       }
       globalSearch.removeClass('globalsearch-close');
+      // Hide the notification and avatar buttons on small screens
+      if (isSmallScreen()) {
+        notificationButton.addClass('hidden');
+        avatar.addClass('hidden');
+      }
       globalSearchInput.focus();
       return false;
     });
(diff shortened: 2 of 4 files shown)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
```diff
@@ -118,19 +118,32 @@
     </property>
     <property>
       <code>require(['jquery', 'bootstrap'], function($) {
+
+  // Returns true if the window inner width is lower than 768.
+  function isSmallScreen() {
+    var windowWidth = window.innerWidth;
+    return windowWidth &lt; 768;
```
evalica: 768 could have also be a parameter, since @screen-sm-min is configurable, and since it's used in at least 2 places.

evalica:
The GIF interaction looks nice.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
slauriere: Thanks for the review, @evalica. Do you mean that we could use @screen-sm-min directly from JavaScript or that we should introduce a variable in the JavaScript itself that we would initialize to 768, or something else?

surli: > Do you mean that we could use @screen-sm-min directly from JavaScript or that we should introduce a variable in the JavaScript itself that we would initialize to 768, or something else?

I guess that if it's configurable then you have a way to retrieve it with a velocity call, so you could put that value in the html as a data-attribute and reuse it in your javascript probably.

slauriere: Thank you Simon. As far as I can see, the screen dimension breakpoints are configurable indeed as variable declarations in the FlamingThemes directly. I could not find a way to retrieve their value from Velocity though. Hence I tried another approach in the commits below: add a CSS class on the parent `ul` element of the search input reflecting the search input state (open/close) and control the with of the input and the display of its adjacent buttons entirely in CSS using this new class. Also updated `suggestSearch.js` and `suggest.css`. The latter has no access to the screen dimension variables so I used a hardcoded width in that one, which is the case already for other CSS files such as `columns.css`, for which we could switch to LESS files I guess.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/action-menus.less
```diff
@@ -80,6 +80,37 @@
   }
 }
 
+ul {
+  @media (max-width: @screen-sm-max) {
+    // On mobile, do not show the avatar nor the notification buttons when the search input is open
```
surli, line 85: Maybe a dummy question, but in your comment you're talking about "not showing" and in the CSS you actually only move them: why not using a display none in that case? Or even a hidden class?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
```diff
@@ -118,19 +118,34 @@
     </property>
     <property>
       <code>require(['jquery', 'bootstrap'], function($) {
+
+  function updateParentListState(globalSearchElement, isSearchOpen) {
```
surli, line 122: Looks like you don't need the `isSearchOpen` parameter since there's a `globalsearch-close` CSS class already. Also would make sense to control there the add/remove of that class here too.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
```diff
@@ -118,19 +118,34 @@
...
+  }
+
   $(function() {
 
     // Some variables used in the next 2 functions
     var globalSearch = $('#globalsearch');
     var globalSearchInput = globalSearch.find('input');
     var globalSearchButton = globalSearch.find('button');
+    var notificationButton = $('#tmNotifications');
+    var avatar = $('li.navbar-avatar');
```
manuelleduc, line 140: Where are these variables used?

PR #1068, opened 2019-02-16
XWIKI-16138: Email addresses are shown in clear in REST results

ClemensRobbenhaar:
First version: 

- email addresses are only shown at all in REST results
   for users who have edit rights on the document (likely
   their own profile)
 - this only happens if email obfuscation is enabled
 - to check for this config settings a new MailGeneralConfiguration
   interface has been created, which for now exposes only this
   setting (the corresponding XClass has only this one)
   and also implements the fallback lookup on the
   XWiki.XWikiPreferences
 - that logic is then reused in the custom email displayer

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-api/src/main/java/org/xwiki/mail/MailGeneralConfiguration.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-api/src/main/java/org/xwiki/mail/MailGeneralConfiguration.java
@@ -1,0 +1,38 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.mail;
+
+import org.xwiki.component.annotation.Role;
+
+/**
+ * Represents all XWiki general configuration options for Mail.
+ *
+ * @version $Id$
+ * @since 11.1
+ * @since 10.11.4
+ */
+@Role
+public interface MailGeneralConfiguration
+{
+    /**
+     * @return true if user emails should only be shown obfuscated
+     */
+    boolean isObfuscateEmails();
+}
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/internal/configuration/GeneralMailConfigClassDocumentConfigurationSource.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/internal/configuration/GeneralMailConfigClassDocumentConfigurationSource.java
@@ -1,0 +1,126 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.mail.internal.configuration;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.configuration.ConfigurationSource;
+import org.xwiki.configuration.internal.AbstractDocumentConfigurationSource;
+import org.xwiki.mail.MailGeneralConfiguration;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.model.reference.LocalDocumentReference;
+import org.xwiki.model.reference.RegexEntityReference;
+import org.xwiki.observation.event.Event;
+
+import com.xpn.xwiki.internal.event.XObjectAddedEvent;
+import com.xpn.xwiki.internal.event.XObjectDeletedEvent;
+import com.xpn.xwiki.internal.event.XObjectUpdatedEvent;
+import com.xpn.xwiki.objects.BaseObjectReference;
+
+/**
+ * Provides configuration from the Mail.GeneralMailConfigClass document in the current wiki.
+ * If the Mail.GeneralMailConfigClass xobject exists in the Mail.MailConfig document
+ * then use configuration values from it and if it doesn't then look up similar values
+ * in the main preferences.
+ *
+ * @version $Id$
+ * @since 11.1
+ * @since 10.11.4
+ */
+@Component(roles = {MailGeneralConfiguration.class})
+@Singleton
+public class GeneralMailConfigClassDocumentConfigurationSource extends AbstractDocumentConfigurationSource
+    implements MailGeneralConfiguration
+{
+    private static final String MAIL_SPACE = "Mail";
+
+    private static final String PROPERTY_OBFUSCATE_NAME = "obfuscate";
+
+    /**
+     * The local reference of the Mail.GeneralMailConfigClass xclass.
+     */
+    private static final LocalDocumentReference CLASS_REFERENCE = new LocalDocumentReference(MAIL_SPACE,
+        "GeneralMailConfigClass");
+
+    /**
+     * The local reference of the Mail.MailConfig document.
+     */
+    private static final LocalDocumentReference DOC_REFERENCE = new LocalDocumentReference(MAIL_SPACE, "MailConfig");
+
+    @Inject
+    @Named("wiki")
+    private ConfigurationSource wikiConfig;
+
+    @Override
+    protected String getCacheId()
+    {
+        return "configuration.document.mail.general";
+    }
+
+    @Override
+    protected DocumentReference getDocumentReference()
+    {
+        return new DocumentReference(DOC_REFERENCE, getCurrentWikiReference());
+    }
+
+    @Override
+    protected LocalDocumentReference getClassReference()
+    {
+        return CLASS_REFERENCE;
+    }
+
+    @Override
+    protected <T> T getPropertyValue(String key, Class<T> valueClass)
+    {
+        T value = super.getPropertyValue(key, valueClass);
+        if (value == null && PROPERTY_OBFUSCATE_NAME.equals(key)) {
+            value = wikiConfig.getProperty("obfuscateEmailAddresses", valueClass);
+        }
+        return value;
+    }
+
+    @Override
+    public boolean isObfuscateEmails()
+    {
+        return getProperty(PROPERTY_OBFUSCATE_NAME, 0) != 0;
+    }
+
+    protected List<Event> getCacheCleanupEvents()
+    {
+        List<Event> events = new ArrayList<>();
+        events.addAll(super.getCacheCleanupEvents());
+
+        // cannot access this:
+        // events.addAll(wikiConfig.getCacheCleanupEvents());
+        // so instead:
+        RegexEntityReference classMatcher = BaseObjectReference.any("XWiki.XWikiPreferences");
+        events.addAll(Arrays.<Event>asList(new XObjectAddedEvent(classMatcher), new XObjectDeletedEvent(classMatcher),
+            new XObjectUpdatedEvent(classMatcher)));
+
+        return events;
+    }
+}
--- a/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/MailSenderScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/MailSenderScriptService.java
@@ -34,6 +34,7 @@ import javax.mail.internet.MimeMessage;
 import org.xwiki.component.annotation.Component;
 import org.xwiki.component.manager.ComponentLookupException;
 import org.xwiki.component.util.DefaultParameterizedType;
+import org.xwiki.mail.MailGeneralConfiguration;
 import org.xwiki.mail.MailListener;
 import org.xwiki.mail.MailSenderConfiguration;
 import org.xwiki.mail.MimeMessageFactory;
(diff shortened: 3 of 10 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -243,6 +249,7 @@ private void fillObjectSummary(ObjectSummary objectSummary, Document doc, BaseOb
                 objectSummary.setHeadline(serializePropertyValue(xwikiObject.get(propertyNames[0])));
```
ClemensRobbenhaar: here is the place where I miss to handle emails as I do not get at the type here. this is only a problem when the email is the first property of a class, but nevertheless ...

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
```diff
@@ -82,18 +85,31 @@ public PropertyClass getPropertyClass()
      */
     public java.lang.Object getValue()
     {
-        BaseProperty baseProperty = getBaseProperty();
+        final BaseProperty<?> baseProperty = getBaseProperty();
+        final XWikiContext context = getXWikiContext();
```
ClemensRobbenhaar, line 89: introduced local variable to escape checkstyle "line too long" warning.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
```diff
@@ -82,18 +85,31 @@ public PropertyClass getPropertyClass()
...
+        if (propertyClass != null && "Email".equals(propertyClass.getClassType())) {
+            if (isObfuscateEmails() && !context.getWiki().getRightService().hasWikiAdminRights(context)) {
                 return null;
             }
         }
 
         return getBaseProperty().getValue();
     }
+
+    private static boolean isObfuscateEmails()
```
ClemensRobbenhaar, line 110: Introduced helper method to avoid "too much complexity" code quality warning.

ClemensRobbenhaar:
@vmassol: sorry that it took so long for me to get back to it. First I thought something complicated was still open and then I forgto about it.

I tried to add a new test case to the ``com.xpn.xwiki.api.DocumentTest`` but that went nowhere. Maybe I should introduce an new ``PropertyTest`` in the same package (though this might test very little except for the stubbing).

Daddy2200 (approved):

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-api/src/main/java/org/xwiki/mail/MailGeneralConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Represents all XWiki general configuration options for Mail.
+ *
+ * @version $Id$
+ * @since 11.1
+ * @since 10.11.4
+ */
+@Role
+public interface MailGeneralConfiguration
```
vmassol, line 32: There's already a `GeneralMailConfiguration` role, see https://github.com/xwiki/xwiki-platform/blob/fb8e06afa1a7e46e33ce311eb4b09be81b0567f6/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-general/src/main/java/org/xwiki/mail/GeneralMailConfiguration.java

Also, the location in mail-send-api seems not the best as it's more generic than for the mail sender.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/internal/configuration/GeneralMailConfigClassDocumentConfigurationSource.java
```diff
@@ -0,0 +1,126 @@
...
+ * then use configuration values from it and if it doesn't then look up similar values
+ * in the main preferences.
+ *
+ * @version $Id$
+ * @since 11.1
+ * @since 10.11.4
+ */
+@Component(roles = {MailGeneralConfiguration.class})
+@Singleton
+public class GeneralMailConfigClassDocumentConfigurationSource extends AbstractDocumentConfigurationSource
```
vmassol, line 56: There's already a class named the same at at https://github.com/xwiki/xwiki-platform/blob/8fc42cf0d60aa6816771c5ee3786709191672c19/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-general/src/main/java/org/xwiki/mail/internal/configuration/GeneralMailConfigClassDocumentConfigurationSource.java#L39

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/MailSenderScriptService.java
```diff
@@ -274,6 +275,24 @@ private MailListener getListener(String hint) throws MessagingException
...
+     */
+    public String obfuscateForDisplay(final String emailAddress)
+    {
+        String obfuscatedEmailAddress;
+        if (emailAddress != null && this.generalConfiguration.isObfuscateEmails()) {
+            obfuscatedEmailAddress = emailAddress.replaceAll("^(.).*@", "$1...@");
+        } else {
+            obfuscatedEmailAddress = emailAddress;
+        }
+        return obfuscatedEmailAddress;
```
vmassol, line 293: We already have an email obfuscator at https://github.com/xwiki/xwiki-platform/blob/e500f31b9b22962519528b7932beef2cf7f23789/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-general/src/main/java/org/xwiki/mail/EmailAddressObfuscator.java#L33

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/MailSenderScriptService.java
```diff
@@ -282,6 +301,17 @@ public MailSenderConfiguration getConfiguration()
         return this.senderConfiguration;
     }
 
+    /**
+     * @return the general configuration about mails; i.e. everything that did not fit elswhere
+     * @since 11.1
+     * @since 10.11.4
+     */
+    public MailGeneralConfiguration getGeneralConfiguration()
```
vmassol, line 309: That doesn't look good to me. We're in the mail sender and there's no reason for it to know the general config.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/MailSenderScriptService.java
vmassol: Also it's already provided in https://github.com/xwiki/xwiki-platform/blob/fb8e06afa1a7e46e33ce311eb4b09be81b0567f6/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-general/src/main/java/org/xwiki/mail/script/GeneralMailScriptService.java#L39

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/script/MailSenderScriptService.java
vmassol: The place would be in https://github.com/xwiki/xwiki-platform/blob/fb8e06afa1a7e46e33ce311eb4b09be81b0567f6/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-general/src/main/java/org/xwiki/mail/script/GeneralMailScriptService.java#L39 but I need to check below why it would be needed.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/integration/ScriptingIntegrationTest.java
```diff
@@ -135,6 +137,15 @@ public void registerConfiguration() throws Exception
...
 
+        this.componentManager.registerComponent(MailGeneralConfiguration.class, new MailGeneralConfiguration()
+        {
+            @Override
+            public boolean isObfuscateEmails()
+            {
+                return true;
+            }
+        });
+
```
vmassol, line 148: It seems it would be better to use a mock for this.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/test/java/org/xwiki/mail/integration/ScriptingIntegrationTest.java
```diff
@@ -311,6 +322,14 @@ public void sendHTMLAndCalendarInvitationMail() throws Exception
         BodyPart calendarBodyPart = ((MimeMultipart) messages[0].getContent()).getBodyPart(1);
         assertEquals("text/calendar;method=CANCEL", calendarBodyPart.getHeader("Content-Type")[0]);
         InputStream is = (InputStream) calendarBodyPart.getContent();
-        assertEquals(calendarContent, IOUtils.toString(is));
+        assertEquals(calendarContent, IOUtils.toString(is, StandardCharsets.UTF_8));
+    }
+
+    @Test
+    public void emailsAreObfuscatedForDisplay()
```
vmassol, line 329: public keyword not needed

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
vmassol: Note that we usually don't use `final` keyword.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
```diff
@@ -82,18 +85,31 @@ public PropertyClass getPropertyClass()
...
         // API level, so that java code using core classes will still have access, regardless or rights.
         if (propertyClass != null && "Password".equals(propertyClass.getClassType())) {
-            if (!getXWikiContext().getWiki().getRightService().hasProgrammingRights(getXWikiContext())) {
+            if (!context.getWiki().getRightService().hasProgrammingRights(context)) {
+                return null;
+            }
+        }
+
+        if (propertyClass != null && "Email".equals(propertyClass.getClassType())) {
+            if (isObfuscateEmails() && !context.getWiki().getRightService().hasWikiAdminRights(context)) {
```
vmassol, line 102: why the admin exception?

vmassol: note that this change doesn't seem related to the `[XWIKI-16138: Email addresses are shown in clear in REST results` issue, is it?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -1081,14 +1088,20 @@ public JobLog toRestJobLog(Collection<LogEvent> logs, URI self)
...
-    private boolean hasAccess(Property restProperty)
+    private boolean hasAccess(Property restProperty, Document doc)
     {
         if (PASSWORD_TYPE.equals(restProperty.getType())) {
             return authorizationManagerProvider.get().hasAccess(Right.ADMIN, xcontextProvider.get().getWikiReference());
         }
 
+        if (EMAIL_TYPE.equals(restProperty.getType())) {
+            return !mailConfigProvider.get().isObfuscateEmails() ||
+                authorizationManagerProvider.get().hasAccess(Right.EDIT, doc.getDocumentReference());
```
vmassol, line 1102: missing some comment to explain it.

Does it mean the REST call fails in this case? It shouldn't. I think the idea is just to obfuscate the email.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/displayer_email.vm
```diff
@@ -4,15 +4,8 @@
...
-  #set ($generalMailConfigObject = $mailConfigDoc.getObject('Mail.GeneralMailConfigClass'))
-  #set ($obfuscate = $generalMailConfigObject.getValue('obfuscate'))
-  #if ("$!obfuscate" == '')
-    ## We handle backward compatibility by also looking in the XWikiPreferences xobject
-    #set ($obfuscate = $xwiki.getXWikiPreferenceAsInt('obfuscateEmailAddresses', 0))
-  #end
-  #if ($obfuscate == 1)
-$!escapetool.xml($value.replaceAll('^(.).*@', '$1...@'))##
+  #if ($services.mailsender.generalConfiguration.obfuscateEmails)
+$!escapetool.xml($services.mailsender.obfuscateForDisplay($value))##
```
vmassol, line 8: Already done in https://github.com/xwiki/xwiki-platform/blob/3ac126fc2f6993b2f848a488a6cf62484d3e06c4/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-general/src/main/resources/templates/displayer_email.vm#L27-L28

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
ClemensRobbenhaar: this change is not about the REST API, but you asked for it to be included in the JIRA ticket.
at leas that is what I understood from https://jira.xwiki.org/browse/XWIKI-16138?focusedCommentId=101848&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-101848

the admin exception is because "admins should be able to do everything".

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -243,6 +249,7 @@ private void fillObjectSummary(ObjectSummary objectSummary, Document doc, BaseOb
                 objectSummary.setHeadline(serializePropertyValue(xwikiObject.get(propertyNames[0])));
             } catch (XWikiException e) {
                 // Should never happen
+                logger.error("Unexpected exception when accessing property [{}]", propertyNames[0], e);
```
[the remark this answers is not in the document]
ClemensRobbenhaar: I never have seen this happening, but I wanted to log anything in case there is an error instead if just swallowing the exception. Then, level "warn" might be more appropriate in this case.

ClemensRobbenhaar:
> I think several portions of this PR were done in [e500f31](https://github.com/xwiki/xwiki-platform/commit/e500f31b9b22962519528b7932beef2cf7f23789)

me thinks so, too. This pull request is simply quite outdated by other developments.
I tried to rebase it on the current head, but that made all changes done on the master in the mean time to be applied in this branch instead of the other way around. I must have misunderstood the `git help rebase` explanations.

So instead of pushing an enormous bulk of redundant stuff to this branch I better open a new PR which takes your comments into account and works with the currently available API instead of creating a new ones.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
vmassol: I raised the fact that we also need to fix it but I didn't say to do it in one issue. You cannot include something that's not related to the issue title ;) Or you need to create different jiras and do a single commit but for several jira issues.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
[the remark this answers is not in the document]
vmassol: I think error is better since it's not supposed to happen and thus if it does it would be quite unexpected and probably something to look at. My point was more that you left the comment untouched but added the logging and the 2 don't make sense. You need find what could happen in term of error and then remove "should never happen" and instead explain when it could happen for ex. Or mention that it shouldn't happen but that you don't know if it could happen and to be extra safe we're logging it as a way to know about it about analyze it to decide what to do in the future.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
vmassol: > the admin exception is because "admins should be able to do everything".

Do we do that elsewhere (show emails in clear for admins)? I just want to make sure we're consistent in the behavior.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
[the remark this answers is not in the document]
ClemensRobbenhaar: ok, then I better skip that change, as I have no idea how the above lines could throw an exception,

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Property.java
ClemensRobbenhaar: Actually we do not do that. Even Admins see the E-mail obfuscated (but of course can go to the object editor for the user to see it there. I guess I should limit the "show emails unobfuscated" to the REST interface, and there only users with programming rights. As requested this will be done in a separate issue.

PR #1071, opened 2019-02-22
bootstrap activity stream icons are not properly vertical aligned

Pbast:


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
```diff
@@ -579,6 +579,10 @@ require(['jquery', 'XWikiNotificationsMacro'], function ($, XWikiNotificationsMa
   width: 2em;
 }
 
+.notification-icon > .img-thumbnail > .fa {
```
evalica: This solution is not generic enough if the users use other Icon Themes, not just the Font Awesome.

evalica:
An image on how it looks with the solution would have been nice :) Thanks Pbast for looking into this issue.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
Pbast: With Icon Themes, img are already well aligned because in style.css:
```
img {
    vertical-align: middle;
}
```

Pbast:
> An image on how it looks with the solution would have been nice :) Thanks Pbast for looking into this issue.

I paste an image on https://jira.xwiki.org/browse/XWIKI-16180

surli:
Thanks for your contribution @Pbast 

> I paste an image on jira.xwiki.org/browse/XWIKI-16180

I'm sorry but I personally don't see the difference in the before/after images. Could you use the browser inspector for the screenshot so we can see the margins info?

Pbast:
> Thanks for your contribution @Pbast
> 
> > I paste an image on jira.xwiki.org/browse/XWIKI-16180
> 
> I'm sorry but I personally don't see the difference in the before/after images. Could you use the browser inspector for the screenshot so we can see the margins info?

here:  https://jira.xwiki.org/browse/XWIKI-16180#comment-102336

Pbast:
Personnaly I prefer this one:

> the selector would be
> `.notification-icon .img-thumbnail span`

because `img-thumbnail span` could be conflict by a custom CSS (especially on welcome page)

surli:
@evalica @Pbast I'm reviewing this old PR and I'm not sure about its status: apparently @evalica requested for changes back then but then added a :+1: on your comment @Pbast. Does it mean the changes are not needed or you agreed with the proposed solution but they were never done? 
So my main question is: is there some more work needed and do you plan to do it if so?

Pbast:
> So my main question is: is there some more work needed and do you plan to do it if so?

@surli oh sorry about the delay ( because I'm working on other project in my job)
Anyway I fixed this issue like @evalica suggest and tested on XWiki 10.11.3

Pbast:
I fixed this issue, could you merge this PR? Thxs

PR #1073, opened 2019-03-02
LICENSE

yashrajbharti:
Corrected some punctuations and typos

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-                  GNU LESSER GENERAL PUBLIC LICENSE
+GNU LESSER GENERAL PUBLIC LICENSE
                        Version 2.1, February 1999
 
  Copyright (C) 1991, 1999 Free Software Foundation, Inc.
@@ -73,7 +73,7 @@ libraries into non-free programs.
   When a program is linked with a library, whether statically or using
 a shared library, the combination of the two is legally speaking a
 combined work, a derivative of the original library.  The ordinary
-General Public License therefore permits such linking only if the
+General Public License, therefore, permits such linking only if the
 entire combination fits its criteria of freedom.  The Lesser General
 Public License permits more lax criteria for linking other code with
 the library.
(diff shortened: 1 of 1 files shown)


vmassol:
Thanks. However I'm not sure we are at liberty to change the license as this is the official LGPL 2.1 license that we just copied.

vmassol:
AFAIK it's coming from https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt

vmassol:
> AFAIK it's coming from https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt

Apparently it's a different one.

yashrajbharti:
Words like "programs enables" are Grammatically wrong and must be changed to "programs enable" ... And I found some comma separations too

vmassol:
Sure but I prefer to stick to the original official translation even if it has mistakes. Changing it is dangerous, it can also change meanings and it's not our text.

Any strong rationale for changing it except fixing some typos?

vmassol:
Now if you can find some literature explaining that it's ok to reword it, please provide the links and we can take a decision.

LICENSE
```diff
@@ -95,9 +95,9 @@ case, there is little to gain by limiting the free library to free
 software only, so we use the Lesser General Public License.
 
   In other cases, permission to use a particular library in non-free
-programs enables a greater number of people to use a large body of
+programs enable a greater number of people to use a large body of
```
sdumitriu: This is correct as originally written, since it's not the programs that enable something, but the **permission** to use the programs.

LICENSE
```diff
@@ -106,7 +106,7 @@ users' freedom, it does ensure that the user of a program that is
 linked with the Library has the freedom and the wherewithal to run
 that program using a modified version of the Library.
 
-  The precise terms and conditions for copying, distribution and
+  The precise terms and conditions for copying, distribution, and
```
sdumitriu: The usage of the sequence comma is not mandatory, so unless there's an ambiguity (and in this case there isn't), both ways are valid.

LICENSE
```diff
@@ -176,7 +176,7 @@ above, provided that you also meet all of these conditions:
     table of data to be supplied by an application program that uses
     the facility, other than as an argument passed when the facility
     is invoked, then you must make a good faith effort to ensure that,
-    in the event an application does not supply such function or
+    in the event, an application does not supply such function or
```
sdumitriu: This is not correct, "in the event an application does not..." is one sentence, "an application does not" is the event, not a follow-up thing to do in another event.

sdumitriu:
Some of the suggestions are not valid, and the rest are purely stylistic. However, this is a standard license text, not written or owned by us, so any change proposals must be submitted to the proper creator of the license, the Free Software Foundation.

PR #1075, opened 2019-03-08
 XWIKI-6665: Saving edits after lock expiration doesn't warn the user that the page may be stale

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-6665

## Design of the solution

The proposed solution is to record in a field of the editor the last version of the working document and to send it to the server when saving.
The server will compare this version number against the latest version of the document: if the latest version is different than the one sent on the request it means someone certainly save the document in the meantime. We also compare the edition dates: in some edge cases we could have the same revision numbers that are not the same revisions (e.g. a deleted and recreated document).
We then compute a content and an object diff to remove the false positive: the doc could have been saved in the meantime because of an attachment added.
Finally in case of conflict we send a 409 status with a JSON containing both the previous and the latest version number.
In case of success we send a JSON the latest version number, so that we can update it.

## Changes

  * XWikiDocument.java: add some utility method to be able to compute the next version, without updating it. Handle of errors in getContentDiff.
  * XWikiAction.java: add an utility method to send a JSON answer.
  * SaveAction.java: add the new method to compute the conflict and change the answer in case of success.
  * SaveAndContinueAction.java: manage the errors in case of conflict.
  * xwiki-meta.js: Allow to bump and set a new version. Fire an event when it's the case. New attribution: isNew to know if a page is already created or not.
  * dataeditor.js: Bump the version when an object is added or removed, since it saves the document. 
  * actionButton.js: 
    * Get the current document version and listen on the new event to update it when it's needed.
    * Create a new form input with this version, the date of the edition and the edit status to provide it when saving.
    * Handle the 409 status answer by displaying a popup with the conflict message and by disabling the form/editor.
    * Save and view now also send an AJAX request to save, instead of submitting the form like it did.
    * Handle the various modes "Save and view / Save and continue / preview". 

## Screenshot

warning_conflict

## Tests

Some selenium tests have been executed to ensure the changes: namely EditWikiTest, InlineEditTest, ObjectEditTest, CompareVersionsTest, LanguageTest.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -1946,16 +1946,33 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
         return this.isContentDirty;
     }
 
+    /**
+     * Increment the current document version.
+     * This method will use {@link #getNextVersion(Version, boolean)} to compute the new version.
+     */
     public void incrementVersion()
     {
-        if (this.version == null) {
-            this.version = new Version("1.1");
+        this.version = getNextVersion(this.version, isMinorEdit());
+    }
+
+    /**
+     * This method computes the next version and returns it, but won't change the current version.
+     * In order to change the current version, see {@link #incrementVersion()}.
+     *
+     * @param version the based version from which to compute the next one.
+     * @param minorEdit true means it's a minor edition.
+     * @return the new version computed based on the current one.
+     * @since 11.2RC1
+     */
+    public static Version getNextVersion(Version version, boolean minorEdit)
+    {
+        if (version == null) {
+            return new Version("1.1");
+        }
+        if (minorEdit) {
+            return version.next();
         } else {
-            if (isMinorEdit()) {
-                this.version = this.version.next();
-            } else {
-                this.version = this.version.getBranchPoint().next().newBranch(1);
-            }
+            return version.getBranchPoint().next().newBranch(1);
         }
     }
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -6323,6 +6323,16 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
     {
         XWikiDocument fromDoc = context.getWiki().getDocument(this, fromRev, context);
         XWikiDocument toDoc = context.getWiki().getDocument(this, toRev, context);
+        if (fromDoc == null) {
+            throw new XWikiException(XWikiException.MODULE_XWIKI_DIFF, XWikiException.ERROR_XWIKI_DIFF_CONTENT_ERROR,
+                String.format("The revision [%s] cannot be found in [%s] for making diff.", fromRev,
+                    this.getDocumentReference()));
+        }
+        if (toRev == null) {
+            throw new XWikiException(XWikiException.MODULE_XWIKI_DIFF, XWikiException.ERROR_XWIKI_DIFF_CONTENT_ERROR,
+                String.format("The revision [%s] cannot be found in [%s] for making diff.", toRev,
+                    this.getDocumentReference()));
+        }
         return getContentDiff(fromDoc, toDoc, context);
     }
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
@@ -19,14 +19,16 @@
  */
 package com.xpn.xwiki.web;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 
 import javax.script.ScriptContext;
-import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.lang3.StringUtils;
+import org.suigeneris.jrcs.rcs.Version;
 import org.xwiki.job.Job;
 import org.xwiki.localization.LocaleUtils;
 import org.xwiki.model.EntityType;
@@ -41,6 +43,8 @@ import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.doc.XWikiDocument;
 import com.xpn.xwiki.doc.XWikiLock;
+
+import net.sf.json.JSONObject;
 
 /**
  * Action used for saving and proceeding to view the saved page.
@@ -185,6 +189,25 @@ public class SaveAction extends PreviewAction
             tdoc.removeXObjects(REDIRECT_CLASS);
         }
 
+        // We only proceed on the check between versions in case of AJAX request, so we currently stay in the edit form
+        // This can be improved later by displaying a nice UI with some merge options in a sync request.
+        // For now we don't want our user to loose their changes.
+        if (Utils.isAjaxRequest(context) && request.getParameter("previousVersion") != null) {
+            // TODO The check of the previousVersion should be done at a lower level or with a semaphore since
+            // another job might have saved a different version of the document
+            Version previousVersion = new Version(request.getParameter("previousVersion"));
+
+            // we ensure that nobody edited the document between the moment the user started to edit and now
+            if (!doc.getRCSVersion().equals(previousVersion)) {
+                JSONObject jsonObject = new JSONObject();
+                jsonObject.element("previousVersion", previousVersion.toString());
+                jsonObject.element("latestVersion", doc.getRCSVersion().toString());
+                this.answerJSON(context, HttpStatus.SC_CONFLICT, jsonObject);
+                return true;
+            }
+        }
+
+
         // We get the comment to be used from the document
         // It was read using readFromForm
         xwiki.saveDocument(tdoc, tdoc.getComment(), tdoc.isMinorEdit(), context);
@@ -221,6 +244,20 @@ public class SaveAction extends PreviewAction
         return false;
     }
 
+    private void answerJSON(XWikiContext context, int status, JSONObject json) throws XWikiException
+    {
+        try {
+            String jsonAnswerAsString = json.toString();
+            context.getResponse().setContentType("application/json");
+            context.getResponse().setContentLength(jsonAnswerAsString.length());
+            context.getResponse().setStatus(status);
+            context.getResponse().setCharacterEncoding(context.getWiki().getEncoding());
+            context.getResponse().getWriter().print(jsonAnswerAsString);
+        } catch (IOException e) {
+            throw new XWikiException("Error while sending JSON answer.", e);
+        }
+    }
+
     @Override
     public boolean action(XWikiContext context) throws XWikiException
     {
@@ -232,9 +269,14 @@ public class SaveAction extends PreviewAction
         if (save(context)) {
             return true;
         }
+
         // forward to view
         if (Utils.isAjaxRequest(context)) {
-            context.getResponse().setStatus(HttpServletResponse.SC_NO_CONTENT);
+            EditForm form = (EditForm) context.getForm();
+            JSONObject jsonAnswer = new JSONObject();
+            jsonAnswer.element("newVersion",
+                XWikiDocument.getNextVersion(context.getDoc().getRCSVersion(), form.isMinorEdit()).toString());
+            answerJSON(context, HttpStatus.SC_OK, jsonAnswer);
         } else {
             sendRedirect(context.getResponse(), Utils.getRedirect("view", context));
         }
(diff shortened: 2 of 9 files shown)


surli:
@mflorea I'm not asking you to review `actionButton.js` since the diff is pretty nasty: there were no indent after the first `{` and my editor added them automatically when I was working on it.
Anyway, I mostly want your opinion first on what I did on `xwiki-meta.js` and how it's used in `dataeditor.js`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1941,16 +1941,33 @@ public boolean isContentDirty()
...
+     */
     public void incrementVersion()
     {
-        if (this.version == null) {
-            this.version = new Version("1.1");
+        this.version = getNextVersion(this.version, isMinorEdit());
+    }
+
+    /**
+     * This method computes the next version and returns it, but won't change the current version.
```
tmortagne: Not sure all this warning is required for a `static` method starting with `get` and taking as parameter the version to increment but OK :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -85,6 +85,8 @@
 import com.xpn.xwiki.objects.BaseObject;
 import com.xpn.xwiki.plugin.fileupload.FileUploadPlugin;
 
+import net.sf.json.JSONObject;
```
tmortagne: You might want to avoid using this very old abandoned library and use Jackson instead for example.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
```diff
@@ -221,6 +241,80 @@ public boolean save(XWikiContext context) throws XWikiException
...
+     * @param context the current context of the request.
+     * @return true in case of conflict. If it's true, the answer is immediately sent to the client.
+     */
+    private boolean isConflictingWithVersion(XWikiContext context, XWikiDocument newDoc) throws XWikiException
+    {
+        XWikiRequest request = context.getRequest();
+        XWikiDocument doc = context.getDoc();
+
+        // the document is new we don't have to check the version date or anything
+        if ("true".equals(request.getParameter("isNew")) && doc.isNew()) {
```
tmortagne: Be careful the context document is usually the default locale and not the translation you might be editing currently.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -944,4 +946,25 @@ private boolean redirectSpaceURLs(String action, XWikiURLFactory urlf, XWiki xwi
...
     }
+
+    /**
+     * Answer to a request with a JSON content.
+     * @param context the current context of the request.
+     * @param status the status code to send back.
+     * @param json the JSON to serialize in the content of the answer.
+     * @throws XWikiException in case of error during the serialization of the JSON.
+     */
+    protected void answerJSON(XWikiContext context, int status, JSONObject json) throws XWikiException
```
tmortagne: I'm never a big fan of exposing publicly 3rd party classes (XWikiAction is public).

You could simply take an `Object` (or a Map since it seems to be your only use case right now) and use Jackson to serialize it.

vmassol: I haven't followed at all the PR but note that anything added to XWikiAction is potentially a technical debt since the new API is the Resource Handler and thus any feature added to XWiki Action also need to be provided to this new API (otherwise it'll make the removal of XWikiAction harder, hence the technical debt).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
surli: OK I'll change that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
surli: Hmm right, I should use the document I have in argument.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
surli: > You could simply take an Object (or a Map since it seems to be your only use case right now) and use Jackson to serialize it.

I will use an Object in the signature.

> any feature added to XWiki Action also need to be provided to this new API 

Good to know. I actually added this method in XWikiAction to avoid duplicating it in SaveAction / SaveAndContinueAction. I can add it too in AbstractResourceReferenceHandler, but for now it won't be used and I'm not sure it will ever be, depending how we move those actions...

vmassol: > Good to know. I actually added this method in XWikiAction to avoid duplicating it in SaveAction / SaveAndContinueAction. I can add it too in AbstractResourceReferenceHandler, but for now it won't be used and I'm not sure it will ever be, depending how we move those actions...

Yes we'll need to convert these actions but we don't know when we'll do it. If the new code can be re-used as is then it'll make it easier when we do the move.

surli:
@tmortagne I made some changes to now use Jackson lib for JSON. I also had to change the way I sent back the version of the doc, by saving it (the version) in the context. Might be good that you check that.

@gdelhumeau I'm not sure Marius will have time to review my javascript code. Would you have some time to check the changes bring on meta.js?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
```diff
@@ -54,8 +61,15 @@
     /** The identifier of the save action. */
     public static final String ACTION_NAME = "save";
 
+    /**
+     * The key to retrieve the saved object version from the context.
+     */
+    public static final String SAVED_OBJECT_VERSION_KEY = "SaveAction.savedObjectVersion";
```
tmortagne: Not super fan of this hack being an API (which this new static is). At the very least it should stay private for now (in which case it's a little better because we can refactor it easily later).

surli: It cannot be private since I need it in SaveActionAndContinue.

tmortagne: You don't really need that field, you need each code to use the same key name. I really prefer duplicate over new API for this one. You could also move it in some internal class.

surli: OK, I'll duplicate it then.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -23,359 +23,518 @@
...
-      if (this.form.xcontinue && this.form.xcontinue.value) {
-        url = this.form.xcontinue.value;
+      // update the version
+      require(['xwiki-meta'], function (xm) {
+        xm.setVersion(response.responseJSON.newVersion);
+      });
+
+      // Announce that the document has been saved
+      // TODO: We should send the new version as a memo field
+      document.fire("xwiki:document:saved");
```
gdelhumeau: It means you might fire the event *before* the new version is set in xwiki-meta.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -23,359 +23,518 @@
...
-      'action_propupdate' : "$services.localization.render('core.shortcuts.edit.saveandview')",
-      'action_saveandcontinue' : "$services.localization.render('core.shortcuts.edit.saveandcontinue')"
+  var actionButtons = XWiki.actionButtons = XWiki.actionButtons || {};
+
+  // we need to handle the creation of document
+  var currentDocument, currentVersion, isNew;
+  var editingVersionDate = new Date();
+
+  require(['xwiki-meta'], function (xm) {
+    currentDocument = xm.documentReference;
```
gdelhumeau: I wonder if you should not wrap the whole file into this require() call. On the other hand, I doubt nobody will try to save a page before xwiki-meta is loaded.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
gdelhumeau: So if someone listen to "xwiki:document:saved", she might have a wrong `xm.version`.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -23,359 +23,518 @@
...
+      document.fire("xwiki:document:saved");
+    },
+    // 409 happens when the document is in conflict: i.e. someone's else edited it at the same time
+    on409 : function(response) {
+      this.progressBox.hide();
+      this.savingBox.hide();
+      this.savedBox.hide();
+      this.failedBox.hide();
+
+      var displayConflictModal = function (jsonAnswer) {
```
gdelhumeau: You could also define this function as a method of the class, to be consistent. But I would not complain if you let it like this (it's more javascript-ish).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/meta.js
```diff
@@ -30,21 +30,51 @@ define(['jquery'], function($) {
...
       'version':           html.data('xwiki-version'),
       'restURL':           html.data('xwiki-rest-url'),
       'form_token':        html.data('xwiki-form-token'),
       // Since 10.4RC1
-      'userReference':     XWiki.Model.resolve(html.data('xwiki-user-reference'), XWiki.EntityType.DOCUMENT)
+      'userReference':     XWiki.Model.resolve(html.data('xwiki-user-reference'), XWiki.EntityType.DOCUMENT),
+      // Since 11.2RC1
+      'isNew': html.data('xwiki-isnew'),
+      'bumpVersion': function (isMinor) {
+        var version = html.data('xwiki-version');
```
gdelhumeau: What happen if you save twice?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/meta.js
```diff
@@ -30,21 +30,51 @@ define(['jquery'], function($) {
...
+          }
+          version = parts[0] + "." + parts[1];
+        }
+        html.data("xwiki-version", version);
+        document.fire('xwiki:document:changeVersion', {'version': version, 'documentReference': documentReference});
+      },
+      'setVersion': function (version) {
+        html.data("xwiki-version", version);
+        document.fire('xwiki:document:changeVersion', {'version': version, 'documentReference': documentReference});
+      }
```
gdelhumeau: Looks good. 

Since we do not really return a "map" kind of objects anymore, I think I would have created a "class" instead, so I could re-use the other methods (like `setVersion()` in `bumpVersion()`). With this solution, you would not need to update the DOM like you do. Instead you would change a field of the object and that's it, since nobody is supposed to rely on the DOM anymore.

Example of javascript "class" without using Prototype: https://github.com/xwiki/xwiki-platform/blob/87bce44d2b261389b565ece64ba1fbffbba01218/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminThemesSheet.xml#L151

Again, it does not mean what you did is wrong, but since you're asking for my feedbacks :)

surli: Thanks for the feedback! I wasn't actually quite happy of my solution and I thought about refactoring it as a real JS object. So ok I'll take time to do it.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.js
```diff
@@ -133,6 +133,9 @@ editors.XDataEditors = Class.create({
               },
               onComplete : function() {
                 item.disabled = false;
+                require(['xwiki-meta'], function (xm) {
```
gdelhumeau: Could we wrap the whole ajax request into the require() call, so we are sure that everything is synchronized?

surli: I wasn't quite sure but I can give it a try.

surli:
@gdelhumeau I don't ignore your feedback but I'll merge the PR right now so I can have enough feedback before the release through the CI. I plan to continue working on it until it's released.

PR #1076, opened 2019-03-11
XWIKI-16054: Additional margin for document title in Inline mode

karannaoh:


karannaoh:
inline
here it is, @surli

karannaoh:
inline
Here it is, @surli

PR #1077, opened 2019-03-13
XWIKI-16220: AWM applications located in extensions are not protected against move/rename anymore

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16220

## Proposal
  * Check if the AWM page is part of an extension: if it's the case, do not hijack the rename button.


surli:
I'm not quite happy with the result (double warning), but I don't really know how to do it: I'm afraid that the users won't see it if I added it the content to the existing warning...

PR #1092, opened 2019-03-28
XWIKI-15450: Browser cache for attachments is never invalidated (needs browser clear cache)

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-15450

## Proposal

Fallback on `createAttachmentRevisionURL` when creating an attachment URL for download. This is an easy solution which use the already existing mechanism to add the revision number at the end of the URL. It should both prevents the cache mechanism in case of a new version has been updated, and use the cache mechanism when downloading several times the same attachment.

## Tests

I only ran oldcore unit tests for it, but perform manual validation by running an instance and downloading/uploading attachments.

## Discussion

@Enygma2002 suggested that we could also use the browser cache mechanism with the right header, but I find it more complicated, especially since apparently the example shown by @mflorea on it use a ResourceHandler that we currently only have for webjars. AFAIU it would imply to create the appropriate resource handler for attachment to use it here.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
@@ -613,17 +613,14 @@ public class XWikiServletURLFactory extends XWikiDefaultURLFactory
             }
         }
 
-        if (context != null) {
-            try {
-                XWikiAttachment attachment = findAttachmentForDoc(context.getDoc(), filename, context);
-                if (attachment != null) {
-                    return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
-                        querystring, xwikidb, context);
-                }
-            } catch (XWikiException e) {
-                if (LOGGER.isErrorEnabled()) {
-                    LOGGER.error("Exception while trying to get latest attachment version !", e);
-                }
+        XWikiAttachment attachment = context.getDoc().getAttachment(filename);
+        if (attachment != null) {
+            return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                querystring, xwikidb, context);
+        } else {
+            if (LOGGER.isErrorEnabled()) {
+                LOGGER.error(String.format("Exception while trying to get attachment [%s] from [%s.%s]!",
+                    filename, spaces, name));
             }
         }
 
@@ -803,18 +800,6 @@ public class XWikiServletURLFactory extends XWikiDefaultURLFactory
     {
         XWikiAttachment attachment = null;
         XWikiDocument rdoc = context.getWiki().getDocument(doc, docRevision, context);
-        if (filename != null) {
-            attachment = rdoc.getAttachment(filename);
-        }
-
-        return attachment;
-    }
-
-    private XWikiAttachment findAttachmentForDoc(XWikiDocument doc, String filename, XWikiContext context)
-        throws XWikiException
-    {
-        XWikiAttachment attachment = null;
-        XWikiDocument rdoc = context.getWiki().getDocument(doc, context);
         if (filename != null) {
             attachment = rdoc.getAttachment(filename);
         }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
@@ -613,6 +613,7 @@ public class XWikiServletURLFactory extends XWikiDefaultURLFactory
             }
         }
 
+        // In the context of the current doc, no need to reload it.
         if (isContextDoc(xwikidb, spaces, name, context)) {
             XWikiAttachment attachment = context.getDoc().getAttachment(filename);
             if (attachment != null) {
@@ -621,6 +622,24 @@ public class XWikiServletURLFactory extends XWikiDefaultURLFactory
             } else {
                 LOGGER.error("Exception while trying to get attachment [{}] from space [{}] and page [{}]!", filename,
                     spaces, name);
+            }
+        } else {
+            String originalWikiId = context.getWikiId();
+            context.setWikiId(xwikidb);
+
+            DocumentReference documentReference = new DocumentReference(xwikidb, spaces, name);
+            try {
+                XWikiDocument document = context.getWiki().getDocument(documentReference, context);
+                XWikiAttachment attachment = document.getAttachment(filename);
+                if (attachment != null) {
+                    return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                        querystring, xwikidb, context);
+                }
+            } catch (XWikiException e) {
+                LOGGER.error("Exception while trying to get out of context attachment [{}] from space [{}] and page "
+                        + "[{}]!", filename, spaces, name, e);
+            } finally {
+                context.setWikiId(originalWikiId);
             }
         }
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
@@ -619,10 +619,8 @@ public class XWikiServletURLFactory extends XWikiDefaultURLFactory
                 return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
                     querystring, xwikidb, context);
             } else {
-                if (LOGGER.isErrorEnabled()) {
-                    LOGGER.error(String.format("Exception while trying to get attachment [%s] from [%s.%s]!",
-                        filename, spaces, name));
-                }
+                LOGGER.error("Exception while trying to get attachment [{}] from space [{}] and page [{}]!", filename,
+                    spaces, name);
             }
         }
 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -613,6 +613,39 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
...
 
+        if (isContextDoc(xwikidb, spaces, name, context)) {
+            XWikiAttachment attachment = context.getDoc().getAttachment(filename);
+            if (attachment != null) {
+                return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                    querystring, xwikidb, context);
+            } else {
+                if (LOGGER.isErrorEnabled()) {
+                    LOGGER.error(String.format("Exception while trying to get attachment [%s] from [%s.%s]!",
+                        filename, spaces, name));
```
vmassol: Best practice: always use SLF4J syntax :)

```
LOGGER.error("Exception while trying to get attachment [{}] from space [{}], page [{}]!", filename, spaces, name));
```

2nd best practice: always separate each variable with `[]`. Imagine case 1 (spaces ="Hello.World" and name = "") and case 2 (spaces="" and name "Hello.World"), you wouldn't be able to differentiate them when debugging a production instance.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -613,6 +613,39 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
             }
         }
 
+        if (isContextDoc(xwikidb, spaces, name, context)) {
+            XWikiAttachment attachment = context.getDoc().getAttachment(filename);
+            if (attachment != null) {
+                return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                    querystring, xwikidb, context);
+            } else {
+                if (LOGGER.isErrorEnabled()) {
```
vmassol: You shouldn't use this with SLF4J when there's only a single statement.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
vmassol: Best practices at https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
surli: Good to know I actually replicate this from https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java#L610 so I guess I can refactor this one too.

vmassol: It's the same cost as the "IF" with SLF4J (that's why the api has positional parameters btw ;)). So it just makes the code harder to read than needed.

Ofc if you have to do some costly computations to perform the logging then yes it can be useful.

tmortagne:
As far as I understand a version will be added to the attachment URL only if that target attachment belong to the context document. Did I missed something ?

surli:
> As far as I understand a version will be added to the attachment URL only if that target attachment belong to the context document. Did I missed something ?

No that's right. Do you think it's a problem?

tmortagne:
Well it's definitely better than nothing but using an attachment located in another page is not exactly a edge case. For example the logo is usually an attachment.

surli:
> Well it's definitely better than nothing but using an attachment located in another page is not exactly a edge case. For example the logo is usually an attachment.

So you think I should always retrieve the attachment version even if it's not in the context of the current doc? What about the perf then?

tmortagne:
I did not said I have the answer right now, just pointing out the limitation of this implementation :)

surli:
> I did not said I have the answer right now, just pointing out the limitation of this implementation :)

yeah and thanks for that :) my question was more to get your opinion: I think I can implement it to always get the attachment, but I'm really wondering about the impact it could have on perf.

Anyway, I will try to implement it so I can have feedback on it: we could always rollback if needed.

tmortagne:
I don't have much other idea than always getting the target XWikiDocument instance which may or may not already be in the cache so indeed it's a potential performance risk.

surli:
So I made the change to also handle the case where we're not the attachment is not in the current document context.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -607,12 +607,60 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
...
+            XWikiAttachment attachment = context.getDoc().getAttachment(filename);
+            if (attachment != null) {
+                return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                    querystring, xwikidb, context);
+            } else {
+                LOGGER.error("Exception while trying to get attachment [{}] from space [{}] and page [{}]!", filename,
+                    spaces, name);
+            }
+        } else {
+            String originalWikiId = context.getWikiId();
```
tmortagne: Better play with the wiki reference instead of the id since that's what is stored in the XWikiContext.

surli: Fixed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -607,12 +608,60 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
                         querystring, xwikidb, context);
                 }
             } catch (XWikiException e) {
-                if (LOGGER.isErrorEnabled()) {
-                    LOGGER.error("Exception while trying to get attachment version !", e);
+                LOGGER.error("Exception while trying to get attachment version !", e);
```
vmassol: extra space before the "!". ;) I also suggest adding more information in the log such as the attachment reference.

vmassol: Actually this is not from you... :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -607,12 +608,60 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
...
+            WikiReference originalWikiReference = context.getWikiReference();
+            context.setWikiId(xwikidb);
+
+            DocumentReference documentReference = new DocumentReference(xwikidb, spaces, name);
+            try {
+                XWikiDocument document = context.getWiki().getDocument(documentReference, context);
+                XWikiAttachment attachment = document.getAttachment(filename);
+                if (attachment != null) {
+                    return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                        querystring, xwikidb, context);
```
vmassol: There seems to be substantial duplication between the 2 conditions. Would be better to introduce a new private method.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -607,12 +608,60 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
...
+            if (attachment != null) {
+                return createAttachmentRevisionURL(filename, spaces, name, attachment.getVersion(),
+                    querystring, xwikidb, context);
+            } else {
+                LOGGER.error("Exception while trying to get attachment [{}] from space [{}] and page [{}]!", filename,
+                    spaces, name);
+            }
+        } else {
+            WikiReference originalWikiReference = context.getWikiReference();
+            context.setWikiId(xwikidb);
```
vmassol: This looks dangerous. I don't see where you put back the previous wiki id? i.e. I don't see the try/finally.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
```diff
@@ -607,12 +608,60 @@ public URL createAttachmentURL(String filename, String spaces, String name, Stri
                         querystring, xwikidb, context);
                 }
             } catch (XWikiException e) {
-                if (LOGGER.isErrorEnabled()) {
-                    LOGGER.error("Exception while trying to get attachment version !", e);
+                LOGGER.error("Exception while trying to get attachment version !", e);
+            }
+        }
+
+        // In the context of the current doc, no need to reload it.
```
vmassol: Why do you need to reload anything? Maybe this needs to be explained in the source.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
surli: There's a finally line 640. But yeah it might miss some comments.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletURLFactory.java
tmortagne: Seems to me "load" would make more sense than "reload".

PR #1095, opened 2019-04-02
XWIKI-14166: Inline edit not saving documents with default language blank and existing translation in the configured default language

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-14166

## Changes
  * Add a new test to show the problem
  * Edit PanelSheet to use tdoc instead of doc: the object are shared
but only the  translation is invalidated in the cache when the
translation is edited, so we need to retrieve their values from the
translation to avoid display inconsistencies.

## Discussion

The root problem of XWIKI-14166 is that objects are shared among documents and their translations, and our cache mechanism does not handle the changes on the objects properly. 
IMO we should invalidate the document and all their translations if the original document or one of its translations has a change on one of its objects. Else we could have inconsistencies such as the one described in XWIKI-14166.

So right now my fix only handle changes made on panels, but any other object can have the same issue if its sheet used doc instead of tdoc. One thing I could do at least might be to also change ObjectSheet to show that we use tdoc in it. 

Now as I said maybe the right fix to do is to handle it in the cache itself. WDYT?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/PanelSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/PanelSheet.xml
@@ -39,14 +39,14 @@
   <content>{{velocity output="false"}}
 #macro(displayPanelProperty $obj $propName)
   ; &lt;label#if ($xcontext.action == 'edit' || $xcontext.action == 'inline') for="${class.getName()}_${obj.number}_${propName}"#end&gt;$services.localization.render("${class.getName()}_${propName}")&lt;/label&gt;
-  : $tdoc.display($propName, $obj)
+  : $doc.display($propName, $obj)
 #end
 
-#set ($paneldoc = $tdoc)
+#set ($paneldoc = $doc)
 {{/velocity}}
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/PanelSheet.xml
```diff
@@ -39,14 +39,14 @@
...
+  : $tdoc.display($propName, $obj)
 #end
 
-#set ($paneldoc = $doc)
+#set ($paneldoc = $tdoc)
 {{/velocity}}
 
 {{velocity}}
-#set ($obj = $doc.getObject('Panels.PanelClass'))
+#set ($obj = $tdoc.getObject('Panels.PanelClass'))
```
tmortagne: You sure about that ? Translation documents are not supposed to contain xobjects.

surli: it's actually what makes it working :) we are always using the tdoc in the SaveAction. So if you are using an edit with `language=something` in a multilingual wiki, the translation will contain an xobject.

PR #1096, opened 2019-04-03
XWIKI-14165: Unable to edit the original document if it has default language blank and en (configured default translation) translation exists

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-14165

## Changes

  * Add a check in XWiki#preparedDocument to use doc instead of tdoc if
an empty language parameter is used.
  * Refactor LanguageTest#translateDocument to check with one original
document and 2 translations.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -2921,7 +2921,7 @@ public class XWiki implements EventListener
         // in use.
         try {
             String language = Util.normalizeLanguage(context.getRequest().getParameter("language"));
-            if (StringUtils.isNotEmpty(language)) {
+            if (language != null) {
                 if ("default".equals(language)) {
                     // forgetting language cookie
                     Cookie cookie = new Cookie("language", "");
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -2921,7 +2921,7 @@ public Locale getLocalePreference(XWikiContext context)
         // in use.
         try {
             String language = Util.normalizeLanguage(context.getRequest().getParameter("language"));
-            if (StringUtils.isNotEmpty(language)) {
+            if (language != null) {
```
tmortagne: I don't think this change is needed anymore.

surli: I don't think it fixes anything, but semantically it looks more correct since we might have an empty language parameter which is correct.
Basically it might avoid the first reaction you had when seeing that :) WDYT?

tmortagne: My point was more that it was not really linked to this issue but sure.

surli: Right. So if you're ok to merge with it I'll keep the change.

tmortagne: As you prefer.

PR #1101, opened 2019-04-08
XWIKI-16307: Unable to view the original document in multilingual wiki

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16307

## Solution 

 * Use the current document if the given locale is ROOT
 * Improve integration tests to check this behaviour

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -1240,12 +1240,21 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
         // rights
         Object currrentSdoc = context.get("sdoc");
         try {
-            // If we execute a translation use translated document as secure document
-            XWikiDocument tdoc = getTranslatedDocument(context);
+            XWikiDocument sdoc;
+            boolean translate;
 
-            context.put("sdoc", tdoc);
+            if (this.locale != Locale.ROOT) {
+                // If the locale is not specified, get the translated document from the context.
+                sdoc = getTranslatedDocument(context);
+                translate = true;
+            } else {
+                // If the locale is ROOT then we want the current document.
+                sdoc = this;
+                translate = false;
+            }
+            context.put("sdoc", sdoc);
 
-            return display(targetSyntax, false, transformationContextIsolated, false, true);
+            return display(targetSyntax, false, transformationContextIsolated, false, translate);
         } finally {
             context.put("sdoc", currrentSdoc);
         }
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1240,12 +1240,21 @@ public String getRenderedContent(Syntax targetSyntax, boolean transformationCont
         // rights
         Object currrentSdoc = context.get("sdoc");
         try {
-            // If we execute a translation use translated document as secure document
-            XWikiDocument tdoc = getTranslatedDocument(context);
+            XWikiDocument sdoc;
+            boolean translate;
 
-            context.put("sdoc", tdoc);
+            if (this.locale != Locale.ROOT) {
```
tmortagne: The locale is always set in a XWikiDocument coming from the DB.

I really don't think there is any criteria you can check for that `if`, the only way is fix this issue IMO is a new method which explicitly take the locale as parameter when you don't want to use the context one.

surli: > The locale is always set in a XWikiDocument coming from the DB.

So it can also be set to ROOT right? 
This change is working on my integration test with a doc with 2 translations.

tmortagne: > So it can also be set to ROOT right?

My point is that it's most use cases it's set to ROOT since we generally use this method from the ROOT XWikiDocument instance.

surli:
ping @tmortagne for me it's ok.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1233,21 +1233,88 @@ public String getRenderedContent(boolean transformationContextIsolated, XWikiCon
...
+     * @param targetSyntax the syntax to use to render the document
+     * @param transformationContextIsolated see {@link DocumentDisplayerParameters#isTransformationContextIsolated()}
+     * @param context the XWiki Context object
+     * @param retrieveTranslation if true retrieve the translation of the document according to the preferences (see
+     *  {@link #getTranslatedDocument(XWikiContext)}). If false, render the current document.
+     * @return the rendered content of the document or its translation.
+     * @throws XWikiException in case of error during the rendering.
+     * @since 11.3RC1
+     */
+    public String getRenderedContent(Syntax targetSyntax, boolean transformationContextIsolated, XWikiContext context,
```
tmortagne: I would keep this new method private as internal helper since that's not the one you are using from what I understand. Otherwise it's not clear why we need to introduce displayDocument since you could use this one.

surli: OK, makes sense.

surli: Done.

PR #1102, opened 2019-04-08
XWIKI-13011: "link":"editor" is not working in Livetable

oanalavinia:
**Issue**
https://jira.xwiki.org/browse/XWIKI-13011

**Proposal**
Display requested field in a modal and update the new value using the existing LiveTableResults.

This is how it looks like:
propEditor

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResults.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResults.xml
@@ -38,25 +38,23 @@
   <hidden>true</hidden>
   <content>{{include reference="XWiki.LiveTableResultsMacros" /}}
 
-{{velocity wiki="false"}}
-## the $extra variable is passed to the call to #gridresultwithfilter below and is used to perform some default
-## filtering on the LiveTable data even when the user has not performed any column filtering yet.
-## When users start using column filtering, the generated DB query will combine both the default filtering and what the
-## user has entered in the column filters.
+{{velocity}}
+## Used in editProperty for getting the field when "link": "editor" property is used.
 #if($request.value)
   #set($itemdoc = $xwiki.getDocument($request.page))
   #set($discard = $itemdoc.use($request.classname))
   #set($discard = $itemdoc.set($request.fieldname, $request.value))
-  #set($ok = $itemdoc.save("Field ${request.fieldname} updated"))
-  #if($ok)
-    SAVED
-  #end
+  #set($discard = $itemdoc.save("Field ${request.fieldname} updated"))
 #elseif($request.fieldname)
   #set($itemdoc = $xwiki.getDocument($request.page))
   #set($discard = $itemdoc.use($request.classname))
   #set($obj = $itemdoc.getObject($request.classname))
   $!itemdoc.display($request.fieldname, "edit", $obj)
 #else
+## the $extra variable is passed to the call to #gridresultwithfilter below and is used to perform some default
+## filtering on the LiveTable data even when the user has not performed any column filtering yet.
+## When users start using column filtering, the generated DB query will combine both the default filtering and what the
+## user has entered in the column filters.
   #set ($extra = '')
   #set ($params = [])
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResults.xml
```diff
@@ -43,24 +43,39 @@
...
-#if ("$!request.space" != '')
-  #set ($extra = "${extra} AND doc.space = ?")
-  #set ($discard = $params.add($request.space))
+#if($request.value)
+  #set($itemdoc = $xwiki.getDocument($request.page))
+  #set($discard = $itemdoc.use($request.classname))
+  #set($discard = $itemdoc.set($request.fieldname, $request.value))
+  #set($ok = $itemdoc.save("Field ${request.fieldname} updated"))
+  #if($ok)
+    SAVED
```
surli: Is it expected, or was it for debugging purpose?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -1239,3 +1239,54 @@ function init() {
...
+          $('#propEditorForm').html(e.responseText);
+        })
+        $('#propEditorSave').one('click', function() {
+          let value = getValue($('#propEditorForm'), fieldname);
+          let surl = doc.getURL('view', 'xpage=plain&page=' + fullname + '&classname=' + classname + '&fieldname='
+            + fieldname + '&value=' + value);
+          $.ajax({
+            url : surl,
+            complete : function(e) {
+              if(e.responseText.search('SAVED') != -1) {
```
surli: I don't know much this part of the code but I find it a bit weird to rely on such a string to assess that an AJAX request was performed correctly. Now maybe there's no other easy solution here to check it?

oanalavinia: Indeed, there could be used "success" and "error" in the ajax request, instead of  "complete".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -1239,3 +1239,53 @@ function init() {
...
+          let value = getValue($('#propEditorForm'), fieldname);
+          let surl = doc.getURL('view', 'xpage=plain&page=' + fullname + '&classname=' + classname + '&fieldname='
+            + fieldname + '&value=' + value);
+          $.ajax({
+            url : surl,
+            success : function() {
+              cb(value);
+            },
+            error : function() {
+              alert("$services.localization.render('platform.livetable.propEditor.failed')");
```
surli: @evalica would know better about the UX but I don't think we use the alerts to display errors. We generally prefer to use the XWiki Notification widget: https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/XWikiNotifications/

evalica:
Some 'design' remarks: 
* Ideally the title should be small. The current title looks more like a Hint, than a Title; Better would be "Modify field" + additional hint if needed (or a label for the input). 
* Button's order is not correct. The primary action is placed at the most right, see https://getbootstrap.com/docs/3.4/javascript/#modals
* The styling for the 'Cancel' button is not correct + there is a small difference in alignment

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
evalica: Ideally we should use the Notification widget. Check also for dependency. 
In case Notification widget will be used, provide error and success ("Field updated", "Field saved", etc.) messages.

evalica:
Also the dependency for https://extensions.xwiki.org/xwiki/bin/view/Extension/LiveTable%20Property%20Editor should be correctly marked in the .pom, in case it is supported.

PR #1104, opened 2019-04-10
XWIKI-16314: Force users to respect the DB limit when creating a page

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16314

## Solution

  * Add new store methods to check the limit size of a property
  * Add new cache store methods to cache the property limits. 
  * Add new checks in CreateAction and in save to check the length of a
page path
  * Edit createinline.vm to handle new exception in case of path too
long
  * Add new unit test to check new behaviours
  * Add new integration test to ensure UI changes

## Tests

Tested with the 4 database supported in docker test:
  * mysql
  * postgresql
  * HSQLDB
  * Oracle DB

## Limitations

HSQLDB showed that we might have exceptions to retrieve the limit based on the DB type, so I don't really know if this code would work on Oracle DB or other DB that are not supported in Docker tests.

### UI in case of path too long

Capture d’écran_2019-04-10_12-09-51

### Test with a REST request

Capture d’écran_2019-04-10_13-40-34

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
@@ -3158,4 +3158,14 @@ public class Document extends Api
     {
         return 1 == this.getDoc().getTranslation();
     }
+
+    /**
+     * @return the maximum authorized length for a document full name (see {@link #getFullName()}).
+     * @since 11.3RC1
+     */
+    @Unstable
+    public int getNameMaxLength()
+    {
+        return this.doc.getNameMaxLength();
+    }
 }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
@@ -3166,6 +3166,6 @@ public class Document extends Api
     @Unstable
     public int getNameMaxLength()
     {
-        return this.doc.getNameMaxLength();
+        return this.doc.getFullNameMaxLength();
     }
 }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
@@ -3161,7 +3161,7 @@ public class Document extends Api
 
     /**
      * @return the maximum authorized length for a document full name (see {@link #getFullName()}).
-     * @since 11.3RC1
+     * @since 11.4RC1
      */
     @Unstable
     public int getLocalReferenceMaxLength()
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
@@ -3164,8 +3164,8 @@ public class Document extends Api
      * @since 11.3RC1
      */
     @Unstable
-    public int getNameMaxLength()
+    public int getLocalReferenceMaxLength()
     {
-        return this.doc.getFullNameMaxLength();
+        return this.doc.getLocalReferenceMaxLength();
     }
 }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -9297,7 +9297,7 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
      * @since 11.3RC1
      */
     @Unstable
-    public int getNameMaxLength()
+    public int getFullNameMaxLength()
     {
         return getStore().getLimitSize(this.getXWikiContext(), this.getClass(), "fullName");
     }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -9297,7 +9297,7 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
      * @since 11.3RC1
      */
     @Unstable
-    public int getFullNameMaxLength()
+    public int getLocalReferenceMaxLength()
     {
         return getStore().getLimitSize(this.getXWikiContext(), this.getClass(), "fullName");
     }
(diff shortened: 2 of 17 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -3255,4 +3256,14 @@ private XWikiAttachmentStoreInterface getXWikiAttachmentStoreInterface(XWikiAtta
 
         return this.attachmentContentStore;
     }
+
+    @Override
+    public int getLimitSize(Class<?> entityType, String propertyName)
+    {
+        PersistentClass persistentClass = getConfiguration().getClassMapping(entityType.getName());
```
tmortagne: Not that this is not really the database but what's in the hbm files.

surli:
@tmortagne fixed

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -9237,4 +9237,17 @@ private XWikiAttachmentStoreInterface getXWikiAttachmentStoreInterface(String st
...
+
+    /**
+     * Compute and return the maximum authorized length for the full name (i.e. the serialized reference of the
+     * document) based on the current store limitation.
+     *
+     * @return the maximum authorized length for a document full name.
+     * @since 11.3RC1
+     */
+    @Unstable
+    public int getNameMaxLength()
```
tmortagne: I don't think this is the right method name since what it return is the **full name** max length, the document **name** is only the last part of it.

surli: `getFullnameMaxLength` or do you have another suggestion?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
```diff
@@ -3158,4 +3158,14 @@ public boolean isTranslation()
...
     }
+
+    /**
+     * @return the maximum authorized length for a document full name (see {@link #getFullName()}).
+     * @since 11.3RC1
+     */
+    @Unstable
+    public int getNameMaxLength()
+    {
+        return this.doc.getFullNameMaxLength();
```
vmassol: This is a weird API. I would never have expected to see this in the Document or XWikiDocument classes....

It seems to be it's related to the store.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -9237,4 +9237,17 @@ private XWikiAttachmentStoreInterface getXWikiAttachmentStoreInterface(String st
...
+     * Compute and return the maximum authorized length for the full name (i.e. the serialized reference of the
+     * document) based on the current store limitation.
+     *
+     * @return the maximum authorized length for a document full name.
+     * @since 11.3RC1
+     */
+    @Unstable
+    public int getFullNameMaxLength()
+    {
+        return getStore().getLimitSize(this.getXWikiContext(), this.getClass(), "fullName");
```
vmassol: This actually proves the API is not in the right place and it should be in the store instead.

vmassol: There's a name in API design about this but it escapes me FTM (when you have an api that actually delegates to some other class). It's an antipattern from what I remember.

vmassol: It's also making our Model API fatter which is the opposite to what we want.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
```diff
@@ -3158,4 +3158,14 @@ public boolean isTranslation()
     {
         return 1 == this.getDoc().getTranslation();
     }
+
+    /**
+     * @return the maximum authorized length for a document full name (see {@link #getFullName()}).
+     * @since 11.3RC1
+     */
+    @Unstable
+    public int getNameMaxLength()
```
tmortagne: You forgot to change that method name.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -9237,4 +9237,17 @@ private XWikiAttachmentStoreInterface getXWikiAttachmentStoreInterface(String st
...
+
+    /**
+     * Compute and return the maximum authorized length for the full name (i.e. the serialized reference of the
+     * document) based on the current store limitation.
+     *
+     * @return the maximum authorized length for a document full name.
+     * @since 11.3RC1
+     */
+    @Unstable
+    public int getFullNameMaxLength()
```
tmortagne: As I said in my previous comment I think `getLocalReferenceMaxLength` would be better since getFullName() is deprecated.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/createinline.vm
```diff
@@ -67,6 +67,17 @@
       )
   </div>
 #end
+## check if the exception is about the document having a path too long
+#if("$!exception" != '' && $exception.code == 11017)
```
vmassol: Another solution would be to let the query execute and we capture the JDBC exception if there's an error instead. And possibly translate it into a user friendly message. This would capture not only size issues but also any other problem, no?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
surli: > At an doc API level I'm not expecting to have any constraint on the doc full name.

I'm not sure to understand your POV there: there's a technical constraint on the name size which was until now hidden, but it has always been there. Here I'm just exposing it to the user so they can use it to create some specific checks.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
surli: Hmm indeed sorry, I saw the comment before you edited it :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/createinline.vm
surli: > This would capture not only size issues but also any other problem, no?

yes it would, but the problem here is about "And possibly translate it into a user friendly message.": it's not necessarily always easy. It's actually easier to check the problem before it happens. 

Besides it's safer to check the problem before the exception, especially when several documents are involved, like in a move of a hierarchy. So this PR is a first step for adding those checks.

PR #1106, opened 2019-04-21
Feature xwiki 13368 - Line numbers in Code macro

ttzn:
JIRA issue [here](https://jira.xwiki.org/browse/XWIKI-13368).

Basically adds a Layout parameter to the Code macro, with a LINE_NUM option to display line numbers next to the highlighted code.

To implement this, I used a simple approach of putting the line numbers and code in two blocks side by side, using flexbox styling (very similar to the `linenos = table` option in Pygments). The other solution was to wrap each line of code into its own block and flank it with a line number, which would have maybe given us more control, but :

- it requires changes deeper in the stack, injecting the layout parameter all the way down to the `PygmentsParser` class where we would need to define a different `PygmentsListener` that splits the incoming content by line before rendering blocks ;
- numbering lines using CSS counters and `:before` pseudo-elements has proven difficult, in particular, achieving right-alignment is not straightforward (see discussion in JIRA).

screenshot_macrocode


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/code/layout/CodeLayoutHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.rendering.block.Block;
+
+/**
+ * Arranges a list of code rendering blocks into a specific layout.
+ * 
+ * @version $Id$
+ * @since 11.3
```
surli: You have to change this: the changes for 11.3 are fixed now that 11.3RC1 has been released. So your change might be only applied to 11.4RC1 now.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroLayout.java
```diff
@@ -0,0 +1,74 @@
...
+        return hint;
+    }
+
+    /**
+     * Convenience class holding hint values for {@link CodeLayoutHandler} components. Links the outer enum class to the
+     * components classes as it is not possible to reference enums in annotations (and avoids using plain strings).
+     * 
+     * @version $Id$
+     */
+    public static class Constants
```
surli: I don't understand why you introduce a new inner class for that, it looks a bit overly complicated just for 2 strings...

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/code/layout/CodeLayoutHandler.java
surli: Also consider using the `@Unstable` annotation since it's a new API

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroLayout.java
ttzn: A third class is [necessary](https://stackoverflow.com/questions/13253624/how-to-supply-enum-value-to-an-annotation-from-a-constant-in-java) to be able to reference the constants from the enum definition (try removing it). If your question is more "why I used constants", well, in general, I avoid copy-pasting String literals, especially when contributing to a project I'm not familiar with.

If you're sure we're better off without that piece of code I'll remove it.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/code/layout/CodeLayoutHandler.java
ttzn: Will do ASAP.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroLayout.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.xwiki.rendering.code.layout.CodeLayoutHandler;
+
+/**
+ * Values allowed in the <code>layout</code> parameter of the code macro.
+ * 
+ * @version $Id$
+ * @since 11.3
+ */
+public enum CodeMacroLayout
```
tmortagne: I guess you used an enum as an easy way to have all the possible layouts listed in the WYSIWYG but then it kind of defeat the purpose of using a public component since it makes impossible to provide more layouts without modifying this enum.

IMO either you make CodeLayoutHandler internal (what I would personally choose as a first version since I doubt adding more layouts than the 2 you implemented is that much of a need) or you move to String.

An alternative which is a bit more work is to introduce a custom type for this property (for example "CodeMacroLayout" but not an enum this time and just some String container) and provide a custom `org.xwiki.displayer.HTMLDisplayer` for this type which would allow you to control exactly what you want to have in he form.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroParameters.java
```diff
@@ -51,4 +56,21 @@ public String getLanguage()
     {
         return this.language;
     }
+
+    /**
+     * @param layout the layout to apply to the code.
```
tmortagne: This will need a `@since`.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroParameters.java
```diff
@@ -51,4 +56,21 @@ public String getLanguage()
...
+     */
+    @PropertyDescription("the layout format (plain or with line numbers)")
+    public void setLayout(CodeMacroLayout layout)
+    {
+        this.layout = layout;
+    }
+
+    /**
+     * @return the layout to apply to the code.
```
tmortagne: This will need a `@since`.

tmortagne:
Would be better to also provide the translation key for the new parameter next to https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties#L1829

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroLayout.java
ttzn: Good point. How would I go about making CodeLayoutHandler internal ? Remove it from DI entirely ?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroLayout.java
surli: > If your question is more "why I used constants"

no it wasn't, you perfectly right to use constants :) 

My bad, indeed in that case it's difficult to do differently. Now I'm not even sure this enum is needed: you could have two static Strings in CodeMacroParameters and rely only on them... It's really a style matter I guess.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/macro/code/CodeMacroLayout.java
tmortagne: Just mean moving that class in an `*.internal.*` package (like the implementations).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-code/src/main/java/org/xwiki/rendering/internal/code/layout/CodeLayoutHandler.java
```diff
@@ -17,20 +17,22 @@
...
 
 /**
  * Arranges a list of code rendering blocks into a specific layout.
  * 
  * @version $Id$
- * @since 11.3
+ * @since 11.4RC1
  */
 @Role
+@Unstable
```
tmortagne: This annotation is only for public APIs (internal stuff can be broken any time without warning by definition)

tmortagne:
The logic looks OK to me but would be better to have someone else validate the actual CSS, @evalica WDYT ?

ttzn:
> Issues with the current PR:
> 
>    1. First we should add
> 
> 
> ```
> .linenoswrapper .linenos {
>   user-select: none;
> }
> ```
> 
> so that the user can copy&paste the selection without copying the line numbers.
> userselect
> 
>     1. Currently the `.code` macro has `white-space: pre` with `auto-scroll`, so if the lines are very big, the macro adds a scroll.
> 
> 
> scroll
> 
> The issue comes for Skins and situations where the white-space property is missing. In this case, the structure looks like:
> whitespace
> 
> So the issue is that there is no correlation between the line number and the code.
> I guess because of this issue the `table` layout is usually prefered. There are alternatives to tables: from `display: table` to using `display: grid`, but in these cases we would need a different HTML structure.
> 
> Let me know what you think about this. Thanks.

These are all valid points. I did try to address the line number-to-code binding, but the only clean way to do it is to make the parser split the content on new line characters (or sequences) before applying the format. This is pretty deep in the stack, and requires injecting the layout parameter all the way down to the parser, which implies a lot of small modifications. It is of course completely doable.

The thing is, in every example I looked up on the web (check out [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) and [this demo of SyntaxHighlighter](http://alexgorbatchev.com/SyntaxHighlighter/manual/demo/m), but there are many more), the authors don't seem to care at all. They're using a layout that is very similar to the one I implemented, with no tie between line numbers and lines of code in the HTML, and those are widely used libraries and visited websites. If it works for them, we should be fine.

As for the `white-space` problem, I don't really have an educated opinion. I guess the property should be set everywhere.

evalica:
So currently the CSS is valid for this PR.

I just want to mention again that with this solution the code text will not wrap and the line numbers will not be associated with long lines. 
This implementation will only work for small chunks of code displayed, and will always have a horizontal-scrollbar for longer lines:
Screenshot 2019-05-15 at 12 57 01

surli:
sorry @ttzn apparently we forgot to merge your PR and the release of 11.4RC1 was yesterday. Could you change the `@since` annotations to put 11.5RC1 now? I'll merge as soon as it's done. Thanks.

ttzn:
> sorry @ttzn apparently we forgot to merge your PR and the release of 11.4RC1 was yesterday. Could you change the `@since` annotations to put 11.5RC1 now? I'll merge as soon as it's done. Thanks.

No problem, will probably do so this evening.

surli:
@ttzn thanks for your contribution!
For future contrib, please use the JIRA issue key + title in message commits, it's easier afterwards to track the associated commits to a given issue. Here I squashed your commits so we have the right commit message.

PR #1109, opened 2019-05-07
XWIKI-16389: Dedicated macro for inline editing of wikimacro content

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16389

## Solution  

  * Create a dedicated macro for wikimacro content that insert the content of a wikimacro and put a dedicated metadata around it for inline editing
  * Process in DefaultWikiMacroRenderer the rendered wikimacro to remove the macro markers that could prevent from inline editing.

## Test

Unit tests has been added to the new module and some manual tests have been performed.
I mostly use this macro for testing manually my changes:
```
{{velocity}}
{{box}}
{{wikimacrocontent/}}
{{/box}}

This content is $xcontext.macro.content.length() characters length.
{{/velocity}}
```

I needed the fix provided for https://jira.xwiki.org/browse/XRENDERING-562 (see https://github.com/xwiki/xwiki-rendering/pull/174) to avoid some issues with the two consecutive metadatablocks created by box macro and wikimacrocontent.
I'm able to inline edit the content of this macro with this change and the fix from xwiki-rendering. 

However I encountered an issue when adding inside the macro box an error macro when inline editing it: it seems that any content added *after* the error box is removed when saving from the inline editor. I suspect another issue from xwiki-rendering.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-macros/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-macros/pom.xml
@@ -19,21 +19,22 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
-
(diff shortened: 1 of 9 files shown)


surli:
@mflorea so after our mail exchange, do you agree with this macro? I don't have much time until the release so I'd like to merge this one.

mflorea:
@surli OK, go. We can improve it later.

PR #1111, opened 2019-05-20
XWIKI-16412: Allow to type a custom value in a StaticListClass

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-16412
It also fixes
https://jira.xwiki.org/browse/XWIKI-16411

## Solution

  * Create the dedicated PropertyValuesProvider in rest-server to reuse
the mechanism already existing for picking a value
  * Create the dedicated displayer_staticlist.vm for it.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
@@ -881,7 +881,7 @@ public class BaseClass extends BaseCollection<DocumentReference> implements Clas
         boolean relationalStorage, String values, String displayType, String separators, String defaultValue)
     {
         return addStaticListField(fieldName, fieldPrettyName, size, multiSelect, relationalStorage, values,
-            displayType, separators, defaultValue, false);
+            displayType, separators, defaultValue, false, false);
     }
 
     /**
@@ -889,7 +889,7 @@ public class BaseClass extends BaseCollection<DocumentReference> implements Clas
      */
     public boolean addStaticListField(String fieldName, String fieldPrettyName, int size, boolean multiSelect,
         boolean relationalStorage, String values, String displayType, String separators, String defaultValue,
-        boolean allowCustomValues)
+        boolean allowCustomValues, boolean largeStorage)
     {
         if (get(fieldName) == null) {
             StaticListClass list_class = new StaticListClass();
@@ -913,6 +913,7 @@ public class BaseClass extends BaseCollection<DocumentReference> implements Clas
                 list_class.setAllowCustomValues(true);
                 list_class.setPicker(true);
             }
+            list_class.setLargeStorage(largeStorage);
             list_class.setObject(this);
             put(fieldName, list_class);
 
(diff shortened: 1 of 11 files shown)


surli:
@mflorea could you tell me if it's the right direction? I tried to reuse as most as possible the mechanism already existing for DBList. I tested it locally with a very simple static list for the wikimacro content type. 

The PR is not ready yet, I need to add unit test, but I wanted to validate it first.

surli:
@tmortagne I just added a new property for storing the list values in a LargeStringProperty. Tell me if it's ok for you.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
```diff
@@ -471,6 +504,9 @@ public BaseProperty newProperty()
             lprop = new DBStringListProperty();
         } else if (isMultiSelect()) {
             lprop = new StringListProperty();
+        // If we allow custom values, they might use a lot of space in DB, so better to put them in a LargeString.
```
tmortagne: This comment does not seems to have anything if the following if, leftover ?

surli: right. I'll remove it.

tmortagne:
It seems to me freeText property already been introduced for enabling/disabling custom value.

surli:
> It seems to me freeText property already been introduced for enabling/disabling custom value.

hmm indeed you're right. I will check for using it then.

surli:
> @surli I've seen it afterwards, much better.

ok if I merge then?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
```diff
@@ -898,6 +909,11 @@ public boolean addStaticListField(String fieldName, String fieldPrettyName, int
             if (defaultValue != null) {
                 list_class.setDefaultValue(defaultValue);
             }
+            list_class.setFreeText(freeText);
+            if (!freeText.equals(ListClass.FREE_TEXT_FORBIDDEN)) {
```
mflorea: It's safer to put the *constant* on the left when checking equality, to prevent a NPE.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/PropertyValue.java
```diff
@@ -100,4 +100,44 @@ public void setMetaData(java.util.Map<String, java.lang.Object> value)
...
     }
+
+    @Override
+    public boolean equals(java.lang.Object o)
+    {
+        if (o instanceof PropertyValue) {
+            PropertyValue other = (PropertyValue) o;
+
+            boolean equals;
+            if (this.getMetaData() != null) {
```
mflorea: Why not use ``EqualsBuilder``?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/BaseClass.java
surli: indeed, fixing.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/java/org/xwiki/rest/model/jaxb/PropertyValue.java
surli: apparently the commons-lang dependency is not available in this module. I can add it, wasn't sure what the best for this case.

PR #1115, opened 2019-06-05
XWIKI-16457: Update the identifier of the extension point used to make the HTML head plugable

slauriere:
This request aims at changing the identifier of the HTML head extension point.

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
@@ -2815,8 +2815,8 @@ Your account on ${wikiname} has been successfully activated. You can now login a
 &lt;meta name="keywords" content="$escapetool.xml($keywords.replaceAll('[|,]', ' ').trim())" /&gt;
 &lt;meta name="rating" content="General" /&gt;
 &lt;meta name="author" content="$escapetool.xml($!xwiki.getUserName($doc.getAuthor(), false))" /&gt;
-## Add HTML tags inside the page's head element, provided by UI Extensions implementing the "org.xwiki.platform.head" Extension Point
-#foreach($uix in $services.uix.getExtensions("org.xwiki.platform.head", {'sortByParameter' : 'order'}))
+## Add HTML tags inside the page's head element, provided by UI Extensions implementing the "org.xwiki.platform.html.head" Extension Point
+#foreach($uix in $services.uix.getExtensions("org.xwiki.platform.html.head", {'sortByParameter' : 'order'}))
   $services.rendering.render($uix.execute(), 'xhtml/1.0')
 #end
 ## We verify that we don't display RSS feeds on the login page since it causes problems if these feeds are


xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/XWikiPreferences.xml
```diff
@@ -2815,8 +2815,8 @@ Your account on ${wikiname} has been successfully activated. You can now login a
 &lt;meta name="keywords" content="$escapetool.xml($keywords.replaceAll('[|,]', ' ').trim())" /&gt;
 &lt;meta name="rating" content="General" /&gt;
 &lt;meta name="author" content="$escapetool.xml($!xwiki.getUserName($doc.getAuthor(), false))" /&gt;
-## Add HTML tags inside the page's head element, provided by UI Extensions implementing the "org.xwiki.platform.head" Extension Point
-#foreach($uix in $services.uix.getExtensions("org.xwiki.platform.head", {'sortByParameter' : 'order'}))
+## Add HTML tags inside the page's head element, provided by UI Extensions implementing the "org.xwiki.platform.html.head" Extension Point
+#foreach($uix in $services.uix.getExtensions("org.xwiki.platform.html.head", {'sortByParameter' : 'order'}))
```
vmassol: @slauriere I didn't check the history but if the previous UIXP id was released then it's a backward compatibility breakage. Is that the case?

vmassol: Just checked and indeed it's a breakage :( The UIXP was introduced in https://jira.xwiki.org/browse/XWIKI-16093 ie 11.1RC1. 

BTW the documentation page for it is in a bad shape :( see https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/HTMLHead

It says it's a work in progress which is not true since it was already committed and it's missing the version when it was introduced.

In addition issue https://jira.xwiki.org/browse/XWIKI-16093 has not documentation fields filled so I don't see how XWiki 11.1RC1 could have been released. Seems the RM failed to check missing doc even though it's marked as checked on https://dev.xwiki.org/xwiki/bin/view/ReleasePlans/ReleasePlan11.1RC1 (cc @mflorea do you remember?).

vmassol: So the canonical solution right now if you want to break an API is to send a vote on the devs list explaining why it's important and evaluating the consequences (at least that's our process).

vmassol: Related: FTR I created an issue about adding aliases for UIXP but it's not implemented: https://jira.xwiki.org/browse/XWIKI-12982

vmassol: Just noticed that Caty commented too at https://jira.xwiki.org/browse/XWIKI-16093?focusedCommentId=102664&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-102664

surli: > In addition issue jira.xwiki.org/browse/XWIKI-16093 has not documentation fields filled so I don't see how XWiki 11.1RC1 could have been released. Seems the RM failed to check missing doc even though it's marked as checked on dev.xwiki.org/xwiki/bin/view/ReleasePlans/ReleasePlan11.1RC1 (cc @mflorea do you remember?).

From what I understood it's Caty who close the issue and marked it as fixed in 11.1RC1 since the commit was pushed in this version. So it couldn't be found for the release since it was still open without a fix version.

vmassol: > From what I understood it's Caty who close the issue and marked it as fixed in 11.1RC1 since the commit was pushed in this version. So it couldn't be found for the release since it was still open without a fix version.

Indeed good point. So what we're missing is a check to find issues with commits and that are open. I can provide a filter for this easily. Will do and add to the Release Plan template.

slauriere: Thanks  @surli  and @vmassol for the guidance, and apologies for the extra work. I sent a vote to the devs list requesting an API breakage indeed, and I updated [the documentation page](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/HTMLHead). I have also filled in the documentation field of both [XWIKI-16093](https://jira.xwiki.org/browse/XWIKI-16093) and [XWIKI-16457](https://jira.xwiki.org/browse/XWIKI-16457). Please let me know if other actions are needed on my end.

vmassol: @slauriere thanks for updating the doc page. However, it's not fully correct on 2 aspects:
1) the version is still missing. Seems you put in the description which is not the right place. It should be put in the xobject ;) Also make sure that all xproperties are filled.
2) it's missing an example with a screenshot. We need a simple example inside the page. It's ok that you provide a link to a real life usage but we also need a simple example.

Thanks!

vmassol: > the version is still missing

You can verify this in the LT at https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/

slauriere: I've updated [the documentation page](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/HTMLHead) with the following changes indeed: version added in the XObject, other properties filled in, screenshot added, with also some code sample. Hoping everything is getting in order.

PR #1116, opened 2019-06-06
XWIKI-13365: Scripts are allowed in Comment previews

surli:
## Issue

https://jira.xwiki.org/browse/XWIKI-13365

## Solution

The template `plain.vm` is currently use to display previewed comment, since we are using a plain parameter. And right now `plain.vm` can use two values for the raw parameters documented in https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/Architecture/URL%20Architecture/Standard%20URL%20Format/:
> raw=1: Returns the page source as is without any formatting but XML-escaped
> raw=2: Returns the page source as is without any formatting (without any escaping)

In order to be able to render a content page in a restricted mode, I propose to add a new value, `raw=3` which would execute the content in restricted mode, and to call the preview of comments with this parameter.

I tested locally the solution and it seems to work well.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
@@ -284,7 +284,7 @@ viewers.Comments = Class.create({
          var notification = new XWiki.widgets.Notification("$services.localization.render('core.viewers.comments.preview.inProgress')", "inprogress");
          new Ajax.Request(previewURL, {
             method : 'post',
-            parameters : {'xpage' : 'plain', 'sheet' : '', 'content' : form.commentElt.value, 'form_token': form.form_token.value},
+            parameters : {'xpage' : 'plain', 'sheet' : '', 'content' : form.commentElt.value, 'form_token': form.form_token.value, 'raw': '3'},
             onSuccess : function (response) {
               this.doPreview(response.responseText, form);
               notification.hide();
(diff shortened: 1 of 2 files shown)


vmassol:
I don't understand the principle. If there's a parameter for the restricted mode then all you need to is to call the URL without this parameter to not be in restricted mode, no?

surli:
> If there's a parameter for the restricted mode then all you need to do is to call the URL without this parameter to not be in restricted mode, no?

I think you understood the contrary of what I'm trying to do :)
There is currently no parameter for the restricted mode. And we want to display the content in restricted mode. 
Hence my proposal to create a parameter for the restricted mode.

vmassol:
> I think you understood the contrary of what I'm trying to do :)
There is currently no parameter for the restricted mode. And we want to display the content in restricted mode.
Hence my proposal to create a parameter for the restricted mode.

I understood that but I guess the reason you want to do this is because it can cause some security issue if not executed in restricted mode. The way you're proposing to fix this is by adding a parameter so what would prevent someone malicious to just call that URL but without the parameter and thus bypass your security?

surli:
> I understood that but I guess the reason you want to do this is because it can cause some security issue if not executed in restricted mode.

wrong guess, read my comment here https://jira.xwiki.org/browse/XWIKI-13365?focusedCommentId=102909&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-102909 it's not a security issue, just a UI issue.

surli:
After discussing it with @tmortagne we decided that a new option parameter to be able to see an entire page content in restricted mode might be useful also for other usecases. So I created https://jira.xwiki.org/browse/XWIKI-16459 and I'll fix XWIKI-13365 with it.

PR #1118, opened 2019-06-27
XWIKI-14819:  The placeholder text from Administration > Extensions does not disappear when start typing on Edge

arcilli:
: The placeholder text from Administration > Extensions does not disappear when start typing on Edge

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extension.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extension.vm
@@ -36,8 +36,9 @@
           <input type="hidden" name="section" value="$escapetool.xml($request.section)" />
         #end
         <label class="hidden" for="extensionSearchInput">$services.localization.render('extensions.search.tip')</label>
-        <input type="text" #if ("$!request.search" == '')class="withTip" #{end}id="extensionSearchInput" name="search"
-          value="#if ("$!request.search" != '')$escapetool.xml($request.search)#{else}$services.localization.render('extensions.search.tip')#{end}" />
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/extension.vm
```diff
@@ -36,8 +36,9 @@
           <input type="hidden" name="section" value="$escapetool.xml($request.section)" />
         #end
         <label class="hidden" for="extensionSearchInput">$services.localization.render('extensions.search.tip')</label>
-        <input type="text" #if ("$!request.search" == '')class="withTip" #{end}id="extensionSearchInput" name="search"
```
surli: Looks like you removed the `#if ("$!request.search" == '')class="withTip" #{end}` part in your changes.

arcilli: [`withTip`](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/SpecialCSSClasses/#HB2.Behaviorclassnames) class was a solving for unavailability (in that version) of HTML5. Adding a placeholder (as the [solution](https://jira.xwiki.org/browse/XWIKI-14819?focusedCommentId=96747&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-96747) proposed by @mflorea) has the same effect and it's a possible fix.

surli: Ok I wasn't aware of this. LGTM then.

PR #1124, opened 2019-07-12
XWIKI-16580 : NPE when using $xwiki.ssfx in an async panel

aubincleme:
Set the default value of forceSkinAction to `true`


tmortagne:
Actually the real issue here is that the parameter set when use() is called was lost by the async rendering framework, it's supposed to be always set when use() is called (false being the default).

PR #1134, opened 2019-08-01
XWIKI-16147: Livetable suggest input field has no title

oanalavinia:


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
```diff
@@ -205,6 +205,11 @@ define('xwiki-selectize', ['jquery', 'selectize', 'xwiki-events-bridge'], functi
     if (this.selectize.settings.takeInputWidth) {
       this.selectize['$control'].width($(this).data('initialWidth'));
     }
+    // Set the title of the input field if it does exists in the select.
+    let inputTitle;
```
surli: Note that right now are functional test are executed on an old version of Firefox, which does not support the "let" keyword, see the comment from @mflorea: https://jira.xwiki.org/browse/XCOMMONS-862?focusedCommentId=98996&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-98996

Now we recently upgraded Guava which mean we should be able to upgrade Selenium and our version of firefox to run those functional test... So I guess we need to agree on that before merging this code.

oanalavinia: Then I will wait to see what's proposed for the upgrade. If it's not planned any time soon, I can change it to 'var'.

mflorea: Yes, better use ``var`` for now, but I'm not sure we need a variable. Do we really need the ``if``? Why not copy directly the value of the ``title`` attribute?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -1563,6 +1563,8 @@ $Msz MB##
       #set ($filterType = 'text')
     #end
   #end
+  #set($title = $services.localization.render('platform.livetable.filtersTitle',
+    [$services.localization.render("${transprefix}$column")]))
```
mflorea: Since you want to remove the duplication, you could include the escaping here also, and have an ``escapedTitle`` variable.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
oanalavinia: Indeed. This way I also won't need let/ var

PR #1135, opened 2019-08-05
XWIKI-16644: Symbolic link from etc/xwiki/xwiki-locales.txt to /usr/l…

ClemensRobbenhaar:
…ib/xwiki/WEB-INF is missing

add file to symlink config section


ClemensRobbenhaar:
Merged with OK in the jira issue.

PR #1139, opened 2019-08-21
XWIKI-16663: Computed field values should be available in REST representations

slauriere:
This pull request introduces method `serializePropertyValue(PropertyInterface property,PropertyClass propertyClass,  XWikiContext context)` in ModelFactory so that ComputedFieldClass  properties get their value evaluated when represented.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
@@ -958,8 +958,8 @@ public class ModelFactory
     }
 
     /**
-     * Serializes the value of the given XObject property. Returns an empty string for ComputedFieldClass properties.
-     * 
+     * Serializes the value of the given XObject property. ComputedField properties are not evaluated.
+     *
      * @param property an XObject property
      * @return the String representation of the property value
      */
@@ -993,9 +993,12 @@ public class ModelFactory
         if (propertyClass instanceof ComputedFieldClass) {
             // TODO: the XWikiDocument needs to be explicitely set in the context, otherwise method
             // PropertyClass.renderInContext fires a null pointer exception: bug?
+            XWikiDocument document = context.getDoc();
             context.setDoc(property.getObject().getOwnerDocument());
             ComputedFieldClass computedFieldClass = (ComputedFieldClass) propertyClass;
             String value = computedFieldClass.displayView(propertyClass.getName(), property.getObject(), context);
+            // Reset the context document to its original value.
+            context.setDoc(document);
             return value;
         } else {
             return serializePropertyValue(property);
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
@@ -958,7 +958,7 @@ public class ModelFactory
     }
 
     /**
-     * Serializes the value of the given XObject property. ComputedField properties are not evaluated.
+     * Serializes the value of the given XObject property. {@link ComputedFieldClass} properties are not evaluated.
      *
      * @param property an XObject property
      * @return the String representation of the property value
@@ -980,8 +980,8 @@ public class ModelFactory
     }
 
     /**
-     * Serializes the value of the given XObject property. In case the property is a ComputedFieldClass, the serialized
-     * value is the computed property value.
+     * Serializes the value of the given XObject property. In case the property is an instance of
+     * {@link ComputedFieldClass}, the serialized value is the computed property value.
      * @param property an XObject property
      * @param propertyClass the PropertyClass of that XObject proprety
      * @param context the XWikiContext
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -956,7 +958,7 @@ public Hierarchy toRestHierarchy(EntityReference targetEntityReference, Boolean
     }
 
     /**
-     * Serializes the value of the given XObject property.
+     * Serializes the value of the given XObject property. Returns an empty string for ComputedFieldClass properties.
```
sdumitriu: Is the javadoc accurate? It seems that an actual value is returned for computed properties, right?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -977,6 +979,30 @@ private static String serializePropertyValue(PropertyInterface property)
...
+     * @param context the XWikiContext
+     * @return the String representation of the property value
+     */
+    private static String serializePropertyValue(PropertyInterface property,
+        com.xpn.xwiki.objects.classes.PropertyClass propertyClass, XWikiContext context)
+    {
+        if (propertyClass instanceof ComputedFieldClass) {
+            // TODO: the XWikiDocument needs to be explicitely set in the context, otherwise method
+            // PropertyClass.renderInContext fires a null pointer exception: bug?
+            context.setDoc(property.getObject().getOwnerDocument());
```
sdumitriu: It's a good practice to always reset the context document to the original value.

slauriere: Thanks for the review, I've added a context document reset in the latest commit.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
slauriere: In the current implementation of BaseProperty:getValue, the stored value of the property is returned, not the one evaluated dynamically in the case of ComputedField properties. So it may not be empty, but it's not evaluated, currently (we may consider evaluating it directly in BaseProperty:getValue but that would be another topic, I guess). I updated the Javadoc accordingly, hoping it reflects better the current situation?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -977,6 +979,33 @@ private static String serializePropertyValue(PropertyInterface property)
...
+    private static String serializePropertyValue(PropertyInterface property,
+        com.xpn.xwiki.objects.classes.PropertyClass propertyClass, XWikiContext context)
+    {
+        if (propertyClass instanceof ComputedFieldClass) {
+            // TODO: the XWikiDocument needs to be explicitely set in the context, otherwise method
+            // PropertyClass.renderInContext fires a null pointer exception: bug?
+            XWikiDocument document = context.getDoc();
+            context.setDoc(property.getObject().getOwnerDocument());
+            ComputedFieldClass computedFieldClass = (ComputedFieldClass) propertyClass;
+            String value = computedFieldClass.displayView(propertyClass.getName(), property.getObject(), context);
```
tmortagne: A general issue is that displayView is designed to return HTML so this will work only for very simple String values. Same issue than when you use it in a script.

slauriere: Indeed... What about introducing a method ComputedFieldClass:getValue (or any similar name) which would execute the current code in ComputedFieldClass:displayView, and make sure that the ComputedFieldClass:displayView actually uses any possible custom displayer when there is one (it seems it's not the case at the moment), so that: "getValue" returns a raw computed value, and displayView returns the view to be displayed in HTML?

tmortagne: > What about introducing a method ComputedFieldClass:getValue

+1 for some ComputedFieldClass:getComputedValue, it's a good idea since we know we are manipulating a ComputedFieldClass

slauriere: I plan to work on that, then. Would it make sense to you also to possibly add the "display" property to the REST output (it could be optional)? In some cases indeed, we want to display the HTML view of a property value in a different context than XWiki, or even within XWiki but loaded via an ajax call. What do you think? (that would be in a dedicated separate issue in any case)

tmortagne: Not sure this is the right place for this need. I feel REST should be about data as much as possible. But we could think about some display resource handler.

slauriere:
I'm creating a new pull request containing the changes requested in this one because I mixed up my local branch.

PR #1140, opened 2019-08-21
Update from Weblate

xwikiorgci:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1671,7 +1671,7 @@ rendering.macroCategory.Internal=Internal
 ### Macro Descriptors
 rendering.macro.groovy.name=Groovy
 rendering.macro.groovy.description=Execute a groovy script.
-rendering.macro.groovy.content.description=the groovy script to execute
+rendering.macro.groovy.content.description=The wiki content to execute.
```
surli: Is this expected?

surli: @mflorea do you confirm that's wrong?

mflorea: @surli the issue is that the ``rendering.macro.groovy.content.description`` key is duplicated in oldcore's ``ApplicationResources.properties``. This commit seems to be the culprit https://github.com/xwiki/xwiki-platform/commit/7fee6198054656f2ec3d47246964e5f890ac4f30 . Bad copy & paste from @tmortagne ?

mflorea: So yes, it's wrong. We need to remove the duplicated key from ``ApplicationResources.properties``.

surli: Aaaah thanks! Actually weblate was telling me an alert with: 
> Duplicated string for translation.

and many translations and I wasn't understanding the reason... now it's clear.

PR #1142, opened 2019-08-27
XWIKI-16663: Computed field values should be available in REST representations

slauriere:
- Add method ComputedFieldClass:getComputedValue() performing an evaluation of the ComputedFieldClass script.
- Introduce method ModelFactor:serializePropertyValue(PropertyInterface, PropertyClass, XWikiContext context) so that the ComputedFieldClass properties get their value evaluated when represented in REST.

(or should we create a distinct issue focusing on the introduction of ComputedFieldClass:getComputedValue?)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
@@ -93,6 +93,7 @@ public class ComputedFieldClass extends PropertyClass
 
     /**
      * Computes and returns the raw value of this property for a given object.
+     *
      * @param name property name
      * @param prefix prefix to be added
      * @param object object for which the property value has to get computed
@@ -105,16 +106,16 @@ public class ComputedFieldClass extends PropertyClass
     {
         String script = getScript();
 
-            ScriptContext scontext = Utils.getComponent(ScriptContextManager.class).getCurrentScriptContext();
-            scontext.setAttribute("name", name, ScriptContext.ENGINE_SCOPE);
-            scontext.setAttribute("prefix", prefix, ScriptContext.ENGINE_SCOPE);
-            scontext.setAttribute("object", new com.xpn.xwiki.api.Object((BaseObject) object, context),
-                    ScriptContext.ENGINE_SCOPE);
+        ScriptContext scontext = Utils.getComponent(ScriptContextManager.class).getCurrentScriptContext();
+        scontext.setAttribute("name", name, ScriptContext.ENGINE_SCOPE);
+        scontext.setAttribute("prefix", prefix, ScriptContext.ENGINE_SCOPE);
+        scontext.setAttribute("object", new com.xpn.xwiki.api.Object((BaseObject) object, context),
+                ScriptContext.ENGINE_SCOPE);
 
-            XWikiDocument classDocument = object.getXClass(context).getOwnerDocument();
+        XWikiDocument classDocument = object.getXClass(context).getOwnerDocument();
 
-            return renderContentInContext(script, classDocument.getSyntax().toIdString(),
-                    classDocument.getAuthorReference(), classDocument.getDocumentReference(), context);
+        return renderContentInContext(script, classDocument.getSyntax().toIdString(),
+                classDocument.getAuthorReference(), classDocument.getDocumentReference(), context);
     }
 
     @Override
@@ -133,7 +134,7 @@ public class ComputedFieldClass extends PropertyClass
 
     @Override
     public void displayView(StringBuffer buffer, String name, String prefix, BaseCollection object,
-        XWikiContext context)
+            XWikiContext context)
     {
         try {
             buffer.append(getComputedValue(name, prefix, object, context));
@@ -145,14 +146,14 @@ public class ComputedFieldClass extends PropertyClass
 
     @Override
     public void displayEdit(StringBuffer buffer, String name, String prefix, BaseCollection object,
-        XWikiContext context)
+            XWikiContext context)
     {
         displayView(buffer, name, prefix, object, context);
     }
 
     @Override
     public void displayHidden(StringBuffer buffer, String name, String prefix, BaseCollection object,
-        XWikiContext context)
+            XWikiContext context)
     {
     }
 }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
@@ -90,6 +90,47 @@ public class ComputedFieldClass extends PropertyClass
         return sValue;
     }
 
+    /**
+     * Computes and returns the raw value of this property for a given object.
+     * @param name property name
+     * @param prefix prefix to be added
+     * @param object object for which the property value has to get computed
+     * @param context current context
+     * @return the computed property value
+     */
+    public String getComputedValue(String name, String prefix, BaseCollection object, XWikiContext context) {
+        String script = getScript();
+
+        try {
+            ScriptContext scontext = Utils.getComponent(ScriptContextManager.class).getCurrentScriptContext();
+            scontext.setAttribute("name", name, ScriptContext.ENGINE_SCOPE);
+            scontext.setAttribute("prefix", prefix, ScriptContext.ENGINE_SCOPE);
+            scontext.setAttribute("object", new com.xpn.xwiki.api.Object((BaseObject) object, context),
+                    ScriptContext.ENGINE_SCOPE);
+
+            XWikiDocument classDocument = object.getXClass(context).getOwnerDocument();
+
+            return renderContentInContext(script, classDocument.getSyntax().toIdString(),
+                    classDocument.getAuthorReference(), classDocument.getDocumentReference(), context);
+
+        } catch (Exception e) {
+            return e.getMessage();
+        }
+    }
+
+    /**
+     * Appends the raw computed property value of this object to the passed buffer.
+     * @param buffer buffer to which the property value has to get appended
+     * @param name property name
+     * @param prefix prefix to be added
+     * @param object object for which the property value has to get computed
+     * @param context current context
+     */
+    public void getComputedValue(StringBuffer buffer, String name, String prefix, BaseCollection object,
+            XWikiContext context) {
+        buffer.append(getComputedValue(name, prefix, object, context));
+    }
+
     @Override
     public BaseProperty fromString(String value)
     {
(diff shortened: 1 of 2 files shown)


slauriere:
This pull request is a follow-up to #1139

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -977,6 +979,33 @@ private static String serializePropertyValue(PropertyInterface property)
...
+        com.xpn.xwiki.objects.classes.PropertyClass propertyClass, XWikiContext context)
+    {
+        if (propertyClass instanceof ComputedFieldClass) {
+            // TODO: the XWikiDocument needs to be explicitely set in the context, otherwise method
+            // PropertyClass.renderInContext fires a null pointer exception: bug?
+            XWikiDocument document = context.getDoc();
+            context.setDoc(property.getObject().getOwnerDocument());
+            ComputedFieldClass computedFieldClass = (ComputedFieldClass) propertyClass;
+            String value = computedFieldClass.getComputedValue(propertyClass.getName(), "", property.getObject(), context);
+            // Reset the context document to its original value.
```
tmortagne: Would be safer to do that in a try/finally.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -90,6 +90,47 @@ public String getScript()
...
+    /**
+     * Appends the raw computed property value of this object to the passed buffer.
+     * @param buffer buffer to which the property value has to get appended
+     * @param name property name
+     * @param prefix prefix to be added
+     * @param object object for which the property value has to get computed
+     * @param context current context
+     */
+    public void getComputedValue(StringBuffer buffer, String name, String prefix, BaseCollection object,
+            XWikiContext context) {
```
tmortagne: Codestyle.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -90,6 +90,47 @@ public String getScript()
...
     }
 
+    /**
+     * Computes and returns the raw value of this property for a given object.
+     * @param name property name
+     * @param prefix prefix to be added
+     * @param object object for which the property value has to get computed
+     * @param context current context
+     * @return the computed property value
+     */
```
tmortagne: Don't forget the `@since`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -90,6 +90,47 @@ public String getScript()
...
+            scontext.setAttribute("object", new com.xpn.xwiki.api.Object((BaseObject) object, context),
+                    ScriptContext.ENGINE_SCOPE);
+
+            XWikiDocument classDocument = object.getXClass(context).getOwnerDocument();
+
+            return renderContentInContext(script, classDocument.getSyntax().toIdString(),
+                    classDocument.getAuthorReference(), classDocument.getDocumentReference(), context);
+
+        } catch (Exception e) {
+            return e.getMessage();
```
tmortagne: You should keep the previous TODO since just returning the exception message is not so great.

tmortagne:
A few minor things but the general logic looks good.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -108,25 +149,7 @@ public BaseProperty newProperty()
...
-
-            String result = renderContentInContext(script, classDocument.getSyntax().toIdString(),
-                classDocument.getAuthorReference(), classDocument.getDocumentReference(), context);
-
-            buffer.append(result);
-        } catch (Exception e) {
-            // TODO: append a rendering style complete error instead
-            buffer.append(e.getMessage());
-        }
+        buffer.append(getComputedValue(name, prefix, object, context));
```
mflorea: Either call ``getComputedValue(buffer, ...)`` here or remove the second signature of ``getComputedValue`` since I don't see it used anywhere.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
slauriere: Indeed, I had removed it by mistake, I restored it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
slauriere: I added it indeed in the latest commit.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
slauriere: This method was removed, taking into account Marius' suggestion indeed.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
slauriere: Thanks but I'm not sure what to put in the "finally" clause. In case an exception is raised, should we propagate the exception, or simply return a fake value containing the exception's message? Or don't catch any exception at all? What'd be your suggestion?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
slauriere: Indeed, I removed the signature altogether since it was not used.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
tmortagne: I did not mentioned `catch` (which is not mandatory). My point is that you should restore the context in a `finally` to be sure you won't leave it with a bad value in case something bad happen.

slauriere: Right, I thought your concern was about the explicit cast to ComputedFieldClass, I had overlooked the context reset aspect. I added a try / finally in the latest commit indeed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -90,6 +90,36 @@ public String getScript()
...
+    /**
+     * Computes and returns the raw value of this property for a given object.
+     * @param name property name
+     * @param prefix prefix to be added
+     * @param object object for which the property value has to get computed
+     * @param context current context
+     * @return the computed property value
+     * @since 11.8
+     */
+    public String getComputedValue(String name, String prefix, BaseCollection object, XWikiContext context) {
```
surli: Codestyle: `{` should be on a new line.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
surli: I would throw an XWikiException here, and keep this with the todo in a try/catch of `displayView` then you should be able to fallback in case of problem in `ModelFactory`. Right now, you risk to put a wrong information in the field in case of error.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
tmortagne: It's actually `11.8RC1` right now (but better wait for the last minute before fixing that to be sure).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
slauriere: Thanks surli, I understand the risk of displaying wrong information is an issue, I tried to get my head around your suggestion but I'm not sure how to fallback. Also, the method PropertyClass:renderContentInContext throws Exception, not XWikiException so I'm wondering what to throw. I brought some changes, please let me know what you think.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
slauriere: Indeed, I fixed that and other formatting issues.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/ModelFactory.java
```diff
@@ -977,6 +979,35 @@ private static String serializePropertyValue(PropertyInterface property)
...
+        com.xpn.xwiki.objects.classes.PropertyClass propertyClass, XWikiContext context)
+    {
+        if (propertyClass instanceof ComputedFieldClass) {
+            // TODO: the XWikiDocument needs to be explicitely set in the context, otherwise method
+            //  PropertyClass.renderInContext fires a null pointer exception: bug?
+            XWikiDocument document = context.getDoc();
+            try {
+                context.setDoc(property.getObject().getOwnerDocument());
+                ComputedFieldClass computedFieldClass = (ComputedFieldClass) propertyClass;
+                return computedFieldClass.getComputedValue(propertyClass.getName(), "", property.getObject(), context);
```
surli: So if I'm right you can now catch an exception here in case `getComputedValue` didn't worked for some reason. And in the catch I guess you can log an error with the actual exception, and fallback on `serializePropertyValue(property);`.

slauriere: Right, thank you, good idea, I added that. I had to turn the method `ModelFactory:serializePropertyValue(PropertyInterface,PropertyClass,XWikiContext)` to not static in order to access the logger field.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -106,23 +134,10 @@ public BaseProperty newProperty()
 
     @Override
     public void displayView(StringBuffer buffer, String name, String prefix, BaseCollection object,
-        XWikiContext context)
+            XWikiContext context)
```
surli: Codestyle: Looks like you added new tabulation in each method in case of line return. You might want to remove it.

slauriere: Indeed, not sure how I come up with this formatting since I has reimported and reapplied the [XWiki IntelliJ code style](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HIntelliJIDEA4.52B), and restarted IntelliJ, but it keeps adding this tab when formatting the code...

surli:
Thanks for the work here @slauriere :+1: For next time, don't forget to prefix each commit message with the code of the issue: I had to squash your commits here to avoid having some commits with no related issue in master.

slauriere:
Thank you @surli, @tmortagne, @mflorea. Regarding the commits: I thought a pull request was always squashed actually, since several commit messages may be irrelevant in the master context, but ok, I'll add the issue number to each commit including in a pull request branch.

surli:
> I thought a pull request was always squashed actually, since several commit messages may be irrelevant in the master context

it's up to us actually. Here there was quite a few commits so I'd eventually squash them anyway, but it's not mandatory at all, and it can help to explain the story to get all the commits on master at the end.

PR #1145, opened 2019-09-03
XWIKI-16663: Computed field values should be available in REST representation

slauriere:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ComputedFieldClass.java
```diff
@@ -99,6 +98,7 @@ public String getScript()
      * @param object object for which the property value has to get computed
      * @param context current context
      * @return the computed property value
+     * @throws Exception in case an error occurs
      * @since 11.8RC1
      */
     public String getComputedValue(String name, String prefix, BaseCollection object, XWikiContext context) throws
```
surli: Actually I notice it only now but you should add an `@Unstable` annotation since it's a new public API.

tmortagne: Yes that too.

PR #1148, opened 2019-09-05
XWIKI-16091: "My groups" tab in user profile displays group fullname,…

oanalavinia:
… not title

* displays group page title when this is defined


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserMembershipSheet.xml
```diff
@@ -93,9 +93,13 @@
   #end
   #set($rows = [])
   #foreach ($group in $subGroups)
+    #set ($title = $xwiki.getDocument($group).title)
+    #if ($title == '')
```
mflorea:
```suggestion
    #if ("$!title" == '')
```
I would use the silent notation to be extra safe (in case ``getDocument`` returns ``null`` for instance).

PR #1150, opened 2019-09-06
[misc] Keep master build containing potentially flickering test…

tmortagne:
Depends on https://github.com/xwiki/xwiki-jenkins-pipeline/pull/3

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,8 @@ def builds = [
     build(
       name: 'Main',
       profiles: 'legacy,integration-tests,snapshotModules',
-      properties: '-Dxwiki.checkstyle.skip=true -Dxwiki.surefire.captureconsole.skip=true -Dxwiki.revapi.skip=true'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -29,7 +29,8 @@ def builds = [
     build(
       name: 'Main',
       profiles: 'legacy,integration-tests,snapshotModules',
-      properties: '-Dxwiki.checkstyle.skip=true -Dxwiki.surefire.captureconsole.skip=true -Dxwiki.revapi.skip=true'
+      properties: '-Dxwiki.checkstyle.skip=true -Dxwiki.surefire.captureconsole.skip=true -Dxwiki.revapi.skip=true',
+      daysToKeepStr: env['BRANCH_NAME'] == 'master' ? '30' : null
```
vmassol: I hope that `env['BRANCH_NAME']` will work. AFAIR it doesn't always depending on the context. At least I remember having issues with it but I don't recall fully the situation.

tmortagne: OK, I guess I could add a print to make sure to that.

PR #1152, opened 2019-09-11
Update from Weblate

xwikiorgci:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources_fr.properties
```diff
@@ -50,7 +50,7 @@
 ###############################################################################
 
 ### Languages
-language=Langue
+language=Langue [Debug of weblate test: please don't change that, I'll remove it myself]
```
surli: This needs to be removed.

PR #1153, opened 2019-09-11
Update from Weblate

xwikiorgci:


xwiki-platform-core/xwiki-platform-flavor/xwiki-platform-flavor-api/src/main/resources/ApplicationResources_fr.properties
```diff
@@ -18,7 +18,7 @@
 # 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 # ---------------------------------------------------------------------------
 
-### Missing: flavor.picker.noflavor=No Flavor, let the wiki be empty.
+flavor.picker.noflavor=Aucune Saveur, garder le wiki vide.
```
surli: Not quite sure "Saveur" is the right word to use there... any idea? :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources_zh_Hans.properties
```diff
@@ -0,0 +1,5412 @@
+# ---------------------------------------------------------------------------
```
surli: @tmortagne I have a doubt: zh_Hans is supposed to be supported by Java?

tmortagne: Doing `new Locale(zh_Hans)` wrong crash but the right code is `zh_CN` (see Locale#SIMPLIFIED_CHINESE) and in the context of XWiki this is actually `zh` since we decided to make it the default variant.

surli: And apparently right now we have a user dedicated to do translations for zh_Hans :-/

tmortagne: Maybe users can create the language they want and the upgrade restored `zh_HANS`.

xwiki-platform-core/xwiki-platform-flavor/xwiki-platform-flavor-api/src/main/resources/ApplicationResources_fr.properties
tmortagne: I think we said that we would keep "Flavor" in French.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources_zh_Hans.properties
surli: @sugov5 thanks for your contribution on XWiki translations! As specified on this thread by @tmortagne for technical reasons the actual translation in Chinese (simplified) have to be made with the code `zh` not `zh_Hans`. The language with the code `zh_Hans` was actually supposed to be deleted from our Weblate platform, but apparently it has been created back. So we are going to delete it again, sorry for the inconveniance.

PR #1157, opened 2019-09-17
XWIKI-16623: Make password changing consistent

arcilli:
Added a template for storing macros that are used to register a user, reset/change the password. 

Since for `reset the password` form no validation was made, I added the same mechanism that is used in registration & changing the password, besides a server validation for minimum 6 characters.

Also, moved the label with original password (at changing the password) at the bottom of the form, before the submit button.


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
```diff
@@ -0,0 +1,219 @@
...
+        },
+        'regex': {
+          'pattern': '/.{6,}/',
+          'failureMessage': $services.localization.render('xe.admin.registration.passwordTooShort')
+        }
+      }
+    })
+  #set ($discard = $fields.add($field))
+##
+## The confirm password field, mandatory, must match password field, and must also be 6+ characters long.
```
Enygma2002: Missing indentation?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
```diff
@@ -0,0 +1,219 @@
...
+  #set ($field =
```
Enygma2002: This `$field` parameter does not really seem used or useful, except for adding to the list. Why not do 
```
#set ($discard = $fields.add({
...
})
```
instead?

It's more explicit to the intent and does not create complications with redefined variables.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * @param $fields The array of fields to use for generating HTML code.
+ * @param $request The request that is made by submitting the form.
+ *#
+#macro (generateHtml, $fields, $request)
+  ## Put the same values back into the fields (if is there any problem with a field from the request that is made).
+  #getParams($fields, $request)
+  <dl>
+  #foreach ($field in $fields)
+    #if ($field.get('name'))
```
Enygma2002: In velocity, you can simplify all those `.get('something')` into `.something` directly and the code should look much more readable.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
```diff
@@ -0,0 +1,219 @@
...
+        ## If a parameter is specified as '' then we don't include it.
+        #if ($entry.getValue() != '')
+          $entry.getKey()="$escapetool.xml($entry.getValue())" ##
+        #end
+      #end
+      ></$tag>
+      #if ($field.get('error'))
+        <span ##
+        #foreach ($entry in $failureMessageParams.entrySet())
+          $entry.getKey()="$entry.getValue()" ##
```
Enygma2002: Same here... `$entry.key`, `$entry.value`

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
```diff
@@ -0,0 +1,219 @@
...
+        #end
+        >$field.get('error')</span>
+      #end
+      </dd>
+    #else
+      ERROR: Field with no name.
```
Enygma2002: Hmm, untranslated message. I know it's not part of this issue, but it would have been good to handle it. (optional)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
arcilli: From what I understood from @mflorea for templates, the blank spaces at the beginning of lines will not be removed. Considering that, spaces before the comments will result in a blank space.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/passwd.vm
```diff
@@ -51,32 +42,44 @@ document.observe('xwiki:dom:loaded', function() {
...
+        #definePasswordFields($fields, 'password', 'password2')
+        ## Change the label for new password fields
+        #set ($fields[1].label = $services.localization.render('platform.core.profile.passwd.newPassword'))
+        ## This key fits better.
+        #set ($fields[2].label = $services.localization.render('xe.admin.passwordReset.step2.newPasswordVerification.label'))
+        <form class="xform third" action="" method="post" autocomplete="off">
         ## CSRF prevention
         <div class="hidden"><input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /></div>
-        <dl>
-          #if($isCurrentUsersProfile || !$hasAdmin)
```
Enygma2002: Just noticed this. Are you still allowing admins to change the password of a user without knowing the old password? AFACIT, you still consider when validating the password change, but I don't see where you consider this in the UI, to avoid asking the admin for the old password.

This is a usecase that was supported before and we must make sure we still support it.

arcilli: I'll check this.

arcilli: @Enygma2002, you're right, I am not considering that use case anymore. I'm reviewing the changes and come back with some commits.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/passwd.vm
```diff
@@ -51,32 +42,49 @@ document.observe('xwiki:dom:loaded', function() {
...
+        <a href = "$doc.getURL("view")">$services.localization.render('platform.core.profile.passwd.return')</a>
       #end
       #if(!$isValidPassword)
-        #set($discard = $xwiki.jsfx.use('uicomponents/widgets/validation/livevalidation_prototype.js'))
-        #set($discard = $xwiki.ssfx.use('uicomponents/widgets/validation/livevalidation.css', true))
         $services.localization.render('platform.core.profile.passwd.instructions')
-
-      <form class="xform third" action="" method="post" autocomplete="off">
+        #set ($fields = [])
+        #if (!$hasAdmin)
```
Enygma2002: Don't forget the `$isCurrentUsersProfile` check, since a user, even if admin, should always be asked for the old password if he's changing his own password.

Most importantly: It's handled on the server side so we need to do it on the UI as well, otherwise an admin user will not be able to change his password (i.e. the server side will always ask for the original password that the UI never offered to fill in).

Generally: You must keep in mind that your refactoring must not affect the previous functionality (and/or reduce the previously supported use cases and edge cases) in any way outside the scope of (or unless it's clearly specified/intended in) the current issue you're fixing.

arcilli: Thanks for your observation, @Enygma2002. I'll investigate it next days.

surli:
@arcilli I kind of broke your PR while working on https://jira.xwiki.org/browse/XWIKI-11309, but I'd need this PR to be finished to be more consistent on the topic. Would that be ok for you if I continue the work you begin there?

arcilli:
@surli, I would appreciate it. If it's helping you, tomorrow I'll make the changes @Enygma2002 was talking about.

surli:
> If it's helping you, tomorrow I'll make the changes @Enygma2002 was talking about.

I think I already took care of it in my latest commit, but I'd be glad if you could review it

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/register_macros.vm
```diff
@@ -0,0 +1,217 @@
...
+    document.observe('xwiki:dom:updated', function(event) {
+      var container = (event && event.memo && event.memo.elements && event.memo.elements[0]) || $('body');
+      if (container.down('form#register')) {
+        initRegistrationFormValidation();
+      }
+    });// ]]>
+    </script>
+#end
+##
+#macro (outputPasswordFields, $fields, $passwordFieldName, $confirmPasswordFieldName, $request)
```
surli: Where is this macro actually used? I cannot find it in the diff or in the project?

arcilli: As the way it was designed first time, it was supposed to be used for change password form. After that, changes were needed and it's not used any more. FTM, it's unused.

Thanks for catching that.

surli: OK I'll remove it on my next commit then. I'm still doing some changes regarding the validation.

PR #1159, opened 2019-09-19
XWIKI-16725: Add action to create Template Provider from the Class UI

acotiuga:
* provide a way to create a template provider from a class
* the template must be created first, in order to be used in the template provider
* by default, the template provider is available only in the root space of the class (for XWiki.Code.MyClass -> XWiki)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-xclass/xwiki-platform-xclass-ui/src/main/resources/XWiki/ClassSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-xclass/xwiki-platform-xclass-ui/src/main/resources/XWiki/ClassSheet.xml
@@ -112,6 +112,11 @@
     #set ($classTemplateReference = $services.model.createDocumentReference("${className}Template",
       $doc.documentReference.parent))
   #end
+  ## Determine the template provider using naming convention.
+  #set ($classTemplateProviderReference = $services.model.createDocumentReference("${className}TemplateProvider",
+    $doc.documentReference.parent))
+  #set ($classTemplateProviderDoc = $xwiki.getDocument($classTemplateProviderReference))
+  #set ($hasClassTemplateProvider = !$classTemplateProviderDoc.isNew())
   #set($classTemplateDoc = $xwiki.getDocument($classTemplateReference))
   #set($hasClassSheets = !$classSheetReferences.isEmpty() || $xwiki.exists($defaultClassSheetReference))
   #set($hasClassTemplate = !$classTemplateDoc.isNew())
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-xclass/xwiki-platform-xclass-ui/src/main/resources/XWiki/ClassSheet.xml
```diff
@@ -318,6 +323,46 @@
...
+    (% id="HClassTemplateProvider" %)
+    = {{translation key="platform.xclass.defaultClassSheet.templateProvider.heading"/}} =
+
+      {{info}}$services.localization.render('platform.xclass.defaultClassSheet.templateProvider.description',
+        ['//']){{/info}}
+
+    #if (!$hasClassTemplateProvider)
+      #set ($templateProviderClassName = 'XWiki.TemplateProviderClass')
+      ## Do the page creation and object addition in one step, providing some default values.
+      #set ($restrictionSpace = $doc.documentReference.spaceReferences.get(0).name)
```
vmassol: This looks like a code smell. Will it work with nested pages?

vmassol: At the very least there needs a comment explaining why we're doing this and picking the first space.

xwiki-platform-core/xwiki-platform-xclass/xwiki-platform-xclass-ui/src/main/resources/XWiki/ClassSheet.xml
```diff
@@ -318,6 +323,46 @@
     [[$services.localization.render('platform.xclass.defaultClassSheet.template.view',
       [$templatePath.trim()]) »&gt;&gt;${classTemplateDoc.fullName}]]
   #end
+  ## Create a template provider only if a template for the current class exists.
+  #if ($classTemplateDoc.getObject(${doc.fullName}))
+    (% id="HClassTemplateProvider" %)
+    = {{translation key="platform.xclass.defaultClassSheet.templateProvider.heading"/}} =
+
+      {{info}}$services.localization.render('platform.xclass.defaultClassSheet.templateProvider.description',
+        ['//']){{/info}}
```
vmassol: why the `//`?

vmassol: If it's for italics, that seems weird.

acotiuga: Yes, this is for italics and I used it to be consistent with the existing approach of the sheet and template designed in this page. We can't use style in translation pages since their syntax is `plain/1.0` and I didn't want to duplicate the parameter as I can use the same for wrapping the text (`{0}some italic text{0}`)

vmassol: I don't fully recall our best practice but I remember we had some rules about wiki markup or HTML in translation. Does someone remember? cc @tmortagne @mflorea

xwiki-platform-core/xwiki-platform-xclass/xwiki-platform-xclass-ui/src/main/resources/XWiki/ClassSheet.xml
acotiuga: I added a comment to explain this in https://github.com/xwiki/xwiki-platform/commit/3de26e46b78a651ba27331528dd0c6e9c28eb53d.

xwiki-platform-core/xwiki-platform-xclass/xwiki-platform-xclass-ui/src/main/resources/XWiki/ClassSheet.xml
mflorea: @vmassol no HTML or wiki syntax in translation messages, that's the best practice I've been using so far.

tmortagne: Yes better avoid any kind of syntax as much as possible to avoid trouble. Also in the case of wiki syntax it can create important security holes.

vmassol: > no HTML or wiki syntax in translation messages,

Yes I remember that part. It just seems a bit weird to me that we use parameters to specify wiki syntax. Seems like we're missing something somewhere since that shouldn't be the role of parameters I think.

tmortagne: Yes what we are missing is a translation syntax.

PR #1190, opened 2019-12-04
XWIKI-16905: Upgrade to CSS4J 2.0.0

carlosame:
[Css4j 2.0.0](https://github.com/css4j/css4j-dist/releases/tag/css4j-2.0.0) is the latest available version of the library, but is not totally backwards-compatible with 1.0. This pull request makes the necessary changes, plus a small addition (commit 3e45129).

Unfortunately I was unable to verify that `xwiki-platform` builds after the change, due to this:

`[ERROR] Failed to execute goal on project xwiki-platform-rendering-macro-script: Could not resolve dependencies for project org.xwiki.platform:xwiki-platform-rendering-macro-script:jar:12.0-SNAPSHOT: Could not find artifact org.xwiki.rendering:xwiki-rendering-transformation-macro:jar:tests:12.0-20191129.175630-16 in xwiki-snapshots (http://nexus.xwiki.org/nexus/content/groups/public-snapshots) -> [Help 1]`

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
@@ -48,6 +48,8 @@ import org.dom4j.io.XMLWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.InputSource;
+import org.xml.sax.SAXNotRecognizedException;
+import org.xml.sax.SAXNotSupportedException;
 import org.xml.sax.XMLReader;
 import org.xwiki.bridge.DocumentAccessBridge;
 import org.xwiki.environment.Environment;
@@ -329,7 +331,11 @@ public class PdfExportImpl implements PdfExport
             SAXReader reader = new SAXReader(docFactory);
             // Dom4J 2.1.1 disable external DTD by default which is required here, putting it back
             // See https://github.com/dom4j/dom4j/issues/51
-            reader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", true);
+            try {
+                reader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", true);
+            } catch (SAXNotSupportedException e) {
+            } catch (SAXNotRecognizedException e) {
+            }
             reader.setEntityResolver(new DefaultEntityResolver());
             XHTMLDocument document = (XHTMLDocument) reader.read(source);
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
@@ -47,7 +47,6 @@ import org.dom4j.io.SAXReader;
 import org.dom4j.io.XMLWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.w3c.dom.css.CSSStyleDeclaration;
 import org.xml.sax.InputSource;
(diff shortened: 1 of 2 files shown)


vmassol:
Thanks a lot @carlosame I'll check this out tomorrow.

xwiki-platform-distribution/pom.xml
```diff
@@ -379,23 +379,6 @@
                       <class>org.w3c.dom.events.UIEvent</class>
                     </classes>
                   </exception>
-                  <!-- This is a PITA. The classes are actually incompatible! We should do something about it but I
```
tmortagne: Unfortunately css4j is not the only thing in the XWiki stack which depends on sac as far as I can see so I don't think we can remove that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
```diff
@@ -322,31 +324,26 @@ String applyCSS(String html, String css, XWikiContext context)
...
-            docFactory.getStyleSheetFactory().getUserAgentStyleSheet(
-                CSSDocument.ComplianceMode.STRICT).getCssRules().clear();
             SAXReader reader = new SAXReader(docFactory);
             // Dom4J 2.1.1 disable external DTD by default which is required here, putting it back
             // See https://github.com/dom4j/dom4j/issues/51
-            reader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", true);
+            try {
+                reader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", true);
+            } catch (SAXNotSupportedException e) {
+            } catch (SAXNotRecognizedException e) {
```
vmassol: @carlosame could you explain why it's ok to ignore these 2 SAX errors? If external DTD loading is required by CSS4J then how would it work if there's a SAX error? Thx

vmassol:
@carlosame would be awesome if you could rebase the PR to include a reference to https://jira.xwiki.org/browse/XWIKI-16905, i.e. add on the first comment line: 

`XWIKI-16905: Upgrade to CSS4J 2.0.0``

I should be able to do it on my side too if need be.

xwiki-platform-distribution/pom.xml
carlosame: That's what I was fearing (I was unable to verify the build). Removed that patch, thank you.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
carlosame: The catch clauses are introduced in case somebody specifies a non-Xerces parser in JAXP. The flows are the following:

A) JAXP gives a Xerces instance:

1. dom4j 2.1.1 sets `"http://apache.org/xml/features/nonvalidating/load-external-dtd"` to false, to disable DTD. No exception is thrown.
2. XWiki sets `"http://apache.org/xml/features/nonvalidating/load-external-dtd"` to true, to enable DTD loading. No exception.

B) JAXP gives a non-Xerces instance:
1. dom4j 2.1.1 sets `"http://apache.org/xml/features/nonvalidating/load-external-dtd"` to false. The feature is not recognized: exception is thrown and ignored.
2. XWiki sets `"http://apache.org/xml/features/nonvalidating/load-external-dtd"` to true. Exception is thrown (and was not being ignored). **Boom**.

vmassol: ok so the idea is that with a non-xerces parser, what dom4j does is just ignore in this case and thus we don't need to compensate for it. 

Got it! Thx, I'll add ample comments since it's not obvious at first sight :)

carlosame:
> @carlosame would be awesome if you could rebase the PR to include a reference to https://jira.xwiki.org/browse/XWIKI-16905

Done.

vmassol:
Applied, thanks @carlosame ! :)

PR #1229, opened 2020-02-20
XWIKI-17041: Office server states are not translated in the administr…

manuelleduc:
…ation

Openoffice server status translation returned instead of the enumeration value.
Consequently, 4 new translation keys are introduced, for each existing server status.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
@@ -23,10 +23,12 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import org.apache.commons.text.CaseUtils;
 import org.slf4j.Logger;
 import org.xwiki.bridge.DocumentAccessBridge;
 import org.xwiki.component.annotation.Component;
 import org.xwiki.context.Execution;
+import org.xwiki.localization.ContextualLocalizationManager;
 import org.xwiki.model.ModelContext;
 import org.xwiki.officeimporter.server.OfficeServer;
 import org.xwiki.officeimporter.server.OfficeServerConfiguration;
@@ -58,6 +60,17 @@ public class OfficeServerScriptService implements ScriptService
      * Error message used to indicate that the current user does not have enough rights to perform the requested action.
      */
     private static final String ERROR_PRIVILEGES = "Inadequate privileges.";
+
+    /**
+     * Prefix of the translation keys for server states.
+     */
+    private static final String TRANSLATION_KEY_SERVER_STATE_PREFIX = "office.config.serverState.";
+
+    /**
+     * The object used to translate translation keys.
+     */
+    @Inject
+    private ContextualLocalizationManager contextualLocalizationManager;
 
     /**
      * The object used to log messages.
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
```diff
@@ -59,6 +60,18 @@
      */
     private static final String ERROR_PRIVILEGES = "Inadequate privileges.";
 
+    /**
+     * Prefix of the translation keys for server states.
+     */
+    private static final String TRANSLATION_KEY_SERVER_STATE_PREFIX =
+        "xe.officeimporter.openoffice.serverstate.";
```
surli: Looks like the prefix does not match the newly introduced keys.

manuelleduc: Indeed. It's should be better now.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-ui/src/main/resources/XWiki/OfficeImporterTranslations.xml
```diff
@@ -44,7 +44,11 @@ office.form.append.hint=The document already exists! Append the result to the ex
 office.config.serverType.hint=The Office server type. It can be internally managed (local) or external (remote).
 office.config.serverPort.hint=The port number used for connecting to the Office server instance
 office.config.autoStart.hint=Specify if the Office server should be started / connected automatically at XWiki startup
-office.config.serverPath.hint=The path to the Office installation</content>
+office.config.serverPath.hint=The path to the Office installation
+office.config.serverstate.connected=Connected
```
vmassol: This is no camel case :) Should be ServerState.

vmassol: FTR there are guidelines on the naming, see https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HTranslationPropertyNaming

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
```diff
@@ -147,7 +159,8 @@ public boolean stopServer()
     public String getServerState()
     {
         this.officeServer.refreshState();
-        return this.officeServer.getState().toString();
+        return contextualLocalizationManager.getTranslationPlain(
```
vmassol: Small detail: missing `this.`. See the line you removed for ex.

manuelleduc:
- The translation keys are now in camel case
 - `this` is now explicit on the returned method call

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
```diff
@@ -23,10 +23,12 @@
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import org.apache.commons.text.CaseUtils;
```
vmassol: Small detail: Missing an explicit dependency in the pom.xml on commons text.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/server/script/OfficeServerScriptService.java
```diff
@@ -147,7 +160,10 @@ public boolean stopServer()
     public String getServerState()
     {
         this.officeServer.refreshState();
-        return this.officeServer.getState().toString();
+        final String normalizedStatusKey =
+            CaseUtils.toCamelCase(this.officeServer.getState().name().toLowerCase(), false, '_');
```
vmassol: I don't know this api. Could you give an example of a state and what it translates it into? Might be good to put as a comment for explanation.

vmassol: Small note: we don't use `final`. It's an antipattern reported by SonarQube.

vmassol: Still we don't really use it FTM as a best practice so for code homogeneity I wouldn't use it. But if you care about it, you could send a proposal to use that to see what others think. What's important is that we all agree about our coding practices since there's no code ownership.

manuelleduc:
* I have documented the enum conversion operation
* the `final` is removed

PR #1236, opened 2020-02-25
XWIKI-11269: WYSIWYG word is not consistent on different pages

sanjithpk:
XWIKI-11269: WYSIWYG word is not consistent on different pages
Link: https://jira.xwiki.org/browse/XWIKI-11269

Before:
image

After:
image

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-test/xwiki-platform-user-profile-test-docker/src/test/it/org/xwiki/user/test/ui/UserProfileIT.java
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-test/xwiki-platform-user-profile-test-docker/src/test/it/org/xwiki/user/test/ui/UserProfileIT.java
@@ -71,7 +71,7 @@ public class UserProfileIT
 
     private static final String USER_BLOGFEED = "http://xwiki.org/feed";
 
-    private static final String WYSIWYG_EDITOR = "Wysiwyg";
+    private static final String WYSIWYG_EDITOR = "WYSIWYG";
 
     private static final String TEXT_EDITOR = "Text";
 


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-test/xwiki-platform-user-profile-test-docker/src/test/it/org/xwiki/user/test/ui/UserProfileIT.java
```diff
@@ -71,7 +71,7 @@
 
     private static final String USER_BLOGFEED = "http://xwiki.org/feed";
 
-    private static final String WYSIWYG_EDITOR = "Wysiwyg";
+    private static final String WYSIWYG_EDITOR = "WYSIWYG";
```
surli: This is actually a piece of code dedicated to test the interface: it will looks for the String "Wysiwyg", changing it is more likely to break some tests since it won't find that String anymore.

It's more likely that the changes should take place in the templates, in xwiki-platform-web or in xwiki-platform-flamingo-skin-resources. It's also possible that some of them are in wiki pages in XML files.

9inpachi:
Hi @sanjithpk 

Please change the Pull Request title to follow the standard practices as mentioned in the issue itself
```
The summary of the Pull Request must have a first line matching the JIRA issue following by a colon and then the title of this issue
```
That is, the commit name and the PR name for this issue would be: `XWIKI-11269: WYSIWYG word is not consistent on different pages`
See: https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HRule:AlwaysputaJIRAissuereferenceincommitmessages

9inpachi:
@surli is right.

Also keep in mind that when you change the value in the settings, you will have to change the value anywhere where we get the text editor preference so nothing is broken.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-test/xwiki-platform-user-profile-test-docker/src/test/it/org/xwiki/user/test/ui/UserProfileIT.java
sanjithpk: > 
> 
> This is actually a piece of code dedicated to test the interface: it will looks for the String "Wysiwyg", changing it is more likely to break some tests since it won't find that String anymore.
> 
> It's more likely that the changes should take place in the templates, in xwiki-platform-web or in xwiki-platform-flamingo-skin-resources. It's also possible that some of them are in wiki pages in XML files.

So what kind of changes should I make? CSS text-transform?

surli: No here most of the work is only to locate where those strings are used in the UI and where they are defined in the code. And, instead of using each time a new version of "WYSIWYG" string, the idea is to always rely on the same translation key: that way we ensure consistency over the whole interface, and it might be change in another language by translating it.

PR #1238, opened 2020-02-26

sanjithpk:

Before:
image

After:
image

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
@@ -553,6 +553,10 @@ div.userDashboard, #dashboardPane .dashboard {
   padding: 5px 10px 10px;
 }
 
+div.userInfo {
+  border-radious: 5px;
+}
+
 /* ----- Profile ----- */
 td.recentChangesLeft, .recentChangesMoreActions {
   display:none;


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -553,6 +553,10 @@ div.userDashboard, #dashboardPane .dashboard {
   padding: 5px 10px 10px;
 }
 
+div.userInfo {
+  border-radious: 5px;
```
mflorea: Why 5px? And is this the only user profile section without round borders?

surli:
Thanks for your proposal here! The work looks good, but you made an arbitrary decision to use a 5px border radius, where as @mflorea said on the issue, this should be at least discussed and agreed on the issue. 

At least what would be good would be to introduce a less variable for the radius and to use it wherever it's needed.

vmassol:
How is it done for the "user profile menu"? How many pixels there? Is there a variable introduced there or is it hardcoded?

PS: I haven't studied the jira but the title says "Use round borders on the user profile sections to be consistent with the user profile menu" which seem to indicate it's already the case for the "user profile menu".

sanjithpk:
> 
> 
> How is it done for the "user profile menu"? How many pixels there? Is there a variable introduced there or is it hardcoded?
> 
> PS: I haven't studied the jira but the title says "Use round borders on the user profile sections to be consistent with the user profile menu" which seem to indicate it's already the case for the "user profile menu".
> 
> PPS: If it's hardcoded there too, we could also apply this PR and create a new jira to introduce a less variable.

image

I think the user was comparing it with above image and saying it was not consistent. I'm not too sure.

sanjithpk:
> 
> 
> Thanks for your proposal here! The work looks good, but you made an arbitrary decision to use a 5px border radius, where as @mflorea said on the issue, this should be at least discussed and agreed on the issue.
> 
> At least what would be good would be to introduce a less variable for the radius and to use it wherever it's needed.

Yeah I did not think of that, because I did it in XML. 

I'll try to implement it in less with a variable.

vmassol:
See https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$15827267191394IyoDM:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=nltrix.net

sanjithpk:
I will come up with a new PR where I implement round borders using the less variable @border-radius-base, so I will be closing this issue.

PR #1241, opened 2020-02-27
XWIKI-15490: Children macro not working for a document with a double quote in its name

manuelleduc:
Escape double quotes before calling documentTree.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/ChildrenMacro.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/ChildrenMacro.xml
@@ -158,7 +158,7 @@
     </class>
     <property>
       <code>{{velocity}}
-{{documentTree root="document:$doc.documentReference" /}}
+{{documentTree root="document:$doc.documentReference.toString().replaceAll('([~""])', '~$1')" /}}
 {{/velocity}}</code>
     </property>
     <property>


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/ChildrenMacro.xml
```diff
@@ -158,7 +158,7 @@
     </class>
     <property>
       <code>{{velocity}}
-{{documentTree root="document:$doc.documentReference" /}}
+{{documentTree root="document:$doc.documentReference.toString().replaceAll('([~""])', '~$1')" /}}
```
vmassol: That looks too hackish to me :)

We should introduce a rendering macro parameter escape tool if we don't have one already.

vmassol: It's also code duplication with https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Navigation.xml#L176

tmortagne: > We should introduce a rendering macro parameter escape tool if we don't have one already.

There is one in the rendering script service.

Another problem is the use of DocumentReference#toString() which is in no way an API, the reference serializer should be used instead.

PR #1245, opened 2020-02-27

sanjithpk:

Used less variable @border-radius-base for rounder borders.

Before:
image

After:
image

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
@@ -459,6 +459,15 @@ return XWiki;
     <property>
       <code>#template("colorThemeInit.vm")
 #set($tabswidth = "130px")
+
+// Variables
+// --------------------------------------------------
+
+//== Components
+//
+
+@border-radius-base:        4px;
+
 /* ----- User menu ----- */
 #user-menu-col{
   float: left;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -459,6 +459,15 @@ return XWiki;
...
       <code>#template("colorThemeInit.vm")
 #set($tabswidth = "130px")
+
+// Variables
+// --------------------------------------------------
+
+//== Components
+//
+
+@border-radius-base:        4px;
```
surli: This variable placed here is actually useless: you only use it at one place... 
Why we asked for a variable, is that it might be used at some other places that needs to be identified. So this variable shouldn't be defined here but in a .less file of the skin. Such as https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less (note: I'm not sure it's the right place, this has to be checked, we have different .less files in that directory)

Now your first work here as @mflorea specified is to ask for an agreement on the issue that we all want rounded borders of 4px. It's not obvious according to Marius' comment.

mflorea: The ``@border-radius-base`` variable is already defined, see https://github.com/xwiki/xwiki-platform/search?q=border-radius-base&unscoped_q=border-radius-base .

@sanjithpk did you test this change? Are you sure the LESS code is evaluated here?

PR #1247, opened 2020-02-27
XWIKI-16345: Use .xform style standard for the edit panels

Haxsen:
Fixes XWIKI-16345.
Link: https://jira.xwiki.org/browse/XWIKI-16345

Before:
Picture 2020-02-25 23_09_37

After:
Picture 2020-02-25 23_09_55

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
@@ -139,7 +139,7 @@
   #end
   #template("header.vm")
   #if($editor == 'wiki' || $editor == 'wysiwyg')
-    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock xform">
+    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock form xform">
     <div class="hidden">
     ## CSRF prevention
     <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
@@ -139,7 +139,7 @@
   #end
   #template("header.vm")
   #if($editor == 'wiki' || $editor == 'wysiwyg')
-    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock form xform">
+    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock xform">
     <div class="hidden">
     ## CSRF prevention
     <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
@@ -139,7 +139,7 @@
   #end
   #template("header.vm")
   #if($editor == 'wiki' || $editor == 'wysiwyg')
-    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock form">
+    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock xform">
     <div class="hidden">
     ## CSRF prevention
     <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />


9inpachi:
You mentioned earlier that adding the `xform` class changed some things in other elements.
Can you share screenshots of that?

Haxsen:
Yes here:
Picture 2020-02-27 23_45_01
(Title is no more Bold)

9inpachi:
I am not sure why adding the `xform` class causes this but this is the right solution.
Let's have some experienced XWikier take a look at this as well.

And, thanks for the PR. :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
```diff
@@ -139,7 +139,7 @@
   #end
   #template("header.vm")
   #if($editor == 'wiki' || $editor == 'wysiwyg')
-    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock form">
+    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock xform">
```
mflorea: You didn't just added the ``xform`` class, you also **removed** the ``form`` class. We may loose styles as a consequence. Did you check if there were styles associated with the ``form`` class? You can also break any JavaScript code that looks for the ``form`` class. Did you check if there is any such JavaScript code?

Haxsen: I just checked in lock.js :
// The page is automatically unlocked when the form is submitted.
    var markUnlocked = this.setLocked.bind(this, false);
    $$('form.withLock').each(function(form) {
      form.observe('submit', markUnlocked);
    });
I am not sure if this requires 'form' coming after withLock.
So just in case I added xform after 'form'. It also worked as intended.
Thanks for letting me know my mistake.

mflorea: ``form.withLock`` is indeed a CSS selector, but ``form`` is not a class name but the tag name. So this JavaScript code is not a good reason to keep the ``form`` class name.

mflorea: So to be clear: I'm not saying you should keep the ``form`` class name. I'm just saying that if you remove it you should check if there are CSS styles that we may loose or if there is some JavaScript code that you may break.

Haxsen: Ok got it! A little question: If it breaks, I will get an error or the styling would be messed up right?

Haxsen:
@mflorea Can this PR be approved by you?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
```diff
@@ -139,7 +139,7 @@
   #end
   #template("header.vm")
   #if($editor == 'wiki' || $editor == 'wysiwyg')
-    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock form">
+    <form id="edit" method="post" action="$doc.getURL('preview')" class="withLock form xform">
```
mflorea: I'd like to know why we need to keep the ``form`` class. Would be good if you can check:
* if there are CSS styles associated with it
* if there is JavaScript code that uses this CSS class

If not then we should remove it. It feels wrong to have both ``form`` and ``xform`` but we need to understand what we loose / break if we remove the ``form`` CSS class.

ClemensRobbenhaar: I looked through the git history of this file, and the `form` CSS class is present from the beginning in the `edit.vm`, but it has not been present in the `xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/edit.vm` where this template originally came from.

One can do a `find . -name '*.css' | xargs fgrep '.form'` to find if this CSS class is used in any CSS file (from within a running XWiki webapp; the source tree also has `*.less` files from which the CSS files are generated. I guess `.css.vm` files are no longer around.)
I am not sure that the same approach works for JavaScript files, however. There are different ways to get at an element which has a CSS class `form`, and just grepping for `form` will create too many hits.
Maybe the WYSIWYG editor stops working, or the Wiki editor, if this class is omitted? If it is not, I feel the CSS class is unused.

The initial commit introducing the `form` CSS class is [dcd85167489885e2017e4c56f6a4768924542927](https://github.com/xwiki/xwiki-platform/commit/dcd85167489885e2017e4c56f6a4768924542927) and has been done by @vmassol but the commit is way too large for me to see why this CSS class might have been introduced.

Haxsen: @ClemensRobbenhaar Big thanks for this!

Haxsen: Yes, the WYSIWYG and Wiki editors are working fine. I believe `form` class is not being used, a hint I took from `editclass` and `editobject` .vm files which have `withLock xform` instead of `withLock form`.
I tested `withLock xform` and `withLock form xform` , they have no difference and both work fine.

mflorea: @Haxsen OK, then remove the ``form`` CSS class.

Haxsen: Done.

Haxsen:
Thanks for the merge!!

mflorea:
Thank you for your contribution and for your patience.

PR #1253, opened 2020-03-03
 XWIKI-15490: Children macro not working for a document with a double quote in its name

manuelleduc:
Escaping parameters is now realized with a proper escape script service.


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/ChildrenMacro.xml
```diff
@@ -158,7 +158,7 @@
     </class>
     <property>
       <code>{{velocity}}
-{{documentTree root="document:$doc.documentReference" /}}
+{{documentTree root="document:$services.rendering.escape($doc.documentReference, $doc.syntax.toIdString())" /}}
```
vmassol: From a pure tech POV this is still using the toString() method and not the serializer ;)

vmassol: `$services.rendering.serialize($doc.documentReference)`

vmassol: So: `$services.rendering.escape($services.rendering.serialize($doc.documentReference), $doc.syntax.toIdString())`

vmassol:
Conflict

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/ChildrenMacro.xml
tmortagne: > $doc.syntax.toIdString()

No, the method takes a Syntax so $doc.syntax (but I still feel it's a good occasion to add a simpler to use `escape(String)`).

ClemensRobbenhaar: Is `$doc.syntax` the right thing to do here? What if the document is e.g. in markdown syntax?
We still need to escape in `xwiki/2.1` syntax here, as the macro is rendered in that syntax. I guess it should be `$wikimacro.doc.syntax` or `$xcontext.macro.doc.syntax` here instead.

vmassol: > No, the method takes a Syntax so $doc.syntax (but I still feel it's a good occasion to add a simpler to use escape(String)).

Indeed. @manuelleduc does it mean you didn't test it? :)

ClemensRobbenhaar: > > No, the method takes a Syntax so $doc.syntax (but I still feel it's a good occasion to add a simpler to use escape(String)).
> 
> Indeed. @manuelleduc does it mean you didn't test it? :)

I think because the `Syntax` class has a static `valueOf(String val)` method, velocity automatically uses this method to transform the String into a Syntax object to match the signature of the method called.

vmassol: > I think because the Syntax class has a static valueOf(String val) method, velocity automatically uses this method to transform the String into a Syntax object to match the signature of the method called.

That would be clever :) Maybe the new Velocity 2.x does this. I don't remember that it was doing that in the past but maybe I just didn't notice.

tmortagne: > Maybe the new Velocity 2.x does this.

No we have automatic parameters conversion since a very long time in Velocity and it's done by us.

> I think because the Syntax class has a static valueOf(String val) method

No it's because we have SyntaxConverter.

tmortagne: > Is $doc.syntax the right thing to do here? What if the document is e.g. in markdown syntax?

Yes `$doc` is wrong since it's the syntax of the document which is using the macro and not the macro syntax. Did not noticed this was a wiki macro.

vmassol: > No it's because we have SyntaxConverter.

ah, I thought we were only converting from String -> Syntax but I guess we also do it on the other way around (Syntax -> String).

tmortagne: Converters convert objects, not just from/to String.

manuelleduc:
```velocity
{{documentTree root="document:$services.model.serialize($doc.documentReference)" /}}
```

I have tested the macro above with markdown and xwiki pages and it seems to be fixing the issue in a more concise way.

WDYT?

vmassol: So it means the serializer will perform escaping for quotes? That seems weird to me.

tmortagne: I don't understand, this does not change anything in practice since current DocumentReference#toString() implementation is the same things as serialize. The point of the escaping was to make sure to support document reference containing things that could break the macro like `"`, serialize() does not do that.

manuelleduc: Indeed, but according to my tests it does fix the problem in practice. I'll dig deeper and let you know when I understand what's going on precisely.

vmassol:
Thanks @manuelleduc!

PR #1254, opened 2020-03-03
XWIKI-12639: Introduce extension point to add custom categories in the user profile

manuelleduc:
https://forum.xwiki.org/t/make-the-user-menu-extensible


xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -175,4 +175,146 @@
...
+      <content></content>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userdashboardsheet</name>
+    </property>
+    <property>
+      <parameters>id=dashboard&#xd;
```
surli: Did you run `mvn xar:format` on it? I'm almost sure those `&#xd;` are automatically removed.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -175,4 +175,146 @@
...
+sheet=Dashboard.XWikiUserDashboardSheet&#xd;
+glyphicon=th&#xd;
+display=#if(($services.model.resolveDocument($xcontext.user) == $doc.documentReference) &amp;&amp; $hasDashboard)true#{else}false#end</parameters>
```
surli: Could be good to add a comment on it to explain the condition.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -46,38 +46,20 @@
...
-    #set($discard = $categories.add({'id':'network', 'sheet':'XWiki.XWikiUserNetworkSheet', 'glyphicon': 'world'}))
-  #end
-  ## TODO: add an enhancement system instead
-  #set($notificationsSheet = $services.model.createDocumentReference('', ['XWiki', 'Notifications', 'Code'], 'XWikiUserNotificationsSheet'))
-  #if ($xwiki.exists($notificationsSheet))
-    #set($discard = $categories.add({'id':'notifications', 'sheet':'XWiki.Notifications.Code.XWikiUserNotificationsSheet', 'glyphicon': 'bell'}))
+
+## load the user menu from the 'org.xwiki.plaftorm.user.profile.menu.content' UIXP.
+#foreach ($uix in $services.uix.getExtensions('org.xwiki.plaftorm.user.profile.menu.content', {
+  'sortByCustomOrder': 'org.xwiki.platform.user.profile.entry.userprofilesheet,org.xwiki.platform.user.profile.entry.userpreferencessheet,org.xwiki.platform.user.profile.entry.usermembershipsheet,org.xwiki.platform.user.profile.entry.userwatchlistsheet,org.xwiki.platform.user.profile.entry.usernetworksheet,org.xwiki.platform.user.profile.entry.usernotificationsheet,org.xwiki.platform.user.profile.entry.userwikisheet,org.xwiki.platform.user.profile.entry.userdashboardsheet'
```
surli: 120 characters per line max in our checkstyle rules :) We don't enforce the style on velocity code, but still it makes it really hard to read right now.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -46,38 +46,20 @@
...
+}))
+  #if($uix.parameters.get('display') == 'true')
+    #if ($uix.parameters.containsKey('name'))
+        ## the name value is derived by default but can be overloaded with the name key.
+        #set($discard = $categories.add({'id': $uix.parameters.get('id'), 'name': $uix.parameters.get('name'), 'sheet': $uix.parameters.get('sheet'), 'glyphicon': $uix.parameters.get('glyphicon')}))
```
surli: Wouldn't it work to just do `$categories.add($uix.parameters)`? AFAICS it's already a dictionary so it should work the same way no? That would allow to remove the if above entirely too I guess.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -175,4 +175,146 @@
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content></content>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userdashboardsheet</name>
```
surli: the UIXP id ends with `profile.menu.content` and here you're using `profile.entry.xxx` it doesn't look really consistent.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/Translations.es.xml
```diff
@@ -44,5 +44,4 @@ dashboard.gadgetEditor.changeGadget.label=Cambiar gadget
 dashboard.gadgetEditor.gadgetTitle.label=Título del gadget
 dashboard.gadgetEditor.gadgetTitle.hint=El título de este gadget aparece en la parte superior del panel del gadget en el panel de control(puede contener un código Velocity).
 dashboard.initialization.saveMessage=Inicializar la configuración del usuario del tablero predeterminado</content>
-
```
surli: I guess it happened with the xar:format, don't hesitate to remove those changes from the commit next time. They are really not related.

surli:
Small note: when possible I generally try to merge pull requests with rebase, to keep the history of changes. 
Here it's not possible since you didn't use the JIRA issue key on your commit. It's not a problem we'll just squash the commits, but I guess you should start to have the reflex to put the JIRA issue key in your commits right now :)

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
manuelleduc: I've added xml comments above each parameters section with non-trivial display conditions.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
vmassol: This id is not correct since this UIX is located in the dashboard module. Thus it should start with `org.xwiki.platform.dashboard`.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
vmassol: Please don't add XML comments, we should not change the XML, it's auto-generated.

vmassol: For me a `display` parameter contains the content to display. If this is meant to be a boolean to decide if the UIX should be displayed then I's use something more explicit such as `isActive`.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -175,4 +175,147 @@
...
+    </property>
+    <property>
+      <!-- display: The user is seeing her own profile and the dashboard document exists -->
+      <parameters>id=dashboard
+sheet=Dashboard.XWikiUserDashboardSheet
+glyphicon=th
+display=#if(($services.model.resolveDocument($xcontext.user) == $doc.documentReference) &amp;&amp; $hasDashboard)true#{else}false#end</parameters>
+    </property>
+    <property>
+      <scope>global</scope>
```
vmassol: Never use the `global` scope. This requires PR rights and this means that it's now broken for the farm use case (since local admins don't have PR rights on farm wikis) :)

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/XWikiUserNotificationsSheet.xml
```diff
@@ -211,4 +211,147 @@
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.usernotificationsheet</name>
```
vmassol: Same comment as above about the UIX id.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/XWikiUserNotificationsSheet.xml
```diff
@@ -211,4 +211,147 @@
...
+    </property>
+    <property>
+      <!-- display: The user is seeing her own profile -->
+      <parameters>id=notifications
+sheet=XWiki.Notifications.Code.XWikiUserNotificationsSheet
+glyphicon=bell
```
vmassol: Why "glyph"? Why not simply `icon`?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserMembershipSheet.xml
```diff
@@ -130,4 +130,146 @@
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.usermembershipsheet</name>
```
vmassol: Same comment.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserMembershipSheet.xml
```diff
@@ -130,4 +130,146 @@
...
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
```
vmassol: I'm not sure why you have the "content" suffix. I'd simply use `org.xwiki.plaftorm.user.profile.menu` same as what we do elsewhere. Note that for `org.xwiki.plaftorm.menu.content` the `menu` part is the module name ;) See https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/#|t=uixs&p=1&l=30&s=doc.title&d=asc&doc.title=&category=menu

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserNetworkSheet.xml
```diff
@@ -521,4 +521,147 @@ require(['jquery', 'xwiki-meta'], function ($, xm) {
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.usernetworksheet</name>
```
vmassol: Same comment

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserNetworkSheet.xml
```diff
@@ -521,4 +521,147 @@ require(['jquery', 'xwiki-meta'], function ($, xm) {
...
+    </property>
+    <property>
+      <!-- display: The user is seeing her own profile and the watch service is available.-->
+      <parameters>id=network
```
vmassol: What is this `id` parameter for? Since there's already the sheet specified, I'm not sure what it brings and why you could use the sheet as id or extract an id from it (I haven't checked the code using it yet, reviewing from the top to bottom ;)). It'll certainly be clear later on in my review ;)

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -198,4 +198,147 @@
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userpreferencessheet</name>
```
vmassol: The module is good here ;) But why `entry`. Entry of what? If the UIXP is named "profileMenu" then I'd use that as prefix, as in : `org.xwiki.platform.user.profileMenu. userPreferencesSheet`.

Comment: I'd also propose to use camel case as we do for translations. I don't see why we need lowercases, makes it harder to read.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -198,4 +198,147 @@
...
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
```
vmassol: I'd do the same as translation and avoid using excessive usage of dots since dots are used to separate portions of the UIXP id. So `<module id>.<uixp name>`, with `org.xwiki.plaftorm.user` as the module id and `profileMenu`  as the uixp name. Thus `org.xwiki.plaftorm.user.profileMenu`. I guess we'll discuss this in the forum ;)

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
```diff
@@ -541,4 +541,146 @@ $services.localization.render('platform.core.profile.category.profile.disabled')
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userprofilesheet</name>
```
vmassol: Same comment as above

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -46,38 +46,23 @@
...
+
+## load the user menu from the 'org.xwiki.plaftorm.user.profile.menu.content' UIXP.
+#set($list_uix = ["org.xwiki.platform.user.profile.entry.userprofilesheet",
+  "org.xwiki.platform.user.profile.entry.userpreferencessheet",
+  "org.xwiki.platform.user.profile.entry.usermembershipsheet",
+  "org.xwiki.platform.user.profile.entry.userwatchlistsheet",
+  "org.xwiki.platform.user.profile.entry.usernetworksheet",
+  "org.xwiki.platform.user.profile.entry.usernotificationsheet",
+  "org.xwiki.platform.user.profile.entry.userwikisheet",
+  "org.xwiki.platform.user.profile.entry.userdashboardsheet"])
```
vmassol: I don't understand this static list. This defeats the concept of UIXP for me :)

xwiki-platform-core/xwiki-platform-watchlist/xwiki-platform-watchlist-ui/src/main/resources/XWiki/XWikiUserWatchListSheet.xml
```diff
@@ -120,4 +120,147 @@ $xwiki.ssx.use("XWiki.XWikiUserSheet")##
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userwatchlistsheet</name>
```
vmassol: same comment as above

xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/UserWikiSheet.xml
```diff
@@ -468,4 +468,148 @@ return XWiki;
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userwikisheet</name>
```
vmassol: same comment as above

xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/UserWikiSheet.xml
```diff
@@ -468,4 +468,148 @@ return XWiki;
...
+    </property>
+    <property>
+      <!-- display: The current wiki is the main wiki and the user sheet document exists.-->
+      <parameters>id=wikis
+name=$services.localization.render('platform.wiki.menu.userwikis')
+sheet=WikiManager.UserWikiSheet
+glyphicon=list
+display=#if($xcontext.isMainWiki() &amp;&amp; $xwiki.exists('WikiManager.UserWikiSheet'))true#{else}false#end</parameters>
```
vmassol: I don't understand the check for ` $xwiki.exists('WikiManager.UserWikiSheet'))`. BTW your XML comment above is not great since it explains the code (which is almost never useful since we can read it ;)). A good comment must explain WHY it's done.

vmassol: If the sheet doesn't exist then this UIX doesn't exist.

tmortagne: > I don't understand the check for $xwiki.exists('WikiManager.UserWikiSheet')). 

Sounds like a too quick copy/pasted from the old version. The good thing with UIX is that you don't need this kind of hack anymore.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -175,4 +175,147 @@
...
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userdashboardsheet</name>
+    </property>
+    <property>
+      <!-- display: The user is seeing her own profile and the dashboard document exists -->
+      <parameters>id=dashboard
+sheet=Dashboard.XWikiUserDashboardSheet
```
tmortagne: I don't understand the point of the "sheet" fields:
* UI extensions already have a dedicated way of providing the executed content to insert (the `content` field), you should not reinvent it in user profile side
* it force using a wiki page which is defeat the purpose of using UIX

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -175,4 +175,147 @@
...
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.plaftorm.user.profile.menu.content</extensionPointId>
+    </property>
+    <property>
+      <name>org.xwiki.platform.user.profile.entry.userdashboardsheet</name>
+    </property>
+    <property>
+      <!-- display: The user is seeing her own profile and the dashboard document exists -->
+      <parameters>id=dashboard
```
tmortagne: Is this id needed given than UI extensions already have an id ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/XWikiUserNotificationsSheet.xml
manuelleduc: Looking at the code, this key is flamingo specific (https://github.com/xwiki/xwiki-platform/blob/71ba7e3e0178d5e3f4aae52805cdf55d145dc5de/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm#L498).
It might be an good idea to change the name of the key but I'm not sure to measure the impact such change would have.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
manuelleduc: The id is used in the url (e.g., http://localhost:8080/xwiki/bin/view/XWiki/Admin?category=notifications, here the id is notifications). I think changing this would break any bookmarked link to the pages of the menu.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
manuelleduc: > Please don't add XML comments, we should not change the XML, it's auto-generated.

I tried to put the comment inline using `##` but I make the whole expression evaluation fails. What's the right place to put such comment?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserNetworkSheet.xml
manuelleduc: See https://github.com/xwiki/xwiki-platform/pull/1254#discussion_r389664987

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
tmortagne: I can understand the retro compatibility need but it should at least not be mandatory and fallback on uix extension id when not provided IMO. WDYT ?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
manuelleduc: That not a closed list, just the list of UIX we know exist and want to keep sorted in the same order are they used to be before the refactoring.
See https://extensions.xwiki.org/xwiki/bin/view/Extension/UIExtension%20Module : 
> **sortByCustomOrder**: takes a list of UI Extension IDs as parameter. The extensions will be in the same order as they've been listed, additional extensions will be put at the end of the list.

An alternative would be to introduce a `priority` key in the UIX and to use it with the **sortByParameter** filter. WDYT?

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
manuelleduc: Sounds good. I'll do this.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
vmassol: yes it's not good at all. We need the UIX to specify the order.

vmassol: The use case is that any UIX needs to be insertable at any position.

vmassol: The alternative is to provide a sort by some criteria but it's less easy since that could depend on the language. Needs to be decided. In any case it must not be hardcoded.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
vmassol: > The id is used in the url (e.g., http://localhost:8080/xwiki/bin/view/XWiki/Admin?category=notifications, here the id is notifications). I think changing this would break any bookmarked link to the pages of the menu.

I'm not sure about this. I don't consider this to be a public API.  AFAIK it's also not documented anywhere.

tmortagne: > I don't consider this to be a public API

It's not something exotic for an extension to lead to a specific well know user profile section.

manuelleduc: Preserving the id while making it optional for new UIX is quite easy (3 LoC) and would preserve existing bookmark to user menu pages.
Additionaly, I personally consider this kind of links are part of a public API and would argue to keep them.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
manuelleduc: I have introduced a `priority` property used in combination with the sortByCustomOrder parameter. The lower the priority is, the higher the UIX is in the menu.
Each existing UIX priority is a multiple of 10, that way external UIX can interleave between existing entries without having to modify "concurrent" UIX.

vmassol: Cool. Just a small detail: I would check how other UIXP have implemented priorities and check that the order is the same (lower is higher) so that we have some consistency ;)

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
manuelleduc: I have replaced the sheet fields with the UIX content.
Now I have "technical" pages with not content of their own, would it be interesting to move the UIX objects declarations somewhere else and remove the unused pages?

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
surli: > What's the right place to put such comment?

Not sure there is. Could be good to check how we parse the parameters, I guess we could improve it to avoid parsing lines starting with `##`.

vmassol: Another idea is to put the logic of "display" (isActive) inside the content and don't display anything if the evaluated content is empty.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
tmortagne: > not content of their own

That's not necessarily a problem, that's what you have in most panels and wiki macros for example (in case on panels we display the panel and macros usually use that for documentation).

You could also keep the content as it is in case someone expect it to be here and simply put an include macro in your content field for the old user profiles sheets.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
tmortagne: > Another idea is to put the logic of "display" (isActive) inside the content

It's too late since you also want to not display the button.

tmortagne: > I tried to put the comment inline using ## but I make the whole expression evaluation fails.

This might be something to fix.

manuelleduc: > Another idea is to put the logic of "display" (isActive) inside the content and don't display anything if the evaluated content is empty

I guess this is doable but it requires to pre-calculate the content of all UIX before deciding to include it or not. Also the result of `UIExtension.execute()` is a `Block` object, do a way to identify if a Block is empty exists?

So in summary I see three candidate solutions:
- @vmassol proposal 
- fixing the parsing of the parameters to ignore end of lines with `##`
- Adding a specific field in the UIX Class that would explicit this logic

tmortagne: > pre-calculate the content of all UIX before deciding to include it or not.

-1 for that, we already have an issue related to execute each section on demand instead of the current "execute all" so we should definitely not rely on it

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtensionParameters.java
```diff
@@ -123,12 +127,18 @@ public WikiUIExtensionParameters(String id, String rawParameters, ComponentManag
...
+        Properties properties = new Properties();
+        try {
+            properties.load(new StringReader(rawParameters));
+            result = new HashMap<>();
+            for (String key : properties.stringPropertyNames()) {
+                if (!Objects.equals(key, "")) {
+                    result.put(key, properties.getProperty(key));
+                }
             }
+        } catch (IOException e) {
```
surli: We shouldn't ignore silently exception: at least add a logger to inform about it.

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtensionParameters.java
```diff
@@ -123,12 +127,18 @@ public WikiUIExtensionParameters(String id, String rawParameters, ComponentManag
...
-            String[] pair = line.split("=", 2);
-            if (pair.length == 2 && !"".equals(pair[0]) && !"".equals(pair[1])) {
-                result.put(pair[0], pair[1]);
+        Map<String, String> result;
+        Properties properties = new Properties();
+        try {
+            properties.load(new StringReader(rawParameters));
+            result = new HashMap<>();
+            for (String key : properties.stringPropertyNames()) {
+                if (!Objects.equals(key, "")) {
```
surli: Why not using something such as `StringUtils.isBlank(key)` here? Looks safer.

surli: Btw Properties extends Hashtable, so it could only have one element with "" as key, no? Wouldn't that work to directly create `new HashMap<>(properties)`?

manuelleduc: Properties inherits `Map<Object, Object>` and `parseParameters` returns `Map<String,String>` so doing `new HashMap<String,String>(properties)` does not typecheck.

surli: Indeed.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
manuelleduc: Seems to be the standard behaviour across the code base, and the way `sortByParameter` works.

For another example see https://github.com/xwiki/xwiki-platform/blob/07c0927c9bf1a6a48db4ef8deb286f16d6e7c1b7/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/PanelsCode/ApplicationsPanelConfigurationSheet.xml#L71

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtensionParameters.java
```diff
@@ -133,11 +133,12 @@ public WikiUIExtensionParameters(String id, String rawParameters, ComponentManag
             properties.load(new StringReader(rawParameters));
             result = new HashMap<>();
             for (String key : properties.stringPropertyNames()) {
-                if (!Objects.equals(key, "")) {
+                if (!StringUtils.isBlank(key)) {
                     result.put(key, properties.getProperty(key));
                 }
             }
         } catch (IOException e) {
+            LOGGER.error("UIX parameters could not be read", e);
```
vmassol: A good practice is to always provide the most information possible in logging. For example here you should pass `rawParameters` IMO. See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtensionParameters.java
```diff
@@ -133,11 +133,12 @@ public WikiUIExtensionParameters(String id, String rawParameters, ComponentManag
...
             result = new HashMap<>();
```
vmassol: Couldn't this be Collections.emptyMap() instead?

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtensionParameters.java
vmassol: Why convert to Map? Why not return a Properties object? In the code we iterate over it later on so you're doing 2 iterations instead of one from what I see.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/XWikiUserNotificationsSheet.xml
manuelleduc: I've changed the key to `icon` and translate it to glyph when needed.

xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/UserWikiSheet.xml
manuelleduc: I've removed the useless condition.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
manuelleduc: I have left the content in the UIX object, do you think it is safer to move to back to its original place?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -46,38 +46,21 @@
...
-    #set($discard = $categories.add({'id':'network', 'sheet':'XWiki.XWikiUserNetworkSheet', 'glyphicon': 'world'}))
-  #end
-  ## TODO: add an enhancement system instead
-  #set($notificationsSheet = $services.model.createDocumentReference('', ['XWiki', 'Notifications', 'Code'], 'XWikiUserNotificationsSheet'))
-  #if ($xwiki.exists($notificationsSheet))
-    #set($discard = $categories.add({'id':'notifications', 'sheet':'XWiki.Notifications.Code.XWikiUserNotificationsSheet', 'glyphicon': 'bell'}))
+
+## load the user menu from the 'org.xwiki.plaftorm.user.profile.menu' UIXP.
+#foreach ($uix in $services.uix.getExtensions('org.xwiki.plaftorm.user.profile.menu', {'sortByParameter': 'priority'}))
+  #if($uix.parameters.get('isActive') == 'true')
```
surli: I'm wondering if it wouldn't be safer to do the contrary: only hide the UIX if `isActive == 'false'`. By definition, the UIX should be injected here we added a parameter to control it, but the fallback should be to view them if the parameter is missing.
I think it would be better especially if we're gonna get rid of this parameter in the future, and transform it to a UIX class parameter: this would allow to create new UIX without needed to remove it. WDYT?

manuelleduc: I find it more interesting and to be strict in the displayed entries as it is undesirable to display entries to an unauthorized user.
On the cons part, the current behaviour has more chances to lead to blocking bug.
I'm voting +0 on the change to `isActive == false`.

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
tmortagne: I don't think there is any risk no, was just a suggestion to change less stuff.

xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/UserWikiSheet.xml
```diff
@@ -468,4 +415,201 @@ return XWiki;
...
+    </property>
+    <property>
+      <name>org.xwiki.platform.wiki.ui.mainwiki.menu</name>
+    </property>
+    <property>
+      <parameters>id=wikis
+name=$services.localization.render('platform.wiki.menu.userwikis')
+icon=list
+# isActive: The current wiki is the main wiki.
+isActive=#if($xcontext.isMainWiki())true#{else}false#end
```
surli: wouldn't `isActive=$context.isMainWiki()` work?

manuelleduc: It works but makes less explicit what the intended output is.

surli: Well it's expected that `isMainWiki()` returns a boolean value, and it's easy to check it in the javadoc. Personally I find it cleaner. Moreover we should understand it with the comment above (which I don't find very clear btw :) you don't explain why it's active only if the current wiki is the main wiki).

Now it's really a small detail.

PR #1257, opened 2020-03-05
Issue XWIKI-16341 " 'Add summary' input not aligned with 'Save' buttons "

Usamaahmad751:
**Issue link:** https://jira.xwiki.org/browse/XWIKI-16341
**Issue Description:** The input is aligned with the button ONLY on the 'editor=wiki' mode. If you switch to WYSIWYG, object, class editors, the alignment is lost. There is a 1px difference.

**Before:**
Before

After:
After

**Solution:** 
Input box "Add Summary" was a little bit above the line of actions-buttons. So 1px from the top is added to keep align it with other action buttons.


mflorea:
Thank you for your contribution!

PR #1260, opened 2020-03-05
XWIKI-16342 "Unaligned buttons for class and object editors"

Usamaahmad751:
**Issue:**  XWIKI-16342 Unaligned buttons for class and object editors
**Issue link:**  https://jira.xwiki.org/browse/XWIKI-16342

**Before:** 

selectClass

addToClass

**After:**

After-Object

After-Class

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
@@ -323,7 +323,10 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
   text-decoration: underline;
   background-color: $theme.highlightColor;
 }
-
+#add_xobject .buttonwrapper, #add_xproperty .buttonwrapper{
+  position: relative;
+  top: -2px;
+}
 #switch-xclass {
   text-align: right;
   margin: 4px 0 -16px;
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
@@ -337,6 +337,7 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
 .suggest-classes.selectize-control {
   /* Display on the same line as the Add button */
   display: inline-block;
+  /* Align with the Add button */
   vertical-align: middle;
 }
 .suggest-classes > .selectize-dropdown > .selectize-dropdown-content,
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
@@ -324,22 +324,22 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
   background-color: $theme.highlightColor;
 }
 #add_xobject .buttonwrapper, #add_xproperty .buttonwrapper{
-  position: relative;
-  top: -2px;
+  vertical-align:top;
+}
+#add_xobject .buttonwrapper input , #add_xproperty .buttonwrapper input {
+    vertical-align: top;
+    margin: 0px;
 }
 #switch-xclass {
   text-align: right;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
```diff
@@ -323,7 +323,10 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
   text-decoration: underline;
   background-color: $theme.highlightColor;
 }
-
+#add_xobject .buttonwrapper, #add_xproperty .buttonwrapper{
+  position: relative;
+  top: -2px;
```
mflorea: This feels like a hack. Why ``-2`` and not ``-3``. Did you test on all [supported browsers](https://dev.xwiki.org/xwiki/bin/view/Community/SupportStrategy/BrowserSupportStrategy)? It needs at least a comment explaining how you ended up with this value (and the reason can't be "because it works"). You couldn't fix this with ``vertical-align``?

Usamaahmad751: Yes, that can also be fixed using` vertical-align: top`  for the 'Add' button. If this seems to be a better solution I will make the changes.

mflorea: Using ``vertical-align`` is better than using magic numbers.

Usamaahmad751: I have tested them on Chrome, Edge, and Firefox and it workes perfectly. 
**Result:** 
Result

**Inspect:** 

Edge-inspect

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
```diff
@@ -323,20 +323,21 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
   text-decoration: underline;
   background-color: $theme.highlightColor;
 }
-
+#add_xobject .buttonwrapper input , #add_xproperty .buttonwrapper input {
```
mflorea: The best practice is to put each selector on a separate line. The comma ``,`` is hard t spot and you can easily be mislead that it's a single long selector.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
```diff
@@ -323,20 +323,21 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
...
+    vertical-align: top;
+    margin: 0px;
```
mflorea: Why not use 2 spaces to indent, like the rest of the CSS code in this file? :) Also, when the value is ``0`` the unit doesn't matter so it can be omitted.

Usamaahmad751: The 'Add' button is given `margin: 2px 0px` on the top of this CSS file. So I prefer to make it zero rather than changing from the top. I thought to omit the `margin: 2px` may affect other components :). But let me make this cleaner.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/editors/dataeditors.css
```diff
@@ -323,20 +323,24 @@ div#xwikiobjects input[type="radio"], div#xwikiclassproperties input[type="radio
...
   background-color: $theme.highlightColor;
 }
-
+#add_xobject .buttonwrapper input {  
+  vertical-align: top;
+}
+#add_xproperty .buttonwrapper input {
+  vertical-align: top;
+  margin: 0px;
+}
```
mflorea: When I said to put the CSS selectors on separate lines I didn't meant to duplicate the styles. I simply wanted you to format the CSS code so that it's easier to read. And again, the unit is not important when the value is ``0`` so it should be omitted.

Usamaahmad751: Oh! Got it. Thanks

mflorea:

PR #1266, opened 2020-03-11
XWIKI-10954: "Panel Title" field is not aligned with "Create" button

Usamaahmad751:
**Issue**: XWIKI-10954 
**Issue link**: https://jira.xwiki.org/browse/XWIKI-10954

**Description:** Text Field and button input elements are aligned. I tested my changes on multiple browsers (Edge, Firefox, Chrome). I have solved both `Create a Job` and `Create a Panel` with a single line of code. `vertical-align: middle`.

**Before:**  

Before vertical_align
Job_before_align

**After:**

after-vertical_align

job_after_align

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/forms.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/forms.less
@@ -7,6 +7,10 @@ textarea, select, input[type="text"], input[type="password"] {
   display: inline-block;
   max-width: 100%; /* Prevent resizing beyond the visible area in modern browsers */
   width: auto;
+}
+
+input.withTip,
+input.panelinput {
   vertical-align: middle; /* Align with other form elements */
 }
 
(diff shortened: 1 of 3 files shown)


Usamaahmad751:
> Looks good, but since this change affects many form fields, can you check other forms in XWiki to see if there are any side effects. Thanks.

To avoid any side effects on other forms I made the selectors specific to `Pane; title` and `Job name` fields. I think that's a good practice to align them without having any risk for other forms.

mflorea:
After more thinking, I'm now more in favor of using the [``form-horizontal``](https://getbootstrap.com/docs/3.3/css/#forms-horizontal) class on the 2 forms.

Usamaahmad751:
@mflorea I applied ` .form-horizontal` on both of these two forms but alignment is still the same without using `vertical-align`. Are you in favor of adding both class `form-horizontal` and style `vertical-align: middle` at the same time?

mflorea:
I meant [``form-inline``](https://getbootstrap.com/docs/3.3/css/#forms-inline), sorry. But you should also:
* remove the class attribute from the input (neither ``withTip`` nor ``panelinput`` are needed anymore)
* remove the ``button`` class from the button, since it's redundant with the ``btn`` class

Usamaahmad751:
@mflorea Please have a look at the required changes. Let me know if I have made any mistakes. Thanks

mflorea:
Thanks for your contribution.

PR #1267, opened 2020-03-11
XWIKI-10952: The "Available package" information is not aligned

Usamaahmad751:
**Issue**: XWIKI-10952
**Issue link:**  https://jira.xwiki.org/browse/XWIKI-10952
**Description:**  Aligned the title and the description rows having only the icon on the left side. Tested on multiple Browsers (IE, Edge, Firefox, and chrome). Responsive across all devices. Responsiveness can be seen in the screenshots.

**Before:** 

Desktop:
Desktop_Before
Tablet : 
Tablet_Before
Mobile:
mobile_before

**After:**

Desktop:
Desktop_After
Tablet:
Tablet_720px_After
Mobile: 
mobile_350px_after

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/xlist.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/xlist.less
@@ -53,14 +53,6 @@ ul.xlist li.create a {
   padding-left: 1.5em;
 }
 
-ul.xlist li.xitem div.xitemcontainer {
-  text-indent: 0;
-}
-
-ul.xlist li.xitem div.xitemcontainer ul.xlist {
-  text-indent: 0;
-}
-
 .xspacer {
   display: none;
 }
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/xlist.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/xlist.less
@@ -54,7 +54,7 @@ ul.xlist li.create a {
 }
 
 ul.xlist li.xitem div.xitemcontainer {
-  text-indent: 1.5em;
+  text-indent: 0;
 }
 
 ul.xlist li.xitem div.xitemcontainer ul.xlist {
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/xlist.less
```diff
@@ -54,7 +54,7 @@ ul.xlist li.create a {
 }
 
 ul.xlist li.xitem div.xitemcontainer {
-  text-indent: 1.5em;
+  text-indent: 0;
```
mflorea: The ``xlist`` class is probably used in other places. Have you checked all of them to see if this change doesn't break anything? See https://github.com/xwiki/xwiki-platform/search?q=xlist&unscoped_q=xlist .

Usamaahmad751: Yes, there is an only single occurrence of  `ul.xlist li.xitem div.xitemcontainer` class in xlist.less with property ` text-indent `.

Usamaahmad751: @mflorea Thanks for reviewing. Let me know If there is still any mistake.

mflorea: > There is an only single occurrence of ``ul.xlist li.xitem div.xitemcontainer`` selector across the XWIKI that's in xlist.less with property ``text-indent``.

I'm not sure you understood. The idea is to check the places where the ``xitemcontainer`` class is used on a ``div`` inside a ``li.xitem`` inside a ``ul.xlist``. Basically to check the places that are affected by this change.

Usamaahmad751: Oh! Got it. Let me have a look.

Usamaahmad751: @mflorea I checked all the places but I didn't find any other place affected by this change. The " Available package" is the only place affected by this change.

mflorea: OK, I trust you. Now, the question is: do you need to set ``text-indent`` to ``0`` or could you simply remove the rule. It makes sense to set to ``0`` only if it inherits a different value.

Take a look at the next rule (below). It looks to me that it overwrites the inherited ``text-indent``, so if you remove the ``text-indent`` from the parent (above) then this rule is not needed anymore either. Would be good to do a quick check.

Usamaahmad751: Yes, you are right! I removed the rule `text-indent: 0` because `text-indent` was not inherited from anywhere. I also removed the next rule (below) that was also not inheriting any `text-indent`. So, basically I removed both of them because `text-indent: 0` was not needed anymore.
@mflorea Please have a look at the changes. Thanks

mflorea:

PR #1282, opened 2020-03-28
XWIKI-15814 - Messages sent to followers only appear in the Notificat…

ClemensRobbenhaar:
…ion List, not in the Dashboard "Activity Stream"

This PR seem to fix the issue, but maybe someone better acquainted with the filter preferences might want to have a look if this makes sense or should be solved differently.


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
```diff
@@ -90,6 +102,24 @@ public void handleUsersParameter(String users, NotificationParameters parameters
...
             addFilterPreference(parameters, userList);
+        } else if (parameters.user != null) {
+            // if we have a user (but no "users") then we should also display personal messages from followed users.
+            // the other types of messages get included, but for personal messages the filter needs
+            // a matching filter preference so we loop though preferences to see if they have
+            // a preference for this (using a copy to guard against unwanted modifications)
+            try {
+                for (NotificationFilterPreference filterPref
+                    : notificationFilterPreferenceManager.getFilterPreferences(parameters.user)) {
+                    if (EventUserFilter.FILTER_NAME.equals(filterPref.getFilterName())) {
```
surli, line 113: I think that's correct by would be good if you could give a second opinion @manuelleduc since you worked on it recently.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
```diff
@@ -90,6 +102,24 @@ public void handleUsersParameter(String users, NotificationParameters parameters
             parameters.filters.add(new FollowedUserOnlyEventFilter(entityReferenceSerializer, userList));
 
             addFilterPreference(parameters, userList);
+        } else if (parameters.user != null) {
+            // if we have a user (but no "users") then we should also display personal messages from followed users.
+            // the other types of messages get included, but for personal messages the filter needs
+            // a matching filter preference so we loop though preferences to see if they have
+            // a preference for this (using a copy to guard against unwanted modifications)
+            try {
+                for (NotificationFilterPreference filterPref
```
surli, line 111: I think you could break the for whenever the filter is found. Would be a small perf improvment.

surli:
@ClemensRobbenhaar you're still missing a test for this one, we have a UsersParameterHandlerTest I'd be really more comfortable to see that covered.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
```diff
@@ -90,6 +102,24 @@ public void handleUsersParameter(String users, NotificationParameters parameters
             parameters.filters.add(new FollowedUserOnlyEventFilter(entityReferenceSerializer, userList));
 
             addFilterPreference(parameters, userList);
+        } else if (parameters.user != null) {
+            // if we have a user (but no "users") then we should also display personal messages from followed users.
+            // the other types of messages get included, but for personal messages the filter needs
+            // a matching filter preference so we loop though preferences to see if they have
+            // a preference for this (using a copy to guard against unwanted modifications)
```
manuelleduc, line 109: Minor comment, sentences begin with a capital letter and end with a period.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
manuelleduc: I did not worked on this aspect of activity stream, but taking into account the current user's preferences when filtering the activity stream seems to be a good solution.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
manuelleduc: I don't think that would work, since IIUC, the goal is to copy the filters for each followed users. Adding a break would only add a filter for the first followed user found, and ignore all the others.

surli: >  the goal is to copy the filters for each followed users

no, you only get the filter for the user in `parameters.user`: that's the filter job to check if the message comes from the followed users.

manuelleduc: I agree with the idea but that's not how it is working currently. If `U1` follows `U2` and `U3`, `notificationFilterPreferenceManager.getFilterPreferences(parameters.user)` will return 2 filterers with a filter name equals to `EventUserFilter.FILTER_NAME`, where the owner is U1 and the user is the followed user.

AFAIK we don't have a generic "followed users" filter.

surli: hmmm right. So indeed what I proposed was a bad idea. I guess it would deserve a comment.

ClemensRobbenhaar:
> @ClemensRobbenhaar you're still missing a test for this one, we have a UsersParameterHandlerTest I'd be really more comfortable to see that covered.

I added a new test case to the `UsersParameterHandlerTest` which checks that only the wanted `NotificationFilterPreference`s are added; however I am not sure if this really tests enough.
I have checked in a local test instance that indeed one gets one filter preference object for every followed user, so testing that only these get through is maybe all that a unit test can do.

Maybe I should undo my change to the `pom.xml` (increased the test coverage by one tick), as this causes conflicts on merge, as I see only now.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
ClemensRobbenhaar: Thanks, fixed.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/UsersParameterHandler.java
ClemensRobbenhaar: I added another sentence to the comment block above. I hope that clarifies things.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/test/java/org/xwiki/notifications/sources/internal/UsersParameterHandlerTest.java
```diff
@@ -119,6 +124,46 @@ public void handlerUsersParameter()
         assertEquals(expectedParameters, notificationParameters);
     }
 
+    @Test
+    public void handlerUsersParameterWhenOnlyTheUserParameterIsGiven() throws NotificationException
```
surli, line 128: note that the best practice for test is now to not use `public`.

surli:
> Maybe I should undo my change to the pom.xml (increased the test coverage by one tick), as this causes conflicts on merge, as I see only now.

I took the liberty to solve the conflict directly on Github.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/test/java/org/xwiki/notifications/sources/internal/UsersParameterHandlerTest.java
ClemensRobbenhaar: ok, updated.

surli:
thanks!

PR #1288, opened 2020-04-06
XWIKI-16303: Make utf8mb4 the default in XWiki Debian packages.

gdelhumeau:
I'm asking a review because these new settings are the ones that I have tested with success on a virtual machine, but I have not tried to build the debian package nor to upgrade an existing installation with the new packages. You could see it as a WIP with a request to point me where I should improve my proposition.

Thank you!

Guillaume

### DIFF
--- a/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/hibernate.cfg.xml.vm
+++ b/xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/hibernate.cfg.xml.vm
@@ -160,7 +160,7 @@
     <property name="hibernate.dbcp.poolPreparedStatements">true</property>
     <property name="hibernate.dbcp.maxOpenPreparedStatements">20</property>
 
-    <property name="hibernate.connection.charSet">UTF-8</property>
+    <property name="hibernate.connection.charSet">utf8mb4</property>
     <property name="hibernate.connection.useUnicode">true</property>
     <property name="hibernate.connection.characterEncoding">utf8</property>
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/hibernate.cfg.xml.vm
```diff
@@ -160,7 +160,7 @@
     <property name="hibernate.dbcp.poolPreparedStatements">true</property>
     <property name="hibernate.dbcp.maxOpenPreparedStatements">20</property>
 
-    <property name="hibernate.connection.charSet">UTF-8</property>
+    <property name="hibernate.connection.charSet">utf8mb4</property>
```
tmortagne, line 163: Not that this change has no impact. This is just a comment. Also from what I tested even if it was in the right place it's not really needed as in both case the connector send the same thing (a String in UTF8), the difference is only on server side.

tmortagne:
> mysql used to have a limit on the number of bytes (not characters) that can be used as a table index, and there are some indexes that would surpass that limit when using 4 bytes instead of 3 for varchar(255) columns.

The limit has been increased a lot in 5.7 so it's not a problem anymore.

xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/hibernate.cfg.xml.vm
gdelhumeau: So if I rollback this change, it would ok for you to merge?

tmortagne: Sure, looks good to me.

tmortagne: We need to decide if we want that in 11.10.x branch too but probably since it's close to a bug.

gdelhumeau: yes please ! (a MySQL bug actually...)

sdumitriu:
> The limit has been increased a lot in 5.7 (3072 instead of 767) so it's not a problem anymore.

Glad to hear it!

PR #1289, opened 2020-04-06
XWIKI-6721: The include macro should have an option to filter out the first heading in the included document or section

Usamaahmad751:
Issue link: https://jira.xwiki.org/browse/XWIKI-6721
After:

Capture

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -222,7 +222,8 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
                 result.removeBlock(heading);
             } else {
                 String includeType = parameters.getSection() != null ? "section" : "document";
-                throw new MacroExecutionException(String.format("The included " + includeType + " doesn't contain any heading"));
+                throw new MacroExecutionException(
+                    String.format("The included [%s] doesn't contain any heading",includeType));
            }
         }
         
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -215,13 +215,13 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
         }
         
         // Exclude the heading from the content
-        if (parameters.isExcludeHeading()) {
+        if (parameters.excludeHeading()) {
             HeaderBlock heading =
                     result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
             if (heading != null) {
                 result.removeBlock(heading);
             } else {
-                throw new MacroExecutionException("The included section doesn't contain main heading");
+                throw new MacroExecutionException("The included section doesn't contain any heading");
             }
         }
         
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -41,6 +41,7 @@ import org.xwiki.properties.BeanManager;
 import org.xwiki.properties.PropertyException;
 import org.xwiki.rendering.block.Block;
 import org.xwiki.rendering.block.MacroBlock;
+import org.xwiki.rendering.block.HeaderBlock;
 import org.xwiki.rendering.block.MacroMarkerBlock;
 import org.xwiki.rendering.block.MetaDataBlock;
 import org.xwiki.rendering.block.XDOM;
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -216,15 +216,15 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
         
         // Exclude the first heading from the content
         if (parameters.excludeFirstHeading()) {
-            HeaderBlock heading =
-                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            HeaderBlock heading = result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class),
+                    Block.Axes.DESCENDANT_OR_SELF);
             if (heading != null) {
-                result.removeBlock(heading);
+                removeBlockRecursively(result, heading);
             } else {
                 String includeType = parameters.getSection() != null ? "section" : "document";
                 throw new MacroExecutionException(
-                    String.format("The included [%s] doesn't contain any heading",includeType));
-           }
+                        String.format("The included [%s] doesn't contain any heading", includeType));
+            }
         }
         
         // Step 5: Wrap Blocks in a MetaDataBlock with the "source" meta data specified so that we know from where the
@@ -238,7 +238,26 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
 
         return Arrays.asList(metadata);
     }
-
+    
+    private boolean removeBlockRecursively(Block block, Block blockToRemove)
+    {
+        // Go through all children and remove the block if found
+        List<Block> allChildren = block.getChildren();
+        for (Block child : allChildren) {
+            if (child == blockToRemove) {
+                block.removeBlock(child);
+                return true;
+            } else {
+                // Check in children of the child
+                if (removeBlockRecursively(child, blockToRemove)) {
+                    // Return without moving on if child found
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+    
     /**
      * Protect form recursive inclusion.
      * 
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -221,7 +221,7 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
             if (heading != null) {
                 result.removeBlock(heading);
             } else {
-                throw new MacroExecutionException("The included section doesn't contain any heading");
+                throw new MacroExecutionException("The included document doesn't contain any heading");
             }
         }
         
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -215,13 +215,13 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
         }
         
         // Exclude the first heading from the content
-        if (parameters.excludeHeading()) {
+        if (parameters.excludeFirstHeading()) {
             HeaderBlock heading =
                     result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
             if (heading != null) {
                 result.removeBlock(heading);
             } else {
-                throw new MacroExecutionException("The included document doesn't contain any heading");
+                throw new MacroExecutionException("The included document or section doesn't contain first heading");
             }
         }
         
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -221,7 +221,7 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
             if (heading != null) {
                 result.removeBlock(heading);
             } else {
-                throw new MacroExecutionException("The included document or section doesn't contain first heading");
+                throw new MacroExecutionException("The included document or section doesn't contain any heading");
             }
         }
         
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -221,8 +221,9 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
             if (heading != null) {
                 result.removeBlock(heading);
             } else {
-                throw new MacroExecutionException("The included document or section doesn't contain any heading");
-            }
+                String includeType = parameters.getSection() != null ? "section" : "document";
+                throw new MacroExecutionException(String.format("The included " + includeType + " doesn't contain any heading"));
+           }
         }
         
         // Step 5: Wrap Blocks in a MetaDataBlock with the "source" meta data specified so that we know from where the
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -214,7 +214,7 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
             }
         }
         
-        // Exclude the heading from the content
+        // Exclude the first heading from the content
         if (parameters.excludeHeading()) {
             HeaderBlock heading =
                     result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -212,7 +212,18 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
                 references.pop();
             }
         }
-
+        
+        // Exclude the heading from the content
+        if (parameters.isExcludeHeading()) {
+            HeaderBlock heading =
+                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            if (heading != null) {
+                result.removeBlock(heading);
+            } else {
+                throw new MacroExecutionException("The included section doesn't contain main heading");
+            }
+        }
+        
         // Step 5: Wrap Blocks in a MetaDataBlock with the "source" meta data specified so that we know from where the
         // content comes and "base" meta data so that reference are properly resolved
         MetaDataBlock metadata = new MetaDataBlock(result.getChildren(), result.getMetaData());
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
@@ -204,13 +204,26 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
             }
         }
         
-        // Exclude the first heading from the content
         if (parameters.excludeFirstHeading()) {
-            HeaderBlock heading = result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class),
-                    Block.Axes.DESCENDANT_OR_SELF);
-            if (heading != null) {
-                removeBlockRecursively(result, heading);
-            } else {
+            List<HeaderBlock> headingList = result.getBlocks(new ClassBlockMatcher(HeaderBlock.class),
+                    Block.Axes.DESCENDANT);
+
+            if (!headingList.isEmpty()) {
+
+                // Get the First Heading Block from list of Heading Blocks
+                HeaderBlock heading = headingList.get(0);
+
+                // Get the parent block in document
+                Block parent = result.getChildren().get(0);
+
+                // Get the first child Block of the document
+                Block child = parent.getChildren().get(0);
+                if (child == heading) {
+                    parent.removeBlock(child);
+                }
+            }
+              
+            else {
                 String includeType = parameters.getSection() != null ? "section" : "document";
                 throw new MacroExecutionException(
                         String.format("The included [%s] doesn't contain any heading", includeType));
@@ -227,25 +240,6 @@ public class IncludeMacro extends AbstractMacro<IncludeMacroParameters>
         }
 
         return Arrays.asList(metadata);
-    }
-    
-    private boolean removeBlockRecursively(Block block, Block blockToRemove)
-    {
-        // Go through all children and remove the block if found
-        List<Block> allChildren = block.getChildren();
-        for (Block child : allChildren) {
-            if (child == blockToRemove) {
-                block.removeBlock(child);
-                return true;
-            } else {
-                // Check in children of the child
-                if (removeBlockRecursively(child, blockToRemove)) {
-                    // Return without moving on if child found
-                    return true;
-                }
-            }
-        }
-        return false;
     }
     
     /**
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -161,7 +161,7 @@ public class IncludeMacroParameters
      * exclude the heading from the section if true.
      * 
      * @param excludeHeading {@code true} whether the heading from the section is
-     *                       excluded, {@code false} otherwise
+     *        excluded, {@code false} otherwise
      */
     @PropertyName("Exclude Heading")
     @PropertyDescription("Exclude any top-level heading from the included section")
@@ -171,6 +171,7 @@ public class IncludeMacroParameters
     
     /**
      * @return {@code true} exclude the heading, {@code false} otherwise
+     *          include the heading
      */
     public boolean isExcludeHeading() {
         return excludeHeading;
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -166,18 +166,19 @@ public class IncludeMacroParameters
     @Unstable
     @PropertyName("Exclude Heading")
     @PropertyGroup("Section")
-    @PropertyDescription("Exclude the heading from the included section.")
-    public void setExcludeHeading(boolean excludeHeading) {
+    @PropertyDescription("Exclude the first heading from the included section.")
+    public void setExcludeHeading(boolean excludeHeading)
+    {
         this.excludeHeading = excludeHeading;
     }
     
     /**
-     * @return exclude the first heading from the included section. If the section 
-     *         doesn't contain the first heading nothing will be excluded.
+     * @return whether the user want to exclude the section headline.
      * @since 12.3RC1 
      */
     @Unstable
-    public boolean excludeHeading() {
+    public boolean excludeHeading()
+    {
         return this.excludeHeading;
     }
     
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -158,10 +158,10 @@ public class IncludeMacroParameters
     }
     
     /**
-     * exclude the heading from the section if true.
+     * Exclude the heading from the section if true.
      * 
      * @param excludeHeading {@code true} whether the heading from the section is
-     *        excluded, {@code false} otherwise
+     *        excluded, {@code false} otherwise 
      */
     @PropertyName("Exclude Heading")
     @PropertyDescription("Exclude any top-level heading from the included section")
@@ -170,8 +170,7 @@ public class IncludeMacroParameters
     }
     
     /**
-     * @return {@code true} exclude the heading, {@code false} otherwise
-     *          include the heading
+     * @return exclude the heading from the section
      */
     public boolean isExcludeHeading() {
         return this.excludeHeading;
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -71,7 +71,7 @@ public class IncludeMacroParameters
     /**
      * @see #excludeHeading()
      */
-    private boolean excludeHeading;
+    private boolean excludeFirstHeading;
     
     /**
      * @see #getSection()
@@ -160,25 +160,25 @@ public class IncludeMacroParameters
     
     /**
      * @param excludeHeading {@code true} to remove the first heading found inside
-     *         the document, {@code false} to keep it 
+     *         the document or a section, {@code false} to keep it 
      * @since 12.3RC1 
      */
     @Unstable
-    @PropertyName("Exclude Heading")
-    @PropertyDescription("Exclude the first heading from the included document/section.")
-    public void setExcludeHeading(boolean excludeHeading)
+    @PropertyName("Exclude First Heading")
+    @PropertyDescription("Exclude the first heading from the included document or section.")
+    public void setExcludeFirstHeading(boolean excludeFirstHeading)
     {
-        this.excludeHeading = excludeHeading;
+        this.excludeFirstHeading = excludeFirstHeading;
     }
     
     /**
-     * @return whether the user want to exclude the document/section headline.
+     * @return whether the user wants to exclude the first heading from the included document or section.
      * @since 12.3RC1 
      */
     @Unstable
-    public boolean excludeHeading()
+    public boolean excludeFirstHeading()
     {
-        return this.excludeHeading;
+        return this.excludeFirstHeading;
     }
     
     /**
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -160,20 +160,19 @@ public class IncludeMacroParameters
     
     /**
      * @param excludeHeading {@code true} to remove the first heading found inside
-     *         the specified section, {@code false} to keep it 
+     *         the document, {@code false} to keep it 
      * @since 12.3RC1 
      */
     @Unstable
     @PropertyName("Exclude Heading")
-    @PropertyGroup("Section")
-    @PropertyDescription("Exclude the first heading from the included section.")
+    @PropertyDescription("Exclude the first heading from the included document/section.")
     public void setExcludeHeading(boolean excludeHeading)
     {
         this.excludeHeading = excludeHeading;
     }
     
     /**
-     * @return whether the user want to exclude the section headline.
+     * @return whether the user want to exclude the document/section headline.
      * @since 12.3RC1 
      */
     @Unstable
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
@@ -69,7 +69,7 @@ public class IncludeMacroParameters
     private Context context = Context.CURRENT;
      
     /**
-     * @see #excludeHeading()
+     * @see #excludeFirstHeading()
      */
     private boolean excludeFirstHeading;
     
@@ -159,8 +159,8 @@ public class IncludeMacroParameters
     }
     
     /**
-     * @param excludeHeading {@code true} to remove the first heading found inside
-     *         the document or a section, {@code false} to keep it 
+     * @param excludeFirstHeading {@code true} to remove the first heading found inside
+     *        the document or the section, {@code false} to keep it 
      * @since 12.3RC1 
      */
     @Unstable
@@ -172,7 +172,7 @@ public class IncludeMacroParameters
     }
     
     /**
-     * @return whether the user wants to exclude the first heading from the included document or section.
+     * @return whether to exclude the first heading from the included document or section, or not.
      * @since 12.3RC1 
      */
     @Unstable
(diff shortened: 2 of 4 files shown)


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -212,7 +213,18 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
...
-
+        
+        // Exclude the heading from the content
+        if (parameters.isExcludeHeading()) {
+            HeaderBlock heading =
+                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            if (heading != null) {
+                result.removeBlock(heading);
+            } else {
+                throw new MacroExecutionException("The included section doesn't contain main heading");
```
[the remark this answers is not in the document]
Usamaahmad751: Actually, I am referring only to the first heading found in the included section. So what about If I write "The included section doesn't found any top-level heading". In this code, I am not removing all headings found in the section as the section can contain subheadings.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
     {
         return this.section;
     }
-
+    
+    /**
+     * Exclude the heading from the section if true.
+     * 
+     * @param excludeHeading {@code true} whether the heading from the section is
+     *        excluded, {@code false} otherwise
```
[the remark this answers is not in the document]
Usamaahmad751: Let me correct this. Thanks

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
...
+    /**
+     * Exclude the heading from the section if true.
+     * 
+     * @param excludeHeading {@code true} whether the heading from the section is
+     *        excluded, {@code false} otherwise 
+     * @since 12.3RC1 
+     */
+    @Unstable
+    @PropertyName("Exclude Heading")
+    @PropertyDescription("Exclude any top-level heading from the included section")
```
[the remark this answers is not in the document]
Usamaahmad751: Yes, I am removing the first heading found inside an included section. Let me correct this to somethings more precise.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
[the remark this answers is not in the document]
vmassol: WDYM by "section"?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,32 @@ public String getSection()
...
+     * Exclude the heading from the section if true.
+     * 
+     * @param excludeHeading {@code true} to remove the first heading found inside
+     *         the specified section, {@code false} to keep it 
+     * @since 12.3RC1 
+     */
+    @Unstable
+    @PropertyName("Exclude Heading")
+    @PropertyDescription("Exclude the first heading from the included section."
+        + " If the section doesn't have first block as heading, simply ignore it")
```
vmassol, line 171: doesn't match the code. Why did you change that?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,32 @@ public String getSection()
...
+    public void setExcludeHeading(boolean excludeHeading) {
```
vmassol, line 172: bad coding style

Usamaahmad751:
> @Usamaahmad751 I think you didn't address my comment about the grouping of section and this new parameter. See my previous review comments.
> 
> Thanks

Section Group Added.
IMO making the exclude parameter advance is a good idea, the user will not found any complexity in including the document. And If he thinks he needs to remove the heading he can come back and check the options.

vmassol:
You need to read all my comments please. You're actually doing the opposite of what I was suggesting ;)

Note: I'll stop reviewing FTM to let you address all comments. Just let me know when you're fully done and then I'll have a look. Thx

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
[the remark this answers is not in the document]
Usamaahmad751: I am referring to the included content as section. What we're doing is removing the "title heading" from the included section.
That is to say, some included sections might have a title that the user doesn't want when including the section.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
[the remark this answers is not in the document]
Usamaahmad751: Done :)

Usamaahmad751:
> @Usamaahmad751 I think you didn't address my comment about the grouping of section and this new parameter. See my previous review comments.
> 
> Thanks

I grouped them inside a "Section" Group. I have updated the screenshot

Usamaahmad751:
> > I think you didn't address my comment about the grouping of section and this new parameter. See my previous review comments.
> 
> Actually, I don't see why you would only want to exclude the first heading when the user is including a section. Could you explain? Thanks

It's because when a user adds a section including a title or headline in some cases there will be double-heading for included section(please see attachment). So to get rid of the section headline he can simply check the "exclude heading" parameter to avoid double-heading.

Usamaahmad751:
@vmassol: Please review it now. I have made changes according to your suggestions. Please correct me if I am wrong at any point. Thank you :)

vmassol:
Yes that's true but it can also be true if you don't use a section, i.e. if you include a whole document inside a location. This is what I was mentioning in a past comment! For example:

Doc1:

```
= doc1 heading =
whatever

```

Doc2:

```
= doc2 heading =
{{include reference="Doc1"/}}
```

Will generate:

```
= doc2 heading =
= doc1 heading =
whatever
```

And you may want to be able to use:

```
= doc2 heading =
{{include reference="Doc1" excludeHeading='true'/}}
```

So that you get:

```
= doc2 heading =
whatever
```

So I'll repeat my question: why only have the `excludeHeading` parameter be valid when there's a `section` specified? Shouldn't it work in all cases?

Usamaahmad751:
Yes, It will work in all cases. If the document/page is included it will remove the document/page headline. And if the section is included It will exclude the section headline. 
So It will work in both cases:

```
{{include reference="Doc1" excludeHeading='true' /}}
{{include reference="Doc1" excludeHeading='true' section ="Hsection1" /}}
```

Usamaahmad751:
> @Usamaahmad751 I think you didn't address my comment about the grouping of section and this new parameter. See my previous review comments.
> 
> Thanks

So, as I said the purpose of this parameter is to exclude the heading either from whole document or a section. So what about removing it from the section group? As it's not only for the section but for the whole included document. Thanks

vmassol:
> So, as I said the purpose of this parameter is to exclude the heading either from whole document or a section

This is what I was suggesting before you started adding inside the section group ;)

> So what about removing it from the section group? As it's not only for the section but for the whole included document. Thanks

Exactly, that was my point...

But you also need to change the text and code since you indicated "section" a bit everywhere, and that doesn't make sense anymore.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,31 @@ public String getSection()
     {
         return this.section;
     }
-
+    
+    /**
+     * @param excludeHeading {@code true} to remove the first heading found inside
+     *         the specified section, {@code false} to keep it
```
vmassol, line 163: to be fixed

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,31 @@ public String getSection()
...
+    
+    /**
+     * @param excludeHeading {@code true} to remove the first heading found inside
+     *         the specified section, {@code false} to keep it 
+     * @since 12.3RC1 
+     */
+    @Unstable
+    @PropertyName("Exclude Heading")
+    @PropertyGroup("Section")
+    @PropertyDescription("Exclude the first heading from the included section.")
```
vmassol, line 169: to be fixed

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,31 @@ public String getSection()
...
+    public void setExcludeHeading(boolean excludeHeading)
+    {
+        this.excludeHeading = excludeHeading;
+    }
+    
+    /**
+     * @return whether the user want to exclude the section headline.
```
vmassol, line 176: to be fixed

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,31 @@ public String getSection()
...
-
+    
+    /**
+     * @param excludeHeading {@code true} to remove the first heading found inside
+     *         the specified section, {@code false} to keep it 
+     * @since 12.3RC1 
+     */
+    @Unstable
+    @PropertyName("Exclude Heading")
+    @PropertyGroup("Section")
```
vmassol, line 168: to be fixed

vmassol:
@Usamaahmad751 I've updated the title for https://jira.xwiki.org/browse/XWIKI-6721 Maybe you could update the commit message for this PR too?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
...
-
+    
+    /**
+     * @param excludeHeading {@code true} to remove the first heading found inside
+     *         the document, {@code false} to keep it 
+     * @since 12.3RC1 
+     */
+    @Unstable
+    @PropertyName("Exclude Heading")
+    @PropertyDescription("Exclude the first heading from the included document/section.")
```
vmassol, line 168: suggestion: "Exclude the first heading from the included document or section."

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
...
+    public void setExcludeHeading(boolean excludeHeading)
+    {
+        this.excludeHeading = excludeHeading;
+    }
+    
+    /**
+     * @return whether the user want to exclude the document/section headline.
```
vmassol, line 175: Same

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
...
+     * @since 12.3RC1 
+     */
+    @Unstable
+    public boolean excludeHeading()
```
vmassol, line 179: I wonder if the parameter should be `excludeHeading` or `excludeFirstHeading` or `skipFirstHeading` or `ignoreFirstHeading`. Any preference? 

I'm fine with any but I think it would be better to include the word "first" in the variable name.

Usamaahmad751: Yeah, that sounds great, I'll prefer to replace it with `excludeFirstHeading`. As we are excluding the first heading.  Thanks for the suggestions :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -212,7 +213,18 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
                 references.pop();
             }
         }
-
+        
+        // Exclude the heading from the content
```
vmassol, line 217: -> "// Exclude the first heading from the content"

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
vmassol, line 175: -> "@return whether the user wants to exclude the first heading from the included document or section"

Usamaahmad751:
@vmassol: You changed the title to `The include macro should have an option to filter out the first section in the included document or section`
It shouldn't be `first heading` in place of `first section` ?

vmassol:
> it shouldn't be first heading in place of first section ?

Fixed, thanks

Usamaahmad751:
@vmassol: I have made the changes. Thanks for suggestions :) 
Here is the final output parameter:

Capture

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -212,7 +213,18 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
...
-
+        
+        // Exclude the first heading from the content
+        if (parameters.excludeFirstHeading()) {
+            HeaderBlock heading =
+                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            if (heading != null) {
+                result.removeBlock(heading);
+            } else {
+                throw new MacroExecutionException("The included document or section doesn't contain first heading");
```
vmassol, line 224: This shouldn't have been changed :) The sentence is not valid English now... grr... "first" --> "any"

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -60,13 +61,18 @@
...
+    
     /**
      * Defines whether the included page is executed in its separated execution context or whether it's executed in the
      * context of the current page.
      */
     private Context context = Context.CURRENT;
-
+     
+    /**
+     * @see #excludeHeading()
```
vmassol, line 72: wrong javadoc

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
     {
         return this.section;
     }
-
+    
+    /**
+     * @param excludeHeading {@code true} to remove the first heading found inside
```
vmassol, line 162: wrong javadoc

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
...
+     *         the document or a section, {@code false} to keep it
```
vmassol, line 163: if you say "the document" then you need to say "the section" too...  You need consistency.

--> "... found inside the included document or section..."

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/macro/include/IncludeMacroParameters.java
```diff
@@ -151,7 +157,30 @@ public String getSection()
...
+    @Unstable
+    @PropertyName("Exclude First Heading")
+    @PropertyDescription("Exclude the first heading from the included document or section.")
+    public void setExcludeFirstHeading(boolean excludeFirstHeading)
+    {
+        this.excludeFirstHeading = excludeFirstHeading;
+    }
+    
+    /**
+     * @return whether the user wants to exclude the first heading from the included document or section.
```
vmassol, line 175: --> "whether to exclude the first heading from the included document or section, or not."

vmassol:
@Usamaahmad751  did you run the build with the quality profile? `-Pquality`. I think you're missing a unit test for it.

@Usamaahmad751 This issue was quite simple. But count the number of times I've had to review or comment. It's huge. You need to improve your coding level because imagine what will happen if you need to code something a bit more complex.

vmassol:
Sorry for being a bit harsh with you and thanks for sticking with me here. I’m trying to evaluate if you can be ready for the gsoc for xwiki and thus I’m pushing you a bit. We have plenty of rules on the xwiki project and are trying to have a high quality level. So that can be seen as a pain. OTOH it should bring you some interesting knowledge in software engineering that could be useful in your career (I hope ;)).

Usamaahmad751:
> @Usamaahmad751 did you run the build with the quality profile? `-Pquality`. I think you're missing a unit test for it.

Yes, I build it using `-Pquality`. It builds successfully. I have attached a screenshot after building using using `mvn clean install -Pquality`.
 
Capture

vmassol:
Ok cool. Would be good to check the existing reste and see if you could add a test.

Usamaahmad751:
> Sorry for being a bit harsh with you and thanks for sticking with me here. I’m trying to evaluate if you can be ready for the gsoc for xwiki and thus I’m pushing you a bit. We have plenty of rules on the xwiki project and are trying to have a high quality level. So that can be seen as a pain. OTOH it should bring you some interesting knowledge in software engineering that could be useful in your career (I hope ;)).

I really apologize for too many revisions. I know it is too irritating to correct me for each and everything. But I have learned from these mistakes and will try my best to minimize the mistakes in further issues. I was not well familiar with JavaDoc so I made irresponsible mistakes. I really appreciate your cooperation. I will boost my knowledge about rules for maintaining great quality for XWiki. Thank you :)

Usamaahmad751:
> Ok cool. Would be good to check the existing reste and see if you could add a test.
Sure, Let me understand the existing tests. I will try my best to add the test for my added code. Thank you

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -212,7 +213,18 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
...
-
+        
+        // Exclude the first heading from the content
+        if (parameters.excludeFirstHeading()) {
+            HeaderBlock heading =
+                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            if (heading != null) {
+                result.removeBlock(heading);
+            } else {
+                throw new MacroExecutionException("The included document or section doesn't contain any heading");
```
vmassol, line 224: Actually this message could be improved a bit since we know if the user is including a section or not.

Usamaahmad751: I tried generating this message dynamically using this approach:
```
                String includeType = parameters.getSection() != null ? "section" : "document";
                throw new MacroExecutionException("The included " + includeType + " doesn't contain any heading"); 
```
But I got mojojo check: `Coverage checks have not been met.` error. And If I use simple if else it builds successfully.

Usamaahmad751: I think the instructions not covered completely and due to some test rules, I am getting this error. Please would you give me some suggestions?

vmassol: suggestion: search a bit and try to understand what's going on. It's very simple. BTW you should always do your homework first and try finding the info for several hours before asking, as otherwise you won't learn (when someone tells you the answer you learn very little in practice, what's important is that you can find the info by yourself, this is how you really progress). Especially here where you need less than 5 minutes to find the answser...

vmassol: Also please use `String.format()` instead for ` throw new MacroExecutionException("The included " + includeType + " doesn't contain any heading"); `.

Usamaahmad751: I totally agree with you!. Sure, I will find the solution by searching for this error. Thanks for the motivation :).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -212,7 +213,19 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
...
+        
+        // Exclude the first heading from the content
+        if (parameters.excludeFirstHeading()) {
+            HeaderBlock heading =
+                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            if (heading != null) {
+                result.removeBlock(heading);
+            } else {
+                String includeType = parameters.getSection() != null ? "section" : "document";
+                throw new MacroExecutionException(String.format("The included " + includeType + " doesn't contain any heading"));
```
vmassol, line 225: nope :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
```diff
@@ -412,6 +412,61 @@ public void testIncludeMacroInsideSourceMetaDataBlockAndWithRelativeDocumentRefe
 
         assertBlocks(expected, blocks, this.rendererFactory);
     }
+    
+    @Unstable
```
vmassol, line 416: Why?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
```diff
@@ -412,6 +412,61 @@ public void testIncludeMacroInsideSourceMetaDataBlockAndWithRelativeDocumentRefe
...
+                will(returnValue(getXDOM("content")));
+                allowing(mockDocument).getRealLanguage();
+                will(returnValue(""));
+            }
+        });
+        try {
+            List<Block> blocks = this.includeMacro.execute(parameters, null, macroContext);
+            assertBlocks(expected, blocks, this.rendererFactory);
+        } catch (MacroExecutionException e) {
+            String includeType = parameters.getSection()!=null ? "section" : "document";
```
vmassol, line 465: Never use a try/catch not supposed to test for an exception condition! Your test is either non-deterministic or it contains code that is never executed!

vmassol: I hope you're were not thinking that this test was testing the two cases.... :)

vmassol: I really don't see what this test is testing, can you explain?

vmassol: It really seems like you copy pasted some existing test and didn't even think about what your test should do

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroParametersTest.java
```diff
@@ -41,4 +41,15 @@ public void setPage()
 
         assertSame(EntityType.PAGE, parameters.getType());
     }
+    
+    @Unstable
```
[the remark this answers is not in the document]
Usamaahmad751: I made changes to the existing code that's why I added `@Unstable`. So It can appear in the backward-compatibility reports.
I read here a few days ago https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices.
I was not really sure to put them for test methods. I am again looking for more information. Thank you!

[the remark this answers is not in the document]
vmassol: Yes you can read about it here: https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility

> I made changes to the existing code that's why I added @Unstable.

But why would making changes to existing code require to add an `@Unstable` annotation? (I know the answer ofc but I'm trying to make it you find it :)).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
vmassol: BTW your test is not passing :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -212,7 +213,20 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
...
+        // Exclude the first heading from the content
+        if (parameters.excludeFirstHeading()) {
+            HeaderBlock heading =
+                    result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);
+            if (heading != null) {
+                result.removeBlock(heading);
+            } else {
+                String includeType = parameters.getSection() != null ? "section" : "document";
+                throw new MacroExecutionException(
+                    String.format("The included [%s] doesn't contain any heading",includeType));
```
vmassol, line 226: Note: bad code style, missing a space after the command. You should apply code style across the full file and make it part of your IDE code style to make sure to not forget it.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
vmassol: I've applied your changes locally and ran your test to see what it does. It tests one case (the negative condition) but not the positive one so we cannot provide it works yet. See the red part in:

Screenshot 2020-04-09 at 22 21 04

vmassol:
ok just spent several hours to convert the IncludeTest for you (was written in old junit4+jmock) so that it uses new technology (junit5+mockito) which is better for you and better so that we don't increase our technical debt more by adding a new test in deprecated tech (normally when you add a new test in a test that uses old tech you convert the test to the new tech but that would have been hard for you ;)).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroParametersTest.java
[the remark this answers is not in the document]
Usamaahmad751: Thanks for the link. What I understand is @Unstable is used to Stabilize the public APIs. So When a user uses a class or method annotated with @Unstable he will not certain about stabilization and he can use it at his own risk. When that method or class is stable to use the @Unstable annotation will be removed. It helps in maintaining the backward-compatibility.
For test methods, we do not add this annotation because Tests are added for developers to test their code and not for the user to use them.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
Usamaahmad751: OK. Thanks, @vmassol for paying a lot of attention and helping me understand all these things. I looked at these things more deeply. And now I proposed a better solution. I have written a more efficient test that covers all of my added code. What I did in the test:

**1.** I have added the expected result that I'll get after removing the heading.
**2.** I passed the heading in the mock document and set the `excludeFirstHeading` to true.
**3.** So when function finds the heading in the document and it will remove the heading from content and return the output.
**4.** I compared that output with the expected one to verify my added code. 
I verify the build using `mvn install -Pquality` and it builds successfully. I also verify the code coverage using command `mvn jacoco:report`. I used the generated HTML file to verify the code coverage. Below is the screenshot:

Capture

Capture2

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -214,7 +228,26 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
 
         return Arrays.asList(metadata);
     }
-
+    
+    private boolean removeBlockRecursively(Block block, Block blockToRemove)
```
vmassol, line 232: I really don't understand this method. Can you explain why you added it?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
```diff
@@ -202,7 +203,20 @@ public void setDocumentDisplayer(DocumentDisplayer documentDisplayer)
                 references.pop();
             }
         }
-
+        
+        // Exclude the first heading from the content
+        if (parameters.excludeFirstHeading()) {
+            HeaderBlock heading = result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class),
+                    Block.Axes.DESCENDANT_OR_SELF);
```
vmassol, line 210: BTW this is wrong. You need to write some test cases to figure it out. Try different examples you can think of content with headings and I'm sure you'll see why.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
```diff
@@ -536,4 +536,46 @@ private XDOM getXDOM(String content) throws Exception
 
         return blocks;
     }
+    
+    @Test
+    void executeIncludeMacroWhenExcludeFirstHeadingTrue() throws Exception
```
vmassol, line 541: You're still missing at least 3-4 other tests... Try to think of all possible inputs and paths taken in the code you added.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
```diff
@@ -536,4 +536,46 @@ private XDOM getXDOM(String content) throws Exception
...
+            + "beginMetaData [[source]=[wiki:space.document][syntax]=[XWiki 2.0]]\n"
+            + "beginSection\n"
+            + "beginSection\n"
+            + "beginParagraph\n"
+            + "onWord [content]\n"
+            + "endParagraph\n"
+            + "endSection\n"
+            + "endSection\n"
+            + "endMetaData [[source]=[wiki:space.document][syntax]=[XWiki 2.0]]\n"
+            + "endDocument";
```
vmassol, line 555: Why are you expecting this? Why expect some sections? Why expect 2 sections? What are you testing?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
```diff
@@ -536,4 +536,46 @@ private XDOM getXDOM(String content) throws Exception
 
         return blocks;
     }
+    
+    @Test
+    void executeIncludeMacroWhenExcludeFirstHeadingTrue() throws Exception 
+    {
+        // Expected content containing no heading.
```
vmassol, line 543: Then why are there sections? This looks weird to have a section inside a section for no reason I can see.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
```diff
@@ -536,4 +536,46 @@ private XDOM getXDOM(String content) throws Exception
...
+        when(this.macroEntityReferenceResolver.resolve("document", EntityType.DOCUMENT,
+                macroContext.getCurrentMacroBlock())).thenReturn(resolvedReference);
+        when(this.contextualAuthorizationManager.hasAccess(Right.VIEW, resolvedReference)).thenReturn(true);
+        when(this.dab.getDocumentInstance((EntityReference) resolvedReference)).thenReturn(this.document);
+        when(this.dab.getTranslatedDocumentInstance(this.document)).thenReturn(this.document);
+        when(this.dab.getCurrentDocumentReference())
+                .thenReturn(new DocumentReference("wiki", "Space", "IncludingPage"));
+        when(this.document.getDocumentReference()).thenReturn(resolvedReference);
+        when(this.document.getSyntax()).thenReturn(Syntax.XWIKI_2_0);
+        when(this.document.getXDOM()).thenReturn(getXDOM("== Heading ==\ncontent")); // To test
```
vmassol, line 574: yes! that's much better.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
Usamaahmad751: This method will find the heading in the children as well as in the children of a child. When a child is matched with the heading block, It will remove that child and exit. The approach I added before was just looking into the child. So this method is a better implementation for removing the first heading.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
Usamaahmad751: **Inputs:** True and false

**True:** covered
**False:** Yet to cover

**Paths:**
**1.** When heading is null(heading not found).
**2.** When heading is not null (heading found).
**3.** When heading is not the first heading.

So, there is a total of 5 test cases that need to be designed. I have covered one till now.
Please correct me if I am wrong.

Usamaahmad751:
Yes, I realize I still have many mistakes in my unit tests. I really apologize for so many revisions. I am looking into it to fix them. Thanks for your reviews and valuable time.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
vmassol: It's actually a very bad implementation IMO (you traverse all children)... and AFAICS it does nothing more than what the caller is already doing already with `result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);`.  I really don't understand why you added it. You already know the block to remove!

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/test/java/org/xwiki/rendering/internal/macro/IncludeMacroTest.java
vmassol: 2 doesn't mean anything, it's not precise. Maybe you meant when the heading the first block in the include?

You're still missing the most important use cases that would show that your code is buggy... I dropped several hints already (more than hint, I already mentioned your code is not correct above)...

vmassol: 3 is also not precise enough.

vmassol: Just write them and you'll see.

vmassol: You're on the right way

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
Usamaahmad751: Yes, I agree, but when I try to remove the using `result.removeBlock(heading)` I got `Provided Block to remove is not a child` exception. `result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);` will get the heading to remove. and to remove that heading we have to find in a child as well as children of a child. So that's why I implement this function. `result.removeBlock(heading)` will only look into its own children. Opinion, please.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacro.java
Usamaahmad751: I tested in all 5 cases but I didn't find any bug in my implemented examples. I can also use `getBlocks()` function to get the heading. But this single line approach was good to use.

Usamaahmad751:
I have added test cases to verify `result.getFirstBlock(new ClassBlockMatcher(HeaderBlock.class), Block.Axes.DESCENDANT_OR_SELF);`. I have added sections in the expected result because `=heading=` create a section with `id="Hheading"`. I look into the macro context but I didn't get desired result. I will look more into it to fix it.

Usamaahmad751:
Hi @vmassol! I have updated the implementation and restricted the exclude heading only to First Block. I have updated tests according to implementation. Please have a look at it. 
If you don't mind please can you tell me why a section is created around a heading when I use `=`. and two sections when I use `==`. I know a section is created using heading But in this case how I can get rid of sections?  I tried a lot but didn't get the desired result.
Thank you so much for your help. :)

excludeFirstHeading

PR #1293, opened 2020-04-13
XWIKI-17197: Make menu colors customization easier for developers

arcilli:
Used variables for menu style. This should allow changing the menu style without override properties in multiple places.


mflorea:
> This should allow changing the menu style without override properties in multiple places.

How can a developer modify these variables? AFAICS these new variables are defined and used in the same place (``Menu.MenuMacro``), so the only way to modify them is by modifying the ``Menu.MenuMacro`` **extension** page. But this page is not meant to be modified. We don't encourage users/developers to modify extension pages (with some exceptions, but it's not the case here) because it causes upgrade problems. Maybe I'm missing something.

PR #1296, opened 2020-04-16
XWIKI-16967: Failed to get the list of flavors on Distribution Wizard on Tomcat

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-16967

  * Don't use AbstractJobRestURLGenerator#getIdStringElement anymore but
    rely on the full list and handle it in Utils#createURI same as we do
for space names
  * Add a unit test for JobStatusRestURLGenerator

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/Utils.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/Utils.java
@@ -360,17 +360,18 @@ public class Utils
             if (pathElement != null) {
                 try {
                     // see generateEncodedSpacesURISegment() to understand why we manually handle "spaceName"
+                    // Note that it would be cleaner to not rely on the variable name and to rely on the instanceof List
+                    // and on generateEncodedListURISegment and to give as argument
+                    // the list of spaces with the intermediate "spaces" segments. We don't go there for now to avoid
+                    // breaking something since this method is heavily used.
                     if (i < pathVariableNames.size() && "spaceName".equals(pathVariableNames.get(i))) {
                         if (!(pathElement instanceof List)) {
                             throw new RuntimeException("The 'spaceName' parameter must be a list!");
                         }
                         encodedPathElements[i] = generateEncodedSpacesURISegment((List) pathElements[i]);
-                    // see generateEncodedJobIdURISegment to understand why we manually handle "jobId"
-                    } else if (i < pathVariableNames.size() && "jobId".equals(pathVariableNames.get(i))) {
-                        if (!(pathElement instanceof List)) {
-                            throw new RuntimeException("The 'jobId' parameter must be a list!");
-                        }
-                        encodedPathElements[i] = generateEncodedJobIdURISegment((List) pathElements[i]);
+                    // see generateEncodedJobIdURISegment to understand why we manually handle the lists
+                    } else if (pathElement instanceof List) {
+                        encodedPathElements[i] = generateEncodedListURISegment((List) pathElements[i]);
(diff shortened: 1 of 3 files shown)


surli:
@tmortagne could you double check this, I think it's ok but maybe not the best option. Note that I tried to directly let JAX-RS manage it by calling:
```
return Utils.createURI(getBaseURI(), JobStatusResource.class, id.toArray()).toURL();
```
but apparently it considers only the first element of the list to create the URL in that case.
Right now all tests are passing when calling `mvn clean install -Pquality,integration-tests` on the xwiki-platform-rest. 

@gdelhumeau @mflorea you might have an opinion too on this stuff since you worked on it.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/Utils.java
```diff
@@ -365,6 +365,12 @@ public static URI createURI(URI baseURI, java.lang.Class< ? > resourceClass, jav
                             throw new RuntimeException("The 'spaceName' parameter must be a list!");
                         }
                         encodedPathElements[i] = generateEncodedSpacesURISegment((List) pathElements[i]);
+                    // see generateEncodedJobIdURISegment to understand why we manually handle "jobId"
+                    } else if (i < pathVariableNames.size() && "jobId".equals(pathVariableNames.get(i))) {
```
tmortagne, line 369: I really don't like this. We should not know about "jobId" parameter in this generic location.

surli: What I could do to make it more generic is to trigger the condition on "instanceof List", but I'm a bit afraid on the possible side effects on the future. 

Would be better if JAX-RS handled it directly: when I try to pass in argument `id.toArray()` the `encodedPathElements` were properly created but then in `uriBuilder.buildFromEncoded(encodedPathElements)` (https://github.com/xwiki/xwiki-platform/pull/1296/files#diff-3d532b9e95503f09519b75a7525fc3beR388) only the first element was used. Maybe you have an idea to tell it to use all elements?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/url/resources/JobStatusRestURLGenerator.java
```diff
@@ -46,7 +46,7 @@ public URL getURL(List<String> id) throws XWikiRestException
         // The idea is to use the UriBuilder of jax-rs to generate URLs that match the resources paths.
         // So it is consistent.
         try {
-            return Utils.createURI(getBaseURI(), JobStatusResource.class, getIdStringElement(id)).toURL();
+            return Utils.createURI(getBaseURI(), JobStatusResource.class, id).toURL();
```
tmortagne, line 49: Seems to me the real fix is to pass `id.toArray()` here instead of a the `id` List, that way instead of passing a list as first element you pass several elements.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/Utils.java
tmortagne: If what you mean is that passing several elements to Utils#createURI does not work then that's a huge bug and it needs to be fixed since it means the contract is not respected and it does not make any sense to take a `java.lang.Object... pathElements`.

surli:
So I edited the solution to handle the same way any argument of type List in the future.

PR #1299, opened 2020-04-20
XWIKI-17225: Annotation link is visible when More Action menu is removed

CamiAndrei:
* Dropped annotation support for older skins (Colibri)


vmassol:
IMO we should not use JS but instead use a UIX. WDYT?

mflorea:
@vmassol it already uses a UIX to inject the link in the menu, but it also needs to add a click listener from JavaScript. The issue here is that the code adding the click listener assumes the skin is Colibri if the link is missing and adds the missing link, which is wrong and @CamiAndrei PR fixes this.

CamiAndrei:
Thank you @mflorea ! I was thinking the same .
Since 7.4M1, the annotations trigger is inserted via an UIX. 
Could this also cherry picked for 11.10.5?

mflorea:
> Could this also cherry picked for 11.10.5?

@CamiAndrei already did.

CamiAndrei:
Great! Thanks!

PR #1301, opened 2020-04-23
XWIKI-16973: Spaces in file names in Zip files cause VFS Tree macro break

surli:
JIRA: 
  * https://jira.xwiki.org/browse/XWIKI-16793
  * https://jira.xwiki.org/browse/XWIKI-17244

This PR fixes two issues related to VFS macro:
  * the fact that we couldn't have a file with a space in a name in an archive (XWIKI-16973)
  * the fact that we couldn't rely on an archive with a space in its name (XWIKI-17244)

The provided fix consists mainly in not relying anymore on URI in VfsResourceReference, but on a new String reference attribute. For the problem of the space name in attachment name, it also needed some work to encode the name and decode it in VFS.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-vfs/xwiki-platform-vfs-api/src/main/java/org/xwiki/vfs/VfsResourceReference.java
+++ b/xwiki-platform-core/xwiki-platform-vfs/xwiki-platform-vfs-api/src/main/java/org/xwiki/vfs/VfsResourceReference.java
@@ -107,7 +107,7 @@ public class VfsResourceReference extends EntityResourceReference
      * @param fullURI the full opaque URI containing both the reference to the archive and the path to the entry inside
      * it, e.g. {@code attach:space.page@attachment/path/to/file}. Note that this constructor requires that the
      * full URL to be URL-encoded.
-     * @deprecated Since 12.4RC1 and 11.10.5, this constructor shouldn't be used anymore, in particular not for internal
+     * @deprecated Since 12.4RC1, this constructor shouldn't be used anymore, in particular not for internal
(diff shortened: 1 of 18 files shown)


xwiki-platform-core/xwiki-platform-vfs/xwiki-platform-vfs-api/src/main/java/org/xwiki/vfs/VfsResourceReference.java
```diff
@@ -95,24 +105,77 @@ public VfsResourceReference(URI uri, String pathSegments)
...
     }
 
     /**
      * @return the URI to the VFS (e.g. {@code attach:space.page@file.zip}, {@code http://server/path/to/zip})
+     * @deprecated Since 12.3 this method shouldn't be used anymore, in favor of {@link #getReference}.
      */
+    @Deprecated
     public URI getURI()
     {
         return this.uri;
```
tmortagne, line 156: Should this try to create a URI with the old logic when it's null for retro compatibility ?

surli: Could be safer indeed.

surli:
@tmortagne change done for `#getURI`. I put the since on 11.10.5 too, since I was thinking about cherry-picking it for LTS too, let me know if you think it's too dangerous.

tmortagne:
Given the number of changes you made it's definitely dangerous from VFS point of view. Now if there is enough tests to back it up...

surli:
>  Now if there is enough tests to back it up...

Well I covered my changes, but I definitely might have missed some corner cases. Going to push and fix only for stable for now then.

PR #1302, opened 2020-04-30
XWIKI-6161: Reset Action doesn't work anymore on page without history

manuelleduc:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -7494,7 +7493,15 @@ public void resetArchive(XWikiContext context) throws XWikiException
     {
         boolean hasVersioning = context.getWiki().hasVersioning(context);
         if (hasVersioning) {
-            getVersioningStore(context).resetRCSArchive(this, true, context);
+            XWikiDocument doc = context.getDoc();
+            if (doc != null) {
+                XWikiDocumentArchive documentArchive = doc.getDocumentArchive();
+
+                if (documentArchive == null) {
+                    doc.setDocumentArchive(new XWikiDocumentArchive());
```
tmortagne, line 7501: I don't really understand why your are doing this.

surli:
btw next time don't forget to assign yourself on the jira issue.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -7494,7 +7493,13 @@ public void resetArchive(XWikiContext context) throws XWikiException
     {
         boolean hasVersioning = context.getWiki().hasVersioning(context);
         if (hasVersioning) {
-            getVersioningStore(context).resetRCSArchive(this, true, context);
+            XWikiDocument doc = context.getDoc();
```
surli, line 7496: I might be missing something, but why are you getting the doc from the context, you're already in an instance of XWikiDocument here.

tmortagne: Yes it's really not clear to me what you are exactly trying to do here.

You seems to try to make sure that the archive is loaded in the XWikiDocument before calling resetRCSArchive but as far as I can see the store is already taking care of it in XWikiHibernateVersioningStore#getXWikiDocumentArchive.

manuelleduc: It is for sure loaded later, but for a reason I fail to understand, without the call to getXWikiDocumentArchive, the hibernate session ends up in a bad state, leading to the `Caused by: org.hibernate.NonUniqueObjectException: A different object with the same identifier value was already associated with the session : [com.xpn.xwiki.doc.rcs.XWikiRCSNodeInfo#com.xpn.xwiki.doc.rcs.XWikiRCSNodeId@57e1e1c8[docId=5330878325779721977,version=1.1]]` error.

But I agree that while this solution is solving the problem, the problem might be fixed in a more elegant what by fixing something elsewhere, but I don't have a clear vision of where and how.

surli (approved):

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -7324,7 +7324,11 @@ public void resetArchive(XWikiContext context) throws XWikiException
     {
         boolean hasVersioning = context.getWiki().hasVersioning(context);
         if (hasVersioning) {
-            getVersioningStore(context).resetRCSArchive(this, true, context);
+            // Pre-load the document's archive to prevent hibernate states inconsistencies
+            // when calling resetRCSArchive.
+            this.loadArchive(context);
```
tmortagne, line 7329: As said in a previous comment resetRCSArchive is already taking care of loading the archive when it's not here already as far as I can see in the code. If there is a bug in XWikiHibernateVersioningStore#getXWikiDocumentArchive it seems to me it should be fixed there instead of trying a random thing without any idea of why it fixes it.

surli: @manuelleduc do you plan to take it back to fix this?

PR #1306, opened 2020-05-18
XWIKI-17067 send notifications only to active users

prmdhost:


surli:
The solution looks good to me, to be super safe I'd improve the existing test to check the query in https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-api/src/test/java/org/xwiki/notifications/notifiers/internal/email/NotificationUserIteratorTest.java#L91 to be sure we don't create a regression in the future.

vmassol:
Thanks.

I was thinking about doing this at a lower level to be more generic. We have a user iterator that is used by the notification code I guess and my idea was to filter at that level to remove inactive users. See https://github.com/xwiki/xwiki-platform/tree/a84ca5353222da77c5a0ad174fdd92155685f315/xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-send/xwiki-platform-mail-send-default/src/main/java/org/xwiki/mail/internal/factory/usersandgroups

With your change, the following use case will still be buggy: https://extensions.xwiki.org/xwiki/bin/view/Extension/Mail/MailSenderAPI#HExample4:SendaTemplateemailtoalistofUsersandGroups

In addition, I'm wondering if doing the query at the DB level (with active==1) is not going to cause problems in the future when we make the user storage independent of wiki pages.

surli:
> In addition, I'm wondering if doing the query at the DB level (with active==1) is not going to cause problems in the future when we make the user storage independent of wiki pages.

I thought about that too, but actually it would involve to completely rewrite the query/the way we get users, so it's not only about the `active == 1` here. IMO it's a good fix for solving this specific issue for now. 

> We have a user iterator that is used by the notification code I guess and my idea was to filter at that level to remove inactive users.

Actually the changed file is actually the user iterator for notification, I'm not sure to understand what you mean there. To use a more generic email iterator and to get rid of our specific notification user iterator? 

> With your change, the following use case will still be buggy: https://extensions.xwiki.org/xwiki/bin/view/Extension/Mail/MailSenderAPI#HExample4:SendaTemplateemailtoalistofUsersandGroups

I don't see the relation with this code.

prmdhost:
I thought about adding a test case for this commit but I don't know how to test the actual query. The current test case completely mocks the result. Maybe you can point me to another test case where a query execution is tested.

surli:
> but I don't know how to test the actual query

I wasn't talking about testing the actual query, but just checking what query is used in the mock. Basically in the link I pasted, we are checking that `anyString` is used for the query and we could be more specific.

prmdhost:
Ok, I see. I will update the test case

vmassol:
> I don't see the relation with this code.

I don't know the notification code but I was assuming it would use the generic user and group iterator and mail sender. Maybe it doesn't (would be interesting to know why it doesn't if that's the case).

So if it doesn't we have 2 choices:
1) fix it in 2 places
2) refactor the notifications code to use the generic code and thus fix the issue in one place

prmdhost:
I've updated the commit with the unit test.

surli:
@vmassol I agree that we'd need something more generic in the future, but IMO it's a good first fix.

vmassol:
> @vmassol I agree that we'd need something more generic in the future, but IMO it's a good first fix.

Sure, it fixes one problem.

surli:
Thanks for your contrib @prmdhost :)

surli:
btw next time don't forget to assign yourself on the jira issue.

PR #1313, opened 2020-06-04
XWIKI-17374: XSS Cross Site Scripting

manuelleduc:


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/PropertyClass.java
```diff
@@ -844,4 +852,23 @@ private String getTypeName()
...
         currentProperty.merge(previousProperty, newProperty, configuration, context, mergeResult);
     }
+
+    /**
+     * Render the provided text to prevent XSS.
+     *
+     * @param text the text to escape.
+     * @param context the xwiki context.
+     * @return the text once sanitized.
+     */
```
vmassol, line 862: Missing `@Unstable`. I don't like the method name. You say it's about rendering and it's named sanitize. IMO the name should not do the rendering at all. 

Also I have the feeling the code is not in the right place. It seems like it should be in XWikiDocument since it calls getRenderedContent() on a XWikiDocument.

Why do you remove a paragraph? Is it always HTML?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/PropertyClass.java
```diff
@@ -844,4 +852,23 @@ private String getTypeName()
     {
         currentProperty.merge(previousProperty, newProperty, configuration, context, mergeResult);
     }
+
+    /**
+     * Render the provided text to prevent XSS.
```
vmassol, line 857: Doesn't explain what it does.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/PropertyClass.java
```diff
@@ -210,6 +213,11 @@ public void displayHidden(StringBuffer buffer, String name, String prefix, BaseC
     @Override
     public void displayView(StringBuffer buffer, String name, String prefix, BaseCollection object,
         XWikiContext context)
+    {
+        displayViewNoEscape(buffer, name, object);
+    }
+
+    protected void displayViewNoEscape(StringBuffer buffer, String name, BaseCollection object)
```
vmassol, line 220: new API, missing unstable annotation too. I'm not fond of adding these 2 new APIs.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/PropertyClass.java
vmassol: Where is the sanitization?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/PropertyClass.java
vmassol: Don't use StringBuffer, StringBuilder is recommended.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
```diff
@@ -123,4 +123,12 @@ public void displayEdit(StringBuffer buffer, String name, String prefix, BaseCol
         buffer.append(input.toString());
     }
 
+    @Override
+    public void displayView(StringBuffer buffer, String name, String prefix, BaseCollection object,
+        XWikiContext context)
+    {
+        StringBuffer result = new StringBuffer();
```
vmassol, line 130: Don't use StringBuffer, StringBuilder is recommended.

PR #1314, opened 2020-06-04
XWIKI-17385 Add Livedata basic Table layout structure with vuejs

ClemDee:


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
```diff
@@ -0,0 +1,90 @@
+<template>
+    <table class="xwiki-livetable">
+
```
mflorea, line 3: What's with the empty lines?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
```diff
@@ -0,0 +1,90 @@
+<template>
+    <table class="xwiki-livetable">
```
mflorea, line 2: We use 2 space indendation for HTML, JavaScript and CSS in XWiki. It would be nice to keep  @the code style consistent.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
```diff
@@ -0,0 +1,90 @@
...
+
+        <thead>
+
+            <tr>
+
+                <th
+                    v-for="(col, cIdx) in cols"
```
mflorea, line 9: I'm not a fan of cryptic variable names. I don't think reducing the variable name by only **2** characters is worth in this case.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
```diff
@@ -0,0 +1,90 @@
...
+                    :key="cIdx"
+                >
+                    <div class="column-header">
+                        <div>{{ col.name }}</div>
```
mflorea, line 13: Do you need the 2 DIVs? Why not put the ``column-header`` class on the TH? Why do you need an additional wrapping for the column name?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
```diff
@@ -0,0 +1,90 @@
...
+            >
+
+                <td
+                    v-for="col in cols"
+                    :key="col.key"
+                >
+                    <div
+                        class="cell"
+                        v-text="row[col.key]"
+                    ></div>
```
mflorea, line 37: Why not put the ``cell`` class on the TD? I think it would make more sense. And why do you need the extra DIV wrapping? I wouldn't complicate the HTML structure unless it's really needed.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
```diff
@@ -0,0 +1,43 @@
...
+    "Vue",
+    "vue!" + BASE_PATH + "layouts/livedata-table.html",
+    "mockData"
+], function (
+    Vue,
+    livedataTable,
+    mockData
+) {
+
+    new Vue({
```
mflorea, line 31: I see you love empty lines :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
```diff
@@ -0,0 +1,43 @@
...
+
+        el: "#livedata",
```
mflorea, line 33: Should be a CSS class not an ID in the future.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/mockData.js
```diff
@@ -0,0 +1,109 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+define({
+    "totalrows": 5,
+    "returnedrows": 5,
+    "offset": 1,
+    "sort": [{property: "age", dir: "desc"}],
+    layouts: ["table"],
+    currentLayout: "table",
```
mflorea, line 26: I would still like to separate the query (properties, sort, filters) from the data (entries, count) and from the meta data (property descriptors, layouts, pagination), like in the spec (which I updated).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/mockData.js
```diff
@@ -0,0 +1,109 @@
...
+
+    properties: [
```
mflorea, line 28: I don't think you can mix ``properties`` with ``propertyDescriptors``. We need both, as they serve different roles. If you're not convinced we can discuss.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
ClemDee: To have the html less compact and thus more readable imo, but if it's a problem for you I can get rid of it :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
ClemDee: I think that `col` is pretty understandable, I changed `cIdx` to `colIdx` if that sounds better, if though it is only used in the next line just to maintain the v-for index

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
ClemDee: As html table element are not very flexible I preferred to create a div element inside the td to avoid troubles in the future. But indeed, as it's not an issue for now I removed it.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
ClemDee: Yes I do :p

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
ClemDee: This is already no longer the case! :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/mockData.js
ClemDee: Sure, it was just in the meantime, now I'm going to use the new specs.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/mockData.js
ClemDee: Yes I totally agree, was just waiting the updated specs, now I have them I'll use them from now :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
mflorea: I don't have a problem with that, but try to be consistent (I haven't seen the pattern / rule yet :P )

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
mflorea: I was only referring to ``colIdx``. I would have used ``colIndex``, which is only 2 characters longer. For me it makes sense to use abbreviation when the win is significant, e.g. i18n vs. internationalization. And if you really want to be concise then you can simply use ``i`` as this is a local short-lived variable.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
mflorea: OK, wrapping the column name makes sense when the table header can contain something else, but that means one DIV. I would still put the ``column-header`` class on the TH if it's possible and if you don't have a good reason to keep it.

mflorea:
Continued by #1316

PR #1315, opened 2020-06-04
XWIKI-17374: Fix escaping in property displayer

manuelleduc:


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -387,18 +387,20 @@ $xwiki.getUserName("xwiki:${username}")
   #end
   #if(!$picturelist.containsKey("${username}--${size}"))
     #getUserAvatarURL($username $avatarURL $size)
+    #set ($resolvedUser = $services.model.resolveDocument($username))
+    #set ($userNameEscaped = $escapetool.xml($escapetool.html($xwiki.getPlainUserName($resolvedUser))))
```
surli, line 391: why escaping both xml and html here? that would need a comment I think.

surli: And `getPlainUserName` is not supposed to perform that escape?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/TextAreaClass.java
```diff
@@ -381,11 +381,16 @@ public void displayView(StringBuffer buffer, String name, String prefix, BaseCol
                 buffer.append(result);
             }
         } else {
-            StringBuffer result = new StringBuffer();
-            super.displayView(result, name, prefix, object, context);
+            String result;
+            BaseProperty prop = (BaseProperty) object.safeget(name);
+            if (prop != null) {
+                result = prop.toText();
```
surli, line 387: I guess the escaped is performed there? Could be nice to comment a bit.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
```diff
@@ -649,16 +649,19 @@ private String getDisplayValue(String value, String name, Map<String, ListItem>
...
+        String syntax = context.getDoc().getSyntax().toIdString();
         if (newresult == null) {
             msgname = "option_" + name + "_" + value;
             newresult = localizePlain(msgname);
             if (newresult == null) {
                 msgname = "option_" + value;
                 newresult = localizePlain(msgname);
                 if (newresult == null) {
-                    newresult = displayValue;
+                    newresult = context.getDoc().getRenderedInlineContent(displayValue, syntax);
```
surli, line 660: Not sure to understand why it needed a change there.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
tmortagne: Since xml and html escaping are pretty much the same thing it means you are doing double escaping here which I really doubt is the intent.

tmortagne: > And getPlainUserName is not supposed to perform that escape?

No, getPlainUserName return plain text and not HTML so it's not escaped for HTML.

surli: > No, getPlainUserName return plain text and not HTML so it's not escaped for HTML.

But don't we want it to do the escape?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
tmortagne: I don't understand what you are doing here, a list does not contain wiki content.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
```diff
@@ -123,4 +123,17 @@ public void displayEdit(StringBuffer buffer, String name, String prefix, BaseCol
...
+        XWikiContext context)
+    {
+        String result;
+        BaseProperty prop = (BaseProperty) object.safeget(name);
+        if (prop != null) {
+            result = prop.toText();
+        } else {
+            result = "";
+        }
+        buffer.append(context.getDoc().getRenderedInlineContent(result, context.getDoc().getSyntax().toIdString()));
```
tmortagne, line 137: Same here, StringClass does not deal with wiki content, only TeatArea does.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
tmortagne: > But don't we want it to do the escape?

Again it return plain text, escaping means returning HTML since the escaping syntax is HTML syntax. This value is used in various contextes which includes not HTML contextes.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1241,6 +1243,31 @@ public String getRenderedContent(boolean transformationContextIsolated, XWikiCon
...
+     */
+    @Unstable
+    public String getRenderedInlineContent(String content, String syntax)
+    {
+        try {
+            XDOM xdom = parseContent(Syntax.valueOf(syntax), content, getDocumentReference());
+            BlockRenderer renderer = Utils.getComponent((Type) BlockRenderer.class, getOutputSyntax().toIdString());
+            WikiPrinter printer = new DefaultWikiPrinter();
+            if (!xdom.getChildren().isEmpty()) {
+                renderer.render(xdom.getChildren().get(0).getChildren(), printer);
```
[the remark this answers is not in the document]
tmortagne: We already have org.xwiki.rendering.util.ParserUtils#removeTopLevelParagraph to convert potentially standalone content to inline. And there is other things to do to be really complete, see DefaultMacroContentParser#convertToInline (which should probably be moved in a more generic place).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
manuelleduc: The double escaping might not be the most elegant solution but is needed for one use case related to the use of the username in the alt attribute of the user profile icon, used in the comments and in the top right header. In this case a simple html escaping is not enough to fix the issue.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/TextAreaClass.java
manuelleduc: to text return a raw content, the escaping is realized by parsing and rendering the text in html using the getRenderedInlineContent method.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
manuelleduc: I thought it could. Is it safe to assume that the content of ListClass and StringClass is always using plain/1.0?

manuelleduc: I have my answer in your next comment. I'll adapt accordingly.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
tmortagne: `alt` does not interpret the HTML as far as I can see so it should not need to be escaped twice (only once as any XML inserted in an attribute value).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -221,7 +221,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
           <input type="hidden" name="${xCommentClass}_author" value="$xcontext.user"/>
         #else
           <label for="${xCommentClass}_author"><input type="text" id="${xCommentClass}_author" name="${xCommentClass}_author" value="$author" maxlength="255" /> says:</label>
-          <strong><a href='$xwiki.getURL('XWiki.XWikiLogin', 'login', "xredirect=$escapetool.url($xwiki.relativeRequestURL)")'>$services.localization.render('login')</a></strong>
+          <strong><a href='$xwiki.getURL('XWiki.XWikiLogin', 'login', "xredirect=$escapetool.url($xwiki.relativeRequestURL)")'>$escapetool.html($services.localization.render('login'))</a></strong>
```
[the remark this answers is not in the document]
mflorea: @surli it is user input as you can overwrite the translation at runtime. The best practice IMO to is always escape translation messages when using them inside HTML, but we normally use ``$escapetool.xml`` and not ``$escapetool.html``.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1241,6 +1243,31 @@ public String getRenderedContent(boolean transformationContextIsolated, XWikiCon
         return getRenderedContent(getOutputSyntax(), transformationContextIsolated, context);
     }
 
+    /**
+     * Parse and render inline contents.
+     *
+     * @param content The content to render.
```
mflorea, line 1249: We usually don't end the parameter JavaDoc with a dot because it's not a sentence.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -202,6 +202,8 @@
 import com.xpn.xwiki.web.Utils;
 import com.xpn.xwiki.web.XWikiRequest;
 
+import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
```
mflorea, line 205: Do we have a rule about importing static methods @vmassol ? I know we allow this for tests, but I don't remember if we allow this for actual code.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -387,18 +387,20 @@ $xwiki.getUserName("xwiki:${username}")
   #end
   #if(!$picturelist.containsKey("${username}--${size}"))
     #getUserAvatarURL($username $avatarURL $size)
+    #set ($resolvedUser = $services.model.resolveDocument($username))
+    #set ($userNameEscaped = $escapetool.html($xwiki.getPlainUserName($resolvedUser)))
```
tmortagne, line 391: I feel you should use getUserName instead of escaping the result of getPlainUserName since getUserName job is to return html content already.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -255,12 +255,11 @@ Object.extend(XWiki, {
...
                       });
 
-                      // Replace the element's content with the temporary container's content, while also evaluating any inline scripts.
-                      // Note: This also does stripScripts internally, but it's OK now, since we have already added the necessary external scripts to head.
-                      $(extraID + "pane").update(container.innerHTML);
-
-                      // Notify the others that the DOM has been updated.
-                      document.fire('xwiki:dom:updated', {elements: [$(extraID + 'pane')]});
+                      require(['jquery'], function (jQuery) {
+                        jQuery("#" + extraID + "pane").html(container.innerHTML);
```
surli, line 259: not sure why you use jquery just for this line: sounds like the other line still use prototype.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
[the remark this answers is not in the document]
surli: > @surli it is user input as you can overwrite the translation at runtime.

indeed, not exactly the original issue description but always better to have it fixed.

> but we normally use $escapetool.xml and not $escapetool.html

ok from what I read on the javadoc for `escapetool.xml`, it might be a bit overkill to use `escapetool.html`.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -255,9 +255,11 @@ Object.extend(XWiki, {
                         }
                       });
 
-                      // Replace the element's content with the temporary container's content, while also evaluating any inline scripts.
-                      // Note: This also does stripScripts internally, but it's OK now, since we have already added the necessary external scripts to head.
-                      $(extraID + "pane").update(container.innerHTML);
+                      // jQuery's html() is used instead of prototype's update to avoid the undesired
```
tmortagne, line 258: The previous comment suggest that the execution of scripts located in the html was actually desired. Wasn't the issue more that it was was also interpreting things which were not really scripts ?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
```diff
@@ -387,18 +387,19 @@ $xwiki.getUserName("xwiki:${username}")
   #end
   #if(!$picturelist.containsKey("${username}--${size}"))
     #getUserAvatarURL($username $avatarURL $size)
+    #set ($userName = $xwiki.getUserName($username, false))
```
tmortagne, line 390: Looks this is just codestyle (and a little bit performances) since it does not seems to have any impact on the produced html.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -221,7 +221,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
           <input type="hidden" name="${xCommentClass}_author" value="$xcontext.user"/>
         #else
           <label for="${xCommentClass}_author"><input type="text" id="${xCommentClass}_author" name="${xCommentClass}_author" value="$author" maxlength="255" /> says:</label>
-          <strong><a href='$xwiki.getURL('XWiki.XWikiLogin', 'login', "xredirect=$escapetool.url($xwiki.relativeRequestURL)")'>$services.localization.render('login')</a></strong>
+          <strong><a href='$xwiki.getURL('XWiki.XWikiLogin', 'login', "xredirect=$escapetool.url($xwiki.relativeRequestURL)")'>$escapetool.xml($services.localization.render('login'))</a></strong>
```
mflorea, line 224: When modifying long lines, it would be nice to also format them so that they don't exceed the 120 char limit. Thanks.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
```diff
@@ -742,7 +742,7 @@ public void displayView(StringBuffer buffer, String name, String prefix, BaseCol
             }
             buffer.append(StringUtils.join(newlist, separator));
         } else {
-            buffer.append(getDisplayValue(prop.getValue(), name, map, context));
+            buffer.append(XMLUtils.escape(getDisplayValue(prop.getValue(), name, map, context)));
```
mflorea, line 745: How come the XML escaping is needed only here and not above? I'm asking because the ``newlist`` above is also built with ``getDisplayValue()``. This lack of consistency needs a comment at least.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -255,9 +255,11 @@ Object.extend(XWiki, {
...
+                      // interpretation of the innerHTML during the dom update.
```
mflorea, line 259: > interpretation of the innerHTML

WDYM by this "interpretation"? What exactly is the difference between jQuery's ``html()`` and Prototype's ``update()``. Last time I checked we still had templates with in-line scripts so not executing in-line scripts is not an option.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
manuelleduc: Fixed

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/macros.vm
manuelleduc: Indeed, it used to be useful but the logic has been moved elsewhere, reverting.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
manuelleduc: I've updated the approach and using the update method with the children of the container element instead of its innerHTML allows scripts to be actually executed where the content of the tags attributes are properly escaped.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
manuelleduc: I've updated the approach and using the update method with the children of the container element instead of its innerHTML allows scripts to be actually executed where the content of the tags attributes are properly escaped.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
manuelleduc: IIRC it is related to https://jira.xwiki.org/browse/XWIKI-9680 and adding an escape no line 743 makes `StaticListClassTest#testDisplayView` fail. But I am not sure to understand the previous decisions. I'll dig deeper to understand what's going on.

manuelleduc: As I understand it, fixing XWIKI-17374 and XWIKI-9680 are mutually exclusive.

manuelleduc:
This PR should not be merge until https://github.com/xwiki/xwiki-commons/pull/101 is merged.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
```diff
@@ -123,4 +124,17 @@ public void displayEdit(StringBuffer buffer, String name, String prefix, BaseCol
...
+    public void displayView(StringBuffer buffer, String name, String prefix, BaseCollection object,
+        XWikiContext context)
+    {
+        BaseProperty prop = (BaseProperty) object.safeget(name);
+        String result;
+        if (prop != null) {
+            result = prop.toText();
+        } else {
+            result = "";
+        }
```
vmassol, line 137: Small note that there seems to be code duplication. I see the same logic at https://github.com/xwiki/xwiki-platform/pull/1315/files#diff-912dd4275a5ac5f2116abef20711b489R384

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -253,8 +253,8 @@ Object.extend(XWiki, {
                       });
 
                       // Replace the element's content with the temporary container's content, while also evaluating any inline scripts.
-                      // Note: This also does stripScripts internally, but it's OK now, since we have already added the necessary external scripts to head.
-                      $(extraID + "pane").update(container.innerHTML);
+                      // Note: This also allows script tag defined in the update html to be loaded.
+                      $(extraID + "pane").update(container.descendants()[0]);
```
vmassol, line 257: Typos: "Note: This also allows script tags defined in the updated html to be loaded."

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
manuelleduc: Safe access to the property's text factored here https://github.com/xwiki/xwiki-platform/pull/1315/files#diff-8574f36522a5e8eab9ad2acf469c1ef0R140

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
manuelleduc: Thanks, typo fixed here https://github.com/xwiki/xwiki-platform/pull/1315/files#diff-3c136b7fe6897859a417c7b58211182dR256

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
```diff
@@ -123,4 +124,27 @@ public void displayEdit(StringBuffer buffer, String name, String prefix, BaseCol
...
+        String content = getPropertyText((BaseProperty) object.safeget(name));
+        buffer.append(XMLUtils.escapeElementText(content));
+    }
+
+    /**
+     * Get the text of the base property. Returns the empty string if the base property is null.
+     * @param property the base property
+     * @return the text of the base property. The empty string is returned if the base property is null
+     */
+    protected String getPropertyText(BaseProperty property)
```
tmortagne, line 140: Missing `since` (this is a new API).

vmassol: And `@Unstable`... However I don't think it's right to make it an API. I'd keep this help method not user public (I wouldn't expose it to the world).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
manuelleduc: Reverted back to avoid the introduction of a undesired public API

surli:
@manuelleduc I guess you need to update that one too now that https://github.com/xwiki/xwiki-commons/pull/102 has been merged?

manuelleduc:
This PR is up to date in regard to  xwiki/xwiki-commons#102.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -253,8 +253,8 @@ Object.extend(XWiki, {
                       });
 
                       // Replace the element's content with the temporary container's content, while also evaluating any inline scripts.
-                      // Note: This also does stripScripts internally, but it's OK now, since we have already added the necessary external scripts to head.
-                      $(extraID + "pane").update(container.innerHTML);
+                      // Note: This also allows script tags defined in the updated html to be loaded.
+                      $(extraID + "pane").update(container.descendants()[0]);
```
surli, line 257: This usage of `.descendants()[0]` need to be explained. I'm not sure to understand it in the right context, so I'm pretty sure I won't understand it in 2 weeks.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
```diff
@@ -123,4 +124,17 @@ public void displayEdit(StringBuffer buffer, String name, String prefix, BaseCol
...
     }
 
+    @Override
+    public void displayView(StringBuffer buffer, String name, String prefix, BaseCollection object,
+        XWikiContext context)
+    {
+        BaseProperty property = (BaseProperty) object.safeget(name);
+        String content;
+        if (property != null) {
+            content = property.toText();
```
tmortagne, line 134: I think I would have put the `buffer.append` directly in the if instead of going through a `content` variable but I guess it's a matter of taste.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/TextAreaClass.java
```diff
@@ -381,13 +381,19 @@ public void displayView(StringBuffer buffer, String name, String prefix, BaseCol
                 buffer.append(result);
             }
         } else {
-            StringBuffer result = new StringBuffer();
-            super.displayView(result, name, prefix, object, context);
+            BaseProperty property = (BaseProperty) object.safeget(name);
+            String content;
+            if (property != null) {
+                content = property.toText();
```
tmortagne, line 387: Similar comment than above but with the following `if` instead of just a buffer.append (i.e. avoid useless work when the property is null).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/xwiki.js
manuelleduc: The script is now using `childElements`. I found out at the last minute that the `descendant()[0]` was breaking the history pane...
See https://github.com/xwiki/xwiki-platform/pull/1315/files#diff-3c136b7fe6897859a417c7b58211182dR255

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/StringClass.java
manuelleduc: Else branch removed https://github.com/xwiki/xwiki-platform/pull/1315/files#diff-8574f36522a5e8eab9ad2acf469c1ef0R133

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/TextAreaClass.java
manuelleduc: Else branch removed here too https://github.com/xwiki/xwiki-platform/pull/1315/files#diff-912dd4275a5ac5f2116abef20711b489R384

tmortagne (approved):

PR #1316, opened 2020-06-05
XWIKI-17387 Add Livedata logic script to load table layout

ClemDee:


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-table.html
```diff
@@ -0,0 +1,90 @@
+<template>
+    <table class="xwiki-livetable">
+
```
ClemensRobbenhaar, line 3: Totally unwanted noob remark: has ``vue-good-table`` been considered for the table layout?

ClemDee: Indeed we considered using existing table plugins at first, but as we are doing most of the operations in the server (filtering, sorting, pagination, ...), the table only receives a list of the rows corresponding to this configuration. So we can't use the common functionalities of the plugin as it would do it on the client side with incomplete data, so at the end it's no longer worth it.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,119 @@
...
+        element.removeAttribute("data-data");
+        this.layouts = {};
+    };
+
+
+    /**
+     * Load a layout, or default layout if none specified
+     * @param {String} layoutName The name of the layout to load with requireJS
+     */
+    Logic.prototype.loadLayout = function (layoutName) {
```
mflorea, line 70: Since this function is asynchronous it needs to return a promise so that you can execute code after the layout is loaded.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,119 @@
...
+        var self = this;
+
+        layoutName = layoutName || this.data.layouts["default"];
```
mflorea, line 73: I don't think we need to have a default layout explicitly specified with a dedicated ``default`` property. I usually do this when reading some config:

    actualConfig = $.extend(true, {}, defaultConfig, configReadFromElement)

Having ``default`` properties for any configuration option is too much for me.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,119 @@
...
+
+            // load success
+            function (layout) {
+                // remove current layout from the page
+                if (self.data.layouts.current && self.layouts[self.data.layouts.current]) {
+                    self.element.removeChild(self.layouts[self.data.layouts.current]);
+                }
+                // add layout element in loaded layouts list if not already loaded on the page
+                if (!self.layouts[layoutName]) {
+                    self.layouts[layoutName] = new layout(self.element);
```
mflorea, line 87: Why ``Logic`` with capital letter and ``layout`` with small letter? And where is ``layout`` defined?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,119 @@
...
+                }
+                // add new layout to the page
+                self.element.appendChild(self.layouts[layoutName]);
+                self.data.layouts.current = layoutName;
+
+                // dispatch events
+                var event = new Event("xwiki:livedata:layoutchange");
```
mflorea, line 94: You should use camel case for the event name. And don't you need to pass some data to the event?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
ClemDee: I just forgot to capitalize the `layout` variable after changing the code from a previous version where it was not a class. `Layout` is the object we get from requirejs, that correspond to the layout we want to load. So it is defined in the layout script that is loaded.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
ClemDee: Sure, it's fixed!

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
ClemDee: No no, I didn't mean to create a default for any configuration option, the `layouts.default` was just in case there are several layouts available, and we want to specify which one should be displayed by default to the user.
If you don't want to have a specific property for this though, we can also say that the first layout in the array of available layouts should be used by default, but this might be less intuitive. Or a mix of both: if default layout is not specified, it takes the first available layout.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
ClemDee: Indeed, fixed.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
mflorea: I don't think we need to have a dedicated "default" property and I don't think we need to look for the first layout or something like that. I would simply do as suggested above:

```
let configReadFromElement = JSON.parse($(element).attr('data-config'));
let defaultConfig = {
  ...
  layout: 'table',
  ...
};
let actualConfig = $.extend(true, {}, defaultConfig, configReadFromElement);
```

This means hard-coding the default config in a single place in the code, where it's easy to read and understand.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,126 @@
...
+   * Load a layout, or default layout if none specified
+   * @param {String} layoutName The name of the layout to load with requireJS
+   * @returns {Object} A jquery promise
+   */
+  Logic.prototype.loadLayout = function (layoutName) {
+    var self = this;
+    var defer = $.Deferred();
+
+    layoutName = layoutName || this.data.meta.layouts[0];
+    if (layoutName === this.data.query.currentLayout) return;
```
mflorea, line 76: You need to return a resolved promise here.

Also, I wouldn't put the **layout** in the **query** because the layout doesn't influence the query, i.e. the same query is executed to fetch the data for the table layout and for the card layout. In other words, the code that executes the query on the server side to get the data from the database doesn't have to know which layout will be used to display the data. For me the layout is only about presentation, so I'd put it in **meta**.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,126 @@
...
+
+    // load layout based on it's filename
+    require([BASE_PATH + "layouts/" + layoutName + ".js"],
+      // load success
+      function (Layout) {
+    // remove current layout from the page
```
mflorea, line 82: Bad alignment.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,126 @@
...
+        });
+        self.element.dispatchEvent(event);
+        defer.resolve();
+      },
+
+      // load failure
+      function (err) {
+        // try to load default layout instead
+        if (layoutName !== self.data.meta.layouts[0]) {
+          self.loadLayout(self.data.meta.layouts[0]);
```
mflorea, line 106: You need to resolve / reject your deferred when the promise returned by this "recursive" ``loadLayout`` call is resolved / rejected.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,126 @@
...
+          livedata: self,
+        });
+        self.element.dispatchEvent(event);
+        defer.resolve();
+      },
+
+      // load failure
+      function (err) {
+        // try to load default layout instead
+        if (layoutName !== self.data.meta.layouts[0]) {
```
mflorea, line 105: I would also check if the layouts array is not empty.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/mockData.js
```diff
@@ -0,0 +1,144 @@
...
+    limit: 10,
+
+  },
+
+
+
+  meta: {
+
+    layouts: ["tableLayout"],
+    defaultLayout: "tableLayout",
```
mflorea, line 51: Sounds good.

PR #1317, opened 2020-06-05
XWIKI-17318: Rich editor integration on comments and annotations textareas

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17318


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-tests/src/test/java/org/xwiki/administration/test/ui/CreatePageIT.java
```diff
@@ -22,6 +22,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import org.junit.Ignore;
```
vmassol, line 25: doesn't look good :)

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
```diff
@@ -162,20 +172,30 @@ protected AnnotationResponse postAnnotation(DocumentReference documentReference,
...
 
             // add the annotation
-            Map<String, Object> annotationMetadata = getMap(request.getAnnotation());
-            annotationService.addAnnotation(documentName, request.getSelection(), request.getSelectionContext(),
+            Map<String, Object> annotationsMap = getMap(request.getAnnotation());
+
+            Map<String, Object> annotationMetadata = new HashMap<>();
+            annotationMetadata.put(DATE_KEY, annotationsMap.get(DATE_KEY));
+            annotationMetadata.put(AUTHOR_KEY, annotationsMap.get(AUTHOR_KEY));
+            if (annotationsMap.containsKey("RequiresHTMLConversion")) {
```
vmassol, line 188: what is this key?

vmassol: Note that I don't know how the annotation code works. Just curious to understand the logic. What is annotationMetadata?

vmassol:
That's a big PR! :)

I don't know enough to review it. I just did a quick shallow review. We need some more in depth one by committers who know more about annotations, REST/JS code.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-tests/src/test/java/org/xwiki/administration/test/ui/CreatePageIT.java
manuelleduc: yes sorry, already removed locally.

manuelleduc:
Note that this a draft and things are note as finalized as I wish, but as it is quite big I want to make it public asap.

xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-api/src/main/java/org/xwiki/edit/script/SyntaxContentEditScriptService.java
```diff
@@ -86,4 +91,23 @@ public String wysiwyg(String content, Syntax syntax, Map<String, Object> paramet
...
+    /**
+     * Generates the HTML code needed to edit the given data while taking into account the preference of the current
+     * user.
+     *
+     * @param content the text content to edit
+     * @param syntax the syntax of the given content 
+     * @param parameters the edit parameters
+     * @return the HTML code that displays the editor for {@link SyntaxContent}
+     * @throws EditException if rendering the editor fails
+     */
```
tmortagne, line 104: Missing a `@since`

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -0,0 +1,39 @@
...
+#set($xCommentClass = 'XWiki.XWikiComments')
+#set ($syntax = $request.syntax)
+#set ($id = $request.id + "$datetool.date.time")
+#set ($number = $numbertool.toNumber($request.number).intValue())
+#set ($name = $request.name)
+#set ($rows = $request.rows)
+#set ($cols = $request.cols)
+#if ($number > -1)
+    #set ($commentObj = $doc.getObject($xCommentClass, $number))
+    #set ($content = $commentObj.getProperty('comment').getValue())
```
tmortagne, line 33: There is a `$commentObj.getValue('comment')` shortcut.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
```diff
@@ -58,9 +62,32 @@ public WebElement getContentField()
         return getContainer().findElement(By.tagName("textarea"));
     }
 
+    /**
+     * Find the content field and runs an action using the field.
+     *
+     * @param action The action.
+     */
+    public void runOnContentField(Consumer<WebElement> action)
```
surli, line 70: That's not really a business API for a PO IMO: would be better to have a more explicit API such as `setContent` or something more aligned with what the users would want to perform as action with that field.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -87,6 +91,10 @@ viewers.Comments = Class.create({
           // without making a new request
           var comment = item.up(this.xcommentSelector);
           comment.hide();
+          const commentNbr = comment.id.match(/.+_(\d+)/)[1];
```
surli, line 94: I guess you're using that one to extract the number from the object name of the comment, right? Seems a bit dangerous to me, but at the very least it needs a comment IMO.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -145,6 +161,9 @@ viewers.Comments = Class.create({
     }
     var comment = editActivator.up(this.xcommentSelector);
     editActivator._x_editForm.hide();
+    const commentNbr = comment.id.match(/.+_(\d+)/)[1];
```
surli, line 164: you reuse the same as defined above, you should use a function for it.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -145,6 +161,9 @@ viewers.Comments = Class.create({
...
+    const name = "XWiki.XWikiComments_ " + commentNbr + "+_comment";
```
surli, line 165: We don't have an hidden field to retrieve it?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -233,15 +259,28 @@ viewers.Comments = Class.create({
...
+              // We enable the form even when it is going to be replaced because the browser caches the state of the
+              // form input fields and the user will get disabled fields when (soft) reloading the page.
               form.enable();
+
+              var comment = form.down('.xwikicomment');
+              var name = "XWiki.XWikiComments_comment";
+              if (comment !== undefined) {
+                const rgx = comment.id.match(/.+_(\d+)/).size();
+                if (rgx > 0) {
+                  const commentNbr = comment.id.match(/.+_(\d+)/)[1];
```
surli, line 271: And here again.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -266,32 +305,47 @@ viewers.Comments = Class.create({
...
-    form.previewButton.observe('click', function() {
-      if (!form.previewButton._x_modePreview && !form.previewButton.disabled) {
-         form.previewButton.disabled = true;
-         var notification = new XWiki.widgets.Notification("$services.localization.render('core.viewers.comments.preview.inProgress')", "inprogress");
-         new Ajax.Request(previewURL, {
-            method : 'post',
-            parameters : {'xpage' : 'plain', 'sheet' : '', 'content' : form.commentElt.value, 'form_token': form.form_token.value, 'restricted': 'true'},
-            onSuccess : function (response) {
+    require(['jquery'], function ($) {
+      if (!form || !XWiki.hasEdit || jQuery("[name='defaultEditorId']").first().val() !== "text") {
```
surli, line 309: you can use directly `$` instead of `jQuery` here no?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -266,32 +305,47 @@ viewers.Comments = Class.create({
...
+      var buttons = form.down('input[type=submit]').up('div');
+      form.previewButton = new Element('span', {'class': 'buttonwrapper'}).update(new Element('input', {
+        'type': 'button',
+        'class': 'button',
+        'value': "$services.localization.render('core.viewers.comments.preview.button.preview')"
+      }));
+      form.previewButton._x_modePreview = false;
+      
+      const buttonName = jQuery(form.previewButton).find("input")[0].value;
+      const existingButton = jQuery(buttons).find("input[value='" + buttonName + "']");
```
surli, line 325: Same here: $ and jQuery

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -198,6 +221,9 @@ viewers.Comments = Class.create({
       // Add listener for submit
       form.down("input[type='submit']").observe('click', function(event) {
         event.stop();
+
+        // triggers the copy of the content of the rich editor to hidden fields
+        jQuery(document).trigger('xwiki:actions:beforeSave');
```
mflorea, line 226: This mix of jQuery and Prototype.js is painful to read.. For small changes, like triggering an event, I don't see why you would switch to jQuery in the middle of Prototype.js code. For large changes, we need to extract the new jQuery based code and place it at the end, after the Prototype.js code, so that it's clear how much code we still need to convert.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -322,9 +378,14 @@ viewers.Comments = Class.create({
...
   doPreview : function(content, form) {
     form.previewButton._x_modePreview = true;
-    form.previewContent.update(content);
-    form.previewContent.show();
-    form.commentElt.hide();
+    const commentElt = jQuery(form).find(".wysiwyg-field");
+    if(jQuery(form).find(".commentPreview").size() === 0) {
+      commentElt.before('<div class="commentcontent commentPreview" style="display: none;"></div>');
+    }
+    const pc = jQuery(form).find(".commentcontainer .commentPreview");
```
surli, line 385: I have a doubt: you're looking for `.commentcontainer` or `.commentcontent` there? since you injected `commentcontent`.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -322,9 +378,14 @@ viewers.Comments = Class.create({
    */
   doPreview : function(content, form) {
     form.previewButton._x_modePreview = true;
-    form.previewContent.update(content);
-    form.previewContent.show();
-    form.commentElt.hide();
+    const commentElt = jQuery(form).find(".wysiwyg-field");
```
surli, line 381: sounds like you're using jquery a lot here too: would be good to encapsulate it in a require(['jquery'] no?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,22 +420,91 @@ viewers.Comments = Class.create({
...
+
+    this.destroyEditor("[name='" + name + "']", name);
+
+    require(['jquery', 'xwiki-events-bridge'], function ($) {
+      if ($(".wysiwyg-field").size() > 0) {
+        $.post(new XWiki.Document().getURL("view") + '?' + $.param({
+          number: commentNbr,
+          xpage: 'xpart',
+          vm: 'commentfield.vm',
+          syntax: 'xwiki/2.1',
```
surli, line 472: The syntax is always necessarily `xwiki/2.1` we don't authorize to comment in `xwiki/2.0` or `markdown`?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -266,32 +305,47 @@ viewers.Comments = Class.create({
...
-            parameters : {'xpage' : 'plain', 'sheet' : '', 'content' : form.commentElt.value, 'form_token': form.form_token.value, 'restricted': 'true'},
-            onSuccess : function (response) {
+    require(['jquery'], function ($) {
+      if (!form || !XWiki.hasEdit || jQuery("[name='defaultEditorId']").first().val() !== "text") {
+        return;
+      }
+
+      var previewURL = "$xwiki.getURL('__space__.__page__', 'preview')".replace("__space__",
+          encodeURIComponent(XWiki.currentSpace)).replace("__page__", encodeURIComponent(XWiki.currentPage));
+      form.commentElt = form.down('textarea');
```
mflorea, line 315: Even more mixing..

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/CommentAsAdminTest.java
```diff
@@ -76,43 +79,33 @@ public void testReplyToCommentAsAdmin()
...
 
     @Test
     public void testDeleteCommentAsAdmin()
     {
-        Assert.assertTrue(this.commentsTab.isCommentFormShown());
+        assertTrue(this.commentsTab.isCommentFormShown());
         this.commentsTab.postComment(COMMENT_CONTENT, true);
         this.commentsTab.deleteCommentByID(this.commentsTab.getCommentID(COMMENT_CONTENT));
-        Assert.assertTrue(this.commentsTab.getCommentID(COMMENT_CONTENT) == -1);
+        assertEquals(this.commentsTab.getCommentID(COMMENT_CONTENT), -1);
```
surli, line 93: sounds like the expected and obtained values are swapped

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,22 +420,91 @@ viewers.Comments = Class.create({
...
         this.startup();
       }
     }.bindAsEventListener(this);
     document.observe("xwiki:docextra:loaded", listener);
+  },
+  destroyEditor: function (selector, name) {
+
+    // TODO: delete and find a way to directly call the editor destroy event.
+    if (typeof CKEDITOR !== 'undefined' && name in CKEDITOR.instances) {
+      CKEDITOR.instances[name].destroy();
```
mflorea, line 442: Bad. See above. We need to trigger a generic event.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,22 +420,91 @@ viewers.Comments = Class.create({
...
+    }
+
+    this.destroyEditor("[name='" + name + "']", name);
+
+    require(['jquery', 'xwiki-events-bridge'], function ($) {
+      if ($(".wysiwyg-field").size() > 0) {
+        $.post(new XWiki.Document().getURL("view") + '?' + $.param({
+          number: commentNbr,
+          xpage: 'xpart',
+          vm: 'commentfield.vm',
```
mflorea, line 471: I think you need to find a way to use the generic ``display.vm``.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
```diff
@@ -69,9 +71,22 @@
...
-
     private static final String ANNOTATION_REQUEST_FILTER_PARAMETER_PREFIX = "filter_";
 
+    private static final String COMMENT_KEY = "comment";
+
+    /**
+     * The name of the request parameter whose multiple values indicate the request parameters that require HTML
+     * conversion. For instance, if this parameter's value is {@code [description, content]} then the request has two
+     * parameters, {@code description} and {@code content}, requiring HTML conversion. The syntax these parameters must
+     * be converted to is found also on the request, under {@code description_syntax} and {@code content_syntax}
```
surli, line 82: Sounds like you encoded a lot of things in a simple String. Wouldn't that be safer for maintenance purpose to encode all that information in a dedicated component whose job would be to retrieve or update the needed information from the parameter map?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Macros.xml
```diff
@@ -20,7 +20,7 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
 
-<xwikidoc version="1.3" reference="AnnotationCode.Macros" locale="">
+<xwikidoc version="1.4" reference="AnnotationCode.Macros" locale="">
```
surli, line 23: No change on the file but its version change?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -229,7 +229,9 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
         <input type="hidden" name="${xCommentClass}_replyto" value="$!replyTo"/>
         <div class="commentcontainer">
            <label for="${xCommentClass}_comment">$services.localization.render('core.viewers.comments.add.comment.label')</label>
-           <textarea id='${xCommentClass}_comment' rows='5' cols="80" name='${xCommentClass}_comment'>$!comment</textarea>
+          <div class="wysiwyg-field">
```
mflorea, line 232: Is it always a WYSIWYG field? What if the current user prefers the text editor?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -292,7 +294,9 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
 #if($date), <span class="commentdate">$!xwiki.formatDate($date)</span>#end
       </div>
     </div>## commentheader
-    <div class="commentcontent">$doc.display('comment', 'edit', $comment)</div>
+    <div class="commentcontent wysiwyg-field-${comment.number}">
```
mflorea, line 297: Same comment as above: is it always a WYSIWYG field?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
```diff
@@ -58,9 +62,54 @@ public WebElement getContentField()
...
+                                         .findElements(By.tagName("textarea"))
+                                         .stream()
+                                         .filter(it -> !it.getAttribute("class").contains("ckeditor-textarea"))
+                                         .collect(Collectors.toList());
+
+        if (textareas.size() == 0 || !textareas.get(0).isDisplayed()) {
+            WebElement iframe = getContainer().findElement(By.tagName("iframe"));
+            getDriver().switchTo().frame(iframe);
+            action.accept(getDriver().findElement(By.tagName("body")));
+            getDriver().switchTo().parentFrame();
```
mflorea, line 82: Maybe we should use the CKEditor page objects here. WDYT?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/display_macros.vm
```diff
@@ -151,4 +151,4 @@
   #set ($macro.importStrings = $stringtool.join($macro.importStrings, ''))
   #set ($return = $NULL)
   #setVariable("$return" $macro.importStrings)
-#end
+#end
```
mflorea, line 154: Insignificant change that pollutes the diff.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -23,349 +23,525 @@ var viewers = XWiki.viewers = XWiki.viewers || {};
...
-    this.addReplyListener();
-    this.addSubmitListener(this.form);
-    this.addCancelListener();
-    this.addEditListener();
-    this.addPreview(this.form);
-  },
+viewers.Comments = Class.create((function () {
+
+  // Private fields and methods
+  const commentNbrRegex = /.+_(\d+)/;
```
mflorea, line 29: You win 3 characters, I lose 3 seconds trying to understand what it means :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -481,7 +657,7 @@ require(['jquery', 'xwiki-events-bridge'], function($) {
     var commentsNumber = $('.xwikicomment').size();
     if(commentsTab) {
       commentsTab.text("$services.localization.render('docextra.extranb', ['__number__'])".replace("__number__", commentsNumber));
-    };
+    }
     if($('#tmComment').length) {
       // All the sub-nodes of tmComment are added in a normalized form.
       $('#tmComment')[0].normalize();
```
mflorea, line 663: This is very hard to review.. since you moved code from one place to another it's hard to see what exactly was changed. Would have been better to try to minimize the diff, especially since this is a large file that serves an important feature...

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
```diff
@@ -162,39 +174,40 @@ protected AnnotationResponse postAnnotation(DocumentReference documentReference,
...
             // check access to this function
-            if (!annotationRightService.canAddAnnotation(documentName, getXWikiUser())) {
+            if (!this.annotationRightService.canAddAnnotation(documentName, getXWikiUser())) {
                 throw new WebApplicationException(Status.UNAUTHORIZED);
             }
 
             // add the annotation
             Map<String, Object> annotationMetadata = getMap(request.getAnnotation());
-            annotationService.addAnnotation(documentName, request.getSelection(), request.getSelectionContext(),
+
```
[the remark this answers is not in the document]
surli: Looks like a typo, I'll remove it.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
```diff
@@ -162,39 +174,40 @@ protected AnnotationResponse postAnnotation(DocumentReference documentReference,
...
-        Map<String, Object> metadataMap = new HashMap<String, Object>();
-        for (AnnotationField f : fields.getFields()) {
+        Map<String, Object> metadataMap = new HashMap<>();
+        for (AnnotationField f : annotation.getFields()) {
             metadataMap.put(f.getName(), f.getValue());
         }
 
+        if (metadataMap.containsKey(REQUIRES_HTML_CONVERSION)
+            && COMMENT_KEY.equals(metadataMap.get(REQUIRES_HTML_CONVERSION))) {
+            String syntax = (String) metadataMap.getOrDefault(String.format("%s_syntax", COMMENT_KEY), "xwiki/2.1");
```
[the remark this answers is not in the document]
surli: > Hard-coding the default syntax doesn't feel right.
Right.

> We should use by default (when not specified) the syntax of the current (target) document or at least the default syntax configured at wiki level.

I follow you on using the target document syntax. Not sure what you mean by "when not specified": there's a way to specify the syntax of an annotation? With a "syntax" parameter maybe? (I haven't checked yet)
Now if we rely on the target document syntax we shouldn't need a fallback on wiki level syntax, should we?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -1528,6 +1545,7 @@ XWiki.Annotation = Class.create({
     this.createPanel.down('input[type=reset]').observe('click', function() {
       this.hideAnnotationCreationForm();
     }.bind(this));
+    jQuery(document).trigger('xwiki:dom:updated', {'elements': jQuery('.annotation-box').toArray()});
```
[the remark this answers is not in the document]
surli: Yeah I was a bit lazy to fix it but I should. Will do it.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -0,0 +1,59 @@
...
+## the number of the comment, if no number is provided we generate a field for a new comment.
+#if ($number && $number > -1)
+  #set ($commentObj = $doc.getObject($xCommentClass, $number))
+  #set ($value = $commentObj.getValue('comment'))
+#else
+  #set ($commentObj = $doc.newObject($xCommentClass))
+  #set ($value = '')
+#end
+
+## We need to know the type of editor used for comments.
```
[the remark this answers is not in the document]
surli: AFAICS it's used for the comments.js at least to enable/disable preview button depending on the editor. Now as I specified on Matrix, I'm not even entirely sure if we should disable preview button when using the WYSIWYG editor... wdyt?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -0,0 +1,59 @@
...
+#set ($defaultEditorId = $services.edit.syntaxContent.defaultEditorId)
+<input type='hidden' name='defaultEditorId' value="$defaultEditorId" />
+
+## copy and rename to the requested name of the comment field
```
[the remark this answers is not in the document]
surli: OK, I will try my best to improve explanations there, but I didn't follow much so maybe I'll let @manuelleduc improves it when he'll come back :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -0,0 +1,59 @@
...
+#set ($ocls = $commentObj.getxWikiClass())
+#set ($discard = $ocls.getXWikiClass().put($name, $commentClass.getXWikiClass().get('comment')))
+#set ($discard = $ocls.getXWikiClass().get($name).setName($name))
+#set ($ppt = $ocls.get($name))
```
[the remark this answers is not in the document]
surli: yeah it's kind of a funny line the mix with ppt and ocls :) I'll try to clarify all that.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -0,0 +1,59 @@
...
+
+## attribution of the value to the new field in the object. 
+#set ($discard = $commentObj.set($name, $value))
+
+## display of the comment field
+$doc.displayEdit($ppt, '', $commentObj)
+
+## raz the temporary added field.
```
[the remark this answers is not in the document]
surli: "Remise à Zéro" in french I guess so "reset" basically.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
surli: ok I need to check back this one specifically. I think the field name is the same even with a simple text editor, but when I tested the javascript handles it properly so I guess Manuel perform the changes to handle this new field class. 
Might be a bit dangerous to change now...

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
surli: That would make sense indeed.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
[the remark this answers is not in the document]
mflorea: >  there's a way to specify the syntax of an annotation?

There's no parameter specific to the annotation, but the WYSIWYG editor submits the syntax of the edited content using ``<fieldName>_syntax``, so ``comment_syntax`` in this case for instance, which the line above reads. This parameter is needed by the ``ConversionFilter`` in order to be able to perform the conversion from HTML to the specified wiki syntax. So the line above should either:

* don't use a default value (expecting the syntax to always be specified, which should be the case normally)
* or don't perform the conversion if the syntax is not specified
* or fallback on the syntax of the current document
* or fallback on the default wiki syntax

> Now if we rely on the target document syntax we shouldn't need a fallback on wiki level syntax, should we?

Of course, it's one or the other. But as I said, normally you don't even need to specify a default value for the ``comment_syntax`` parameter because the WYSIWYG editor should always send a value. This is just bullet proofing, to ensure nothing can break, but still it shouldn't be done by hard-coding the syntax.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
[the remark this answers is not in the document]
mflorea: > it's used for the comments.js at least to enable/disable preview button depending on the editor.

I figured it out, but the comment should mention **why** it's needed, not only that it's needed :)

> I'm not even entirely sure if we should disable preview button when using the WYSIWYG editor... wdyt?

The preview button doesn't bring any value to the WYSIWYG editor so I would keep it hidden in order to have a cleaner / simpler UI. I don't remember deciding to keep the preview button in the stand-alone WYSIWYG mode. We didn't remove it there just because we were lazy I think.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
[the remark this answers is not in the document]
surli: OK, I'll just check that the syntax is specified and don't perform conversion if that's not the case.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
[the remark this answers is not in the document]
surli: So actually I'm not comfortable on commenting those lines: I don't really understand why those manipulations are performed on the XClass... I understand that the goal of the file is to display a specific XWikiComment xobject field in edit.
Now I really don't get those lines: 
```
#set ($ocls = $commentObj.getxWikiClass())
#set ($discard = $ocls.getXWikiClass().put($name, $commentClass.getXWikiClass().get('comment')))
#set ($discard = $ocls.getXWikiClass().get($name).setName($name))
```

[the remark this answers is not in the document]
mflorea: The name comes from ``comments.js``' ``reloadEditor`` function:

```
var name = 'XWiki.XWikiComments_comment';
if (commentNbr) {
  name = 'XWiki.XWikiComments_' + commentNbr + '_comment';
}
```
As I mentioned above, we're supporting 2 use cases:

* editing a comment, in which case the field name looks like ``XWiki.XWikiComments_0_comment``
* adding a new comment, in which case the field name looks like ``XWiki.XWikiComments_comment``; we can't use an object number because multiple users can add comments at the same time

The first use case is easy. You just need to call ``$doc.display()``. The second case is a bit tricky because we want to display a comment without having a comment. So the first thing we do is we add a new temporary (unsaved) comment object, see above:

    #set ($commentObj = $doc.newObject($xCommentClass))

But we can't simply call ``$doc.display()`` on this object because we would get ``XWiki.XWikiComments_N_comment`` as the field name, where N is the number of the new temporary object. Instead we want the field name to be ``XWiki.XWikiComments_comment``. To achieve this we add a new property to the comment class by copying the comment property and changing the name. Then we display this new property.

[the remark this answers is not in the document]
surli: Thanks for the explanation, so if I follow you, all that stuff we're doing before the display we normally need to do it only in case of newObject, so in the same `#else` than in https://github.com/xwiki/xwiki-platform/pull/1317/files/4bd7613af141305df55f32d398206afed5d8c329#diff-118419b8c5fd09db95b383d36464a362R38 right? 

That would feel cleaner to me to only do it in that case wdyt?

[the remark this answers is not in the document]
mflorea: Yes, we should do it only when there's no comment number specified.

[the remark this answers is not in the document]
surli: OK going to clarify the code then.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
surli: So I checked locally and I confirm we always obtain a wysiwyg-field whatever the editor is. This classname is currently used in comments.js, and in the pageobject. I guess we should be able to rename it for something more generic such as "comment-editor".

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
surli: So I just checked and the problem is that the pageobjects module of CKeditor actually depends on test-ui module from platform. So it would introduce a cyclic dependency. So keeping it like that for now.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -54,6 +74,11 @@ viewers.Comments = Class.create({
     this.addCancelListener();
     this.addEditListener();
     this.addPreview(this.form);
+    this.reloadEditor({
+      callback: function () {
+        this.addPreview(this.form);
```
surli, line 79: I don't understand this callback: we don't want the addPreview for WYSIWYG editor and if I understand properly the reloadEditor is only useful for WYSIWYG editor...

surli: I read back reloadEditor and I got it wrong: it's used for any editor. Now I guess there is a problem in it since we need to call explicitely addPreview.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,20 +458,80 @@ viewers.Comments = Class.create({
...
+    require(['jquery', 'xwiki-events-bridge'], function ($) {
+      $(document).trigger('xwiki:actions:cancel');
+      if ($(selector).size() > 0) {
+        const parent = $(selector).parent(".wysiwyg-field");
+        parent.empty();
+        $(document).trigger('xwiki:dom:updated', {'elements': parent.toArray()});
+      }
+    });
+  },
+  reloadEditor: function (options) {
```
surli, line 486: Not sure about the name: it's actually a `loadEditor` more than a "re" load for me...

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,20 +458,80 @@ viewers.Comments = Class.create({
...
+              var url = $(this).attr('href') || $(this).attr('src');
+              return existingSkinExtensions.indexOf(url) < 0;
+            }).appendTo('head');
+          }
+
+          const wf = $(wfClass);
+          wf.empty();
+          wf.append(data);
+          $(document).trigger('xwiki:dom:updated', {'elements': wf.toArray()});
+          loadRequiredSkinExtensions(jqXHR.getResponseHeader('X-XWIKI-HTML-HEAD'));
```
surli, line 528: If I'm right this is only needed for WYSIWYG editor, I don't know how it would behave for the text editor, but feels wrong to call it every time.

mflorea: The text editor could require, in theory, some skin extensions, so I don't see the problem with calling this for any editor. The ``X-XWIKI-HTML-HEAD`` header is simply not set or empty when the text editor is used I think, so there's no performance penalty.

surli: Ok I'll try to debug it to ensure it doesn't lead to an error with text editor (would explain why the callback isn't properly called to display preview in that case). But keeping it.

surli: Ok it's working properly.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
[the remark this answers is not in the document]
surli: > Yes, we should do it only when there's no comment number specified.

So actually we were wrong about that: Manuel was changing the name for both editing and adding, because else we'd obtain the field name `comment` and not `XWiki.XWikiComments_N_comment`. So the problem is that in case of editing you won't send the complete name of the comment to the server. 
Going to revert my change, I'm just improving comments and variable names there.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
mflorea: I'm not sure why you resolved this conversation. It's still not clear to me why this new code requires jQuery when it continues to use Prototype.js heavily. Why exactly is jQuery required here?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -321,25 +409,44 @@ viewers.Comments = Class.create({
...
+      form.previewButton._x_modePreview = true;
+      const commentElt = $(form).find(".wysiwyg-field");
+      if ($(form).find(".commentPreview").size() === 0) {
+        commentElt.before('<div class="commentcontent commentPreview" style="display: none;"></div>');
+      }
+      const pc = $(form).find(".commentcontent.commentPreview");
+      pc.html(content);
+      pc.show();
+      commentElt.hide();
+      form.previewButton.down('input').value =
```
mflorea, line 422: More mixing of jQuery and Prototype.js .

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -321,25 +409,44 @@ viewers.Comments = Class.create({
...
-    form.previewContent.show();
-    form.commentElt.hide();
-    form.previewButton.down('input').value = "$services.localization.render('core.viewers.comments.preview.button.back')";
+    require(['jquery'], function ($) {
+      form.previewButton._x_modePreview = true;
+      const commentElt = $(form).find(".wysiwyg-field");
+      if ($(form).find(".commentPreview").size() === 0) {
+        commentElt.before('<div class="commentcontent commentPreview" style="display: none;"></div>');
+      }
+      const pc = $(form).find(".commentcontent.commentPreview");
```
mflorea, line 418: ``pc`` as in "previewComment" or "previewContent" or ?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -321,25 +409,44 @@ viewers.Comments = Class.create({
...
   cancelPreview : function(form) {
-    form.previewButton._x_modePreview = false;
-    form.previewContent.hide();
-    form.previewContent.update('');
-    form.commentElt.show();
-    form.previewButton.down('input').value = "$services.localization.render('core.viewers.comments.preview.button.preview')";
+    require(['jquery'], function ($) {
+      if (form.previewButton) {
+        form.previewButton._x_modePreview = false;
+        form.previewButton.down('input').value =
```
mflorea, line 435: Again, mixing jQuery and Prototype.js .

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,20 +457,80 @@ viewers.Comments = Class.create({
...
+    if (commentNbr) {
+      name = 'XWiki.XWikiComments_' + commentNbr + '_comment';
+      wfClass = wfClass + '-' + commentNbr;
+    }
+
+    this.destroyEditor("[name='" + name + "']", name);
+
+    require(['jquery', 'xwiki-events-bridge'], function ($) {
+      if ($(".wysiwyg-field").size() > 0) {
+        $.post(new XWiki.Document().getURL("view") + '?' + $.param({
```
mflorea, line 500: We normally use the ``get`` action for AJAX requests in order to not increase the view statistics artificially.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -0,0 +1,66 @@
...
+#end
+
+## We need to know the type of editor used for comments in order to hide/display the preview button.
+#set ($defaultEditorId = $services.edit.syntaxContent.defaultEditorId)
+<input type='hidden' name='defaultEditorId' value="$defaultEditorId" />
+
+## display of the comment field
+$doc.displayEdit($fieldToDisplay, '', $commentObj)
+
+## reset the temporary added field.
```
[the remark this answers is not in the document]
surli: Keeping it to be safe.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
surli: I renamed it to "commenteditor".

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
surli: I guess I did it by mistake, I just found it back when reviewing again the code... I agree the mix is wrong here. I try to refactor it to jQuery but it's an heavy task. So going to remove the call to jquery instead.

surli:
@mflorea waiting your confirmation that it's clean to merge this one.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-rest/src/main/java/org/xwiki/annotation/rest/internal/AnnotationsRESTResource.java
```diff
@@ -162,39 +178,46 @@ protected AnnotationResponse postAnnotation(DocumentReference documentReference,
...
-        for (AnnotationField f : fields.getFields()) {
+        Map<String, Object> metadataMap = new HashMap<>();
+        for (AnnotationField f : annotation.getFields()) {
             metadataMap.put(f.getName(), f.getValue());
         }
 
+        // We perform conversion only if:
+        //   1. there is a RequiresHTMLConversion parameter
+        //   2. the value of this parameter is exactly "comment"
+        //   3. the syntax of comment is given in the parameters
```
mflorea, line 211: AFAIK the annotation module can be configured to use a different class (than ``XWikiComments``). So it's good to know, and maybe write down, that this code will work when annotations are saved as comments or, to be precise, when the annotation class has a ``comment`` property and that is used for storing the annotation text.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -1542,8 +1562,9 @@ XWiki.Annotation = Class.create({
   },
 
   onAnnotationAdd : function(event) {
-    // don't submit
     event.stop();
+    // don't submit
```
mflorea, line 1566:
```suggestion
    // Notify the others that we're about to submit the annotation, in order to give them the chance to update the form
    // fields before the submit.
```

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -320,26 +414,49 @@ viewers.Comments = Class.create({
...
+    require(['jquery'], function ($) {
+      if (form.previewButton) {
+        form.previewButton._x_modePreview = false;
+        $(form.previewButton).find('input')
+          .val("$services.localization.render('core.viewers.comments.preview.button.preview')");
+      }
+      const pc = $(form).find(".commentPreview");
+      if (pc) {
+        pc.hide();
+        pc.html('');
```
mflorea, line 451:
```suggestion
        pc.empty();
```

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -350,20 +467,81 @@ viewers.Comments = Class.create({
...
     }.bindAsEventListener(this);
     document.observe("xwiki:docextra:loaded", listener);
+  },
+  destroyEditor: function (selector, name) {
+    require(['jquery', 'xwiki-events-bridge'], function ($) {
+      $(document).trigger('xwiki:actions:cancel');
+      if ($(selector).size() > 0) {
+        const parent = $(selector).parent(".commenteditor");
+        parent.empty();
+        $(document).trigger('xwiki:dom:updated', {'elements': parent.toArray()});
```
mflorea, line 491: I don't see the point of triggering the ``'xwiki:dom:updated'`` for an element that was just emptied. ``'xwiki:dom:updated'`` is useful when you add or change the DOM, because it allows others to enhance the added DOM. But here a listener that is called for an empty element won't be able to do much. And doing cleanup is not a valid reason because the listener won't know what content was removed (it just receives the empty parent). I would simply drop this line.

PR #1319, opened 2020-06-09
XWiki-17388 - Add switch functionality between Table and Cards layouts

ClemDee:


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdown-menu.html
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div class="livedata-dropdown btn-group">
+
+    <a class="btn btn-default dropdown-toggle" title="More Actions" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true" role="button">
+      <span class="fa fa-ellipsis-v"></span>
+    </a>
+
+    <ul class="dropdown-menu">
+
+      <li class="dropdown-header">Change Layout</li>
```
mflorea, line 10: This needs to be translated. Same for "More Actions" above. The easiest is to use Velocity:

    $escapetool.xml($services.localization.render('someKey'))

and add ``evaluate=true`` as query string when loading this with RequrieJS.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdown-menu.html
```diff
@@ -0,0 +1,81 @@
...
+<style>
+
+.livedata-dropdown .btn-default {
+  background-color: #f8f8f8;
+  background-image: none;
+  border-color: #e5e5e5;
+  box-shadow: none;
+  color: #333333;
+  text-shadow: none;
+}
```
mflorea, line 75: Why do you need this custom styles? I'm worried because this doesn't respect the configured color theme.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdownMenu.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+define([
+    "Vue",
+    "vue!" + BASE_PATH + "dropdown-menu.html",
```
mflorea, line 23: Here you could add ``?evaluate=true`` in order to support translations in drop down menu.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-cards.html
```diff
@@ -0,0 +1,115 @@
...
+        v-for="(entry, entryId) in entries"
+        :key="entryId"
+      >
+        <div
+          class="card-property"
+          v-for="property in properties"
+          :key="property.id"
+        >
+          <strong>{{ property.name }}:</strong>
+          <span>{{ entry[property.id] }}</span>
```
mflorea, line 21: Perfect for now.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-cards.html
```diff
@@ -0,0 +1,115 @@
...
+
+.livedata-cards-layout .livedata-cards {
+  margin-top: 1rem;
+}
+
+.livedata-cards-layout .card {
+  display: inline-block;
+  padding: 1rem 2rem;
+  margin: 1rem;
+  border: 1px solid lightgray;
```
mflorea, line 96: We need to find a way to use the color theme here. Ideally it would be nice to be able to use LESS here and have access to the skin properties.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/mockData.js
```diff
@@ -0,0 +1,157 @@
...
+
+    filters: {},
+
+    sort: [ {properties: "age", descending: true} ],
+
+    offset: 0,
+    limit: 10,
+
+    layouts: ["table", "cards"],
+    defaultLayout: "table",
```
mflorea, line 45: I think this needs to go in **meta** because the layout doesn't influence the data or how the data is fetched from the database on the server. The layout controls how the data is displayed so for me it naturally goes inside **meta**. ``properties`` vs. ``propertyDescriptors`` is different, because you need the list of properties on the server side when fetching the data. I don't see any problem having both ``layouts`` and ``layoutDescriptors`` in **meta**. We need both later when from design mode you may want to allow the user to choose the layouts that should be available to the end user.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -72,10 +72,10 @@ define(["jquery"], function ($) {
     var self = this;
     var defer = $.Deferred();
 
-    layoutId = layoutId || this.data.query.defaultLayout;
-    if (layoutId === this.data.query.currentLayout) return;
-    if (!this.data.meta.layoutDescriptors[layoutId]) return;
-    if (this.data.query.layouts.indexOf(layoutId) === -1) return;
+    layoutId = layoutId || this.data.meta.defaultLayout;
+    if (layoutId === this.data.query.currentLayout) return defer.reject();
```
mflorea, line 76: Why not resolve? The layout is basically already loaded.

ClemDee: I returned a rejected promise because if we want to execute code after the layout actually changed, in that case nothing happens, and it's still the same layout. But indeed this is not right with the name of the function (**loadlayout**), so I should either I return `defer.resolve()`, or I change the name of the function to `changeLayout`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdown-menu.html
ClemDee: It's very strange I can't parse the velocity code, even with the `BASE_PATH` being `/xwiki/bin/skin/resources/uicomponents/livedata/`. I have currently no idea why it' not working, so I'm going to leave it like that, as it might probably be a configuration problem in the server, and so that you could have a look at my code to see if there is anything that seems wrong.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdown-menu.html
ClemDee: I was trying to get the same style of the existing dropdown menu on the page, because the default one (from bootstrap) looks strange and not coherent with the page design. But yes, it should be better to use scss for this, I'm working on it!

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/livedata-cards.html
ClemDee: Yes, that's the plan. Are less files automatically compiled or do I need to specify something to the server? In addition as the style is inside a vue file it might gets more complicated.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdown-menu.html
ClemDee: Turns out translation is only working with webjars. (which is not very convenient)
As there are more important things to take care of atm, the issue is reported.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdown-menu.html
ClemDee: Need to find a solution to be able to use LESS effectively for vue component. As there are higher priority issues, it is delay for now.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
mflorea: > either return defer.resolve(), or change the name of the function to changeLayout

I would do both :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,128 @@
...
+        });
+        self.element.dispatchEvent(event);
+        defer.resolve();
+      },
+
+      // load failure
+      function (err) {
+        // try to load default layout instead
+        if (layoutId !== self.data.meta.defaultLayout) {
+          return self.changeLayout(self.data.meta.defaultLayout);
```
mflorea, line 108: You need to resolve / reject your promise after the recursive ``changeLayout`` finishes (with success or failure).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+        else {
```
mflorea, line 110: Should be on the same line as ``}``, at least that's our current code style.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdownMenu.js
```diff
@@ -0,0 +1,51 @@
...
+    "vue!" + BASE_PATH + "dropdown-menu.html?evaluate=true",
+  ], function (
+    Vue,
+    dropdownMenu
+  ) {
+
+    return function (logic) {
+
+      /**
+       * Create the table layout from Vuejs
```
mflorea, line 32: Bad copy & paste?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/dropdownMenu.js
```diff
@@ -0,0 +1,51 @@
...
+        // Constructs a dropdown-menu component and passes it the data
+        template: '<dropdown-menu :logic="logic"></dropdown-menu>',
+
+        data: {
+          logic: logic,
+        },
+
+      }).$mount();
+
+      // return the HTML Element of the layout for the layout
```
mflorea, line 45: Bad comment.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/cardsLayout.js
```diff
@@ -0,0 +1,53 @@
...
+      template: '<livedata-cards :logic="logic"></livedata-cards>',
+
+      data: {
+        logic: Logic(element),
+      },
+
+    }).$mount();
+
+    // return the HTML Element of the layout for the logic script
+    return vueCardsLayout.$el;
```
mflorea, line 48: This means this function is not a constructor and thus you shouldn't use it as a constructor in the logic code.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
```diff
@@ -0,0 +1,54 @@
...
+  "vue!" + BASE_PATH + "layouts/livedata-table.html",
+  "Logic"
+], function (
+  Vue,
+  livedataTable,
+  Logic
+) {
+
+  return function (element) {
+    this.logic = Logic(element);
```
mflorea, line 32: Any reason for saving the reference to the logic? It seems to be used only to initialize the data below.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,128 @@
...
+   * Change layout to specified one, or default one if none specified
+   * @param {String} layoutId The id of the layout to load with requireJS
+   * @returns {Object} A jquery promise
+   */
+  Logic.prototype.changeLayout = function (layoutId) {
+    var self = this;
+    var defer = $.Deferred();
+
+    layoutId = layoutId || this.data.meta.defaultLayout;
+    if (layoutId === this.data.query.currentLayout) { return defer.reject(); }
```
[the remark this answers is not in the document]
ClemDee: Yes, indeed we should be able to run code after the function succeeded (layout changed or stayed the same).
But in the first case the event is not send and I wanted to have the same behavior for the promise and the event. However I didn't want the event to be triggered when the layout didn't changed, because this is not intuitive (for instance, the native `change` event does is not triggered when an input has been modified but the value didn't changed).
So I think I'll change the return to resolve, and keep the event system as it is (triggered only when the layout actually changed)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,128 @@
...
+        if (!self.layouts[layoutId]) {
+          self.layouts[layoutId] = new Layout(self.element);
+        }
+        // add new layout to the page
+        self.element.appendChild(self.layouts[layoutId]);
+        self.data.query.currentLayout = layoutId;
+        // dispatch events
+        var event = new CustomEvent("xwiki:livedata:layoutChange", {
+          layoutId: layoutId,
+          livedata: self,
```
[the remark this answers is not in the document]
ClemDee: Very easy fix, done!

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
ClemDee: Oh yeah indeed, I don't know what I did there!
It's fully working now! :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
ClemDee: oh noooo, I hate this code style I find it very unreadable imo :(
But if it's the xwiki code style, I'll stick to it of course.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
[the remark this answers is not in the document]
ClemDee: Also the event system changed in a more recent commit (not in this PR), now it's delegating the work to a dedicated function in `Logic`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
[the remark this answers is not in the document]
mflorea: > So I think I'll change the return to resolve, and keep the event system as it is (triggered only when the layout actually changed)

Definitely.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
mflorea: There is no definitive code style for JavaScript but we try to use as much as possible the same code style as on Java (server side). One exception is the indentation (2 spaces on JavaScript and 4 on Java) but there are others. I don't see why we wouldn't use the same code style for ``if / then / else`` in Java and JavaScript since it's the same syntax. Now, these rules are not set in stone but the process of changing them is not easy as it requires the vote of the majority of committers.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -0,0 +1,128 @@
...
+        // remove current layout from the page
+        if (self.data.query.currentLayout && self.layouts[self.data.query.currentLayout]) {
+          self.element.removeChild(self.layouts[self.data.query.currentLayout]);
+        }
+        // add layout element in loaded layouts list if not already loaded on the page
+        if (!self.layouts[layoutId]) {
+          self.layouts[layoutId] = new Layout(self.element);
+        }
+        // add new layout to the page
+        self.element.appendChild(self.layouts[layoutId]);
```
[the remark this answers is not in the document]
ClemDee: Yes, that was an issue I thought about.
I first tried to return the whole layout object, but that would have implied that every layout should follow the same architecture, with an `element` property for instance containing the layout root element to be used in the Logic script. And also, the Logic script doesn't have to know much more about the layout than its HTML element. So that's why I decided to return only the layout element instead of the whole object. 

I know this is not a very common pattern, but that was the easiest way to generate a new HTML object through a class. I can change it to a simple function like the one you proposed if you prefer.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/layouts/tableLayout.js
ClemDee: No, not anymore.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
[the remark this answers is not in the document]
mflorea: Returning the DOM element is good. I would drop the ``new`` operator.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -80,32 +80,39 @@ define(["jquery"], function ($) {
...
         // dispatch events
         var event = new CustomEvent("xwiki:livedata:layoutChange", {
+          layout: self.layouts[layoutId],
           layoutId: layoutId,
+          previousLayoutId: previousLayoutId,
           livedata: self,
         });
         self.element.dispatchEvent(event);
-        defer.resolve();
+        defer.resolve(self.layouts[layoutId]);
```
mflorea, line 104: That's a good idea, but you need to make sure you pass the layout element to all resolve calls, otherwise some function that expects the layout on resolve will fail if the promise is resolved in a different place (like when you load the default layout).

ClemDee: Yes, and I also forgot to return the resolved promise at the top! I'm just dumb, but it's fixed now!

PR #1327, opened 2020-06-30
Update from Weblate

xwikiorgci:


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources_es.properties
```diff
@@ -2765,14 +2765,14 @@ xe.panels.rights.welcome=Bienvenido al editor de privilegios.
...
 xe.panels.rights.accesslevelsexplanation=Este campo debe contener una lista de los niveles de acceso que se quieren aplicar a los usuarios y grupos especificados. Los niveles de acceso disponibles son: admin (administraci\u00F3n), programming (proramaci\u00F3n), register (registro), edit (edici\u00F3n), view (visualizaci\u00F3n), comment (comentarios). Para proteger su Wiki en modo vista y edici\u00F3n utilice "view, edit". Para proteger de a\u00F1adir comentarios utilice "comment".
 xe.panels.rights.allowdeny=Permitir/Denegar
-xe.panels.rights.allowdenyexplanation=Este campo deber\u00EDa contener 'Permitir' para indicar que se dispone del derecho y 'Denegar' para indicar que no se dispone de este derecho. Si se dispone del derecho el Wiki o el Espacio es *exclusivamente* visible o editable para los usuarios o grupos que disponen del derecho.
+xe.panels.rights.allowdenyexplanation=Este campo deber\u00EDa contener <em>Permitir</em> para indicar que se dispone del derecho, y <em>Denegar</em> para indicar que no se dispone de ese derecho. Un derecho <em>permitido</em> significa: "esta wiki, espacio o p\u00E1gina es *solo" visible o editable por los usuarios o grupos especificados".
```
surli, line 2775: Still have to decide about that.

surli: This is actually part of the source translation, if we want to change it we have to do it on the source first.

vmassol: `<em>` is actually fine since it's semantic (in HTML5) and not styling. If the translation had put `<i>` or `<b>` it would have been a problem. We'll need to decide what type of HTML is supported in comments. Since XWiki outputs HTML5 ATM (it used to output HTML4 and then XHTML in the past), it means that the HTML in the translations should be consisdered HTML5 since I believe we output directly and don't parse the translations FTM.

mflorea: > it means that the HTML in the translations should be consisdered HTML5 since I believe we output directly and don't parse the translations FTM.

I always XML escape the translations I use. I wouldn't trust the translator to write valid HTML, so I would prefer not to have HTML at all in translations.

PR #1333, opened 2020-07-07

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17524


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsEventExecutor.java
```diff
@@ -0,0 +1,98 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.mentions;
```
tmortagne, line 20: Does this really need to be a public API ?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsThreadFactory.java
```diff
@@ -0,0 +1,49 @@
...
+{
+    private static final String THREAD_NAME = "Mentions pool thread";
+
+    private final ThreadFactory threadFactory = Executors.defaultThreadFactory();
+
+    @Override
+    public Thread newThread(Runnable runnable)
+    {
+        Thread thread = this.threadFactory.newThread(runnable);
+        thread.setDaemon(true);
```
tmortagne, line 43: Note that this means the thread is killed right away when Java stops. Is this OK in your use case or do you need to wait for the thread to finish processing the mentions ?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/listeners/MentionsCreatedEventListener.java
```diff
@@ -74,16 +76,28 @@ public void onEvent(Event event, Object source, Object data)
...
+    private void traverseXObjects(Map<DocumentReference, List<BaseObject>> xObjects, DocumentReference authorReference,
+        DocumentReference documentReference)
+    {
+        for (Map.Entry<DocumentReference, List<BaseObject>> entry : xObjects.entrySet()) {
+            for (BaseObject baseObject : entry.getValue()) {
+                if (baseObject != null) {
+                    for (Object o : baseObject.getProperties()) {
+                        if (o instanceof LargeStringProperty) {
+                            String content = ((LargeStringProperty) o).getValue();
+                            this.executor.executeCreate(content, authorReference, documentReference, AWM_FIELD);
```
tmortagne, line 96: I'm wondering if putting all that in the queue is not a bit expensive in term of memory (imagine what will end up in the queue when we import a big backup).

Might be safer to just queue the reference of documents to process and then get the document and its previous version in the background thread. I guess it should not increase too much the processing time if the mentions found in the previous version of the document are cached.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsEventExecutor.java
manuelleduc: Only the `getQueueSize` and `clearQueue` methods that are required by the script service. I'll simplify the interface.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsThreadFactory.java
manuelleduc: Indeed, that's probably better to remove this line. Is it necessary to define a timeout mechanism in case a thread is hanging for too long?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsEventExecutor.java
tmortagne: What the script service uses does not need to be public.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsThreadFactory.java
tmortagne: For the details I'm not sure since it depends on your need but what you need to have is a way to tell the processing thread that it needs to stop when it's done processing. To trigger this clean stop you need to implement org.xwiki.component.phase.Disposable#dispose() in the component in charge of telling this thread to stop.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionNotificationService.java
```diff
@@ -42,6 +42,6 @@
      * @param location The location of the mention.
      * @param anchorId The anchor link to use.
      */
-    void sendNotif(DocumentReference authorReference, DocumentReference documentReference,
-        DocumentReference mentionedIdentity, MentionLocation location, String anchorId);
+    void sendNotif(String authorReference, String documentReference, DocumentReference mentionedIdentity,
```
surli, line 45: Since you refactor it, I would have renamed it `sendNotification`: no real need to use an abbreviation there...
Now I'm not sure about this refactoring: what's the need for not using a real reference type there? (looking in order so I'll probably get the answer below)

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/internal/MentionsEventExecutor.java
```diff
@@ -0,0 +1,56 @@
...
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Execution the notifications for the mentions asynchronously.
+ *
+ * @version $Id$
+ * @since 12.6RC1
+ */
+@Unstable
```
surli, line 32: No need for unstable here since it's internal.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/script/MentionsScriptService.java
```diff
@@ -62,4 +66,28 @@ public String getSelfMentionsColor()
...
+
+    /**
+     * Clear the queue of elements with mentions to analyze.
+     *
+     * @see MentionsEventExecutor#clearQueue()
+     * @since 12.6RC1
+     *
+     */
+    @Unstable
+    public void clearQueue()
```
surli, line 89: Sounds like something that would request some specific rights to trigger, maybe Admin rights at least?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/pom.xml
```diff
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
   <description>Default implementation of the Mentions components.</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.83</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.62</xwiki.jacoco.instructionRatio>
```
surli, line 35: Not nice...

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/pom.xml
```diff
@@ -54,5 +54,16 @@
       <artifactId>xwiki-platform-annotation-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!-- third party-->
```
surli, line 57: Would be good to provide some info on why you need those.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionNotificationService.java
```diff
@@ -50,12 +50,12 @@
...
 
     @Override
-    public void sendNotif(DocumentReference authorReference, DocumentReference documentReference,
+    public void sendNotif(String authorReference, String documentReference,
         DocumentReference mentionedIdentity, MentionLocation location, String anchorId)
     {
         MentionEventParams params = new MentionEventParams()
-                                        .setUserReference(authorReference.toString())
-                                        .setDocumentReference(documentReference.toString())
+                                        .setUserReference(authorReference)
```
surli, line 57: So it was wrong to use `toString` before, now it would still make sense to have the method signature using real reference types, and to use the serializer here.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionsBlockingQueueProvider.java
```diff
@@ -0,0 +1,44 @@
...
+import org.xwiki.mentions.internal.async.MentionsData;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Blocking queue provider.
+ *
+ * @version $Id$
+ * @since 12.6RC1
+ */
+@Unstable
```
surli, line 34: Internal, no need.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionsDataConsumer.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.xpn.xwiki.XWikiException;
+
+/**
+ * Consume a mention data and analyse it.
+ *
+ * @version $Id$
+ * @since 12.6RC1
+ */
+@Unstable
```
surli, line 34: Still internal.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionsThreadPoolProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provide a thread pool to analyze the mentions.
+ *
+ * @version $Id$
+ * @since 12.6RC1
+ */
+@Unstable
```
surli, line 33: Internal.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-test/xwiki-platform-mentions-test-docker/pom.xml
```diff
@@ -46,6 +46,13 @@
       <type>xar</type>
     </dependency>
 
+    <!--  set an explicit version to avoid inconsistencies -->
```
surli, line 49: Isn't it a dependency available in mentions-notifications?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
```diff
@@ -0,0 +1,300 @@
...
+            if (doc.getPreviousVersion() != null) {
+                oldDoc = this.documentRevisionProvider
+                             .getRevision(dr, doc.getPreviousVersion());
+            } else {
+                oldDoc = null;
+            }
+            DocumentReference documentReference = doc.getDocumentReference();
+            DocumentReference authorReference = doc.getAuthorReference();
+
+            String documentReferenceStr = documentReference.toString();
```
surli, line 136: So that's not good, you should use a call to a serializer. Now I still think that it should be serialized in sendNotif and you should continue to propagate DocumentReference type.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
```diff
@@ -0,0 +1,300 @@
...
+import static java.util.Optional.ofNullable;
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
+import static org.xwiki.annotation.Annotation.SELECTION_FIELD;
+import static org.xwiki.mentions.MentionLocation.ANNOTATION;
+import static org.xwiki.mentions.MentionLocation.AWM_FIELD;
+import static org.xwiki.mentions.MentionLocation.COMMENT;
+
+/**
+ * Default implementation of {@link MentionsDataConsumer}.
```
surli, line 65: This is a complex one. A quick explanation of what it actually does would be useful IMO.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/DefaultMentionsBlockingQueueProvider.java
```diff
@@ -0,0 +1,53 @@
...
+public class DefaultMentionsBlockingQueueProvider implements MentionsBlockingQueueProvider
+{
+    @Inject
+    private Environment environment;
+
+    @Override
+    public BlockingQueue<MentionsData> initBlockingQueue()
+    {
+        File queueDirectory = new File(new File(this.environment.getPermanentDirectory(), "mentions"), "queue");
+        return new XodusBlockingQueue<>(queueDirectory.getAbsolutePath(), MentionsData.class, Long.MAX_VALUE);
```
surli, line 51: Why do you need this specific blocking queue implem?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsThreadFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static java.lang.Thread.NORM_PRIORITY;
+
+/**
+ * {@link ThreadFactory} dedicated to asynchronous tasks needed for the mentions.
+ *
+ * @version $Id$
+ * @since 12.6RC1
+ */
+class MentionsThreadFactory implements ThreadFactory
```
surli, line 33: Could have been an inner class in MentionThreadPoolExecutor.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/DefaultMentionsBlockingQueueProvider.java
tmortagne: I guess and a filesystem queue but it would worth some explanation.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/internal/MentionsEventExecutor.java
```diff
@@ -0,0 +1,58 @@
...
+    void execute(DocumentReference documentReference, DocumentReference authorReference, String version);
+
+    /**
+     *
+     * @return the current size of the queue of mentions to analyze.
+     */
+    long getQueueSize();
+
+    /**
+     * Clear the queue of mentions to analyze. This operation is only allowed for users with
```
surli, line 53: That's not good, when I mentioned about checking rights I meant exactly inside the script service: anyone with Script right can use the script service, now interacting directly with the Java API means that you might already have Programming rights so it's not really useful to perform checks there :) 
So basically you should just add an explicit check in the script service, you might find some examples in the various script services we have.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        /**
+         * Consume a single queue task.
+         */
+        public void runOnce()
+        {
+            MentionsData data = null;
+            try {
+                data = DefaultMentionsEventExecutor.this.queue.take();
```
surli, line 137: Not sure about this: there's always a data to take? Since it's running indefinitely isn't there some cases where the data is null?

manuelleduc: That's why we are using a BlockingQueue here, the take method is waiting passively if no element is available.

> Retrieves and removes the head of this queue, waiting if necessary until an element becomes available.
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingQueue.html#take--

surli: Makes sense. Thanks

surli: Just for info, the name of the method `runOnce` is a bit misleading then :) You could put everything in the while loop of run method.

surli:
@tmortagne waiting your review for merging.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
manuelleduc: It just a convenient way to isolate an iteration. That's mainly for testing purpose.

surli: Ok then.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -0,0 +1,159 @@
...
+ * analysis.  
+ *
+ * @version $Id$
+ * @since 12.6RC1
+ */
+@Component
+@Singleton
+public class DefaultMentionsEventExecutor implements MentionsEventExecutor, Initializable, Disposable
+{
+    private ThreadPoolExecutor executor;
```
tmortagne, line 58: I don't really understand why you use an executor since you don't really use it as such. From what I understand all you do in the end is starting several persistent threads.

By the way you don't shutdown the executor or trigger a stop of the threads in dispose().

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -55,7 +54,11 @@
 @Singleton
 public class DefaultMentionsEventExecutor implements MentionsEventExecutor, Initializable, Disposable
 {
-    private ThreadPoolExecutor executor;
+    static final int NB_THREADS = 4;
```
surli, line 57: Sounds like something that should be configurable.

surli: And 4 threads for Mentions by default sounds like a lot IMO...

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -119,27 +129,32 @@ public long getQueueSize()
...
+        public void consume()
         {
             MentionsData data = null;
             try {
-                data = DefaultMentionsEventExecutor.this.queue.take();
-                DefaultMentionsEventExecutor.this.logger
-                    .debug("[{}] - [{}] consumed. Queue size [{}]", data.getDocumentReference(), data.getVersion(),
-                        DefaultMentionsEventExecutor.this.queue.size());
+                // slightly active wait in order to let the loop be stopped when halt is set to true, once every second.
+                data = DefaultMentionsEventExecutor.this.queue.poll(10, SECONDS);
```
surli, line 150: Doc says every second but you specified 10 seconds?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/jmx/JMXMentionsMBean.java
```diff
@@ -37,4 +37,16 @@
      * Clear the mentions analysis tasks queue.
      */
     void clearQueue();
+
+    /**
+     * Update the number of thread consuming the user mentions analysis tasks.
+     *
+     * @param number the new number of threads.
+     */
+    void updateThreadNumber(int number);
```
surli, line 46: Well that's an interesting approach to allow changing dynamically the number of threads, but that's not really what I had in mind :) was more thinking about a config in xwiki.properties to specify the number of threads to use as startup, as we have for notifications for example.

surli: I'd say that we can keep your implem but we should still have the default number of threads computed from a property.

manuelleduc: mentions.poolSize property added. I've kept the dynamic jmx configuration for now.

xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.properties.vm
```diff
@@ -1033,6 +1033,18 @@ edit.defaultEditor.org.xwiki.rendering.block.XDOM#wysiwyg=$xwikiPropertiesDefaul
...
+# Mentions
+#-------------------------------------------------------------------------------------
+
+#-# [Since 12.6RC1]
+#-# The detection and notification for the user mentions in created or edited content is done asynchronously 
+#-# by a pool of consumers.
+#-# This properties controls the size of this pool.
+#-#
+#-# The default is :
+# notifications.poolSize = 2
```
surli, line 1046: Still not sure about this default size: it's something not critical so I'd allow the minimum of resources (i.e. 1 thread only). WDYT @tmortagne ?

surli:
Unless you have other comments on it @tmortagne else I'll merge it.

xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.properties.vm
tmortagne: I don't think I would have even implemented a pool system :)

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        /**
+         * Consume a single queue task.
+         */
+        public void consume()
+        {
+            MentionsData data = null;
+            try {
+                // slightly active wait in order to let the loop be stopped when halt is set to true, once every
```
tmortagne, line 181: An alternative is to put in the queue an elements used to both wake up the thread and order it to stop.

manuelleduc: In theory, by putting "stop" element in the queue for each consumer, they would all be stopped one by one.
The only issue is if the server halt in the middle of the stopping process. In this case some stop elements would be persisted on disk (since the queue is persistent). The works would stop their work when started again. But I guess it can be prevented by cleaning the queue of unwanted elements at startup.

tmortagne: > In this case some stop elements would be persisted on disk (since the queue is persistent). The works would stop their work when started again.

Not if you use == on a static to check if a task is a STOP task (but you will still need to ignore "bad" tasks).

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -108,31 +107,14 @@ public void dispose()
     {
         this.jmxRegistration.unregisterMBean(MBEAN_NAME);
         for (MentionsConsumer consumer : this.consumers) {
+            // We ask the consumers to stop directly with halt, in case their is many elements in the queue.
+            // Additionally we fill the queue with stop element in order to wake up consumers and notify them
+            // to stop if the queue is empty.
             consumer.halt();
+            this.queue.add(new MentionsData().stop());
```
tmortagne, line 114: Does not make much sense to create a new instance of MentionsData to return a static.

xwiki-platform-core/xwiki-platform-mentions/pom.xml
```diff
@@ -43,12 +43,49 @@
...
     <module>xwiki-platform-mentions-ui</module>
     <module>xwiki-platform-mentions-default</module>
   </modules>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.revapi</groupId>
+        <artifactId>revapi-maven-plugin</artifactId>
+        <configuration>
+          <analysisConfiguration><![CDATA[
```
surli, line 67: Revapi ignores are put in xwiki-platform-core/pom.xml so they can be processed all at once during release. And I think the new format for them is XML but maybe it changed again lately.

tmortagne: Yes it's XML and that's also what revapi suggest you in the error AFAIK.

PR #1336, opened 2020-07-10
XWIKI-17518: There should be no mentions in a clean wiki for the Admin user

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17518

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/WebHome.xml
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/WebHome.xml
@@ -36,7 +36,5 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content>To mention a user on a WYSIWYG field, type the ##@## character followed by the name of the user to mention. Then you can select a user in the presented list, once selected the mention is inserted in the text.
-
-For instance, typing ##@adm## should propose to mention the Administrator.</content>
+  <content>This space contains the code of the User Mentions application.</content>
 </xwikidoc>
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/WebHome.xml
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/WebHome.xml
@@ -36,7 +36,7 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/WebHome.xml
```diff
@@ -36,5 +36,7 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content>{{mention reference="XWiki.Admin" style="FULL_NAME" anchor="mention-example-1" /}} {{mention reference="XWiki.Admin" style="FIRST_NAME" anchor="mention-example-2" /}} {{mention reference="XWiki.Admin" style="LOGIN" anchor="mention-example-3" /}}</content>
+  <content>To mention a user on a WYSIWYG field, type the {{{@}}} character followed by the name of the user to mention. Then you can select a user in the presented list, once selected the mention is inserted in the text.
+
+For instance, typing {{{@adm}}} should propose to mention the Administrator.</content>
```
vmassol, line 41: would be better to use monospace:

```
...type the ##@## character...

... typing ##@adm## should...
```

However, I don't think this page should be about WYSIWYG. It's the macro page so if you want to put some text it should be about the usage of the macro only.

vmassol: Actually it might be good to just mention that this page contains the definition of the mentions macro.

It's also possible to have a simple example on how to use it but I wouldn't link to the documentation though (as we don't do that in general and the link can become broken).

WDYT?

surli: > Actually it might be good to just mention that this page contains the definition of the mentions macro.

Could be enough, I think the important thing to do is to add a tips for using mentions.

manuelleduc: The Macro is in XWiki.Mentions.MentionsMacro and it content is currently empty. The goal of the content of XWiki.Mentions.WebHome was just to show some very basic information about the nominal use of the extension from a user point of view (ie, using the wysiwyg).

But I can also add a short text to XWiki.Mentions.MentionsMacro explaining that it contains the macro definition.

WDYT ?

manuelleduc: I like the idea of a tip. If it's ok for everybody i'll create an issue.

surli: >  The goal of the content of XWiki.Mentions.WebHome was just to show some very basic information about the nominal use of the extension from a user point of view (ie, using the wysiwyg).

It should be very rare that people ends up there, so you can just indicate that this space contains the code for Mentions, should be enough. (That's usually what we do)

PR #1338, opened 2020-07-15
XWIKI-17553: Mention summary on the notification

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-17553


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/rendering/PlainTextMentionsBlockRenderer.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.rendering.internal.renderer.plain.PlainTextBlockRenderer;
+import org.xwiki.rendering.renderer.PrintRendererFactory;
+
+/**
+ * Mentions specific plain text block renderer.
+ *
+ * @version $Id$
+ * @since X.Y.Z
```
manuelleduc, line 34: I'll fix this in the next push

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultQuoteService.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private BlockMatcher mentionsBlocksMatcher(String anchorId)
+    {
+        return block -> {
+            boolean ret;
+            if (block instanceof MacroBlock) {
+                ret = ((MacroBlock) block).getId().equals("mention")
+                          && Optional.ofNullable(block.getParameter("anchor"))
+                                 .map(it -> it.equals(anchorId)).orElse(false);
```
surli, line 68: Not sure if I agree with this `orElse(false)`: note that the anchor are an optional parameter. So their value might be null or empty. In that case we cannot distinguish two mentions on the same page with an empty anchor of course, but at least we should take the first one.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionNotificationService.java
```diff
@@ -35,13 +36,14 @@
...
+     *  @param authorReference the reference of the author of the mention.
      * @param documentReference the document in which the mention has been done.
      * @param mentionedIdentity the identity of the mentioned user.
      * @param location The location of the mention.
      * @param anchorId The anchor link to use.
+     * @param xdom the content xdom
      */
-    void sendNotif(DocumentReference authorReference, DocumentReference documentReference,
-        DocumentReference mentionedIdentity, MentionLocation location, String anchorId);
+    void sendNotification(DocumentReference authorReference, DocumentReference documentReference,
```
surli, line 46: instead of adding more and more arguments there, an object should probably be sent.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm
```diff
@@ -41,6 +41,9 @@
     <td class="description">$summary</td>
     #insertEventDate($event)
 </tr>
+    <tr>
```
surli, line 44: since the quote might be null I guess it would need a if to not display this line at all if it's the case

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultQuoteService.java
```diff
@@ -0,0 +1,124 @@
...
+        @Override
+        public boolean match(Block block)
+        {
+            boolean ret;
+            if (block instanceof MacroBlock) {
+                ret =
+                    ((MacroBlock) block).getId().equals("mention")
+                        && Optional.ofNullable(block.getParameter("anchor"))
+                               .map(it -> it.equals(this.anchorId))
+                               .orElse(false);
```
surli, line 97: sounds to me that if the anchor parameter is null you keep returning false here.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm
```diff
@@ -41,6 +41,15 @@
     <td class="description">$summary</td>
     #insertEventDate($event)
 </tr>
+#if($services.mentions.isQuoteActivated() && $eventParams.getQuote())
+  <tr>
+    <td colspan="3">
+      <blockquote style="font-size: inherit">
```
surli, line 47: This style should be in a dedicated CSS.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionNotificationService.java
```diff
@@ -49,17 +49,32 @@
...
+    @Inject
+    private MentionsConfiguration configuration;
+
     @Override
-    public void sendNotif(DocumentReference authorReference, DocumentReference documentReference,
-        DocumentReference mentionedIdentity, MentionLocation location, String anchorId)
+    public void sendNotification(MentionNotificationParameters mentionNotificationParameters)
     {
+        String extractedQuote;
+        if (this.configuration.isQuoteActivated()) {
```
surli, line 62: I'm not entirely sure about that one: right now, if you had the config off and you enable it later, it means that all the already computed mentions won't have any associated quote. Only the new ones will have quotes. 
On the other hand, without this if you might compute quotes that you will never display if you don't change the config.

IMO it really depends why we have this config: if this is for perf reason, then I agree we should have this if. If it's for display reason, then we shouldn't and it should be only a switch for the velocity template.

surli:
@manuelleduc you need to rebase this PR.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionNotificationService.java
manuelleduc: imo notifications are transient and consequently it it not very useful to compute a quote which will never be display during the short lifespan during which the notification might be displayed to the user.
But we can easily separate the settings, one for the generation of the quotes (regarding performances), and a second one for the quotes display (regarding UX).

surli: > But we can easily separate the settings, one for the generation of the quotes (regarding performances), and a second one for the quotes display (regarding UX).

A bit overkill for now. I'm ok for keeping it like that, we just need to document properly the behaviour of this option.

PR #1341, opened 2020-07-16
XWIKI-17559: Allow to register a right as an implied right of an already existing right

surli:
Jira issue: https://jira.xwiki.org/browse/XWIKI-17559

  * Add a new API to register a right with a set of existing Rights that
    should imply it
  * Refactor existing register method so that by default Admin Right
    automatically imply the newly registered right
  * Refactor Right impliedRight to make it mutable internally
  * Add a new disabled test to check changes: disabled because Rights
    are static and it currently breaks integrations tests
  * Add more fine-grained checkstyle ignore and refactor checkstyle
    problems on DefaultSecurityCache based on that

### DIFF
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
@@ -19,7 +19,9 @@
  */
 package org.xwiki.security.authorization;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import javax.inject.Inject;
@@ -48,6 +50,11 @@ import org.xwiki.security.internal.XWikiBridge;
 @Singleton
 public class DefaultAuthorizationManager implements AuthorizationManager
 {
+    /**
+     * List of rights that should imply any newly registered right.
+     */
+    private static final List<Right> DEFAULT_IMPLIED_BY_RIGHTS = Arrays.asList(Right.ADMIN, Right.PROGRAM);
+
     /** Logger. **/
     @Inject
     private Logger logger;
@@ -178,7 +185,7 @@ public class DefaultAuthorizationManager implements AuthorizationManager
     public Right register(RightDescription rightDescription) throws UnableToRegisterRightException
     {
         // By default Admin should imply any newly registered right.
-        return register(rightDescription, Collections.singleton(Right.ADMIN));
+        return register(rightDescription, new HashSet<>(DEFAULT_IMPLIED_BY_RIGHTS));
     }
 
     @Override
@@ -186,7 +193,11 @@ public class DefaultAuthorizationManager implements AuthorizationManager
         throws UnableToRegisterRightException
     {
         try {
-            Right newRight = new Right(rightDescription, impliedByRights);
+            // Ensure that the default implied by rights are in the given set.
+            Set<Right> augmentedImpliedByRights = new HashSet<>(impliedByRights);
+            augmentedImpliedByRights.addAll(DEFAULT_IMPLIED_BY_RIGHTS);
+
+            Right newRight = new Right(rightDescription, augmentedImpliedByRights);
             // cleanup the cache since a new right scheme enter in action
             securityCache.remove(securityReferenceFactory.newEntityReference(xwikiBridge.getMainWikiReference()));
             return newRight;
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
@@ -19,6 +19,9 @@
  */
 package org.xwiki.security.authorization;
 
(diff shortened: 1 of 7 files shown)


xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
```diff
@@ -173,9 +176,17 @@ private boolean evaluateSecurityAccess(Right right, DocumentReference userRefere
 
     @Override
     public Right register(RightDescription rightDescription) throws UnableToRegisterRightException
+    {
+        // By default Admin should imply any newly registered right.
+        return register(rightDescription, Collections.singleton(Right.ADMIN));
```
tmortagne, line 181: PROGRAM too.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/Right.java
```diff
@@ -162,6 +162,11 @@
     /** Additional rights implied by this right. */
     private final Set<Right> impliedRights;
 
+    /**
+     * Mutable instance of the Additional rights implied by this right.
```
tmortagne, line 166: You kept the old javadoc ;)

surli: Woups

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
surli: ok adding it

surli: Note that the current logic is to add Admin (and Program) only if the list of impliedBy right is not given. If a list is given I consider that the developer chose to not specify Admin/Program right. That sounds good to you?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/Right.java
```diff
@@ -373,7 +413,10 @@ public String toString()
     @Override
     public Set<Right> getImpliedRights()
     {
-        return impliedRights;
+        if (this.immutableImpliedRights == null && !impliedRights.isEmpty()) {
```
tmortagne, line 416: The if should only have `this.immutableImpliedRights == null`. If the set is empty anyone will be able to add new rights in it through the Set returned by getImpliedRights() returns.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
tmortagne: Hmm not sure. Feels weird to have a right which is not implied by PROGRAM, I'm pretty sure people will forget to add PROGRAM and ADMIN when they want their right to be in implied by EDIT.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/Right.java
surli: >  If the set is empty anyone will be able to add new rights in it through the Set returned by getImpliedRights() returns.

I don't understand: the idea here is to keep returning null if `impliedRights` has always been empty.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
surli: So you suggest to always force adding Program and Admin in register?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/Right.java
tmortagne: Right, I misread. For some reason in my mind it meant we would return impliedRights directly if it was empty but immutableImpliedRights is always null when impliedRights is empty.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java
tmortagne: It should definitely be force for PROGRAM since that's kind of the definition of PROGRAM. Admin I'm not sure.

surli: Forcing them both for now.

PR #1342, opened 2020-07-17

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17524


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -85,12 +85,17 @@ public void initialize()
     {
         this.queue = this.blockingQueueProvider.initBlockingQueue();
         this.consumers = new ArrayList<>();
+        JMXMentionsMBean mbean = new JMXMentions(this.queue, () -> this.consumers.size());
+        this.jmxRegistration.registerMBean(mbean, MBEAN_NAME);
+    }
+
+    @Override
+    public void startThreads()
```
surli, line 93: what happens if you call this several times? It will recreate again and again the asked number of threads. You should have some flag to specify that it's already started no?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -80,19 +80,33 @@
...
-            startConsumer();
-        }
         JMXMentionsMBean mbean = new JMXMentions(this.queue, () -> this.consumers.size());
         this.jmxRegistration.registerMBean(mbean, MBEAN_NAME);
     }
 
+    @Override
+    public void startThreads()
+    {
+        if (!this.threadStarted) {
```
surli, line 97: Sounds like you forgot to remove this one

manuelleduc: This one is to prevent to enter the synchronized block when threadStarted is already set to true, for performance reason. Otherwise it is necessary to enter the synchronized block at every call.
But since this method is not supposed to be called very often anyway, it could be removed from simplicity's sake

surli: I think it's better to remove it yes.

PR #1346, opened 2020-07-20
XWIKI-17529: Parsing of content for mentions analysis is wrongly using syntax xwiki/2.1

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17529


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
```diff
@@ -63,8 +65,7 @@
     private Logger logger;
 
     @Inject
-    @Named("xwiki/2.1")
-    private Parser parser;
+    private ComponentManager componentManager;
```
mflorea, line 68: I'm wondering if you shouldn't use https://extensions.xwiki.org/xwiki/bin/view/Extension/Component%20Module#HContextComponentManager instead.

PR #1348, opened 2020-07-21

manuelleduc:
- various fixes and defensive code added
- improved internal APIs

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
@@ -21,6 +21,7 @@ package org.xwiki.mentions.internal;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 
 import javax.inject.Inject;
@@ -45,7 +46,7 @@ import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMess
  * This class is in charge of the management of mentions task analysis.
  * First, when {@link DefaultMentionsEventExecutor#execute(DocumentReference, DocumentReference, String)} is called,
  * a {@link MentionsData} is added to a queue.
- * Then, a pool of {@link MentionsConsumer} workers is consuming the queue and delegate the actual analyis to
+ * Then, a pool of {@link MentionsConsumer} workers is consuming the queue and delegate the actual analysis to
  * {@link MentionsDataConsumer#consume(MentionsData)} which deals with actual implementation of the user mentions 
  * analysis.  
  *
@@ -87,8 +88,13 @@ public class DefaultMentionsEventExecutor implements MentionsEventExecutor, Init
     {
         this.queue = this.blockingQueueProvider.initBlockingQueue();
         this.consumers = new ArrayList<>();
-        JMXMentionsMBean mbean = new JMXMentions(this.queue, () -> this.consumers.size());
+        JMXMentionsMBean mbean = new JMXMentions(this::getQueueSize, this::clearQueue, () -> this.consumers.size());
         this.jmxRegistration.registerMBean(mbean, MBEAN_NAME);
+    }
+
+    private void clearQueue()
+    {
+        this.queue.clear();
     }
 
     @Override
@@ -131,16 +137,17 @@ public class DefaultMentionsEventExecutor implements MentionsEventExecutor, Init
     @Override
     public void execute(DocumentReference documentReference, DocumentReference authorReference, String version)
     {
+        MentionsData data = new MentionsData()
+                                .setDocumentReference(documentReference.toString())
+                                .setAuthorReference(Objects.toString(authorReference.toString(), ""))
+                                .setVersion(version)
+                                .setWikiId(documentReference.getWikiReference().getName());
         try {
-            this.queue.put(new MentionsData()
-                               .setDocumentReference(documentReference.toString())
-                               .setAuthorReference(authorReference.toString())
-                               .setVersion(version)
-                               .setWikiId(documentReference.getWikiReference().getName())
-            );
+            this.queue.put(data);
         } catch (InterruptedException e) {
             this.logger
-                .warn("Error while adding a task to the mentions analysis queue. Cause [{}]", getRootCauseMessage(e));
+                .warn("Error while adding a task to the mentions analysis queue. Cause [{}]",
+                    getRootCauseMessage(e));
         }
     }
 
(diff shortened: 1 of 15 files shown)


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -131,16 +137,17 @@ public void dispose()
     @Override
     public void execute(DocumentReference documentReference, DocumentReference authorReference, String version)
     {
+        MentionsData data = new MentionsData()
+                                .setDocumentReference(documentReference.toString())
```
mflorea, line 141: `toString()`` should be used only when you need to "print" the reference (like in a log message). Otherwise, if some other code may be resolving the string reference later then you should use a proper reference serializer.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -131,16 +137,17 @@ public void dispose()
...
+                                .setAuthorReference(Objects.toString(authorReference, ""))
```
mflorea, line 142: Same as above.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -189,13 +196,18 @@ public void consume()
                 DefaultMentionsEventExecutor.this.logger
                     .warn("Error during mention analysis of task [{}]. Cause [{}].", data, getRootCauseMessage(e));
                 if (data != null) {
-                    // push back the failed task at the beginning of the queue.
-                    try {
-                        DefaultMentionsEventExecutor.this.queue.put(data);
-                    } catch (InterruptedException interruptedException) {
+                    if (!data.isFailed()) {
+                        // push back the failed task at the beginning of the queue.
```
mflorea, line 200: The comment is a bit confusing because you're testing for not failed and the comment says it failed.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsData.java
```diff
@@ -168,6 +170,23 @@ public MentionsData setWikiId(String wikiId)
...
+     * Increase the attempts counter.
+     */
+    public void increaseAttempts()
+    {
+        this.attempts++;
+    }
+
+    /**
+     *
+     * @return true if the number of attempts for the current data has reached 10.
```
mflorea, line 183:
```suggestion
     * @return {@code true} if the number of attempts for the current data has reached 10. 
```

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsData.java
```diff
@@ -168,6 +170,23 @@ public MentionsData setWikiId(String wikiId)
...
+    public void increaseAttempts()
+    {
+        this.attempts++;
+    }
+
+    /**
+     *
+     * @return true if the number of attempts for the current data has reached 10. 
+     */
+    public boolean isFailed()
```
mflorea, line 185: I would try to find a better name. See also my comment above (when you put back the task at the beginning of the queue). If you need to read the JavaDoc to understand what the method does then it means something is not right.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/test/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutorTest.java
```diff
@@ -107,6 +112,30 @@ void execute() throws InterruptedException
                      .setVersion("1.0"));
     }
 
+    @Test
+    void executeInterruptedException() throws InterruptedException
+    {
+
```
mflorea, line 118: Useless empty line.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
manuelleduc: Replaced by `EntityReferenceSerializer<String>`

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MentionsData.java
manuelleduc: Renamed `tooManyAttempts()`

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/test/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutorTest.java
```diff
@@ -125,10 +167,44 @@ void consume() throws Exception
...
+        assertEquals(ERROR, this.logCapture.getLogEvent(22).getLevel());
+        IntStream.of(0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20)
+            .forEach(
+                i -> assertEquals("[xwiki:XWiki.Doc] - [1.0] consumed. Queue size [0]", this.logCapture.getMessage(i)));
+        IntStream.of(1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21)
+            .forEach(i -> assertEquals("Error during mention analysis of task "
+                                           + "[documentReference = [xwiki:XWiki.Doc], version = [1.0], authorReference = [xwiki:XWiki.Author], wikiId = [xwiki]]. "
+                                           + "Cause [RuntimeException: ...].", this.logCapture.getMessage(i)));
+
+        assertEquals("[documentReference = [xwiki:XWiki.Doc], version = [1.0], authorReference = [xwiki:XWiki.Author], wikiId = [xwiki]] abandoned because it has failed to many times.", this.logCapture.getMessage(22));
```
mflorea, line 208: Are you sure this doesn't break checkstyle (line too long, try build with ``-Pquality``)? I guess it's because we don't verify checkstyle on tests. It would still be nice to not exceed the 120 char limit.

PR #1358, opened 2020-08-12
XWIKI-17636: Use a lighter library to perform disk-persistent blocking queue in mention

surli:
* Provide a dedicated Map-based implementation of LinkedBlockingQueue
  * Use H2 MVStore as the underlying lib to persist the map of data on
    disk

(marking as draft since it's still missing the unit tests)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
@@ -1,0 +1,364 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.mentions.internal.async;
+
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * A dedicated implementation of {@link BlockingQueue} internally based on a {@link LinkedBlockingQueue} and on
+ * a {@link ConcurrentMap}indexed by {@code Long}.
+ * The idea is to be able to obtain a {@link java.util.concurrent.BlockingQueue} from a {@code Map} persisted on disk.
+ * @param <T> The type of object manipulated by this queue.
+ *
+ * @version $Id$
+ * @since 12.7RC1
+ * @since 12.6.1
+ */
+public class MapBasedLinkedBlockingQueue<T> implements BlockingQueue<T>
+{
+    private final ConcurrentMap<Long, T> internalMap;
+    private final AtomicLong highestKey;
+    private final LinkedBlockingQueue<Pair<Long, T>> internalQueue;
+
+    /**
+     * Default constructor. It uses the provided map to populate the queue information, and then keep in sync the
+     * map and the queue.
+     *
+     * @param internalMap the map in charge of the disk persistency.
+     */
+    public MapBasedLinkedBlockingQueue(ConcurrentMap<Long, T> internalMap)
+    {
+        super();
+        this.internalMap = internalMap;
+        this.highestKey = new AtomicLong(computeHighestKey());
+        this.internalQueue = new LinkedBlockingQueue<>();
+
+        // if the map contains some element, we want to populate the actual queue with those info.
+        this.populateUnderlyingQueue();
+    }
+
+    /**
+     * Compute from the internal map what's the highest key number.
+     *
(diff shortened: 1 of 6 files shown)


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
```diff
@@ -0,0 +1,270 @@
...
+     *
+     * @param t the element to remove.
+     */
+    private void internalRemove(Object t)
+    {
+        if (t != null) {
+            this.internalMap.remove(t);
+
+            // decrement the key if we removed the highest key element.
+            if (!this.internalMap.containsKey(this.highestKey.get())) {
```
tmortagne, line 139: I don't understand this logic. What's removed is the first element of the queue so the element associated with the lowest key and not the highest.

surli: It's an internal remove method to be called whenever an element is removed from the queue, so it might be the head or the tail of the queue. Since I don't want to make prediction, I'm just checking that the highestkey is still present or not to update it.

tmortagne: My main issue here is that you don't know the key so you can't remove the object from the map, `this.internalMap.remove(t);` makes no sense since ti is a value and not a key.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
```diff
@@ -0,0 +1,364 @@
...
+    {
+        this.internalMap.clear();
+        this.highestKey.set(0);
+    }
+
+    @Override
+    public void put(T t) throws InterruptedException
+    {
+        Long key = this.internalPut(t);
+        this.internalQueue.put(Pair.of(key, t));
```
tmortagne, line 168: I would put in the key before the map so that it's available faster to whoever is waiting for new entries.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
```diff
@@ -0,0 +1,364 @@
...
+    public boolean contains(Object o)
+    {
+        // We check on the map since most likely the given object is not a Pair but a T.
+        return this.internalMap.containsValue(o);
+    }
+
+    @Override
+    public Iterator<T> iterator()
+    {
+        return this.internalMap.entrySet().stream().sorted(new DefaultComparator()).map(Map.Entry::getValue).iterator();
```
tmortagne, line 270: Why not `internalQueue.iterator()` ?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    @Override
+    public Object[] toArray()
+    {
+        return this.internalMap.values().toArray();
```
tmortagne, line 276: Same question, why use the map here ? The map should be used only to add/remove elements IMO.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
surli: Because the internal queue is of type `Pair<Long, T>` where I need an Iterator of T.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
surli: It might also lead to some inconsistency: you would retrieve faster the new entry to whoever is waiting, but since you would use the same blocking queue you'd need to remove the key from the map. So you could have a race condition leading to having the entry in the map while it has already been consumed.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
surli: Same answer, because of types. Here it's a bit less of an issue since it returns Object[] but I don't think the users would expect to have Pair in the result instead of T.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
tmortagne: You could map it. It would still be much faster than manipulating serialized elements.

By the way to be complete you should also wrap the Iterator because of Iterator#remove().

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
tmortagne: Same answer that above, you can easily use stream mapping and it would be much faster than the map that might be slow (definitely is when the map is on filesystem).

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/MapBasedLinkedBlockingQueue.java
tmortagne: Yes you are right, it's safer like this.

PR #1360, opened 2020-08-15
XWIKI-17649: The DefaultMentionsEventExecutor produces 100% CPU load …

ClemensRobbenhaar:
seems a simple switch to another method solves the issue.


manuelleduc:
iirc, the previous queue implementation was passively waiting when no data were available, even on `poll`. Looking at the documentation of `Queue`, `take` seems to be more adequate.

surli:
Thanks @ClemensRobbenhaar

PR #1370, opened 2020-08-26
XWIKI-17699 Page tree is not sorted by title when using Oracle DB

ClemensRobbenhaar:
Added a condition for `doc.XWD_LANGUAGE is null` beside of `doc.XWD_LANGUAGE = ''` to handle cases where the DB stores a null in the language field instead of an empty string.
Oracle does that; see also comment at https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xwiki.oracle.hbm.xml#L37

As I am not fully comfortable with the document tree part of the code it would be nice if someone else could double check the two changes, small as they are.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-api/src/main/resources/org/xwiki/index/tree/internal/nestedpages/query/queries.hbm.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-api/src/main/resources/org/xwiki/index/tree/internal/nestedpages/query/queries.hbm.xml
@@ -72,7 +72,7 @@
             left outer join xwikidoc doc on (
                 doc.XWD_WEB = XWS_REFERENCE and
                 doc.XWD_NAME = 'WebHome' and
-                doc.XWD_LANGUAGE = '')
+                (doc.XWD_LANGUAGE = '' OR doc.XWD_LANGUAGE is NULL))
             left outer join xwikidoc tdoc on (
                 tdoc.XWD_WEB = XWS_REFERENCE and
                 tdoc.XWD_NAME = 'WebHome' and
@@ -134,7 +134,7 @@
         left outer join xwikidoc doc on (
             doc.XWD_WEB = XWS_REFERENCE and
             doc.XWD_NAME = 'WebHome' and
-            doc.XWD_LANGUAGE = '')
+            (doc.XWD_LANGUAGE = '' OR doc.XWD_LANGUAGE is NULL))
         left outer join xwikidoc tdoc on (
             tdoc.XWD_WEB = XWS_REFERENCE and
             tdoc.XWD_NAME = 'WebHome' and
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-api/src/main/resources/org/xwiki/index/tree/internal/nestedpages/query/queries.hbm.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-api/src/main/resources/org/xwiki/index/tree/internal/nestedpages/query/queries.hbm.xml
@@ -55,7 +55,10 @@
   </resultset>
 
   <!-- Note that we use 0 and 1 instead of false and true because Oracle doesn't support the boolean constants. The
-    returned value is boolean though because we specify the type in the result set meta-data. -->
+    returned value is boolean though because we specify the type in the result set meta-data.
+    The double check on doc.XWD_LANGUAGE being the empty string or null to find out the original language page is needed
+    because Oracle stores empty strings as nulls unless a not-null constraint is present, which is not the case for the
+    language field. -->
   <!-- This query follows the SQL-2013 ANSI standard, but it uses a non-core feature F591 "Derived tables" which is
     implemented by most of the databases (HSQLDB, MySql, Oracle, PostgreSQL) -->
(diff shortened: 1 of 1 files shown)


surli (approved):
Sounds good to me but not an expert with Oracle DB or Tree code so not 100% sure. Maybe a good fit for a docker test since it would be tested on various DB including Oracle.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-api/src/main/resources/org/xwiki/index/tree/internal/nestedpages/query/queries.hbm.xml
```diff
@@ -72,7 +72,7 @@
             left outer join xwikidoc doc on (
                 doc.XWD_WEB = XWS_REFERENCE and
                 doc.XWD_NAME = 'WebHome' and
-                doc.XWD_LANGUAGE = '')
+                (doc.XWD_LANGUAGE = '' OR doc.XWD_LANGUAGE is NULL))
```
vmassol, line 75: We need a comment to explain these changes  so that we remember them. Thanks!

vmassol: It should also explain why it's not a problem with other DBs.. hmm I wonder if it is. I see in the HBM file that XWD_LANGUAGE could be null:

`<property name="language" type="string" column="XWD_LANGUAGE" index="DOC_LANGUAGE" length="5" />`  

?

tmortagne: It technically can be null but it never is in practice because XWikiDocument#getLanguage() never return null.

mflorea (approved):

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-api/src/main/resources/org/xwiki/index/tree/internal/nestedpages/query/queries.hbm.xml
mflorea: @ClemensRobbenhaar can you add the comment that @vmassol requested. Unfortunately you can't add it in-line but only before the full query. Still useful though.

ClemensRobbenhaar: Sorry for the delay. Comments have been added.

tmortagne: > unless a not-null constraint is present

Really ? I tough it Oracle was always behaving like this. So that would be that a cleaner fix would be to add this constraint then ? After all this is is never supposed to be null.

ClemensRobbenhaar: It sure would be a cleaner fix to add the constraint to the hibernate definition of the schema in ``.../xwiki-platform-oldcore/src/main/resources/xwiki.oracle.hbm.xml`` but I have no idea how to update the existing Oracle databases where currently a ``null`` is stored for the ``XWD_LANGUAGE`` field instead of an empty string.

tmortagne: Not sure about the exact request but it would be done in an hibernate migration.

ClemensRobbenhaar: I thought the hibernate migrations only run after the schema has been validated/updated which would be too late here, but I see some liquibase migrations can be run before that.
If I write the migration this might need some time as it is my first DB migration for XWiki and also testing it with Oracle might be ... time consuming.

tmortagne: Why too late ? The automated hibernate migrator is very (very) basic: all it does is adding missing columns. So what will happen is that the automatic migrator will run (and do nothing) and then in your migrator you will alter the column to make it not nullable.

ClemensRobbenhaar: Ah, I thought hibernate checks the schema at startup and fails if not all constraints are present. You see, I have only a very vague idea how the database migrations work in XWiki.

PR #1374, opened 2020-08-27
XWIKI-15833: Users shouldn't be able to pick a group as a subgroup of itself

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-15833


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/XWikiGroupSheet.xml
```diff
@@ -210,14 +215,20 @@
...
       // Clear the list of selected users and groups.
       $('#userInput')[0].selectize.clear(true);
       $('#groupInput')[0].selectize.clear(true);
       // Reload the live table.
       editgrouptable.refresh();
       // Show notification message.
-      new XWiki.widgets.Notification("$services.localization.render('xe.admin.groups.addSuccess')", 'done', {timeout: 5});
+      if(response.ignored.size() &gt; 0) {
+        var concat = response.ignored.join(', ');
+        new XWiki.widgets.Notification("$services.localization.render('xe.admin.groups.addIgnored') (" + concat + ")",
```
surli, line 227: sounds better to create a translation with a parameter than using a concat here. Besides "concat" is not really a good variable name.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
```diff
@@ -17,22 +17,49 @@
...
+## Checks if this is valid to add the addedGroup to the currentGroup.
+## This verification is based on three criteria:
+## - the current and added groups are not the same group
+## - the current group does not already have the added group in its sub-groups
+## - this current group is not sub-group of the added group
+##
+## The boolean result is set on the $accept variable.
+##
+#macro (checkGroups $currentGroup $addedGroup)
+  #set ($members = $services.user.group.getMembers($currentGroup, true))
```
surli, line 30: This recursive method can be very expensive, I'm not sure using a velocity macro for this is the right approach: IMO this kind of check should rely on some cache mechanisms to avoid computing back everything when it's possible. And it needs to have really good coverage, so I'd provide the check as a Java API.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
```diff
@@ -17,22 +17,49 @@
...
+  #set ($membersOf = $services.user.group.getGroupsFromAllWikis($currentGroup))
+  #set ($currentGroupDR = $services.model.resolveDocument($currentGroup))
+  #set ($addedGroupDR = $services.model.resolveDocument($addedGroup))
+  #set ($accept = $currentGroup != '' && $addedGroup != '' && $currentGroupDR != $addedGroupDR
```
surli, line 34: At the very least the tree first check should be performed first in your macro so you don't call getMembers and getGroupsFromAllWikis if you don't need to.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/XWikiGroupSheet.xml
tmortagne: +1 for parameters, you are never really sure if the placement you have in English is going to work for all locales.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
tmortagne: Not sure what you mean, I don't see any Velocity recursion here or maybe I missed it. `getMembers` is what is taking care of the recursion and it's a Java API which rely on caching. Sure we could add a `isMemberOf` helper but it's not going to change much in term of performance, but making the Velocity code more readable is good too.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
```diff
@@ -17,22 +17,49 @@
...
+## - the current and added groups are not the same group
+## - the current group does not already have the added group in its sub-groups
+## - this current group is not sub-group of the added group
+##
+## The boolean result is set on the $accept variable.
+##
+#macro (checkGroups $currentGroup $addedGroup)
+  #set ($members = $services.user.group.getMembers($currentGroup, true))
+  #set ($membersOf = $services.user.group.getGroupsFromAllWikis($currentGroup))
+  #set ($currentGroupDR = $services.model.resolveDocument($currentGroup))
```
tmortagne, line 32: You should move the resolves before the calls to getMembers and getGroupsFromAllWikis so that you don't end up doing them twice (once automatically and once explicitly).

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
tmortagne: +1, always order conditions from least expensive to most expensive

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
surli: > getMembers is what is taking care of the recursion and it's a Java API which rely on caching.

Was talking about this recursion not about a velocity recursion. I remembered it being expensive but maybe I'm wrong there. In general this `checkGroups` sounds more like something we'd want at Java level than at velocity level: we could need it when performing an add member at API level too for example.

tmortagne: > I remembered it being expensive but maybe I'm wrong there.

It could of course be expensive by definition if the member cache is totally empty and the entity is member of a lot of groups but there is not much else we can do, it should not be very noticeable in practice. We could of course have a helper in the API for checkGroups but with a much better name then :)

manuelleduc: I'm considering moving the logic to a method called `GroupScriptService#canAddNewChild(DocumentReference parent, DocumentReference child)`, and of course taking advantage of this code update to take into account the other remarks. WDYT?

tmortagne: Sounds good (not sure it's the best name but I'm not against it except that the naming everywhere is "member" and not "child"). Note that you should keep the check related to empty string on Velocity side since it does not make much sense after the conversion to DocumentReference.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/XWikiGroupSheet.xml
manuelleduc: I agree, the problem is that IIUC the velocity code is evaluated at the generation of the script, but the content to used as the translation parameter is only available at runtime.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
```diff
@@ -140,4 +140,29 @@
...
+        final boolean ret;
+        if (target == null || member == null || member.equals(target)) {
+            ret = false;
+        } else {
+            // check if the candidate member is not already one of the members of the target
+            if (getGroupsFromAllWikis(target).contains(member)) {
+                ret = false;
+            } else {
+                // check if the target is not a member of the member
+                ret = !getMembers(target, true).contains(member);
```
tmortagne, line 163: I don't really understand why you are using both `getGroupsFromAllWikis` and `getMembers`. I would use `getMembers` for both checks since it's generally much faster (especially in an instances with many wikis).

tmortagne (changes requested):

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
manuelleduc: I agree, I'll update to only use getMembers.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/XWikiGroupSheet.xml
manuelleduc: The translation is now done server side (see https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-675b8cf1e4313bb08807f805a03daa40R50)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
manuelleduc: The logic is now moved to a dedicated method in the groups script service (see https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-8dc9d53466040f00f533fc828198b1dfR154)

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
```diff
@@ -140,4 +140,27 @@
...
+     * @param candidate The group to be added
+     * @return {@code true} if the candidate can be added to the members of the target, {@code false} otherwise.
+     * @since 12.8RC1
+     */
+    public boolean canAddAsMember(DocumentReference target, DocumentReference candidate) throws GroupException
+    {
+        boolean ret;
+        if (target == null || candidate == null || candidate.equals(target)) {
+            ret = false;
+        } else if (getMembers(target, false).contains(candidate)) {
```
tmortagne, line 159: Should be `true` here, you want all the members of that group, not just the direct ones.

tmortagne: You you are right, I was too focused on the "make sure it's not already a member".

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2352,6 +2352,7 @@ xe.admin.groups.addUser=Users to add
 xe.admin.groups.addUser.submit=Add
 xe.admin.groups.addGroup=Subgroups to add
 xe.admin.groups.addSuccess=Members successfully added
+xe.admin.groups.addIgnored=Members successfully added but some groups have been ignored ({0})
```
surli, line 2355: So here the naming convention of https://dev.xwiki.org/xwiki/bin/view/Community/L10N%20Conventions is not respected. I guess if we change it we should also migrate the other ones in the future. You can check https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HMovingTranslations for this.

surli (changes requested):
Minor changes to be done mostly for doc and translation key.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
```diff
@@ -140,4 +140,27 @@
...
+
+    /**
+     * Checks if a group (candidate) can be added to the members of another group (target).
+     * A group can be added to the members of another group if neither group already has the other in its members.
+     * A group cannot be added the its own members.
+     *
+     * @param target The targeted group
+     * @param candidate The group to be added
+     * @return {@code true} if the candidate can be added to the members of the target, {@code false} otherwise.
+     * @since 12.8RC1
```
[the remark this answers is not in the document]
manuelleduc: Added

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
```diff
@@ -140,4 +140,27 @@
...
+     */
+    public boolean canAddAsMember(DocumentReference target, DocumentReference candidate) throws GroupException
```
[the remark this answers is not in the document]
manuelleduc: Added here https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-8dc9d53466040f00f533fc828198b1dfR159

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
```diff
@@ -140,4 +140,27 @@
     {
         return getMembers(group, true);
     }
+
+    /**
+     * Checks if a group (candidate) can be added to the members of another group (target).
+     * A group can be added to the members of another group if neither group already has the other in its members.
+     * A group cannot be added the its own members.
```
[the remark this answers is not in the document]
manuelleduc: You are right, javadoc updated (see https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-8dc9d53466040f00f533fc828198b1dfR146)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
```diff
@@ -20,19 +20,36 @@
...
       #set ($dirty = true)
+    #elseif($name != '')
+      #set ($discard = $ignoredGroups.add($name))
     #end
   #end
   #if ($dirty)
     #set ($discard = $doc.save())
   #end
-  #if (!$isAjaxRequest)
+  #if(!$isAjaxRequest)
```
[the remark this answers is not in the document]
manuelleduc: Fixed

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
manuelleduc: Updated to `web.groups.administration.groupsIgnored`
- Declaration: https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-fffbf37334c4d7a178b13f4ccc313f86R2358
- Use: https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-675b8cf1e4313bb08807f805a03daa40R50

vmassol: AFAIK we don't use the `web` prefix.

vmassol: Note: I see `web` was used at https://github.com/xwiki/xwiki-platform/commit/749f52d0870e7014142890b10045afcedb3dd252#diff-cf04ec11293bf524b894888450e2230cR357 (but that's a very old commit). So not sure but needs to be checked and decided.

manuelleduc: The most recent use of the web prefix are from 2020 (https://github.com/xwiki/xwiki-platform/commit/e0b57863b3d8ff57e1a07e060d763b88d1baad3a#diff-fffbf37334c4d7a178b13f4ccc313f86R1669) but I agree that they are rarely used recently. Most of them have been introduced in 2015.

Would `groups.administration.groupsIgnored` be acceptable even though we don't have a module named `xwiki-platform-groups` yet?

tmortagne: This commit seems to be a move of existing code so that translation key is much older.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
```diff
@@ -20,19 +20,36 @@
...
 #if ($services.csrf.isTokenValid($request.form_token))
   #set ($dirty = false)
+  #set ($ignoredGroups = [])
   #foreach ($name in $request.getParameterValues('name'))
     ## Verify if the user or group is already in the target group.
-    #if ($name.trim() != '' && !$doc.getObject('XWiki.XWikiGroups', 'member', $name))
+    #if ($name.trim() == '')
+      #set ($accept = false)
+    #else
+      #set ($accept = $services.user.group.canAddAsMember($name, $doc.getDocumentReference()))
```
surli, line 29: since canAddAsMember could throw exception I'd protect it, see: https://extensions.xwiki.org/xwiki/bin/view/Extension/Velocity%20Module#HVelocityDirectives

manuelleduc: Instead I have captured the exception inside canAddAsMember (see https://github.com/xwiki/xwiki-platform/pull/1374/files#diff-8dc9d53466040f00f533fc828198b1dfR176). WDYT ?

surli: Sounds ok to me, now I still have a question: this API is supposed to be called whenever you add group to another group, or whenever you add a member (user or group) in another group. In other word, is that possible that canAddAsMember will be called with $name being a reference to a user instead of a group and in that case how does it behave, in particular would the call to `getMembers(candidate, true)` throw an exception?

tmortagne: > in particular would the call to getMembers(candidate, true) throw an exception

AFAIK it should simply return an empty list because no XWiki.XWikiGroups object is found under this reference.

surli: Ok was expecting this but wanted to be sure.

surli (approved):

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/adduorg.vm
manuelleduc: You were right to ask the question @surli , when trying to add a use that is already a member, the warning message is `Members successfully added but some groups have been ignored (XWiki.U2)`. I'll revert it back to the current message (`Members successfully added`).

tmortagne (approved):

PR #1386, opened 2020-09-09
XWIKI-17771: UserReferenceConverter resolution is always done in the context of the main wiki

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17771


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/UserReferenceConverter.java
```diff
@@ -42,13 +46,17 @@
...
     @Override
     protected UserReference convertToType(Type targetType, Object value)
     {
         if (value == null) {
             return null;
         }
 
-        return this.userReferenceResolver.resolve(value.toString());
+        WikiReference currentWikiReference = this.xcontextProvider.get().getWikiReference();
+        return this.userReferenceResolver.resolve(value.toString(), currentWikiReference);
```
vmassol, line 60: hmmm not sure this is correct. Shouldn't we use a `current` `UserReferenceResolver<String>` implementation instead?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/test/java/org/xwiki/user/internal/UserReferenceConverterTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    void convertToType()
+    {
+        XWikiContext context = mock(XWikiContext.class);
+        when(this.xcontextProvider.get()).thenReturn(context);
+        when(context.getWikiReference()).thenReturn(new WikiReference("s1"));
+
+        this.converter.convertToType(mock(Type.class), "XWiki.U1");
+
+        verify(this.userReferenceResolver).resolve("XWiki.U1", new WikiReference("s1"));
```
vmassol, line 70: I dont understand why you use `verify()` instead of doing an `assert` on the result.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/UserReferenceConverter.java
vmassol: > Shouldn't we use a current UserReferenceResolver<String> implementation instead?

Actually this is already existing ;)

vmassol: Forget what I said, the fix is correct.

vmassol: Well there's only 1 potential improvement which is to get the current wiki from the ModelContext instead of from XWikiContext....

tmortagne: Seems to be the right fix would be to provide a UserReferenceResolver<String> with hint "current" and use that like we do for DocumentReferenceResolver for example.

manuelleduc: Logic changed, but after some though I might have misinterpreted you explanations. Do you propose to introduce a new class implementing `UserReferenceResolver<String>`?

vmassol: Thomas said exactly the same thing I said at https://github.com/xwiki/xwiki-platform/pull/1386#discussion_r485602221. However, I checked the code and `UserReferenceResolver`  is working differently than entity reference resolvers.

vmassol: The hardest part would be to choose a name for the `current` hint class since `CurrentUserReferenceResolver`  already exists but it's a resolver for `CurrentUserReference` and not for `UserReference` ;)

tmortagne: > Do you propose to introduce a new class implementing UserReferenceResolver<String>?

Yes.

vmassol: > The hardest part would be to choose a name for the current hint class since CurrentUserReferenceResolver already exists but it's a resolver for CurrentUserReference and not for UserReference ;) 

Actually we should call the existing class `CurrentUserReferenceUserReferenceResolver`. And the new one would be `CurrentUserReferenceResolver` (actually it would be `ConfiguredCurrentUserReferenceResolver` and forward to the impls).

vmassol: > Actually we should call the existing class CurrentUserReferenceUserReferenceResolver

And change the other impl for consistency ofc.

vmassol: > And the new one would be CurrentUserReferenceResolver

Actually it would be CurrentUserReferenceUserReferenceResolver... haha same name.. so it doesn't work for the naming part ;)

tmortagne: I would name the new one (implementation of `UserReferenceResolver<String>` with hint "current") `CurrentStringUserReferenceResolver` (following CurrentStringDocumentReferenceResolver logic).

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/document/CurrentStringUserReferenceResolver.java
```diff
@@ -0,0 +1,68 @@
...
+import org.xwiki.model.reference.EntityReferenceProvider;
+import org.xwiki.user.CurrentUserReference;
+import org.xwiki.user.UserReference;
+import org.xwiki.user.UserReferenceResolver;
+
+/**
+ * Converts a {@link String} representing a user id into a {@link UserReference}.
+ * <p>
+ * The conversion takes into account the current wiki.
+ * Thus, "XWiki.U1" will be resolved to "xwiki:XWiki.U1" on the main wiki, and to "s1:XWiki.U1" in the context of the
```
vmassol, line 38: This example and the code below is only valid for the "document" store for users. This is what I was saying in my comments: we need a generic implementation independent of the store!

Imagine for a second that we have a "file" store. Do you think the code above in UserReferenceConverter would work?

Thanks

vmassol (changes requested):

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/test/java/org/xwiki/user/internal/DefautConfiguredStringUserReferenceResolverTest.java
```diff
@@ -39,15 +39,16 @@
...
 /**
- * Unit tests for {@link ConfiguredStringUserReferenceResolver}.
+ * Unit tests for {@link DefautConfiguredStringUserReferenceResolver}.
  *
  * @version $Id$
+ * @since 12.8RC1
  */
 @ComponentTest
-public class ConfiguredStringUserReferenceResolverTest
+public class DefautConfiguredStringUserReferenceResolverTest
```
vmassol, line 48: Typo --> `Default`

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/test/java/org/xwiki/user/internal/DefautConfiguredStringUserReferenceResolverTest.java
```diff
@@ -39,15 +39,16 @@
...
 {
     @InjectMockComponents
-    private ConfiguredStringUserReferenceResolver resolver;
+    private DefautConfiguredStringUserReferenceResolver resolver;
```
vmassol, line 51: Same

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/test/java/org/xwiki/user/internal/UserReferenceConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        UserReference actual = this.converter.convertToType(mock(Type.class), "XWiki.U1");
+        assertEquals(userReference, actual);
+    }
+
+    @Test
+    void convertToTypeValueNull()
+    {
+        assertNull(this.converter.convertToType(mock(Type.class), null));
+
```
vmassol, line 68: extra new line!

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/test/java/org/xwiki/user/internal/document/CurrentDocumentStringUserReferenceResolverTest.java
```diff
@@ -0,0 +1,119 @@
...
+class CurrentDocumentStringUserReferenceResolverTest
+{
+    @InjectMockComponents
+    private CurrentDocumentStringUserReferenceResolver resolver;
+
+    @MockComponent
+    @Named("current")
+    private DocumentReferenceResolver<String> documentReferenceResolver;
+
+
```
vmassol, line 60: extra new lines

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/test/java/org/xwiki/user/internal/CurrentConfiguredStringUserReferenceResolverTest.java
```diff
@@ -0,0 +1,87 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
```
vmassol, line 20: extra new line

vmassol:
hmm I probably didn't do it correctly since there are now conflicts and I cannot close the PR. I made changes after applying the PR locally and push all to master....

vmassol:
@manuelleduc Here are my changes: https://github.com/xwiki/xwiki-platform/commit/15d993c444e37710061c73cb1226ebac48d4e199

PR #1387, opened 2020-09-09
XWIKI-17718: Duplicate user when editing the mentions macro in WYSIWYG

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17718

Currently the mention is fully qualified identifier (eg, "xwiki:XWiki.U1") to reference the mentioned user, even when the user is local to the wiki.
The user picker is expecting to see local users identified by their short qualifier (eg, "XWiki.U1").
Consequently,  when the user picker is displayed twice, once for the short form and once for the fully qualified one.

The solution is to use the short for the mention's identifier.
As a consequence, the wiki in which the mention occurred is now needed for the resolution of the user.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
@@ -40,6 +40,7 @@ import org.xwiki.component.manager.ComponentLookupException;
 import org.xwiki.component.manager.ComponentManager;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.DocumentReferenceResolver;
+import org.xwiki.model.reference.WikiReference;
 import org.xwiki.rendering.block.Block;
 import org.xwiki.rendering.block.MacroBlock;
 import org.xwiki.rendering.block.XDOM;
@@ -85,12 +86,13 @@ public class DefaultMentionXDOMService implements MentionXDOMService
     }
 
     @Override
-    public Map<DocumentReference, List<String>> countByIdentifier(List<MacroBlock> mentions)
+    public Map<DocumentReference, List<String>> countByIdentifier(List<MacroBlock> mentions,
+        WikiReference wikiReference)
     {
         Map<DocumentReference, List<String>> ret = new HashMap<>();
         for (MacroBlock block : mentions) {
             String macroReference = block.getParameter(REFERENCE_PARAM_NAME);
-            DocumentReference reference = this.documentReferenceResolver.resolve(macroReference);
+            DocumentReference reference = this.documentReferenceResolver.resolve(macroReference, wikiReference);
             String anchor = block.getParameter(ANCHORID_PARAM_NAME);
             ret.merge(reference, new ArrayList<>(Collections.singletonList(anchor)), (l1, l2) -> {
                 l1.addAll(l2);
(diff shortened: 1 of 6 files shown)


manuelleduc:
This PR should be merged after https://github.com/xwiki/xwiki-platform/pull/1386 otherwise this fix will create a regression somewhere else

surli:
@manuelleduc have you checked it after Vincent's change on master following the close of #1386 ?

vmassol:
Normally my changes shouldn't affect it since the changes were internal ones and this PR is only supposed to use "API"s.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionXDOMService.java
```diff
@@ -48,11 +49,13 @@
     /**
      * Count the number of mentions per reference.
```
vmassol, line 50: I don't understand what is meant by "reference", could you explain?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionXDOMService.java
```diff
@@ -48,11 +49,13 @@
...
      *
-     * @param mentions the list of mentions.
+     * @param mentions the list of mentions
```
vmassol, line 52: it's actually the list of mention macros.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionXDOMService.java
```diff
@@ -48,11 +49,13 @@
...
+     * @param wikiReference the reference to the wiki in which the mentions occured
```
vmassol, line 53: occured -> occurred

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionXDOMService.java
```diff
@@ -48,11 +49,13 @@
...
      * @return the map of anchors of mentions per reference. If the anchor is not specified, then the list must contain
```
vmassol, line 54: Might be nice to explain this a bit more. I see it's returning a map with a ref as the key. What is this ref corresponding to? (basically what is an "anchor"?).

vmassol:
Note: I'm pretty sure that `traverseXObjectsOnCreate` is not going to pass checkstyle (too many nesting levels). It needs to be modified IMO.

vmassol:
> Normally my changes shouldn't affect it since the changes were internal ones and this PR is only supposed to use "API"s.

Actually I've checked the PR and I don't see the relationship at all with https://github.com/xwiki/xwiki-platform/pull/1386 :)

manuelleduc:
By moving from fully qualified user references to short one for local users, I discovered that they were resolved in the context of the main wiki, even on sub-wikis. 
Consequently, the value of the reference attribute of the mention macro was resolving to the wrong user.

manuelleduc:
- `countByIdentifier` renamed `groupAnchorsByUserReference` and has a improved javadoc
- `DefaultMentionsDataConsumer` private methods renamed and documented

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
```diff
@@ -85,12 +86,13 @@ private static boolean matchMentionMacro(Block block)
     }
 
     @Override
-    public Map<DocumentReference, List<String>> countByIdentifier(List<MacroBlock> mentions)
+    public Map<DocumentReference, List<String>> groupAnchorsByUserReference(List<MacroBlock> mentions,
+        WikiReference wikiReference)
     {
         Map<DocumentReference, List<String>> ret = new HashMap<>();
         for (MacroBlock block : mentions) {
             String macroReference = block.getParameter(REFERENCE_PARAM_NAME);
```
surli, line 94: This variable is very badly name if it's supposed to be the serializedUserReference.

PR #1389, opened 2020-09-13
XWIKI-14210: container macro doesn't work properly with dynamic layout

lucaa:
Fixed regression.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/container/columns.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/container/columns.css
@@ -21,15 +21,6 @@
   float: left;
 }
 
-/* Responsiveness, upon the same limit than Twitter Bootstrap (768px) */
-@media (max-width: 768px) {
-  .container-columns .column {
-    width: 100%;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/container/columns.css
```diff
@@ -55,3 +46,12 @@
...
   }
 #end
+
+/* Responsiveness, upon the same limit than Twitter Bootstrap (768px) */
+@media (max-width: 768px) {
+  .container-columns .column {
+    width: 100%;
+    padding-right: 0;
+    padding-left: 0;
+  }
```
vmassol, line 56: If the position is important maybe we need a comment to make sure it's not moved again in the future?

lucaa: it's actually the C part of CSS = Cascading :) the fact that the order matters is part of the CSS basics, so I trust that whoever would touch a CSS file would be aware of that. 
I don't think the regression was done on purpose, whoever did it just didn't realise which rule from under it is relevant...

PR #1391, opened 2020-09-14
XWIKI-17726: CommentAsAdminTest#testReplyToCommentAsAdmin is flickering

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17726


vmassol:
@manuelleduc this is to get more information? (I don't see how it fixes the problem)

manuelleduc:
> @manuelleduc this is to get more information? (I don't see how it fixes the problem)

IIUC, the test is flickering because it starting interacting with the UI before the comment form is fully loaded.
By adding a call to `isCommentFormShown` at the beginning of the test, we ensure that the test waits until the comment form is loaded before continuing.

vmassol:
This still wrong. A `isXXX` API (ie a getter) is not supposed to have side effects and when you read the line that you added nobody can understand that it's supposed to have side effects... In other words, your intent needs to be obvious just by reading the test code. The issue seems to be in the setup where you open the comment tab (this is in the PO that it needs to wait)...

manuelleduc:
The opening of the comments form is now explicitly done with the call to `openCommentForm`.

PR #1392, opened 2020-09-14
XWIKI-17796: Use ComponentManager to set the XMLReader used by PDF export

carlosame:
Creates a `XMLReader` and sets it to dom4j's `SAXReader`. It sets the `http://xml.org/sax/features/namespaces` feature which is generally a good idea (and compatible with what _dom4j_ did to its default reader).

The new reader uses a _Xerces-j_ fixed configuration which causes whitespace to be different to the previous test output, so the text in the test is now compared after doing a `replaceAll` that removes whitespace between tags. This also prevents a bogus test fail depending on which XML parser is used.

This pull request depends on xwiki/xwiki-commons#103.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
@@ -48,8 +48,6 @@ import org.dom4j.io.XMLWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.InputSource;
-import org.xml.sax.SAXNotRecognizedException;
-import org.xml.sax.SAXNotSupportedException;
 import org.xml.sax.XMLReader;
 import org.xwiki.bridge.DocumentAccessBridge;
 import org.xwiki.environment.Environment;
@@ -329,26 +327,10 @@ public class PdfExportImpl implements PdfExport
             XHTMLDocumentFactory docFactory = XHTMLDocumentFactory.getInstance();
             SAXReader reader = new SAXReader(docFactory);
 
-            // Dom4J 2.1.1 disables external DTDs by default, which is required here, putting it back.
+            // Dom4J 2.1.1 disables external DTDs by default, so we set our own XMLReader.
             // See https://github.com/dom4j/dom4j/issues/51
-            try {
-                reader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", true);
-            } catch (SAXNotSupportedException | SAXNotRecognizedException e) {
-                // We ingore these 2 exceptions to properly handle the case when a non-Xerces parser is provided by
-                // JAXP. Here are the possible flows:
-                //
-                // A) JAXP gives a Xerces instance:
-                // dom4j 2.1.1 sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to false,
-                // to disable DTD. No exception is thrown.
-                // XWiki sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to true, to enable DTD
-                // loading. No exception.
-                //
-                // B) JAXP gives a non-Xerces instance:
-                // dom4j 2.1.1 sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to false. The
-                // feature is not recognized: exception is thrown and ignored.
-                // XWiki sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to true. Exception is
-                // thrown (and was not being ignored). Boom.
-            }
+            XMLReader xmlReader = Utils.getComponent(XMLReaderFactory.class).createXMLReader();
+            reader.setXMLReader(xmlReader);
 
             reader.setEntityResolver(new DefaultEntityResolver());
             XHTMLDocument document = (XHTMLDocument) reader.read(source);
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
@@ -24,6 +24,7 @@ import org.xwiki.bridge.DocumentAccessBridge;
 import org.xwiki.environment.Environment;
 import org.xwiki.model.reference.DocumentReferenceResolver;
 import org.xwiki.model.reference.EntityReferenceSerializer;
+import org.xwiki.test.annotation.ComponentList;
 import org.xwiki.velocity.VelocityManager;
 
 import com.xpn.xwiki.XWikiContext;
@@ -42,6 +43,9 @@ import static org.mockito.Mockito.when;
  *
  * @version $Id$
  */
+@ComponentList({
+    org.xwiki.xml.internal.DefaultXMLReaderFactory.class,
+})
 @OldcoreTest
 public class PdfExportImplTest
 {
@@ -109,13 +113,14 @@ public class PdfExportImplTest
 
         String expected = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
             + "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" "
-                + "\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">"
-                + "<html xmlns=\"http://www.w3.org/1999/xhtml\"><head><title>\n"
+            + "\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">"
+            + "<html xmlns=\"http://www.w3.org/1999/xhtml\"><head>\n"
+            + "<title>\n"
             + "  Main.ttt - ttt\n"
-            + "</title>"
-            + "<meta content=\"text/html; charset=UTF-8\" http-equiv=\"Content-Type\"/>"
-            + "<meta content=\"en\" name=\"language\"/>"
-            + "</head><body class=\"exportbody\" id=\"body\" pdfcover=\"0\" pdftoc=\"0\">"
+            + "</title>\n"
+            + "<meta content=\"text/html; charset=UTF-8\" http-equiv=\"Content-Type\"/>\n"
+            + "<meta content=\"en\" name=\"language\"/>\n\n"
+            + "</head><body class=\"exportbody\" id=\"body\" pdfcover=\"0\" pdftoc=\"0\">\n\n"
             + "<div id=\"xwikimaincontainer\">\n"
             + "<div id=\"xwikimaincontainerinner\">\n\n"
             + "<div id=\"xwikicontent\">\n"
@@ -125,7 +130,7 @@ public class PdfExportImplTest
                 + "background-attachment: scroll; \">Hello Clément</span></p>\n"
             + "          </div>\n"
             + "</div>\n"
-            + "</div>"
+            + "</div>\n\n"
             + "</body></html>";
 
         assertEquals(expected, pdfExport.applyCSS(html, css, xcontext));


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
```diff
@@ -329,26 +329,14 @@ String applyCSS(String html, String css, XWikiContext context)
...
-                // to disable DTD. No exception is thrown.
-                // XWiki sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to true, to enable DTD
-                // loading. No exception.
-                //
-                // B) JAXP gives a non-Xerces instance:
-                // dom4j 2.1.1 sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to false. The
-                // feature is not recognized: exception is thrown and ignored.
-                // XWiki sets "http://apache.org/xml/features/nonvalidating/load-external-dtd" to true. Exception is
-                // thrown (and was not being ignored). Boom.
+                xmlReader.setFeature("http://xml.org/sax/features/namespaces", true);
```
vmassol, line 336: Could you explain why we set this here?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
```diff
@@ -329,26 +329,14 @@ String applyCSS(String html, String css, XWikiContext context)
...
+            } catch (SAXNotRecognizedException | SAXNotSupportedException e) {
```
vmassol, line 337: And also explain why it's ok to ignore these exceptions and what will be the consequences?

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
```diff
@@ -63,6 +65,7 @@ public void applyCSSWhenExistingStyleDefinedUsingShorthandNotation() throws Exce
         this.oldcore.getMocker().registerMockComponent(Environment.class);
         this.oldcore.getMocker().registerMockComponent(VelocityManager.class);
         this.oldcore.getMocker().registerMockComponent(XSLFORenderer.class, "fop");
+        this.oldcore.getMocker().registerComponent(XMLReaderFactory.class, new DefaultXMLReaderFactory());
```
vmassol, line 68: this is not great since DefaultXMLReaderFactory is a component (thus we should not use `new`).

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
```diff
@@ -128,6 +131,7 @@ public void applyCSSWhenExistingStyleDefinedUsingShorthandNotation() throws Exce
             + "</div>"
             + "</body></html>";
 
-        assertEquals(expected, pdfExport.applyCSS(html, css, xcontext));
+        assertEquals(expected.replaceAll(">\\s+<", "><"),
```
vmassol, line 134: Could you explain this?

vmassol: ok just seen the explanation in the PR. I still think the code requires a comment, it's far from obvious and we need to also understand why the behavior change is not a problem for backward compatibility.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
vmassol: > It sets the http://xml.org/sax/features/namespaces feature which is generally a good idea (and compatible with what dom4j did to its default reader).

Why is it a good idea?

carlosame: Supporting namespaces is a good idea, if you do not know in advance whether the processed XML is going to use namespaces. In any case, as mentioned in the commit message this patch is replacing something that **did** support namespaces. Removing namespace support is what should be justified, but I'm just keeping what was being done.

I did not explain it in comments because the older namespace default wasn't either (it was a dom4j default).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
carlosame: IIRC all current parsers support this (and the default value for that feature is `true` anyway), so all this is basically a no-op in practice, which is kept for backwards compatibility.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
carlosame: I'll prepare a new version of the patch with a comment (and removing the `http://xml.org/sax/features/namespaces`) if you do not deem it necessary.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
vmassol: > I did not explain it in comments because the older namespace default wasn't either (it was a dom4j default).

Sure but that's a big difference! Before, we didn't have to care about it and we were trusting the 3rd party software to do the right thing, to have its own tests and its own documentation.

With your change it now becomes XWiki's responsibility and the extra code is now on xwiki's side, so we need to do the work of explaining why we do that (and ideally even test it to verify it works, which is something we didn't need to do before).

So I suggest we add your comment to the code "Supporting namespaces is a good idea, if you do not know in advance whether the processed XML is going to use namespaces.".

Last, I think this should be done in a central place so that all code requiring an XMLReader shouldn't have to redo it. WDYT? Don't we have that in commons-xml already?

Thanks

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/pdf/impl/PdfExportImpl.java
vmassol: cool so let's add this as comment!

BTW checkstyle is supposed to fail on empty blocks (not sure when we lost that check, I'll need to do some archeology ;)).

vmassol: > IIRC all current parsers support this (and the default value for that feature is true anyway)

So why do we need it?

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
carlosame: Not sure how to register the component then.

vmassol: Since this is an integration tests (not a unit test), I think you want to use the real component implementation and thus not mock it. So the way to do this is to add a `@ComponentList()` annotation to the class:

```
@ComponentList({
  org.xwiki.xml.internal.DefaultXMLReaderFactory
})
```

See https://dev.xwiki.org/xwiki/bin/view/Community/Testing/JavaUnitTesting/#HComponentManager for details.

thanks!

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
```diff
@@ -128,6 +132,11 @@ public void applyCSSWhenExistingStyleDefinedUsingShorthandNotation() throws Exce
             + "</div>"
             + "</body></html>";
 
-        assertEquals(expected, pdfExport.applyCSS(html, css, xcontext));
+        // Depending on the configuration of the parser that is used in the test, the element content
```
vmassol, line 135: Is that not fixed (the parser used in the test)? AFAIK we control that. I'd feel better to me to have a single expectations and to have a test failure if one day the parser changes the way it handles whitespaces, so that we notice the change. 

BTW I still wonder if this parser behavior changes will lead to other problems down the road. Arguably if it does, it probably means we have a bug in the code that parses.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
```diff
@@ -109,23 +113,24 @@ public void applyCSSWhenExistingStyleDefinedUsingShorthandNotation() throws Exce
 
         String expected = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
             + "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" "
-                + "\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">"
-                + "<html xmlns=\"http://www.w3.org/1999/xhtml\"><head><title>\n"
+            + "\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">"
+            + "<html xmlns=\"http://www.w3.org/1999/xhtml\"><head>\n"
+            + "<title>\n"
```
vmassol, line 118: indentation :)

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
```diff
@@ -109,23 +113,24 @@ public void applyCSSWhenExistingStyleDefinedUsingShorthandNotation() throws Exce
...
             + "<div id=\"xwikimaincontainer\">\n"
             + "<div id=\"xwikimaincontainerinner\">\n\n"
             + "<div id=\"xwikicontent\">\n"
             + "      <p><span style=\"color: #f00; background-color: #fff; background-image: none; "
-                + "background-position: 0% 0%; background-size: auto auto; background-origin: padding-box; "
-                + "background-clip: border-box; background-repeat: repeat repeat; "
-                + "background-attachment: scroll; \">Hello Clément</span></p>\n"
+            + "background-position: 0% 0%; background-size: auto auto; background-origin: padding-box; "
+            + "background-clip: border-box; background-repeat: repeat repeat; "
+            + "background-attachment: scroll; \">Hello Clément</span></p>\n"
```
vmassol, line 130: indentation broken here too.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
carlosame: > wonder if this parser behavior changes will lead to other problems

It respects the original whitespace. Less problems, probably.

vmassol: ok, let's hope it's the case ;)

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
carlosame: The previous indentation was arbitrary. Why indent `<html` more than its descendants? Makes no sense to me, and the original document had the same indentation there (so indentation is also broken on line 81 which I did not write).

vmassol: What I did was indent when it was supposed to be text on the same line. It wasn't arbitrary ;)

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
carlosame: > indentation

Matter of taste :)

vmassol: The hardest problems ;)

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/pdf/impl/PdfExportImplTest.java
carlosame: > It wasn't arbitrary ;)

Well if the indentation criteria is based on element content whitespace (which is quite arbitrary in the original document) then it is. The DOCTYPE originally had a newline after it, which is lost in the process. Do I change that indent too?

vmassol: Don't worry I'm on it! Thx

vmassol:
Thanks for the PR @carlosame  and for bearing with me ;)

PR #1394, opened 2020-09-16
XWIKI-15833: Users shouldn't be able to pick a group as a subgroup of itself

manuelleduc:
- Fixes the use of invalid tags in the javadoc.

https://jira.xwiki.org/browse/XWIKI-15833

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
@@ -147,16 +147,18 @@ public class GroupScriptService implements ScriptService
     }
 
     /**
-     * Checks if a group (candidate) can be added to the members of another group (target).<br />
-     * A group (candidate) can be added to the members of another group (target) if the target is not, transitively,
-     * already part of the members of candidate.<br />
-     * A group cannot be added to its own members.<br />
-     * A group (candidate) cannot be added to the members of a group (target) if candidate is already one of the direct
-     * members of target.<br />
+     * Checks if a group (candidate) can be added to the members of another group (target).
+     * <ul>
+     * <li>A group (candidate) can be added to the members of another group (target) if the target is not, transitively,
+     * already part of the members of the candidate.</li>
+     * <li>A group cannot be added to its own members.</li>
+     * <li>A group (candidate) cannot be added to the members of a group (target) if the candidate is already one of the
+     * direct members of the target.</li>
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-script/src/main/java/org/xwiki/user/script/GroupScriptService.java
```diff
@@ -147,16 +147,16 @@
...
 
     /**
-     * Checks if a group (candidate) can be added to the members of another group (target).<br />
+     * Checks if a group (candidate) can be added to the members of another group (target).<br>
      * A group (candidate) can be added to the members of another group (target) if the target is not, transitively,
-     * already part of the members of candidate.<br />
-     * A group cannot be added to its own members.<br />
-     * A group (candidate) cannot be added to the members of a group (target) if candidate is already one of the direct
-     * members of target.<br />
+     * already part of the members of the candidate.<br>
```
surli, line 152: that don't fix the issue at all. Here the problem is about `<` and `>` which are not encoded.

surli, line 152: why don't you just remove those br tags? they are not really useful are they?

vmassol: if you need a list you can use `<ul>` and `<li>`

tmortagne: In general think of Javadoc as a piece of HTML with a few helpers based on the `{@something}` syntax.

vmassol: It's not HTML. You don't close elements for example and only a small subset is supported.

BR should never be used. If you need items you should use a list.

manuelleduc: Javadoc updated with ul/li tags. For now on I'll add a `mvn javadoc:javadoc` to my pre-push tests.

tmortagne: > BR should never be used. If you need items you should use a list.

My point is that to would not do a list using `*` and `<br/>` in HTML.

PR #1396, opened 2020-09-16
XWIKI-17780: Comment 'Reply' button is missing after adding a comment and refreshing the page

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17780

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
@@ -243,7 +243,7 @@ viewers.Comments = Class.create({
    * Inline reply: Move the form under the replied comment and update the hidden "replyto" field.
    */
   addReplyListener : function() {
-    if ($("commentscontent")) {
+    if (this.form) {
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -242,7 +242,7 @@ viewers.Comments = Class.create({
    * Inline reply: Move the form under the replied comment and update the hidden "replyto" field.
    */
   addReplyListener : function() {
-    if (this.form) {
+    if ($("commentscontent")) {
```
surli, line 245: What's `commentscontent` refers to?
When looking on the else part of this if, it says:
>  // If, for some reason, the form is missing, hide the reply functionality from the user

I have the feeling that the whole if is now useless no? And the "for some reason" is a bit weird.

manuelleduc: I have this feeling but until I'm 200% I'll keep the else branch. The PR now proposes an alternative fix that is IMHO less prone to create unexpected regressions

PR #1403, opened 2020-09-18
XWIKI-17726: CommentAsAdminTest#testReplyToCommentAsAdmin is flickering

surli:
* Clarify variable name in comments.js
  * Move when the success notification message is sent when a comment is
    posted, to ensure the comment is already displayed before, since
tests are waiting for it.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
@@ -325,16 +325,13 @@ viewers.Comments = Class.create({
                   "$services.localization.render('core.viewers.comments.add.inProgress')",
                   "inprogress");
           form.disable();
-          this.restartNeeded = false;
+          this.requestSucceeded = false;
           new Ajax.Request(url, {
             method : 'post',
             parameters : formData,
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/comments.js
```diff
@@ -396,6 +393,12 @@ viewers.Comments = Class.create({
 
                 // Notify any displayed CAPTCHA that it was reloaded and it might need to reinitialize its JS.
                 this.container.fire('xwiki:captcha:reloaded');
+
+                // We send success notification only when everything is done: our integration tests relies on it
+                // for waiting a comment added.
```
vmassol, line 398: Note that I don't think this is just for our tests but for anyone listening to this event.

surli: Well here it's a UI event, not even a JS event.

vmassol: Indeed. However it's probably better to display that event to the user only when the action is fully done, so I believe what you did is improving that.

PR #1408, opened 2020-09-24
XWIKI-17754: Allow to force-download attachments with a given mimetype even if they are added by a PR user

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-17754

## Description of the fix

The fix consists in adding a new property to always force-download some attachments based on their mimetype. The following changes have been made:
      * introduce a new property to list the mimetypes to force-download
      * improve a bit the documentation about whitelist/blacklist properties
        to explain the limitations for PR users
      * Refactor the check in DownloadAction to decide if the file should be
        served inline or as attachment
      * Refactor DownloadActionTest and provide some more tests to check the
        various conditions

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DownloadAction.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DownloadAction.java
@@ -47,6 +47,7 @@ import org.xwiki.model.reference.EntityReference;
 import org.xwiki.resource.ResourceReference;
 import org.xwiki.resource.ResourceReferenceManager;
 import org.xwiki.resource.entity.EntityResourceReference;
+import org.xwiki.stability.Unstable;
 
 import com.xpn.xwiki.XWiki;
 import com.xpn.xwiki.XWikiContext;
@@ -81,6 +82,13 @@ public class DownloadAction extends XWikiAction
 
     /** Key of the blacklist in xwiki.properties. */
     public static final String BLACKLIST_PROPERTY = "attachment.download.blacklist";
+
+    /**
+     * Key of the forceDownload list in xwiki.properties.
+     * @since 12.10RC1
+     */
+    @Unstable
+    public static final String FORCE_DOWNLOAD_PROPERTY = "attachment.download.forceDownload";
 
     /** The URL part separator. */
     private static final String SEPARATOR = "/";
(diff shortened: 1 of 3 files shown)


tmortagne (approved):

surli:
Converting to draft since it needs first a decision. See the following proposal: https://forum.xwiki.org/t/download-attachments-and-whitelist-blacklist-strategy/7526

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DownloadAction.java
```diff
@@ -82,6 +82,11 @@
     /** Key of the blacklist in xwiki.properties. */
     public static final String BLACKLIST_PROPERTY = "attachment.download.blacklist";
 
+    /**
+     * Key of the forceDownload list in xwiki.properties
+     */
+    public static final String FORCE_DOWNLOAD_PROPERTY = "attachment.download.forceDownload";
```
tmortagne, line 88: Missing since annotation. I guess unstable would make sense too.

surli: Indeed. I guess I didn't get complains from revapi because I didn't try to build on legacy.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DownloadAction.java
```diff
@@ -483,15 +488,43 @@ private boolean isValidRange(Long start, Long end)
...
     }
 
-    private boolean isAuthorized(String mimeType)
+    /**
+     * Check if an attachment should be downloaded or can be displayed inline.
+     * Attachments should be downloaded if:
+     * <ul>
+     *     <li>the request contains a force-download parameter with the value 1</li>
+     *     <li>or the mimetype is part of the forceDownload property list</li>
+     *     <li>or the attachments has not been added by a user with programming rights</li>
```
tmortagne, line 497: The javadoc is not fully accurate. For example this line suggest that if the author of the attachment does not have PR it's always downloaded which is not really true.

surli: Indeed it's a bit ambiguous here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DownloadAction.java
tmortagne: There is not reason for revapi to complain about a new static field.

PR #1409, opened 2020-09-24
XWIKI-17658: Unsaved content changes are not preserved anymore when using the browser's Back button to return to edit mode

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-17658 and https://jira.xwiki.org/browse/XWIKI-17871

## Description of the fix

The idea of this fix is to rely on the cached data of previousVersion field in case of using the back button to ensure
  1. that the content is sent with the right version info to avoid a conflict
  2. that the user will not send data with wrong previous information which could override existing data (situation before the check conflict mechanism)

In order to do that, I had to inject the needed fields for the check conflict mechanism in the edit template, since they need to be in the original DOM to be properly cached by the browser: fields injected with Javascript are not cached. 
I also had to change the way we reload the editor in actionButtons.js to ensure to force the reload without the cached information. Finally I also changed the format of the date we send to the server to use something more standard and easy to use.

Since I use a javascript API that is not supported by IE11 (`URLSearchParams`) I also introduced in this PR a polyfill through a Webjar dependency. 

## Tests

 - [x] `xwiki-platform-flamingo-skin-test-docker` (contains EditIT tests)
 - [x] `xwiki-platform-appwithinminutes-test` (AWM uses actionButtons.js)
 - [x] `xwiki-platform-administration-test` (Administration uses actionButtons.js)
 - [x] `xwiki-platform-distribution-flavor-test-ui` (various tests uses actionButtons.js)

Manual testing:
  - Steps  from https://test.xwiki.org/xwiki/bin/view/Nested%20Spaces/Warning%20conflict%20when%20editing%20a%20page
    - [x] With wiki editor
    - [x] With inline editing

  - [x] Steps from https://test.xwiki.org/xwiki/bin/view/CKEditor/Preview%20A%20Page

## List of changes

  * Remove in actionButtons.js all the mechanism to detect if the page
    was reached with back button
  * Remove in actionsButtons.js the injection of fields needed for
    conflict check mechanism
  * Put in edit.vm the fields needed for conflict check mechanism in
    order to be able to rely on cache for them in case of usage of back
button
  * Change the format of Date send to SaveAction to rely on milliseconds
    since Epoch instead of sending a formatted date
  * Change the reload editor function in actionButtons.js to ensure that
    the cache is discarded

### DIFF
--- a/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
+++ b/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
@@ -817,8 +817,7 @@ define('editInPlace', [
       // Check for merge conflicts only if the document is not new and we know the current version.
       if (!xwikiDocument.isNew &amp;&amp; xwikiDocument.version) {
         formData.previousVersion = xwikiDocument.version;
-        // It would have been easier to send the timestamp but that's what the Save action expects.
-        formData.editingVersionDate = new Date(xwikiDocument.modified).toISOString();
+        formData.editingVersionDate = new Date(xwikiDocument.modified).getTime();
       }
       return $.extend(formData, extra);
     }
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
@@ -100,6 +100,7 @@
     <div class="hidden">
     ## CSRF prevention
     <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />
+
     ## Pass the section number when the form is submitted in order to affect only the specified section.
     ## We need to test if the section number is present to prevent a NumberFormatException for the empty string.
     #if("$!{request.section}" != '')
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editactions.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editactions.vm
@@ -37,6 +37,12 @@
     ## This allows the preview form to redirect back to the correct editor when pressing "Back to edit":
     <span class="meta-versionSummary metamodifs">
       <input type="hidden" name="xeditaction" value="$!xcontext.action" />
+      ## Current version of the document for the conflict check
+      <input type="hidden" id="previousVersion" name="previousVersion" value="$doc.version" />
+      ## If the document is new or not when the user started the edition, for the conflict check too.
+      <input type="hidden" id="isNew" name="isNew" value="$doc.isNew()" />
+      ## We store the date of the beginning of the edition for the conflict check mechanism.
+      <input type="hidden" id="editingVersionDate" name="editingVersionDate" value="$datetool.date.getTime()" />
       #if ((!$commentFieldAdded) && $xwiki.hasEditComment())
         #set ($commentFieldAdded = 1)
         #if ($xwiki.isEditCommentFieldHidden())
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
@@ -24,6 +24,7 @@ import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -34,7 +35,6 @@ import javax.script.ScriptContext;
 
 import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.lang3.StringUtils;
-import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.suigeneris.jrcs.diff.DifferentiationFailedException;
@@ -402,11 +402,11 @@ public class SaveAction extends PreviewAction
         Version previousVersion = new Version(request.getParameter("previousVersion"));
         Version latestVersion = originalDoc.getRCSVersion();
 
-        DateTime editingVersionDate = new DateTime(request.getParameter("editingVersionDate"));
-        DateTime latestVersionDate = new DateTime(originalDoc.getDate());
+        Date editingVersionDate = new Date(Long.parseLong(request.getParameter("editingVersionDate")));
+        Date latestVersionDate = originalDoc.getDate();
 
         // we ensure that nobody edited the document between the moment the user started to edit and now
-        if (!latestVersion.equals(previousVersion) || latestVersionDate.isAfter(editingVersionDate)) {
+        if (!latestVersion.equals(previousVersion) || latestVersionDate.after(editingVersionDate)) {
             try {
                 XWikiDocument previousDoc =
                     getDocumentRevisionProvider().getRevision(originalDoc, previousVersion.toString());
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
@@ -26,25 +26,24 @@ var XWiki = (function(XWiki) {
   var actionButtons = XWiki.actionButtons = XWiki.actionButtons || {};
 
   // we need to handle the creation of document
-  var currentDocument, currentVersion, isNew, versionRefreshed;
-  var editingVersionDate = new Date();
+  var currentDocument;
+  var editingVersionDateField = $('editingVersionDate');
+  var previousVersionField = $('previousVersion');
+  var isNewField = $('isNew');
 
-  var getCurrentVersion = function (event) {
+  var refreshVersion = function (event) {
     if (currentDocument.equals(event.memo.documentReference)) {
-      // We are certainly coming from a back button navigation: the content of the editor might not reflect the real
-      // content, let's reload the editor.
-      if (!versionRefreshed && currentVersion != event.memo.version) {
-        window.location.reload();
+      if (previousVersionField) {
+        previousVersionField.setValue(event.memo.version);
       }
-      currentVersion = event.memo.version;
-      editingVersionDate = new Date();
-      isNew = "false";
-      versionRefreshed = true;
+      if (isNewField) {
+        isNewField.setValue(false);
+      }
     }
   };
 
-  // in object editor the version is updated when an object is added.
-  document.observe('xwiki:document:changeVersion', getCurrentVersion);
+  // Listen for versions change to update properly the version fields.
+  document.observe('xwiki:document:changeVersion', refreshVersion);
 
   actionButtons.EditActions = Class.create({
     initialize : function() {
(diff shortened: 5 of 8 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -425,7 +399,13 @@ var XWiki = (function(XWiki) {
     },
     // Reload the editors in case the user want to loose his change.
     reloadEditor : function () {
-      window.location.reload();
+      // We don't rely on window.location.reload() since it might keep cached data from the form
+      // we don't rely on window.location.reload(true) either since it's unclear if it's properly supported by
+      // all browsers. Instead we rely on a query parameter with the current date.
+      // URLSearchParams is not supported by IE6 but a polyfill must be available
+      var params = new URLSearchParams(window.location.search);
```
surli, line 406: @mflorea I'd need your input on this one about the polyfill.

surli: So after discussing it on #xwiki it appears that I shouldn't use this API since it's not documented on https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/js/polyfills.txt

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
```diff
@@ -402,11 +402,11 @@ private boolean isConflictingWithVersion(XWikiContext context, XWikiDocument ori
         Version previousVersion = new Version(request.getParameter("previousVersion"));
         Version latestVersion = originalDoc.getRCSVersion();
 
-        DateTime editingVersionDate = new DateTime(request.getParameter("editingVersionDate"));
-        DateTime latestVersionDate = new DateTime(originalDoc.getDate());
+        Date editingVersionDate = new Date(Long.parseLong(request.getParameter("editingVersionDate")));
```
tmortagne, line 405: Does this really makes a difference or was it just to remove a dependency on joda (which perfectly make sense too) ? It's not obvious to me when looking at the diff.

surli: It wasn't just to remove the dependency. The main idea was to be able to easily send a date from Java or from Javascript without having to mind about the exact format. I was using `editingVersionDate.toISOString()` in Javascript before, and when I wrote the velocity code I didn't find immediately the exact same API for it. So all in all I decided to rely on a solution that I found a bit more standard.

mflorea: This can be considered a breaking change and should be documented in the release notes. We need to update at least [this place](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml#L820).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
```diff
@@ -100,6 +100,13 @@
     <div class="hidden">
     ## CSRF prevention
     <input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" />
+    ## Current version of the document for the conflict check
+    <input type="hidden" id="previousVersion" name="previousVersion" value="$doc.version" />
+    ## If the document is new or not when the user started the edition, for the conflict check too.
+    <input type="hidden" id="isNew" name="isNew" value="$doc.isNew()" />
+    ## We store the date of the beginning of the edition for the conflict check mechanism.
+    <input type="hidden" id="editingVersionDate" name="editingVersionDate" value="$datetool.date.getTime()" />
```
mflorea, line 108: What about the places where ``actionButtons.js`` is used outside of edit mode:
* preview
* administration sections
* repository application
* in-place editing

If these hidden fields are used by ``actionButtons.js``, why not generate them in the same place where the action buttons are generated, i.e. ``editactions.vm``?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -425,7 +399,15 @@ var XWiki = (function(XWiki) {
     },
     // Reload the editors in case the user want to loose his change.
     reloadEditor : function () {
-      window.location.reload();
+      // We don't rely on window.location.reload() since it might keep cached data from the form
```
mflorea, line 402:
```suggestion
      // We don't rely on window.location.reload() since it might keep cached data from the form.
```

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -425,7 +399,15 @@ var XWiki = (function(XWiki) {
...
+      // we don't rely on window.location.reload(true) either since it's unclear if it's properly supported by
```
mflorea, line 403:
```suggestion
      // We don't rely on window.location.reload(true) either since it's unclear if it's properly supported by
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
surli: > why not generate them in the same place where the action buttons are generated, i.e. editactions.vm?

So I use to have a special check in actionButtons.js to not generate those fields outside an /edit/ action (see: https://github.com/xwiki/xwiki-platform/pull/1409/files#diff-c96ca895f6b4766e80a3282f4b68203aL247), that's why I put it here. 
Now for sure I should also support it on preview and in-place editing. I will check to put them directly in editaction.vm, but to be honest I'm a bit afraid about the impact on AWM and Administration. Now a solution could be to do the contrary than https://github.com/xwiki/xwiki-platform/pull/1409/files#diff-c96ca895f6b4766e80a3282f4b68203aL247 i.e. to remove the fields if I'm not in edit.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
surli: Thanks for the pointer. It's a breaking change but I'm not sure it has ever been documented, so it would need to be documented in RN to be safe but the impact should be limited. Or do you think we should keep it like that?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/edit.vm
surli: Actually from what I see for preview we don't use editactions.vm but previewactions.vm. Sounds like I will have to multiply those fields in all templates where I want to cache the info.

surli: > Actually from what I see for preview we don't use editactions.vm but previewactions.vm. Sounds like I will have to multiply those fields in all templates where I want to cache the info.

Apparently all editors (including inline) uses editactions.vm, so I'd need to put them as mentioned also in `previewactions.vm`, now IMO it makes sense to not put it for other cases you mentioned (Administration / Repository application). Do you agree?

surli: Actually I shouldn't even need to put them in `previewactions.vm` since the parameters are automatically added as hidden field for preview. So should be ok only with `editactions.vm` for that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
tmortagne: So this is indeed a breaking change of what is usually seen as an API (standard action parameters). Sounds a bit dangerous but I don't know how old and used this is.

surli:
@mflorea IMO it's ready I've been tested it as much as possible to ensure I didn't break anything so it should be safe.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
mflorea: It's relatively recent and I doubt it's used outside platform. I'm fine to change the format (I prefer the timestamp over the ISO format), as long as we update the in-place editing code.

surli: >  as long as we update the in-place editing code.

Note that I did edited it: https://github.com/xwiki/xwiki-platform/pull/1409/files#diff-ff70cc3aca6d799932e43f371b498c5eR820

xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
```diff
@@ -817,8 +817,7 @@ define('editInPlace', [
       // Check for merge conflicts only if the document is not new and we know the current version.
       if (!xwikiDocument.isNew &amp;&amp; xwikiDocument.version) {
         formData.previousVersion = xwikiDocument.version;
-        // It would have been easier to send the timestamp but that's what the Save action expects.
-        formData.editingVersionDate = new Date(xwikiDocument.modified).toISOString();
+        formData.editingVersionDate = new Date(xwikiDocument.modified).getTime();
```
mflorea, line 820: ``xwikiDocument.modified`` is already a timestamp AFAICS, but that's really minor.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editactions.vm
```diff
@@ -37,6 +37,12 @@
     ## This allows the preview form to redirect back to the correct editor when pressing "Back to edit":
     <span class="meta-versionSummary metamodifs">
       <input type="hidden" name="xeditaction" value="$!xcontext.action" />
+      ## Current version of the document for the conflict check
+      <input type="hidden" id="previousVersion" name="previousVersion" value="$doc.version" />
+        ## If the document is new or not when the user started the edition, for the conflict check too.
```
mflorea, line 42: This line is indented in order to be seen as a "child" of the previous comment? It's a bit strange that the next line with the input is not indented.

mflorea (approved):
Looks good. Just some minor comments.

xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
surli: I didn't check it back tbh, at least it's obvious by reading the code that we get a timestamp there.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editactions.vm
surli: I think it's an indent created when I copy/pasted the code there. Will fix it.

PR #1411, opened 2020-10-02
XWIKI-15815: The visible to Group option doesn't send any message to any user from the selected group

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-15815

- Enables the group message descriptor
- Defines the group messages filter
- Introduces a group message notification template, and update of the common message macros to support more than one translation parameter.
- Introduces a common AbstractMessageStreamNotificationFilter for all the messages
- Migration of the remaining JUnit 4 tests to JUnit 5

### DIFF
--- a/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java
+++ b/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java
@@ -1,0 +1,72 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.messagestream.internal;
+
+import java.util.Collection;
+import java.util.Objects;
+
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.notifications.NotificationFormat;
+import org.xwiki.notifications.filters.NotificationFilter;
+import org.xwiki.notifications.filters.NotificationFilterPreference;
+import org.xwiki.notifications.filters.NotificationFilterType;
+import org.xwiki.notifications.filters.expression.ExpressionNode;
+import org.xwiki.notifications.preferences.NotificationPreference;
+import org.xwiki.notifications.preferences.NotificationPreferenceProperty;
+
+/**
+ * Base class for the message notification filters.
+ *
+ * @version $Id$
+ * @since 12.10RC1
+ */
+public abstract class AbstractMessageStreamNotificationFilter implements NotificationFilter
+{
+    @Override
+    public boolean matchesPreference(NotificationPreference preference)
+    {
+        return Objects.equals(getEventType(), getEventType(preference));
+    }
+
+    @Override
+    public ExpressionNode filterExpression(DocumentReference user,
+        Collection<NotificationFilterPreference> filterPreferences, NotificationFilterType type,
+        NotificationFormat format)
+    {
+        return null;
+    }
+
+    @Override
+    public ExpressionNode filterExpression(DocumentReference user,
+        Collection<NotificationFilterPreference> filterPreferences, NotificationPreference preference)
+    {
+        return null;
+    }
+
+    /**
+     * @return the event type handled by the filter
+     */
+    abstract String getEventType();
+
+    private String getEventType(NotificationPreference preference)
+    {
+        return (String) preference.getProperties().get(NotificationPreferenceProperty.EVENT_TYPE);
+    }
+}
--- a/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java
+++ b/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java
@@ -1,0 +1,105 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.messagestream.internal;
+
+import java.util.Collection;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import com.xpn.xwiki.api.User;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.eventstream.Event;
+import org.xwiki.messagestream.GroupMessageDescriptor;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.notifications.NotificationFormat;
+import org.xwiki.notifications.filters.NotificationFilterPreference;
+
+import com.xpn.xwiki.XWikiContext;
+
+/**
+ * Filter that make sure a group message (to a group) from the message stream is visible only to members of this group.
+ *
+ * @version $Id$
+ * @since 12.10RC1
+ */
+@Component
+@Singleton
+@Named("GroupMessageStreamNotificationFilter")
+public class GroupMessageStreamNotificationFilter extends AbstractMessageStreamNotificationFilter
+{
+    @Inject
+    private Provider<XWikiContext> xcontextProvider;
+
+    @Override
+    public FilterPolicy filterEvent(Event event, DocumentReference userDocumentReference,
+        Collection<NotificationFilterPreference> filterPreferences, NotificationFormat format)
+    {
+        FilterPolicy ret;
+        // Don't handle events that are not group messages.
+        if (!getEventType().equals(event.getType())) {
+            ret = FilterPolicy.NO_EFFECT;
+        } else if (userDocumentReference != null) {
+            XWikiContext xWikiContext = this.xcontextProvider.get();
+            // If the user is not in the messaged group, the event is filtered.
+            // If the user is in the group, we let it pass and let downstream filter analyze the event.
+            User user = xWikiContext.getWiki().getUser(userDocumentReference, xWikiContext);
+            if (user == null) {
+                ret = FilterPolicy.NO_EFFECT;
+            } else {
+                boolean userInGroup = user.isUserInGroup(event.getStream());
+                if (!userInGroup) {
+                    ret = FilterPolicy.FILTER;
+                } else {
+                    ret = FilterPolicy.NO_EFFECT;
+                }
+            }
+        } else {
+            ret = FilterPolicy.FILTER;
+        }
+
+        return ret;
+    }
+
+    @Override
+    public String getName()
+    {
+        return "Group Message Stream Notification Filter";
+    }
+
+    @Override
+    String getEventType()
+    {
+        return GroupMessageDescriptor.EVENT_TYPE;
+    }
+
+    @Override
+    public int getPriority()
+    {
+        // Makes the priority higher than org.xwiki.notifications.filters.internal.user.EventUserFilter.getPriority, to
+        // force the group memberships to be checked before other filters.
+        // EventUserFilter in particular because it keeps events if they are send by followers of the sender, even
+        // if the follower is not part of the targeted group in our context (and consequently, skipping this filter).
+        return 2001;
+    }
+}
--- a/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-ui/src/main/resources/Main/MessageSenderMacro.xml
+++ b/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-ui/src/main/resources/Main/MessageSenderMacro.xml
@@ -943,7 +943,7 @@ Valid values are: 'everyone', 'followers', 'group' or 'user'. </description>
     <property>
       <description>Comma separated list of visibility options that the macro should allow the user to choose from.
 
-This list should be a sublist of the defualt ones: 'everyone', 'followers', 'group', 'user'.</description>
+This list should be a sublist of the default ones: 'everyone', 'followers', 'group', 'user'.</description>
     </property>
     <property>
       <mandatory/>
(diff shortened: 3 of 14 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/macros.vm
```diff
@@ -193,17 +193,29 @@ $xwiki.getPlainUserName($user)##
...
+ * @param icon the icon of the message
+ * @param translationParameters Optional array parameter.
+ *                              Used as parameters for the translation of translationKey.
+ *                              When this parameters is absent, the user sending the message is used as translation
+ *                              parameter.
  * @since 10.5RC1
  * @since 9.11.6
  *#
-#macro(displayMessageNotification $compositeEvent $translationKey $icon)
+#macro(displayMessageNotification $compositeEvent $translationKey $icon $translationParameters)
```
surli, line 208: I'm not sure how we handle this, since velocity allow to call a macro with a missing parameter, but maybe it would be better for tracability to have a new macro with the new parameter and a proper since on it?

xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java
```diff
@@ -0,0 +1,72 @@
...
+        NotificationFormat format)
+    {
+        return null;
+    }
+
+    @Override
+    public ExpressionNode filterExpression(DocumentReference user,
+        Collection<NotificationFilterPreference> filterPreferences, NotificationPreference preference)
+    {
+        return null;
```
surli, line 60: I'm not sure that's good: I can see that `DirectMessageStreamNotificationFilter` is actually overriding it, I'm wondering why it's not the case in the others. And why not on the other filterExpression method. I don't know much this API so not sure when those methods are actually used.

xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java
```diff
@@ -0,0 +1,88 @@
...
+    {
+        FilterPolicy ret;
+        // Don't handle events that are not group messages.
+        if (!getEventType().equals(event.getType())) {
+            ret = FilterPolicy.NO_EFFECT;
+        } else if (user != null) {
+            XWikiContext xWikiContext = this.xcontextProvider.get();
+            // If the user is not in the messaged group, the event is filtered.
+            // If the user is in the group, we let it pass and let downstream filter analyze the event.
+            boolean userInGroup = xWikiContext.getWiki().getUser(user, xWikiContext).isUserInGroup(event.getStream());
```
surli, line 64: Sounds potentially dangerous: getUser can return null AFAIR. And don't we have a better API from GroupManager or something like this to check if a user belongs to a group?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/groupMessage.vm
```diff
@@ -0,0 +1,26 @@
...
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+## TODO: find a way to move it to the "message stream" module
+#template('notification/macros.vm')
+#set ($translationParameters = [
+    "#displayNotificationEventUser($event.events[0].user, false)",
+    "#displayNotificationEventUser($event.events[0].stream, false)"
+])
+#displayMessageNotification($event, 'messagestream.notification.groupMessage.description', 'group', $translationParameters)
```
surli, line 26: Don't we need a notification RSS template too?

manuelleduc: Indeed I missed that. I've tested the message stream RSS and it is not well handled for all the types of messages, with displayed contents looking like this:

Title: \*\*\* No title \*\*\*
Content:
```MessageStream: [publicMessage] cd94f7c8-af2b-4c78-92e9-34ca168fb125 .
edited by Administrator.

2020/10/08 09:28 
Changes: ${newdoc.version}```

Might be better to create another issue for this one.

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java (hunk 3 above)
Comment (manuelleduc) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java:64]: I'll add a null check on the result of get user.
Regarding `isUserInGroup` I didn't found anything better.

File: xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/groupMessage.vm (hunk 4 above)
Comment (surli) [reply xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/groupMessage.vm:26]: I already created an issue for the title problem in RSS notifications templates for RecordableEvent (https://jira.xwiki.org/browse/XWIKI-17851) but I guess we need a specific handle of it for MessageStream events.

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java (hunk 3 above)
Comment (surli) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java:64]: > Regarding isUserInGroup I didn't found anything better.

interesting, IMO this is a standard API that we'd need either in User API or in Group API.

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java (hunk 2 above)
Comment (manuelleduc) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: IIUC, this contributes to the store filtering when retrieving the message send to an user activity stream.
But AFAICS, the messages are already send only to the relevant recipients, so as long a we do not provide more specific message stream preferences, retuning null (ie, no filter) is fine.
And AFAIU (but I might be missing something), the `DirectMessageStreamNotificationFilter` filter is a tautology and I could not observe different results by simply replacing the body with `return null` here too.

File: xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/macros.vm (hunk 1 above)
Comment (manuelleduc) [reply xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/macros.vm]: Replaced by `displayGroupMessageNotification` a dedicated macro for the messages sent to a group.

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java (hunk 2 above)
Comment (surli) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: > But AFAICS, the messages are already send only to the relevant recipients, so as long a we do not provide more specific message stream preferences, retuning null (ie, no filter) is fine.

Be careful, maybe it's not used anymore after the changes performed by @tmortagne to use pre-filtering, but it's certainly used when you disable the pre-filtering. You should make a manual test by disabling it in xwiki.properties and ensure it's still working properly.

Comment (ClemensRobbenhaar) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: I have been using a similar construct for a custom filter some time ago for an 11.x XWiki instance. Returning ``null`` here is fine, and handles by the calling side.

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java (hunk 3 above)
Comment (ClemensRobbenhaar) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java]: I agree about some ``isUserInGroup`` "should be API". However I am sure it is not; there are several extensions like the [Publication Workflow](https://github.com/xwiki-contrib/workflow-publication/blob/5c975c988d0ae86da2a67299810e008ca8c74baa/xwiki-workflow-publication-api/src/main/java/org/xwiki/workflowpublication/internal/DefaultPublicationRoles.java#L98) who do a manual check (recursing over possible sub groups etc.) to implement this feature. @manuelleduc 's solution is rather short in comparison :).

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java (hunk 2 above)
Comment (manuelleduc) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: I've tested without pre-filtering and couldn't find an issue. In addition, my changes are pure refactoring and does not modify the implementations of the existing filter. The most common methods are moved to the abstract method and only personal override `filterExpression`.
If something is wrong with how things are currently working, it should probably be addresses in another issue.

File: xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/groupMessage.vm (hunk 4 above)
Comment (manuelleduc) [reply xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/notification/groupMessage.vm]: PR for the improved display of the message stream rss entries https://github.com/xwiki/xwiki-platform/pull/1419

File: xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java (hunk 2 above)
Comment (surli) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: So from what I see there's 2 different cases: 
`NotificationFilter#filterExpression(DocumentReference, Collection<NotificationFilterPreference>, NotificationFilterType, NotificationFormat)` (not that one, but the one above) should be ok to return null since it's only use for inclusive/exclusive filters: basically it's used in ScopeNotificationFilter only. 

Now this method we're discussing here is used when computing the query in `QueryExpressionGenerator`: when you don't override this filter, calling `EventSearcher#searchStoreEvents` will return wrong results, it will also include some more events that shouldn't be when you don't use pre-filtering.
You see it working in XWiki because we actually do not only rely on `EventSearcher#searchStoreEvents` but we also call the `NotificationFilter#filterEvent` method in `DefaultParametrizedNotificationManager` after getting the results. 

IMO as part of your refactoring you should provide clean filters with this method properly overridden in each classes, so that we wouldn't get more results than needed from the query when not using pre-filtering.

Comment (surli) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: Note that I think we are doing this double filtering (from query first, and then from filterEvent) because some notifications couldn't be filtered only from the query (can you confirm @gdelhumeau or @tmortagne ?). Now it looks like those filters you're working on can.

Comment (manuelleduc) [reply xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/AbstractMessageStreamNotificationFilter.java:60]: The filtering is now realized fully in `filterEvent`, and nothing more is needed in the `filterExpression` method that are now all retuning `null` for the message stream filters.

File: xwiki-platform-core/pom.xml [hunk 5]
```diff
@@ -517,6 +517,11 @@
                 <new>method long org.xwiki.like.LikeManager::countUserLikes(org.xwiki.user.UserReference) throws org.xwiki.like.LikeException</new>
                 <justification>Unstable API.</justification>
               </item>
+              <item>
+                <code>java.method.removed</code>
+                <old>method boolean org.xwiki.messagestream.GroupMessageDescriptor::isEnabled(java.lang.String)</old>
+                <justification>Not a breakage. Inherits isEnable from AbstractMessageDescriptor.</justification>
```
surli, line 523: I'm surprised that it needs a justification, it doesn't break the API AFAIU since the method is exposed in the abstract. Maybe we should report it to revapi.

xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+            ret = FilterPolicy.NO_EFFECT;
+        } else if (userDocumentReference != null) {
+            XWikiContext xWikiContext = this.xcontextProvider.get();
+            // If the user is not in the messaged group, the event is filtered.
+            // If the user is in the group, we let it pass and let downstream filter analyze the event.
+            User user = xWikiContext.getWiki().getUser(userDocumentReference, xWikiContext);
+            if (user == null) {
+                ret = FilterPolicy.NO_EFFECT;
+            } else {
+                boolean userInGroup = user.isUserInGroup(event.getStream());
```
surli, line 70: Just to be sure: from what I see there's no exception triggers by this method, you know how it handles bad values such as empty string / null values / bad group name, etc?

xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-ui/src/main/resources/Main/MessageSenderMacro.xml
```diff
@@ -943,7 +943,7 @@ Valid values are: 'everyone', 'followers', 'group' or 'user'. </description>
     <property>
       <description>Comma separated list of visibility options that the macro should allow the user to choose from.
 
-This list should be a sublist of the defualt ones: 'everyone', 'followers', 'group', 'user'.</description>
+This list should be a sublist of the default ones: 'everyone', 'followers', 'group', 'user'.</description>
```
surli, line 946: No translation key for that?

manuelleduc: Just corrected a typo when I saw one :) The key `rendering.macro.messageSender.parameter.visibilityOptions.description`is defining the same sentences without the line break. I don't know if it's a good idea to use it here.

xwiki-platform-core/pom.xml
vmassol: @surli why do you say it doesn't break the API? Removing a method can always break some usage IMO (both binary and source: https://revapi.org/revapi-java/0.22.0/differences.html#_method_removed_java_method_removed).

For example imagine that there's some code using reflection on the GroupMessageDescriptor class to get the `isEnabled` method to call it. It'll result in an exception at runtime. For the source use case, maybe if some code was calling `super.isEnabled()`, idk. There are probably cases I don't see. But I'm quite confident it's not a bug and the justification is not quite correct.

Interesting BTW that java lists the binary incompatible changes that they support: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html. See also https://www.oracle.com/java/technologies/javase/8-compatibility-guide.html

xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/src/main/java/org/xwiki/messagestream/internal/GroupMessageStreamNotificationFilter.java
manuelleduc: Tried by replaced `event.getStream()`, with null, the empty string and a random string, `isUserInGroup` is returning `false` without throwing an exception.
`isUserInGroup` is catching an logging the exceptions thrown by the internal call to `com.xpn.xwiki.user.api.XWikiUser#isUserInGroup` (https://github.com/xwiki/xwiki-platform/blob/5f2c29cc3eae719f8e7634df7bb9a38914e53870/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/User.java#L127)

xwiki-platform-core/pom.xml
surli: > For example imagine that there's some code using reflection on the GroupMessageDescriptor class to get the isEnabled method to call it. It'll result in an exception at runtime.

Well it would depend if you retrieve it by using `getDeclaredMethod` or `getMethod` since the method is declared in the superclass. Still I agree this would be a possible breakage.

xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-ui/src/main/resources/Main/MessageSenderMacro.xml
surli: Ok, might be something to check but out of the scope of your PR anyway.

xwiki-platform-core/pom.xml
surli: @manuelleduc so here you could actually keep the method and just call `super.isEnabled()`, or you explain a bit more why you're breaking it. Personally I'm fine with this breakage since the chances are almost null that anyone is performing reflection on this one, and it was overridden just to disable the feature half implemented...

surli:
@manuelleduc just to be sure this one depends on #1419 right?

surli:
@manuelleduc this one needs to be rebased and its conflicts fixed, also make sure to fix the since to 12.10RC1 now.

PR #1412, opened 2020-10-05
XWIKI-17820: Make the refactoring configuration more fine grained

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17820

- Introduces the main wiki refactoring configuration source that always look at the main wiki documents
- The configuration is now resolved by looking for a value:
  - In the current wiki
  - Then in the main wiki
  - Then in xwiki.properties
If no value is found, returns the default value.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/AbstractRefactoringConfigurationSource.java
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/AbstractRefactoringConfigurationSource.java
@@ -19,35 +19,31 @@
  */
 package org.xwiki.refactoring.internal;
 
-import javax.inject.Named;
-import javax.inject.Singleton;
+import java.util.Arrays;
+import java.util.List;
 
-import org.xwiki.component.annotation.Component;
-import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.configuration.internal.AbstractDocumentConfigurationSource;
+import org.xwiki.model.reference.LocalDocumentReference;
 
 /**
- * Provides configuration from the {@code Refactoring.Code.RefactoringConfigurationClass} document in the current wiki.
- * If the {@code Refactoring.Code.RefactoringConfigurationClass} xobject exists in the
- * {@code Refactoring.Code.RefactoringConfiguration} document then always use configuration values from it and if it
- * doesn't then use the passed default value.
+ * Provides configuration from the {@code Refactoring.Code.RefactoringConfiguration} document.
  *
  * @version $Id$
- * @since 12.8RC1
+ * @since 12.9RC1
  */
-@Component
-@Singleton
-@Named("refactoring")
-public class RefactoringConfigurationSource extends AbstractRefactoringConfigurationSource
+public abstract class AbstractRefactoringConfigurationSource extends AbstractDocumentConfigurationSource
 {
-    @Override
-    protected DocumentReference getDocumentReference()
-    {
-        return new DocumentReference(DOCUMENT_REFERENCE, this.getCurrentWikiReference());
-    }
+    protected static final List<String> SPACE_NAMES = Arrays.asList("Refactoring", "Code");
+
+    protected static final LocalDocumentReference DOCUMENT_REFERENCE =
+        new LocalDocumentReference(SPACE_NAMES, "RefactoringConfiguration");
+
+    private static final LocalDocumentReference CLASS_REFERENCE =
+        new LocalDocumentReference(SPACE_NAMES, "RefactoringConfigurationClass");
 
     @Override
-    protected String getCacheId()
+    protected LocalDocumentReference getClassReference()
     {
-        return "configuration.document.refactoring";
+        return CLASS_REFERENCE;
     }
 }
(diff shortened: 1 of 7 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/AbstractRefactoringConfigurationSource.java
```diff
@@ -0,0 +1,51 @@
...
+import org.xwiki.model.reference.LocalDocumentReference;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provides configuration from the {@code Refactoring.Code.RefactoringConfiguration} document.
+ *
+ * @version $Id$
+ * @since 12.9RC1
+ */
+@Unstable
```
tmortagne, line 35: No need since this is not an API (internal package).

manuelleduc: Removed.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultRefactoringConfiguration.java
```diff
@@ -38,13 +44,65 @@
...
+     * </ul>
+     * The search stops when a value is found.
+     * If no value is found, the default value is used.
+     *
+     * @param property the property name
+     * @param defaultValue the default value, used if no property declaration is found in the hierarchy
+     * @param <T> the type of the property
+     * @return the value of the property
+     */
+    private <T> T defaultPropertyAccess(String property, T defaultValue)
```
surli, line 91: Sounds like you're reimplementing the job of CompositeConfigurationSource: @vmassol would know better that stuff, but I do think you don't need to do that.

vmassol: It's actually what we do in most places but if you wanted you could create a new ConfigurationSource component for your need that implements CompositeConfiguration as it's done at https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-configuration/xwiki-platform-configuration-default/src/main/java/org/xwiki/configuration/internal/DefaultConfigurationSource.java

That would probably be nicer I guess. I guess you could do that in the `@Named("refactoring")`  implementation.

manuelleduc: IIUC, I will add 3 configuration sources to the composite:
1. currentWikiConfigurationSource
2. mainWikiConfigurationSource
3. xwikiPropertiesSource

In the case of a call to `getProperty` from the main wiki, `currentWikiConfigurationSource` and `mainWikiConfigurationSource` are pointing to the same `xwiki:Refactoring.Code.RefactoringConfiguration` document. Consequently, if the default value is set in this document, it will be checked twice before trying `xwikiPropertiesSource`.

The impact on the performance is probably minimal and the code would be simplified. WDYT?

manuelleduc: Also, the configuration key would not be prefixed with `refactoring` when calling `xwikiPropertiesSource`.

vmassol: Indeed, interesting point ;) 

As you wish. As I said we don't introduce new composite CS in general so we don't have a best practice about that.

vmassol: > Also, the configuration key would not be prefixed with refactoring when calling xwikiPropertiesSource.

That's a good point. We're still lacking a generic mechanism for that. But this is true for any code that uses the `default` ConfigurationSource.

surli: > Also, the configuration key would not be prefixed with refactoring when calling xwikiPropertiesSource

Ok so that making the usage of CompositeConfigurationSource not possible right now. I guess we'd need an improvment issue for that in the future: would be nice to be able to use CompositeConfigurationSource for that kind of usecase.

surli (approved):

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/DefaultRefactoringConfiguration.java
surli: Anyway that's something internal so we could change its implem later if we want. I'm gonna merge it now so we can have it for 12.9.

surli:
@manuelleduc apparently you commited with both XWIKI-17820 and XWIKI-17821 and that one. Can you confirm that all changes are only related to XWIKI-17820 there?

manuelleduc:
@surli my bad, I fixed the commit logs, all the changes are related to XWIKI-17820. Thanks for spotting this.

PR #1413, opened 2020-10-05
XWIKI-17821: Hide the recycle bin skipping configuration when the recycle bin is not activated

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17821

- Hides the "skip the recycle bin" configuration if the recycle bin is not activated.
- Hides the delete section if the recycle bin is not activated as it would currently display an empty page.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
@@ -450,9 +450,10 @@ $services.localization.render('refactoring.configuration.delete.emptySection')
       <code>#if (!$xwiki.hasRecycleBin())
 ## Hides the section of the menu in which the refactoring is displayed.
 ## This should be removed if new configuration values must be displayed regardless of the recycle bin status.
-#set ($doc = $xwiki.getDocument('Refactoring.Code.RefactoringConfiguration'))
-#set ($section = $doc.getObject('XWiki.ConfigurableClass').displayInSection)
-#panel-body-content a[data-id="$section"] {
+#set ($refactoringConfigurationDoc = $xwiki.getDocument('Refactoring.Code.RefactoringConfiguration'))
+#set ($refactoringConfigurationSection =
+        $refactoringConfigurationDoc.getObject('XWiki.ConfigurableClass').displayInSection)
+#panel-body-content a[data-id="$refactoringConfigurationSection"] {
   display: none;
 }
 #end</code>
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
@@ -36,7 +36,47 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content/>
+  <content>{{velocity}}
+#if ($xwiki.hasRecycleBin())
+
+#set ($configurationObject = 'Refactoring.Code.RefactoringConfiguration')
+
+{{html clean='false'}}
+&lt;form id='delete_Refactoring.Code.RefactoringConfiguration' method='post' action="${xwiki.getURL($configurationObject, 'save')}" class='xform'&gt;
+  &lt;fieldset&gt;
+    &lt;dl&gt;
+      &lt;dt&gt;
+        &lt;label for='Refactoring.Code.RefactoringConfiguration_Refactoring.Code.RefactoringConfigurationClass_0_isRecycleBinSkippingActivated'&gt;
+          $services.localization.render('Refactoring.Code.RefactoringConfigurationClass_isRecycleBinSkippingActivated')
+        &lt;/label&gt;
+        &lt;span class='xHint'&gt;
+          $services.localization.render('Refactoring.Code.RefactoringConfigurationClass_isRecycleBinSkippingActivated.hint')
+        &lt;/span&gt;
+      &lt;/dt&gt;
+      &lt;dd&gt;
+{{/html}}
+        $doc.display('isRecycleBinSkippingActivated', 'edit')
+{{html clean='false'}}
+      &lt;/dd&gt;
+    &lt;/dl&gt;
+    &lt;input type='hidden' id='Refactoring.Code.RefactoringConfiguration_redirect' name='xredirect' value='$xwiki.getRequestURL()'&gt;
+    &lt;input type='hidden' name='form_token' value="$!{services.csrf.getToken()}"&gt;
+  &lt;/fieldset&gt;
+  &lt;div class='bottombuttons'&gt;
+    &lt;p class='admin-buttons'&gt;
+      &lt;span class='buttonwrapper'&gt;
+        &lt;input type='submit' class='button' value="$services.localization.render('admin.save')" name='formactionsac'&gt;
+      &lt;/span&gt;
+    &lt;/p&gt;
+  &lt;/div&gt;
+&lt;/form&gt;
+{{/html}}
+#else
+{{warning}}
+$services.localization.render('refactoring.configuration.delete.emptySection')
+{{/warning}}
+#end
+{{/velocity}}</content>
   <object>
     <name>Refactoring.Code.RefactoringConfiguration</name>
     <number>0</number>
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
```diff
@@ -302,4 +338,131 @@
...
+      <contentType>CSS</contentType>
+    </property>
+    <property>
+      <name/>
+    </property>
+    <property>
+      <parse>1</parse>
+    </property>
+    <property>
+      <use>always</use>
```
surli, line 465: Shouldn't you use it only on demand? This is a very specific use case only for this page IMO.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
```diff
@@ -36,7 +36,43 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content/>
+  <content>{{velocity}}
+#if ($xwiki.hasRecycleBin())
+
+#set ($configurationObject = 'Refactoring.Code.RefactoringConfiguration')
+
+{{html clean="false"}}
```
surli, line 44: why do you need a clean false here?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
```diff
@@ -36,7 +36,43 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content/>
+  <content>{{velocity}}
+#if ($xwiki.hasRecycleBin())
```
surli, line 40: No else to handle?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
vmassol: BTW always use single quotes in velocity when you don't use variables inside. It's faster.

vmassol: @surli our rule is to always use `clean='false'`  in our code (when we can) since that's faster (no need to run the cleaner) and we control the content. I don't know if this applies here (didn't check the code)

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
manuelleduc: Currently this is the only value configurable, when the recycle bin is not activated, their is nothing to display. This is why we are also hiding the entry in the menu.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
manuelleduc: I don't think that's possible unless I missed a UIX to add some content to the global administration page.
The goal here is to load a css that hide the delete section in the left menu of the global administration when the recycle bin is not activated.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
surli: > This is why we are also hiding the entry in the menu.

Ok I missed that part. I would still display a warning message in the else part, in case someone end up there.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
```diff
@@ -302,4 +338,131 @@
...
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>#if (!$xwiki.hasRecycleBin())
+## Hides the section of the menu in which the refactoring is displayed.
+## This should be removed if new configuration values must be displayed regardless of the recycle bin status.
+#set ($section = $doc.getObject('XWiki.ConfigurableClass').displayInSection)
```
surli, line 449: Have you tested it? I'm not sure to understand the logic here. You're calling this code in all pages right now, so it means if I'm in administration accessing any page with a XWiki.ConfigurableClass, it will hides the section of this page. 
Here it doesn't apply only to the recycle bin configurable class.

surli: Unless I'm wrong because `$doc` only apply to  RefactoringConfiguration document, because the CSSX is used with "Always". Still sounds a bit fragile.

manuelleduc: I did some more test with a "no cache" policy and `$doc` always refers to `Refactoring.Code.RefactoringConfiguration`.
I can hardcore I you think it makes the code safer.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
manuelleduc: Alternative message added https://github.com/xwiki/xwiki-platform/pull/1413/files#diff-3c735d937df88dc14ac06a727039ad99c06968b61044fa89899f7fb5564be338R76

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
surli: > I can hardcore it if you think it makes the code safer.

Yeah I do think it would be safer and more readable here: you only want that specific document to be checked. Note that it has nothing to do with the cache policy here, it all depends when the UIX is executed (e.g. with what doc in the context).

manuelleduc: `$xwiki.getDocument('Refactoring.Code.RefactoringConfiguration')` is now affected to `$doc`.
https://github.com/xwiki/xwiki-platform/pull/1413/files#diff-3c735d937df88dc14ac06a727039ad99c06968b61044fa89899f7fb5564be338R453

surli: Hmmmm. Not good: `$doc` is the standard Velocity variable for the current document, here you're overriding it so if you have any velocity script executed in the same context after this one and using `$doc`, it will be a big mess. Better introduce a dedicated variable name for it. In general you don't want to override variable names in velocity...

manuelleduc: `$doc` replaced by `$refactoringConfigurationDoc`
https://github.com/xwiki/xwiki-platform/pull/1413/files#diff-3c735d937df88dc14ac06a727039ad99c06968b61044fa89899f7fb5564be338R453

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/src/main/resources/Refactoring/Code/RefactoringConfiguration.xml
surli: thanks

PR #1418, opened 2020-10-08
XWIKI-17920: Notifications events marked as read are not filtered any more

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-17920

  * Introduce a new enum and method in NotificationFilter to define when
    the filter should be applied (pre/post filtering or both)
  * Override this new method in EventReadAlert/EmailFilter/ForUserEvent
  * Add a new method in NotificationFilterManager to allow retrieve the
    filters based on the NotificationFilter phase
  * Use this new method in DefaultNotificationParametersFactory and
    UserEventManager
  * Change DefaultNotificationParametersFactory to always request
    getAllFilters for post filtering phase
  * Also change it a bit to always put a set instead of a list for
    NotificationParameters#filter (interface is a collection)
  * Provide new tests covering those changes

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
@@ -21,7 +21,6 @@ package org.xwiki.notifications.filters;
 
 import java.util.Collection;
 import java.util.Map;
-import java.util.Set;
 import java.util.stream.Stream;
 
 import org.xwiki.component.annotation.Role;
@@ -30,6 +29,7 @@ import org.xwiki.notifications.NotificationException;
 import org.xwiki.notifications.filters.internal.ToggleableNotificationFilter;
 import org.xwiki.notifications.preferences.NotificationPreference;
 import org.xwiki.rendering.block.Block;
+import org.xwiki.stability.Unstable;
 
 /**
  * Provide an interface for interacting with user notification filters.
@@ -65,13 +65,15 @@ public interface NotificationFilterManager
      *
      * @param user reference to the user
      * @param onlyEnabled either or not only filters enabled for the user should be collected
-     * @param filteringMoments when the filter should be used
+     * @param filteringPhase when the filter should be used (use null for any phase)
      * @return the collection of notification filters enabled to the user.
      * @throws NotificationException if an error happens
-     * @since 10.4RC1
+     * @since 12.9RC1
+     * @since 12.6.3
      */
+    @Unstable
     default Collection<NotificationFilter> getAllFilters(DocumentReference user, boolean onlyEnabled,
-        Set<NotificationFilter.FilteringMoment> filteringMoments)
+        NotificationFilter.FilteringPhase filteringPhase)
         throws NotificationException
     {
         return getAllFilters(user, onlyEnabled);
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
@@ -65,7 +65,7 @@ public interface NotificationFilterManager
      *
      * @param user reference to the user
      * @param onlyEnabled either or not only filters enabled for the user should be collected
-     * @param filteringPhase when the filter should be used (use null for any phase)
+     * @param filteringPhase when the filter should be used ({@code null} means any phase)
      * @return the collection of notification filters enabled to the user.
      * @throws NotificationException if an error happens
      * @since 12.9RC1
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
@@ -21,6 +21,7 @@ package org.xwiki.notifications.filters;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Stream;
 
 import org.xwiki.component.annotation.Role;
@@ -58,6 +59,23 @@ public interface NotificationFilterManager
      */
     Collection<NotificationFilter> getAllFilters(DocumentReference user, boolean onlyEnabled)
         throws NotificationException;
+
+    /**
+     * Get all notifications filters that are enabled to the given user and matching the requested filtering moment.
+     *
+     * @param user reference to the user
+     * @param onlyEnabled either or not only filters enabled for the user should be collected
+     * @param filteringMoments when the filter should be used
+     * @return the collection of notification filters enabled to the user.
+     * @throws NotificationException if an error happens
+     * @since 10.4RC1
+     */
+    default Collection<NotificationFilter> getAllFilters(DocumentReference user, boolean onlyEnabled,
+        Set<NotificationFilter.FilteringMoment> filteringMoments)
+        throws NotificationException
+    {
+        return getAllFilters(user, onlyEnabled);
+    }
 
     /**
      * Get from the filters the one that match the given notification preference.
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterManager.java
@@ -165,8 +165,8 @@ public class DefaultNotificationFilterManager implements NotificationFilterManag
         // by definition if the filteringPhase provided is null we return all filters
         if (filteringPhase == null) {
             result = true;
-        // just bulletproof in case a filter didn't properly override the method
-        } else if (filterFilteringPhases == null) {
+        // we consider that filters should properly set the filtering phases to be selected.
+        } else if (filterFilteringPhases == null || filterFilteringPhases.isEmpty()) {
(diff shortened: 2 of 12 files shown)


surli:
Created as draft for now since I still need to check manually if it fixes the problem.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
```diff
@@ -59,6 +60,23 @@
...
 
+    /**
+     * Get all notifications filters that are enabled to the given user and matching the requested filtering moment.
+     *
+     * @param user reference to the user
+     * @param onlyEnabled either or not only filters enabled for the user should be collected
+     * @param filteringMoments when the filter should be used
+     * @return the collection of notification filters enabled to the user.
+     * @throws NotificationException if an error happens
+     * @since 10.4RC1
```
tmortagne, line 71: Not sure about that one :)

surli: Oups, copy/pasted too fast, I forgot unstable too.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterManager.java
```diff
@@ -59,6 +60,23 @@
...
+     */
+    default Collection<NotificationFilter> getAllFilters(DocumentReference user, boolean onlyEnabled,
+        Set<NotificationFilter.FilteringMoment> filteringMoments)
```
tmortagne, line 74: I don't understand why we need a Set here. I feels like you would use this method to get filters for:
* pre filtering only
* post filtering only
* all

And this can be expressed with a single FilteringMoment parameter already.

surli: Because you usually want the filters for pre-filtering AND both, or post-filtering AND both. You can check how I used it in UserEventManager and DefaultNotificationParametersFactory.

tmortagne: > Because you usually want the filters for pre-filtering AND both

Sure but my point is that it's what the parameter value pre-filtering should already mean. IMO it's just bad naming in the FilteringMoment enum, it should probably just be `PREFILTERING`, `POSTFILTERING` and `BOTH` (not even sure we really need BOTH actually since we could use null as "any phase" instead).

surli: > Sure but my point is that it's what the parameter value pre-filtering should already mean. 

I don't agree but in this case you would have two different meaning depending where the enum value is used: 
using `PREFILTERING` in `NotificationFilter#getPhase` would mean **only** `PREFILTERING` phase, while using `NotificationFilterManager#getAllFilters` with `PREFILTERING` parameter would return `PREFILTERING` **and** `BOTH` filters. 
I prefer if the phase names are explicit (hence the `ONLY_` prefix), as the API to use them, even if it makes it a bit more verbose to use.

tmortagne: What about having NotificationFilter#isPhaseSupported instead of NotificationFilter#getPhase ?

surli: > What about having NotificationFilter#isPhaseSupported instead of NotificationFilter#getPhase ?

could be an elegant solution indeed. Not sure if we should keep `BOTH` to be explicit if you want to use `isPhaseSupported(BOTH)` or if we should just drop it.

tmortagne: I don't think BOTH is needed in that configuration, you would just pass `null` as "I don't want to filter based on the phase, just give me everything".

surli: > I don't think BOTH is needed in that configuration

ok dropping it for now then

surli: Note that now the problem is that you cannot get the filters that are ONLY working on postfiltering, you always get all those supporting postfiltering  (meaning both + postfiltering)

surli: Which is not right I think actually... since it implies that you will call much filters than only the post-filtering ones when you don't need need to..

surli: So I just pushed the new solution, but I'm not entirely sure about it, see my comment here: https://github.com/xwiki/xwiki-platform/pull/1418/files#diff-2a2836c0e09aac5b088303fcf643a499R398

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/internal/DefaultNotificationParametersFactory.java
```diff
@@ -393,6 +395,9 @@ public void useUserPreferences(NotificationParameters parameters) throws Notific
                 enableAllEventTypes(parameters);
 
                 parameters.filters.add(new ForUserEventFilter(parameters.format, null));
+                // FIXME: This doesn't feel right: we actually also retrieve the filters for pre-filtering.
```
tmortagne, line 398: We are back to what I was talking about in the chat: the fact that IMO `getAllFilters` should check the configured mode to know if the post filterinig phase is supposed to use all filters supporting post filtering or only filters supporting post filtering. That way you can remove the `if` on the mode in `useUserPreferences`.

surli: > if the post filterinig phase is supposed to use all filters supporting post filtering or only filters supporting post filtering

The problem is that such check is not made possible with NotificationFilter#isPhaseSupported

tmortagne: Yes and no, since we have only two phases right now you would invert it (check if it's not supporting PREFILTERING when you want only POSTFILTERING) but it indeed won't work if we have more than two phase and it's a bit hackish.

surli: > I don't think moving the configuration check to the filter would be a good idea. We might need to know what a filter support whatever the current configuration.

I agree on that one.

> I guess we could have both a NotificationFilter#isPhaseSupported and NotificationFilter#isOnlyPhaseSupported and choose one of the other depending on the configuration.

Not a big fan of the second one. Sounds like I could reintroduce `NotificationFilter#getSupportedPhase` and even use it in most cases for `NotificationFilter#isPhaseSupported`. It would probably mean reintroducing BOTH phase too. I think it's better like that.

tmortagne: I think NotificationFilter#getSupportedPhases would make more sense. That way you keep the enum and getAllFilters as it is.

surli:
@tmortagne so PR ready again, I checked that notifications docker tests are ok and the bug is fixed when tested manually. The only remaining thing that I'm not entirely sure is about this case: https://github.com/xwiki/xwiki-platform/pull/1418/files#diff-0f27e8c7c4ae4f4bbd88af6d59a7e530R177

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterManager.java
```diff
@@ -111,21 +113,80 @@
...
+        // just bulletproof in case a filter didn't properly override the method
+        } else if (filterFilteringPhases == null) {
+            result = false;
+        // if pre-filtering is enabled and we're requesting for post-filtering phases, we return only
+        // the filters that only supports post-filtering: we don't need the other ones since they have been already
+        // applied.
+        } else if (prefilteringEnabled && filteringPhase == NotificationFilter.FilteringPhase.POST_FILTERING) {
+            result = Collections.singleton(NotificationFilter.FilteringPhase.POST_FILTERING)
+                .equals(filterFilteringPhases);
+        // if pre-filtering is not enabled, but we request pre-filtering filters only, it doesn't really make sense
```
tmortagne, line 177: I think I would simply apply the standard logic and return filters which only support prefiltering. Of course in practice there is none right now (but there could be soon) and more importantly no code is supposed to ask for it in postfiltering mode so it does not matter that much.

surli: > more importantly no code is supposed to ask for it in postfiltering mode

Yeah actually it sounded to me that it would have been almost a bug to perform such request hence this decision to not return anything for now.

PR #1428, opened 2020-10-14
XWIKI-17895: Modularize the mentions data consumers

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17895

- Introduces a new mention macro parameter: the user type
- The notification service is now called according to the mentioned user type
- The rendering of the user mentions is now done according to the mentioned user type
- The mention notifier now receive a collection of all the mentions and newly introduced mentions in an entity (page body, comment, large string field...)
- MentionsFormatter is moved outside of internal to allow external contributors to define their own mention rendering when providing new user types.


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
```diff
@@ -193,17 +193,15 @@ public void consume()
             MentionsData data = null;
             try {
                 data = DefaultMentionsEventExecutor.this.queue.take();
-                if (data != null) {
```
surli, line 196: Why removing this null check here?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionedUserReference.java
```diff
@@ -0,0 +1,104 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.mentions.internal;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.xwiki.text.XWikiToStringBuilder;
+
+/**
+ * A user reference.
```
surli, line 27: Would need a bit more explanation...

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionNotificationService.java
```diff
@@ -0,0 +1,122 @@
...
+ */
+@Component
+@Named(TYPE)
+@Singleton
+public class UserMentionNotificationService implements MentionNotificationService
+{
+    /**
+     * Type of the component.
+     */
+    public static final String TYPE = "user";
```
surli, line 59: Well, here it's the default implem for XWiki Users, right? "user" sounds a bit too generic to me.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
```diff
@@ -365,32 +418,66 @@ private void handleBaseObjectOnUpdate(List<BaseObject> oldEntry, BaseObject base
...
+    private void sendNotification(List<MentionNotificationParameters> notificationParametersList)
+    {
+        for (MentionNotificationParameters mentionNotificationParameters : notificationParametersList) {
+
+            // If new types are are found, that means new mentions have been found in the entity.
+            // In this case, we retrieved the mention notification services for the type of users mentioned and call
+            // them with the list of mentions analyzed.
+            for (String type : mentionNotificationParameters.getNewTypes()) {
+                this.mentionNotificationServiceProvider.get(type)
+                    .ifPresent(it -> {
```
surli, line 458: sounds like we should have a debug log if no service is found.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
```diff
@@ -365,32 +418,66 @@ private void handleBaseObjectOnUpdate(List<BaseObject> oldEntry, BaseObject base
...
+                        try {
+                            it.sendNotifications(mentionNotificationParameters);
```
surli, line 460: So you actually send back all the mentions informations, for any services. Meaning if in the same comment I have one Mention to an XWiki user, and one Mention to an AP user, both services will be called with both mentions info? Doesn't sound really right.

surli (changes requested):
I didn't went yet deep down into details, but I already have questions about the proposed architecture.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsEventExecutor.java
manuelleduc: Because AFAIK `take()` does not return null values.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
manuelleduc: The provider is already logging the lookup fail. Adding another log here would lead to two log traces for the same cause.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsFormatter.java
```diff
@@ -0,0 +1,52 @@
...
+public interface MentionsFormatter
+{
+    /**
+     * Format a user mention.
+     *
+     * @param userReference The user reference
+     * @param style The display style of the mention
+     * @return the formatted mention
+     */
+    default String formatMention(String userReference, DisplayStyle style)
```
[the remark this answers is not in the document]
manuelleduc: This is now even more generic than that, in the context of activitypub, the reference will be an URI or a webfinger identifier.
I'll update the documentation to make it clear.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
surli: > For instance, the ActivityPub implementation will want to convert local actors to their activity pub identifier to add them to the mentioned actors of the entity, event if they have been mentioned as local actors.

The usecase is interesting, I'm not yet entirely sure to agree with it though. Sounds almost like an unwanted side-effect.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameter.java
```diff
@@ -0,0 +1,171 @@
...
+
+    private final MentionLocation location;
+
+    private final String anchorId;
+
+    private final XDOM xdom;
+
+    /**
+     * @param authorReference the reference of the author of the mention
+     * @param documentReference the document in which the mention has been done
```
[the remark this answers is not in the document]
manuelleduc: `MentionNotificationParameters` is now holding an `EntitityReference` to the most specific place where the mention occurred (a `DocumentReference` if in the document body, an `ObjectPropertyReference` if in a comment for instance).

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
```diff
@@ -19,105 +19,108 @@
...
 import org.apache.commons.lang3.builder.HashCodeBuilder;
-import org.xwiki.mentions.MentionLocation;
-import org.xwiki.model.reference.DocumentReference;
-import org.xwiki.rendering.block.XDOM;
 import org.xwiki.stability.Unstable;
 import org.xwiki.text.XWikiToStringBuilder;
 
 /**
- * Object holding the parameters of the users mentions.
+ * Holds the mentions for a given entity update. Two maps of mentions, grouped by the type of user mentioned.
```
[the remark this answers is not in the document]
manuelleduc: Done

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
manuelleduc: The mentions of an entity are still sent together but we now rely on the existing event mechanism to notify listeners of the new mentions.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
```diff
@@ -19,105 +19,153 @@
...
     {
-        return this.anchorId;
+        addToMap(type, mentionedActorReference, this.newMentions);
+        return this;
     }
 
     /**
+     * @return the map of new mentions
+     */
+    public Map<String, Set<MentionNotificationParameter>> getNewMentions()
```
surli, line 133: The doc needs to specify what the keys of the map are used for. From a quicklook I think it's the type of user but I'm not sure.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
```diff
@@ -19,105 +19,153 @@
...
+    {
+        return this.newMentions;
+    }
+
+    /**
+     * @return the map of all the mentions
+     */
+    public Map<String, Set<MentionNotificationParameter>> getMentions()
```
surli, line 141: Same comment for the key of the map. I guess all mentions include new mentions too? Wouldn't that be better to have newMentions and oldMentions? I'm not sure about the usecases here.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
```diff
@@ -86,20 +83,18 @@ private static boolean matchMentionMacro(Block block)
...
             // We are currently resolving to DocumentReference to allow us to support the mention of groups.
-            DocumentReference reference =
-                this.documentReferenceResolver.resolve(serializedUserReference, wikiReference);
-            String anchor = block.getParameter(ANCHORID_PARAM_NAME);
-            ret.merge(reference, new ArrayList<>(Collections.singletonList(anchor)), (l1, l2) -> {
-                l1.addAll(l2);
-                return l1;
-            });
+
+            ret.merge(new MentionedActorReference(userReference, block.getParameter(TYPE_PARAM_NAME)),
```
surli, line 93: Note that the type parameter is new, so you need to consider that this parameter might not exist for some old mentions: maybe it's handled later in the code?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsFormatter.java
```diff
@@ -0,0 +1,57 @@
...
+ * Unless you know statically the type of user to format, {@link MentionFormatterProvider} should be used to resolve
+ * the formatter to use according to the type of the user.
+ *
+ * @since 12.10RC1
+ * @see MentionFormatterProvider
+ * @version $Id$
+ */
+@Role
+@Unstable
+public interface MentionsFormatter
```
surli, line 38: Sounds a bit overkill to have a dedicated role for it: I doubt we'll ever have another implem of it. But it's not harmful IMO.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/MentionedActorReference.java
```diff
@@ -0,0 +1,106 @@
...
+import org.xwiki.stability.Unstable;
+import org.xwiki.text.XWikiToStringBuilder;
+
+/**
+ * Represent the unique reference of a mentioned actor, composed of its serialized reference and its type.
+ *
+ * @version $Id$
+ * @since 12.10RC1
+ */
+@Unstable
```
surli, line 33: internal: no need for unstable.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionEventListener.java
```diff
@@ -0,0 +1,200 @@
...
+    public List<Event> getEvents()
+    {
+        return singletonList(new NewMentionsEvent());
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        // We send notification to the mentions explicitly typed as local user, but also untyped mentions (i.e., with a
+        // null type).
```
surli, line 117: So I guess this answers my comment before about the handle of old mentions. Can you confirm?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionEventListener.java
```diff
@@ -0,0 +1,200 @@
...
+            .setDocumentReference(
+                this.entityReferenceSerializer.serialize(entityReference.extractReference(EntityType.DOCUMENT)))
+            .setLocation(mentionNotificationParameters.getLocation())
+            .setAnchor(mentionNotificationParameter.getAnchorId())
+            .setQuote(extractedQuote);
+        String mentionedIdentity = mentionNotificationParameter.getReference();
+
+        DocumentReference documentUserReference = this.documentReferenceResolver.resolve(mentionedIdentity,
+            entityReference.extractReference(EntityType.WIKI));
+        String identity = this.entityReferenceSerializer.serialize(documentUserReference);
```
surli, line 164: Why don't use a UserReferenceResolver and a UserReferenceSerializer here? a priori you're only handling user mentions there.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionsFormatter.java
```diff
@@ -66,7 +69,7 @@ public String formatMention(String userReference, DisplayStyle style)
         if (Objects.equals(style, LOGIN) || Objects.equals(firstName, "") && Objects.equals(lastName, "")) {
             // if the login is asked explicitly we display it
             // if the user has no first name and no last name, we display the login by default
-            content = this.documentReferenceResolver.resolve(userReference).getName();
+            content = this.documentReferenceResolver.resolve(actorReference).getName();
```
surli, line 72: That would need a todo for later: I guess you're doing it like that because the User API doesn't allow to retrieve the login of a user, right? But right now it will work only because we know that any user reference we obtain is also a document reference.

surli (changes requested):

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsFormatter.java
manuelleduc: I already have one locally for ActivityPub ;)

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
manuelleduc: You are right, if the type is not defined, `MentionedActorReference` is initialized with a `null`, the `null` case is handled downstream.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionEventListener.java
manuelleduc: You are right, `MentionedActorReference` is initialized with a `null` type if the value is not defined in the macro. Mentions with a `null` type are then handled here.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
manuelleduc: Updated.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
manuelleduc: Updated as well.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionsFormatter.java
manuelleduc: TODO added

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionEventListener.java
manuelleduc: You are right, user resolver/serializer used instead.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsFormatter.java
surli: Ah sorry I didn't meant that one: I wanted to place it for MentionFormatterProvider.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/internal/MentionFormatterProvider.java
```diff
@@ -20,22 +20,23 @@
...
+ * Provides a {@link MentionsFormatter} component according to the type of actor to format.
  *
  * @version $Id$
- * @since 12.6
+ * @since 12.10RC1
+ * @see MentionsFormatter
  */
 @Role
-public interface MentionsFormatter
+public interface MentionFormatterProvider
```
surli, line 33: Sounds a bit overkill to have a dedicated role for it: I doubt we'll ever have another implem of it. But it's not harmful IMO.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
surli: Wouldn't that be safer to check here that the param is null and to use a default "user" value type to put in the map in that case? Because right now you're creating a Map that is available in different APIs and it's rarely expected to have a null value as key.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/internal/MentionFormatterProvider.java
manuelleduc: I agree, but I wouldn't bet on the future. I added it mainly for uniformity's sake, but I agree it is a matter of taste.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsFormatterProvider.java
```diff
@@ -0,0 +1,82 @@
...
+    private ComponentManager componentManager;
+
+    @Inject
+    private Logger logger;
+
+    /**
+     * Anonymous instance of the mention formater. Used as a default implementation if no specific implementation is
+     * found for a given type.
+     */
+    private final MentionsFormatter defaultFormatter = new MentionsFormatter()
```
surli, line 56: I'm not a big fan, why not having a `DefaultMentionsFormatter` component that you'd inject here? It sounds safer to future extensions of this work.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/internal/MentionFormatterProvider.java
surli: Ok so let's keep it like that.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionsFormatter.java
surli: Ok maybe create a JIRA for it too so we don't forget

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/test/java/org/xwiki/mentions/internal/DefaultMentionNotificationParameterServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test of {@link UserMentionEventListener}.
+ *
+ * @version $Id$
+ * @since 12.5RC1
```
surli, line 65: Wrong since.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/test/java/org/xwiki/mentions/internal/DefaultMentionNotificationParameterServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+import static java.util.Collections.emptyList;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test of {@link UserMentionEventListener}.
```
surli, line 62: Bad copy/paste I guess.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/test/java/org/xwiki/mentions/internal/UserMentionEventListenerTest.java
```diff
@@ -0,0 +1,172 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.xwiki.mentions.events.MentionEvent.EVENT_TYPE;
+
+/**
+ * Test of {@link UserMentionEventListener}.
+ *
+ * @version $Id$
+ * @since 12.5RC1
```
surli, line 61: Wrong since.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionsFormatter.java
manuelleduc: https://jira.xwiki.org/browse/XWIKI-17999

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionXDOMService.java
manuelleduc: That might avoid some inconsistencies later, when the type is not defined, the "user" values is now used instead.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsFormatterProvider.java
manuelleduc: `DefaultMentionsFormatter` (re-)introduced.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionsFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.xwiki.mentions.DisplayStyle.LOGIN;
+
+/**
+ * User implementation of {@link MentionsFormatter}.
+ *
+ * @version $Id$
+ * @since 12.10RC1
+ */
+@Singleton
+@Named("user")
```
surli, line 47: You're using the "user" hint as a fallback at different places (in provider and in consumer if I'm correct), so IMO it deserves a public constant.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsFormatterProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Inject
+    private MentionsFormatter defaultFormatter;
+
+    @Override
+    public MentionsFormatter get(String type)
+    {
+        String hint = type;
+        if (type == null) {
+            hint = "user";
```
surli, line 60: I'm now wondering if that's still correct, or if you shouldn't use the default formatter in that case. Normally now this should never happen with the default consumer.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/UserMentionsFormatter.java
manuelleduc: `org.xwiki.mentions.MentionsConfiguration#USER_MENTION_TYPE` introduced.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsFormatterProvider.java
manuelleduc: The default formatter shouldn't be used here, otherwise mentions defined before 12.10RC1 would be badly formatter when upgrading.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/script/MentionsScriptService.java
```diff
@@ -98,13 +98,29 @@ public boolean isQuoteActivated()
...
-     * @see MentionsFormatter#formatMention(String, DisplayStyle)
      * @param userReference the user reference
      * @param style the display style
      * @return the formatted mention
+     * @see MentionsFormatter#formatMention(String, DisplayStyle)
      */
     public String format(String userReference, DisplayStyle style)
     {
-        return this.formatter.formatMention(userReference, style);
+        return format(userReference, style, null);
```
surli, line 108: So that's actually where there's still a risk to call the FormatterProvider with a null type: shouldn't we deprecate this method, and force it to "user" instead of null? It should be enough no, in that case we shouldn't even need to handle the special case of a null type in the FormatterProvider, no?

manuelleduc: I'm ok with deprecating this format method, but I still think the special case should be handled in `FormatterProvider`, because it is not only used by the script service, but also by the renderer used to generate the mentions quotes.
We could also move the substitution of `null` to `"user"` outside of `FormatterProvider`, but that would lead to a small code duplication.

surli: > but also by the renderer used to generate the mentions quotes

IMO there should be a check here too to use "user" in case null is encountered. Basically as much as possible I'd avoid propagating a null value. I'm ok to keep checking it on the FormatterProvider for bulletproof reasons though.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsDataConsumer.java
```diff
@@ -181,42 +199,48 @@ private void traverseXObjectsOnCreate(Map<DocumentReference, List<BaseObject>> x
     /**
      * Handles an object of a created document and search for mentions to notify its content.
      * @param baseObject the object
-     * @param documentReference the reference of the document
+     * @param version the version of the document holding the base object when it was created
      * @param authorReference the reference of the author of the document
      * @param syntax the syntax of the document
+     * @param mentionNotificationParameters hold the list of identified mentions
```
surli, line 205: We agree that this map contains the mentions we already know about to compare if the ones we traverse are new?

manuelleduc: This store a list of `MentionNotificationParameters`. They are created for each traversed entity, regardless of if they have new mentions.
A the end of the analysis, we only send events for the `MentionNotificationParameters` holding new mentions.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
```diff
@@ -19,105 +19,161 @@
...
+     * Returns a map of all the mentions, including the new ones. The type of the mentioned actors are used as keys, and
+     * the values are {@link MentionNotificationParameter}, identifying a unique mention in a page by its actor
+     * reference and its anchor.
      *
-     * @return the content xdom
+     * @return the map of all the mentions
+     */
+    public Map<String, Set<MentionNotificationParameter>> getMentions()
+    {
+        return this.mentions;
```
surli, line 151: Wondering if it wouldn't be safer here to return an immutable map of the mentions: those are not supposed to be modified externally.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/analyzer/CreatedDocumentMentionsAnalyzer.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param version the version of the created document
+     * @param authorReference the reference of the author of the document
+     * @return the list of {@link MentionNotificationParameters}, holding the new mentions identified for each entity of
+     *     the document.
+     */
+    public List<MentionNotificationParameters> analyze(XWikiDocument doc,
+        DocumentReference documentReference, String version, String authorReference)
+    {
+        Syntax syntax = doc.getSyntax();
+        List<MentionNotificationParameters> ret = new ArrayList<>();
```
surli, line 77: Note that you could simplify a lot the code of those analyzers, by considering they are not singleton: if you're using a `ComponentInstantiationStrategy.PER_LOOKUP` and you explicitely call the component manager to initialize them in the consumer, then you can consider that `ret` (bad name btw ;) ) is a field on those instances. And you don't have to pass it on each method. I don't know much how it will impact the memory footprint since I'm not sure if the consumer can handle several job at once. I guess that if the pool of thread is limited it shouldn't be an issue, you should have at most the same number of analyzer instance than the size of the pool of threads for the consumer.

surli (approved):
That's a big one so I certainly missed some things but I don't see anything obvious anymore.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
manuelleduc: Interesting point. I'll add a `lock()` operation. When `lock` is called, `addMention` and `addNewMention` cannot add new `MentionNotificationParameter` objects to the maps. In addition I'll make `getMentions` and `getNewMentions` return unmodifiable maps.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/analyzer/CreatedDocumentMentionsAnalyzer.java
manuelleduc: `ret` is something we can discuss, bit I often call the variable that will be returned by the current method like this.
For the rest, I'm not convinced trading an encapsulated aggregation of values (`ret` is not passed around) for a PER_LOOKUP strategy is a good thing

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
surli: > I'll add a lock() operation. When lock is called, addMention and addNewMention cannot add new MentionNotificationParameter objects to the maps

Not sure that's needed but why not. 

> In addition I'll make getMentions and getNewMentions return unmodifiable maps.

ok, I suggested this one because it's a common mistake to call update operation on a map returned by an API.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/analyzer/CreatedDocumentMentionsAnalyzer.java
surli: > For the rest, I'm not convinced trading an encapsulated aggregation of values (ret is not passed around) for a PER_LOOKUP strategy is a good thing

why not? That's the perf impact that bothers you here or another reason? I'm not entirely against keeping it like that, it just that I think it would be easier to maintain with the `ret` considered as a field of the class.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/notifications/MentionNotificationParameters.java
manuelleduc: ok, I've limited my changes to making the returned maps unmodifiable.

PR #1429, opened 2020-10-16
XWIKI-17948: LoginIT is not executed

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17948

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
@@ -25,7 +25,6 @@ import org.junit.jupiter.api.Test;
 import org.xwiki.administration.test.po.GlobalRightsAdministrationSectionPage;
 import org.xwiki.test.docker.junit5.TestReference;
 import org.xwiki.test.docker.junit5.UITest;
-import org.xwiki.test.docker.junit5.servletengine.ServletEngine;
 import org.xwiki.test.integration.junit.LogCaptureConfiguration;
 import org.xwiki.test.ui.TestUtils;
 import org.xwiki.test.ui.XWikiWebDriver;
@@ -43,16 +42,16 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
  * @version $Id$
  * @since 2.3M1
  */
-@UITest(servletEngine = ServletEngine.EXTERNAL,
+@UITest(
     properties = {
         // The RightsManagerPlugin is needed to change rights in the UI
         "xwikiCfgPlugins=com.xpn.xwiki.plugin.rightsmanager.RightsManagerPlugin"
     }
 )
-public class LoginIT
+class LoginIT
 {
     @BeforeAll
-    static void beforeAll(TestUtils setup, XWikiWebDriver driver)
+    static void beforeAll(TestUtils setup)
     {
         // By default the minimal distribution used for the tests doesn't have any rights setup. For this test class
         // we'll need:
@@ -148,6 +147,8 @@ public class LoginIT
             setup.loginAsAdmin();
             GlobalRightsAdministrationSectionPage grasp = GlobalRightsAdministrationSectionPage.gotoPage();
             grasp.unforceAuthenticatedView();
+            // Make sure to log out to leave a deterministic state for the following tests.
+            setup.gotoPage(setup.getURLToLogout());
         }
     }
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
```diff
@@ -164,6 +162,8 @@ void correctUrlIsAccessedAfterLogin(TestUtils setup, TestReference testReference
         // Navigate to it and verify we're asked to log in and that we're not logged-in.
         setup.gotoPage(testReference);
         LoginPage loginPage = new LoginPage();
+        // Waits for the page the redirection before asserting the location.
+        setup.getDriver().waitUntilPageIsReloaded();
```
surli, line 166: It's not good you need to add a reload marker first, else this wait is not taken into account. Check the documentation of `waitUntilPageIsReloaded`.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
```diff
@@ -43,16 +42,16 @@
  * @version $Id$
  * @since 2.3M1
  */
-@UITest(servletEngine = ServletEngine.EXTERNAL,
+@UITest(
     properties = {
         // The RightsManagerPlugin is needed to change rights in the UI
         "xwikiCfgPlugins=com.xpn.xwiki.plugin.rightsmanager.RightsManagerPlugin"
```
surli, line 48: Note that this one should be added on AllIT too if I'm correct, unless we want to only rely on the merge of properties now. ping @vmassol  I don't recall the latest status on that.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
manuelleduc: `addPageNotYetReloadedMarker` added above.

vmassol: Also (I know I keep repeating this but it's important), don't explain what the code does, that's not very useful. Instead explain why you do it. In this case does it mean there's some JS executed in between? If you wan't write a proper explanation to why then it means you don't understand what you're doing! At least mention that so that others will not think you knew what you were doing ;) But best is to not put stuff you don't understand and thus always explain the why.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
surli: I will merge the PR I think it should be ok for now since we have the merging of properties, I assume you ran the whole module docker-tests and it was ok @manuelleduc right?

manuelleduc: From https://dev.xwiki.org/xwiki/bin/view/Community/Testing/DockerTesting/ 
"When using nested tests, only configure any @UITest parameter in the nested tests themselves and not in the parent (e.g. at the AllIT level the @UITest annotation should be empty). This is because @UITest annotation parameters are automatically merged."

surli: Ah thanks :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/LoginIT.java
```diff
@@ -155,10 +154,15 @@ void redirectBackAfterLogin(TestUtils setup, XWikiWebDriver driver, TestReferenc
     @Order(4)
     void correctUrlIsAccessedAfterLogin(TestUtils setup, TestReference testReference)
     {
+        // Logout, then login again to refresh the csrf token required to perform the following operations on
+        // testReference.
+        setup.gotoPage(setup.getURLToLogout());
```
surli, line 159: From what I see the previous test is actually removing cookies: that would explain the issue here. Have you checked if you reproduce the problem when the previous test is disabled? It might be something to fix in the previous test.

PR #1440, opened 2020-10-28
XWIKI-17915: Message stream messages are not displayed correctly in the rss feeds

manuelleduc:
- Defines dedicated templates to display the message in RSS feeds
- Defines dedicated titles for the event, used as the titles of the message in the RSS feeds

### DIFF
--- a/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/pom.xml
@@ -31,7 +31,7 @@
   <name>XWiki Platform - Message Stream - API</name>
   <description>API to publish and display user status messages in different streams</description>
   <properties>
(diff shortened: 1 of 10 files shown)


xwiki-platform-core/xwiki-platform-messagestream/xwiki-platform-messagestream-api/pom.xml
```diff
@@ -31,7 +31,7 @@
   <name>XWiki Platform - Message Stream - API</name>
   <description>API to publish and display user status messages in different streams</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.69</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.70</xwiki.jacoco.instructionRatio>
```
surli, line 34: did you run it on 11.10.x? We have surprises sometimes with the coverage on different branches...

manuelleduc: Good question. I've run the test locally with success.

PR #1455, opened 2020-11-09
XWIKI-18038: Livetable "html" field parameter doesn't generate HTML code anymore

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18038

Retrieve the unescaped "colname"_value field when available during the livetable client side generation.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
@@ -441,7 +441,13 @@ XWiki.widgets.LiveTable = Class.create({
         }
         // The value can be passed as a string..
         if (descriptor.html + '' === 'true') {
-          container.innerHTML = row[fieldName] || '';
+          // The value of the column must be unescaped to allow it to be used as html.
+          // For the values coming from XObjects fields, the unescaped value is found in a field suffixed with '_value'.
+          if (!table.options.columnDescriptors[column + '_value']) {
+            container.innerHTML = row[fieldName + '_value'] || row[fieldName] || '';
+          } else {
+            container.innerHTML = row[fieldName] || '';
+          }
         } else if (row[fieldName] !== undefined && row[fieldName] !== null) {
           var text = row[fieldName] + '';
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -441,7 +441,14 @@ XWiki.widgets.LiveTable = Class.create({
         }
         // The value can be passed as a string..
         if (descriptor.html + '' === 'true') {
-          container.innerHTML = row[fieldName] || '';
+          // Select the unescaped value for the current column.
+          // If a columns is already declared with the '_value' suffix, the value is directly retrieved
```
vmassol, line 445: s/columns/column

Also, as usual, you must explain why and not what! :)

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -441,7 +441,13 @@ XWiki.widgets.LiveTable = Class.create({
         }
         // The value can be passed as a string..
         if (descriptor.html + '' === 'true') {
-          container.innerHTML = row[fieldName] || '';
+          // The value of the column must be unescaped to allow it to be used as html.
```
vmassol, line 444: Could you explain why it's escaped? Is it JSON-escaped? If so that's true for any value not because it's used as HTM. What unescaping are you referring to? I don't see any unescape call below in your changes.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -441,7 +441,13 @@ XWiki.widgets.LiveTable = Class.create({
...
+          // For the values coming from XObjects fields, the unescaped value is found in a field suffixed with '_value'.
```
vmassol, line 445: Why is there a different between xobjects fields and something else? What is this something else? What are values not coming from xobject fields (I assume that you mean xproperty values?)? And where is the unescaped value found for them?

vmassol:
Still needs comment improvement :)

PR #1464, opened 2020-11-14
XWIKI-18063 : Allow wiki macros to change their macro content during their execution

aubincleme:
* Update the macro content with the content provided by the wikimacro
  binding after the execution of a macro
* Add unit test

This fixes XWIKI-18063. Even though the fix is easy in itself, could someone review it before it going into the main tree, just to make sure that we're not introducing an unexpected behavior or a security flaw ?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/DefaultWikiMacroRenderer.java
+++ b/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/DefaultWikiMacroRenderer.java
@@ -592,6 +592,10 @@ public class DefaultWikiMacroRenderer extends AbstractBlockAsyncRenderer
             restoreBindingsOrClean();
         }
 
+        // Update the macro content to take into consideration any change that could have happened to the content
+        // during the macro execution.
+        this.macroContent = macroBinding.getContent();
+
         // Replace macro placeholders (content and parameters)
         block = resolveMacroPlaceholders(block);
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/DefaultWikiMacroRenderer.java
```diff
@@ -592,6 +592,10 @@ private Block transform(XDOM macroXDOM, boolean async, boolean cached) throws Re
             restoreBindingsOrClean();
         }
 
+        // Update the macro content to take into consideration any change that could have happened to the content
+        // during the macro execution.
+        this.macroContent = macroBinding.getContent();
```
tmortagne, line 597: Might make more sense to put that in #extractResult since that's where we deal with this kind of things (the result binding for example).

tmortagne (approved):
I'm fine with the concept and the implementation but maybe this deserve a forum post to expose this new concept to more people.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/DefaultWikiMacroRenderer.java
aubincleme: Should be fine now

aubincleme:
Created forum thread to discuss about it : https://forum.xwiki.org/t/allow-wiki-macros-to-update-their-contents-and-parameters-upon-execution/7789

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-api/src/main/java/org/xwiki/rendering/macro/wikibridge/binding/WikiMacroBinding.java
```diff
@@ -90,7 +90,12 @@ public String getContent()
         return (String) this.get(CONTENT);
     }
 
-    protected void setContent(String content)
+    /**
+     * Set the content of the macro.
+     *
+     * @param content the new macro content
+     */
+    public void setContent(String content)
```
tmortagne, line 98: The use of setContent should be limited to wiki macro authors with programming right as otherwise a wiki macro could use the right of whoever uses it (since the content is executed with the right of its author).

tmortagne (changes requested):

PR #1470, opened 2020-11-26
Add mandatory features for Livedata release

ClemDee:
Add date and suggest filters for the Livedata


ClemDee:
Actually don't pay attention to the two commits named "mend". Those are just "--amend" commits I failed somehow and didn't see until now.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterSuggest.vue
```diff
@@ -0,0 +1,186 @@
...
+  methods: {
+    // Transform fetch data to xwiki.selectize parsable data
+    getSuggestion (propertyValue) {
+      const metaData = propertyValue.metaData || {};
+      return {
+        value: propertyValue.value,
+        label: metaData.label,
+        icon: metaData.icon,
+        url: metaData.url,
+        hint: metaData.hint,
```
mflorea, line 80: This should be handled by the live data source. ``logic.getPropertyValues()`` should return ``{value, label, icon, url, hint}``. See below.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterSuggest.vue
```diff
@@ -0,0 +1,186 @@
...
+          if (response && Array.isArray(response.propertyValues)) {
+            return response.propertyValues.map(this.getSuggestion);
+          } else {
+            return [];
+          }
+        }).done(callback).fail(callback);
+      };
+    },
+
+    // Taken from xwiki-platform-web/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js
```
mflorea, line 97: I don't like this. We should instead find a way to trigger this code from the ``xwiki.selectize.js``. One idea is to add an event that ``xwiki.selectize.js`` listen to and reloads the selected values as a result. ATM ``xwiki.selectize.js`` loads the selected values only when initialized. But someone could change the value afterwards. The simplest solution I see is to modify ``xwiki.selectize.js`` to listen to ``change`` event on the underlying input/select and to reload the selected values when this happens. With this we will only have to trigger the ``change`` event from the live data filter.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterSuggest.vue
```diff
@@ -0,0 +1,186 @@
...
+      if (selectize.getValue() === newValue) { return; }
+      // Clear all items and add all new items
+      selectize.clear(true);
+      valueArray.forEach((val) => {
+        if (selectize.options.hasOwnProperty(val)) {
+          selectize.addItem(val, true);
+        } else {
+          selectize.createItem(val, false);
+        }
+      });
```
mflorea, line 163: This looks like a hack. As mentioned above, I think it should be enough to trigger an event that reloads the selected values from ``xwiki.selectize.js``.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/filterMixin.js
```diff
@@ -64,6 +70,20 @@ export default {
     applyFilter (newValue) {
       this.logic.filter(this.propertyId, this.index, {value: newValue});
     },
+
+    // Call applyFilter method, but using a delay
+    // This can be used when we want to call the applyFilter method inside an input event
+    applyFilterWithDelay (newValue) {
+      // Clear existing timeout
+      if (this._applyFilterTimeoutId) {
```
mflorea, line 78: This ``if`` is not needed. ``clearTimeout`` is null safe AFAIK.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -621,6 +680,23 @@ define([
...
+     * Return whether selecting properties is enabled
+     * If entry is given, return whether this entry can be selected
+     * @param {Object} [parameters]
+     * @param {Object} [parameters.entry]
+     */
+    isSelectionEnabled ({ entry } = {}) {
+      // TODO: Ensure entry is valid (need other current PR)
+      if (entry !== undefined) {
+        return entry.age < 40;
+        // return this.isEditable({ entry });
```
mflorea, line 693: You should check if the entry has a value set for the configured id property (from the entry descriptor). We can't select entries that don't have an id. Then, if the id is specified we look at the configuration (TODO).

mflorea (changes requested):

PR #1481, opened 2020-12-16
Create FUNDING.yml

evalica:
* Displaying a sponsor button in your repository, read documentation at https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/displaying-a-sponsor-button-in-your-repository
* Using the OpenCollective service, see https://opencollective.com/xwiki

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+# You can add one username per supported platform and one custom link
+open_collective: xwiki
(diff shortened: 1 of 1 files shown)


vmassol:
Thanks @evalica. AFAIR we didn't fully finish the offer which is why we haven't advertised it yet (no post on the forum). Before we apply this PR I'd say we should do that, we've waited too long already. I remember we had said that we wanted to have the xwiki.org current funding be listed for example.

surli:
@vmassol you opened a vote for adding it on XWiki.org (see: https://forum.xwiki.org/t/adding-an-opencollective-donate-button-on-xwiki-org/9099/) but sounds like you forgot about this PR, do you think we can merge it now? IMO we probably don't need a new vote.

PR #1484, opened 2020-12-29
XWIKI-11205: If view rights are not allowed for guest users then ResetPassword doesn't work

surli:
## Jira
https://jira.xwiki.org/browse/XWIKI-11205

## Global description

This work consists in implementing Reset password and Forgot username with a dedicated ResourceHandler. Moreover, the reset action procedure has been entirely reimplemented with Java component to ease the testing, and to allow providing new implementations in the future for different authentication mechanism. 

More specifically the following work has been done:
  * Introduce an AuthenticationResourceReference supporting two actions
    (resetpassword and forgotusername) for now: this could be improved
later to also support the login.
  * Introduce the handler, serializer and resolver components related to
    the new AuthenticationResourceReference
  * Introduce some script service methods to provide URLs for the new
    resources
  * Introduce the new template pages inspired from the old wiki pages
  * Refactor the common templates htmlheader, javascript, and layoutvars
    to support a resource reference not related to a document.
  * Refactor the login template to use the right URLs
  * Fix a bug in ExtendedURL to take into account the query parameters
    in equals and hashcode
  * Edit ForgotUsername, ResetPassword and ResetPasswordComplete wiki
    pages to redirect towards the new authentication action for backward compatibility reasons
  * Introduce ResetPasswordManager and its default implementation
  * Refactor test page objects


xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/AuthenticationResourceReference.java
```diff
@@ -0,0 +1,142 @@
...
+ * @version $Id$
+ * @since 13.0RC1
+ */
+@Unstable
+public class AuthenticationResourceReference extends AbstractResourceReference
+{
+    /**
+     * Identifier of the resource type also used for associated components.
+     */
+    public static final String RESOURCE_TYPE_ID = "authenticate";
```
tmortagne, line 42: "authenticate" does not feel right, but I don't have a perfect fit to propose. I'm wondering if it should not be more about "user" instead.

surli: Note that the Maven module is called "security-authentication" and IMO it's really about handling action related to the authentication. Now I chose "authenticate" instead of "authentication" because the other actions are verbs not nouns.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/resetpasswordinline.vm
```diff
@@ -0,0 +1,293 @@
...
+            $services.localization.render('xe.admin.passwordReset.error.noEmail')</div>
+        #else
+            ## Find the object that will hold the verification string
+            #set ($verifObj = '')
+            #set ($verifObj = $userDoc.getObject($verifClass, true))
+            ## Generate a random string
+            #set ($verifStr = $util.generateRandomString(30))
+            ## If the class is correctly configured, the string should automatically be stored as a hash
+            #set ($discard = $verifObj.set('verification', $verifStr))
+            #set ($discard = $userDoc.saveAsAuthor($services.localization.render('xe.admin.passwordReset.versionComment'), true))
```
tmortagne, line 93: Using save() instead of saveAsAuthor() would also fix https://jira.xwiki.org/browse/XWIKI-16661.

Enygma2002:
@surli: What about preserving backwards compatibility for the old URLs? There may be existing logic in place that could rely on those links continuing to answer, e.g. links in emails for users to easily reset their passwords or even well known password reset keys used as first time easy initialization of passwords. Those cases will be broken if the links are no longer working, so maybe it would be good to continue providing those pages, but to change the content to delegate to the new implementation.

WDYT?

surli:
> or even well known password reset keys used as first time easy initialization of passwords.

I'm not sure to understand this example, since it almost sounds like a security flaw to me to reuse a password reset key. 

> There may be existing logic in place that could rely on those links continuing to answer, e.g. links in emails for users to easily reset their passwords

I agree with that one.

>  so maybe it would be good to continue providing those pages, but to change the content to delegate to the new implementation.

Indeed. I can see two options:
  - either I provide a message in the page informing to now redirect the user to the new URL, with a link to that URL. We could decide to remove it from XWiki in the future. 
  - or I provide a redirection to the right location: in that case, users won't even see the old page so they might not be aware that the page is actually not used. I'm a bit less fan of this option personally.

Enygma2002:
>>  or even well known password reset keys used as first time easy initialization of passwords.

> I'm not sure to understand this example, since it almost sounds like a security flaw to me to reuse a password reset key.

I know, it's a bit less intuitive, but it works nicely in practice. I actually have a use-case in production where new users (new wikis actually, but the point is the same) are notified by email that their user account was created and instead of being given a default password that they should change once they log in, they are given a direct one-time link to the password reset link already containing a pre-seeded password reset key (e.g. `https://host/xwiki/bin/view/XWiki/ResetPasswordComplete?u=xwiki%3AXWiki.<userName>&v=<preGeneratedKeyStoredInTheUserProfilePage>`). This works out much nicer, as it forces the user to initialize their password and contributes to improved security. For completeness, I also mention that a 2nd link is also mentioned to the user (in the same email) to the `XWiki/ResetPassword` page, that they can use at any time to reset their password.

This is just one example that will certainly break if we just drop the current 3 (?) password reset related pages and that the backwards compatibility approach would be more in line with what XWiki does in general (e.g. legacy module, etc.).

Generally speaking, dropping things forces applications/integrators to rely on IFs or complicated logic to cover all the possible cases (and it does not always work out well for them). Worst case scenario, integrators need a heads up of at least 6-12 months (of deprecation period) before their current apps stop working in new versions of the platform, since they can not always control the deployed platform (XWiki) version where their integrations are running, so they need time to release new versions that address the changes.

surli:
> I actually have a use-case in production where new users (new wikis actually, but the point is the same) are notified by email that their user account was created and instead of being given a default password that they should change once they log in, they are given a direct one-time link to the password reset link already containing a pre-seeded password reset key

ok I see the usecase now, for info I created https://jira.xwiki.org/browse/XWIKI-18221 for it since I haven't found a jira for that (maybe I just missed it). 

> This is just one example that will certainly break if we just drop the current 3 (?) password reset related pages and that the backwards compatibility approach would be more in line with what XWiki does in general (e.g. legacy module, etc.).

yeah I agree I need to update my PR to still support those, now you didn't provide feedback whether an automatic redirection or a manual one is better there. I still think the manual one is better (note that the manual one would be still filled with the parameters so it wouldn't involve any other actions for user than clicking on the link).

Enygma2002:
> you didn't provide feedback whether an automatic redirection or a manual one is better there. I still think the manual one is better (note that the manual one would be still filled with the parameters so it wouldn't involve any other actions for user than clicking on the link).

I honestly see no reason to complicate things for users. We don't do that in any other backwards compatibility we offer, even if most are at an API level. It would be the dev's responsibility to upgrade to the new API, while the users should ideally see no difference. Like any other deprecated API, we can throw a warning in the logs to avoid using the deprecated pages, but that should be it, IMO.

surli:
> Like any other deprecated API, we can throw a warning in the logs to avoid using the deprecated pages, but that should be it, IMO.

Ok I missed that option, going in that direction then.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/javascript.vm
```diff
@@ -218,7 +218,13 @@ XWiki.editor = "$!{escapetool.javascript($editor)}";
 XWiki.viewer = "$!{escapetool.javascript($viewer)}";
 XWiki.contextaction = "${xcontext.getAction()}";
 XWiki.skin = '$escapetool.javascript($xwiki.skin)';
+## Special case for when the request is about a resource reference not related to a document.
+## Note that we don't need to do it for the other variables since they are all of type string.
+#if ("$!doc" != "")
 XWiki.docisnew = $!doc.isNew();
+#else
+XWiki.docisnew = false;
```
tmortagne, line 226: What's the relational for choosing false (which means current document exists) instead of true ?

surli: The rationale was to say that if there's no document, then it cannot be new. Note that the fact there's no document is clarified by the fact that all the other values (such as XWiki.currentSpace and XWiki.currentPage) are just empty.

Now I do think that this should be better handled in the future by skipping entirely all those variables and not setting them in such case, if that's properly supported in our scripts which I'm really not sure of right now.

tmortagne: bq. The rationale was to say that if there's no document, then it cannot be new.

It cannot really exist either :)

surli: I'm not I agree with you on this one. It's possible that there's more `if (!XWiki.docisnew) ... display something` than the contrary but I'm pretty sure the contrary also exists. Moreover semantically speaking it would be really wrong to have a `docisnew` set to true in that case, and probably a bit WTF in case of bug related to it. So I'd be really -1 for it.

tmortagne: > semantically speaking it would be really wrong to have a docisnew set to true

I really don't see how indicating that the current document exist while there is no such thing as a current document is semantically correct. None of true or false is semantically correct.

surli: > how indicating that the current document exist while there is no such thing as a current document is semantically correct

Well you infer that the document exist because it's not new, while I'm saying that it's not new because it doesn't exist :) But yeah I agree it's a bit playing with words there. I would have preferred to just skip the else part on this if, but AFAIR it's causing some problems. Maybe I could use a null value here, wdyt @mflorea ?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
```diff
@@ -37,7 +37,8 @@
   #set($showRightPanels = $xwiki.getXWikiPreference("showRightPanels"))
   #set($leftPanelsWidth = $xwiki.getXWikiPreference("leftPanelsWidth"))
   #set($rightPanelsWidth = $xwiki.getXWikiPreference("rightPanelsWidth"))
-#else
+#elseif ("$!doc" != "") ## We only rely on the space reference if we are displaying a doc
+                        ## we want to be able to control this value in case of special resources such as authenticate.
```
vmassol, line 41: The indentation looks very weird here :)

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
```diff
@@ -0,0 +1,109 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.security.authentication.api;
```
tmortagne, line 20: Why ".api" ? Everything which is not internal is an API.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/AuthenticationResourceReference.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Dedicated resource type used for authentication.
+     */
+    public static final ResourceType TYPE = new ResourceType(RESOURCE_TYPE_ID);
+
+    /**
+     * Possible actions for authentication resource reference.
+     */
+    public enum AuthenticationAction
```
vmassol, line 52: I would put this one outside of AuthenticationResourceReference. Since it's an API it's best to do that now.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
vmassol, line 20: shouldn't all classes related to reset be in a "reset" package to separate them from other auth apis?

vmassol: Also, we don't use `api` as a package name.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
```diff
@@ -0,0 +1,109 @@
...
+public interface ResetPasswordManager
+{
+    /**
+     * Dedicated response interface for the first reset password request.
+     *
+     * @version $Id$
+     * @since 13.1RC1
+     */
+    @Unstable
+    interface ResetPasswordRequestResponse
```
vmassol, line 50: Should be outside. I'm really not fond of nested classes. Most of the time we end up wanting to extract them and this will break any API so best to do that right now IMO.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
```diff
@@ -121,4 +140,94 @@ public void resetAuthenticationFailureCounter(String username)
...
+     * This will result in computing a verification code and sending the appropriate link by email to the user.
+     * This method returns the email address used, so that we can display it to the user.
+     *
+     * @param user the user for which to perform a reset password request.
+     * @return the email address used to send the verification code.
+     * @throws ResetPasswordException if any error occurs for performing the reset password request.
+     * @since 13.1RC1
+     */
+    @Unstable
+    public InternetAddress requestResetPassword(UserReference user) throws ResetPasswordException
```
tmortagne, line 188: Shouldn't this method be limited to author with programming right ?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+        /**
+         * @return the reference of the user for whom the reset password request have been performed.
+         */
+        UserReference getUserReference();
+
+        /**
+         * @return the email address of the user for whom the reset password request have been performed.
+         */
+        InternetAddress getUserEmail();
```
vmassol, line 60: a request response is weird as name :) Why not a `ResetPasswordResponse`?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
vmassol: A request response is weird as name :) Why not a ResetPasswordResponse?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
vmassol, line 60: Do we need this, knowing that the UserReference allows getting the email of the user? I don't know enough to judge, just asking in case i's useful.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/checkstyle/checkstyle-suppressions.xml
```diff
@@ -26,4 +26,5 @@
 
 <suppressions>
   <suppress checks="FanOutComplexity" files="DefaultAuthenticationFailureManager.java"/>
+  <suppress checks="FanOutComplexity" files="DefaultResetPasswordManager.java"/>
```
vmassol, line 29: not great :)

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
```diff
@@ -0,0 +1,151 @@
...
+
+import com.xpn.xwiki.XWikiContext;
+
+/**
+ * Component dedicated to send reset password information by email.
+ *
+ * @version $Id$
+ * @since 13.1RC1
+ */
+@Component(roles = ResetPasswordMailSender.class)
```
vmassol, line 61: I'm personally not fond of this. I much prefer a clean interface. Otherwise if you're not careful you end up adding new method that shouldn't be APIs. I know this is internal but still.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
```diff
@@ -0,0 +1,151 @@
...
+        MimeMessage message;
+        try {
+            message =
+                this.mimeMessageFactory.createMessage(
+                    this.documentReferenceResolver.resolve(RESET_PASSWORD_MAIL_TEMPLATE_REFERENCE), parameters);
+        } catch (MessagingException e) {
+            throw new ResetPasswordException(localizedError, e);
+        }
+        try {
+            MailListener mailListener = this.componentManager.getInstance(MailListener.class, "database");
```
vmassol, line 133: why not inject it and avoid the try/catch?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    {
+        ExecutionContext context = this.execution.getContext();
+        if (context == null) {
+            context = new ExecutionContext();
+        }
+        try {
+            XWikiContext xWikiContext = this.xWikiContextInitializer.initialize(context);
+            // We are directly relying on Utils#parseTemplate because we want the plugin manager to properly
+            // handle the javascript placeholders and it avoids duplicating code.
+            Utils.parseTemplate(templateName, true, xWikiContext);
```
vmassol, line 101: old api usage. Should probably use the TemplateManager.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceResolver.java
```diff
@@ -0,0 +1,72 @@
...
+                result = new AuthenticationResourceReference(authenticationAction);
+                copyParameters(representation, result);
+                return result;
+            } catch (IllegalArgumentException e) {
+                throw new CreateResourceReferenceException(
+                    String.format("Cannot find an authentication action for name [%s]", actionName));
+            }
+        } else {
+            throw new CreateResourceReferenceException(
+                String.format("Invalid Authentication URL format: [%s]", resourceType.toString()));
```
vmassol, line 69: I think it's the representation that needs to be passed, no?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceSerializer.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public ExtendedURL serialize(AuthenticationResourceReference resource)
+        throws SerializeResourceReferenceException, UnsupportedResourceReferenceException
+    {
+        return new ExtendedURL(Arrays.asList(
+            contextProvider.get().getWikiId(),
+            AuthenticationResourceReference.TYPE.getId(),
+            resource.getAction().getRequestParameter()),
+            resource.getParameters());
```
vmassol, line 59: isn't this missing the the servlet context?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/test/java/org/xwiki/security/authentication/internal/DefaultResetPasswordManagerTest.java
```diff
@@ -0,0 +1,364 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link DefaultResetPasswordManager}.
+ *
+ * @version $Id$
+ */
+@ComponentTest
+public class DefaultResetPasswordManagerTest
```
vmassol, line 68: public is not needed

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/test/java/org/xwiki/security/authentication/internal/ResetPasswordMailSenderTest.java
```diff
@@ -0,0 +1,240 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link ResetPasswordMailSender}.
+ *
+ * @version $Id$
+ */
+@ComponentTest
+public class ResetPasswordMailSenderTest
```
vmassol, line 76: same

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/test/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceHandlerTest.java
```diff
@@ -0,0 +1,145 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link AuthenticationResourceReferenceHandler}.
+ *
+ * @version $Id$
+ * @since 13.0RC1
+ */
+@ComponentTest
+public class AuthenticationResourceReferenceHandlerTest
```
vmassol, line 66: same

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/test/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceResolverTest.java
```diff
@@ -0,0 +1,93 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests for {@link AuthenticationResourceReferenceResolver}.
+ *
+ * @version $Id$
+ */
+@ComponentTest
+public class AuthenticationResourceReferenceResolverTest
```
vmassol, line 46: same (stopping here and not repeating for the other ones ;))

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
```diff
@@ -121,4 +140,94 @@ public void resetAuthenticationFailureCounter(String username)
     {
         this.authenticationFailureManager.resetAuthenticationFailureCounter(username);
     }
+
+    /**
+     * Compute a relative URL for an {@link AuthenticationResourceReference} based on the given action string.
+     * See {@link AuthenticationResourceReference.AuthenticationAction} for more information.
+     *
+     * @param action the authentication action from which to build the right URL.
```
vmassol, line 148: Could you give an example in the javadoc? What is the "action"? Usually the action has a meaning for "bin" URL types (e.g. "view" or "edit" actions).

vmassol: There's the AuthenticationResourceReference.AuthenticationAction link though which probably is enough.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
```diff
@@ -121,4 +140,94 @@ public void resetAuthenticationFailureCounter(String username)
...
+                new AuthenticationResourceReference(authenticationAction);
+            if (params != null) {
+                for (Map.Entry<String, Object> entry : params.entrySet()) {
+                    resourceReference.addParameter(entry.getKey(), entry.getValue());
+                }
+            }
+            ExtendedURL extendedURL = this.defaultResourceReferenceSerializer.serialize(resourceReference);
+            return extendedURL.serialize();
+        } catch (IllegalArgumentException | SerializeResourceReferenceException
+            | UnsupportedResourceReferenceException e) {
```
vmassol, line 170: Normally our best practices is to go to the next line for `{`  when it spans.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
```diff
@@ -121,4 +140,94 @@ public void resetAuthenticationFailureCounter(String username)
...
         this.authenticationFailureManager.resetAuthenticationFailureCounter(username);
     }
+
+    /**
+     * Compute a relative URL for an {@link AuthenticationResourceReference} based on the given action string.
+     * See {@link AuthenticationResourceReference.AuthenticationAction} for more information.
+     *
+     * @param action the authentication action from which to build the right URL.
+     * @param params the query string parameters of the URL.
+     * @return a relative URL for the current wiki.
```
vmassol, line 150: apparently it can return null, this needs to be explained.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
```diff
@@ -121,4 +140,94 @@ public void resetAuthenticationFailureCounter(String username)
...
+                for (Map.Entry<String, Object> entry : params.entrySet()) {
+                    resourceReference.addParameter(entry.getKey(), entry.getValue());
+                }
+            }
+            ExtendedURL extendedURL = this.defaultResourceReferenceSerializer.serialize(resourceReference);
+            return extendedURL.serialize();
+        } catch (IllegalArgumentException | SerializeResourceReferenceException
+            | UnsupportedResourceReferenceException e) {
+            logger.error("Error while getting authentication URL for action [{}]: [{}]", action,
+                ExceptionUtils.getRootCauseMessage(e));
```
vmassol, line 172: This is not correct. Either it returns a warning and we display only the root cause or it's a non recoverable error and we log an error and display the full stack trace

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/ExtendedURL.java
```diff
@@ -286,6 +287,7 @@ public boolean equals(Object object)
         return new EqualsBuilder()
             .append(getURI(), rhs.getURI())
             .append(getSegments(), rhs.getSegments())
+            .append(getParameters(), rhs.getParameters())
```
vmassol, line 290: This should probably have been done in a separate commit but ok :)

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
surli: So sadly that's not something I introduced here, but I introduced this package in 11.6 when working on the authentication failure manager, see: https://github.com/xwiki/xwiki-platform/tree/master/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api. 
We can change it now but it would be breaking... I felt that it would be cleaner to have all our APIs in the same package, even if the name is wrong.

surli: I answered to Thomas about the exact same comment here: https://github.com/xwiki/xwiki-platform/pull/1484#discussion_r562646934

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/AuthenticationResourceReference.java
surli: OK I'll extract it.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
surli: > Should be outside for me.

Yeah I hesitated about that one, but indeed I'll extract it too. 

> Why not a ResetPasswordResponse?

Because that's not a response when you reset the password, but when you perform a request to reset it. So before you're able to reset it.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
vmassol: I think we should fix it now. Cannot we keep the .api one in legacy for example?

In any case I think it's better for the future to have a clean api , especially as it's growing. 

@tmortagne  WDYT?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceHandler.java
surli: So I explained on the comment why I did not use the TemplateManager: as a matter of fact I tried first to use the template manager, but I needed more stuff to handle the JSX placeholders and I discovered that this methoud was already doing all the plumbing I needed.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
surli: Because my MailSender component is a singleton, and this component should be instantiated each time I call the method.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceHandler.java
vmassol: ok, I reviewed quickly and missed the comment. We'll need to do something at some point since we need to deprecate and remove `Utils.parseTemplate`. Either put that it does in TemplateManager somewhere, or introduce some new component.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
surli: Hmmmm good catch, I've been careful to check possible attacks on the other ones but looks like I missed the obvious one...

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
vmassol: For that you can use:

```
@Inject
@Named("database")
private Provider<MailListener> databaseMailListenerProvider;
```

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
surli: That mainly avoids having to compute back the email address from the user reference in the script service when it has always been computed in the main component.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
surli: Well for this one I really do prefer keeping it like that and not exposing it as an interface since there's no real need of using it externally. At first it was even implemented as part of the DefaultResetPasswordManager, I put it outside because it was becoming really too big.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/test/java/org/xwiki/security/authentication/internal/DefaultResetPasswordManagerTest.java
surli: Right, I keep forgetting to remove it on the test class...

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-script/src/main/java/org/xwiki/security/authentication/script/AuthenticationScriptService.java
surli: Right

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
tmortagne: New API like ResetPasswordManager should definitely not go in this bad package IMO. The existing ones also don't really have much to do with ResetPasswordManager except for the fact that they are generally about authentication. There is no rule that say all APIs of a modules should go in the same package and I would really not see the point.

Now for what to do about the existing ones it does not have anything to do with this PR, but I suggest deprecating them and copy them in a more accurate package.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
tmortagne: > Otherwise if you're not careful you end up adding new method that shouldn't be APIs

As you said since it's internal, so it's not an API. I really don't see the point of having an interface/implementation for something like this, it's painful and does not bring any kind of value.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/ResetPasswordMailSender.java
surli: Ah indeed I missed that option.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/api/ResetPasswordManager.java
surli: > New API like ResetPasswordManager should definitely not go in this bad package IMO.

Ok I'll fix this one here. 

> Now for what to do about the existing ones it does not have anything to do with this PR, but I suggest deprecating them and copy them in a more accurate package.

Agreed that it shouldn't be part of this PR, I'll open a ticket for it to not forget.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/resource/AuthenticationResourceReferenceSerializer.java
surli: Not sure what you mean, it's just an URL there.

PR #1488, opened 2021-01-11
XWIKI-17037: Vote isn't saved when the Page or it parent Space has a double quote in its name

manuelleduc:
Escapes the page name when displaying the displayFullRating macro.

https://jira.xwiki.org/browse/XWIKI-17037

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
+++ b/xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
@@ -21,34 +21,30 @@ package org.xwiki.ratings;
 
 import java.util.Arrays;
 import java.util.Optional;
-import java.util.Properties;
 
 import javax.inject.Named;
 
-import org.apache.velocity.runtime.RuntimeConstants;
-import org.apache.velocity.runtime.resource.loader.ClasspathResourceLoader;
-import org.junit.jupiter.api.BeforeEach;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.select.Elements;
 import org.junit.jupiter.api.Test;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.script.ModelScriptService;
 import org.xwiki.ratings.script.RatingsScriptService;
-import org.xwiki.rendering.block.XDOM;
 import org.xwiki.rendering.macro.Macro;
 import org.xwiki.rendering.macro.descriptor.MacroDescriptor;
 import org.xwiki.rendering.syntax.Syntax;
 import org.xwiki.script.service.ScriptService;
-import org.xwiki.template.TemplateManager;
 import org.xwiki.test.junit5.mockito.MockComponent;
 import org.xwiki.test.page.PageTest;
 import org.xwiki.test.page.XHTML10ComponentList;
 import org.xwiki.test.page.XWikiSyntax21ComponentList;
-import org.xwiki.velocity.VelocityConfiguration;
 import org.xwiki.velocity.tools.EscapeTool;
 import org.xwiki.xml.html.filter.HTMLFilter;
 
 import com.xpn.xwiki.doc.XWikiDocument;
 
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        MacroDescriptor templateMacroDescriptor = mock(MacroDescriptor.class);
+        when(this.templateMacro.getDescriptor()).thenReturn(templateMacroDescriptor);
+        when(templateMacroDescriptor.getParametersBeanClass()).thenReturn((Class) "".getClass());
+
+        registerVelocityTool("escapetool", new EscapeTool());
+
+        // Initialize a document reference to be used as parameter when calling displayFullRating.
+        ModelScriptService modelScriptService = mock(ModelScriptService.class);
+        when(modelScriptService.createDocumentReference("a", "b", "c\"d"))
```
vmassol, line 101: I'm curious: Why do you need a setup() since there's only a single test?

Also the `"c\"d"` seems very related to an escape test and should be in that test IMO.

WDYT?

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
```diff
@@ -0,0 +1,133 @@
...
+            + "#displayFullRating($tdoc){{/velocity}}";
+
+        XWikiDocument document = new XWikiDocument(new DocumentReference("xwiki", "XWiki", "Test"));
+        document.setSyntax(Syntax.XWIKI_2_0);
+        document.setContent(script);
+        this.xwiki.saveDocument(document, "registering document", true, this.context);
+        String content = document.getRenderedContent(this.context);
+
+        // Verify that the displayFullRating parameter is actually escaped when used.
+        assertTrue(content.contains("data-reference=\"a:b.c&#34;d\""));
```
vmassol, line 131: Maybe something more precise in term of HTML asserting? Some code could use a `data-reference` attribute not related to ratings for ex, so doing a check more related to ratings could make the test more stable. WDYT?

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
manuelleduc: I've set on the setup what's generic and could be useful to future tests, but that might be (too) early optimization.
And indeed, the call to createDocumentReference is specific to my test, I'll remove the setup for now.

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
manuelleduc: I've replaced this test with an inspection of the DOM structure with Jsoup.

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
manuelleduc: And actually, having a second look help I found out I could drastically simplify the test setup.

xwiki-platform-core/xwiki-platform-ratings/xwiki-platform-ratings-ui/src/test/java/org/xwiki/ratings/RatingsTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Page test of the ratings.
+ *
+ * @version $Id$
+ * @since 13.0RC1
```
surli, line 56: Since should be updated.

surli (approved):

PR #1497, opened 2021-01-27
XWIKI-18273 The LD date filter doesn't behave as expected

ClemDee:


ClemDee:
There is still a point about convert filter value when changing between "contains" and "isBetween" operators that can be discussed.

For now, we either get `Invalid date` or something like `1611615600000-1611701999999` depending on the operator we're switching to (respectively "contains" to "isBetween", and "isBetween" to "contains" in these examples).

I think there are 2 possibilities:

- keep it simple, stupid: keep the raw value in the input, the user will modify it so that it meet his needs
- smart filters: transform the raw value so that it correspond to the operator
  - from "isBetween" to "contains": keep the formatted value of the "from" date
  - from "contains" to "isBetween": if the input can be parsed as a date, use it, else set range to "today"

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -1114,6 +1114,10 @@ define([
       if (!newEntryValidOperator) {
         newEntry.operator = self.getFilterDefaultOperator(newEntry.property);
       }
+      // If operator has changed, reset value
```
mflorea, line 1117: This is going to be very annoying when switching the operator doesn't have an impact on the value. For instance the value shouldn't change when switching between ``lowerThan`` and ``greaterThan`` operators of the number filter. Same for the "before" and "after" date operators. So would be better to clear the value only for some operators. E.g. clear the value if "contains" is ether the old or the new operator. Of course, this information is known by the filter not by the logic. So the filter needs to detect that the old/new operator is "contains" and tell logic to clear the value, somehow. WDYT?

PR #1502, opened 2021-01-27
XWIKI-18266 Improve the L&F of the live data macro

ClemDee:


ClemDee:
About this one statement in this issue
> The filter inputs are a bit too close to each other

Actually, instead of separating them even more, I decided to bring them closer together and merge their borders, so that the result looks more cleaner (less lines visually), and we even gain space inside the filters inputs.

Before:
image

After:
image

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderFilters.vue
```diff
@@ -80,7 +80,8 @@ export default {
 <style>
 
 .layout-table .column-filters th {
-  padding: 2px;
+  padding: 0px;
+  border: none;
```
mflorea, line 84: We should keep the border in order to preserve the standard XWiki table styles. The table layout should look as much as possible like any other table inside XWiki. Moreover, I think it looks better if the header and the filter row are visually separated from the table body with this thin border.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableRow.vue
```diff
@@ -90,6 +90,10 @@ export default {
   height: 100%;
 }
 
+.layout-table tbody tr:first-child .cell {
+  border: none;
```
mflorea, line 94: Same as above: we shouldn't overwrite the standard XWiki table styles.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderFilters.vue
```diff
@@ -91,6 +92,14 @@ export default {
   */
   width: 100vw;
 }
+.layout-table .column-filters th:not(:empty) {
+  border: 1px solid #ddd;
```
mflorea, line 96: Square corners on filter inputs is not consistent with XWiki style: we use rounded corners on form fields everywhere else.

mflorea (changes requested):
I would keep the current styles for the filter row (rounded corners, small space between the filter inputs, standard table border). For the rest of the changes we should ask for an agreement on forum.xwiki.org . I'll start a discussion.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/LayoutLoader.vue
```diff
@@ -0,0 +1,129 @@
...
+  methods: {
+    watchLoading (propertyName, eventBefore, eventAfter) {
+      let timeoutId;
+
+      this.logic.onEvent(eventBefore, () => {
+        console.log("BEFORE", propertyName, eventBefore, eventAfter);
+        clearTimeout(timeoutId);
+        timeoutId = setTimeout(() => {
+          this[propertyName] = true;
+        }, this.delay);
```
mflorea, line 61: Why do you wait 200ms to mark the loading state? This needs an explanation. Note that this approach makes writing functional tests a pain. It's fine for changing the layout because there are some CSS classes to wait for (``layout-cards``) but for data it's different. If you write a test that checks the pagination you have this situation:

* the test clicks on the next page link and needs to wait for something that says the live data is ready
* if you wait for ``waiting`` class to disappear then it may return before the loading starts (because the live data is not waiting right after clicking on the next page link)
* if you wait for ``waiting`` class to appear and then disappear then this may happen too fast and the test will just see no waiting class and thus fail

If you don't want to show the loading animation right away, I'm fine with that, but the live data should be marked as loading right away (in the HTML). So you may need 2 classes for this (one to mark the loading state and one to show it to the user).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
```diff
@@ -35,48 +35,52 @@
...
     <!-- Entry selector info bar -->
-    <LivedataEntrySelectorInfoBar/>
+    <LivedataEntrySelectorInfoBar />
 
+    <!-- Loading bar -->
+    <LayoutLoader />
 
     <!-- Table layout root -->
-    <table class="layout-table-root">
+    <div class="layout-table-wrapper">
```
mflorea, line 52: Would be nice to explain the need for this new wrapper.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
```diff
@@ -123,6 +127,10 @@ export default {
 
 <style>
 
+.layout-table-wrapper {
+  overflow: auto;
```
mflorea, line 131: Would be nice to explain the need for this style.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -142,17 +163,53 @@ export default {
...
+      /*
+        As the draggable plugin is taking in account every child it has for d&d
+        and there is the select-entry-all component as first child
+        we need to substract 1 to the indexes that the draggable plugin handles
+        so that it matches the true property order
+        When selection is disabled (and the select-entry-all component hidden)
+        we don't need to readjust the offset of the indexes
+        as Vue handily creates a html comment where the component shoud be
+        So that it does not messes up with indexes
+      */
```
mflorea, line 175: The previous style was OK. We use multi-line comment syntax only for documenting functions / methods and classes. For in-line comments we use the single line syntax.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -142,17 +163,53 @@ export default {
...
       this.logic.reorderProperty(e.moved.oldIndex - 1, e.moved.newIndex - 1);
     },
 
+    getNextvisibleProperty (th) {
```
mflorea, line 179:
```suggestion
    getNextVisibleProperty (th) {
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -368,7 +368,15 @@ define('xwiki-livedata', [
 
 
     fetchEntries () {
-      return liveDataSource.getEntries(this.data.query);
+      // Before fetch event
+      this.triggerEvent("beforeentryfetch");
```
mflorea, line 372:
```suggestion
      this.triggerEvent("beforeEntryFetch");
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -368,7 +368,15 @@ define('xwiki-livedata', [
...
+      // Fetch entries from data source
+      return liveDataSource.getEntries(this.data.query)
+        .then(data => {
+          // After fetch event
+          this.triggerEvent("afterentryfetch");
```
mflorea, line 377:
```suggestion
          this.triggerEvent("afterEntryFetch");
```

mflorea (changes requested):

PR #1504, opened 2021-01-28
XWIKI-18272 The list (dropdown) live data filter doesn't include the actual values

ClemDee:
* Modify BaseSelect to use objects instead of strings

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
@@ -36,7 +36,7 @@
     :selected-values="values"
     :multiple="true"
     :sort="true"
-    @change="applyFilter($event.toString())"
+    @change="applyListFilter"
   >
 
     <!--
@@ -95,10 +95,28 @@ export default {
     // but if no Array is found, the string is parsed as a singleton
     // so that it always return an array
     values () {
-      if (this.filterEntry.value) {
-        return this.filterEntry.value.split(",");
+      try {
+        // Try to parse the string as an array
+        const values = JSON.parse(this.filterEntry.value);
+        return values;
+      } catch (err) {
+        if (!this.filterEntry.value) {
+          return [];
+        } else {
+          return [this.filterEntry.value];
+        }
+      }
+    },
+  },
+
+  methods: {
+    applyListFilter (values) {
+      if (values.length === 0) {
+        this.applyFilter("");
+      } else if (values.length === 1) {
+        this.applyFilter(values[0]);
       } else {
-        return [];
+        this.applyFilter(JSON.stringify(values));
       }
     },
   },
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
@@ -33,22 +33,22 @@
   <BaseSelect
     class="filter-list"
     :options="options"
-    :selected="values"
+    :selected-values="values"
     :multiple="true"
     :sort="true"
-    @change="applyFilter(JSON.stringify($event))"
+    @change="applyFilter($event.toString())"
   >
 
     <!--
       Provide a title for the select
       List the selected options, or display "Select Values" if none is selected
     -->
-    <template #title=data>
-      <div v-if="data.selected.length === 0">
+    <template #title="data">
+      <div v-if="data.selectedValues.length === 0">
         Select Values
       </div>
       <div v-else class="selected-enum">
-        {{ data.selected.join(", ") }}
+        {{ data.selectedValues.join(", ") }}
       </div>
     </template>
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/BaseSelect.vue
```diff
@@ -54,17 +54,18 @@
...
+            :value="option.value"
+            :label="option.name"
             :checked="isSelected(option)"
             :toggle="toggleSelect"
           >
-            {{ option }}
+            {{ option.name }}
           </slot>
         </a>
       </li>
```
mflorea, line 71: You also need to add the selected values that are not in the list of known/supported options. It's not common but it can happen that the filter / constraint value (either from the configuration given by the user as input or from the live data configuration saved in the URL) becomes obsolete, meaning that it's no longer a value that is accepted on new live data entries for that property, but that doesn't mean there can't be existing entries that are still using it and we shouldn't loose this configuration, not until the user selects a different value for the filter.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/BaseSelect.vue
```diff
@@ -101,30 +102,51 @@ export default {
...
+    },
+
     isSelected (option) {
-      return this.selected.includes(option);
+      return this.selectedValues.includes(option.value);
     },
 
     select (option) {
-      if (!this.options.includes(option)) { return; }
+      if (!this.isValidOption(option)) { return; }
```
mflorea, line 134: See my comment above, the live data configuration could use obsolete values for some query filter and we shouldn't drop these values.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/BaseSelect.vue
```diff
@@ -101,30 +102,51 @@ export default {
...
+      const _selected = this.multiple ? this.selectedValues.slice() : [];
+      _selected.push(option.value);
+      this.sortSelected(_selected);
       this.$emit("change", _selected);
     },
 
     deselect (option) {
-      if (!this.options.includes(option)) { return; }
-      const index = this.selected.indexOf(option);
+      if (!this.isValidOption(option)) { return; }
```
mflorea, line 144: We shouldn't  prevent the user from "fixing" a filter / constraint by removing an obsolete value.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
```diff
@@ -95,16 +95,9 @@ export default {
...
-        const values = JSON.parse(this.filterEntry.value || "[]");
-        if (values instanceof Array) {
-          return values;
-        } else {
-          return [values];
-        }
-      } catch (err) {
-        console.warn(err);
+      if (this.filterEntry.value) {
+        return this.filterEntry.value.split(",");
```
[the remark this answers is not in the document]
ClemDee: Ah I didn't know it was actually possible to have a value with a comma!

When you say using an array as the filter value, do you mean:
- like the previous version (a stringified array as filter value):
```js
{ operator: "equals", value: "[\"option1\"]" }
```
- a true array object:
```js
{ operator: "equals", value: ["option1"] }
```

For now, I was considering that every filter values were string, because it is consistent (and easier to handle for storing the config in the url).
But if for some reason you prefer to use an array instead of a stringified array, I can change it.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/BaseSelect.vue
```diff
@@ -54,17 +54,18 @@
...
-        :key="option"
+        :key="option.value"
       >
         <a href="#"
           @click.prevent.stop="toggleSelect(option)">
           <slot
             name="option"
-            :value="option"
+            :value="option.value"
+            :label="option.name"
```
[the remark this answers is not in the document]
ClemDee: Well, because I didn't remember we agreed on that! but np it's changed! :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/BaseSelect.vue
```diff
@@ -101,30 +102,51 @@ export default {
...
       default: false,
-    }
+    },
+  },
+
+  computed: {
+    optionsValues () {
+      console.log("optionsValues", this.options.map(option => option.value));
+      console.log("this.options", this.options);
+      console.log("this", this);
```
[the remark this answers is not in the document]
ClemDee: What a horror! I didn't even catch that when adding the changes on git! :(

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/BaseSelect.vue
ClemDee: Oh, I didn't thought about this use case!

But in this case, even if we still have the option value from the saved configuration, how do we get the option pretty name to display, as it is no longer in the LD meta config ?

Also, what happens when the user deselect the option from the widget? Is it supposed to disappear from the list of available options, or does it stays in the list, even if it is no longer an option and is not selected anymore?

ClemDee:
Actually, I've realized that my branch name is incorrect: `XWIKI-18727` instead of `XWIKI-18272`. Hopefully, the commit and PR names are right, but is this wrong branch name going to be a problem for logging the issue?

PR #1512, opened 2021-02-02
XWIKI-17975: Office export is using way too much memory

surli:
## JIRA

https://jira.xwiki.org/browse/XWIKI-17975

##  Description

This work consists mainly in refactoring OfficeConverter API to not
return a byte[] that has to be kept in memory, but to return instead a
proper OfficeConverterResult object that contains the paths of the
various files resulting from the conversion. This new API is then used
wherever it's needed and the cleanup of the temporary files is only
performed at the very end of the usecases.

More specifically this has been done:
  * Add class OfficeConverterResult
  * Use it as return type of OfficeConverter#convertDocument and
    deprecate OfficeConverter#convert
  * Refactor OfficeDocument interface to deprecate getArtifacts which is
    returning a byte[] in favor of a new method getArtifactsFiles which
returns a set of file
  * Refactor the internal components to use the new APIs and in
    particular refactor completely the logic of DefaultPresentationBuilder

## TODO
- [x] Move deprecated stuff to legacy
- [x] Review velocity scripts that use script service method for conversion
- [x] Run IT tests
- [x] Add missing revapi ignores (check when building legacy-office project)


surli:
Opening the PR for review since the tests are all passing. I still need to review the scripts using office script services to ensure that the clean up is properly performed.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @return all files created during the conversion, including {@link #getOutputFile()}.
+     */
+    Set<File> getAllFiles();
+
+    /**
+     * Allows to completely delete {@link #getOutputDirectory()}.
+     */
+    void cleanup();
```
tmortagne, line 56: Instead of a custom cleanup() I think I would make OfficeConverterResult Closeable to make clear that this is something that need to be released when you don't need it. It also allows using try/resource and make possible for tools like SonarLint to catch leaks caused by missing close() calls.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @return the directory containing all output files.
+     */
+    File getOutputDirectory();
+
+    /**
+     * @return all files created during the conversion, including {@link #getOutputFile()}.
+     */
+    Set<File> getAllFiles();
```
tmortagne, line 51: Maybe we should expose Path instead of File in new APIs. It's not like File was deprecated but using NIO APIs is kind of the recommended way now.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * @return the output file of the conversion.
+     */
+    File getOutputFile();
+
+    /**
+     * @return the directory containing all output files.
+     */
+    File getOutputDirectory();
```
tmortagne, line 46: Do we really need to expose that with getAllFiles() ?

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
surli: Hmmm I like the idea but it could be a bit complicated with the current architecture, because of OfficeDocument#getArtifactFiles. This API is actually the one used in script service and those files are directly the ones coming from OfficeConverterResult. So if I perform the cleanup too soon, then the files from OfficeDocument#getArtifactFiles won't exist either which is not right. That's why I added a new API OfficeDocument#cleanupArtifacts which calls the OfficeConverterResult#cleanup, and the former method is called pretty late in the script. 

I'm not a huge fan of this overly complicated archi, but not quite sure how to simplify it since we need: 
  1. a OfficeConverterResult providing info when OfficeConverter#converDocument is called
  2. an OfficeDocument which is returned in script service and should provide all the necessary information

I could make both closeable I guess but then I'd probably need to perform a file copy when building the OfficeDocument, it sounded a bit expensive to me. If you have other ideas to simplify it, please tell me :)

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
surli: > Maybe we should expose Path instead of File in new APIs

I actually started the implem with File and I finally reverted it to use Path because all our usages actually need File, so it sounded easier to directly use File than calling everywhere `toFile`.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
surli: Yeah I'm using it in https://github.com/xwiki/xwiki-platform/pull/1512/files#diff-eb88f1ba13c6d5f799cc81ac5bf67d9dbf6123cd0fd56a985020742a7e39b311R153. It allows me to ensure that the other  temporary files that I might create will be stored in a temporary directory.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/converter/OfficeConverterResult.java
tmortagne: Well in my first message I was only suggesting replacing whatever cleanup() does by a call to Closeable#close().

surli: Oh I thought you were suggesting to only use OfficeConverterResult in try with resources.

surli: But yeah my reaction shows that I don't like really this current archi anyway :)

tmortagne: It would definitely be nice if it was possible to release every resource exposed by OfficeConverterResult with a call to close().

surli: So I actually implemented it in both OfficeConverterResult and OfficeDocument: the latter relies on the first one, so developers will have to be careful when to call the close. But normally if you're relying on the existing API you're safe.

PR #1515, opened 2021-02-10
XWIKI-17606: The Tag page should use a document tree or LiveTable to list all documents having a given tag

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-17606

- TagPlugin deprecated and replaced by TagScriptService
- Updated calls to the plugin with calls to the service in velocity scripts
- Introduction of the taggedDocument live data source, to list the documents containing a given tag
- Integration of the taggedDocument live data in the tagged documents page

### DIFF
--- a/xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataConfigurationProvider.java
+++ b/xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataConfigurationProvider.java
@@ -17,25 +17,36 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
+package org.xwiki.tag.internal.livedata;
 
-package com.xpn.xwiki.plugin.tag;
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.livedata.LiveDataConfiguration;
+import org.xwiki.livedata.LiveDataQuery;
 
 /**
- * The result of a {@link TagPlugin} operation.
- * 
+ * Live data configuration for the tag.
+ *
  * @version $Id$
+ * @since 13.1RC1
  */
-public enum TagOperationResult
+@Component
+@Named("taggedDocument")
+@Singleton
+public class TaggedDocumentLiveDataConfigurationProvider implements Provider<LiveDataConfiguration>
 {
-    /** The operation was successful, and data changes occurred. */
-    OK,
-
-    /** The operation did not change any data. */
-    NO_EFFECT,
-
-    /** The current user does not have permission to perform the operation. */
-    NOT_ALLOWED,
-
-    /** The operation failed with an exception. */
-    FAILED
+    @Override
+    public LiveDataConfiguration get()
+    {
+        LiveDataConfiguration liveDataConfiguration = new LiveDataConfiguration();
+        LiveDataQuery query = new LiveDataQuery();
+        LiveDataQuery.Source source = new LiveDataQuery.Source();
+        source.setId("tag");
+        query.setSource(source);
+        liveDataConfiguration.setQuery(query);
+        return liveDataConfiguration;
+    }
 }
--- a/xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataEntryStore.java
+++ b/xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataEntryStore.java
@@ -1,0 +1,107 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.tag.internal.livedata;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+
+import org.slf4j.Logger;
+import org.xwiki.component.annotation.Component;
+import org.xwiki.component.annotation.InstantiationStrategy;
+import org.xwiki.component.descriptor.ComponentInstantiationStrategy;
+import org.xwiki.livedata.LiveData;
+import org.xwiki.livedata.LiveDataEntryStore;
+import org.xwiki.livedata.LiveDataException;
+import org.xwiki.livedata.LiveDataQuery;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.tag.TagException;
+import org.xwiki.tag.internal.TagQueryManager;
+
+import com.xpn.xwiki.XWiki;
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
+
+/**
+ * Provides the paginated list of tags for a given page.
+ *
+ * @version $Id$
+ * @since 13.1RC1
+ */
+@Component
+@Named("taggedDocument")
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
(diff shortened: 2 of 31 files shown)


xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataEntryStore.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Provides the paginated list of tags for a given page.
+ *
+ * @version $Id$
+ * @since 13.1RC1
+ */
+@Component
+@Named("taggedDocument")
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
```
mflorea, line 59: It can be a singleton in your case because it doesn't have a state.

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataPropertyDescriptorStore.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * {@link LiveDataPropertyDescriptorStore} implementation that exposes the known tag live data columns.
+ *
+ * @version $Id$
+ * @since 13.0RC1
+ */
+@Component
+@Named("taggedDocument")
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
```
mflorea, line 42: Same, it should be a singleton.

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataSource.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Live data source for the tagged document table.
+ *
+ * @version $Id$
+ * @since 13.1RC1
+ */
+@Component
+@Named(TaggedDocumentLiveDataSource.HINT)
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
```
mflorea, line 40: It should be a singleton.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataConfigurationProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        liveDataPropertyDescriptor.setEditable(false);
+        liveDataPropertyDescriptor.setVisible(true);
+        liveDataPropertyDescriptor.setFilterable(false);
+        liveDataPropertyDescriptor.setName("Page");
+        liveDataPropertyDescriptor.setType("document");
+
+        LiveDataMeta meta = new LiveDataMeta();
+        liveDataConfiguration.setMeta(meta);
+        meta.setPropertyDescriptors(Collections.singletonList(
+            liveDataPropertyDescriptor
```
vmassol, line 72: Stupid question: why is this all done in Java? It seems a regression from before when we could have that in a wiki page and thus be editable to adjust to the user's needs.

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataEntryStore.java
```diff
@@ -0,0 +1,106 @@
...
+            XWikiContext context = this.xcontextProvider.get();
+            XWiki wiki = context.getWiki();
+            liveData.getEntries().addAll(pages.stream().map(page -> {
+                Map<String, Object> entry = new HashMap<>();
+                try {
+                    XWikiDocument document = wiki.getDocument(page, context);
+                    entry.put("page", document.getRenderedTitle(context));
+                    entry.put("page_link", wiki.getURL(page, context));
+                } catch (XWikiException e) {
+                    this.logger.warn("Failed to initialize the live table entry for the page [{}]. Cause: [{}].", page,
```
vmassol, line 95: As usual any warning means that it's a recoverable problem and that we continue doing something. We should always specify what we do in this case and what will happen for the user (in this case I think it means that data from the tags from the failing page won't be included in the LD display, right?). Also, how will the user know that the displayed table is missing some entries? Do we have a way to convey that info visually?

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataPropertyDescriptorStore.java
```diff
@@ -0,0 +1,60 @@
...
+import org.xwiki.livedata.LiveDataConfiguration;
+import org.xwiki.livedata.LiveDataException;
+import org.xwiki.livedata.LiveDataPropertyDescriptor;
+import org.xwiki.livedata.LiveDataPropertyDescriptorStore;
+
+/**
+ * {@link LiveDataPropertyDescriptorStore} implementation that exposes the known tag live data columns.
+ *
+ * @version $Id$
+ * @since 13.0RC1
```
vmassol, line 39: wrong since now I think.

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataPropertyDescriptorStore.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+@Component
+@Named(TaggedDocumentLiveDataPropertyDescriptorStore.HINT)
+@Singleton
+public class TaggedDocumentLiveDataPropertyDescriptorStore implements LiveDataPropertyDescriptorStore
+{
+    /**
+     * This component's hint.
+     */
+    public static final String HINT = "taggedDocument";
```
vmassol, line 49: Personally I don't like to make public something that shouldn't be (I prefer small duplication over extra public). Something to discuss all together probably. This is used in some places already.

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/test/java/org/xwiki/tag/TagScriptServiceTest.java
```diff
@@ -0,0 +1,295 @@
...
+        TagOperationResult tagOperationResult = this.tagScriptService.removeTagFromDocument("t1", documentReference);
+        assertEquals(TagOperationResult.NOT_ALLOWED, tagOperationResult);
+    }
+
+    @Test
+    void removeTagFromDocumentFails() throws TagException
+    {
+        DocumentReference documentReference = new DocumentReference("xwiki", "XWiki", "Doc");
+        when(this.contextualAuthorization.hasAccess(Right.EDIT, documentReference)).thenReturn(true);
+        when(this.tagDocumentManager.removeTagFromDocument("t1", documentReference)).thenThrow(TagException.class);
```
vmassol, line 216: I would write: `thenThrow(new TagException("error"))` WDYT?

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataConfigurationProvider.java
mflorea: @vmassol this is the **default** configuration for the ``tag`` live data source, so you don't have to duplicate it whenever you want to display the tagged documents using the live data macro: you just need to specify the source and the list of properties. The fact that it is done in Java doesn't imply it cannot be done from a wiki page. You can overwrite this default configuration when calling the live data macro.

Now, the class name should be improved (``Default...``) and the source id should probably be "taggedPages" or "taggedDocuments" because it doesn't return tags but pages / documents. I mean, a live data entry is a page / document, not a tag.

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataEntryStore.java
```diff
@@ -0,0 +1,106 @@
...
+            List<DocumentReference> pages =
+                this.tagQueryManager.getPages(tag, query.getLimit(), query.getOffset().intValue());
+
+            XWikiContext context = this.xcontextProvider.get();
+            XWiki wiki = context.getWiki();
+            liveData.getEntries().addAll(pages.stream().map(page -> {
+                Map<String, Object> entry = new HashMap<>();
+                try {
+                    XWikiDocument document = wiki.getDocument(page, context);
+                    entry.put("page", document.getRenderedTitle(context));
```
mflorea, line 92: Feels strange to return the page title as ``page``. It should be more explicit (e.g. ``pageTitle``). And you should return more information, like the page reference (which can serve as an id for the entry).

xwiki-platform-core/xwiki-platform-tag/xwiki-platform-tag-api/src/main/java/org/xwiki/tag/internal/livedata/TaggedDocumentLiveDataEntryStore.java
```diff
@@ -0,0 +1,106 @@
...
+                    entry.put("page_link", wiki.getURL(page, context));
```
mflorea, line 93: This is data so you should focus on the semantic: the value represents the page URL. The fact that the page URL is used later to display a link to the page is not important at this point.

mflorea (changes requested):
This PR includes formatting & refactoring changes that are not strictly related to the target issue, which makes the PR review harder. I prefer to separate them so that it's more clear what exactly are the changes required to fix the issue.

PR #1517, opened 2021-02-10
XWIKI-18304: Misalignment issue and unneeded scroll bar on the Version conflict box

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18304

- Introduction of a new diff-line-decision class to help the css selection
- Prevent the white-space pre-wrap on the diff decision td and apply it on the proposed decisions instead
- Uniformize the padding to keep an uniform view when switching between the proposed decisions.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
@@ -27,12 +27,15 @@
   border: 1px solid $theme.borderColor;
   font-family: Consolas, "Liberation Mono", "Courier New", monospace;
   font-size: 12px;
-  overflow: visible;
+  overflow: auto;
 }
 
 .diff-container table {
   /* Prevent scroll bars when the table fills the entire width. */
   margin: 0;
+  /* Having a table size of 100% leads to the display of an horizontal scroll bar when a td has a border. A 99.99% width
+     is visually indistinguishable but prevent this overflow. */
+  width: 99.99%
 }
 
 .diff-container td {
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
@@ -27,7 +27,7 @@
   border: 1px solid $theme.borderColor;
   font-family: Consolas, "Liberation Mono", "Courier New", monospace;
   font-size: 12px;
-  overflow: auto;
+  overflow: visible;
 }
 
 .diff-container table {
@@ -55,10 +55,9 @@ td.diff-line {
   white-space: pre-wrap;
 }
 
-/*
-Prevent the code formatting for diff lines containing a decision form as it breaks the display of the form.
-The code formatting and long line wrapping is instead applied on the proposed decisions preview contained in the line. 
-*/
+/* Prevent the code formatting for diff lines containing a decision form as it breaks the display of the form.
+   The code formatting and long line wrapping is instead applied on the proposed decisions preview contained in the 
+   line. */
 td.diff-line.diff-line-decision {
   white-space: normal;
   padding: 0.5em;
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
@@ -55,6 +55,21 @@ td.diff-line {
   white-space: pre-wrap;
 }
 
+/*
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
```diff
@@ -55,6 +55,21 @@ td.diff-line {
   white-space: pre-wrap;
 }
 
+/*
+Prevent the code formatting for diff lines containing a decision form as it breaks the display of the form.
```
vmassol, line 59: Maybe we should use a common comment style. Above in the same file I see for ex:

```
/**
 * Raw (Wiki Syntax) Diff
 */
```

I don't know what's the coding style for CSS files though. Seems we're missing comments in https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/

manuelleduc: `/**` is used very rarely afaics and does not feel "css" to me. It seems to be used most often to delimit subsections of the file (@mflorea can you confirm?).

manuelleduc: `/*` directly followed by a line break is used ever less often apparently. I have the habit to use this form for multi line comments. I'll do a proposal.

manuelleduc: The proposal: https://forum.xwiki.org/t/css-comments-code-style/8266

mflorea: > It seems to be used most often to delimit subsections of the file

Yes.

mflorea: On my side the expected style for this comment is this https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-13.0/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css#L209 (below on the same CSS file).

mflorea (changes requested):

manuelleduc:
Screenshots of the display:
- [before](https://up1.xwikisas.com/#L2F3KKPlyXNvtSSIN9xKkQ)
- [after](https://up1.xwikisas.com/#wQVb1zQktdUjM0e5Y9R-NA)

- The select field is position at the top right of the line
- The scrollbar at the bottom of the bottom is not displayed anymore

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
```diff
@@ -27,7 +27,7 @@
   border: 1px solid $theme.borderColor;
   font-family: Consolas, "Liberation Mono", "Courier New", monospace;
   font-size: 12px;
-  overflow: auto;
+  overflow: visible;
```
mflorea, line 30: This doesn't feel right. Any reason for this change?

manuelleduc: The intent is to avoid having a visible scrollbar at the bottom on Edge.

mflorea: Sure, but doesn't this mean that the diff can overflow the container? We show the diff in a few places where the width is limited. Have you checked that there's no overflow?

manuelleduc: I've replaced this overflow: visible with a table size of 99.99%.
I'm not a big fan of this solution but it has the advantage to preserve the current visual aspect of the table while:
- displaying the dashed border of the four sides (otherwise, the right border was hidden).
- prevent the display of an horizontal scrollbar at the bottom of the table

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/diff.css
```diff
@@ -33,6 +33,9 @@
 .diff-container table {
   /* Prevent scroll bars when the table fills the entire width. */
   margin: 0;
+  /* Having a table size of 100% leads to the display of an horizontal scroll bar when a td has a border. A 99.99% width
+     is visually indistinguishable but prevent this overflow. */
```
mflorea, line 37:
```suggestion
     is visually indistinguishable but prevents this overflow. */
```

mflorea: Looks good otherwise.

PR #1521, opened 2021-02-17
Introduce title attribute into "Go to permalink" modal dialogue

BjoernLudwigPTB:
We are working our way through our XWiki instance with the [WAVE plugin](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh) at the moment to increase accessibility, that is resolve all errors thrown by their standards.We found one very minor issue is a missing label or title attribute in a modal dialogue to copy a comments permalink. This PR is meant to fix this. It is my very first contribution to XWiki and I would be glad if there are any standards I missed, if you could advice me on where to look and what to fix.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
@@ -338,7 +338,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
         <div class="modal-body">
           <div class="input-group">
             <div class="input-group-addon">$services.icon.renderHTML('link')</div>
-            <input type="text" class="form-control"/>
+            <input type="text" class="form-control" title="Permalink"/>
           </div>
         </div>
         <div class="modal-footer">
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
@@ -338,7 +338,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
         <div class="modal-body">
           <div class="input-group">
             <div class="input-group-addon">$services.icon.renderHTML('link')</div>
-            <input type="text" class="form-control" title="Permalink"/>
+            <input type="text" class="form-control" title="$services.localization.render('core.viewers.comments.permalink')"/>
           </div>
         </div>
         <div class="modal-footer">


vmassol:
@BjoernLudwigPTB thanks a lot of the PR. However I think you'll need to use a translation for the title instead of a hardcode English translation. You could maybe check how it's done for the other titles in that same file. Let us know if you need more help. Thanks!

surli:
> We are working our way through our XWiki instance with the WAVE plugin at the moment to increase accessibility, that is resolve all errors thrown by their standards.

That's interesting, thanks for the pointer!

> We found one very minor issue is a missing label or title attribute in a modal dialogue to copy a comments permalink.

Usually we open a JIRA ticket when an issue is found, it allows us to follow our bugs and also to properly backport the fixes on the proper branches, so would be nice if you could open it on https://jira.xwiki.org. 

> It is my very first contribution to XWiki and I would be glad if there are any standards I missed, if you could advice me on where to look and what to fix.

Looks good to me after your last changes, I'm ok to merge it once you created the JIRA ticket so we can assign the commit to it.

Thanks!

BjoernLudwigPTB:
Thanks for the heads up to the Jira issue. I was not sure what process to follow for such a minor change. I opened the issue [here](https://jira.xwiki.org/browse/XWIKI-18355). If there is something unclear about it or if I should improve geiven information, let me know! We probably will propose some other changes related to this.

surli:
> I opened the issue here. If there is something unclear about it or if I should improve geiven information, let me know! We probably will propose some other changes related to this.

Some information for next time (here I directly edited the ticket):
  - type: we consider those issues as accessibility bug, so it should be of type Bug (note that we also only backport bugs usually to our supported branches, so it's better to tag those as bugs to get the fix on all supported versions)
  - component: the general component for this kind of bug is "Web - Templates & Resources" 
  - labels: for accessibility tickets we have the label "wcag" 
  - there's a pull request status field that you can use it when you create the ticket
  - assignee: if you create a PR for it, then it means you're working on it so you should assign the ticket to yourself

Also note that the commits should contain the ticket number, so they are automatically linked. We also generally reuse the ticket title as the main title of the commits, we put the needed information about what has been done in the commit in the description of the commit only. 
So for example, your first commit should have been: 
```
XWIKI-18355: Missing "title" attribute in "Go to permalink" modal dialogue

  * Introduce title into Go to permalink modal dialogue 
```

Here I will squash the commits and provides the right information when merging, but following those rules will help merging faster ;)

vmassol:
@surli you merged but I don't see the key being added to the resource file. Was that done too?

vmassol:
@BjoernLudwigPTB as to the process, see https://dev.xwiki.org/xwiki/bin/view/Community/Contributing#HFixingbugsornewFeatures2FImprovements

surli:
>  I don't see the key being added to the resource file. Was that done too?

That's an already existing key AFAICS: https://github.com/xwiki/xwiki-platform/blob/28b61b904a7e345033e55d28baf1d7d4ce747030/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties#L1032

vmassol:
> That's an already existing key AFAICS:

ah I missed that, all is good then, thanks!

BjoernLudwigPTB:
Thanks for all the hints  and heads up! The Contributing guide I already read, just did not identify the necessary hints on correct issues and commit messages. We will follow these in the future. I [already opened a more aligned issue](https://jira.xwiki.org/browse/XWIKI-18356) about something more challenging. We are looking forward to keep supporting your great work!

PR #1532, opened 2021-02-24
[misc] Integration of jest to unit test vue components

manuelleduc:
@mflorea @ClemDee This is a first prototype of the integration of jest to unit test the Vue components.
Let me know what you think but I believe this is something important to have.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
@@ -95,6 +95,23 @@
             </configuration>
           </execution>
           <execution>
+            <!-- We keep the Vue files and the build configuration files in separate folders in the source tree but
+              we need them in the same folder when building the Vue components. -->
+            <id>copy-vue-tests</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/vue/tests</outputDirectory>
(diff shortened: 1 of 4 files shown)


manuelleduc:
Currently the tests cannot be run manually using npm (and are not recognized by intellij for instance), but can be executed using `mvn test`.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
```diff
@@ -94,6 +94,23 @@
               </resources>
             </configuration>
           </execution>
+          <execution>
+            <!-- We keep the Vue files and the build configuration files in separate folders in the source tree but
+              we need them in the same folder when building the Vue components. -->
```
mflorea, line 99: This was copied from above, but I think it needs to be updated.

mflorea (commented):
Looks good, only one minor comment.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
manuelleduc: ok, it don't think we need to merge this PR, but I'll reuse this codebase when providing tests for the next evolutions. I'll also wait for @ClemDee input since he spent more time than us on it :)

mflorea: I think we should merge it. But sure, we can wait for @ClemDee opinion first.

PR #1534, opened 2021-02-25
Ld rework isview

manuelleduc:
@mflorea @ClemDee 
This is a draft of a proposition to move isView outside of displayerMixing.js

After refactoring, DefaultDisplayer is not in charge of managing the state if `isView`, but instead receives in as a props.
The users of `DefaultDisplayer` are in charge of managing the state of `isView` and to pass it to the DefaultDisplayer.
When DefaultDisplayer needs to changes it's view state, it sends at `update:isView` event, which is received by the parent component which acts accordingly.
Currently this is done automatically using the `.sync` modifier (see https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier).

I believe this refactoring simplifies the current code. Additionally, it makes it easier to watch `isView`, in particular in `DisplayHtml`, where we need to retrieve the edit form of the cell when it switches to edit mode.

WDYT?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
@@ -95,6 +95,23 @@
             </configuration>
           </execution>
           <execution>
+            <!-- We keep the Vue files and the build configuration files in separate folders in the source tree but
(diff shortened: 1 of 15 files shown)


mflorea:
> I believe this refactoring simplifies the current code.

It removes some hacks from ``displayerMixin.js`` but it also adds some code duplication in each displayer. I think it's important to reduce the complexity of writing a displayer and adding more code to the existing displayers doesn't go in this direction.

PR #1537, opened 2021-03-01
XWIKI-18255: RegisterFromAdministrationTest#testRegisterExistingUser is flickering

manuelleduc:
Sets explicitly all the registration form fields to be independent of the state of the form before the begining of the test.

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/RegisterTest.java
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/RegisterTest.java
@@ -88,12 +88,13 @@ public class RegisterTest extends AbstractTest
 
     @Test
     @IgnoreBrowsers({
-    @IgnoreBrowser(value = "internet.*", version = "8\\.*", reason="See https://jira.xwiki.org/browse/XE-1146"),
-    @IgnoreBrowser(value = "internet.*", version = "9\\.*", reason="See https://jira.xwiki.org/browse/XE-1177")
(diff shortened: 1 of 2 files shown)


xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/RegisterTest.java
```diff
@@ -97,7 +93,8 @@ public void testRegisterJohnSmith()
     })
     public void testRegisterExistingUser()
     {
-        registrationPage.fillRegisterForm(null, null, "Admin", null, null, null);
+        // Uses the empty string instead of the null value to empty the form fields (the null value just keep the value filled from the previously run test).
```
mflorea, line 96: This line is a bit too long. Do you know why the null values are ignored? Maybe  the text framework should handle them as empty strings?

manuelleduc: Currently `AbstractRegistrationPage#fillRegisterForm` is just skipping the field when the corresponding parameter is `null`.
For instance:
```java
if (firstName != null) {
  map.put("register_first_name", firstName);
}
// [...]
getForm().fillFieldsByName(map);
```
It seems to be done this way on purpose, but looking at the way this method is used, it shouldn't be too risky to consider the `null` value as the empty string for instance.

mflorea: OK, so ``null`` could be useful to keep the value unchanged because due to the way the method was designed you need to pass some values. E.g. you may want to change only the first name but not the rest when calling this method. Would be good to have this in the JavaDoc of ``fillRegisterForm``.

mflorea (approved):

PR #1542, opened 2021-03-01
Xwiki 18097

manuelleduc:


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/pom.xml
```diff
@@ -64,5 +64,11 @@
       <version>${commons.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-skin-skinx</artifactId>
+      <version>13.2-SNAPSHOT</version>
+      <scope>provided</scope>
+    </dependency>
```
manuelleduc, line 72: Currently required to populate `X-XWIKI-HTML-HEAD`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataEntryStore.java
```diff
@@ -190,4 +199,110 @@ private String getLiveTableResultsFromPage(String page, LiveDataQuery query) thr
         }
         return entries;
     }
+
+    @Override
+    public void updateAll(String entryId, Map<String, Object> newProperties)
```
manuelleduc, line 204: This might not be generic enough

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataEntryStore.java
```diff
@@ -190,4 +199,110 @@ private String getLiveTableResultsFromPage(String page, LiveDataQuery query) thr
...
+                xcontext.getWiki().saveDocument(document, xcontext);
+            }
+        } catch (AccessDeniedException | XWikiException e) {
+            // TODO log
+            e.printStackTrace();
+        }
+    }
+
+    @Override
+    public Optional<LiveDataEntryStoreEditDescriptor> getEdit(String entryId, String propertyId)
```
manuelleduc, line 240: The is not generic enough either

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataEntryStore.java
```diff
@@ -190,4 +199,110 @@ private String getLiveTableResultsFromPage(String page, LiveDataQuery query) thr
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public Optional<Object> save(Map<String, Object> entry) throws LiveDataException
+    {
+        // TODO: where do we save the new document?
+        return Optional.empty();
+    }
```
manuelleduc, line 307: should be remove as it is not relate to the current task.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntryPropertyResource.java
```diff
@@ -59,6 +66,40 @@ public Object getProperty(String sourceId, String namespace, String entryId, Str
...
+        try {
+            Optional<LiveDataEntryStoreEditDescriptor> o = this.liveDataSourceManager.get(sourceId, namespace)
+                .map(source -> {
+                    if (source instanceof WithParameters) {
+                        for (Entry<String, List<String>> stringListEntry : ui.getQueryParameters().entrySet()) {
+                            if (stringListEntry.getKey().startsWith("sourceParams.")) {
+                                String s = stringListEntry.getKey().replaceFirst("sourceParams\\.", "");
+                                ((WithParameters) source).setParameter(s, stringListEntry.getValue());
+                            }
+                        }
```
manuelleduc, line 82: populate the parameters with the livetable source params

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataEntryStore.java
```diff
@@ -190,4 +199,110 @@ private String getLiveTableResultsFromPage(String page, LiveDataQuery query) thr
...
+    {
+        String importString = "";
+        for (String s : Arrays.asList("ssrx", "ssfx", "ssx", "linkx", "jsrx", "jsfx", "jsx")) {
+
+            importString +=
+                ((SkinExtensionPluginApi) xcontext.getWiki().getPluginApi(s, xcontext)).getImportString();
+        }
+        return importString;
+    }
+
```
manuelleduc, line 281: Iintialize the javascript and css dependencies of the edit form. This is not working well, and the dependencies are always returned even if they have already been returned for a previous request on the same page.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntryResource.java
```diff
@@ -72,14 +73,14 @@ public Response updateEntry(String sourceId, String namespace, String entryId, E
...
             LiveDataEntryStore entryStore = source.get().getEntries();
-            Optional<Object> updatedEntryId = entryStore.save(entry.getValues());
-            if (updatedEntryId.isPresent()) {
-                Optional<Map<String, Object>> values = entryStore.get(updatedEntryId.get());
-                if (values.isPresent()) {
-                    Entry updatedEntry =
-                        createEntry(values.get(), updatedEntryId.get(), config.getQuery().getSource(), namespace);
-                    return Response.status(Status.ACCEPTED).entity(updatedEntry).build();
-                }
+            entryStore.updateAll(entryId, entry.getValues());
```
manuelleduc, line 76: Update the properties of the entry instead of saving a new entry.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/rest/LiveDataEntryResource.java
```diff
@@ -70,7 +70,7 @@ Entry getEntry(
      * @return the response
      * @throws Exception if updating the specified entry fails
      */
-    @PUT
+    @POST
```
manuelleduc, line 73: PUT is for creating a new resource, POST to update an existing one.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
```diff
@@ -94,6 +94,23 @@
               </resources>
             </configuration>
           </execution>
+          <execution>
+            <!-- We keep the Vue files and the build configuration files in separate folders in the source tree but
```
manuelleduc, line 98: this comment should be updated or removed

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/mocks/daterangepicker.js
```diff
@@ -0,0 +1,19 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
manuelleduc, line 19: This empty files are currently needed to mock the dependencies. Look how to avoid that.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -28,61 +28,67 @@
...
+          :value="value"
+          @focusout="applyEdit($event.target.value)"
+          @keypress.enter="applyEdit($event.target.value)"
+          @keydown.esc="cancelEdit"
       />
     </slot>
 
+    <slot name="loading" v-if="isLoading">
+      <XWikiLoader></XWikiLoader>
+    </slot>
```
manuelleduc, line 83: currently unused, but should be used to show a loading content while the form is saved or loaded

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerHtml.vue
```diff
@@ -56,15 +65,58 @@
...
+    isView: function(isView) {
+      if (isView) {
+        this.editField = "<span>loading...</span>"
+      } else {
+        const source = this.data.query.source;
+        const entryId = this.logic.getEntryId(this.entry);
+        this.logic.getEditEntryProperty(source, entryId, this.propertyId).then(res => {
+          $("head").append(res.dependencies);
+          this.editField = res.body;
+        })
```
manuelleduc, line 115: query and initialize the edit form

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerHtml.vue
```diff
@@ -31,22 +31,31 @@
...
-    <template #editor></template>
+    <!-- TODO: implement the edit widget fetch from the server -->
+    <template #editor>
+      <div v-html="editField"
+           tabindex="0"
+           v-autofocus
+           @focusout="applyEdit($event.target.value)"
+           @keypress.enter="applyEdit($event.target.value)"
+           @keydown.esc="cancelEdit">
+      </div>
```
manuelleduc, line 57: the focus management is currently poorly handled, the focus is something lost when interacting with the edit form labels, and something clicking outside of the cell does not focus out.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterBoolean.vue
```diff
@@ -0,0 +1,60 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+
+<!--
+  BooleanFilter is a custom filter that allow to filter boolean values.
+-->
+<template>
+  <div>
+    <label>
```
manuelleduc, line 27: should be updated according to the forum proposal

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -25,7 +25,7 @@ define('xwiki-livedata', [
   "xwiki-livedata-source",
   "xwiki-json-merge",
   "xwiki-livedata-polyfills"
-], function (
+], function(
```
manuelleduc, line 28: sorry for the bad large diff, I'll fix that. The only interesting change is the introduction of the getEditEntryProperty method at the bottom of the file.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -77,8 +106,33 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
...
 
+  const updateEntry = function(source, entryId, entry) {
+    return Promise.resolve($.post({
+      url: getEntryPropertiesURL(source, entryId),
+      contentType: 'application/json',
+      data: JSON.stringify({
+        values: entry
+      })
+    }));
+  }
```
manuelleduc, line 117: not sure if it is best to only update the edited entry from the response of the request, or to simply reload LD.
The second solution is easier and help keeping the displayed data in sync with the server, but might move the edited entry out of view of the user with some combinations of filters and sort.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntryResource.java
mflorea: See above. Save is not just for new entry.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/rest/LiveDataEntryResource.java
mflorea: On the contrary, I think it's the other way around in the scope of REST resources. See for instance https://restfulapi.net/rest-put-vs-post/ . I've used ``PUT`` precisely because of that:
* you create new resources with ``POST`` from the parent resource (POST request to ``/entries``)
* you update an existing resource with ``PUT`` targeting the resource itself (PUT to ``/entry/{entryId}``)

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -119,27 +125,52 @@ export default {
   // The methods for specific displayers can be found in the displayerMixin
   methods: {
 
+    toggleIsView() {
+      this.$emit('update:isView', !this.isView);
+    },
+
+    setView() {
```
mflorea, line 132: I don't see how this is a better name than ``view`` below..

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -119,27 +125,52 @@ export default {
...
+    },
+
     // Trigger View mode (switch from Editor widget to Viewer widget)
     // This should rarely be used directly as it does not validate modified data
     // Used the `applyEdit` method instead (found in the displayerMixin)
     // which call this view function after validating data
-    view () {
-      if (this.isView) { return; }
-      this.isView = true;
+    view() {
```
mflorea, line 144: Why would you call ``view`` if not to switch to view mode? You removed the line that was setting ``isView``.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -119,27 +125,52 @@ export default {
...
+    cancelEdit() {
+      // Go back to view mode
+      this.setView();
+    }
   },
-
+  watch: {
+    isView: function(newIsView) {
+      if (newIsView) this.view();
+    }
```
mflorea, line 172: This looks strange to me.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerDate.vue
```diff
@@ -77,6 +78,11 @@ export default {
   // Add the displayerMixin to get access to all the displayers methods and computed properties inside this component
   mixins: [displayerMixin],
 
+  data() {
+    return {
+      isView: true
+    }
+  },
```
mflorea, line 85: Why is this needed?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerLink.vue
```diff
@@ -62,12 +63,18 @@ export default {
     BaseDisplayer,
   },
 
+  data() {
+    return {
+      isView: true
+    }
+  },
```
mflorea, line 70: Same, why do we need to duplicate this. We didn't had to do this before.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerLink.vue
```diff
@@ -85,7 +92,7 @@ export default {
       // See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties
       // See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#interactive_content
       const interactiveContent = 'a, button, details, embed, iframe, keygen, label, select, textarea, audio[controls],'
-        + 'img[usemap], input, menu[type=toolbar], object[usemap], video[controls]';
+          + 'img[usemap], input, menu[type=toolbar], object[usemap], video[controls]';
```
mflorea, line 95: Why the increased indentation?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerLink.vue
```diff
@@ -30,9 +30,10 @@
     all the displayer default behavior
   -->
   <BaseDisplayer
-    class="displayer-link"
-    :property-id="propertyId"
-    :entry="entry"
+      class="displayer-link"
+      :property-id="propertyId"
+      :entry="entry"
+      :is-view.sync="isView"
```
mflorea, line 36: The indentation was good IMO.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerText.vue
```diff
@@ -57,6 +58,12 @@ export default {
   components: {
     BaseDisplayer,
   },
+  
+  data() {
+    return {
+      isView: true
+    }
+  },
```
mflorea, line 66: Same duplication. This is not right.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataEntryStore.java
manuelleduc: isn't this solution very specific to the liveTable source?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -1245,15 +1284,12 @@ define('xwiki-livedata', [
       });
     },
 
-
-
-
+    getEditEntryProperty(source, entryId, propertyId) {
+      return liveDataSource.getEditEntryProperty(source, entryId, propertyId);
```
mflorea, line 1288: As mentioned previously, it doesn't have to go through the live data source. This needs to be a configuration of the HTML displayer.

mflorea (changes requested):

mflorea:
After a discussion @manuelleduc and I had, we came to the conclusion that the simplest solution is to have a live data displayer dedicated to xclass properties (that could handle also document properties). This xproperty displayer could "extend" the generic HTML displayer (that we still need, check the Logging administration section for instance). The advantage is that we reuse the same "service" (``xpage=display``) used by in-place editing to edit and view xproperties. Making this "service" generic as Manuel tried is not worth the effort ATM because the xproperty displayer is actually the exception not the rule with respect to server-side rendering. None of the other displayers need to call the server to get the view/edit rendering.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataException.java
```diff
@@ -45,6 +45,19 @@ public LiveDataException(Throwable source)
         super(source);
     }
 
+    /**
+     * Constructs a new live data exception with the specified detail message.
+     *
+     * @param message the detail message
+     * @since 13.2RC1
```
manuelleduc, line 52: Of course the versions are outdated but I'll update them as soon as we are ok for the rest.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
```diff
@@ -31,10 +32,10 @@
         "displayer": {"id": "link", "propertyHref": "doc.author_url"}
       },
       {"id": "doc.objectCount", "type": "Number", "sortable": false, "filterable": false, "editable": false},
-      {"id": "_images", "displayer": "html"},
-      {"id": "_attachments", "displayer": "html"},
+      {"id": "_images", "displayer": "xClassProperty"},
+      {"id": "_attachments", "displayer": "xClassProperty"},
       {"id": "_actions", "displayer": {"id": "actions", "actions": ["edit", "delete"]}},
-      {"id": "_avatar", "displayer": "html"}
+      {"id": "_avatar", "displayer": "xClassProperty"}
```
mflorea, line 38: Same as above.

mflorea (changes requested):
Partial review.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
```diff
@@ -45,7 +45,9 @@
       {"id": "text"},
       {"id": "link"},
       {"id": "html"},
-      {"id": "actions"}
+      {"id": "xClassProperty"},
```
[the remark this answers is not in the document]
manuelleduc: I agree, renaming to `xObjectProperty`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/XClassPropertyService.java
```diff
@@ -0,0 +1,246 @@
...
+        return changedValue;
+    }
+
+    private Object updateXObject(String property, Object value, XWikiDocument document,
+        DocumentReference classReference, int objectNumber) throws XWikiException
+    {
+        Object changedValue = null;
+        XWikiContext xcontext = this.xcontextProvider.get();
+        BaseObject baseObject = document.getXObject(classReference, objectNumber);
+        BaseClass xClass = baseObject.getXClass(xcontext);
```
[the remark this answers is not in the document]
manuelleduc: Indeed, control added https://github.com/xwiki/xwiki-platform/pull/1542/files#diff-d5f05f03b0e30c52832e1eb3549e68f82cf851729fa4eec26ef1b1bf8118341bR208

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
```diff
@@ -13,13 +13,14 @@
...
       {"id": "doc.title", "type": "String", "displayer": "link"},
+      {"id": "doc.hidden", "type": "Boolean", "displayer": "boolean"},
       {
         "id": "doc.space",
         "type": "String",
         "editable": false,
         "displayer": {"id": "link", "propertyHref": "doc.space_url"}
       },
-      {"id": "doc.location", "type": "String", "editable": false, "displayer": "html"},
+      {"id": "doc.location", "type": "String", "editable": false, "displayer": "xClassProperty"},
```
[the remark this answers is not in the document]
manuelleduc: Indeed, reverted to html, thanks

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
manuelleduc: Indeed, reverted to html, thanks

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
```diff
@@ -51,7 +52,7 @@
         "sortable": false,
         "filterable": false,
         "editable": false,
-        "displayer": "html"
+        "displayer": "xClassProperty"
```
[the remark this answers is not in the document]
manuelleduc: Indeed, reverted to html, thanks

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+import org.xwiki.security.authorization.Right;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+import com.xpn.xwiki.objects.classes.BaseClass;
+
+/**
+ * Provides the operations to interact with a XClass properties of the rows of a liveTable live data.
```
mflorea, line 50:
```suggestion
 * Used by the live table source to update the live data entries that are mapped to XWiki documents (i.e. when entry properties are either document fields or xobject properties).
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+
+    @Inject
+    private Provider<XWikiContext> xcontextProvider;
+
+    @Inject
+    private Logger logger;
+
+    /**
+     * Update a property of an XObject located in the provided document reference. If the property starts with {@code
+     * doc} the corresponding document's property is updated instead. The first object of the request type if updated,
```
mflorea, line 73:
```suggestion
     * Update a property of an xobject located in the specified document. If the property starts with {@code
     * doc.} the corresponding document field is updated instead. The first xobject of the specified type (xclass reference) is updated,
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+
+        Object changedValue = updateProperty(property, value, classReference, objectNumber, document);
+
+        saveDocument(document);
+        return Optional.ofNullable(changedValue);
+    }
+
+    /**
+     * Update all the provided entries of the document. The entiers can either be from the document itself if prefixed
+     * with {@code doc.} or from an XObject instance of the classReference parameter. The first XObject found is used,
```
mflorea, line 125:
```suggestion
     * Update multiple properties of a document. The properties can either be from the document itself, if prefixed
     * with {@code doc.}, or from the first xobject instance of the specified type (class reference) found on the document;
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+     * to update another XObject, see {@link #updateAll(Map, DocumentReference, DocumentReference, int)}
+     *
+     * @param entries tge list of entries to update
```
mflorea, line 128:
```suggestion
     * @param properties the properties to update
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+     */
+    public void updateAll(Map<String, Object> entries, DocumentReference documentReference,
+        DocumentReference classReference) throws AccessDeniedException, XWikiException, LiveDataException
+    {
+        updateAll(entries, documentReference, classReference, 0);
+    }
+
+    /**
+     * Update all the provided entries of the document. The entiers can either be from the document itself if prefixed
+     * with {@code doc.} or from an XObject instance of the classReference parameter. The XObject at the provided index
```
mflorea, line 144:
```suggestion
     * Update all the provided properties of the document. The properties can either be from the document itself, if prefixed
     * with {@code doc.}, or from an XObject instance of the specified type (class reference). The XObject at the provided index
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+     * number is updated.
+     *
+     * @param entries tge list of entries to update
```
mflorea, line 147:
```suggestion
     * @param properties the properties to update
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+    {
+        this.authorization.checkAccess(Right.EDIT, documentReference);
+        XWikiContext xcontext = this.xcontextProvider.get();
+        XWikiDocument document = xcontext.getWiki().getDocument(documentReference, xcontext);
+
+        if (document.isNew()) {
+            throw new LiveDataException(NEW_DOCUMENT_UPDATE_ERROR);
+        }
+
+        for (Map.Entry<String, Object> entry : entries.entrySet()) {
```
mflorea, line 168:
```suggestion
        for (Map.Entry<String, Object> property : properties.entrySet()) {
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+            document.setAuthorReference(xcontext.getUserReference());
+            xcontext.getWiki().saveDocument(document, "LiveData update.", true, xcontext);
+        }
+    }
+
+    private Object updateProperty(String property, Object value, DocumentReference classReference, int objectNumber,
+        XWikiDocument document) throws XWikiException, LiveDataException
+    {
+        Object changedValue;
+        if (StringUtils.defaultIfEmpty(property, "").startsWith("doc")) {
```
mflorea, line 193:
```suggestion
        if (StringUtils.defaultIfEmpty(property, "").startsWith("doc.")) {
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+            changedValue = updateDocument(property.replaceFirst("doc\\.", ""), value, document);
```
mflorea, line 194: Using ``substring`` is faster and cleaner IMO.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+                String.format("XObject [%s] not found at index [%d] in [%s]", classReference, objectNumber, document));
+        }
+
+        BaseClass xClass = baseObject.getXClass(xcontext);
+
+        List<Object> properties = Arrays.asList(baseObject.getPropertyNames());
+        if (properties.contains(property)) {
+            changedValue = baseObject.get(property).toFormString();
+
+            Object nvalue;
```
mflorea, line 219:
```suggestion
            Object newValue;
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+            if (value instanceof List) {
+                List list = (List) value;
+                nvalue = IntStream.range(0, list.size())
+                    .mapToObj(i -> String.valueOf(list.get(i)))
+                    .toArray(String[]::new);
+            } else {
+                nvalue = value;
+            }
+            Map<String, Object> valueMap = new HashMap<>();
+            valueMap.put(property, nvalue);
```
mflorea, line 229:
```suggestion
            Map<String, Object> valueMap = Collections.singletonMap(property, newValue);
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+        return changedValue;
+    }
+
+    private Object updateDocument(String property, Object value, XWikiDocument document)
+    {
+        Object changedValue = null;
+        if (Objects.equals(property, "hidden")) {
+            changedValue = document.isHidden();
+            document.setHidden(Objects.equals(String.valueOf(value), "true"));
```
mflorea, line 240:
```suggestion
            document.setHidden(Boolean.valueOf(String.valueOf(value)));
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
```diff
@@ -13,6 +13,7 @@
     "propertyDescriptors": [
       {"id": "doc.name", "type": "String", "editable": false, "displayer": "link"},
       {"id": "doc.title", "type": "String", "displayer": "link"},
+      {"id": "doc.hidden", "type": "Boolean", "displayer": "boolean"},
```
mflorea, line 16: The ``Boolean`` type should use the ``boolean`` displayer by default so we can omit it, right?

mflorea (changes requested):
Partial review.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/pom.xml
```diff
@@ -73,6 +73,13 @@
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
+    <!-- Used for the translations. -->
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-localization-rest-default</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
```
mflorea, line 82: I'm not so sure about this. The live data WebJar doesn't depend on a **specific** implementation of the localization REST API. It just needs an implementation to be present. It doesn't feel right to ask explicitly for the default implementation. @tmortagne WDYT? You have module A that calls a REST URL and module B that implements the REST resource targeted by that URL. Should A depend on B? It doesn't feel right for me.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntryResource.java
```diff
@@ -79,6 +79,8 @@ public Response updateEntry(String sourceId, String namespace, String entryId, E
                     Entry updatedEntry =
                         createEntry(values.get(), updatedEntryId.get(), config.getQuery().getSource(), namespace);
                     return Response.status(Status.ACCEPTED).entity(updatedEntry).build();
+                } else {
+                    return Response.status(Status.ACCEPTED).build();
```
mflorea, line 83: Why? Because the entry can be deleted by someone else between the ``save`` and the ``get`` calls? Deserves a comment at least.

mflorea (changes requested):
Next batch.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/displayers/DisplayerBoolean.spec.js
```diff
@@ -0,0 +1,101 @@
...
+    const viewerDiv = wrapper.find('div[tabindex="0"]');
+    await viewerDiv.trigger('dblclick')
+
+    const checkbox = wrapper.find('input');
+
+    await checkbox.setChecked(false);
+
+    await wrapper.find('div[tabindex="0"]').trigger('keypress.enter');
+
+    expect([...values]).toMatchObject([{
```
mflorea, line 94: Why do you need to copy the array of values?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/displayers/DisplayerDate.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+describe('DisplayerDate.vue', () => {
+  it('Renders an entry in view mode', () => {
+    const wrapper = initWrapper(DisplayerDate, {
+      logic: {
+        isEditable() {
+          return true;
+        }
+      }
```
mflorea, line 47: Why is this needed?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/displayers/DisplayerLink.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+    const viewerDiv = wrapper.find('div[tabindex="0"]');
+    await viewerDiv.trigger('dblclick');
+
+    let inputField = wrapper.find('input');
+
+    await inputField.setValue('New Value');
+    await wrapper.find('div[tabindex="0"]').trigger('keypress.enter');
+
+    expect([...values]).toMatchObject([{
```
mflorea, line 77: Why do you need to copy the array of values?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
-    isEditable () {
-      return this.logic.isEditable({
+    // Checks if the displayed is allowed to be edited and if the livedata is in a state where this displayer can be
+    // edited.
+    isEditable() {
+      const editable = this.logic.isEditable({
         entry: this.entry,
         propertyId: this.propertyId,
       });
+      // Checks that no other cell is already editing.
```
mflorea, line 125:
```suggestion
      // Checks that no other property is currently being edited.
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
+      type: Boolean,
+      default: false
+    }
   },
 
   computed: {
-    isEditable () {
-      return this.logic.isEditable({
+    // Checks if the displayed is allowed to be edited and if the livedata is in a state where this displayer can be
+    // edited.
```
mflorea, line 119: Could you improve this comment. We don't edit the displayer. We edit the value of a property.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
+      const editable = this.logic.isEditable({
         entry: this.entry,
         propertyId: this.propertyId,
       });
+      // Checks that no other cell is already editing.
+      const noOtherEditing = this.editBus.isEditable()
+      return editable && noOtherEditing;
+    },
+    // The base value uses the value provided in the props the initial value of the form input.
+    // Once the form is edited, `this.editedValue` is defined and is used instead.
```
mflorea, line 130: Could you improve this comment. It's not clear at all **why** you need this ``baseValue``.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -24,57 +24,60 @@
...
+  instead of reimplementing the whole displayer logic each time.
 -->
 <template>
-  <div
-    :class="isView ? 'view' : 'edit'"
-    @dblclick="edit"
-    @keypress.self.enter="edit"
-    :tabindex="isEditable ? 0 : ''"
-  >
+  <div :class="{view: isView, edit: !isView, editing: isView}" ref="displayerRoot">
```
mflorea, line 30: What does ``div.view.editing`` mean? Why is this ``editing`` CSS class needed on view mode?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
mflorea: This is still an issue. The function name along with its documentation are very misleading. You have a function named ``view`` that doesn't do anything on view mode and focuses the element on edit / loading mode?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
-    // and should not be used inside a specific displayer
-    edit () {
-      if (!this.isEditable) { return; }
-      if (!this.isView) { return; }
-      this.isView = false;
+    // This method should be used to apply edit and go back to view mode.
+    // The validation of the edited property is done once the whole entry is done editing.
+    applyEdit() {
+      // When edit slot is redefined by the parent component, the edited value is always undefined and 
+      // can is ignored by the parent, which has access the the value of its own edit slot.
```
mflorea, line 173: This comment needs to be improved. I don't understand much of it.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
+      },
+      set(value) {
+        this.editedValue = value;
+      }
+    }
+  },
+
+  data() {
+    return {
+      editedValue: undefined,
```
mflorea, line 143: Why is this needed?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
+      // can is ignored by the parent, which has access the the value of its own edit slot.
+      this.$emit('saveEdit', this.editedValue);
+      // Go back to view mode
+      this.$emit('update:isView', true);
     },
 
+    // This method should be used to cancel edit and go back to view mode.
+    // This is like applyEdit but it does not save the entered value
+    cancelEdit() {
+      // Notifies the edit bus that the entry is canceled.
```
mflorea, line 182: "the entry is canceled"?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -91,55 +95,127 @@ export default {
...
+  },
+  watch: {
+    // Focus on a cell when it passes to edit mode.
+    isView: function(newIsView) {
+      // Focuses in the current cell.
+      if (newIsView) this.view();
+    }
   },
+  mounted() {
+    // Monitors clicks outside of the current cell. We switch back to edit mode whenever a click is done outside of
```
mflorea, line 197: > We switch back to edit mode

You mean view mode?

mflorea (changes requested):
Next batch.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
manuelleduc: Because `data` attributes cannot be created at runtime in `Vue`. And when the component is initialized, no value has been edited.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataConfigurationResolver.java
```diff
@@ -319,7 +328,11 @@ private DisplayerDescriptor getDisplayerConfig(String column, ObjectNode columnP
             displayerConfig.setParameter("propertyHref", getLinkTarget(column, columnProperties.get(LINK).asText()));
             displayerConfig.setParameter(HTML, columnProperties.path(HTML).booleanValue());
         } else if (columnProperties.path(HTML).booleanValue()) {
-            displayerConfig.setId(HTML);
+            if (StringUtils.isNotEmpty(options.path(CLASS_NAME).textValue())) {
+                displayerConfig.setId("xObjectProperty");
```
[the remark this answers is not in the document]
manuelleduc: That's was my intent, having a single server side resource to retrieve the edit html.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/XWikiLivedata.vue
```diff
@@ -61,6 +61,8 @@ export default {
 
   props: {
     logic: Object,
+    xObjectPropertyHelper: Object,
```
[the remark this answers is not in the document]
manuelleduc: `displayerXObjectPropertyHelper.js` is currently provided by the webjar because it needs 'jquery', 'xwiki-meta', and 'xwiki-entityReference', but afaik the only way we have to pass a webjar dependency to a Vue component is to add it to the `provide` attributes of the root component (XWikiLivedata).
I agree this does not look great but I haven't find an cleaner way to far.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,294 @@
...
+                String cacheKey = propertyName + "_cache";
+                if (entry.containsKey(propertyName)) {
+                    String property = (String) entry.get(propertyName);
+                    entry.put(propertyName, this.htmlConverter.fromHTML(property, (String) entry.get(syntaxKey)));
+                }
+                entry.remove(syntaxKey);
+                entry.remove(cacheKey);
+            }
+        }
+    }
```
manuelleduc, line 211: This PR is already large. I propose to improve this code to avoid code duplication on another commit.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntryResource.java
manuelleduc: This was because `liveTable` does not implements the `get` method. I've updated the code to rely on the `UnsupportedOperationException` exception:
https://github.com/xwiki/xwiki-platform/pull/1542/files#diff-d5f0a1eac1f978ab5a56bc091e47a16d7317ce6614b9fb207c6e350407525a82R80-R84

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -74,8 +93,17 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
   };
 
   var addEntry = function(source, entry) {
-    return Promise.resolve($.post(getEntriesURL(source), entry).then(entry => entry.values));
+    return Promise.resolve($.post(getEntriesURL(source), entry).then(e => e.values));
```
mflorea, line 96: What was wrong with the old code?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/editBus.js
```diff
@@ -0,0 +1,232 @@
...
+   * Utility method to aggregate the provided content into a vals object.
+   * For each key of the content object, a corresponding key is created in vals. If the key is found only once, the
+   * corresponding value is assigned in vals. If the key is found more than once, an array of the found values is
+   * set for the corresponding key in vals.
+   * @param content the content to add to vals
+   * @param vals the object in which to aggregate the provided content.
+   */
+  function aggregate(content, vals) {
+    for (const key in content) {
+      const value = content[key];
```
mflorea, line 46: I don't get it. If you do ``content[key]`` then the ``key`` must be **unique** in ``content``, so the only way to have a "duplicated" key (but it's not actually duplicated) is if the key already exists in ``vals`` before this function is called, but it seems you're calling with an empty object below so you never have "duplicated" keys.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/displayerMixin.js
```diff
@@ -73,27 +74,11 @@ export default {
...
-      this.$el.__vue__.view();
-    },
-
-  },
-
+    /**
+     * Generic save operation where the saved value is the 
+     */
+    genericSave() {
+      this.editBus.save(this.entry, this.propertyId, [{[this.propertyId]: this.editedValue}])
```
mflorea, line 81: Why are you sending an array of objects? I noticed this in the ``xobjectProperty`` displayer also. As I said there it would have made sense if the objects had all the same structure. But it's not the case. I find this overly complicated.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/editBus.js
```diff
@@ -0,0 +1,232 @@
...
+      }
+
+      // If a cell to save is found, we aggegrate the content to save and save it. 
+      if (canBeSaved && keyEntry) {
+        // Aggregates the content of the form values of each property.
+
+        const vals = {};
+        values[keyEntry].content.forEach((content) => {
+          aggregate(content, vals);
+        })
```
mflorea, line 142: I find this overly complicated and the root cause is the format of "content" you pass to the same function. IMO you should either pass an plain object or an array of objects with the same structure (as in the case of ``serializeArray``).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/editBus.js
```diff
@@ -0,0 +1,232 @@
...
+
+  }
+
+  /**
+   * Notifies the save of the a cell. With the current save strategy, the cell is directly save and the table is reload
+   * after this notification.
+   * @param entry the entry of the edit row
+   * @param propertyId the property id of the edited cell
+   * @param content the content of the edited cell, an array of objects with a single key and value pair,
+   *                for instance `[{"staticList1":"comedy"},{"staticList1":"romance"},{"staticList1":""}]
```
mflorea, line 210: The format of the ``content`` parameter is unexpected and unnatural. I don't see why it has to be so complicated.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
```diff
@@ -0,0 +1,294 @@
...
+        convertPropertiesFromHtml(entry, this.localSerializer.serialize(classReference), objectNumber);
+
+        for (Map.Entry<String, Object> property : entry.entrySet()) {
+            this.updateProperty(property.getKey(), property.getValue(), classReference, objectNumber, document);
+        }
+
+        saveDocument(document);
+    }
+
+    private void convertPropertiesFromHtml(Map<String, Object> entry, String className, int objectNumber)
```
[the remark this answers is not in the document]
manuelleduc: I'm not sure the interface of `update` provides enough information to call `HTMLConverter#fromHTML`. Where would I find the `RequiresHTMLConversion` value?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
manuelleduc: The lambda parameter `entry` shadows the `addEntry` method `entry` parameter.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/editBus.js
manuelleduc: You are right, this is not useful anymore. It used to be needed where I was updating a full line at first, but now that we update a single cell at a time it is not. I'll remove it.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/ModelBridge.java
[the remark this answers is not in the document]
mflorea: Indeed, for this use case the conversion to HTML needs to be done before. If the servlet filter is not an option then we need to find something else. We can leave this for later, as editing large text from the live data is not common.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -106,4 +106,13 @@ livedata.panel.sort.direction.descending=Descending
...
 
-livedata.displayer.link.noValue=(no value)
\ No newline at end of file
+livedata.displayer.link.noValue=(no value)
+
+livedata.displayer.boolean.true=True
+livedata.displayer.boolean.false=False
+
+livedata.displayer.xObjectProperty.missingDocumentName.errorMessage=Can't save an XObject property with a missing document name.
+livedata.displayer.xObjectProperty.failedToRetrieveField.errorMessage=Failed to retrieve the {mode} field.
```
mflorea, line 115: The problem is that contributors will translate also "mode" so best is to use:

```suggestion
livedata.displayer.xObjectProperty.failedToRetrieveField.errorMessage=Failed to retrieve the {0} field.
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -106,4 +106,13 @@ livedata.panel.sort.direction.descending=Descending
...
+
+livedata.filter.boolean.yes=Yes
+livedata.filter.boolean.no=No
```
mflorea, line 118: Shouldn't this be consistent with the boolean displayer? If I see ``true`` and ``false`` as values in the live data then I expect to filter by ``true`` and ``false``. So we should either use ``true`` / ``false`` everywhere (displayer and filter) or ``yes`` / ``no``.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/displayers/DisplayerLink.spec.js
```diff
@@ -0,0 +1,83 @@
...
+  it('Send events after the end of the edit', async () => {
+    const values = [];
+    const wrapper = initWrapper(DisplayerLink, {
+      logic: {
+        getDisplayerDescriptor() {
+          return {
+            html: false
+          };
+        }
+      }, editBus: {
```
mflorea, line 62:
```suggestion
      },
      editBus: {
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/displayers/DisplayerLink.spec.js
```diff
@@ -0,0 +1,83 @@
...
+        start() {
+        }, save(entry, propertyId, v) {
```
mflorea, line 64:
```suggestion
        },
        save(entry, propertyId, v) {
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/editBus.js
```diff
@@ -0,0 +1,205 @@
...
+        const editing = entryValue.editing;
+        const tosave = entryValue.tosave;
+        canBeSaved = !editing && tosave;
+
+        if (canBeSaved) break;
+      }
+
+      // If a cell to save is found, we get its content and save it. 
+      if (canBeSaved && keyEntry) {
+        // Aggregates the content of the form values of each property.
```
mflorea, line 114: Should be removed.

mflorea (changes requested):
Last remarks.

PR #1546, opened 2021-03-02
[Misc] Fixed indentation of RuleState.java

plan3d:


mflorea:
I think the right alignment was intentional. @AndreasJonsson probably felt the hexadecimal values are easier to parse visually when aligned to the right. It doesn't follow our code style but I'm not sure it's worth changing this code just for that.

plan3d:
Ok. Would you prefer to add a line in checkstyle-suppressions.xml for indentation in RuleState?

mflorea:
The ``xwiki-platform-security-api`` build is not failing ("You have 0 Checkstyle violations") so I see no reason why we would modify ``checkstyle-suppressions.xml``.

plan3d:
I'm working on a pull request for checkstyle which will fix a bug with indentation of enum constant definitions (https://github.com/checkstyle/checkstyle/pull/9331). This fix finds indentation violations in RuleState. Checkstyle uses xwiki for regression tests therefore I would like to fix xwiki code base.

tmortagne:
It indeed feels strange that checkstyle was not complaining about this indentation.

tmortagne:
A middle ground might be to do something like:

```
    /** Right state undetermined. */
    UNDETERMINED(0x0),
    /** Right is denied. */
    DENY        (0x2),
    /** Right is allowed. */
    ALLOW       (0x3);
```

WDYT @mflorea ?

tmortagne:
I would be fine to apply this pull request. Still against it @mflorea ?

mflorea:
> I'm working on a pull request for checkstyle which will fix a bug with indentation of enum constant definitions (checkstyle/checkstyle#9331)

It would have been good to provide this information in the PR description :) . It's important to know the motivation behind this change: the version of Checkstyle we're using doesn't fail the build for the bad enum indentation but future versions will fail the build, as soon as https://github.com/checkstyle/checkstyle/pull/9331 is fixed and released. So the proposed change is preparing us for the Checkstyle upgrade.

mflorea (approved):

plan3d:
Sorry for that, next time I will reference issue in PR description :-)

PR #1554, opened 2021-03-09
XWIKI-17386 Livedata table Layout Responsiveness

ClemDee:


ClemDee:
Table layout is now responsive for small layouts.

The header still support common action, like sort, order, and filter.

image

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerHtml.vue
```diff
@@ -74,4 +74,12 @@ export default {
 
 <style>
 
+/*
+ * Fix styles of xwiki html widgets when needed
+ * So that it renders best as displayer
+ */
+.xwiki-livedata .displayer-html .breadcrumb {
+  margin-bottom: 0;
```
mflorea, line 82: You should go all the way and remove also the padding and the background. See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/js/xwiki/table/livetable.css#L396 .

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
```diff
@@ -138,4 +144,49 @@ export default {
...
+    flex-direction: row;
+  }
+  .layout-table thead tr {
+    display: flex;
+    flex-direction: column;
+  }
+  /*
+   * Align the property name with the filter
+   * 38px correspond to the height of the filters inputs
+   * This is hard coded and not scalable, but I don't have better for IE :(
```
mflorea, line 162: Add a TODO to fix then using the color theme (i.e. by moving the styles to the skin where we have a less variables with the line height).

mflorea (commented):
Looks good, only 2 minor comments.

mflorea (approved):

mflorea:
Note that I reverted this PR in the end because:
* it was using CSS custom properties that unfortunately break the LESS parser we're currently using (less4j, dead since 5 years ago..). We need to fix [XWIKI-18310 (Replace less4j by an alternative)](https://jira.xwiki.org/browse/XWIKI-18310) (move to ``jlessc``) before we can use CSS custom properties (because ATM all live data styles are parsed as LESS)
* it wasn't using the standard XWiki styles for [responsive tables](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/SpecialCSSClasses/#HClassesfortheresponsiveness). The styles from this PR were probably better than the standard XWiki styles, but I prefer consistency and we should improve the standard styles.

PR #1559, opened 2021-03-15
XWIKI-17588: Display some text when there's no tag

CatalinStratu:
**- Display some text when there's no tag for guest**
Guest_user
 **-Display some text when there's no tag for logged user**
Logged_user

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
@@ -131,6 +131,10 @@
         #else
             ## TODO
         #end
+    #else
+       #if($tags.size() == 0)
+        <div class="tag-tool tag-add"><a>None</a></div>
+       #end    
     #end
 </div>
     #if($tagErrorMessage != '')
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
@@ -132,7 +132,7 @@
             ## TODO
         #end
     #else
-       #if($tags.size() == 0)
+       #if($tags.isEmpty())
         <div class="tag-tool tag-add"><a>None</a></div>
        #end    
     #end
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
@@ -132,9 +132,9 @@
             ## TODO
         #end
     #else
-       #if($tags.isEmpty())
-        <div class="tag-tool tag-add"><a>None</a></div>
-       #end    
+        #if($tags.isEmpty())
+          <div class="tag-tool tag-add"><a>None</a></div>
+        #end    
     #end
 </div>
     #if($tagErrorMessage != '')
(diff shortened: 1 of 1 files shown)


vmassol:
Thanks for the PR. However it doesn't obey what's been decided on the jira issue in the comments (which is to not display anything for guest users when there are no tags). Could you update your PR? Thx

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -131,9 +131,13 @@
         #else
             ## TODO
         #end
+    #else
+       #if($tags.size() == 0)
```
vmassol, line 135: why not `$tags.isEmpty()` which is the traditional way of doing it?

vmassol:
Note that ideally we also need a test to prove that it works. You could check the existing tag tests to see if you could easily add one?

CatalinStratu:
@vmassol 
**TEST**
**Logged user**
test_logged
**Guest User**
test_guest

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
CatalinStratu: In new commit I use this

vmassol:
> Thanks for the PR. However it doesn't obey what's been decided on the jira issue in the comments (which is to not display anything for guest users when there are no tags). Could you update your PR? Thx

This is still not done. This PR cannot be applied ATM. 

Note: It's also missing tests. What's your plan and strategy for that? Thx

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -131,9 +131,13 @@
         #else
             ## TODO
         #end
+    #else
+       #if($tags.isEmpty())
+        <div class="tag-tool tag-add"><a>None</a></div>
```
vmassol, line 136: BTW why one space for indentation instead of 2?

CatalinStratu:
Guest 0 tags
guest_no_tags

Guest one tag for test
guest_test_tag

Logged 1 tag
logged_no_tags

Logged 1 tag 
logged_1_tag

CatalinStratu:
Hi, @vmassol  can you do a review on my PR?

vmassol:
> Hi, @vmassol can you do a review on my PR?

Note: As I said it's missing some tests to prove it works. Normally any bug fix should have a test to prove that the issue is fixed and won't happen again. Since this is a vm file change, it's not that easy to write. So I'm trying to write a PageTest. Will take some time.

vmassol:
I've now added a test, please verify if it still passes (it shouldn't!) and update it accordingly with your changes :)

https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java

CatalinStratu:
> I've now added a test, please verify if it still passes (it shouldn't!) and update it accordingly with your changes :)
> 
> https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java

Thanks :))

Enygma2002:
> Guest 0 tags
> guest_no_tags

This does not look right. Careful not to introduce new bugs ("regressions") in your attempt to improve/fix another.

The `Created by ...` text needs to stay in its place, in all cases.

vmassol:
> This does not look right. Careful not to introduce new bugs ("regressions") in your attempt to improve/fix another.

I was suspecting this which is why I wrote a test for the documentTags.vm :) (which should fail with the PR).

PR #1568, opened 2021-03-22
XWIKI-18270 Live Data macro styles don't use the color theme

ClemDee:
Export style as less files instead of css

For now, every style file (even only css ones) is built as less file, because the configuration of the MiniCssExtractPlugin plugin only allow to specify the filename globally, and not according to a specific rule. This is not really an issue, as less is a superset of css, but his could lower performance as every css files will have to be parsed through the less compiler.

Also strangely, the `"[name]"` template string of webpack resolves to `xwiki-livedata-vue.umd.min` for the main less filename, instead of `xwiki-livedata-vue`. But this is just a cosmetic issue.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/config/vue.config.js
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/config/vue.config.js
@@ -24,6 +24,23 @@ module.exports = {
   // __webpack_public_path__ variable (see https://webpack.js.org/configuration/output/#outputpublicpath).
   publicPath: '',
   filenameHashing: false,
+
+  configureWebpack: config => {
+    // skip less parsing
+    const lessRules = config.module.rules.find(rule => /less/.test(rule.test));
+    lessRules.oneOf.forEach(context => {
+      const loaders = context.use;
+      const lessLoaderIndex = loaders.findIndex(loader => /less-loader/.test(loader.loader));
+      if (lessLoaderIndex !== -1) {
+        loaders.splice(lessLoaderIndex, 1);
+      }
+    });
+    // export style as less files
+    const miniCssExtractPlugin = config.plugins.find(plugin => plugin.constructor.name === "MiniCssExtractPlugin");
+    miniCssExtractPlugin.options.filename = "[name].less";
+    miniCssExtractPlugin.options.chunkFilename = "less/[name].less";
+  },
+
   chainWebpack: config => {
     // Provided dependencies (that shouldn't be bundled).
     config.externals({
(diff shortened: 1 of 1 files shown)


manuelleduc:
less-loader 8.0.0 is not supported by webpack 4, see https://github.com/webpack-contrib/less-loader/releases/tag/v8.0.0
7.3.0 is the last version supporting webpack 4 and should be used instead.

manuelleduc:
Also I have tried this change by setting the language to less in an arbitrary component.
See:  https://github.com/xwiki/xwiki-platform/pull/1586/files#diff-de33ce229b04fedd6e0b0a75393ce54859b42d887f518345d1c1e4cacca27e90R358-R369

The result is not what I expected.
- the content of the style section of the components is copied in `less/[name].less` is copied if the language is **not** less (ie, is css). Consequently, the less files contains css code
- the content of the style section of the components is not copied in the `less` directory when `lang="less"`.

Another lead is `META-INF/resources/webjars/xwiki-platform-livedata-webjar/13.3-SNAPSHOT/xwiki-livedata-vue.umd.min.less` which contains the concatenation of all the styles.

ClemDee:
> less-loader 8.0.0 is not supported by webpack 4, see https://github.com/webpack-contrib/less-loader/releases/tag/v8.0.0
> 7.3.0 is the last version supporting webpack 4 and should be used instead.

Actually, this should not be a problem, as the loader is disabled in order to get the un-parsed less files at the end. I'm a bit surprised the less-loader version is not matching the webpack one, because this comes directly from the setup from the `vue create` command, which should have ensured that all dependencies were matching.


> Also I have tried this change by setting the language to less in an arbitrary component

I wasn't able to reproduce the exact behavior you were describing, on my side everything is working as expected. On which component did you write your example less styles?

Concerning css files coming as less files, I already explained the behavior [here](https://github.com/xwiki/xwiki-platform/pull/1568#issue-597867596). If you think this is a naming issue, I can rename the "less" folder to "styles".

> Another lead is META-INF/resources/webjars/xwiki-platform-livedata-webjar/13.3-SNAPSHOT/xwiki-livedata-vue.umd.min.less which contains the concatenation of all the styles.

Reading what you were expecting, I think there is a misunderstanding.

There are two possible "outputs" for a component:
- Every component that has to be loaded right away are bundled inside one main file, named: `xwiki-livedata-vue.umd.min.js`. The styles of those components are also bundled under one file: `xwiki-livedata-vue.umd.min.less`
- Asynchronous components come in their own webpack chunk files, named `xwiki-livedata-vue.umd.min.x.js`. Those components are for example filters, displayers, or layouts. Their corresponding styles are located in the `less` folder.


I hope this can help, if it is still not clear, or the issue does not come to what I explained, please do not hesitate to come back to me.

PR #1575, opened 2021-03-25
[Misc] Migrates LiveTableGeneratorTest to a docker integration test

manuelleduc:
LiveTableGeneratorTest ported to a docker integration test in xwiki-platform-appwithinminutes-test-docker


vmassol:
Sounds good if it passes and there's no flicker! :)

PR #1579, opened 2021-03-27
XWIKI-17588: Display some text when there's no tag

CatalinStratu:
Guest:
image

Logged user:
image

Test: 
image

Detailed explanation about test modification : https://forum.xwiki.org/t/new-solution-for-xwiki-17588-issue/8526

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
@@ -88,14 +88,14 @@
 ##
 #macro(displayAddForm)
 <form action="$doc.getURL('view', "xpage=documentTags&amp;xaction=add&amp;xredirect=${xredirect}")" method="post" class="tag-add-form">
-  <div>
-      ## CSRF prevention
-    <div class="hidden"><input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /></div>
-    <label for="tag">$services.localization.render('core.tags.add.label')<br/>
-      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label><br/>
-    <span class="buttonwrapper"><input class="button button-add-tag" type="submit" value="$services.localization.render('core.tags.add.submit')"/></span>
-    <span class="buttonwrapper"><a class="button button-add-tag-cancel" href="$doc.getURL('view', "#${tagsId}")">$services.localization.render('core.tags.add.cancel')</a></span>
-  </div>
+    <div>
+        ## CSRF prevention
+        <div class="hidden"><input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /></div>
+        <label for="tag">$services.localization.render('core.tags.add.label')<br/>
+            <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label><br/>
+        <span class="buttonwrapper"><input class="button button-add-tag" type="submit" value="$services.localization.render('core.tags.add.submit')"/></span>
+        <span class="buttonwrapper"><a class="button button-add-tag-cancel" href="$doc.getURL('view', "#${tagsId}")">$services.localization.render('core.tags.add.cancel')</a></span>
+    </div>
 </form>
 #end
 ##
@@ -109,7 +109,7 @@
 #else ## display
     $xwiki.ssfx.use('uicomponents/viewers/tags.css', {'forceSkinAction': true, 'colorTheme': "$!{themeDocFullName}"})##
     $xwiki.jsfx.use('uicomponents/viewers/tags.js', true)##
-<div class="doc-tags" id="${tagsId}">
+
     #if($xwiki.tag)
         #set($hasTagsPlugin = true)
         #set($tags = $xwiki.tag.getTagsFromDocument($doc.fullName))
@@ -120,19 +120,21 @@
     ##   #if($!tags.size() == 0)
     ##     No tags
     ##   #else
-    $services.localization.render('core.tags.list.label')
-    #foreach($tag in $tags)
-        #displayTag($tag)
-    #end
     ##   #end
-    #if($hasedit)
-        #if($xwiki.tag)
-          <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-        #else
-            ## TODO
+    <div class="doc-tags" id="${tagsId}">
+        #if($hasedit || !$tags.isEmpty())
+            $services.localization.render('core.tags.list.label')
+            #foreach($tag in $tags)
+                #displayTag($tag)
+            #end
+            #if($hasedit && $xwiki.tag)
+                <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
+            #else
+                ## TODO
+            #end
         #end
-    #end
-</div>
+    </div>
+
     #if($tagErrorMessage != '')
         #error($tagErrorMessage)
     #end
--- a/xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
+++ b/xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
@@ -75,7 +75,7 @@ class DocumentTagsTest extends TemplateTest
         // - The tag label is displayed
         // - No tag is listed after the tag label
         // - No "+" link is displayed since the user doesn't have edit rights
-        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label </div>"));
+        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> </div>"));
     }
 
     @Test


xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -88,14 +88,14 @@
...
-    <span class="buttonwrapper"><a class="button button-add-tag-cancel" href="$doc.getURL('view', "#${tagsId}")">$services.localization.render('core.tags.add.cancel')</a></span>
-  </div>
+    <div>
+        ## CSRF prevention
+        <div class="hidden"><input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /></div>
+        <label for="tag">$services.localization.render('core.tags.add.label')<br/>
+            <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label><br/>
+        <span class="buttonwrapper"><input class="button button-add-tag" type="submit" value="$services.localization.render('core.tags.add.submit')"/></span>
+        <span class="buttonwrapper"><a class="button button-add-tag-cancel" href="$doc.getURL('view', "#${tagsId}")">$services.localization.render('core.tags.add.cancel')</a></span>
+    </div>
```
Enygma2002, line 98: Why did you feel the need to change the existing indentation from 2 spaces (that we use on html and velocity code) to 4 spaces?

Please limit your changes to the issue at hand and respect the existing coding conventions. https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/

Enygma2002: Note that it does seem the original change was introduced in this commit https://github.com/xwiki/xwiki-platform/commit/fbe352330de2179e1e9f451b772cad9bbc8664a1#r48788975 and it seems that the file was already using bag indentation even before your change.

CatalinStratu: > Why did you feel the need to change the existing indentation from 2 spaces (that we use on html and velocity code) to 4 spaces?
> 
> Please limit your changes to the issue at hand and respect the existing coding conventions. https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/

I apologize for this, I probably opened this file in two code editors and probably for these reasons the code was arranged differently. I made a new commit and I solved it

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -120,20 +120,22 @@
...
-    #foreach($tag in $tags)
-        #displayTag($tag)
-    #end
     ##   #end
-    #if($hasedit)
-        #if($xwiki.tag)
-          <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-        #else
-            ## TODO
+    <div class="doc-tags" id="${tagsId}">
```
surli, line 124: AFAIU you don't want to remove this div to avoid a CSS issue, shown in https://forum.xwiki.org/t/new-solution-for-xwiki-17588-issue/8526/4 I'm ok with that, but not sure why you don't use the a hidden class on it instead of displaying the label in the div only if there's some tags. IMO that's better to rely on a hidden div containing the right label: the semantic is preserved and it's easier to understand the HTML. Note that we have a dedicated "hidden" CSS class.

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -75,7 +75,7 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndNoTags() throws Exceptio
         // - The tag label is displayed
```
surli, line 75: You need to edit this.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/resources/uicomponents/viewers/tags.css
```diff
@@ -7,6 +7,9 @@
     and the next footer item (e.g. the document creator). */
   padding: 3px 2em 3px 0;
 }
+.tag-visibility {
```
surli, line 10: you don't need this class, as I mentioned there's a dedicated CSS class named `hidden` in xwiki for this.

CatalinStratu: Hi, please check again

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -120,20 +120,22 @@
...
-    #foreach($tag in $tags)
-        #displayTag($tag)
-    #end
     ##   #end
-    #if($hasedit)
-        #if($xwiki.tag)
-          <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-        #else
-            ## TODO
+    <div class="doc-tags #if(!$hasedit)hidden#end" id="${tagsId}">
```
surli, line 124: I don't get the condition here: you don't want the div to be hidden if user has not edit but there's tags.

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -120,20 +120,22 @@
...
+        $services.localization.render('core.tags.list.label')
+        #if($hasedit || !$tags.isEmpty())
+            #foreach($tag in $tags)
+                #displayTag($tag)
+            #end
+            #if($hasedit && $xwiki.tag)
+                <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
+            #else
+                ## TODO
```
surli, line 133: I know you haven't added it, but IMO this else part should be removed. It's useless and the TODO comment doesn't provide any info about what should be done next.

CatalinStratu: check again

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
CatalinStratu: check again, Please

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
surli: you're still missing some comments adjustments here.

CatalinStratu: can you tell me what else I should edit?

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -75,7 +75,7 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndNoTags() throws Exceptio
...
         // - No tag is listed after the tag label
         // - No "+" link is displayed since the user doesn't have edit rights
-        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label </div>"));
+        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags hidden\" id=\"xdocTags\"> core.tags.list.label </div>"));
```
vmassol, line 78: You should also add a new test to prove that it works when there are tags.

vmassol: @CatalinStratu for me this is the only remaining point: making sure that it still works when there are tags. Please try to add a test and if you cannot, let me know.

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
CatalinStratu: @surli  Please check

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -120,20 +120,20 @@
     ##   #if($!tags.size() == 0)
     ##     No tags
     ##   #else
```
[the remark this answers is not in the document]
CatalinStratu: @vmassol Please check

surli (changes requested):
I agree with Vincent that a new test would be better here.

CatalinStratu:
> I agree with Vincent that a new test would be better here.

Ok, I try to make this

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
CatalinStratu: > @CatalinStratu for me this is the only remaining point: making sure that it still works when there are tags. Please try to add a test and if you cannot, let me know.

Hi, I tried to do a unittest but I can't, can you help me please?

vmassol: Could you show me the code that you wrote so far?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -109,31 +109,27 @@
...
-            ## TODO
+    <div class="doc-tags #if(!$hasedit && $tags.isEmpty() )hidden#end" id="${tagsId}">
+        $services.localization.render('core.tags.list.label')
+        #if($hasedit || !$tags.isEmpty())
+            #foreach($tag in $tags)
+                #displayTag($tag)
+            #end
+            #if($hasedit && $xwiki.tag)
+                <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
+            #end
```
mflorea, line 128: This is overly complicated.. AFAICS the only change required to the original source code is to wrap the ``$services.localization`` line inside an ``#if``. Basically to show the "Tags:" label only if there are tags or the user can add a new tag.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
CatalinStratu: Check

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
CatalinStratu: I try to use this piece of code to try to test, but it doesn't work.
Source : https://github.com/xwiki/xwiki-platform/blob/0e3ec41cda7ac82f14e8cc092ff2e5f059ed1fc1/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-ui/src/test/it/org/xwiki/test/ui/tag/AddRemoveTagsTest.java

image

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
mflorea: Maybe I wasn't clear enough, but for me the diff should show only a **single** line as modified, and that line is the ``$services.localization`` line. I don't think you need other changes. I suggest you start over from the **original** source. Don't create a new PR, just read my comment above and apply my suggestion on the original source.

CatalinStratu:
image

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -109,31 +109,29 @@
...
-    #foreach($tag in $tags)
-        #displayTag($tag)
-    #end
-    ##   #end
-    #if($hasedit)
-        #if($xwiki.tag)
-          <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-        #else
-            ## TODO
+    <div class="doc-tags #if(!$hasedit && $tags.isEmpty() )hidden#end" id="${tagsId}">
```
vmassol, line 120: 2 small problems:
* Extra trailing space if the condition is not satisfied
* Extra space at `isEmpty() )`

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -109,31 +109,29 @@
...
+        $services.localization.render('core.tags.list.label')
+        #if(!$tags.isEmpty())
+            #foreach($tag in $tags)
+                #displayTag($tag)
+            #end
         #end
-    #end
-</div>
+        #if($hasedit)
+            #if($xwiki.tag)
```
vmassol, line 128: Seems the 2 conditions could be grouped, no?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -109,31 +109,29 @@
...
+                <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
```
vmassol, line 129: does that work: `<a href="$doc.getURL('view', "showTagAddForm=true` ? It seems there's a double quote problem to me.

vmassol: extra space at: `#else #displayAddForm()#end`   (you need to use `#{else}`instead)

vmassol: Also the line is too large, it would be better to simplify it and set some variables to make it more legible.

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -95,7 +95,7 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndNoTags() throws Exceptio
         // - The tag label is displayed
```
vmassol, line 95: This doesn't look good since the tag label is supposed to not be displayed!

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -95,7 +95,7 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndNoTags() throws Exceptio
...
         // - No tag is listed after the tag label
         // - No "+" link is displayed since the user doesn't have edit rights
-        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label </div>"));
+        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags hidden\" id=\"xdocTags\"> core.tags.list.label </div>"));
```
vmassol, line 98: hmm I don't see a `\\E`

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -112,7 +112,7 @@ void displayTagsWhenEditRightsAndTagPluginAvailableAndNoTags() throws Exception
         // - The tag label is displayed
         // - No tag is listed after the tag label
         // - The "+" link is displayed since the user has edit rights
-        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label "
+        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags \" id=\"xdocTags\"> core.tags.list.label "
```
vmassol, line 115: Why do say that it's good to have an extra space???

vmassol: You should not do stuff that make it easier for you when you don't understand something but instead do the right things! :)

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -135,8 +135,9 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndTags() throws Exception
         // Verify that the generated HTML matches the expectation:
         // - The tag label is displayed
         // - The tags after the tag label
-        // - No "+" link is displayed since the user doesn't have edit rights
-        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label "
+        // - No "+" link is displayed since the user doesn't have edit right
+        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags \" id=\"xdocTags\"> core.tags.list.label "
```
vmassol, line 139: Same problem here.

xwiki-platform-core/xwiki-platform-web/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -135,8 +135,9 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndTags() throws Exception
...
+
```
vmassol, line 140: why an extra empty new line?

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
```diff
@@ -117,17 +117,19 @@
...
         #end
-        #if($hasedit)
-            #if($xwiki.tag)
-                <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-            #end
+        #if($hasedit && $xwiki.tag)
+            <div class="tag-tool tag-add">
+                #if("$!{request.showTagAddForm}" == '')
+                    <a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>
+                #{else}#displayAddForm()#end
```
vmassol, line 131: Note that this will probably generate some useless extra new lines in the generated HTML. You should keep it on one line (as suggested by Marius previously). What I suggested was to introduce some velocity variables. Thanks

CatalinStratu: @vmassol Can you tell me what variables I need to add? I didn't quite understand what Marius said above about changing just one line of code

vmassol: The goal is to make the code the most readable possible. Always a good thing to do :)

For example, something like:

```
#set ($url = "$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")")
#if("$!{request.showTagAddForm}" == '')
  <a href="$url" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>
#else
  #displayAddForm()
#end
```

BTW I'm not sure about the extra new lines since the velocity `#if` shouldn't add a new line I think, did you check if that was the case or not?

vmassol:
@CatalinScr Where do we stand on this PR? what's left? Thx

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
CatalinStratu: Ok, I'll work on this! Thanks for the hint

CatalinStratu:
> @CatalinScr Where do we stand on this PR? what's left? Thx
It is finished, please see if anything needs to be added. Thank you
image

xwiki-platform-core/xwiki-platform-web/src/main/webapp/templates/documentTags.vm
mflorea: @CatalinStratu my comment was not about the line that generates the + link. My comment was about **all your changes** done to the ``documentTags.vm``: there are **too many** and I'm not convinced that all of them are needed to fix the problem. I'll repeat: all you need to do is to wrap the line (a single line!) that outputs "Tags:" with a ``#if`` so that it gets displayed only when the user has edit right (can add tags) or there are existing tags. So that means one line removed and 3 lines added. I really don't see how I can be more explicit than than, except by making the change myself..

CatalinStratu: @mflorea 
` #if(!$hasedit && $tags.isEmpty())
         $services.localization.render('core.tags.list.label')
    #end`

mflorea:
@CatalinStratu it's not enough to pass the tests, the code needs to be of good quality. ATM, as I mentioned above, your PR does too many changes that are not really required. Please check the diff (Files changed): it should contain only relevant changes (for which you have a good justification).

CatalinStratu:
> @CatalinStratu it's not enough to pass the tests, the code needs to be of good quality. ATM, as I mentioned above, your PR does too many changes that are not really required. Please check the diff (Files changed): it should contain only relevant changes (for which you have a good justification).

In such a situation do I have to do a new PR?

mflorea:
> In such a situation do I have to do a new PR?

No, you can fix the current one. I advise you to take the clean version of ``documentTags.vm`` from master and make the minimal changes that are required to fix the problem. It doesn't matter how many commits you make because they will be squashed in the end on merge. What matters is the changes listed on the "Files changed" tab of the PR.

PR #1581, opened 2021-03-29
XWIKI-5826: Send event when user authenticate

aemi-dev:
XWIKI-5826: Send event when user authenticate
- Created: UserAuthenticatedEvent.java - Create a new Event
- Modified: XWiki.java - Notify event with ObservationManager


xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+ * <p>
+ * The event also send the following parameters:
+ * </p>
+ * <ul>
+ * <li>source: the current {com.xpn.xwiki.user.api.XWikiUser} instance which has been authenticated.</li>
+ * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
+ * </ul>
+ * 
+ * @version $Id$
+ * @since 2.7RC1
```
vmassol, line 35: that doesn't look right ;)

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.xwiki.model.reference.DocumentReference;
+
+/**
+ * An event triggered when a user is authenticated.
+ * <p>
+ * The event also send the following parameters:
+ * </p>
+ * <ul>
+ * <li>source: the current {com.xpn.xwiki.user.api.XWikiUser} instance which has been authenticated.</li>
```
vmassol, line 30: Missig `{@code ...}`.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.bridge.event;
+
+import org.xwiki.model.reference.DocumentReference;
+
+/**
+ * An event triggered when a user is authenticated.
+ * <p>
+ * The event also send the following parameters:
```
vmassol, line 27: s/send/sends

vmassol:
Sounds good to me. Is this the only place where a user is created? 

Note: We may want to use the new User API instead and put this event in the user module to prepare for the future and avoid having to move it soon.

vmassol:
> Note: We may want to use the new User API instead and put this event in the user module to prepare for the future and avoid having to move it soon.

Actually this is not correct. This is about authentication, not user creation.

So maybe this code should go in the xwiki-security module somewhere?

Related to user creation: http://jira.xwiki.org/browse/XWIKI-18165

vmassol:
@aemi-dev Nice first PR! ;) We just need to agree about the location. The jira issue was created a long time ago and I'm not sure what was mentioned in it is the best location. Needs  a confirmation with the other committers.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
tmortagne, line 30: Really not a fan of this parameter. `com.xpn.xwiki.user.api.XWikiUser` hardly a reference these days to represent a user in Java. Also, it's located in xwiki-platform-oldcore and xwiki-platform-bridge does not depend on it (that's actually the whole point of xwiki-platform-bridge to not depending on xwiki-platform-oldcore).

It would probably make much more sense to put this event in xwiki-platform-user-api and pass nothing as parameter since you already have the reference in the event itself.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+ * </p>
+ * <ul>
+ * <li>source: the current {com.xpn.xwiki.user.api.XWikiUser} instance which has been authenticated.</li>
+ * <li>data: the current {com.xpn.xwiki.XWikiContext} instance</li>
```
tmortagne, line 31: Don't. There is no point in passing the XWikiContext. Those which does it are very old events and keep doing that for retro compatibility reasons.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Constructor initializing the event filter with a {@link org.xwiki.observation.event.filter.FixedNameEventFilter},
+     * meaning that this event will match only delete events affecting the same user.
+     *
+     * @param userReference the reference of the user to match
+     */
+    public UserAuthenticatedEvent(DocumentReference userReference)
```
tmortagne, line 51: Don't we want to work with UserReference instead ? I'm surprised @vmassol did not commented about that :)

vmassol: > I'm surprised @vmassol did not commented about that :)

I did actually ;)

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
vmassol: > It would probably make much more sense to put this event in xwiki-platform-user-api and pass nothing as parameter since you already have the reference in the event itself. It would increase a lot the number of modules which could use this event. I updated the jira issue description which was very old.

Yes I mentioned this in comments too but I'm not sure since authentication has its own module so right now it feels better in the security module than in the user module to me.

aemi-dev: I got your different points of view.

So far I'm trying to implement in through authentication module in security since we're talking about authentication but does the event still have to be sent in oldcore ?

Is it the only place where the user is authenticated ?

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
aemi-dev: Oups. 😅

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
aemi-dev: I just noticed that bunch of files have this error since I just really duplicate an other event file and others that I am currently working on have this typo too.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
tmortagne: > does the event still have to be sent in oldcore ?

Sure. What produces the event and where the event is defined are two different things.

aemi-dev: Of course, just wanted to be sure of that. Thanks!

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/UserAuthenticatedEvent.java
aemi-dev: What is a good example of recent Event management ?

tmortagne: Depends on which aspect but regarding the parameters the idea is that we should pass only things which bring value. The reason why old events are passing the XWikiContext is because it was the only way for the listener to access it basically while it's easy now using the Provider<XWikiContext>.

In general if you are not sure if something worth being sent with the event don't send it. It's very easy to add something as a parameter later (as long as the event is not already associated with two parameters), but it's very hard to remove it since that would be an API breakage.

aemi-dev: Okay, I fully understand the problem now. I'll take care to avoid such a thing next times.

vmassol:
> Moved: UserAuthentiatedEvent.java to UserAuthenticationEvent.java in xwiki-platform-security-authentication-api module

Well we still need to decide where it should go ;)

Also usually we use `XXXingEvent` for events before the action and `XXXedEvent` for events after the action. So before auth happened it would be `UserAuthenticatingEvent` and after auth it would be `UserAuthenticatedEvent`.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-default/src/main/java/org/xwiki/security/authentication/internal/UserAuthenticationLoggerListener.java
```diff
@@ -0,0 +1,62 @@
...
+ * Logger listener for {@link UserAuthenticationEvent} that warn about the user authentication.
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Component
+@Singleton
+@Unstable
+@Named("org.xwiki.security.authentication.internal.UserAuthenticationLoggerListener")
+public class UserAuthenticationLoggerListener extends AbstractEventListener
```
surli, line 44: Well I suggested to create it for you to test, I don't think we need it in XWiki standard. At the very least not with a warning for each authentication.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -4354,8 +4370,11 @@ public boolean checkAccess(String action, XWikiDocument doc, XWikiContext contex
         if (action.equals("skin") && SKIN_RESOURCE_SPACE_NAMES.contains(firstSpaceName)) {
             // We still need to call checkAuth to set the proper user.
             XWikiUser user = checkAuth(context);
+
             if (user != null) {
                 context.setUser(user.getUser());
+                UserReference userReference = getUserReferenceResolver().resolve(user.getUserReference());
+                getObservationManager().notify(new UserAuthenticationEvent(userReference), user);
```
surli, line 4377: Note that here you're still sending the event with an XWikiUser as source, contrary to the javadoc of the event. You don't need to send anything else except the UserReference that you put in parameter as Thomas mentioned. So you can probably just send the event (constructed with the UserReference) without more info.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
```diff
@@ -0,0 +1,70 @@
...
+import org.xwiki.user.UserReference;
+import org.xwiki.user.UserReferenceSerializer;
+
+/**
+ * This event is triggered every time a user is authenticated.
+ * <p>
+ * The event also sends the following parameters:
+ * </p>
+ * <ul>
+ * <li>source: the user reference (as a {@code org.xwiki.user.UserReference}) used for the authentication.</li>
```
surli, line 33: This is not needed since the UserReference needs to be given in contructor of the event.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * The version identifier for this Serializable class. Increment only if the <i>serialized</i> form of the class
+     * changes.
+     */
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Used to serialize user reference.
+     */
+    private static final UserReferenceSerializer<String> SERIALIZER = null;
```
surli, line 51: You forgot to remove it I guess?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
```diff
@@ -0,0 +1,70 @@
...
+ * </p>
+ * <ul>
+ * <li>source: the user reference (as a {@code org.xwiki.user.UserReference}) used for the authentication.</li>
+ * </ul>
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Unstable
+public class UserAuthenticationEvent extends AbstractCancelableEvent
```
surli, line 40: That's not good: it's not and should not be a CancelableEvent.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -739,6 +744,17 @@ private JobProgressManager getProgress()
         return this.relativeEntityReferenceResolver;
     }
 
+    private UserReferenceResolver<DocumentReference> getUserReferenceResolver()
```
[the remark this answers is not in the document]
tmortagne: Yes XWikiContext should definitely expose a UserReference like it expose the user DocumentReference.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
aemi-dev: Okay, I'll fix that

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
aemi-dev: Do I need not the serializer to serialize `UserReference` at the `UserAuthenticationEvent` creation ?

surli: No you can just store the UserReference as a field of your event and provide a getter for it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
tmortagne: Note that there might be other locations where #checkAuth is called. It would be better to place yourself in XWiki#checkAuth instead of XWiki#checkAccess IMO. Also, I feel you should compare the context user before and after the call to `getAuthService().checkAuth(context)` to send an event only when the user was not already authenticated.

tmortagne: > I feel you should compare the context user before and after the call to getAuthService().checkAuth(context) to send an event only when the user was not already authenticated.

Actually it's not going to be enough since this will still produce a new event for every single HTTP request. Right now only the XWikiAuthService knows if it's the same user or a new authentication. @surli where is the authentication protection catch failing auth ?

surli: > @surli where is the authentication protection catching failing auth ?

It's performed in various places: the authenticators (MyFormAuthenticator and MyBasicAuthenticator) and also in XWikiSecretVerifier. There was no single point for this AFAIR.

tmortagne: Indeed there is also https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/XWikiSecretVerifier.java#L85. I though the REST service was going through MyBasicAuthenticator since it is a BASIC auth in practice.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
aemi-dev: But ObservationManager#notify can not notify an event without a source as far as I tried to use it that way.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
aemi-dev: Do I need to setup a whole AuthenticationFailure-like management to achieve the UserAuthenticated thing ? Moreover, the UserAuthenticatedEvent could be a AuthenticationSuccessEvent, couldn't it be ?

surli: > Do I need to setup a whole AuthenticationFailure-like management to achieve the UserAuthenticated thing ?

Not sure what you mean: here you only need to trigger the event, but yes you might need to do it in the various location pointed by Thomas. @tmortagne if I'm correct it means we won't get the event for specific authenticators, which doesn't feel right in that case: are you sure it's not possible to just trigger the event where it was originally here?

> Moreover, the UserAuthenticatedEvent could be a AuthenticationSuccessEvent, couldn't it be ?

I'm mixed. I agree that if both events (AuthenticationFailureEvent and the new one) are triggered from same places they should have the same name pattern. Now both events have different usages and are not constructed the same way, so not entirely sure.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
tmortagne: You can pass null.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
tmortagne: > @tmortagne if I'm correct it means we won't get the event for specific authenticators, which doesn't feel right in that case: are you sure it's not possible to just trigger the event where it was originally here?

Not really sure what you are proposing. I explained what was the problem with the way it was done originally.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticationEvent.java
aemi-dev: Actually I'm feeling stupid, thanks 😅

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -109,7 +109,6 @@
 import org.xwiki.container.servlet.HttpServletUtils;
 import org.xwiki.context.Execution;
 import org.xwiki.edit.EditConfiguration;
-import org.xwiki.extension.event.ExtensionInitializedEvent;
```
tmortagne, line 112: You might want to reset any change made to this file since they are not related to your pull request.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiAuthServiceImpl.java
```diff
@@ -57,19 +66,26 @@
  *
  * @version $Id$
  */
-public class XWikiAuthServiceImpl extends AbstractXWikiAuthService
+public class XWikiAuthServiceImpl extends AbstractXWikiAuthService implements EventListener
```
tmortagne, line 69: Why ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiAuthServiceImpl.java
```diff
@@ -606,4 +669,26 @@ protected String stripContextPathFromURL(URL url, XWikiContext context)
...
+    public List<Event> getEvents()
+    {
+        return LISTENER_EVENTS;
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        if (event instanceof UserAuthenticatedEvent) {
+            LOGGER.info("Authenticated user [{}]", source.toString());
```
tmortagne, line 685: Is this some debug leftover ?

aemi-dev: At the moment yes, since I know I have to implement tests to make it a acceptable PR.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiAuthServiceImpl.java
```diff
@@ -309,6 +353,8 @@ public XWikiUser checkAuth(String username, String password, String rememberme,
                 return null;
             }
 
+            om.notify(new UserAuthenticatedEvent(),getContextUserReference(principal, context));
```
tmortagne, line 356: Wasn't the plan to put that in MyFormAuthenticator and MyBasicAuthenticator ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
aemi-dev: I didn't notice this one and yes for sure.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiAuthServiceImpl.java
aemi-dev: As the following, bug leftover.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiAuthServiceImpl.java
aemi-dev: Why not in XWikiAuthServiceImpl ? It is used by both, so it fulfills the need to get an event in these and it allows us to trigger any use of this auth service.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
```diff
@@ -219,4 +243,45 @@ public void showLogin(HttpServletRequest request, HttpServletResponse response,
     {
         showLogin(request, response, this.realmName);
     }
+
+    private ObservationManager getObservationManager() {
```
tmortagne, line 247: Not applying the XWiki codestyle.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyFormAuthenticator.java
```diff
@@ -292,4 +315,45 @@ public boolean processLogout(SecurityRequestWrapper securityRequestWrapper,
         }
         return result;
     }
+
+    private ObservationManager getObservationManager() {
```
tmortagne, line 319: There is a lot of duplication, seems to me we would need to put all that in a component used by both MyFormAuthenticator an MyBasicAuthenticator.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,78 @@
...
+import org.xwiki.stability.Unstable;
+import org.xwiki.user.UserReference;
+
+/**
+ * This event is triggered every time a user is authenticated.
+ * <p>
+ * The event also sends the following parameters:
+ * </p>
+ * <ul>
+ * <li>source: the user reference (as a {@code org.xwiki.user.UserReference}) used for the authentication.</li>
```
tmortagne, line 32: The javadoc is not accurate.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,78 @@
...
+ * </ul>
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Unstable
+public class UserAuthenticatedEvent extends AbstractFilterableEvent
```
tmortagne, line 39: Extending AbstractFilterableEvent is totally useless since you don't use anything in it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
```diff
@@ -56,6 +76,8 @@ public boolean processLogin(SecurityRequestWrapper request, HttpServletResponse
             // show the basic authentication window again.
             showLogin(request.getCurrentRequest(), response);
             return true;
+        } else {
+            getObservationManager().notify(new UserAuthenticatedEvent(getContextUserReference(principal)), null);
```
tmortagne, line 80: Seems to me this should be located in checkLogin instead. In general your  criteria to find the location should be "right after the call to request.setUserPrincipal() which register the user in the session" IMO.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
```diff
@@ -73,6 +95,8 @@ public boolean processLogin(String username, String password, String rememberme,
             // make sure the Principal contains wiki name information
             if (!StringUtils.contains(principal.getName(), ':')) {
                 principal = new SimplePrincipal(context.getWikiId() + ":" + principal.getName());
+
+                getObservationManager().notify(new UserAuthenticatedEvent(getContextUserReference(principal)), null);
```
tmortagne, line 99: The point of this if is to modify the principal, but in both cases the user is authenticated.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
aemi-dev: Since `MyBasicAuthenticator#checkLogin` is a static method, I can't access getObservationManger nor getContextUserReference. However, `MyBasicAuthenticator#checkLogin` is only used twice and I can add this event sending right after it each time and that is what I did in the next commit.

aemi-dev:
> This is more an `equals` implementation than a `matches` one. Keep in mind that most listeners will declare a `UserAuthenticatedEvent` where the user reference is null because they want to be called for any authenticated user.

Okay, I'll fix this and prior comments about code styles and also components.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyFormAuthenticator.java
```diff
@@ -292,4 +317,45 @@ public boolean processLogout(SecurityRequestWrapper securityRequestWrapper,
...
+    private UserReference getContextUserReference(Principal principal)
+    {
+        UserReference contextUserReference;
+
+        if (principal != null) {
+            // Ensures that the wiki part is removed if specified in the Principal name and if it's not the same wiki
+            // as the current wiki.
+            DocumentReference userDocumentReference =
+                getDocumentReferenceResolver().resolve(principal.getName());
+            contextUserReference = getUserReferenceResolver().resolve(userDocumentReference);
```
surli, line 354: So here since you want to get the UserReference from  a String directly (principal.getName()), you don't need to go through the DocumentReference resolution first: you can directly use a `UserReferenceResolver<String>`. This is actually even better since `UserReferenceResolver<DocumentReference>` should not really be used.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Wraps an {@link org.xwiki.observation.ObservationManager} and a {@link org.xwiki.user.UserReferenceResolver} to
+ * notify about user who are authenticated through {@link com.xpn.xwiki.user.impl.xwiki.MyFormAuthenticator} and {@link
+ * com.xpn.xwiki.user.impl.xwiki.MyBasicAuthenticator}.
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Unstable
+public class UserAuthenticatedManager
```
tmortagne, line 45: I really don't think we want to make that public right now, so I would move it in an internal package (for example `com.xpn.xwiki.internal.user`). I would also make it a component instead of a simple Java object so that it can use injection (see https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java#L99 for a common example of a utility internal component declaring itself as role to avoid useless plumbing for something which is internal).

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -78,9 +67,21 @@ public UserReference getUserReference()
 
     @Override
     public boolean matches(Object other)
+    {
+        return other instanceof UserAuthenticatedEvent;
+    }
+
+    @Override
+    public boolean equals(Object other)
```
tmortagne, line 75: My point in the previous comment was that the implementation of matches was wrong and not that you should implement equals (it's useless IMO). Also, your hashCode implementation was wrong (it's supposed to be consistent with equals) but anyway you should remove both equals and hashCode since we don't really need them.

aemi-dev: I got it. It will be fixed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+ * Wraps an {@code ObservationManager} and a {@code UserReferenceResolver} to
+ * notify about user who are authenticated through {@code MyFormAuthenticator} and
+ * {@code MyBasicAuthenticator}.
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Component(roles = UserAuthenticatedManager.class)
+@Singleton
+@Unstable
```
surli, line 46: This is an internal component so we don't need Unstable annotation here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+public class UserAuthenticatedManager
```
surli, line 47: Not a big fan of this name, it looks like you handle user authentication while here you only want to handle events when user is authenticated. I'd prefer something like `UserAuthenticatedEventManager`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.xpn.xwiki.internal.user;
```
surli, line 21: We want to move things out of oldcore so if possible it would be better to put that new component in another module.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
tmortagne: +1

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+@Component(roles = UserAuthenticatedManager.class)
+@Singleton
+@Unstable
+public class UserAuthenticatedManager
+{
+    /**
+     * The unique instance of this class.
+     */
+    public static final UserAuthenticatedManager INSTANCE = new UserAuthenticatedManager();
```
surli, line 52: That's really wrong: it's the component manager's role to handle the creation of the instance.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
```diff
@@ -35,10 +36,15 @@
 
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.internal.user.UserAuthenticatedManager;
 import com.xpn.xwiki.web.Utils;
 
 public class MyBasicAuthenticator extends BasicAuthenticator implements XWikiAuthenticator
 {
+
+    @Inject
```
tmortagne, line 45: You cannot do that, MyBasicAuthenticator is not a component. You need to use the static getComponent like you had before the refactoring for other components.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyFormAuthenticator.java
```diff
@@ -36,14 +42,13 @@
...
 
-import com.xpn.xwiki.XWikiContext;
-import com.xpn.xwiki.XWikiException;
-import com.xpn.xwiki.web.Utils;
-
 public class MyFormAuthenticator extends FormAuthenticator implements XWikiAuthenticator
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(MyFormAuthenticator.class);
 
+    @Inject
```
tmortagne, line 49: Same comment.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Inject
+    private Logger logger;
+
+    @Inject
+    @Named("observer")
+    private ObservationManager observationManager;
```
surli, line 59: Here you want to inject the default `ObservationManager` implementation so you shouldn't rely on `@Named`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Inject
+    @Named("referenceresolver")
+    private UserReferenceResolver<String> userReferenceResolver;
```
surli, line 63: Same here: we should rely on the default implem of the role.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * The unique instance of this class.
+     */
+    public static final UserAuthenticatedManager INSTANCE = new UserAuthenticatedManager();
+
+    @Inject
+    private Logger logger;
+
+    @Inject
+    @Named("observer")
```
tmortagne, line 58: This is the same as the second parameter of getComponent (that's right, you don't have any :)).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+    private ObservationManager observationManager;
+
+    @Inject
+    @Named("referenceresolver")
+    private UserReferenceResolver<String> userReferenceResolver;
+
+    /**
+     * Nothing to pass to this class.
+     */
+    public UserAuthenticatedManager()
```
tmortagne, line 68: I don't see the point of this constructor since you don't have any other it's already implicit.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
```diff
@@ -117,6 +126,8 @@ public static Principal checkLogin(SecurityRequestWrapper request, HttpServletRe
 
                 request.setUserPrincipal(principal);
 
+                UserAuthenticatedManager.INSTANCE.notify(principal.getName());
```
tmortagne, line 129: You need to use it as a component otherwise the injections in it won't work.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        // Voluntarily empty. We want to have a single instance of this class.
+    }
+
+    /**
+     * @param userReference {@code UserReference} of user who triggers a UserAuthenticatedEvent
+     */
+    public void notify(UserReference userReference)
```
surli, line 76: This should not be public since it's never used externally and it's an internal API.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+    public void notify(String stringUserReference)
+    {
+        UserReference userReference = this.userReferenceResolver.resolve(stringUserReference);
+        this.notify(new UserAuthenticatedEvent(userReference));
+    }
+
+    /**
+     * @param event {@code UserAuthenticatedEvent} that has already been created
+     */
+    public void notify(UserAuthenticatedEvent event)
```
surli, line 93: Same: this should remain private for now, it's not needed externally.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * @param userReference {@code UserReference} of user who triggers a UserAuthenticatedEvent
+     */
+    public void notify(UserReference userReference)
+    {
+        this.notify(new UserAuthenticatedEvent(userReference));
+    }
+
+    /**
+     * @param stringUserReference string form of the reference of user who triggers a UserAuthenticatedEvent
```
surli, line 82: This doc is wrong, and you should provide more information on what the method actually does (and link the UserAuthenticatedEvent).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+        UserReference userReference = this.userReferenceResolver.resolve(stringUserReference);
+        this.notify(new UserAuthenticatedEvent(userReference));
+    }
+
+    /**
+     * @param event {@code UserAuthenticatedEvent} that has already been created
+     */
+    public void notify(UserAuthenticatedEvent event)
+    {
+        this.logger.debug("User authenticated for [{}]", event.getUserReference());
```
surli, line 95: This logger debug is not needed, now I'm not quite sure if it needs to be removed or not. I'm not quite sure about our best practices for it, wdyt @tmortagne @vmassol .

tmortagne: I would say there is never too much debug log, so I leave it to developer choice.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/XWikiAuthServiceImpl.java
```diff
@@ -606,4 +606,4 @@ protected String stripContextPathFromURL(URL url, XWikiContext context)
 
         return strippedURL;
     }
-}
+}
```
surli, line 609: Change to be removed from the PR.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,74 @@
...
+ * This event is triggered every time a user is authenticated.
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Unstable
+public class UserAuthenticatedEvent implements Event
+{
+    /**
+     * Reference of the user who triggered the authentication event.
```
surli, line 36: It's not the user *who* triggered the authentication event, it's the user who gets authenticated.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
```diff
@@ -0,0 +1,99 @@
...
+import org.xwiki.observation.ObservationManager;
+import org.xwiki.security.authentication.UserAuthenticatedEvent;
+import org.xwiki.stability.Unstable;
+import org.xwiki.user.UserReference;
+import org.xwiki.user.UserReferenceResolver;
+
+/**
+ * Wraps an {@code ObservationManager} and a {@code UserReferenceResolver} to
+ * notify about user who are authenticated through {@code MyFormAuthenticator} and
+ * {@code MyBasicAuthenticator}.
```
vmassol, line 39: Small note that the best javadoc comments explain WHY something is needed. I haven't really followed the PR but just reading this javadoc and seeing the notify methods, it makes me wonder why we need a new ObservationManager, since the default one should already be sending the notifications.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    private UserReference userReference;
+
+    /**
+     * This event will match any other document event of the same type.
+     */
+    public UserAuthenticatedEvent()
+    {
+        super();
```
surli, line 45: This is a useless call.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * This event will match only events of the same type affecting the same document.
+     *
+     * @param userReference the reference related to the user
+     */
+    public UserAuthenticatedEvent(UserReference userReference)
+    {
+        super();
```
surli, line 55: same

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-api/src/main/java/org/xwiki/security/authentication/UserAuthenticatedEvent.java
```diff
@@ -0,0 +1,74 @@
...
+        this.userReference = userReference;
+    }
+
+    /**
+     * @return the {@link UserReference} of the source user reference.
```
surli, line 60: This doc doesn't provide right information

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
aemi-dev: It seems to be the right name yes.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/user/UserAuthenticatedManager.java
aemi-dev: Trying to move it to `xwiki-platform-security-authentication-default` but apparently there is a cyclic dependency

tmortagne: Since it's an internal tool used only in oldcore right now I really don't see the point of trying that hard to move it elsewhere. We can think about it if it's needed by something else one day.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyBasicAuthenticator.java
```diff
@@ -35,10 +35,21 @@
...
 import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.internal.user.UserAuthenticatedEventNotifier;
 import com.xpn.xwiki.web.Utils;
 
 public class MyBasicAuthenticator extends BasicAuthenticator implements XWikiAuthenticator
 {
+
+    private UserAuthenticatedEventNotifier userAuthenticatedEventNotifier;
+
+    private UserAuthenticatedEventNotifier getUserAuthenticatedEventNotifier() {
```
tmortagne, line 46: This is not following XWiki codestyle.

aemi-dev: Is this not intended to be reviewed by checkstyle at build time ?

Btw I'll fix it.

tmortagne: MyBasicAuthenticator is skipped in the checkstyle configuration as far as I can see.

But anyway relying on the build for that is not very effective, you should configure your IDE formatter and use it instead so that you don't have to worry about that at build time.

aemi-dev: Yes of course, I need to dig deeper in IntelliJ settings.

vmassol: See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HIntelliJIDEA

surli (approved):
LGTM

PR #1586, opened 2021-03-30
XWIKI-18492: Add support for compiling LESS files from WebJar at runtime in the context of the skin

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18492

- pass the servlet response to filterResource if it exists
- less files are compiled to css when evaluate=true

### DIFF
--- a/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
+++ b/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
@@ -42,6 +42,7 @@ import org.xwiki.resource.ResourceReferenceHandler;
 import org.xwiki.resource.ResourceReferenceHandlerChain;
 import org.xwiki.resource.ResourceReferenceHandlerException;
 import org.xwiki.resource.ResourceType;
+import org.xwiki.stability.Unstable;
 import org.xwiki.tika.internal.TikaUtils;
 
 /**
@@ -82,9 +83,7 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
             InputStream resourceStream = getResourceStream(typedResourceReference);
             if (resourceStream != null) {
                 try {
-                    HttpServletResponse response = getHttpServletResponse();
-                    serveResource(typedResourceReference,
-                        filterResource(typedResourceReference, resourceStream, response));
+                    serveResource(typedResourceReference, filterResource(typedResourceReference, resourceStream));
                 } catch (ResourceReferenceHandlerException e) {
                     this.logger.error(e.getMessage(), e);
                     sendError(HttpStatus.SC_INTERNAL_SERVER_ERROR, e.getMessage());
@@ -120,7 +119,8 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
         Request request = this.container.getRequest();
         if (request instanceof ServletRequest
             && ((ServletRequest) request).getHttpServletRequest().getHeader("If-Modified-Since") != null
-            && isResourceCacheable(resourceReference)) {
+            && isResourceCacheable(resourceReference))
+        {
             // The user probably used F5 to reload the page and the browser checks if there are changes.
             Response response = this.container.getResponse();
             if (response instanceof ServletResponse) {
@@ -164,8 +164,7 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
         throws ResourceReferenceHandlerException
     {
         InputStream resourceStream = rawResourceStream;
-        String resourceName = getResourceName(resourceReference);
-
+        
         // Make sure the resource stream supports mark & reset which is needed in order be able to detect the
         // content type without affecting the stream (Tika may need to read a few bytes from the start of the
         // stream, in which case it will mark & reset the stream).
@@ -176,9 +175,10 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
         try {
             Response response = this.container.getResponse();
             setResponseHeaders(response, resourceReference);
-            response.setContentType(TikaUtils.detect(resourceStream, resourceName));
+            response.setContentType(getContentType(resourceStream, resourceReference));
             IOUtils.copy(resourceStream, response.getOutputStream());
         } catch (Exception e) {
+            String resourceName = getResourceName(resourceReference);
             throw new ResourceReferenceHandlerException(String.format("Failed to read resource [%s]", resourceName), e);
         } finally {
             IOUtils.closeQuietly(resourceStream);
@@ -186,18 +186,20 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
     }
 
     /**
-     * Filter the resource before sending it to the client.
+     * Computes the content type of the resource. By default the content type is inferred by {@link
+     * TikaUtils#detect(InputStream, String)} based on the resource content and name.
      *
-     * @param resourceReference the resource to filter
-     * @param resourceStream the resource content
-     * @return the filtered resource content
-     * @deprecated use {@link #filterResource(ResourceReference, InputStream, HttpServletResponse)} instead
+     * @param resourceStream the stream of the requested resource
+     * @param resourceReference the reference of the request resource
+     * @return the content type of the resource
+     * @throws IOException in case of error during the content type analysis
+     * @since 13.3RC1
      */
-    @Deprecated
-    protected InputStream filterResource(R resourceReference, InputStream resourceStream)
-        throws ResourceReferenceHandlerException
+    @Unstable
+    protected String getContentType(InputStream resourceStream, R resourceReference)
+        throws IOException
     {
-        return resourceStream;
+        return TikaUtils.detect(resourceStream, getResourceName(resourceReference));
     }
 
     /**
@@ -205,12 +207,9 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
      * 
      * @param resourceReference the resource to filter
      * @param resourceStream the resource content
-     * @param response the servlet response if it exists, {@code null} otherwise
      * @return the filtered resource content
-     * @since 13.3RC1
      */
-    protected InputStream filterResource(R resourceReference, InputStream resourceStream,
-        HttpServletResponse response)
+    protected InputStream filterResource(R resourceReference, InputStream resourceStream)
         throws ResourceReferenceHandlerException
     {
         return resourceStream;
@@ -259,20 +258,5 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
                     String.format("Failed to return status code [%s].", statusCode), e);
             }
         }
-    }
-
-    /**
-     * @return the {@link HttpServletResponse} of the query if the response is an instance of {@link ServletResponse},
-     *     {@code null} otherwise.
-     */
-    private HttpServletResponse getHttpServletResponse()
-    {
-        HttpServletResponse response;
-        if (this.container.getResponse() instanceof ServletResponse) {
-            response = ((ServletResponse) this.container.getResponse()).getHttpServletResponse();
-        } else {
-            response = null;
-        }
-        return response;
     }
 }
--- a/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
+++ b/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
@@ -82,7 +82,9 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
             InputStream resourceStream = getResourceStream(typedResourceReference);
             if (resourceStream != null) {
                 try {
-                    serveResource(typedResourceReference, filterResource(typedResourceReference, resourceStream));
+                    HttpServletResponse response = getHttpServletResponse();
+                    serveResource(typedResourceReference,
+                        filterResource(typedResourceReference, resourceStream, response));
                 } catch (ResourceReferenceHandlerException e) {
                     this.logger.error(e.getMessage(), e);
                     sendError(HttpStatus.SC_INTERNAL_SERVER_ERROR, e.getMessage());
@@ -185,12 +187,30 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
 
     /**
      * Filter the resource before sending it to the client.
-     * 
+     *
      * @param resourceReference the resource to filter
      * @param resourceStream the resource content
      * @return the filtered resource content
+     * @deprecated use {@link #filterResource(ResourceReference, InputStream, HttpServletResponse)} instead
      */
+    @Deprecated
     protected InputStream filterResource(R resourceReference, InputStream resourceStream)
+        throws ResourceReferenceHandlerException
+    {
+        return resourceStream;
+    }
+
+    /**
+     * Filter the resource before sending it to the client.
+     * 
+     * @param resourceReference the resource to filter
+     * @param resourceStream the resource content
+     * @param response the servlet response if it exists, {@code null} otherwise
+     * @return the filtered resource content
+     * @since 13.3RC1
+     */
+    protected InputStream filterResource(R resourceReference, InputStream resourceStream,
+        HttpServletResponse response)
         throws ResourceReferenceHandlerException
     {
         return resourceStream;
@@ -239,5 +259,20 @@ public abstract class AbstractServletResourceReferenceHandler<R extends Resource
                     String.format("Failed to return status code [%s].", statusCode), e);
             }
         }
+    }
+
+    /**
+     * @return the {@link HttpServletResponse} of the query if the response is an instance of {@link ServletResponse},
+     *     {@code null} otherwise.
+     */
+    private HttpServletResponse getHttpServletResponse()
+    {
+        HttpServletResponse response;
+        if (this.container.getResponse() instanceof ServletResponse) {
+            response = ((ServletResponse) this.container.getResponse()).getHttpServletResponse();
+        } else {
+            response = null;
+        }
+        return response;
     }
 }
--- a/xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
+++ b/xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
@@ -98,9 +98,9 @@ public class WebJarsResourceReferenceHandler extends AbstractServletResourceRefe
         if (!isResourceCacheable(resourceReference)) {
             String resourceName = getResourceName(resourceReference);
             if (resourceName.endsWith(LESS_FILE_EXTENSION)) {
-                stream = this.lessFilter.handle(resourceStream, resourceName);
+                stream = this.lessFilter.filter(resourceStream, resourceName);
             } else {
-                stream = this.velocityFilter.handle(resourceStream, resourceName);
+                stream = this.velocityFilter.filter(resourceStream, resourceName);
             }
         } else {
             stream = super.filterResource(resourceReference, resourceStream);
@@ -121,6 +121,9 @@ public class WebJarsResourceReferenceHandler extends AbstractServletResourceRefe
         WebJarsResourceReference resourceReference) throws IOException
     {
         String mimeType;
+        // When the resource is not cacheable and the request resource is a less file, we compile it to css.
+        // In this case, the content type must be explicitly set to text/css instead of the default text/x-less
+        // that would otherwise be computed from the resource filename.
         if (!isResourceCacheable(resourceReference) && getResourceName(resourceReference)
             .endsWith(LESS_FILE_EXTENSION))
         {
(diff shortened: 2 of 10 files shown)


xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
```diff
@@ -185,17 +187,35 @@ private void serveResource(R resourceReference, InputStream rawResourceStream)
...
      * @param resourceStream the resource content
      * @return the filtered resource content
+     * @deprecated use {@link #filterResource(ResourceReference, InputStream, HttpServletResponse)} instead
      */
+    @Deprecated
     protected InputStream filterResource(R resourceReference, InputStream resourceStream)
         throws ResourceReferenceHandlerException
     {
         return resourceStream;
     }
```
manuelleduc, line 201: I'll move this to legacy if we agree to add the `HttpServletResponse` parameter.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
```diff
@@ -90,27 +97,79 @@ protected String getResourceName(WebJarsResourceReference resourceReference)
...
-                this.velocityManager.getVelocityEngine().evaluate(this.velocityManager.getVelocityContext(), writer,
-                    resourceName, new InputStreamReader(resourceStream, UTF8));
-                return new ByteArrayInputStream(writer.toString().getBytes(UTF8));
-            } catch (Exception e) {
-                throw new ResourceReferenceHandlerException(
-                    String.format("Failed to evaluate the Velocity code from WebJar resource [%s]", resourceName), e);
+            if (resourceName.endsWith(".less")) {
+                stream = lessResourceHandler(resourceReference, resourceStream, response);
+            } else {
+                stream = defaultResourceHandler(resourceStream, resourceName);
```
manuelleduc, line 114: WDYT of moving these two calls to dedicated internal components?

mflorea: +1 to introduce a ``WebJarsResourceFilter`` component with two implementations ("velocity" and "less").

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
manuelleduc: see 
- the role: https://github.com/xwiki/xwiki-platform/pull/1586/files#diff-16deb1721a8b7b94d4d569c7592eedd1b3791da4bba55054b5cfc6e71e02e50bR35
- the less component: https://github.com/xwiki/xwiki-platform/pull/1586/files#diff-5a5014f40184c0b97b9f3a265b874c94aa5a8eb26cb713872f0cfea5f05600bdR48
- the velocity component: https://github.com/xwiki/xwiki-platform/pull/1586/files#diff-259a2897677e7d165d728f64a24176c9b19af4450037396d41349f93eeb01d16R46

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
manuelleduc: Indeed, way cleaner that way, thanks for the suggestion.
In practice, `setResponseHeaders` is called before et content type definition, so setting the `Content-Type` here would have no effect since the value would be overridden by `response.setContentType`.
Instead I have introduced a protected `getContentType` method (see https://github.com/xwiki/xwiki-platform/pull/1586/files#diff-4ab936b179b6d22f41cb869da8f1e16018f3c43e9e8d3f3d19e8279e1ca0de5aR199)

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
```diff
@@ -120,4 +115,19 @@ protected ClassLoader getClassLoader(String namespace)
...
+
+    @Override
+    protected String getContentType(InputStream resourceStream,
+        WebJarsResourceReference resourceReference) throws IOException
+    {
+        String mimeType;
+        if (!isResourceCacheable(resourceReference) && getResourceName(resourceReference)
+            .endsWith(LESS_FILE_EXTENSION))
+        {
+            mimeType = "text/css";
```
mflorea, line 127: I would add a comment explaining why we change the content type from LESS to CSS.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/filter/LessWebJarsResourceFilter.java
```diff
@@ -0,0 +1,98 @@
...
+        try {
+            String compile = this.lessCompiler.compile(lessResourceReference, true, false, false);
+            return IOUtils.toInputStream(compile, UTF_8);
+        } catch (LESSCompilerException e) {
+            throw new ResourceReferenceHandlerException(
+                String.format("Error when compiling the resource [%s]", resourceName), e);
+        }
+    }
+
+    private static class WebjarLESSResourceReference implements LESSResourceReference
```
mflorea, line 67: We usually put nested classes at the top. Are you sure there's no Checkstyle violation?

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/filter/VelocityWebJarsResourceFilter.java
```diff
@@ -0,0 +1,70 @@
...
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.resource.ResourceReferenceHandlerException;
+import org.xwiki.velocity.VelocityManager;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Evaluates the velocity content inside the request webjars resource.
```
mflorea, line 38:
```suggestion
 * Evaluates the velocity content inside the requested WebJars resource.
```

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/filter/WebJarsResourceFilter.java
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * Handle the filtering of the request resource, returning an new content instead of the one initially requested.
+     *
+     * @param resourceStream the input stream of the webjar resource initially requested
+     * @param resourceName the name of the webjar resource initially requested
+     * @return an input stream of the filtered resource
+     * @throws ResourceReferenceHandlerException in case of error during the filtering
+     */
+    InputStream handle(InputStream resourceStream, String resourceName) throws ResourceReferenceHandlerException;
```
mflorea, line 45: I would have called it ``filter()``, but since it's an internal component (for now) it's not that important.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/filter/LessWebJarsResourceFilter.java
manuelleduc: No checkstyle violation, and this is not true 100% of the time in XS afaics. But if that helps readability, I'll move it at the top of the class.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/filter/WebJarsResourceFilter.java
manuelleduc: I was actually hesitating between handle and filter. I'll rename it.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/filter/LessWebJarsResourceFilter.java
mflorea: @tmortagne @vmassol @surli do we have a best practice / code style regarding nested classes? I'm fine to say "if Checkstyle doesn't complain then it's OK" :)

tmortagne: I don't remember seeing a nested class not defined at the beginning and my first reflex was also to think that it would fail checkstyle. But I also don't remember ever discussing this.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
manuelleduc: Comment added https://github.com/xwiki/xwiki-platform/pull/1586/files#diff-45ea9c87d5fb68cd5db0da7f78cf25e76f1325f5fe56e21618b21786fc706236R124-R126

mflorea (approved):
Looks good! (except for the location of the nested class but if Checkstyle doesn't complain then I won't complain either :) ).

surli (approved):

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
```diff
@@ -174,15 +175,33 @@ private void serveResource(R resourceReference, InputStream rawResourceStream)
...
+            String resourceName = getResourceName(resourceReference);
             throw new ResourceReferenceHandlerException(String.format("Failed to read resource [%s]", resourceName), e);
         } finally {
             IOUtils.closeQuietly(resourceStream);
         }
     }
 
+    /**
+     * Computes the content type of the resource. By default the content type is inferred by {@link
+     * TikaUtils#detect(InputStream, String)} based on the resource content and name.
```
vmassol, line 190: Maybe the javadoc should mention that this is very costly and if the content type is known, then the RRH should set it,WDYT?

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
```diff
@@ -120,4 +115,22 @@ protected ClassLoader getClassLoader(String namespace)
...
     }
+
+    @Override
+    protected String getContentType(InputStream resourceStream,
+        WebJarsResourceReference resourceReference) throws IOException
+    {
+        String mimeType;
+        // When the resource is not cacheable and the request resource is a less file, we compile it to css.
+        // In this case, the content type must be explicitly set to text/css instead of the default text/x-less
+        // that would otherwise be computed from the resource filename.
```
vmassol, line 126: I guess you mean by Tika because `getContentType()` should guess the right value since it looks at both the name and the content. Or maybe rewrite it the comment as:

> In this case, we set the content type text/css explicitly in order to be safe since otherwise it's possible that the default implementation of getContentType() will infer the type only by looking at the resource file name extension and not check its content (or fail to discover the right content).

?

vmassol: BTW this makes it even more important that the javadoc of getContentType() mentions that it should be overridden when the type is known (see https://github.com/xwiki/xwiki-platform/pull/1586/files#r607793345)?

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/test/java/org/xwiki/webjars/internal/filter/LessWebJarsResourceFilterTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @MockComponent
+    private LESSCompiler lessCompiler;
+
+    @Test
+    void handle() throws Exception
+    {
+        when(this.lessCompiler.compile(any(LESSResourceReference.class), eq(true), eq(false), eq(false)))
+            .thenReturn("some content");
+
+        InputStream stream = this.resourceFilter.filter(mock(InputStream.class), "resourceName.less");
```
vmassol, line 66: Small note that the stream is never closed... it would be more correct to put this in a try-with-resource block. Same for the other tests

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java
mflorea: > this can be very costly

Why?

vmassol: Because reading the stream content to infer the type is costly.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
tmortagne: > getContentType() should guess the right value since it looks at both the name and the content.

Not in this case. Tika guess the content only for types which use a magic marker at the beginning, I don't think it ever tries to guess what's in a text file.

mflorea: @vmassol I don't see any problem with @manuelleduc 's comment. The resource is **transformed** (LESS -> CSS) and the default implementation of ``getContentType()`` returns the content type of the **original** resource (LESS), before the transformation. We need the content type to match the result of the transformation (CSS).

mflorea: > important that the javadoc of getContentType() mentions that it should be overridden when the type is known

No, you need to overwrite ``getContentType()`` only if you transform the resource, i.e. you change it's content type. Which is what happens here. We change the content type form LESS to CSS. The resource file in the WebJar has content type LESS while we produce CSS because we transform the resource.

vmassol: @mflorea The javadoc for `getContentType()` says:

> * Computes the content type of the resource. By default the content type is inferred by {@link TikaUtils#detect(InputStream, String)} based on the resource content and name.

So it's supposed to infer the type not just from the name but also from the resource content according to the javadoc.

vmassol: > No, you need to overwrite getContentType() only if you transform the resource, 

I disagree with this. It's always better to set it if you know it, you'll avoid the cost of Tika trying to guess it.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
```diff
@@ -90,27 +87,25 @@ protected String getResourceName(WebJarsResourceReference resourceReference)
...
-                this.velocityManager.getVelocityEngine().evaluate(this.velocityManager.getVelocityContext(), writer,
-                    resourceName, new InputStreamReader(resourceStream, UTF8));
-                return new ByteArrayInputStream(writer.toString().getBytes(UTF8));
-            } catch (Exception e) {
-                throw new ResourceReferenceHandlerException(
-                    String.format("Failed to evaluate the Velocity code from WebJar resource [%s]", resourceName), e);
+            if (resourceName.endsWith(LESS_FILE_EXTENSION)) {
+                stream = this.lessFilter.filter(resourceStream, resourceName);
+            } else {
+                stream = this.velocityFilter.filter(resourceStream, resourceName);
```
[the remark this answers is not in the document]
mflorea: Indeed, good catch. Should be fixed here https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java#L84 with a try-with-resource.

mflorea (changes requested):
Please close the resource stream after filtering, as pe @vmassol comment.

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
vmassol: > Not in this case. Tika guess the content only for types which use a magic marker at the beginning, I don't think it ever tries to guess what's in a text file.

@tmortagne ok but still, I think my comment is better and takes into account different default implementations (you cannot guarantee that Tika in the future will not check the content for ex. or that we'll stop using Tika and guess it ourselves).

mflorea: > So it's supposed to infer the type not just from the name but also from the resource content according to the javadoc.

And? The content is LESS but we **transform the content**. So the content type determined based on the content (which is LESS) can't be right in this case (we output CSS), which is why we need to overwrite ``getContentType()``.

vmassol: > And? The content is LESS but we transform the content. So the content type determined based on the content (which is LESS) can't be right in this case (we output CSS), which is why we need to overwrite getContentType().

ok, indeed, that's a good reason :) Now we still have this small perf disagreement but it's orthogonal.

mflorea: > It's always better to set it if you know it

You "know it" only if you need to filter / transform the resource (i.e. to handle it in a special way). There's no need to overwrite ``getContentType`` otherwise (we would duplicate Tika's code).

mflorea:
I merged the PR in order to allow @ClemDee progress on the front end ([XWIKI-18270](http://jira.xwiki.org/browse/XWIKI-18270)). @manuelleduc can improve the code afterwards (make sure input streams are properly closed).

xwiki-platform-core/xwiki-platform-webjars/xwiki-platform-webjars-api/src/main/java/org/xwiki/webjars/internal/WebJarsResourceReferenceHandler.java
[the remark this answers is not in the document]
manuelleduc: The resource is closed here https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-servlet/src/main/java/org/xwiki/resource/servlet/AbstractServletResourceReferenceHandler.java#L184 but this is very indirect and it's probably best to convert to a `try-with-resource` statement.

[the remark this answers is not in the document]
manuelleduc: Especially because `closeQuietly` is deprecated

PR #1587, opened 2021-03-31
XWIKI- 15049: NotificationsApplicationsPreferencesMacro stylesheet not using Color Themes variables

DamianArado:
Issue Link: https://jira.xwiki.org/browse/XWIKI-15049

Color themes variables are defined here: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/colorThemeInit.vm

Could not test it, as I am currently having issues in build.
It would be great if a reviewer can check if these changes pass the build.

Also, please let me know if something needs to change.


mflorea (changes requested):
Since the styles are inside a SSX object then best is to use LESS (change the SSX content type) than to use Velocity (parse=1) and the **old** color theme variables.

DamianArado:
> Since the styles are inside a SSX object then best is to use LESS (change the SSX content type) than to use Velocity (parse=1) and the **old** color theme variables.

So, I should not use these color theme variables?

mflorea:
You should use the **LESS** (color theme) variables instead.

PR #1588, opened 2021-04-02
XWIKI-15119: The Applications panel entries are not properly aligned

GwennanS:
Issue: https://jira.xwiki.org/browse/XWIKI-15119

I know it is an old issue but it is still relevant.

image (1)
Old vs new

In the old, the text after an icon is a set amount of pixels after the icon, and this is not aligned with the other titles when the icons are not the same size
In the new, the width property of the icon is a set amount of pixels so the text always starts at the same position.

I set it at 20px since this was also already the height of the box so the icon is in a square now.
Since there is already some space left now as icons are smaller than 20px, the left margin of the following text can be reduced (to .1em).


mflorea:
Would be good to provide some before and after screenshots. Also, you should explain your solution, either in a comment in the code or in the PR description.

GwennanS:
> Would be good to provide some before and after screenshots. Also, you should explain your solution, either in a comment in the code or in the PR description.

Updated!

mflorea (approved):

mflorea:

PR #1590, opened 2021-04-06
Add round borders to preference boxes

RaoulKalisvaart:

This PR rounds borders using a border radius of 4 pixels. This value is chosen because it is used in more places in the project.


mflorea (changes requested):
You need to use LESS and the corresponding ``@border-radius-base``. Either convert the existing SSX object to LESS or add another SSX object to hold the LESS code.

RaoulKalisvaart:
> You need to use LESS and the corresponding `@border-radius-base`. Either convert the existing SSX object to LESS or add another SSX object to hold the LESS code.

I will look into this!

BhargavaCreates (commented):
Changes Look Good

PR #1591, opened 2021-04-06
Display fifth step in Distribution Wizard header

RaoulKalisvaart:
Issue: Distribution Wizard shows 4 steps but then displays a fifth one for the report
Link: https://jira.xwiki.org/browse/XWIKI-17600

This PR makes sure that the distribution wizard header shows five steps instead of four. The fifth step, the report, is currently namely not listed in the wizard.


tmortagne:
@RaoulKalisvaart could you make a screenshot of the result

mflorea:
See the comments on the issue. Before you do the implementation we need to agree on the solution.

PR #1592, opened 2021-04-08
XWIKI-10309: Check URL domains based on a whitelist

surli:
Introduce a new property for listing the trusted domains and API to
check an URL against that list and the aliases used in subwikis.

  * Add new property url.trustedDomains in xwiki.properties
  * Add new API in URLConfiguration to retrieve this configuration value
  * Create a new URLSecurityManager responsible to check if an URL can
    be trusted based on this property and on the subwikis configurations
  * Introduce a new listener to invalidate the cache of
    URLSecurityManager whenever a XWikiServerClass xobject is
added/updated/deleted
  * Introduce a new xwiki-platform-url-default module to avoid cyclic
    dependencies
  * Use the new URLSecurityManager API in XWikiServletResponse to
    perform a check before doing a redirection. Took inspiration on
@sdumitriu commit for it.


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
```diff
@@ -66,9 +68,21 @@ public void sendRedirect(String redirect) throws IOException
             LOGGER.warn("Possible HTTP Response Splitting attack, attempting to redirect to [{}]", redirect);
             return;
         }
+        if (redirect.matches("[a-z0-9]+://.*")) {
```
vmassol, line 71: Shouldn't we display something if we don't redirect? At least some warning log?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
```diff
@@ -66,9 +68,21 @@ public void sendRedirect(String redirect) throws IOException
...
+            // Full URL, check that the hostname matches
+            if (!getURLSecurityManager().isDomainTrusted(new URL(redirect))) {
```
vmassol, line 73: Do we need the "IF" above? What about wrapping in a try/catch for URLMalformedException instead?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
```diff
@@ -66,9 +68,21 @@ public void sendRedirect(String redirect) throws IOException
...
+                LOGGER.warn("Possible phishing attack, attempting to redirect to [{}]", redirect);
```
vmassol, line 74: I think the message should be improved and we should mention it's not in the list of trusted domains and explain ho to add to it.

vmassol: Also the message should say what is done, i.e. that the redirect has not been performed.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLConfiguration.java
```diff
@@ -47,4 +51,15 @@ default boolean useResourceLastModificationDate()
     {
         return true;
     }
+
+    /**
+     * @return the list of trusted domains that can be used in the wiki.
```
vmassol, line 56: I think a definition of what is a trusted domain would be nice since otherwise the javadoc is not really interesting ;)

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+        for (String domain : domains) {
+            if (ACCEPTED_DOMAIN_PATTERN.matcher(domain).matches()) {
+                this.trustedDomains.add(domain);
+            }
+        }
+    }
+
+    @Override
+    public boolean isDomainTrusted(URL urlToCheck)
```
tmortagne, line 87: I also feel that `isDomainTrusted` should take a `String domain` and not a whole URL and the user of the API should decide how to get the domain to check (it might not always be a URL).

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @param urlToCheck the URL for which we want to know if the domain is trusted or not.
+     * @return {@code true} if the URL domain can be trusted, {@code false} otherwise
+     */
+    boolean isDomainTrusted(URL urlToCheck);
+
+    /**
+     * For performance reasons the URLSecurityManager might use a cache, this method allows to invalidate it.
+     */
+    void invalidateCache();
```
vmassol, line 51: That doesn't seem great to me as I don't find this a valid business API.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,110 @@
...
+    {
+        if (this.trustedDomains == null) {
+            this.computeTrustedDomains();
+        }
+
+        String host = urlToCheck.getHost();
+
+        while (host.contains(DOT)) {
```
tmortagne, line 95: `StringUtils.contains(host, '.')` should be a bit better for performances.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.platform</groupId>
+    <artifactId>xwiki-platform-url</artifactId>
+    <version>13.3-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-platform-url-default</artifactId>
+  <name>XWiki Platform - URL - Default</name>
+  <packaging>jar</packaging>
+  <description>Allows configuration of the URL scheme used by XWiki to parse/serialize URLs</description>
```
vmassol, line 33: desc needs to be updated

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  <properties>
+    <xwiki.jacoco.instructionRatio>0.66</xwiki.jacoco.instructionRatio>
+    <!-- Name to display by the Extension Manager -->
+    <xwiki.extension.name>URL Default</xwiki.extension.name>
```
vmassol, line 37: This name needs improving, actually we dont need an extension name probably since we don't import impl modules on e.x.o.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,110 @@
...
+    @Inject
+    private URLConfiguration urlConfiguration;
+
+    @Inject
+    private WikiDescriptorManager wikiDescriptorManager;
+
+    @Inject
+    private Logger logger;
+
+    private Set<String> trustedDomains;
```
tmortagne, line 63: Should it be a `Cache<Boolean>` ? Right now you are only caching the fact that a domain is trusted but caching that a domain is not trusted would be interesting for performances too.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,110 @@
...
+    {
+        Set<String> domains;
+        domains = new HashSet<>(this.urlConfiguration.getTrustedDomains());
+
+        try {
+            for (WikiDescriptor wikiDescriptor : wikiDescriptorManager.getAll()) {
+                domains.addAll(wikiDescriptor.getAliases());
+            }
+        } catch (WikiManagerException e) {
+            logger.warn("Error while getting wiki descriptor to fill list of trusted domains.", e);
```
vmassol, line 75: warn should not display a stack trace. Also, as always, the message needs to explain what is the action done since it's not an error and we fallback to something.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        this.trustedDomains = new HashSet<>();
+
+        for (String domain : domains) {
+            if (ACCEPTED_DOMAIN_PATTERN.matcher(domain).matches()) {
+                this.trustedDomains.add(domain);
+            }
```
vmassol, line 82: what if it doesn't match? Nothing is displayed in logs or anywhere, that doesn't seem good.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    @Override
+    public boolean isDomainTrusted(URL urlToCheck)
+    {
+        if (this.trustedDomains == null) {
+            this.computeTrustedDomains();
```
vmassol, line 90: small detail: extra `this.`

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/XWikiServerClassListener.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+@Component
+@Singleton
+@Named(XWikiServerClassListener.NAME)
+public class XWikiServerClassListener extends AbstractEventListener
+{
+    /**
+     * Name of the listener.
+     */
+    public static final String NAME = "org.xwiki.url.internal.XWikiServerClassListener";
```
vmassol, line 55: Interesting rule that I hadn't noticed. Maybe we should define it as a best practice and generalize it for all listeners (I remember writing listeners with ids and not FQN of internal packages in the past).

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/XWikiServerClassListener.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private static final LocalDocumentReference XWIKISERVER_CLASS =
+        new LocalDocumentReference("XWiki", "XWikiServerClass");
```
vmassol, line 58: Small detail Ideally this would need to come from elsewhere so that if we change the class location or name we need to update less places.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/test/java/org/xwiki/url/internal/DefaultURLSecurityManagerTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+        WikiDescriptor wikiDescriptor2 = mock(WikiDescriptor.class);
+        when(wikiDescriptor2.getAliases()).thenReturn(Collections.singletonList(
+            "enterprise.eu"
+        ));
+
+        when(this.wikiDescriptorManager.getAll()).thenReturn(Arrays.asList(wikiDescriptor1, wikiDescriptor2));
+
+        assertTrue(this.urlSecurityManager
+            .isDomainTrusted(new URL("http://www.xwiki.org/xwiki/bin/view/XWiki/Login")));
```
vmassol, line 80: We should never use `assertTrue` without a message. Instead use `assertThat` so that junit displays the diff between expected and actual.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/pom.xml
```diff
@@ -335,6 +335,13 @@
       <version>${project.version}</version>
       <scope>runtime</scope>
     </dependency>
+    <!-- URL Default Module -->
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-url-default</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
```
vmassol, line 344: Maybe check the minimal distribution. It has:

```
    <!-- URL -->
    <dependency>
      <groupId>org.xwiki.platform</groupId>
      <artifactId>xwiki-platform-url-scheme-standard</artifactId>
      <scope>runtime</scope>
    </dependency>
```

I think `xwiki-platform-url-scheme-standard` should now have a dep on the url impl instead of the url api.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/PackageMojo.java
```diff
@@ -750,6 +750,10 @@ private String getContextPath(Artifact warArtifact)
         mandatoryTopLevelArtifacts.add(this.repositorySystem.createArtifact("org.xwiki.platform",
             "xwiki-platform-user-script", getXWikiPlatformVersion(), null, "jar"));
 
+        // Default URL API implementation
+        mandatoryTopLevelArtifacts.add(this.repositorySystem.createArtifact("org.xwiki.platform",
+            "xwiki-platform-url-default", getXWikiPlatformVersion(), null, "jar"));
+
```
vmassol, line 756: why is this needed since we were supposed to already have a dep for URLs?

vmassol:
General comment: you introduced a default impl but kept the impl in the url api module. That feels a bit weird to me....

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
tmortagne: > we should mention it's not in the list of trusted domains

+1

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
tmortagne: `sendRedirect` is very often used with relative URLs (most of the time actually) so relying on exception is not great for performances.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
tmortagne, line 71: You should have a pre compiled `Pattern` as constant field for performances.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/XWikiServerClassListener.java
tmortagne: Maybe the best is to introduce WikiDescriptor/Added/Deleted/UpdatedEvent events since that what we are actually checking here (and it's not the only place where we could need that).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
```diff
@@ -31,6 +32,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.xwiki.url.URLSecurityManager;
 
 public class XWikiServletResponse implements XWikiResponse
```
tmortagne, line 37: Doing it at this level is quite a breaking change since XWikiServletResponse is not only used by xredirect but also by perfectly valid cases: for example this is going to break the OpenID Connect Authenticator.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
surli: Not sure what you suggest to handle it. Note that we do have several APIs with `clearCache` or `invalidateCache`.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: I'm not a big fan of this. First I don't really the usecases where you'd prefer to give a String instead of a URL to be checked. Second, this API will be in a URL module, so it feels more accurate to take a URL.
IMO allowing to check any String could lead to wrong assumptions like putting a URL serialized as string for the check.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
vmassol: Then they also need to be fixed ;) It's clearly an impl detail.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
surli: Not sure to understand what you suggest here: we do output a warning log if we don't redirect. Here this if is just to know if we need to check the domain.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: Thanks for the hint I'll check that.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/pom.xml
surli: Ok I'll remove it.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: Indeed I should handle an else part here.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: What's the issue with it?

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/XWikiServerClassListener.java
surli: > Maybe the best is to introduce WikiDescriptor/Added/Deleted/UpdatedEvent events since that's what we are actually checking here (and it's not the only place where we could need that).

Could be interesting indeed.

surli:
> General comment: you introduced a default impl but kept the impl in the url api module. That feels a bit weird to me....

I didn't want to pollute the PR with lots of file move, especially since I wasn't sure we all agree on the locations. If we do agree I'll move the impl in the right module.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
surli: That's interesting, so you suggest it should be moved to the handling of xredirect parameters only? AFAIR there was no single point for that, I guess it would need a proper refactoring to be handled at a single place.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
surli: Indeed.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/PackageMojo.java
surli: You mean `xwiki-platform-url-scheme-standard` that you mentioned on the other comment?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
vmassol: ok I was just checking the diff and saw no ELSE. I've checked the file now and it seems that if it doesn't match we redirect without a check. So my understanding now (BTW a comment would be nice just before the match() call to explain what it's doing and why it's there) is that we check only when there's a URL scheme and the reason is probably because we don't want to perform the check for relative URLs since they're on the same domain (BTW I had to check `sendRedirect` again and indeed it supports relative URLs: `This method can accept relative URLs;the servlet container must convert the relative URL to an absolute URL before sending the response to the client. If the location is relative without a leading '/' the container interprets it as relative to the current request URI. If the location is relative with a leading '/' the container interprets it as relative to the servlet container root.`).

So it looks good, except the missing comment. 

Thanks

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
vmassol: We don't use `this.` for methods in our code style. Only for variables.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/PackageMojo.java
vmassol: yes

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/pom.xml
surli: > I think xwiki-platform-url-scheme-standard should now have a dep on the url impl instead of the url api.

Not sure, if the purpose is just to avoid adding a dep in the minimal distrib. In terms of archi, it feels better to have a url API used by both url default and url-scheme-standard: the purpose of both modules is specific.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
tmortagne: Yes supporting this only for xredirect is not an easy fix and won't work for any extension which uses xredirect. A possibility could be to compare the `redirect` value with the `xredirect` value from the context request, but that's hackish...

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/XWikiServerClassListener.java
surli: I created https://jira.xwiki.org/browse/XWIKI-18552 for it.

tmortagne: Note that in the meantime you could use XWikiServerClassDocumentInitializer#SERVER_CLASS unless you don't have a dependency on old core, did not checked.

surli: Indeed I missed it.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLConfiguration.java
```diff
@@ -60,4 +63,16 @@ public boolean useResourceLastModificationDate()
...
     }
+
+    @Override
+    public List<String> getTrustedDomains()
+    {
+        return this.configuration.get().getProperty(PREFIX + "trustedDomains", Collections.emptyList());
+    }
+
+    @Override
+    public boolean skipTrustedDomainsChecks()
```
tmortagne, line 74: The name of the method sounds weird, I would expect it to skip the trusted domain check and not indicate if it's skipped. What about something like isTrustedDomainsCheckSkipped (not really sure what "checks" was plural too) ?

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -0,0 +1,137 @@
...
+
+        while (StringUtils.contains(host, DOT)) {
+            if (trustedDomains.contains(host)) {
+                return true;
+            } else {
+                host = host.substring(host.indexOf(DOT) + 1);
+            }
+        }
+
+        String bypassCheckProperty = (String) execution.getContext()
```
tmortagne, line 117: I don't understand. Should this context property simply be a Boolean instead of parsing a String ?

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLConfiguration.java
vmassol: yeah I don't like the name too much too. I'd make it shorter and remove "check" which doesn't mean much IMO, and instead talk about activating/enabling the feature or not, with something like `enableTrustedDomains` or `isTrustedDomainsEnabled`. WDYT?

surli: Ok to go with `isTrustedDomainsEnabled` but then the default would be true.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: Yeah don't know why I thought the values should be all strings....

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/XWikiServerClassListener.java
surli: Actually to use it I'd need to add a dependency to `xwiki-platform-wiki-default` and I don't really want this dependency just for that. This would be probably better if this constant would be a `xwiki-platform-wiki-api`.

tmortagne: > I don't really want this dependency just for that

Definitely, I just though it was in oldcore for some reason.

> This would be probably better if this constant would be a xwiki-platform-wiki-api.

I don't know about that, I prefer not making this public too much and instead provide the missing APIs (like the events in your case which would pass the new WikiDescriptor).

PR #1593, opened 2021-04-08
XWIKI-18510: A right object with like level will not be displayed in page administration

arcilli:
As it is exposed in https://jira.xwiki.org/browse/XWIKI-18510, there is an inconsistency of naming the Like right with the rest of the rights from the platform (those are with lowercase).

The changing of the name with lowercase will not impact the checking of access (since for those cases we're working with the object itself).

Note that the current PR is only a draft, it was not extensively tested, even if the unit testing are passing.


surli:
The fix looks good but I don't think it's enough: we probably need a migration to ensure all the `Like` rights record are converted to `like`.

PR #1595, opened 2021-04-13
XWIKI-18491: Add a localization REST resource to expose translation messages

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18491

- Introduces xwiki-platform-localization-rest
  - Provides a REST endpoint to retrieve the raw translation values of the requested translation keys
- Introduces xwiki-platform-localization-test-docker
  - Adds functional tests for the localization REST endpoint
- Declares xwiki-platform-localization-rest in the runtime dependencies of xwiki-platform-distribution-war-dependencies


xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,117 @@
...
+            // Set the requested wiki id.
+            xWikiContext.setWikiId(wikiId);
+
+            ComponentManager componentManager = this.componentManagerProvider.get();
+            LocalizationContext localizationContext = componentManager.getInstance(LocalizationContext.class);
+            LocalizationManager localizationManager = componentManager.getInstance(LocalizationManager.class);
+
+            // Resolve the local.
+            Locale local;
+            if (StringUtils.isNotEmpty(locale)) {
```
tmortagne, line 85: Empty string is a valid locale, so this `if` should rely on `null` instead IMO.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,117 @@
...
+                local = Locale.forLanguageTag(locale);
```
tmortagne, line 86: You should use `org.xwiki.localization.LocaleUtils#toLocale` instead.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,117 @@
...
+        XWikiContext xWikiContext = this.contextProvider.get();
+
+        // Save the current context wiki id. It will be restored on the method is finished.
+        String oldWikiId = xWikiContext.getWikiId();
+        try {
+            // Set the requested wiki id.
+            xWikiContext.setWikiId(wikiId);
+
+            ComponentManager componentManager = this.componentManagerProvider.get();
+            LocalizationContext localizationContext = componentManager.getInstance(LocalizationContext.class);
```
tmortagne, line 80: Why don't you inject those components ?

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,117 @@
...
+            LocalizationManager localizationManager = componentManager.getInstance(LocalizationManager.class);
+
+            // Resolve the local.
+            Locale local;
```
tmortagne, line 84: `local` is a very misleading variable name :) You could name the method parameter `localString` instead for example.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,117 @@
...
+
+            // Resolve the raw translation of the requested keys.
+            for (String key : keys) {
+                String fullKey = cleanPrefix + key;
+                Translation translation = localizationManager.getTranslation(fullKey, local);
+                if (translation != null) {
+                    Object rawSource = translation.getRawSource();
+                    result.put(fullKey, String.valueOf(rawSource));
+                } else {
+                    result.putNull(fullKey);
```
tmortagne, line 104: Maybe add a comment to explain why it's in the result at all.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: Good catch, thanks.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: Because they need to be loaded in the right context. But their is maybe a simpler way to do that?

tmortagne: No they don't :)

They need to be **used** in the right context. Whatever context you are calling getInstance from you will always get `com.xpn.xwiki.internal.localization.XWikiLocalizationContext`.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-118a400f3886da870ebde91eda0d6269f44a19975059f7d189964ac02c48d623R96

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-118a400f3886da870ebde91eda0d6269f44a19975059f7d189964ac02c48d623R82

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-118a400f3886da870ebde91eda0d6269f44a19975059f7d189964ac02c48d623R56-R60

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-118a400f3886da870ebde91eda0d6269f44a19975059f7d189964ac02c48d623R80

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-118a400f3886da870ebde91eda0d6269f44a19975059f7d189964ac02c48d623R100-R101

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/resources/META-INF/components.txt
```diff
@@ -0,0 +1 @@
+org.xwiki.localization.internal.DefaultLocalizationSource
```
surli, line 1: I'm surprised that Spoon is not complaining in theory you need a blank line at the end of components.txt.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,113 @@
...
+            // Resolve the local.
+            Locale locale;
+            if (localeString != null) {
+                locale = LocaleUtils.toLocale(localeString);
+            } else {
+                locale = this.localizationContext.getCurrentLocale();
+            }
+
+            // Computes the prefix.
+            String cleanPrefix = Objects.toString(prefix, "");
```
surli, line 88: You're doing that because prefix is nullable? It's not indicated in javadoc AFAICS.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/resources/META-INF/components.txt
manuelleduc: I agree Spoon must complain if a blank line is missing at the end of `components.txt`.
This is not the case here, though I agree github's UI is not really explicit about it.
I think a missing trailing blank line is represent by a small icon at the end of the document.
Here for instance: https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-57ce2b01a0829f38517cf869d7b3bfa5260c57a1c2d2f0f52a0c7e3f6ce0af35R104

surli: Ah you're right, all good then.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: Exactly, and all jaxrs query params are nullable so I didn't repeat it in the javadoc, but explicit is always better, I'll update the javadoc.

surli: > all jaxrs query params are nullable

so you're missing a check for keys below no?

surli: (and wikiId?)

manuelleduc: wikiId is mandatory since it's a path param. I have a doubt regarding `keys`. when no `key` parameter is provided, it is initialized with and list containing a single null element, but indeed in this case the response is `{"null": null}` which is not what we want, fixing.

manuelleduc: Missing key parameter fix https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-118a400f3886da870ebde91eda0d6269f44a19975059f7d189964ac02c48d623R94
Improved javadoc https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-d3378dd806e951aaca607e611516c5e5bc45d33347bba4e1eaf0a8909578c803R42-R57

surli (approved):

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,115 @@
...
+
+            ObjectNode result = JsonNodeFactory.instance.objectNode();
```
tmortagne, line 90: The best practice is actually to define a strict types JAX-RS model and return this model instead. The model is then serialized  in XML or Json depending on what the HTTP client asked for. You can find an example of a non xwiki-platform-rest-server module which define a model in xwiki-platform-annotation-rest.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>jsr311-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
```
mflorea, line 49: You shouldn't need this.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-localization-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-oldcore</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
mflorea, line 69: You shouldn't need this either.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Exposes the wiki translations through REST.
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Path("/wikis/{wikiName}/localization/translations")
+@Unstable
+public interface LocalizationSource
```
mflorea, line 40: Doesn't match the URL and we use "**Re**source" everywhere else. So it should be ``TranslationsResource``. The ``LocalizationResource`` will be used for ``/wikis/{wikiName}/localization`` (i.e. is the parent resource). @tmortagne suggested to have two separate resources for ``localization`` and ``translations`` because localization is much more than translations and we wanted to keep it open to add more later. For the scope of this PR you only need to implement the translations resource.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param wikiName the name of the wiki holding the translation keys
+     * @param locale the locale of the translation values, when {@code null} the current locale of the user is used
+     * @param prefix a common prefix, concatenated to each translation key before resolving their values (can be
+     *     {@code null} in which case nothing is concatenated)
+     * @param keys the translation keys to resolve. If no key is passed, and empty object is returned
+     * @return the response, an {@link Response.Status#INTERNAL_SERVER_ERROR} in case of component resolution issue, or
+     *     a {@link Response.Status#OK} response with an object holding the translation values otherwise
+     */
+    @GET
+    Response translations(@PathParam("wikiName") String wikiName, @QueryParam("locale") String locale,
```
mflorea, line 59: It should return a ``org.xwiki.rest.model.jaxb.Map`` or some type dedicated to translations.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Path("/wikis/{wikiName}/localization/translations")
+@Unstable
+public interface LocalizationSource
+{
+    /**
+     * Returns the raw translation values of the request keys. In other words, the translations values without the
```
mflorea, line 43:
```suggestion
     * Returns the raw translation values of the requested keys. In other words, the translations values without the
```

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,115 @@
...
+    @Inject
+    private Logger logger;
+
+    @Override
+    public Response translations(String wikiId, String localeString, String prefix, List<String> keys)
+    {
+        XWikiContext xWikiContext = this.contextProvider.get();
+
+        // Save the current context wiki id. It will be restored on the method is finished.
+        String oldWikiId = xWikiContext.getWikiId();
```
mflorea, line 74: You should use the ``ModelContext`` so that you don't depend on old core.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,115 @@
...
+        try {
+            // Set the requested wiki id.
+            xWikiContext.setWikiId(wikiId);
+
+            // Resolve the local.
```
mflorea, line 79:
```suggestion
            // Resolve the locale.
```

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
```diff
@@ -0,0 +1,115 @@
...
+                    String fullKey = cleanPrefix + key;
+                    Translation translation = this.localizationManager.getTranslation(fullKey, locale);
+                    if (translation != null) {
+                        Object rawSource = translation.getRawSource();
+                        result.put(fullKey, String.valueOf(rawSource));
+                    } else {
+                        // When a translation key is not found, it is still added to the result object, associated with
+                        // a null value.
+                        result.putNull(fullKey);
+                        this.logger.warn("Key [{}] not found for local [{}] in wiki [{}].", key, locale, wikiId);
```
mflorea, line 104: I would rather fall-back on the English translation. Actually, would be good to follow the full chain: locale, parent locale, .., root locale. Also, logging a warning for this feels to much because having translation keys with missing values for a specific locale is very common. @tmortagne WDYT?

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
tmortagne: > would be good to follow the full chain: locale, parent locale, .., root locale

That's already what `LocalizationManager.getTranslation`. If you get null it means this tranlation simply does not exist at all (in this context at least).

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/pom.xml
tmortagne: It comes with xwiki-platform-rest-server anyway but yes you don't seem to be using it directly in your code.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
tmortagne: Any XWikiRestComponent will depend on oldcore anyway right now.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
mflorea: OK, good to know. I would still use debug level and not warning though.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
tmortagne: Would be nicer for reusability to have this API (and the model) in an API module which does not force you to depend on oldcore (which is the case of any implementation of XWikiRestComponent right now).

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
mflorea: Yes, but at least it will allow us to remove the direct dependency to old core (even if it comes as a transitive dependency).

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
manuelleduc: Would it be ok to move this class to `xwiki-platform-localization-api` in this case?

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
manuelleduc: I'm currently going with a dedicated Translation type. I'll push the update version soon.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
mflorea: I'm not sure, it would force a dependency on ``javax.ws.rs`` for ``localization-api``.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
tmortagne: > I would still use debug level and not warning though.

Not sure, having a UI which request a non-existing translation seems like a mistake to me. Do you have a use case in mind ?

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
tmortagne: Yeah I'm not a fan either of the extra dependencies to add to xwiki-platform-localization-api.

manuelleduc: Or should I two sub modules `xwiki-platform-localization-rest-api` and `xwiki-platform-localization-rest-default`?

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
mflorea: Could happen with a cached (outdated) UI (JavaScript) and deleted translation keys that didn't went through deprecation. But indeed it would mean something is wrong (i.e. the front-end is not in sync with the back-end) so the warning could help the user raise and fix the problem.

I just don't want to see this warning when a translation is missing for a specific locale but it's available in English, but that's not the case AFAIU, so I'm fine with warning.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
manuelleduc: @mflorea let me know if the new jaxb model is ok for you :)

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/internal/DefaultLocalizationSource.java
manuelleduc: I stayed with the warn for now.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/src/main/java/org/xwiki/localization/rest/LocalizationSource.java
manuelleduc: I'm working on this part.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-api/src/main/resources/xwiki.rest.localization.model.xsd
```diff
@@ -0,0 +1,64 @@
...
+  </complexType>
+
+  <!--
+  *
+  * Elements
+  *
+  -->
+
+  <element name="translations" type="localization:Translations"/>
+</schema>
```
manuelleduc, line 64: Could also be moved to a dedicated `-model` module. WDYT?

tmortagne: xwiki-platform-localization-rest-api is already a pretty dedicated module IMO.

xwiki-platform-core/xwiki-platform-localization/pom.xml
```diff
@@ -33,10 +33,20 @@
   <description>XWiki Platform - Localization - Parent POM</description>
   <modules>
     <module>xwiki-platform-localization-api</module>
+    <module>xwiki-platform-localization-rest</module>
```
tmortagne, line 36: This needs to be updated, right ?

manuelleduc: No because `xwiki-platform-localization-rest` contains `xwiki-platform-localization-rest-api` and `xwiki-platform-localization-rest-default`.
But https://github.com/xwiki/xwiki-platform/pull/1595/files#diff-e0609ee1c594bb59216372bd806f653443f22af8e1c91d94fe3b28fd30cda1b5R436 must be updated though

manuelleduc: done

tmortagne: Indeed sorry, I did not notice that you put them one level below.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-api/src/main/java/org/xwiki/localization/rest/TranslationResource.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param wikiName the name of the wiki holding the translation keys
+     * @param locale the locale of the translation values, when {@code null} the current locale of the user is used
+     * @param prefix a common prefix, concatenated to each translation key before resolving their values (can be
+     *     {@code null} in which case nothing is concatenated)
+     * @param keys the translation keys to resolve. If no key is passed, and empty object is returned
+     * @return the list of resolved translations
+     */
+    @GET
+    Translations translations(@PathParam("wikiName") String wikiName, @QueryParam("locale") String locale,
```
mflorea, line 59: We use ``get`` prefix for GET requests everywhere in ``xwiki-platform-rest-api``. So it should be ``getTranslations``.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-default/src/main/java/org/xwiki/localization/internal/DefaultTranslationResource.java
```diff
@@ -0,0 +1,122 @@
...
+
+            // Computes the prefix.
+            String cleanPrefix = Objects.toString(prefix, "");
+
+            ObjectFactory objectFactory = new ObjectFactory();
+            Translations result = objectFactory.createTranslations();
+
+            // Resolve the raw translation of the requested keys.
+            for (String key : keys) {
+                if (key != null) {
```
mflorea, line 92: I would add a comment about https://github.com/restlet/restlet-framework-java/issues/922 .

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-default/src/main/java/org/xwiki/localization/internal/DefaultTranslationResource.java
```diff
@@ -0,0 +1,122 @@
...
+                    String fullKey = cleanPrefix + key;
+                    Translation translation = this.localizationManager.getTranslation(fullKey, locale);
+                    if (translation != null) {
+                        String rawSource = (String) translation.getRawSource();
+                        result.getTranslations().add(createTranslation(objectFactory, fullKey, rawSource));
+                    } else {
+                        // When a translation key is not found, it is still added to the result object, associated with
+                        // a null value.
+                        result.getTranslations().add(createTranslation(objectFactory, fullKey, null));
+                        this.logger.warn("Key [{}] not found for local [{}] in wiki [{}].", key, locale, wikiId);
```
mflorea, line 102:
```suggestion
                        this.logger.warn("Translation key [{}] not found for locale [{}] in wiki [{}].", key, locale, wikiId);
```

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-default/src/test/java/org/xwiki/localization/internal/DefaultTranslationResourceTest.java
```diff
@@ -0,0 +1,242 @@
...
+        when(this.localizationContext.getCurrentLocale()).thenReturn(this.defaultLocale);
+        when(this.modelContext.getCurrentEntityReference()).thenReturn(CURRENT_ENTITY);
+    }
+
+    @Test
+    void translations() throws Exception
+    {
+        Translation translationKey1 = mock(Translation.class);
+        Translation translationKey2 = mock(Translation.class);
+        ObjectFactory objectFactory = new ObjectFactory();
```
mflorea, line 94: This could be moved as a field since it is used in multiple tests.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-api/src/main/java/org/xwiki/localization/rest/TranslationResource.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Exposes the wiki translations through REST.
+ *
+ * @version $Id$
+ * @since 13.3RC1
+ */
+@Path("/wikis/{wikiName}/localization/translations")
+@Unstable
+public interface TranslationResource
```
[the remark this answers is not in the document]
manuelleduc: my bad, fixed

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-default/src/test/java/org/xwiki/localization/internal/DefaultTranslationsResourceTest.java
```diff
@@ -0,0 +1,238 @@
...
+    {
+        when(this.localizationContext.getCurrentLocale()).thenReturn(this.defaultLocale);
+        when(this.modelContext.getCurrentEntityReference()).thenReturn(CURRENT_ENTITY);
+    }
+
+    @Test
+    void getTranslations() throws Exception
+    {
+        ObjectFactory objectFactory = new ObjectFactory();
+        Translations expected = objectFactory.createTranslations();
```
mflorea, line 99: These are used in many of the tests below so I would move them as fields in the test class.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-test/xwiki-platform-localization-test-docker/src/test/it/org/xwiki/localization/test/ui/TranslationsRestIT.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Test
+    @Order(1)
+    void translationsNoKeyParam(TestUtils testUtils) throws Exception
+    {
+        Map<String, Object[]> queryParams = new HashMap<>();
+        GetMethod getMethod = testUtils.rest().executeGet(TranslationsResource.class, queryParams, "xwiki");
+        String body = getMethod.getResponseBodyAsString();
+        assertEquals("<Translations><translation/></Translations>", body);
```
mflorea, line 58: I don't understand where the **upper case** ``Translations`` comes from but I doesn't look right to me. In your schema you have:

    <element name="translations" type="localization:Translations"/>

as root element, so the name should be "translations" not "Translations". For instance the REST URL ``rest/wikis/xwiki/spaces/Sandbox/pages/WebHome/attachments`` outputs ``<attachments><attachment>...`` and we have:

    <element name="attachments" type="xwiki:Attachments"></element>

in the [schema](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/resources/xwiki.rest.model.xsd#L435)

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-default/src/main/java/org/xwiki/localization/internal/DefaultTranslationsResource.java
```diff
@@ -0,0 +1,125 @@
...
+            // Resolve the raw translation of the requested keys.
+            for (String key : keys) {
+                // When no key parameter is passed to the request, the list of keys is initialized with a single null 
+                // value. See https://github.com/restlet/restlet-framework-java/issues/922
+                if (key != null) {
+                    String fullKey = cleanPrefix + key;
+                    Translation translation = this.localizationManager.getTranslation(fullKey, locale);
+                    if (translation != null) {
+                        String rawSource = (String) translation.getRawSource();
+                        result.getTranslation().add(createTranslation(objectFactory, fullKey, rawSource));
```
mflorea, line 99: This is not right. It should have generated ``getTranslations()`` (plural). So it looks like the schema is still not right.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-test/xwiki-platform-localization-test-docker/src/test/it/org/xwiki/localization/test/ui/TranslationsRestIT.java
```diff
@@ -0,0 +1,100 @@
...
+    @Order(2)
+    void translationsKeyMissing(TestUtils testUtils) throws Exception
+    {
+        Map<String, Object[]> queryParams = new HashMap<>();
+        queryParams.put("key", new Object[] { "key1" });
+        GetMethod getMethod = testUtils.rest().executeGet(TranslationsResource.class, queryParams, "xwiki");
+        String body = getMethod.getResponseBodyAsString();
+        assertEquals(
+            "<Translations>"
+                + "<translation><translation><key>key1</key><rawSource/></translation></translation>"
```
mflorea, line 71: ``<translation><translation>`` is definitely not right.

mflorea (changes requested):
There's something wrong with the schema but I can't see why. It seems to follow the pattern from [Attachment](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/resources/xwiki.rest.model.xsd#L169) and [Attachments](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-model/src/main/resources/xwiki.rest.model.xsd#L424) but the result is different than what we get on ``xwiki-platform-rest-model`` (both on the generated Java classes and on the output XML).

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-test/xwiki-platform-localization-test-docker/src/test/it/org/xwiki/localization/test/ui/TranslationsRestIT.java
manuelleduc: Mystery solved, activating `xjc:simple` (https://javaee.github.io/jaxb-v2/doc/user-guide/ch05.html#simple) allow to obtains the desired serialization and java code generation.

xwiki-platform-core/xwiki-platform-localization/xwiki-platform-localization-rest/xwiki-platform-localization-rest-default/src/test/java/org/xwiki/localization/internal/DefaultTranslationsResourceTest.java
manuelleduc: I'm not sure it's simplifying the tests, especially regarding readability and changeability.

mflorea: Let's keep it like that then.

mflorea (approved):

PR #1598, opened 2021-04-16
XWIKI-18542: AWM tests failing on chrome

manuelleduc:
Moving to the top of the screen is required for the drag and drop action to work, but using sendKey(HOME) wasn't working on Chrome (the drag and drop action was performed while the scroll up is still moving). sendKey replaced with executeScript("window.scrollTo(0, 0)")

### DIFF
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
@@ -102,9 +102,10 @@ public class ApplicationClassEditPage extends ApplicationEditPage
     {
         String fieldXPath = "//li[@class = 'field' and normalize-space(.) = '%s']";
         WebElement field = palette.findElement(By.xpath(String.format(fieldXPath, fieldType)));
-        // NOTE: We scroll the page up because the drag&drop fails sometimes if the dragged field and the canvas (drop
-        // target) are not fully visible. See https://code.google.com/p/selenium/issues/detail?id=3075 .
-        getDriver().executeScript("window.scrollTo(0, 0)");
+        // NOTE: We scroll to the top of the page (coordinates (0, 0)) because the drag&drop fails sometimes if the 
+        // dragged field and the canvas (drop target) are not fully visible. See
+        // https://code.google.com/p/selenium/issues/detail?id=3075 .
+        getDriver().scrollTo(0, 0);
         getDriver().dragAndDrop(field, fieldsCanvas);
         final WebElement addedField = fieldsCanvas.findElement(By.xpath("./ul[@id='fields']/li[last()]"));
 
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
@@ -20,7 +20,6 @@
 package org.xwiki.appwithinminutes.test.po;
 
 import org.openqa.selenium.By;
-import org.openqa.selenium.Keys;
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
```diff
@@ -97,14 +96,15 @@ public boolean hasPreviousStep()
...
+     * @return the newly added field page object
      */
     public ClassFieldEditPane addField(String fieldType)
     {
         String fieldXPath = "//li[@class = 'field' and normalize-space(.) = '%s']";
         WebElement field = palette.findElement(By.xpath(String.format(fieldXPath, fieldType)));
         // NOTE: We scroll the page up because the drag&drop fails sometimes if the dragged field and the canvas (drop
         // target) are not fully visible. See https://code.google.com/p/selenium/issues/detail?id=3075 .
-        getDriver().createActions().sendKeys(Keys.HOME).perform();
+        getDriver().executeScript("window.scrollTo(0, 0)");
```
manuelleduc, line 107: Might be interesting to move this to a `moveTop` method in `BaseElement`.
WDYT?

surli: Well actually at the very least you should use `getDriver().scrollTo()` method here. Not sure if we need it in BaseElement now: that's a pretty low-level API so it makes sense if it reminds in XWikiWebDriver IMO.

manuelleduc: `XWikiWebDriver#scrollTo` takes a WebElement, not coordinates.
But we could introduce a `XWikiWebDriver#scrollTo(int xCoord, int yCoord)` method.

manuelleduc: Moving to `(0, 0)` is low level. Moving to the top of a page is more business related imo.

surli: > XWikiWebDriver#scrollTo takes a WebElement, not coordinates.

Right, read too fast. Would be a `scrollToTop` method then. +1 to add that: the test would still be fragile but it was already and at least it's easier to spot where this hack is used if we ever find a way to improve it (which I doubt since the drag&drop can be difficult even manually sometimes...).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiWebDriver.java
```diff
@@ -649,6 +649,24 @@ public WebElement scrollTo(WebElement element)
...
+     * @param xCoord is the pixel along the horizontal axis of the web page that you want displayed in the upper
+     *     left
+     * @param yCoord is the pixel along the vertical axis of the web page that you want displayed in the upper left
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo">MDN Web Docs -
+     *     Element.scrollTo()</a>
+     * @see #scrollTo(WebElement)
+     * @since 13.3RC1
+     * @since 12.10.7
+     */
+    public void scrollTo(int xCoord, int yCoord)
```
surli, line 665: Hmm my last comment was proposing to add a `scrollToTop` but if you think this one add more value I'm ok to add it.

manuelleduc: No no, I agree with you actually, just a misunderstanding.
I'll add it too but that does not prevent to keep this one too.

manuelleduc: `scrollToTop` added https://github.com/xwiki/xwiki-platform/pull/1598/files#diff-3f695424dbbf520ad4644d8d288aafd9ee5d18652641363395e48112b3b1e55eR258

PR #1599, opened 2021-04-16
XWIKI-18542: AWM tests failing on chrome (#1598)

manuelleduc:
Moving to the top of the screen is required for the drag and drop action to work, but using sendKey(HOME) wasn't working on Chrome (the drag and drop action was performed while the scroll up is still moving). sendKey replaced with executeScript("window.scrollTo(0, 0)")
- scrollTo(int,int) moved to a dedicated method in XWikiWebDriver.
- Introducing a `scrollToTop` method in `ViewPage`.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
@@ -20,7 +20,6 @@
 package org.xwiki.appwithinminutes.test.po;
 
 import org.openqa.selenium.By;
-import org.openqa.selenium.Keys;
 import org.openqa.selenium.NotFoundException;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-pageobjects/src/main/java/org/xwiki/appwithinminutes/test/po/ApplicationClassEditPage.java
```diff
@@ -97,14 +96,15 @@ public boolean hasPreviousStep()
...
     public ClassFieldEditPane addField(String fieldType)
     {
         String fieldXPath = "//li[@class = 'field' and normalize-space(.) = '%s']";
         WebElement field = palette.findElement(By.xpath(String.format(fieldXPath, fieldType)));
-        // NOTE: We scroll the page up because the drag&drop fails sometimes if the dragged field and the canvas (drop
-        // target) are not fully visible. See https://code.google.com/p/selenium/issues/detail?id=3075 .
-        getDriver().createActions().sendKeys(Keys.HOME).perform();
+        // NOTE: We scroll to the top of the page because the drag&drop fails sometimes if the dragged field and the
+        // canvas (drop target) are not fully visible. See https://code.google.com/p/selenium/issues/detail?id=3075 .
+        scrollToTop();
```
vmassol, line 107: Would be good to provide more explanations. The issue seems to indicate it's related to chrome but I don't see any mention of that in the comment for ex.

vmassol: It would be good to explain also why we use JS  in scrollToTop().

vmassol: These are tricky stuff and we need the max info that we can review in the future.

manuelleduc: I've made a tentative documentation improvement in another PR https://github.com/xwiki/xwiki-platform/pull/1600

PR #1600, opened 2021-04-16
[Misc] Page object documentation improvements

manuelleduc:
- Explains why a call to the javascript window.scrollTo method is the best way to scroll
- Documents why using XWikiWebDriver#scrollTo(int, int) is the best way to move to the top of the page
- Documents why we stopped using a home page key press action to move the view to the top of the page in ApplicationClassEditPage


vmassol:
@manuelleduc sounds good to me (I have no idea if the content of the doc is correct, i.e. whether the issue is only on chrome and that it's a chrome limitation that the home key press is async, but it's clear and makes sense). Only potential improvement would be a link to a resource on the internet showing this but don't spend time on it, if you have it you could use it. Otherwise it's fine and good already. Thanks!

PR #1610, opened 2021-04-26
XWIKI-14811:  Bad layout when using a font larger than the default one

slauriere:
-  Replace float positioning of the extension status element by a block display on small screens and a flex layout on lg screens


slauriere:
I see GitHub mentions conflicting files, sorry about that but I can't spot any conflict on my side when comparing these two files with the master branch. Any hint on how to fix this would be appreciated (GitHub won't show me the conflicts it detects since the "Resolve conflicts" action is only for users with write access to the target repository).

slauriere:
Closing this issue as invalid since the files got moved, I'll create a new one without conflict.

PR #1613, opened 2021-04-26
XWIKI-18546: Inconsistencies between frontend and backend registration check

slauriere:
- Use regex /^[a-zA-Z0-9_]+$/ to validate a username (same as the
  default one used by XWiki#createUser)
- Add i18n label for notifying the user about an invalid username
  directly in the form live validation step

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
@@ -192,6 +192,10 @@
         'mandatory' : {
           'failureMessage' : $services.localization.render('core.validation.required.message')
         },
+        'regex' : {
+          'pattern' : '/^[a-zA-Z0-9_]+$/',
+          'failureMessage' : $services.localization.render('xe.admin.registration.invalidUsername')
+        },
         'programmaticValidation' : {
           'code' : '#nameAvailable($request.get("xwikiname"))',
           'failureMessage' : $services.localization.render('core.register.userAlreadyExists')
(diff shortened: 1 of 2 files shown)


slauriere:
PS: as mentioned in a comment on [XWIKI-18546](https://jira.xwiki.org/browse/XWIKI-18546), accessing the property `xwiki.validusername` would be more appropriate than declaring the same default regex in `XWiki/Registration` but I could not figure out so far how to access that variable from Velocity.

mflorea:
Accessing the ``xwiki.validusername`` configuration property won't help as long as the default value is hard-coded in ``createUser``. The configuration can be missing, so you would still need to specify the default value, unless we introduce a dedicated ``getValidUserNameRegEx()`` somewhere (to have the default value in a single place).

mflorea (approved):

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
```diff
@@ -192,6 +192,10 @@
         'mandatory' : {
           'failureMessage' : $services.localization.render('core.validation.required.message')
         },
+        'regex' : {
+          'pattern' : '/^[a-zA-Z0-9_]+$/',
```
mflorea, line 196: You're forcing the default value from ``createUser`` but this doesn't mean that you remove the inconsistency: you still have it if the user configures a different value for the ``xwiki.validusername`` property.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/Registration.xml
slauriere: Indeed, so if I understand correctly we need to introduce a `getValidUserNameRegEx()` somewhere as you suggest. I'm wondering what could be the best place. Could it be in `DefaultUserConfiguration` and `UserScriptService` maybe?

PR #1614, opened 2021-04-26

slauriere:
- Replace float positioning of the extension status by a block display on small screens and a flex layout display on large ones with "space-between" content justification


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/extension.vm
```diff
@@ -1037,13 +1037,15 @@ $namespace##
...
-        <p class="extension-status">$escapetool.xml($extensionStatusMessage)</p>
-      #end
+      <div class="extension-title-and-status">
+        <h2 class="extension-title">
+          <span class="extension-name">#displayExtensionName($extension)</span>
+          <span class="extension-version">$escapetool.xml($extension.id.version)</span>
+        </h2>
+        #if ($extensionStatusMessage)
+          <p class="extension-status">$escapetool.xml($extensionStatusMessage)</p>
+        #end
```
mflorea, line 1047: I would rather move ``extension-status`` inside ``extension-title`` than introducing another wrapper.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.css
```diff
@@ -475,3 +471,10 @@ ul.collapsible.document-tree {
   font-weight: normal;
   padding-right: .3em;
 }
+
+@media (min-width: 992px) {
```
mflorea, line 475: You should mention in a comment the Bootstrap LESS variable that corresponds to this value (otherwise it looks too much like a magic number). I'm also wondering whether it's not better to move this to the skin (Flamingo) in order to be able to use LESS (introducing a WebJar for the Extension Manager is more work).

mflorea (changes requested):
As with any UI change, it would be nice to show some before/after screenshots.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/extension.vm
slauriere: I agree we should use wrappers wisely. I'm trying to implement a layout where there are two elements on the left (the extension name and its version) and one on the right (the extension status). Won't we end up grouping the two first ones in a common container from which the third one is excluded? Or do you see an alternative?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.css
slauriere: 1c504fe adds a comment indeed. We could move the CSS rules to the skin indeed, should that be in `resources/flamingo/admin.less`? The commit also applies the same layout change to the extension dependency panel, I'm adding a screenshot to the related JIRA issue itself.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/extension.vm
mflorea: It might be possible to use flex on the ``extension-title`` and set ``flex-grow`` on the ``extension-status`` to make it grow to fill the remaining space. Then align the text to the right on the ``extension-status``. I haven't tried it though.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.css
mflorea: A comment is fine for now I think, taking into account that this value is rarely changed from the color theme (close to never).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/extension.vm
slauriere: Indeed, good idea, thanks. The latest commit follows this path. It also sets the `dependency-item` background image position to make sure it remains aligned at the top even when the item spans across multiple lines, just like how the `extension-title` behaves. I'm adding screenshots to the JIRA issue to clarify.

slauriere: Meantime I actually removed the `dependency-item` background image positionining rule because it does not align the icon correctly in case of large font. This might be the subject of another issue.

slauriere: In the new layout, the extension version number is not wrapped (eg "Refactoring application"), while it was when the extension name and its version were grouped in a common block:

## New layout without additional wrapper

extension-updates-ld

However in the dependencies panel, the extension name and versions are wrapped, which can result in a slightly different display (eg with the "Administration Application"), this is something we may want to align visually in the scope of a different issue maybe:

extension-dependencies-ld

## Previous layout where the title and the version were wrapped

large-screen-display

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm
```diff
@@ -613,10 +613,12 @@ $namespace##
     #set ($dependencyName = "<a href=""$dependencyURL"" class=""extension-link"">#displayExtensionName($dependencyExtension)</a>")
   #end
   <div class="dependency-item extension-item-$dependencyStatus">
-    <span class="extension-name">${dependencyName}</span><span class="extension-version">$!dependencyVersion</span>
-    #if ($extensionNamespace.startsWith('wiki:'))
-      <span class="extension-namespace">$services.localization.render('extensions.info.dependency.wiki', ["#wikiHomePageLink($extensionNamespace)"])</span>
-    #end
+    <div class="dependency-item-title">
```
mflorea, line 616: Why do you need to group these? It goes against the extension display where the name, version and status are siblings. I think we should be consistent: if we keep them as siblings in the extension display we should do the same in the extension dependency display. Basically, in both cases we have a "line" that shows extension name, version, name space (optional) and status (optional).

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm
slauriere: Indeed, using siblings in both places will be more consistent, I updated the structure accordingly in the latest commit, which results in the display below with default font / large font. Many thanks for your guidance.

extension-dependency

extension-dependency-large-font

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.css
```diff
@@ -475,3 +476,21 @@ ul.collapsible.document-tree {
   font-weight: normal;
   padding-right: .3em;
 }
+
+/* Media query for screens wider than Bootstrap @screen-md-min */
+@media (min-width: 992px) {
+  /* Display the extension title and status on a single row on md screens */
+  .extension-title, .dependency-item {
+    display: flex;
```
mflorea, line 484:
```suggestion
    display: flex;
    align-items: center;
```

I think you're just missing the vertical alignment.

mflorea (changes requested):
Just one small remark.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.css
slauriere: Indeed thx

mflorea (approved):

PR #1615, opened 2021-04-26
XWIKI-18556: Rights UI only display rights with the proper case

slauriere:
- In the UI JavaScript, convert the rights to lowercase before comparing them with the
  reference rights ('admin', 'view', etc.)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
@@ -399,7 +399,7 @@ function displayUsersAndGroups(row, i, table, idx, form_token, targetDocument)
   var uorg = table.json.uorg;
   // Rights are case insensitive in XWiki security module, and the reference rights defined in rightsUI.vm
   // are lowercase, hence we make sure to convert all applying rights to lowercase before comparing them with the
-  // reference ones (see XWIKI-18556).
+  // reference ones.
   var allows = row.allows.map(entry => entry.toLowerCase());
   var denys = row.denys.map(entry => entry.toLowerCase());
 
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
@@ -397,8 +397,11 @@ function displayUsersAndGroups(row, i, table, idx, form_token, targetDocument)
 {
   var userurl = row.userurl;
   var uorg = table.json.uorg;
-  var allows = row.allows;
-  var denys = row.denys;
+  // Rights are case insensitive in XWiki security module, and the reference rights defined in rightsUI.vm
+  // are lowercase, hence we make sure to convert all applying rights to lowercase before comparing them with the
+  // reference ones (see XWIKI-18556).
+  var allows = row.allows.map(entry => entry.toLowerCase());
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -397,8 +397,11 @@ function displayUsersAndGroups(row, i, table, idx, form_token, targetDocument)
 {
   var userurl = row.userurl;
   var uorg = table.json.uorg;
-  var allows = row.allows;
-  var denys = row.denys;
+  // Rights are case insensitive in XWiki security module, and the reference rights defined in rightsUI.vm
+  // are lowercase, hence we make sure to convert all applying rights to lowercase before comparing them with the
+  // reference ones (see XWIKI-18556).
```
tmortagne, line 402: I try to avoid reference to jira issue in the code and instead make sure that everything needed is in the comment (which seems to be the case here).

slauriere: Thank you Thomas for the review. I removed the jira reference indeed.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -397,8 +397,11 @@ function displayUsersAndGroups(row, i, table, idx, form_token, targetDocument)
 {
   var userurl = row.userurl;
   var uorg = table.json.uorg;
-  var allows = row.allows;
-  var denys = row.denys;
+  // Rights are case insensitive in XWiki security module, and the reference rights defined in rightsUI.vm
+  // are lowercase, hence we make sure to convert all applying rights to lowercase before comparing them with the
+  // reference ones.
+  var allows = row.allows.map(entry => entry.toLowerCase());
+  var denys = row.denys.map(entry => entry.toLowerCase());
```
vmassol, line 404: FWIW `toLowerCase()` should be avoided. The signature taking a locale should be used. Here it's fine because we're pretty sure that the right will not contain some characters. However it's best to use `Locale.ROOT`. See https://rules.sonarsource.com/java/RSPEC-1449

vmassol: EDIT: I thought this was java... I have no idea about javascript, needs to be researched....

vmassol: Seems the issue is the same: https://www.w3schools.com/jsref/jsref_tolocalelowercase.asp But I don't know what's the recommended solution.

vmassol: So it seems that in javascript `toLowerCase()` is the method to use when case insensitive and `toLocaleLowerCase()` is locale sensitive. So it looks good here. We should review and fix our java usages though.

vmassol: > We should review and fix our java usages though.

For ex: https://github.com/xwiki/xwiki-platform/blob/bc6fb452ef10414c1756b2b547e8129ec1e35c9d/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/uorgsuggest.vm#L22

mflorea (approved):

PR #1616, opened 2021-04-26
XWIKI-18399: Admin Rights UI : rights of local users/groups are not updated properly if prefixed with the wiki name

slauriere:
- In subwikis, use regex to match a user or a group prefixed or not with the current wiki identifier
  (same approach as code fixing XWIKI-9730)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
```diff
@@ -28,8 +28,15 @@
 
 #if ($!{services.csrf.isTokenValid("$!{request.getParameter('form_token')}")})
 #foreach($obj in $doc.getObjects($clsname)) ## XWiki.XWikiGlobalRights or XWiki.XWikiRights
+  ## In subwikis, users or groups may be referenced by "wiki:Space.Username" or "Space.Username"
+  #set ($wikiname = $xcontext.database)
```
mflorea, line 32: Please use camel-case as per our [Java code style](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
```diff
@@ -28,8 +28,15 @@
...
+  #if ($wikiname == "xwiki")
```
mflorea, line 33: You shouldn't hard-code "xwiki". If you want to treat the main wiki differently then we have ``$xcontext.mainWikiName``.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
```diff
@@ -28,8 +28,15 @@
...
+    #set ($uorgmatch = "${regextool.quote($fullname)}")
+  #else
+    #set ($uorgmatch = "(" + ${regextool.quote($wikiname)} + ":)?" + "${regextool.quote($fullname)}")
+  #end
   #set($pers = "$!obj.getProperty($uorg).getValue()")
-  #if($pers.matches("${regextool.quote($fullname)},?")) ## found the rule containing that person
+  #if($pers.matches("${uorgmatch},?")) ## found the rule containing that person
```
mflorea, line 39: What happens if there are two users ``XWiki.Bob`` and ``XWiki.Bobby``? Using regular expressions for this feels error-prone to me. Why not resolve each user/group reference from the value (relative to the current wiki) and compare the absolute reference?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
```diff
@@ -90,9 +97,9 @@
         #end
       #end
     #end ## end get property
-  #elseif($pers.matches("^(.*,)?${regextool.quote($fullname)}(,.*)?$")) ## The person is in a multiple selection rule, remove it from there.
-    #set($pers = $pers.replaceAll("^${fullname}(,|$)", ""))
-    #set($pers = $pers.replaceAll("(^|,)${fullname}(,|$)", "$2"))
+  #elseif($pers.matches("^(.*,)?${uorgmatch}(,.*)?$")) ## The person is in a multiple selection rule, remove it from there.
+    #set($pers = $pers.replaceAll("^${uorgmatch}(,|$)", ""))
+    #set($pers = $pers.replaceAll("(^|,)${uorgmatch}(,|$)", "$3"))
```
mflorea, line 102: Again, this feels error-prone. I find it cleaner and safer to resolve the list of user/group references and then compare. It's not like we have thousands of selected users/group, but even if it were the case, resolving a document reference is fast enough.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
slauriere: I agree with you that using references is more appropriate and will do so. I just checked the case you mention to make sure the existing product code (which is based on regular expressions) was not misconfiguring access rights in such a case and it is ok, both with single and multiple user(s)/group(s) rules, because of the following:

```
#set ($pers = 'XWiki.bobby')
#set ($regex = $regextool.quote('XWiki.bob'))
#set ($result = $pers.matches("$regex,?")) ## Does not match. Would match with "$regex,?.*"

#set ($pers = 'XWiki.bobby,XWiki.alice')
#set ($result = $pers.matches("^(.*,)?$regex(,.*)?$")) ## Does not match

#set ($pers = 'XWiki.bobby,XWiki.bob,XWiki.alice')
#set ($result = $pers.matches("^(.*,)?$regex(,.*)?$")) ## Matches, however the subsequent replace calls do not replace anything:
#set ($pers = $pers.replaceAll("^${regex}(,|$)", "")) ## No match because the regex needs to be followed by a comma, a pipe, or end of line
#set ($pers = $pers.replaceAll("(^|,)${regex}(,|$)", "$2")) ## No match either for the same reason
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
slauriere: Indeed, and I realized in the meantime that there was not always a third group in the match, since I introduced two regex: one for the main wiki, one for the subwikis, so this code was broken. This will be handled differently with references.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
slauriere: dce6e16 uses references instead of regular expressions. I tested the code with single and multiple user/group access rules on both a subwiki and the main wiki with relative or absolute references. There might be shorthand methods for dealing in a uniform manner with property values consisting of strings separated with comma or pipes but couldn't find any so far. 

Other templates use regular expressions for matching a given user or group identifier, eg `rightsUI.vm`, `getusersandgroups.vm`, I'll create issues for using references in these templates as well.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
slauriere: Well noted thanks. In the latest commit, there is no need to retrieve the wiki name anymore.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
slauriere: Well noted as well. The latest commit proposes a more generic approach which does not treat the main wiki differently, so I'm resolving this conversation.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/saverights.vm
slauriere: The latest change in this pull request resolves each user/group into a reference indeed.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/saverights.vm
```diff
@@ -27,11 +27,28 @@
 #set($foundd = false)
 
 #if ($!{services.csrf.isTokenValid("$!{request.getParameter('form_token')}")})
+## Convert the user or group identifier present in the request to a DocumentReference relative to the current wiki
+## if no wiki is specified.
+#set ($userOrGroupReference = $services.model.resolveDocument($fullname, 'current'))
```
mflorea, line 32:
```suggestion
#set ($userOrGroupReference = $services.model.resolveDocument($fullname))
```

``current`` is already the default resolver used [it seems](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/script/ModelScriptService.java#L69).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/saverights.vm
```diff
@@ -27,11 +27,28 @@
...
+  ## in order to compare these references to the reference of the user or group for which the rights may need to
+  ## be updated. User or groups can be separated by a comma or a pipe (but not both in the same string),
+  ## so we split the string once for each.
+  #set ($userOrGroupIdentifiers = $pers.split(','))
+  #if ($$userOrGroupIdentifiers.length == 1)
+    #set ($userOrGroupIdentifiers = $ids[0].split('|'))
+  #end
+  #set ($ruleUserOrGroupReferences = [])
+  #foreach ($userOrGroupIdentifier in $userOrGroupIdentifiers)
+    #set ($reference = $services.model.resolveDocument($userOrGroupIdentifier, 'current'))
```
mflorea, line 45:
```suggestion
    #set ($reference = $services.model.resolveDocument($userOrGroupIdentifier))
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/saverights.vm
```diff
@@ -27,11 +27,28 @@
...
+    #set ($discard = $ruleUserOrGroupReferences.add($reference))
+  #end
+  #if ($ruleUserOrGroupReferences.size() == 1 && $ruleUserOrGroupReferences.contains($userOrGroupReference))
+    ## Found the rule containing the given user or group and this user or group only
+    #set($levels = "$obj.getValue('levels')") ## get the levels
+    #if($obj.getValue('allow') == 1) ## found allow rule
```
mflorea, line 51:
```suggestion
    #set ($levels = "$obj.getValue('levels')") ## get the levels
    #if ($obj.getValue('allow') == 1) ## found allow rule
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/saverights.vm
```diff
@@ -90,10 +107,14 @@
         #end
       #end
     #end ## end get property
-  #elseif($pers.matches("^(.*,)?${regextool.quote($fullname)}(,.*)?$")) ## The person is in a multiple selection rule, remove it from there.
-    #set($pers = $pers.replaceAll("^${fullname}(,|$)", ""))
-    #set($pers = $pers.replaceAll("(^|,)${fullname}(,|$)", "$2"))
-    $doc.getObject($clsname, $obj.number).set($uorg, $pers)
+  #elseif($ruleUserOrGroupReferences.contains($userOrGroupReference))
```
mflorea, line 110:
```suggestion
  #elseif ($ruleUserOrGroupReferences.contains($userOrGroupReference))
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/saverights.vm
```diff
@@ -27,11 +27,28 @@
...
-    #if($obj.getProperty('allow').getValue() == 1) ## found allow rule
+  #set ($pers = "$!obj.getValue($uorg)")
+  ## Convert the user or group identifiers contained in the current access rule to an array of DocumentReferences
+  ## in order to compare these references to the reference of the user or group for which the rights may need to
+  ## be updated. User or groups can be separated by a comma or a pipe (but not both in the same string),
+  ## so we split the string once for each.
+  #set ($userOrGroupIdentifiers = $pers.split(','))
+  #if ($$userOrGroupIdentifiers.length == 1)
+    #set ($userOrGroupIdentifiers = $ids[0].split('|'))
+  #end
```
mflorea, line 42:
```suggestion
  #set ($userOrGroupIdentifiers = $stringtool.split($pers, ',|'))
```

See http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#split-java.lang.String-java.lang.String- .

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/saverights.vm
slauriere: Thank you for these change requests, which I applied in the latest commit.

slauriere: (I executed a new run of tests after the latest changes with local / global users or groups, both with single and multiple targets per rule)

mflorea (approved):

PR #1618, opened 2021-04-27
[misc] Integrates a page test for the mentions notifications template

manuelleduc:
- Defines a page test
- Removes a docker test, now better tested with the page test
- Improved velocity template documentation


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/test/java/org/xwiki/mentions/templates/MentionTest.java
```diff
@@ -0,0 +1,150 @@
...
+            cleanup(IOUtils.toString(getClass().getResourceAsStream("/templates/mentions/mention.mention.html"),
+                UTF_8));
+
+        assertEquals(expected, actual);
+    }
+
+    private String cleanup(String render)
+    {
+        return render.trim().replaceAll("\\s+", " ");
+    }
```
manuelleduc, line 149: This should be moved to a test util method, we need to define where.

manuelleduc: This is already a code duplication and it will be useful for future template page tests.

vmassol: see https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$16195289384221TGmYu:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=dodsorf.as

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/test/java/org/xwiki/mentions/templates/MentionTest.java
```diff
@@ -74,24 +75,16 @@ void mention() throws Exception
     {
         TemplateManager templateManager = this.oldcore.getMocker().getInstance(TemplateManager.class);
         VelocityManager velocityManager = this.oldcore.getMocker().getInstance(VelocityManager.class);
-        VelocityContext velocityContext = velocityManager.getVelocityContext();
+        this.componentManager.registerComponent(ScriptService.class, "icon", this.iconManager);
+        this.componentManager.registerComponent(ScriptService.class, "date", this.dateScriptService);
+        this.componentManager.registerComponent(ScriptService.class, "localization", this.localizationScriptService);
```
vmassol, line 80: I don't understand this. It shouldn't be needed if you have the component in the `@ComponentList`.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/pom.xml
```diff
@@ -73,19 +73,14 @@
...
-      <artifactId>xwiki-platform-web-templates</artifactId>
-      <version>${project.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.xwiki.platform</groupId>
-      <artifactId>xwiki-platform-icon-script</artifactId>
+      <artifactId>xwiki-platform-date</artifactId>
       <version>${project.version}</version>
       <scope>test</scope>
```
vmassol, line 78: I don't understand this scope. Why is it for testing only?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/test/java/org/xwiki/mentions/templates/MentionTest.java
```diff
@@ -0,0 +1,147 @@
...
+     * Needs to be mocked because don't want the test result to be dependent of the current time.
+     */
+    @Mock
+    private DateScriptService dateScriptService;
+
+    @BeforeEach
+    void setUp() throws Exception
+    {
+        // Initializes then environment for the icon extension.
+        IconSetup.setUp(this.xwiki, this.context, this.oldcore);
```
manuelleduc, line 77: Icon environment setup. It could be moved to `PageTest`, to initialize the icons even if they are not always used in the test, to prevent testers to forget to call this initializer.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/IconSetup.java
```diff
@@ -0,0 +1,72 @@
...
+     * @throws XWikiException when an error occurs when accessing a document
+     * @throws ComponentLookupException when an error occurs when resolving a component
+     * @throws IOException when a error occurs when loading the icon page content from the resources
+     */
+    public static void setUp(XWiki xwiki, XWikiContext context, MockitoOldcore oldcore)
+        throws XWikiException, ComponentLookupException, IOException
+    {
+        XWikiDocument document =
+            xwiki.getDocument(new DocumentReference("xwiki", "Fake", "IconTheme"), context);
+        document.setContent(IOUtils.toString(IconSetup.class.getResourceAsStream("/iconPage.txt"), UTF_8));
```
manuelleduc, line 66: Load the fake icon theme page content from the file system. `iconPage.txt` is probably not the best name for it though.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/PageTest.java
```diff
@@ -139,6 +187,9 @@ public void setUpComponentsForPageTest(MockitoComponentManager componentManager)
 
         CacheManager cacheManager = componentManager.registerMockComponent(CacheManager.class);
         when(cacheManager.createNewCache(any(CacheConfiguration.class))).thenReturn(mock(Cache.class));
+        CacheFactory cacheFactory = mock(CacheFactory.class);
+        when(cacheManager.getCacheFactory()).thenReturn(cacheFactory);
+        when(cacheFactory.newCache(any())).thenReturn(mock(Cache.class));
```
vmassol, line 192: Why is this needed in PageTest and not in OldCoreTest for example?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/PageTest.java
```diff
@@ -331,41 +382,13 @@ private void initializeSkinEnvironment() throws ComponentLookupException
...
-                }
-                return is;
-            });
+        Environment environment = this.oldcore.getMocker().getInstance(Environment.class);
+        when(environment.getResource(any(String.class))).thenAnswer(new GetResourceAnswer());
+        when(environment.getResourceAsStream(any(String.class))).thenAnswer(new GetResourceAsStreamAnswer());
+
+        XWikiEngineContext engineContext = mock(XWikiEngineContext.class);
+        when(engineContext.getResourceAsStream(any())).thenAnswer(new GetResourceAsStreamAnswer());
+        this.xwiki.setEngineContext(engineContext);
```
manuelleduc, line 391: `Answer<?>` anonymous classes moved to nested classes at the top of the class. 
`GetResourceAsStreamAnswer` reused for the class to `getResourceAsStream` from xwiki engine context.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/IconSetup.java
```diff
@@ -0,0 +1,72 @@
...
+        xwiki.saveDocument(document, context);
```
vmassol, line 67: Maybe we don't need to do this and instead load the real page using `loadPage()`.

manuelleduc: Indeed, that's way simpler, thanks.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/pom.xml
manuelleduc: Should all the script services used in the templates of a module be added to the module's dependencies?
If that's the case, then I must remove the test scope for this one and add some other dependencies (icons + localization).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/IconSetup.java
```diff
@@ -0,0 +1,72 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Setup an icon environment for the page tests.
+ *
+ * @version $Id$
+ * @since 13.4RC1
+ */
+public final class IconSetup
```
vmassol, line 44: We need to discuss this strategy/pattern of using new classes for setup. So far everything has been done in OldCoreTest or PageTest AFAIK.

vmassol: > 

Not sure. Just seen that I added LocalizationSetup in the past to fix a class fan out. We still need to discuss whether we want to call it in `setUpForPageTest` or not.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/LocalizationSetup.java
```diff
@@ -52,15 +55,19 @@ public static void setUp(TestComponentManager tcm) throws Exception
...
+        when(lss.render(anyString(), anyCollection())).thenAnswer((Answer<String>) invocationOnMock -> {
+            // Displays the comma separated list of parameters between squared brackets after the translation key as 
+            // the value.
+            // For instance: my.key [paramA, paramB]
+            Object key = invocationOnMock.getArgument(0);
+            Collection<?> parameters = invocationOnMock.getArgument(1);
+            return parameters.stream()
+                .map(String::valueOf)
+                .collect(Collectors.joining(", ", key + " [", "]"));
+        });
```
vmassol, line 71: I think we need to review the mocking we do in PageTest since we talked about mocking the minimum now.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-page/src/main/java/org/xwiki/test/page/ScriptServiceComponentList.java
```diff
@@ -0,0 +1,73 @@
...
+import org.xwiki.skinx.internal.LinkSkinExtension;
+import org.xwiki.test.annotation.ComponentList;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Pack of default {@link ScriptService} implementations.
```
vmassol, line 47: I think I'd prefer to group by domain, i.e. by top level module in xwiki-platform for ex. It doesn't make sense to me to group by function since it means test code cannot decide to use only one domain and not the other.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/pom.xml
vmassol: What I meant is that if the vm uses the icon script service, then surely this is a runtime depenency and thus needs to be added to the module that has that template, and not as a "test" scope.

PR #1619, opened 2021-04-29
XWIKI-18615: Notification widget styling improvement

slauriere:
- Add spacing at the bottom and around each widget
- Add padding and border radius
- Use normal font weight
- Set notification container maximum width to 40%


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/notification.css
```diff
@@ -2,24 +2,27 @@
...
   bottom: 0;
-  left: 0%;
-  width: 100%;
+  left: 0;
+  right: 0;
+  margin: auto;
   text-align: center;
   z-index: 1200;
   display: block !important;
+  max-width: 40%;
```
mflorea, line 11: Is this OK on small screens? We might want to use a different max width in this case.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/notification.css
```diff
@@ -2,24 +2,27 @@
...
 }
 .xnotification {
   cursor: pointer;
-  display: inline;
+  display: inline-block;
   background: #333 2px center no-repeat;
   border: none;
-  padding: 0px 6px;
+  padding: 14px 28px;
+  border-radius: 3px;
```
mflorea, line 22: Even if we can't use LESS here it would be good to use the values from the skin. We use 4px border radius by default for instance, as for the padding, you should maybe look for some UI element that is similar (uses similar padding, maybe breadcrumbs). My point is to try to have a consistent style across XWiki.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/notification.css
slauriere: Thank you for the review. I made this rule responsive indeed, turning it to 80% for screens narrower than 992px (@screen-md-min).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/notification.css
slauriere: I made this more consistent indeed in the latest commit, using the same padding as the one in the breadcrumb, and the default border radius.

mflorea (approved):

PR #1625, opened 2021-05-03
XWIKI-18598: Replace the like pages live table in the user profiles with live data

manuelleduc:


vmassol:
@manuelleduc BTW could you show a before/after screenshot? I'd like to verify that the "after" looks at least as good as the "before". Thanks!

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/pom.xml
```diff
@@ -45,6 +45,13 @@
       <version>${project.version}</version>
       <type>xar</type>
     </dependency>
+    <!-- Required to display the like pages user profile UIX. -->
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-user-profile-ui</artifactId>
+      <type>xar</type>
+      <version>${project.version}</version>
+    </dependency>
```
vmassol, line 54: Shouldn't this dep be declared somewhere else in the like modules? why is it only needed for testing and not for runtime?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/src/test/it/org/xwiki/like/test/ui/LikeIT.java
```diff
@@ -129,4 +132,28 @@ void likeUnlikeDefaultConfiguration(TestUtils testUtils, TestReference testRefer
...
+    void userProfileUIXLiveData(TestUtils testUtils, TestReference testReference)
+    {
+        testUtils.login(USER1, USER1);
+
+        // Re-creates and likes a page.
+        testUtils.deletePage(testReference);
+        testUtils.createPage(testReference, "", "");
+        LikeButton likeButton = new LikeButton();
+        likeButton.clickToLike();
+        assertEquals(1, likeButton.getLikeNumber());
```
vmassol, line 147: Why is this needed? We're supposed to be in a scenario and it would be weird that we don't already like a page in LikeIT elsewhere.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/src/test/it/org/xwiki/like/test/ui/LikeIT.java
```diff
@@ -129,4 +132,28 @@ void likeUnlikeDefaultConfiguration(TestUtils testUtils, TestReference testRefer
...
+        // Go to the Like Pages user profile tab and assert that the like page is correctly displayed in the live data.
+        UserProfileLikePagesPage userProfileLikePagesPage = new UserProfileLikePagesPage(USER1);
+        userProfileLikePagesPage.gotoPage();
+        LiveDataElement likedPages = userProfileLikePagesPage.getLiveData();
+        likedPages.waitUntilHasContentReady();
+        assertEquals(1, likedPages.countRows());
+        assertTrue(likedPages.hasLinkRow(TITLE_COLUMN_NAME, testReference.getLastSpaceReference().getName(),
+            testUtils.getURL(testReference.getLastSpaceReference())));
+        assertTrue(likedPages.hasRow(LIKES_COLUMN_NAME, "1"));
+    }
```
vmassol, line 158: Generally speaking I don't understand this test method. For me it's part of testing the like feature:
- you like a page somewhere
- you then go to the use profile and navigate to the like tab and verify it's displayed

vmassol: The way you've written it is clearly as a unit test to me and not as a functional test (i.e. scenario). I've seen that in other tests you added in the past but I didn't report it back then.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-pageobjects/src/main/java/org/xwiki/like/test/po/UserProfileLikePagesPage.java
```diff
@@ -0,0 +1,81 @@
...
+import org.xwiki.livedata.test.po.LiveDataElement;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.test.ui.po.BaseElement;
+
+/**
+ * Provides the operations to interact with the Like Pages tab of a user profile.
+ *
+ * @version $Id$
+ * @since 13.4RC1
+ */
```
vmassol, line 34: I guess this means that @surli forgot to write some tests for the Like tab in the user profile in the past ;)

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-pageobjects/src/main/java/org/xwiki/like/test/po/UserProfileLikePagesPage.java
```diff
@@ -0,0 +1,81 @@
...
+     * The Likes column name.
+     */
+    public static final String LIKES_COLUMN_NAME = "Likes";
+
+    private final String userName;
+
+    /**
+     * Default constructor.
+     *
+     * @param userName a user name
```
vmassol, line 57: Doesn't feel useful to me...

vmassol: See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HJavadocBestPractices

vmassol: what is a "user name"? Is it "Vincent" (for ex) or "XWiki.Vincent"? At least you should explain that and always give an example.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-pageobjects/src/main/java/org/xwiki/like/test/po/UserProfileLikePagesPage.java
```diff
@@ -0,0 +1,81 @@
...
+        this.userName = userName;
+    }
+
+    /**
+     * Go to the Liked Pages tab of the user profile of the user passed in the constructor.
+     */
+    public void gotoPage()
+    {
+        Map<String, Object> queryParameters = new HashMap<>();
+        queryParameters.put("category", "org.xwiki.platform.like.userProfileMenu");
```
vmassol, line 70: small detail, you could use `Collections.singletonMap`.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/pom.xml
```diff
@@ -48,5 +48,18 @@
...
+      <artifactId>xwiki-platform-livedata-macro</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-livedata-livetable</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
```
vmassol, line 63: why don't you add an optional dep on the user profile since we add a new tab for likes?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
```diff
@@ -254,30 +254,53 @@
...
+  'xpage': 'plain',
+  'outputSyntax': 'plain'
+}))
+##
+{{liveData
+  id='likedPages'
+  source='liveTable'
+  sourceParameters="$sourceParameters"
+  properties='doc.title,doc.location,like'
+}}$jsontool.serialize($liveDataConfig){{/liveData}}
```
vmassol, line 303: Interestingly it doesn't seem that using LD is simpler than using the previous LT. There's actually more lines to write it seems... Isn't there a simpler way?

Also I see `source='liveTable'`. Does it mean we cannot yet use a LD source or are there other considerations (maybe some backward compat) that made you reuse the LT source?

Thanks

vmassol: > Interestingly it doesn't seem that using LD is simpler than using the previous LT

it seems actually a lot more complex seen that we also need java code...

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/XWikiLivedata.vue
```diff
@@ -25,7 +25,7 @@
   totally autonomous.
 -->
 <template>
-  <div class="xwiki-livedata">
+  <div class="xwiki-livedata" :id="data.id">
```
vmassol, line 28: I'm curious: is this related to this PR or is it some limitation/bug that we had on the LD component?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/pom.xml
manuelleduc: The user profile is only need to add the UIXP the load the UIX I want  to test. Should the UIX modules declare an optional dependency to the corresponding UIXP modules?

vmassol: See https://github.com/xwiki/xwiki-platform/pull/1625#discussion_r625027599

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/src/test/it/org/xwiki/like/test/ui/LikeIT.java
manuelleduc: At this point in the test user1 has 0 pages likes. I could indeed use user2 instead, which already should have a liked page.

vmassol: yes, in the same test and not in a new `@Test`.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/src/test/it/org/xwiki/like/test/ui/LikeIT.java
manuelleduc: I'm in addition testing that the displayed content is the one expected (the right liked page with the right likes number). If that's over-testing I'm ok to remove these assertions.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/XWikiLivedata.vue
manuelleduc: Adding the id is needed to select the right DOM element in the page.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-test/xwiki-platform-like-test-docker/src/test/it/org/xwiki/like/test/ui/LikeIT.java
vmassol: That seems good

manuelleduc:
> @manuelleduc BTW could you show a before/after screenshot? I'd like to verify that the "after" looks at least as good as the "before". Thanks!

**Before:**
Screenshot_2021-05-03 Profile of Administrator - XWiki

**After:**
Screenshot_2021-05-03 Profile of Administrator - XWiki(1)

vmassol:
Thanks for the screenshot. The cell alignment of headings and text don't seem as good as before, WDYT?

vmassol:
Note that the location column doesn't have links anymore. Is that wanted?

manuelleduc:
> Thanks for the screenshot. The cell alignment of headings and text don't seem as good as before, WDYT?
> 
> For title and location it's not great but for Likes it's quite bad IMO...

The small gap at the left of the header is for an element allowing to reorder columns, and only visible on hover (cc @mflorea @ClemDee )

vmassol:
what options do you have when you click the kebab menu?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/pom.xml
manuelleduc: > what options do you have when you click the kebab menu?

Screenshot_2021-05-03 Profile of Administrator - XWiki(3)

manuelleduc:
> Note that the location column doesn't have links anymore. Is that wanted?

Screenshot_2021-05-03 Profile of Administrator - XWiki(2)

Is seems to be following the breadcrumb style where the spaces are in blue, and the document is in grey (the grey text is a link).

manuelleduc:
> BTW would be good to add column sorting and filtering IMO, for all columns (which we didn't seem to have before with the LT)

@surli would it be difficult to add sorting and filtering?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/pom.xml
vmassol: ah so we can filter and sort now? I was surprised to not see any filtering inputs in the view UI as we had for LT. Is that a choice we have now? Marius explained that to me but I don't remember ;)

manuelleduc: Filtering and sorting are supported by LD, it just happens that the migrated LT did not support them.

vmassol:
> @surli do you remember why sorting and filtering wasn't supported previously?

just more work ;)

vmassol:
> @surli do you remember why sorting and filtering wasn't supported previously?

This seems to indicate that we don't have sorting nor filtering  with the LD, yet you've shown the kebab menu showing iflteirnf and sorting. Doesn't seem great and they should be hidden or greyed out if they're not active/usable. WDYT?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/pom.xml
vmassol: > Filtering and sorting are supported by LD,

AFAIR we support inputs for simple search and the kebab menu for advanced filtering, but not sure.

manuelleduc: Yes, if a column can be filtered a filter header is displayed, allowing to filter on a single criteria. Using the filter menu from the kebab button allows for advanced filter, not sure how advanced though.

vmassol: > not sure how advanced though.

You can combine several filters for one column with AND or OR AFAIR.

manuelleduc:
> > @surli do you remember why sorting and filtering wasn't supported previously?
> 
> @manuelleduc This seems to indicate that we don't have sorting nor filtering with the LD, yet you've shown the kebab menu showing iflteirnf and sorting. Doesn't seem great and they should be hidden or greyed out if they're not active/usable. WDYT?

Screenshot_2021-05-03 Profile of Administrator - XWiki(4)

This is the current result after clicking on filter and sort from the kebab menu. The buttons can indeed probably be greyed when noting more is proposed in the sort and filter screens.

surli:
>   @surli do you remember why sorting and filtering wasn't supported previously?

So as Vincent mentioned it was more work, but also I was not quite sure how to do it properly for all columns: this LT is mixing data from Solr (number of likes and references) and from the main DB (title of the page, human-readable location). So right now both column Locations and Title would be sorted the same way if I use only the reference, which might be a bit weird for end users. And filtering cannot be easily done since it would involve to filter first all pages from main DB and then to perform a new request to get the likes from Solr.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
```diff
@@ -254,30 +165,53 @@
...
-  'url': $pageToCall.getURL('get', $escapetool.url($queryParams))
+#set ($currentUser = $services.model.serialize($doc.documentReference, 'default'))
+#set ($liveDataConfig = {
+  'meta': {
+    'propertyDescriptors': [
+       {
+         'id': 'doc.title',
+         'displayer': 'link',
+         'sortable': false,
+         'filterable': false,
```
vmassol, line 180: why don't we want that sortable or filterable?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
```diff
@@ -254,30 +165,53 @@
...
+         'name': $docTitleDisplayName
+       },
+       {
+         'id': 'doc.location',
+         'displayer': 'html',
+         'sortable': false,
+         'filterable': false,
```
vmassol, line 187: same question

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
```diff
@@ -254,30 +165,53 @@
...
+         'name': $docLocationDisplayName
+       },
+       {
+         'id': 'like',
+         'displayer': 'number',
+         'sortable': false,
+         'filterable': false,
```
vmassol, line 194: same

vmassol: shouldn't the types be autodiscovered?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
manuelleduc: Because the result page does not support filtering and sorting.

vmassol: I guess that would be one improvement: to have a better way to write custom results page (in LD or LT or both) so that it's easy to support filtering and sorting, especially for well-known fields.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/pom.xml
mflorea: @vmassol the (old) "Liked Pages" live **table** was neither filterable nor sortable so you shouldn't be surprised that the (new) live **data** is neither filterable nor sortable. Both use the same source (which doesn't support sort and filter ATM).

@manuelleduc BTW, when migrating live tables, please compare the result you get after your changes (with the live data macro) with the result you get before your changes but with ``?useLiveData=true`` in the URL. It should be the same.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
mflorea: @vmassol as mentioned on #xwiki, we're reusing the live table source for now since we haven't rewritten it as a generic live data source (there's a bridge). As for the live data configuration, most of it should be automatically generated (when using an xclass) so we should be writing less lines of code than before, if not then there's probably a mistake.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
surli: >  to have a better way to write custom results page (in LD or LT or both) so that it's easy to support filtering and sorting, especially for well-known fields.

Don't forget that here the results are coming from a mix of two DBs results, so a generic solution probably wouldn't fit here.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
```diff
@@ -254,30 +165,49 @@
...
+         'id': 'doc.location',
+         'sortable': false,
+         'filterable': false,
+         'name': $docLocationDisplayName
+       },
+       {
+         'id': 'like',
+         'displayer': 'number',
+         'sortable': false,
+         'filterable': false,
```
mflorea, line 192: These 2 can be omitted, as properties need to be made sortable and filterable explicitly. For the doc title and location is different because the default configuration makes them sortable and filterable so we need to overwrite that I think.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
vmassol: yes sure, my comment was not about the like but more generic. I put it after the column declarations, my bad it wasn't clear. In any case you explained it to me and I saw the code re the defaults. Thx

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/Code/LiveTableResultPage.xml
```diff
@@ -0,0 +1,135 @@
...
+  <content>{{velocity}}
+#if ($xcontext.action == 'get')
+  #template('hierarchy_macros.vm')
+  #if("$!{request.xpage}" == 'plain')
+    $response.setContentType('application/json')
+  #end
+  #if($request.user)
+    #set ($userReference = $services.model.resolveDocument($request.user))
+  #else
+    #set ($userReference = $services.model.resolveDocument($xcontext.userReference))
```
mflorea, line 48: ``$xcontext.userReference`` is already a ``DocumentReference``.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
mflorea: > Now for the like column we could imagine that a default would be defined by the source (with a future LD source). It would allow to specify it in one place and have all LD using it inherit from it (unless they really need to change the type but it's a rare case IMO).

Yes, that's the plan.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
mflorea: I can see ``sortable`` and ``filterable`` are still explicitly set to ``false`` for the ``like`` column. Why?

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/UserProfileUIX.xml
```diff
@@ -254,30 +165,43 @@
...
-#set ($pageToCall = $xwiki.getDocument($pageToCallReference))
-#set ($queryParams = {
-  "user": $currentUser,
-  "livetable": "true",
-  "xpage": "plain",
-  "outputSyntax": "plain"
-})
-#set($options = {
-  'url': $pageToCall.getURL('get', $escapetool.url($queryParams))
+#set ($currentUser = $services.model.serialize($doc.documentReference, 'default'))
```
mflorea, line 172: This doesn't seem to be used, is it?

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-ui/src/main/resources/XWiki/Like/Code/LiveTableResultPage.xml
vmassol: Also note that `getUserLikes` takes a `UserReference` as input. Using a document reference is a bit the old way. 

To get the current user:

`UserReference userReference = $services.user.currentUserReference`

To convert from a user as a string to a `UserReference`, well you don't need to do anything since we have a converter that converts from String to `UserReference` so you can just pass `$request.user` :)

See https://extensions.xwiki.org/xwiki/bin/view/Extension/User%20Module/API/ for more details.

PR #1627, opened 2021-05-05
XWIKI-18501: Extension updater propose to upgrade Refactoring UI to 13.1 when 12.10.5 flavor is installed

slauriere:
- Add dependency on `xwiki-platform-refactoring-api`

### DIFF
--- a/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/pom.xml
@@ -36,6 +36,8 @@
     <xwiki.extension.name>Refactoring Application</xwiki.extension.name>
   </properties>
   <dependencies>
+    <!-- xwiki-platform-refactoring-api is a soft dependency: while the UI does not use it directly, having the
+     UI installed without the API is pointless. -->
     <dependency>
       <groupId>org.xwiki.platform</groupId>
       <artifactId>xwiki-platform-refactoring-api</artifactId>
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-ui/pom.xml
```diff
@@ -36,6 +36,12 @@
     <xwiki.extension.name>Refactoring Application</xwiki.extension.name>
   </properties>
   <dependencies>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-refactoring-api</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
```
mflorea, line 44: Where exactly is the Refactoring API used in the wiki pages from Refactoring UI? AFAICS the UI only has a configuration class, a configuration section and some translation keys, neither using the Refactoring API.

slauriere: You're right. I thought this could be conisdered as an implicit functional requirement, but it is irrelevant. I'm trying to identify a minimal requirement by such a module containing just pages, in order to solve the related upgrade issue, unless another path would be more appropriate. Would it be relevant to add a dependency on`xwiki-platform-web-war`? What do you think?

vmassol: Does it make sense to install refactoring-ui without refactoring-api? I think there's a dep between both even if it not an API dependency. For me it's fine to have this dep with a comment explaining the type of dep that it is maybe.

what do other committers think?

mflorea: I'm OK to keep the dependency if we add a comment to explain that it's not a hard dependency (it's not used directly) but a soft one (it doesn't make sense to have the UI installed without the API).

mflorea (approved):

PR #1628, opened 2021-05-05
XWIKI-18500: Extension updater propose to upgrade Sandbox to 13.1 when 12.10.5 flavor is installed

slauriere:
- Add dependency on `xwiki-platform-uiextension-ui`, which is needed by the Sandbox application to get listed in the Applications panel via a `UserExtensionClass` object.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-sandbox/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-sandbox/pom.xml
@@ -41,7 +41,6 @@
       <groupId>org.xwiki.platform</groupId>
       <artifactId>xwiki-platform-uiextension-api</artifactId>
       <version>${project.version}</version>
-      <type>xar</type>
     </dependency>
   </dependencies>
   <build>
--- a/xwiki-platform-core/xwiki-platform-sandbox/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-sandbox/pom.xml
@@ -39,7 +39,7 @@
     <dependency>
       <!-- Needed to display the Sandbox in the Applications panel list -->
       <groupId>org.xwiki.platform</groupId>
-      <artifactId>xwiki-platform-uiextension-ui</artifactId>
+      <artifactId>xwiki-platform-uiextension-api</artifactId>
       <version>${project.version}</version>
       <type>xar</type>
     </dependency>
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-sandbox/pom.xml
```diff
@@ -35,6 +35,15 @@
...
   </properties>
+  <dependencies>
+    <dependency>
+      <!-- Needed to display the Sandbox in the Applications panel list -->
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-uiextension-ui</artifactId>
+      <version>${project.version}</version>
+      <type>xar</type>
+    </dependency>
+  </dependencies>
```
mflorea, line 46: ``xwiki-platform-uiextension-ui`` only has a sheet which I doubt is used by the Sandbox application. The class is generated by ``UIExtensionClassDocumentInitializer`` from the API module.

slauriere: Wrong pick by me indeed, sorry, fixed in latest commit.

xwiki-platform-core/xwiki-platform-sandbox/pom.xml
```diff
@@ -35,6 +35,15 @@
     <!-- Name to display by the Extension Manager -->
     <xwiki.extension.name>Sandbox Application</xwiki.extension.name>
   </properties>
+  <dependencies>
+    <dependency>
+      <!-- Needed to display the Sandbox in the Applications panel list -->
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-uiextension-api</artifactId>
+      <version>${project.version}</version>
+      <type>xar</type>
```
mflorea, line 44:
```suggestion
```

Wrong type.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-sandbox/pom.xml
slauriere: Damn, sorry and thank you

mflorea (approved):

PR #1629, opened 2021-05-05
XWIKI-18479: Attachment count is not updated when the Attachments tab…

slauriere:
… is loaded

- Listen to event 'xwiki:docextra:loaded' and update attachment counter on attachment tab content loading completion

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
@@ -256,6 +256,16 @@ require(['jquery', 'xwiki-events-bridge'], function($) {
   $(document).on('xwiki:html5upload:done', function() {
     updateCount();
   });
+
+  /**
+   * Call "updateCount" when the attachment tab content gets loaded, in order to make sure that the counter
+   * reflects the actual number of attachments, which may have changed since the initial page load.
+   */
+  document.observe("xwiki:docextra:loaded", function(event) {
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
```diff
@@ -256,6 +256,16 @@ require(['jquery', 'xwiki-events-bridge'], function($) {
   $(document).on('xwiki:html5upload:done', function() {
     updateCount();
   });
+
+  /**
+   * Call "updateCount" when the attachment tab content gets loaded, in order to make sure that the counter
+   * reflects the actual number of attachments, which may have changed since the initial page load.
+   */
+  document.observe("xwiki:docextra:loaded", function(event) {
```
mflorea, line 264: Why use Prototype.js in a Require block with jQuery dependency? See a few lines above. See also https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/JavaScriptAPI/#HBridgingcustomXWikieventsbetweenPrototypeandjQuery .

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
slauriere: Indeed, a lack of observation by me, thank you for the review, the latest commit uses jQuery instead.

mflorea (approved):

PR #1631, opened 2021-05-06

slauriere:
- Use "justify-content: start" for the "extension-title" and "dependency-item" sections


slauriere:
This change allows to keep the extension name, version and optional namespace at the beginning of the line even in the case when there is no extension status (while with `justify-content: space-between` they were distributed evenly in the line, which resulted in having the version and the optional namespace pushed to the end, as illustrated by the before / after screen captures below).

## Before

before

## After

after

mflorea (approved):

PR #1632, opened 2021-05-06
XWIKI-18620: Replace the SendMailStatus live table with live data

manuelleduc:


xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-ui/src/main/resources/Mail/SendMailStatus.xml
```diff
@@ -76,32 +76,36 @@
...
-   'resultPage': 'Mail.SendMailStatusJSON',
-   'translationPrefix' : 'Mail.SendMailStatus_livetable.',
-   'tagCloud' : false,
-   'selectedColumn' : 'date',
-   'defaultOrder' : 'desc',
-   'outputOnlyHtml': true
+#set ($liveDataConfig = {
+  'meta': {
+    'propertyDescriptors': [
+      { 'id': 'date', 'displayer': 'text' },
```
manuelleduc, line 88: I didn't find a way to configure this column to have a working date date filter.

manuelleduc:
@mflorea this this the result with `useLiveData=true`

- the date filter is not working 
- the actions' html is displayed in raw

Screenshot_2021-05-06 XWiki - Mail - SendMailStatus

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-test/xwiki-platform-mail-test-pageobjects/src/main/java/org/xwiki/mail/test/po/MailStatusAdministrationSectionPage.java
```diff
@@ -42,12 +42,10 @@ public static MailStatusAdministrationSectionPage gotoPage()
...
-     * @return the status live table
+     * @return the status live data
      */
-    public LiveTableElement getLiveTable()
+    public LiveDataElement getLiveData()
     {
-        LiveTableElement lt = new LiveTableElement("sendmailstatus");
-        lt.waitUntilReady();
-        return lt;
+        return new LiveDataElement("sendmailstatus");
```
mflorea, line 49: Shouldn't you wait until the live data is ready?

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-ui/src/main/resources/Mail/SendMailStatus.xml
mflorea: WDYM by "working date filter"? The date filter is loading for me when using ``?useLiveData=true``. Sure, it doesn't filter the rows because the source (``Mail.SendMailStatusJSON``) doesn't support filtering on the date column, but that's not different than with the old live table. So you just need to make the date column un-filterable.

xwiki-platform-core/xwiki-platform-mail/xwiki-platform-mail-ui/src/main/resources/Mail/SendMailStatus.xml
```diff
@@ -76,32 +76,36 @@
...
+      { 'id': 'batchId', 'displayer': 'text' },
+      { 'id': 'type', 'displayer': 'text' },
+      { 'id': 'recipients', 'displayer': 'html' },
+      { 'id': 'state', 'displayer': 'text' },
+      { 'id': 'wiki', 'displayer': 'text' },
+      { 'id': 'error', 'displayer': 'text' },
```
mflorea, line 94: ``text`` displayer is used by default so there's no need to specify it. Moreover, there's no need to specify a property descriptor if you don't have any (custom) configuration for it, so most of the property descriptors above can be omitted.

mflorea (changes requested):

mflorea:
> the actions' html is displayed in raw

@manuelleduc that's because of the live table configuration:

> 'actions' : { 'type' : 'text', 'html' : **'true'**, 'filterable' : 'false', 'sortable' : 'false' }

it should have been:

> 'actions' : { 'type' : 'text', 'html' : **true**, 'filterable' : 'false', 'sortable' : 'false' }

(unquoted) because the value should be boolean. Anyway, using the live data macro won't have this problem. It affects only the automatic conversion.

mflorea (commented):
Live data configuration looks good.

PR #1633, opened 2021-05-06

slauriere:
- Adapt tests in xwiki-platform-extension-test to the updated HTML
  structure of the extension / extension dependencies panes

### DIFF
--- a/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/DependencyPane.java
+++ b/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/DependencyPane.java
@@ -92,7 +92,7 @@ public class DependencyPane extends BaseElement
         }
         // Unknown dependency, with no link.
         String innerText = container.getText();
-        return innerText.substring(0, innerText.indexOf(getVersion()));
+        return innerText.substring(0, innerText.indexOf(getVersion())).trim();
     }
 
     /**
--- a/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/DependencyPane.java
+++ b/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/DependencyPane.java
@@ -86,13 +86,7 @@ public class DependencyPane extends BaseElement
      */
     public String getName()
     {
-        WebElement link = getLink();
-        if (link != null) {
-            return link.getText();
-        }
-        // Unknown dependency, with no link.
-        String innerText = container.getText();
-        return innerText.substring(0, innerText.indexOf(getVersion())).trim();
+        return getDriver().findElementWithoutWaiting(container, By.className("extension-name")).getText();
     }
 
     /**
--- a/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/ExtensionPane.java
+++ b/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/ExtensionPane.java
@@ -187,10 +187,11 @@ public class ExtensionPane extends BaseElement
      */
     private String getXPath()
     {
-        String nameAndVersion =
-            getDriver().findElementWithoutWaiting(container, By.className("extension-title")).getText();
+        String name = getDriver().findElementWithoutWaiting(container, By.className("extension-name")).getText();
+        String version = getDriver().findElementWithoutWaiting(container, By.className("extension-version")).getText();
         return String.format("//form[contains(@class, 'extension-item') and descendant::*["
-            + "contains(@class, 'extension-title') and normalize-space(.) = '%s']]", nameAndVersion);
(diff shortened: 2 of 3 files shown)


slauriere:
A `trim` was introduced in [`DependencyPane#getName`](https://github.com/xwiki/xwiki-platform/pull/1633/commits/64c4ea5944e90c2151dbe27399b1268a7bbf71b1#diff-3ca1ed632f1598afe4fb5ca342beab93049c28fc85d4c1997c69291c39d16f4b): that's because when using `display: flex`, `WebElement#getText()` can include line feeds, which are absent with the default `display`, it seems.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/DependencyPane.java
```diff
@@ -92,7 +92,7 @@ public String getName()
         }
         // Unknown dependency, with no link.
         String innerText = container.getText();
-        return innerText.substring(0, innerText.indexOf(getVersion()));
+        return innerText.substring(0, innerText.indexOf(getVersion())).trim();
```
mflorea, line 95: This feels overly complicated. Maybe there was no ``extension-name`` wrapper when this method was written but we have it now so we should get the extension name from there. This method should have a single line, like the next one (``getVersion``). Can you make this change?

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/ExtensionPane.java
```diff
@@ -187,10 +187,11 @@ private ExtensionPane clickAndWaitUntilElementIsVisible(WebElement button, Strin
...
     private String getXPath()
     {
-        String nameAndVersion =
-            getDriver().findElementWithoutWaiting(container, By.className("extension-title")).getText();
+        String name = getDriver().findElementWithoutWaiting(container, By.className("extension-name")).getText();
+        String version = getDriver().findElementWithoutWaiting(container, By.className("extension-version")).getText();
         return String.format("//form[contains(@class, 'extension-item') and descendant::*["
-            + "contains(@class, 'extension-title') and normalize-space(.) = '%s']]", nameAndVersion);
+            + "contains(@class, 'extension-name') and normalize-space(.) = '%s'] and descendant::*["
+            + "contains(@class, 'extension-version') and normalize-space(.) = '%s']]", name, version);
```
mflorea, line 194: This is not good because it can match a **dependency** name or version. You need to limit the scope of the search, e.g. by looking only in the ``extension-header`` or ``extension-title``.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/SearchResultsPane.java
```diff
@@ -71,10 +71,10 @@ public String getNoResultsMessage()
      */
     public ExtensionPane getExtension(String name, String version)
     {
-        String nameAndVersion = name + " " + version;
-        By xpath =
-            By.xpath("//form[contains(@class, 'extension-item') and descendant::*[contains(@class, "
-                + "'extension-title') and normalize-space(.) = '" + nameAndVersion + "']]");
+        String selector = String.format("//form[contains(@class, 'extension-item') and descendant::*["
+            + "contains(@class, 'extension-name') and normalize-space(.) = '%s'] and descendant::*["
+            + "contains(@class, 'extension-version') and normalize-space(.) = '%s']]", name, version);
```
mflorea, line 76: Same as above, you could match a dependency, so you need to restrict the search scope.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/ExtensionPane.java
slauriere: Indeed. I just tried to fix this but XPath is new to me and I couldn't find the correct syntax so far. Can I ask you what would be a possible syntax for limiting the scope?

I tried this but there's something wrong with the `descendant` I guess and couldn't find much documentation yet.

```
return String.format("//form[contains(@class, 'extension-item') and descendant::*["
 + "*[@class = 'extension-header']//*[@class = 'extension-name'] and normalize-space(.) = '%s'] and "
 + "descendant::*[*[@class = 'extension-header']//*[@class = 'extension-version'] and normalize-space(.) = '%s']]",
```

manuelleduc: I have the feeling this complex xpath expression could be replaced by a simpler css selector. 
I haven't looking in more details at the code calling `getXPath` but that could be a way to solve your issue.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-test/xwiki-platform-extension-test-pageobjects/src/main/java/org/xwiki/extension/test/po/DependencyPane.java
slauriere: Indeed, updated in the latest commit

mflorea (approved):

PR #1636, opened 2021-05-19
XWIKI-18653: The naming strategies cache is only reloaded when using the administration UI

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18653

Introduces a listener to update the model validation cache on change

### DIFF
--- a/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/pom.xml
+++ b/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/pom.xml
@@ -30,7 +30,7 @@
   <description>XWiki Platform - Model Validation API to handle model validation in XWiki</description>
   <packaging>jar</packaging>
   <properties>
-    <xwiki.jacoco.instructionRatio>1.00</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.97</xwiki.jacoco.instructionRatio>
     </properties>
   <dependencies>
     <dependency>
--- a/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/EntityNameValidation.java
+++ b/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/EntityNameValidation.java
@@ -67,4 +67,14 @@ public interface EntityNameValidation
      * @return {@code true} if the policy is respected.
      */
     boolean isValid(EntityReference entityReference);
+
+    /**
+     * Clean the configuration cache of a wiki.
+     * @param wikiId the identifier of the wiki's cache to clean (e.g., {@code "wikiA"})
+     * @since 13.4
+     */
+    default void cleanConfigurationCache(String wikiId)
+    {
+        // Does nothing by default.
+    }
 }
--- a/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/EntityNameValidationManager.java
+++ b/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/EntityNameValidationManager.java
@@ -19,6 +19,7 @@
  */
 package org.xwiki.model.validation;
 
+import java.util.List;
 import java.util.Set;
 
 import org.xwiki.component.annotation.Role;
@@ -51,7 +52,8 @@ public interface EntityNameValidationManager
     Set<String> getAvailableEntityNameValidations();
 
     /**
-     * Allow to reset the configuration of the strategies.
+     * @return the list of available {@link EntityNameValidation} components.
+     * @since 13.4
      */
-    void resetStrategies();
+    List<EntityNameValidation> getAvailableEntityNameValidationsComponents();
 }
--- a/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/DefaultEntityNameValidationManager.java
+++ b/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/DefaultEntityNameValidationManager.java
@@ -20,10 +20,10 @@
 package org.xwiki.model.validation.internal;
 
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 
 import javax.inject.Inject;
-import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.slf4j.Logger;
@@ -51,14 +51,7 @@ public class DefaultEntityNameValidationManager implements EntityNameValidationM
     private EntityNameValidationConfiguration entityNameValidationConfiguration;
 
     @Inject
-    private ReplaceCharacterEntityNameValidationConfiguration replaceCharacterEntityNameValidationConfiguration;
-
-    @Inject
     private Logger logger;
-
-    @Inject
-    @Named(ReplaceCharacterEntityNameValidation.COMPONENT_NAME)
-    private EntityNameValidation replaceCharacterValidator;
 
     @Override
     public EntityNameValidation getEntityReferenceNameStrategy()
@@ -89,10 +82,13 @@ public class DefaultEntityNameValidationManager implements EntityNameValidationM
     }
 
     @Override
-    public void resetStrategies()
+    public List<EntityNameValidation> getAvailableEntityNameValidationsComponents()
     {
-        ((ReplaceCharacterEntityNameValidation) this.replaceCharacterValidator)
-            .setReplacementCharacters(this.replaceCharacterEntityNameValidationConfiguration
-                .getCharacterReplacementMap());
+        try {
+            return this.componentManager.getInstanceList(EntityNameValidation.class);
+        } catch (ComponentLookupException e) {
+            this.logger.error("Error while getting the instance list of the EntityNameValidation", e);
+            return Collections.emptyList();
+        }
     }
 }
--- a/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationListener.java
+++ b/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationListener.java
@@ -1,0 +1,118 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.model.validation.internal;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.xwiki.bridge.event.DocumentCreatedEvent;
+import org.xwiki.bridge.event.DocumentDeletedEvent;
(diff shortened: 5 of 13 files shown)


xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/pom.xml
```diff
@@ -30,7 +30,7 @@
   <description>XWiki Platform - Model Validation API to handle model validation in XWiki</description>
   <packaging>jar</packaging>
   <properties>
-    <xwiki.jacoco.instructionRatio>1.00</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.97</xwiki.jacoco.instructionRatio>
```
manuelleduc, line 33: Decreased due to the introduction of `cleanConfigurationCache`, a default method with an empty body.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/DefaultEntityNameValidationManager.java
```diff
@@ -89,10 +82,13 @@ public EntityNameValidation getEntityReferenceNameStrategy(String hint)
     }
 
     @Override
-    public void resetStrategies()
+    public List<EntityNameValidation> getAvailableEntityNameValidationsComponents()
     {
-        ((ReplaceCharacterEntityNameValidation) this.replaceCharacterValidator)
-            .setReplacementCharacters(this.replaceCharacterEntityNameValidationConfiguration
-                .getCharacterReplacementMap());
```
manuelleduc, line 96: Removed because it is not needed now that we monitor directly the documents.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationSource.java
```diff
@@ -43,11 +43,9 @@
 {
     private static final List<String> SPACE_NAMES = Arrays.asList("XWiki", "EntityNameValidation");
 
-    private static final LocalDocumentReference CLASS_REFERENCE =
-        new LocalDocumentReference(SPACE_NAMES, "ConfigurationClass");
+    static final LocalDocumentReference CLASS_REFERENCE = new LocalDocumentReference(SPACE_NAMES, "ConfigurationClass");
 
-    private static final LocalDocumentReference DOC_REFERENCE =
-        new LocalDocumentReference(SPACE_NAMES, "Configuration");
+    static final LocalDocumentReference DOC_REFERENCE = new LocalDocumentReference(SPACE_NAMES, "Configuration");
```
manuelleduc, line 48: Scope changed to reuse the references in `EntityNameValidationConfigurationListener`.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/pom.xml
vmassol: Note that default methods should ideally be tested too ;)

manuelleduc: I agree but I'm not sure how to test a method with an empty body :/

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/EntityNameValidation.java
```diff
@@ -67,4 +67,16 @@
...
     boolean isValid(EntityReference entityReference);
+
+    /**
+     * Clean the configuration cache of a wiki.
+     *
+     * @param wikiId the identifier of the wiki's cache to clean (e.g., {@code "wikiA"})
+     * @since 13.4
+     * @since 12.10.8
+     */
+    default void cleanConfigurationCache(String wikiId)
```
vmassol, line 78: hmm this doesn't look very good to me. This seems to be a management API and that should go in a JMX MBean. It's definitely wrong to me to have that in a "business" interface such as EntityNameValidation.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/pom.xml
vmassol: Indeed :) I've replied at https://github.com/xwiki/xwiki-platform/pull/1636/files#r635320431

tmortagne: > but I'm not sure how to test a method with an empty body :/

Not a very useful test, but I guess all it takes in terms of coverage is to call it in some test.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationListener.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Creates a {@link RegexEventFilter} matching a local reference on any wiki. Schematically, for a local reference
+     * {@code XWiki.Doc}, this method created a {@link RegexEventFilter} with the following regex: {@code
+     * ".*:XWiki\.Doc"}.
+     *
+     * @param localDocumentReference the local document reference to filter on
+     * @return the even filter matching the local document reference events
+     */
+    private RegexEventFilter createRegexEventFilter(LocalDocumentReference localDocumentReference)
```
tmortagne, line 111: You seems to more or less reinvent `BaseObjectReference#any` here.

tmortagne, line 111: Maybe this helper should be moved in a more reusable location since it looks like a very generic need.

Having a more specific method name like `anyLocalDocument` might help readability (as you probably noticed from my previous deleted comments :) )

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/EntityNameValidationManager.java
```diff
@@ -51,7 +52,9 @@
     Set<String> getAvailableEntityNameValidations();
 
     /**
-     * Allow to reset the configuration of the strategies.
+     * @return the list of available {@link EntityNameValidation} components.
+     * @since 13.4
+     * @since 12.10.8
      */
-    void resetStrategies();
+    List<EntityNameValidation> getAvailableEntityNameValidationsComponents();
```
surli, line 59: Not sure to understand the purpose of this method, you could achieve the same by calling the component manager and getting the list of components of type `EntityNameValidation` no?

surli: AFAICS that's even exactly your implem so I don't see why it needs to be declared in the API.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationListener.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Override
+    public List<Event> getEvents()
+    {
+        LocalDocumentReference classReference = EntityNameValidationConfigurationSource.CLASS_REFERENCE;
+        LocalDocumentReference documentReference = EntityNameValidationConfigurationSource.DOC_REFERENCE;
+        RegexEventFilter configurationClassEventFilter = createRegexEventFilter(classReference);
+        RegexEventFilter configurationObjectEventFilter = createRegexEventFilter(documentReference);
+        return Arrays.asList(
```
surli, line 84: I don't understand why you're listening on those events, IMO you could just listen on the `XObjectAdded/Updated/DeletedEvent` with a `XWiki.EntityNameValidation.ConfigurationClass` xobject reference.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/ReplaceCharacterEntityNameValidation.java
```diff
@@ -42,58 +42,82 @@
...
      * Name of the component.
      */
     protected static final String COMPONENT_NAME = "ReplaceCharacterEntityNameValidation";
 
-    private String[] forbiddenCharacters;
-    private String[] replacementCharacters;
+    /**
+     * A map used as a cache of the current configurations of the wikis.
+     */
+    private final Map<String, Configuration> wikiConfigurationCache = new ConcurrentHashMap<>();
```
surli, line 64: Interesting, contrary than before you're allowing to have a different configuration for each wiki. It makes sense since right now I think we display the administration page in all subwikis, now I'm wondering if that's something we really want, or if we'd want to be able to configure it on the farm only: for example, in myxwiki.org you don't want the admin of a wiki to override the global farm entity name validation since it's based on some server restrictions in general.

manuelleduc: Interesting. In this case we might want to hide the configuration in subwikis (or at least make them edit the configuration located in the main wiki + relevant rights).
For now, after any wikis edit its configuration, it becomes the global one which is very confusing.

surli: > For now, after any wikis edit its configuration, it becomes the global one which is very confusing.

That's clearly a bug and it could make sense to report it, so we can discuss how we want to fix it there.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationListener.java
surli: > The rational is to listen to the single class and object of this configuration (does not make sens to invalidate on an XObject that will never be read), on all the wikis. 

Actually what I don't get is why you're listening on the configuration class document? Even if this one changes it doesn't make the config itself changing.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/ReplaceCharacterEntityNameValidation.java
manuelleduc: https://jira.xwiki.org/browse/XWIKI-18670

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-default/src/main/java/org/xwiki/model/validation/internal/EntityNameValidationConfigurationListener.java
manuelleduc: In a minimal configuration, the XObject can be missing, or one of the properties left blank. In this case is it interesting to rely on the XClass defaults values.
This is the can on docker tests with a minimal distribution for instance.

surli: bq. In this case is it interesting to rely on the XClass defaults values.

well maybe I missed the change but the default values are empty in the xclass, that's the configuration components implementation which gives default values.

manuelleduc: My mistake, I don't know why I though XClass properties could define default values... Removing.

surli: > Could be interesting to add two keys in xwiki.properties to provide default values for the configuration, That would also make testing easier.

That's actually an option that I'm thinking about to actually fix the problem I raised in my comment on https://jira.xwiki.org/browse/XWIKI-18670?focusedCommentId=109158&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-109158

PR #1637, opened 2021-05-21
XWIKI-18628: Field pretty name should not take precedence over the specified translation prefix

manuelleduc:
Update the livetable data configuration resolver to apply the following precedence for the property names:
- name defined in the Live Data macro
- name defined in the translations
- pretty name of the property
- identifier of the property

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
@@ -49,7 +49,7 @@ import org.xwiki.localization.ContextualLocalizationManager;
 
 /**
  * Adds missing live data configuration values specific to the live table source.
- * 
+ *
  * @version $Id$
  * @since 12.10.4
  * @since 13.0
@@ -80,7 +80,7 @@ public class DefaultLiveDataConfigurationResolver implements LiveDataConfigurati
     /**
      * Used to merge the default configuration with the provided configuration.
      */
-    private JSONMerge jsonMerge = new JSONMerge();
+    private final JSONMerge jsonMerge = new JSONMerge();
 
     @Override
     public LiveDataConfiguration resolve(LiveDataConfiguration config) throws LiveDataException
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
```diff
@@ -199,12 +199,26 @@ private LiveDataPropertyDescriptor getDefaultPropertyDescriptor(String property)
         return propertyDescriptor;
     }
 
-    private LiveDataConfiguration translate(LiveDataConfiguration config)
+    /**
+     * Updates the {@code mergedConfig} with translated property names and descriptions.
+     *
+     * @param mergedConfig the configuration to update
+     * @param config the base configuration of the live data macro
```
mflorea, line 206:
```suggestion
     * @param config the live data configuration provided by the user through the live data macro
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
```diff
@@ -199,12 +199,26 @@ private LiveDataPropertyDescriptor getDefaultPropertyDescriptor(String property)
...
-        String translationPrefix = (String) config.getQuery().getSource().getParameters().get("translationPrefix");
-        for (LiveDataPropertyDescriptor property : config.getMeta().getPropertyDescriptors()) {
-            if (property.getName() == null) {
-                property.setName(this.l10n.getTranslationPlain(translationPrefix + property.getId()));
+        String translationPrefix =
+            (String) mergedConfig.getQuery().getSource().getParameters().get("translationPrefix");
+        for (LiveDataPropertyDescriptor property : mergedConfig.getMeta().getPropertyDescriptors()) {
+            // We set a name only if the property does not already have a name or if it does not have a default 
+            // name.
+            if (property.getName() == null || !hasDefaultName(config, property.getId())) {
```
mflorea, line 216:
```suggestion
            // If the property name is not set then we default on the configured translation key or the property id.
            // Otherwise, if the property name is set but not by the user (i.e. the name comes from the default source
            // configuration) then we want to give priority to the configured translation key if available.
            if (property.getName() == null || !hasSpecifiedName(config, property.getId())) {
```

mflorea (changes requested):
Two small suggestions to make things clear.

PR #1638, opened 2021-05-25
XWIKI-18676: Missing placeholder for the first load of a LD table

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18676

Displays a loader:
1. in the Live Data macro: before vue and its dependencies are loaded client side
2. after vue has replaced the placeholder provided by the Live Data macro by its own template, and before the first layout and its dependencies are loaded client side

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
@@ -327,6 +327,7 @@ dl.export-tree-legend dt + dd {
 // Loading --------------------------------------------------------------------
 .loading {
   background: transparent url(@xwiki-icon-spinner) no-repeat center;
+  min-height: @xwiki-icon-spinner-height;
 }
 
 /* Make sure the loading animation is visible when a tab is loaded asynchronously. */
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
@@ -330,11 +330,6 @@ dl.export-tree-legend dt + dd {
   min-height: @xwiki-icon-spinner-height;
 }
 
-/* Make sure the loading animation is visible when a tab is loaded asynchronously. */
-.tab-content > .tab-pane.loading {
-  min-height: 16px;
-}
-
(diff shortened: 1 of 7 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/java/org/xwiki/livedata/internal/macro/LiveDataMacro.java
```diff
@@ -116,9 +123,11 @@ public LiveDataMacro()
         // Load the JavaScript code of the Live Data widget.
         Map<String, Object> skinExtensionParameters = Collections.singletonMap("forceSkinAction", Boolean.TRUE);
         this.jsfx.use("uicomponents/widgets/liveData.js", skinExtensionParameters);
+        // Load the CSS code of the Live Data widget.
+        this.ssfx.use("uicomponents/widgets/liveData.css");
```
mflorea, line 127: The loading styles should be generic so I would keep them in the skin. Moreover, I don't see what else we could add to ``liveData.css`` in the future to justify its existence beyond the live data loading animation. The minimum height is an issue that needs to be fixed for any loading block: it doesn't make sense to use a spinner icon without making sure the container has enough height to show it, so IMO this needs to be fixed in [``misc.less``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less#L327). Basically we need to drop the custom ``.tab-content > .tab-pane.loading`` fix and make it generic for any ``.loading``.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -327,6 +327,7 @@ dl.export-tree-legend dt + dd {
 // Loading --------------------------------------------------------------------
 .loading {
   background: transparent url(@xwiki-icon-spinner) no-repeat center;
+  min-height: @xwiki-icon-spinner-height;
```
manuelleduc, line 330: I've introduced a new less variable to configure the spinner height. It is initialized to the height of the current spinner icon and needs to be synchronized if a new icon if `@xwiki-icon-spinner` is changed.
I guess the `.tab-content > .tab-pane.loading` block just below can be removed, but I'm not sure if it should be kept for legacy compatibility.

mflorea: > I guess the .tab-content > .tab-pane.loading block just below can be removed, but I'm not sure if it should be kept for legacy compatibility.

+1 to remove it.

mflorea (approved):

PR #1640, opened 2021-06-07
Update README.md

kitarp29:
Hey @piis3 
Sorry, but the Readme file had one grammar mistake.
I took care of it in this.
Sorry for making a PR for such a small thing but this was the first mistake I found out.
I am figuring my way around the Repo.
Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [XWiki Platform](https://www.xwiki.org/xwiki/bin/view/Documentation/) is a generic wiki platform offering runtime services for applications built on top of it.
 
-XWiki Commons, XWiki Rendering and XWiki Platform are part of the [XWiki.org](http://www.xwiki.org/) software forge. They are released together and share the same version.
+XWiki Commons, XWiki Rendering, and XWiki Platform are part of the [XWiki.org](http://www.xwiki.org/) software forge. They are released together and share the same version.
 
 ## Documentation
 * [API](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/API/)


piis3:
Hello,
I'm not sure why you included me on this, I just contributed a small patch to the project years ago. Thanks though.

kitarp29:
Sorry I tagged the people who are active.

vmassol:
@kitarp29 thx. Please try to have the commit message be as descriptive as possible. In this case I would have used for example "Fix grammar mistake" instead of "update README.md". It's more precise.

README.md
```diff
@@ -2,7 +2,7 @@
 
 [XWiki Platform](https://www.xwiki.org/xwiki/bin/view/Documentation/) is a generic wiki platform offering runtime services for applications built on top of it.
 
-XWiki Commons, XWiki Rendering and XWiki Platform are part of the [XWiki.org](http://www.xwiki.org/) software forge. They are released together and share the same version.
+XWiki Commons, XWiki Rendering, and XWiki Platform are part of the [XWiki.org](http://www.xwiki.org/) software forge. They are released together and share the same version.
```
acotiuga, line 5: Why is this a grammar mistake? IIRC both `XWiki Commons, XWiki Rendering and XWiki Platform ` and `XWiki Commons, XWiki Rendering, and XWiki Platform ` are correct. I would rather consider that the comma is not needed in a list, before `and`.

mflorea: @vmassol: @acotiuga is right. The comma before AND is **optional** and really not needed in this case. See https://www.grammarly.com/blog/comma-before-and/ .

vmassol: Thanks @mflorea I had actually checked it and found this link too but all I could extract from it is that both are allowed. We can revert if you prefer, I have no problem with that.

PR #138, opened 2013-08-01
XWIKI-6769 : Login is not protected against CSRF attacks

tdelafosse:

PR #1641, opened 2021-06-08
XWIKI-6769: Login is not protected against CSRF attacks

aemi-dev:
- Updated code from [old XWIKI-6769 PR](https://github.com/xwiki/xwiki-platform/pull/138) to make it work with latest version of XWiki


xwiki-platform-core/xwiki-platform-csrf/src/main/java/org/xwiki/csrf/internal/DefaultCSRFToken.java
```diff
@@ -193,9 +197,14 @@ public String getResubmissionURL()
 
         // back URL is the URL of the document that was about to be modified, so in most
         // cases we can redirect back to the correct document (if the user clicks "no")
-        String backUrl = getDocumentURL(this.docBridge.getCurrentDocumentReference(), null);
+        DocumentReference currentDocRef = this.docBridge.getCurrentDocumentReference();
+        // If the current document is null (on the login page for example), let's redirect to the default page
+        if (currentDocRef == null) {
+            currentDocRef = resolver.resolve("");
```
vmassol, line 203: Note 1: You can always resolve any reference, including null so there's no need for an IF. We should mention it in the javadoc.
Note 2: getDocumentURL() should work with null already. I think there's a bug in `DAB#getDocumentURL` which leads to an infinite loop if `this.getContext().getDoc().getDocumentReference()` is null. We need to fix that IMO but I'm not sure what behavior we should have then. Maybe do as you did and go to the default page.

@tmortagne WDYT?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyFormAuthenticator.java
```diff
@@ -185,6 +188,16 @@ public boolean processLogin(SecurityRequestWrapper request, HttpServletResponse
 
         // process login form submittal
         if ((this.loginSubmitPattern != null) && request.getMatchableURL().endsWith(this.loginSubmitPattern)) {
+
+            // Placed here to not interfere with persistentLoginManager
```
vmassol, line 192: I find this comment hard to understand (I don't understand it). Could you explain in comment what is the problem?

xwiki-platform-core/xwiki-platform-csrf/src/main/java/org/xwiki/csrf/internal/DefaultCSRFToken.java
```diff
@@ -193,9 +197,14 @@ public String getResubmissionURL()
...
+        }
+        String backUrl = getDocumentURL(currentDocRef, null);
         query += "&xback=" + urlEncode(backUrl);
-
+
```
manuelleduc, line 207: I know this kind of changes are easy to introduce by accident, but the PRs and commit diffs are easier to read when there is no other changes than the one strictly related to the PR :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/user/impl/xwiki/MyFormAuthenticator.java
tmortagne: Would indeed be nice to elaborate a bit on what kind of problem you are avoiding by putting it here. Note that since this is the location where we read user/pass parameters, this is the natural location to put that token validation, so I don't think I would have put a comment explaining why it's here in the first place.

aemi-dev: In fact, working with the old PR, the token verification was placed at the top of the method. Doing it so caused the verification to fail when persistentLoginManager was intended to be used. Since, it's the natural location you may be right and I should remove this comment.

xwiki-platform-core/xwiki-platform-csrf/src/main/java/org/xwiki/csrf/internal/DefaultCSRFToken.java
aemi-dev: Actually, I can't pass `currentDocRef` to the resolver because it can only accept strings and even if the DocumentReference can be converted to string, Java doesn't do it implicitly. By the way, I can only pass null or empty string to get the current behaviour.

xwiki-platform-core/xwiki-platform-csrf/src/main/java/org/xwiki/csrf/internal/DefaultCSRFToken.java
```diff
@@ -191,11 +195,15 @@ public String getResubmissionURL()
     {
         String query = "resubmit=" + urlEncode(getRequestURI());
 
-        // back URL is the URL of the document that was about to be modified, so in most
+        // backURL is the URL of the document that was about to be modified, so in most
         // cases we can redirect back to the correct document (if the user clicks "no")
-        String backUrl = getDocumentURL(this.docBridge.getCurrentDocumentReference(), null);
+        DocumentReference currentDocRef = this.docBridge.getCurrentDocumentReference();
+        // If the current document is null (on the login page for example), let's redirect to the default page
```
surli, line 201: Wouldn't be cleaner to get the default page reference by accessing the current wiki descriptor? Something like: 
```
@Inject
private WikiDescriptorManager wikiDescriptorManager;

....

this.wikiDescriptorManager.getCurrentWikiDescriptor().getMainPageReference();
```

surli (commented):
Sounds good to me, but have you checked back the integration tests? AFAIR there was some issues when running them.

xwiki-platform-core/xwiki-platform-csrf/src/main/java/org/xwiki/csrf/internal/DefaultCSRFToken.java
aemi-dev: I'll try that.

tmortagne: Yes, being redirected to the current wiki home page will probably be less of a WTF than ending up on a completely different wiki :)

tmortagne:
Same here, the general thing looks ok, but without some update of TestUtils it will most probably break most of the UI tests.

Enygma2002:
Please also consider my [comment on the issue](https://jira.xwiki.org/browse/XWIKI-6769?focusedCommentId=109826&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-109826), as "fixing" this would break an existing valid usecase (that has no workaround that I know of) without really bringing much in terms of security.

PR #1642, opened 2021-06-08
XWIKI-14947: Allow EntityResourceReferenceHandler to call XWikiActions

surli:
## JIRA

* https://jira.xwiki.org/browse/XWIKI-14947
* https://jira.xwiki.org/browse/XWIKI-15091

## Description

The goal of this PR is to finish the work on resource module to
allow using a `ResourceReferenceHandler` for URL such as /bin/action. It
was not possible until now to have a fallback on `XWikiAction`: this works
allows to keep the fallback, but also to have new resource reference
handler for new /bin/ actions.
The main part of the work consists in modifying
`ResourceReferenceHandlerManager#canHandle` to be able to check the full
`ResourceReference`: that way it allows the
`MainResourceReferenceHandlerManager` to match an `EntityResourceReference`
only if there's an appropriate `ResourceReferenceHandler` for it, which is
done by asking to the `EntityResourceReferenceHandlerManager`. If we don't
perform this check in `MainResourceReferenceHandler`, we cannot properly
fallback on `XWikiAction`.

This PR also contains a fix for XWIKI-15091 related to using `WikiReferenceExtractor` when the context is not initialized since it was preventing using properly the `ResourceReferenceHandler`. 

Note that the main thing to check on this PR is the backward compatibility, in particular with the new API defined in `ResourceReferenceHandler`.


xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-default/src/main/java/org/xwiki/resource/internal/AbstractResourceReferenceHandlerManager.java
```diff
@@ -88,10 +88,17 @@ public void handle(ResourceReference reference) throws ResourceReferenceHandlerE
 
     @Override
     public boolean canHandle(T resourceReferenceQualifier)
+    {
+        // FIXME: this is not good, it means we don't support backward compatibility.
```
surli, line 92: Reviewers: that's the part I'm not a big fan of. I see two choices to fix it:
  1. we have two branches in the code, one using the full `ResourceReference` to check the handler (which is the current implem you can see below), and another one using the old implem which only uses the type `T` by extracting the ResourceReferenceQualifier. It probably means a bit of duplicating code, for something which I'm not even sure is implemented somewhere (we're talking about an implem of `ResourceReferenceHandlerManager` here, not just an handler)
  2. We break the API of `ResourceReferenceHandlerManager` to remove entirely `#canHandle(T resourceReferenceQualifier)`, maybe we'll break someone (I doubt it), but at least it's clear what we're doing and we don't need to maintain two artificial branches.

tmortagne: When strictly speaking about retro compatibility another thing to worry about would not so much be the implementation of #canHandle(T) (since this method would never been called anymore by XWiki Standard) but the default implementation of the new `ResourceReferenceHandlerManager#canHandle(ResourceReference)` which is close to useless right now since it will never have the expected behavior for a ResourceReferenceHandlerManager provided by some extension since this extension will only implement the old one.

That being said, I'm pretty sure the chance that anyone implemented or is planning to implement `ResourceReferenceHandlerManager` in an extension is very close to 0.

Back to your question: the chance that anyone directly use ResourceReferenceHandlerManager at all in an extension is most probably very close to 0 too, so I'm fine with `2.`

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-default/src/main/java/org/xwiki/resource/internal/MainResourceReferenceHandlerManager.java
```diff
@@ -37,10 +41,19 @@
...
+    private ResourceReferenceHandlerManager<EntityResourceAction> entityResourceReferenceHandlerManager;
+
     @Override
-    protected boolean matches(ResourceReferenceHandler handler, ResourceType resourceType)
+    protected boolean matches(ResourceReferenceHandler handler, ResourceReference reference)
     {
-        return handler.getSupportedResourceReferences().contains(resourceType);
+        boolean handlerResourceSupported = handler.getSupportedResourceReferences()
+            .contains(extractResourceReferenceQualifier(reference));
+        if (handlerResourceSupported && handler instanceof EntityResourceReferenceHandler) {
```
vmassol, line 52: This is not good IMO. The MainResourceReferenceHandlerManager should not know about EntityResourceReferenceHandler

surli: I was expecting this comment, but I've no idea how to avoid that check if we want `#canHandle` to properly return false right away and not when `#handle` is called.

vmassol: yes it means we need to review the design and possibly change it.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-schemes/xwiki-platform-url-scheme-standard/src/main/java/org/xwiki/url/internal/standard/AbstractWikiReferenceExtractor.java
```diff
@@ -46,6 +48,11 @@
     @Inject
     private Execution execution;
 
+    protected boolean isContextAvailable()
+    {
+        return execution.getContext() != null && execution.getContext().hasProperty(XWikiContext.EXECUTIONCONTEXT_KEY);
```
tmortagne, line 53: Hmm feels like this code was not really meant to be aware of XWikiContext, I wonder if all this bulletproofing should not be moved to WikiDescriptorManager.

PR #1646, opened 2021-06-17
XWIKI-18732: Live Data source should be able to display contextual error/warning messages

manuelleduc:
Introduce a DocTitle component dedicated to displaying the doc.title property.
Introduce a Footnotes component to display footnotes
Introduce the logic to check if an entry is empty because the current user does not have the right to view it

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
@@ -47,7 +47,8 @@
       {"id": "html"},
       {"id": "xObjectProperty"},
       {"id": "actions"},
-      {"id": "boolean"}
+      {"id": "boolean"},
+      {"id": "docTitle"}
     ],
 
     "defaultDisplayer": "text",
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
@@ -45,10 +45,8 @@
       {"id": "text"},
       {"id": "link"},
       {"id": "html"},
-      {"id": "xObjectProperty"},
       {"id": "actions"},
-      {"id": "boolean"},
-      {"id": "docTitle"}
+      {"id": "boolean"}
     ],
 
     "defaultDisplayer": "text",
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
@@ -177,6 +177,7 @@
 
     "displayers": [
       {"id": "link", "propertyHref": "doc.url"},
+      {"id": "xObjectProperty"},
       {"id": "docTitle", "propertyHref": "doc.url"}
     ],
 
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
@@ -12,7 +12,7 @@
 
     "propertyDescriptors": [
       {"id": "doc.name", "type": "String", "editable": false, "displayer": "link"},
-      {"id": "doc.title", "type": "String", "displayer": "link"},
+      {"id": "doc.title", "type": "String", "displayer": "docTitle"},
       {
         "id": "doc.hidden",
         "type": "Boolean",
@@ -176,7 +176,8 @@
     ],
 
     "displayers": [
-      {"id": "link", "propertyHref": "doc.url"}
+      {"id": "link", "propertyHref": "doc.url"},
+      {"id": "docTitle", "propertyHref": "doc.url"}
     ],
 
     "filters": [
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
@@ -106,7 +106,7 @@ livedata.panel.sort.direction.descending=Descending
 livedata.panel.sort.add=Add sort
 livedata.panel.sort.delete=Delete sort
 
-livedata.displayer.emptyValue=N/A
+livedata.displayer.emptyValue=N/A<sup>*</sup>
 
 livedata.displayer.link.noValue=(no value)
 
@@ -119,4 +119,4 @@ livedata.displayer.xObjectProperty.failedToRetrieveField.errorMessage=Failed to 
 livedata.filter.list.emptyLabel=---
 
 livedata.footnotes.computedTitle=Some pages have a computed title. Filtering and sorting by title will not work as expected for these pages.
-livedata.footnotes.propertyNotViewable=Displayed when some pages require special rights to be viewed and thus fields cannot be extracted from them.
+livedata.footnotes.propertyNotViewable=Displayed when some entries require special rights to be viewed and thus fields cannot be extracted from them.
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
@@ -106,6 +106,8 @@ livedata.panel.sort.direction.descending=Descending
 livedata.panel.sort.add=Add sort
 livedata.panel.sort.delete=Delete sort
 
+livedata.displayer.emptyValue=N/A
+
 livedata.displayer.link.noValue=(no value)
 
 livedata.displayer.boolean.true=True
@@ -117,4 +119,4 @@ livedata.displayer.xObjectProperty.failedToRetrieveField.errorMessage=Failed to 
 livedata.filter.list.emptyLabel=---
 
 livedata.footnotes.computedTitle=Some pages have a computed title. Filtering and sorting by title will not work as expected for these pages.
-livedata.footnotes.propertyNotViewable=Some pages require special rights to be viewed.
+livedata.footnotes.propertyNotViewable=Displayed when some pages require special rights to be viewed and thus fields cannot be extracted from them.
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
@@ -115,3 +115,6 @@ livedata.displayer.xObjectProperty.missingDocumentName.errorMessage=Can't save a
 livedata.displayer.xObjectProperty.failedToRetrieveField.errorMessage=Failed to retrieve the {0} field.
 
 livedata.filter.list.emptyLabel=---
+
+livedata.footnotes.computedTitle=Some pages have a computed title. Filtering and sorting by title will not work as expected for these pages.
+livedata.footnotes.propertyNotViewable=Some pages require special rights to be viewed.
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
@@ -46,7 +46,7 @@
         -->
         <span>{{ value }}</span>
       </slot>
-      <span v-if="!isViewable">{{ $t('livedata.displayer.emptyValue') }}</span>
+      <span v-if="!isViewable" v-html="$t('livedata.displayer.emptyValue')"></span>
     </div>
 
     <!-- The slot containing the displayer Editor widget -->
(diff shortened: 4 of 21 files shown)


manuelleduc:
Preview of the UI after change:
Screenshot from 2021-06-17 11-03-30

vmassol:
> Preview of the UI after change:
> Screenshot from 2021-06-17 11-03-30

Doesn't look very nice for the `(no value)` lines. How is it rendered in the LT? Also there's no `*` mention after `(no value)` so we don't know if the `(*)` legend part is referring to this or not

vmassol:
> > Preview of the UI after change:
> > Screenshot from 2021-06-17 11-03-30
> 
> Doesn't look very nice for the `(no value)` lines. How is it rendered in the LT? Also there's no `*` mention after `(no value)` so we don't know if the `(*)` legend part is referring to this or not

Also does the order means that the "hidden" page name have a first letter between P and S? Should we keep the order? Feels a bit strange too.

manuelleduc:
> > > Preview of the UI after change:
> > > Screenshot from 2021-06-17 11-03-30
> > 
> > 
> > Doesn't look very nice for the `(no value)` lines. How is it rendered in the LT? Also there's no `*` mention after `(no value)` so we don't know if the `(*)` legend part is referring to this or not
> 
> Also does the order means that the "hidden" page names have a first letter between P and S? Should we keep the order? Feels a bit strange too.

I agree, this part is a consequence of the result returned Livetable source and we'd have the same result with a Livetable.
The best way to improve this is probably to introduce a new source dedicated to document listing (or fixing the Livetable queries to take into account rights).

manuelleduc:
> Doesn't look very nice for the `(no value)` lines. How is it rendered in the LT? Also there's no `*` mention after `(no value)` so we don't know if the `(*)` legend part is referring to this or not

This is the result with a Livetable:
image

The '*' messages is displayed but we don't know which field it refers to. We can either hide the '*' in the cell, or also show them aside the `(no value)` text too if this is clearer.

vmassol:
So for me we could do at least this to improve it:
* Improve the message and instead of "(no value)", use something more explanatory like `N/A` or `Cannot be computed` or something like that.
* Make sure that the legend clearly refers to it. For example `(N/A) Displayed when some pages require special rights to be viewed and thus fields cannot be extracted from them`.

Is that possible? Do we know when it's a right issue? I guess we can since the source could return something specific in this case.

cc @mflorea any idea on your side to improve this?

PS1: The LT display with empty lines is even worse IMO, I don't see how users can know that the `(*) ...` legend refers to these empty lines.
PS2: How could the user do a "copy" action if the page doesn't have rights to view it?

manuelleduc:
> Is that possible? Do we know when it's a right issue? I guess we can since the source could return something specific in this case.

Yes, for instance, if a property is viewable but empty, we do not display the (`*` / `N/A`)

> PS2: How could the user do a "copy" action if the page doesn't have rights to view it?

I need to create an issue for this one, this is not related to this PR and a consequence of inconsistent values sent from the livetable source.

manuelleduc:
The result using `N/A` instead of `*` + `(no value)` replaced by `N/A` in the link:

image

mflorea:
> Doesn't look very nice for the (no value) lines.

@vmassol "(no value)" is a special output of the **link** displayer when the link label is missing or empty but the link URL is available. Since you cannot navigate to a link with no label the link displayer is using the "(no value)" label for the link. There's a link there to the page! In this case the page title (which is the label) is missing because the user doesn't have the right to view it. At the same time the page URL is available because it is not protected by rights ATM (the page URL is computed from the page reference and the page reference is not protected).

> Also there's no * mention after (no value)

Yes, because the "value" is **partially missing**. The link label is missing but the link URL is not. If the link URL was missing also then you would see just the ``*``.

> Also does the order means that the "hidden" page names have a first letter between P and S? Should we keep the order? Feels a bit strange too.

You know very well that ATM we cannot check access rights at query time (due to rights inheritance) so we cannot exclude the protected pages easily without either breaking the pagination or affecting the performance. Since the protected pages are "included" in the live data results (with partial information) we can:

* either not show them at all but then the users will complain they don't see all 15 results that are advertised at the top of the table (the pagination doesn't match the display)
* or show them masked using ``*`` and adding a footnote / legend at the bottom of the page, which @manuelleduc did

> use something more explanatory like N/A

I really don't see how ``N/A`` is more explanatory than "(no value)" for the link displayer. I much prefer the output @ClemDee and I chose. I'm fine to use ``N/A`` instead of ``*`` when the value is **completely** missing, if you think it's better, but I'd like to keep "(no value)" for the link displayer when the label is empty but there is an URL.

manuelleduc:
> I really don't see how `N/A` is more explanatory than "(no value)" for the link displayer. I much prefer the output @ClemDee and I chose. I'm fine to use `N/A` instead of `*` when the value is **completely** missing, if you think it's better, but I'd like to keep "(no value)" for the link displayer when the label is empty but there is an URL.

Note that in this PR, `(no value)` is displayed for the link when it has actually no defined value, whereas `N/A` is displayed when there is no value because of the rights.

vmassol:
> You know very well that ATM we cannot check access rights at query time (due to rights inheritance) so we cannot exclude the protected pages easily without either breaking the pagination or affecting the performance.

I didn't say to exclude them... I mentioned that I was finding that the order was not the best and that it was also revealing information about the page names because of the order. Moving the n/a pages to the bottom of the list would not break nor affect the performance to me. I don't know if it's a good idea or not.

vmassol:
> I really don't see how `N/A` is more explanatory than "(no value)" for the link displayer. I much prefer the output @ClemDee and I chose. I'm fine to use `N/A` instead of `*` when the value is **completely** missing because of a rights limitation, if you think it's better, but I'd like to keep "(no value)" for the link displayer when the label is empty but there is an URL.

My main comments were:
* The line looks  strange with a mix of "no value" and "*"
* The user cannot link the "no value" with the legend information that starts with `(*)``
* No value just means that there's no value! it doesn't explain that the issue is that the user doesn't have view rights to the page. A user will want to know why there's no value.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
```diff
@@ -47,7 +47,8 @@
       {"id": "html"},
       {"id": "xObjectProperty"},
       {"id": "actions"},
-      {"id": "boolean"}
+      {"id": "boolean"},
+      {"id": "docTitle"}
```
mflorea, line 51: ``docTitle`` (and ``xObjectProperty``) displayers are not generic (from the point of view of the live data API) so they shouldn't be here. They are specific to a live data source. I'd move this configuration to the livetable module for now if possible (hoping that the merge will work as expected, i.e. keep the standard displayers and add the source-specific ones).

vmassol:
> Note that in this PR, `(no value)` is displayed for the link when it has actually no defined value, whereas `N/A` is displayed when there is no value because of the rights.

There's something I don't understand:
* Why would the user have access to the URL of a page if he doesn't have view rights on it
* What does it mean that the link has no value? How can a page not have a URL?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/resources/liveTableLiveDataConfiguration.json
```diff
@@ -176,7 +176,8 @@
     ],
 
     "displayers": [
-      {"id": "link", "propertyHref": "doc.url"}
+      {"id": "link", "propertyHref": "doc.url"},
+      {"id": "docTitle", "propertyHref": "doc.url"}
```
mflorea, line 180: Here is where I would list the ``docTitle`` and ``xObjectProperty`` displayers (not in the live data API).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -114,4 +116,7 @@ livedata.displayer.boolean.false=False
 livedata.displayer.xObjectProperty.missingDocumentName.errorMessage=Can't save an XObject property with a missing document name.
 livedata.displayer.xObjectProperty.failedToRetrieveField.errorMessage=Failed to retrieve the {0} field.
 
-livedata.filter.list.emptyLabel=---
\ No newline at end of file
+livedata.filter.list.emptyLabel=---
+
+livedata.footnotes.computedTitle=Some pages have a computed title. Filtering and sorting by title will not work as expected for these pages.
+livedata.footnotes.propertyNotViewable=Displayed when some pages require special rights to be viewed and thus fields cannot be extracted from them.
```
mflorea, line 122: Unlike the previous message, which is specific to page title, this one is generic, independent of the live data source: the live data configuration allows you to express that a given entry is not viewable. So we should:

* either rephrase it to refer to **entries** rather than **pages**
* or add support for "overwriting" the message with a more specific one "from the live data source"

I'd keep it simple an rephrase the message.

manuelleduc:
> * Why would the user have access to the URL of a page if he doesn't have view rights on it (the other fields are not shown, suggesting the user doesn't have the rights to view the page)

This is a consequence of what is returned by the livetable source, only the page title is filtered but not its url.
 
> * What does it mean that the link has no value? How can a page not have a URL?

By "no value" I mean nothing to display inside the `a` tag, `(no value)` is here to prevent the situation where a link is present in the html but impossible to click because empty (`<a href="..."><!-- nothing here --></a>` )

What we could do is: 
1. if the user does not have the view right on the document, `N/A` is displayed (without a link)
2. if the user have the view right, we keep the current behavior.

vmassol:
Note: Regarding `N/A`, I think it's better to use `N/A*` (the star is a superscript) and have the legend be `(*) ...`

WDYT?

manuelleduc:
> Note: Regarding `N/A`, I think it's better to use `N/A*` (the star is a superscript) and have the legend be `(*) ...`
> 
> WDYT?

sorry missed that part, +1 for me

vmassol:
> This is a consequence of what is returned by the livetable source, only the page title is filtered but not its url.

WDYM by filtered? You mean that the page title is not displayed because the user doesn't have the rights on the page? And we currently allow the user to see the URL even if he/she doesn't have view rights on the page?

> What we could do is:
>
> * if the user does not have the view right on the document, N/A is displayed (without a link)
> * if the user have the view right, we keep the current behavior.

Yes sounds good. For me it's a lot more important for the user to understand that the title is not displayed because he doesn't have view rights on the page than because it's empty (which I believe is just a consequence of the user not allowed to see the page).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -67,6 +67,33 @@ define('xwiki-livedata', [
...
+  /**
+   * A service providing footnotes related operations. 
+   */
+  class FootnotesService {
+    constructor(logic) {
+      this._logic = logic;
+    }
+
+    /**
+     * Register a new footnote. If a footnote with the same translationKey is already registered the new
```
mflorea, line 79: Something is missing at the end.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -67,6 +67,33 @@ define('xwiki-livedata', [
...
+   */
+  class FootnotesService {
+    constructor(logic) {
+      this._logic = logic;
+    }
+
+    /**
+     * Register a new footnote. If a footnote with the same translationKey is already registered the new  
+     * @param prefix
+     * @param translationKey
```
mflorea, line 81: Would be good to add some documentation if you list them :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -76,7 +103,9 @@ define('xwiki-livedata', [
    */
   const Logic = function (element) {
     this.element = element;
-    this.data = JSON.parse(element.getAttribute("data-config") || "{}");
+    const data = JSON.parse(element.getAttribute("data-config") || "{}");
+    data.footnotesSet = [];
```
mflorea, line 107: Why not have this as an **internal** (private) field of ``FootnotesService``? It's only used there AFAICS.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/BaseDisplayer.vue
```diff
@@ -123,7 +128,17 @@ export default {
...
     },
-    
+    isViewable() {
+      var empty = this.isEmpty;
+      if (empty === undefined) {
+        empty = !this.value
+      }
+      const isViewable = this.logic.isActionAllowed('view', this.entry) || !empty;
+      if (!isViewable) {
+        this.logic.footnotes.put(this.$t('livedata.displayer.emptyValue'), 'livedata.footnotes.propertyNotViewable');
```
mflorea, line 138: The fact that this function has side effects (adds the footnote) is a but unexpected (for a computed property that returns boolean). I don't have a better idea though.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/displayers/DisplayerDocTitle.vue
```diff
@@ -0,0 +1,29 @@
...
+import displayerLinkMixin from "./displayerLinkMixin.js";
+
+export default {
+  name: "displayer-docTitle",
+  components: {DisplayerLink},
+  mixins: [displayerMixin, displayerLinkMixin],
+  computed: {
+    html() {
+        if (this.entry['doc.title_raw']) {
+        this.logic.footnotes.put('1', 'livedata.footnotes.computedTitle');
```
mflorea, line 17: Same as above. The side effect might be unexpected. I guess we can live with this for now.

mflorea (changes requested):

mflorea:
> And we currently allow the user to see the URL even if he/she doesn't have view rights on the page?

Yes, because the information is in the live data JSON (live table results). Hiding it would just mean security by obscurity. The user can check the response JSON and access the page URL anyway (based on the page reference).

Allowing the user to access the URL even if the page is protected can be helpful if the user wants to ask an administrator to give them access or to ask why they don't have access.

PR #1647, opened 2021-06-17
XWIKI-17571: Allow to unregister a right

surli:
## JIRA

https://jira.xwiki.org/browse/XWIKI-17571

## Work done

  * Add a new API in AuthorizationManager to unregister a right
  * Provide some new package private method in Right to unregister it
  * Refactor tests to use junit5 and provide new tests for
    register/unregister
  * Use the new API when the DefaultLikeManager is disposed to
    unregister the right.

## Tests

Unit tests provided, also tested manually by uninstalling Like API and checking the UI for Extension Rights.


tmortagne:
I wonder if we should not also make sure to invalidate from the security cache any entry related to this right.

surli:
> I wonder if we should not also make sure to invalidate from the security cache any entry related to this right.

Isn't it enough to call https://github.com/xwiki/xwiki-platform/pull/1647/files#diff-1f6ca2d0723f4b628e991417281f14f9ff201aec0f65d8bd085de2c50c984aacR222 from what I understand this method also remove all children rules from the cache, so basically everything. We're using the same line of code for register, which would have some problems because the cache would contain wrong values for the implied rights, no?

tmortagne:
It's too much actually, but indeed I don't know how I missed that line.

PR #1650, opened 2021-06-22
XWIKI-16544: Improve getdocuments template

aemi-dev:
Previously, `getdocuments` was permitting to retrieve non-viewable
documents reference whereas this update does not include these results.
A drawback of this change is that a pagination relying on `getdocuments`
can not use offset number and returnedrows as previously. Since the
number of undisclosed items is unknown, the total number of checked rows
is included in the answer in order to be used as the next offset in the
pagination.

Also, totalrows can not be used reliably since it does not differentiate
between viewable and non-viewable documents because this verification is
done after the query.

About concrete changes, most of the query-related code is now wrapped in a for loop that looks for documents in order to reach the limit. The loop stops if queried items are fewer than the limit or if the limit is reached.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
@@ -23,70 +23,11 @@
 ### XWiki.LivetableResults. It supports basic document listing, sorting and pagination.
 ###
 $response.setContentType("application/json")
-##
-## WHERE
-##
-#set ($whereQueryPart = 'WHERE 1=1')
-#set ($queryParams = [])
-## Display all documents in an exact given space (doesn't support Nested Spaces). Can be useful for backward
-## compatibility for example or if there's a need to list exactly the content of a given space.
-#if ("$!request.space" != '')
-  #set ($discard = $queryParams.add($request.space))
-  #set ($whereQueryPart = "${whereQueryPart} AND doc.space = ?${queryParams.size()}")
-#end
-## Display all documents in a space, including Nested Spaces
-#if ("$!request.childrenOf" != '')
-  #set ($discard = $queryParams.add($services.query.parameter().literal("$!{request.childrenOf}.").anyChars()))
-  #set ($whereQueryPart = "${whereQueryPart} AND doc.fullName LIKE ?${queryParams.size()}")
-  #set ($discard = $queryParams.add("${request.childrenOf}.WebHome"))
-  #set ($whereQueryPart = "${whereQueryPart} AND doc.fullName <> ?${queryParams.size()}")
-#end
-## Exclude some documents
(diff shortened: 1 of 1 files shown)


surli:
So if I understand properly, the solution you propose breaks usage of `offset` and `totalrows`, but does provide a new `checkedrows` to mitigate it, is that correct? 
I don't think that's enough to make the solution acceptable: breaking the offset and totalrows is a big issue, especially since those are used in the livetables if I'm correct, so those values should stay correct.

aemi-dev:
The offset itself is not broken in the way I see it, but the livetable pagination would be broken or really confusing (e.g. some results could be displayed in more than a single page).
Anyway, I have to investigate LiveTable a bit but the fix appears to be not as easy as I proposed.

mflorea:
Breaking the pagination is not acceptable.

> a for loop that looks for documents in order to reach the limit. The loop stops if queried items are fewer than the limit or if the limit is reached.

This can be very very costly in its worse case: you have thousands of documents and the current user can't view any of them. You'll end-up loading all documents in memory and performing thousands of access rights checks. While this can be viewed as more user friendly (minus the performance hit which could lead to an unresponsive wiki), this is not what we currently do in other places such as Solr search or live data/table (where we simply return the rows from the current page that are accessible, without trying to fill the page).

aemi-dev:
> Breaking the pagination is not acceptable.
> 
> > a for loop that looks for documents in order to reach the limit. The loop stops if queried items are fewer than the limit or if the limit is reached.
> 
> This can be very very costly in its worse case: you have thousands of documents and the current user can't view any of them. You'll end-up loading all documents in memory and performing thousands of access rights checks. While this can be viewed as more user friendly (minus the performance hit which could lead to an unresponsive wiki), this is not what we currently do in other places such as Solr search or live data/table (where we simply return the rows from the current page that are accessible, without trying to fill the page).

That's why and after some discussions, I came back to an simpler fix that does not affect much how things work.
Thanks for your comment.

surli (changes requested):
I don't see the relation between your changes here and XWIKI-16544. The issue is about listing documents, not displaying author names. Moreover if you'd want to use a string to display be sure to use a translation key.

aemi-dev:
> I don't see the relation between your changes here and XWIKI-16544. The issue is about listing documents, not displaying author names.

Since my previous proposal is not compatible with livetable and livedata and so removing elements from the queried results is not on option.
Since the policy in regards of documents is that we don't hide their references, I can't hide them, it has to be consistent.

So there remains only two options:
- hide what should not be publicly available and If I'm not wrong, these few fields are the only one
or
- decide to change a bit our policy in terms of what should be available to anyone about documents in order to make hiding references, the best way to fix this issue.

These are the reasons why finally, it is just about hiding full name in my opinion. But if you have an other ideas, be sure I would be glad to try them.

> Moreover if you'd want to use a string to display be sure to use a translation key.

I'll have a look at that.

manuelleduc:
Replaced by https://github.com/xwiki/xwiki-platform/pull/1696

PR #1653, opened 2021-06-28
XWIKI-18605: Live Data macro doesn't use the configured Icon Theme

manuelleduc:
This is a Draft PR to discuss the migration from fixed Font Awesome icons to icons selected for XWiki's Icon Sets.


xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
```diff
@@ -101,6 +101,7 @@ eject = eject
 step-forward = step-forward
 step-backward = step-backward
 fast-forward = fast-forward
+fast-backward = fast-backward
```
manuelleduc, line 104: Everything related to icon sets will be moved to a dedicated commit once the PR is in a ready to merge state.

manuelleduc:
Using `more-vertical` and `fa-caret-up`/`fa-caret-down` is find with Font Awesome but is not with Silk because `more-vertical` and `fa-caret-down` are mapped to the same icon.

Peek 2021-06-28 15-11
Peek 2021-06-28 15-09

manuelleduc:
Using `caret-right`/`caret-down` for the action to minimize/maximize the advanced setting panels is not ok since the icons (both in Silk and Font Awesome) does not have exactly the same color:

Peek 2021-06-28 15-33
Peek 2021-06-28 15-17

manuelleduc:
For filtering, I prefer the funnel over the magnifying glass, but I could get used to it if I'm the only one to complain.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -72,23 +72,19 @@
...
           v-if="logic.isPropertySortable(property.id)"
-          :class="[
-            'sort-icon',
-            'fa',
-            { 'sorted': isFirstSortLevel(property) },
-            (isFirstSortLevel(property) && firstSortLevel.descending) ? 'fa-caret-up' : 'fa-caret-down',
-          ]"
-        ></span>
+          :icon-descriptor="{name: isFirstSortLevel(property) && firstSortLevel.descending? 'caret-up': 'caret-down'}"
+          :additional-classes="['sort-icon',  isFirstSortLevel(property)? 'sorted': '']"/>
```
ClemDee, line 87: Why do you need here to create an `additional-classes` property on the XWikiIcon component? Can't you simply use the regular `class` attribute here?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/XWikiIcon.vue
```diff
@@ -19,47 +19,113 @@
  -->
 
 <template>
+  <img v-if="isImage" :src="url" :class="additionalClasses"/>
```
ClemDee, line 22: Maybe it comes from the auto-formatting you are using, but it is considered as a [good practise](https://vuejs.org/v2/style-guide/#Multi-attribute-elements-strongly-recommended) to keep each HTML attribute on a separate line, because this is easier to read and easier to keep track of changes in diffs. This would also ensure that the formatting stays consistent with what's already existing!

manuelleduc: I tend to keep the attributes on a single line as long as it stays below the column limit, but that's only a personal taste.
I'll update to stick to the best practice :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
manuelleduc: I yes, I forgot classes of a component call are passed to the component root. Indeed that would be simple and more idiomatic that way. I'll fix that.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/utilities/XWikiIcon.spec.js
```diff
@@ -0,0 +1,105 @@
...
+    const wrapper = initWrapper({
+      propsData: {
+        additionalClasses: ["newClass"],
+        iconDescriptor: {
+          cssClass: "fa fa-plus"
+        }
+      }
+    });
+
+    expect(wrapper.html()).toBe('<span class="fa fa-plus newClass"></span>');
```
ClemDee, line 48: I'm not sure it is a really good practice to evaluate the whole html of the component, this seems very fragile to me.
I would rather focus the tests on what's being actually tested, in that case that the [class](https://vue-test-utils.vuejs.org/api/wrapper/#classes) attribute:
```js
expect(wrapper.classes()).toContain('fa')
expect(wrapper.classes()).toContain('fa-plus')
expect(wrapper.classes()).toContain('newClass')
```
So that if one day we decide to change the tag name, or add an attribute to the element, we won't have to update the tests.

manuelleduc: I don't think we have a rule for that. When the html of the component is small, I find this kind of assertions easier to read and more exhaustive.
What you propose does not assert:
- the kind of produced tag
- the absence of any other attribute

Of course whenever the html is getting larger, this does not work, or only by first selecting sub elements, for instance `expect(wrapper.find('.someclass).html()).toBe('...')`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/utilities/XWikiIcon.vue
ClemDee: Great, thanks! :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/tests/unit/utilities/XWikiIcon.spec.js
ClemDee: > What you propose does not assert:
>
> - the kind of produced tag
> - the absence of any other attribute

Yes, this is exactly my point!

The test is testing: "Displays an icon using a font", so what it should assert is whether the icon is displayed on the component. It should not care about the other component attributes or the component tag, it is not part of its concern.

If the element tag is really important though, you can create another test, that is going to assert that specifically:
```js
  it('Has correct tag', () => {
    const wrapper = initWrapper({ /* ... */ });

    expect(wrapper.name()).toBe('span')
  });
```

But actually the [`name` method is deprecated](https://vue-test-utils.vuejs.org/api/wrapper/#name) because this is considered a [bad practice](https://vue-test-utils.vuejs.org/upgrading-to-v1/#name) to go this far into implementation details.

ClemDee: > What you propose does not assert:
>
> - the kind of produced tag
> - the absence of any other attribute

Yes, this is exactly my point!

The test is testing: "Displays an icon using a font", so what it should assert is whether the icon is displayed on the component. It should not care about the other component attributes or the component tag, it is not part of its concern.

If the element tag is really important though, you can create another test, that is going to assert that specifically:
```js
  it('Has correct tag', () => {
    const wrapper = initWrapper({ /* ... */ });

    expect(wrapper.name()).toBe('span')
  });
```

But actually the [`name` method is deprecated](https://vue-test-utils.vuejs.org/api/wrapper/#name) because this is considered a [bad practice](https://vue-test-utils.vuejs.org/upgrading-to-v1/#name) to go this far into implementation details.

manuelleduc:
Font Awesome preview
image
And Silk preview
image

PR #1655, opened 2021-06-29
XWIKI-18722: Current wiki not passed to the context when calling a Livetable result template or page

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-18722

- If the namespace if not null and of type wiki, the context is updated with the namespace's wiki.
- The namespace parameter is now sent when when creating Live Data requests

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntriesResource.java
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntriesResource.java
@@ -34,6 +34,8 @@ import javax.ws.rs.core.Response;
 
 import org.apache.commons.lang3.StringUtils;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.component.namespace.Namespace;
+import org.xwiki.component.namespace.NamespaceUtils;
 import org.xwiki.livedata.LiveData;
 import org.xwiki.livedata.LiveDataConfiguration;
 import org.xwiki.livedata.LiveDataEntryStore;
@@ -48,9 +50,12 @@ import org.xwiki.livedata.rest.LiveDataEntryResource;
 import org.xwiki.livedata.rest.LiveDataSourceResource;
 import org.xwiki.livedata.rest.model.jaxb.Entries;
 import org.xwiki.livedata.rest.model.jaxb.Entry;
+import org.xwiki.model.namespace.WikiNamespace;
 import org.xwiki.rest.Relations;
 import org.xwiki.rest.internal.Utils;
 import org.xwiki.rest.model.jaxb.Link;
+
+import com.xpn.xwiki.XWikiContext;
 
 /**
  * Default implementation of {@link LiveDataEntriesResource}.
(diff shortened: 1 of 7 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntriesResource.java
```diff
@@ -66,25 +71,22 @@
...
-            getConfig(sourceId, actualProperties, actualMatchAll, actualSort, actualDescending, offset, limit);
-        Optional<LiveDataSource> source = this.liveDataSourceManager.get(config.getQuery().getSource(), namespace);
-        if (source.isPresent()) {
-            String idProperty = config.getMeta().getEntryDescriptor().getIdProperty();
-            LiveDataEntryStore entryStore = source.get().getEntries();
-            return createEntries(entryStore.get(config.getQuery()), idProperty, config.getQuery().getSource(),
-                namespace).withOffset(offset).withLimit(limit);
-        } else {
-            throw new WebApplicationException(Response.Status.NOT_FOUND);
+        XWikiContext xWikiContext = this.xcontextProvider.get();
```
mflorea, line 77: Did you try using the ``ModelContext`` to set the current entity instead (passing the wiki reference)? It would be a bit more cleaner I think.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntriesResource.java
```diff
@@ -66,25 +71,22 @@
...
+            // Switch the wikiId of the context if a namespace is provided and is of type wiki (for instance "wiki:s1").
+            Namespace namespace = NamespaceUtils.toNamespace(namespaceStr);
+            if (namespace != null && Objects.equals(namespace.getType(), WikiNamespace.TYPE)) {
+                xWikiContext.setWikiId(namespace.getValue());
+            }
+
+            LiveDataConfiguration config = initConfig(sourceId, properties, matchAll, sort, descending, offset, limit);
+            return getEntries(namespaceStr, offset, limit, config);
+        } finally {
+            xWikiContext.setWikiId(savedWikiId);
```
mflorea, line 89: This rollback is not needed in practice because each request calls a single REST resource so there's nothing else executed after, that would use the current wiki.

manuelleduc: I understand this finally is not useful now but  couldn't this kind of call be wrapped or followed by some generic code using the context in the future?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-rest/src/main/java/org/xwiki/livedata/internal/rest/DefaultLiveDataEntriesResource.java
manuelleduc: Indeed that looks cleaner, testing.

manuelleduc: Importing `ModelContext` requires to also import `WikiReference`, leading to a `ClassFanOutComplexity` checkstyle error. 
I'll see how it can refactor the code to avoid this issue.

mflorea (approved):

PR #1663, opened 2021-07-09
XWIKI-18723: Make ConfigurableClass support all space and wiki administration cases

surli:
# JIRA

https://jira.xwiki.org/browse/XWIKI-18723

# Work done

  * Provide a new scope property in ConfigurableClass with 4 possible
    values: WIKI, SPACE, ALL_SPACES and WIKI+ALL_SPACES. This property
replace the old boolean property configureGlobally which was standing
for WIKI (true) and SPACE (false).
  * Refactor a bit ConfigureClassMacros to use the new property
  * Refactor all existing ConfigureClass in xwiki-platform to use the
    new property

# TODO

 - [x] Provide a migration for other ConfigurableClass 
 - [x] Provide translations for ConfigureClass properties
 - [x] Check retro compatibility in case of installing an extension containing an old ConfigureClass


tmortagne:
While checking if you added translation key for the new `scope` field and its values, I noticed that we actually don't seem to have any standard translation keys for the XWiki.Configurable class fields.

tmortagne:
Hmm, I'm afraid a migration for the new `scope` field is not enough: if you install an extension which use the old way to control where the section should be displayed (so all extensions currently) it won't be taken into account anymore (what's the default by the way ?).

surli:
> Hmm, I'm afraid a migration for the new scope field is not enough from script retro compatibility point of view: if you install an extension which use the old way to control where the section should be displayed (so all extensions currently) it won't be taken into account anymore (what's the default by the way ?).

I guess I'll modify a bit the ConfigurableClassMacros to keep looking in the DB for configureGlobally. Now we might have some unexpected behaviour if someone starts using the scope property without removing the deprecated configureGlobally property. But I'm not sure we can really avoid that and it's dev jobs to be careful with what they put on those classes I guess. Re the default value, I guess I'll keep it empty to avoid any issue with this usecase: I don't want an installed doc which have a configureGlobally set to false gets a default value set to WIKI, it would create exactly the kind of wrong behaviour I want to avoid. 

Note that another way for dealing that would be to add a listener to apply the migration automatically whenever a new extension is installed but it might be a bit overkill.

tmortagne:
> Note that another way for dealing that would be to add a listener to apply the migration automatically whenever a new extension is installed but it might be a bit overkill.

Does not sound that bad, it feels like an easy thing to implement and it simply a lot the life of the rest of the code in charge of dealing with Configurable objects. The only downside is that it will be seen by Extension Manager as a customization of an extension page, but it should not be a big problem.

PR #1666, opened 2021-07-13
XWIKI-18002: Add autocomplete attributes to fields when editing the user profile personal information

slauriere:
This PR proposes to add the autocomplete attributes in JavaScript once the form is loaded. Alternatively these attributes could be added directly on form creation in Velocity but this would probably require more code.


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
```diff
@@ -141,6 +141,27 @@
       <code>require(['jquery', 'xwiki-meta'], function ($, xm) {
   'use strict';
   $(document).ready(function() {
+
+    /** Add autocomplete attributes for accessibility purpose. */
```
surli, line 145: Minor remark but I don't think autocomplete it's an accessibility feature only: for me it's mainly something handy to avoid retyping stuff. At least there's no mention of accessibility when I read the doc in https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete.

surli: Now I feel a bit strange to have that feature only at User profile edition: it means that when you register you don't have autocomplete features filling the names / emails and you have it when you edit the profile. It might feel a bit inconsistent.

slauriere: Thanks Simon for your feedback. You're right, these attributes are useful also for avoiding retyping, I've updated the comment. As for registration: indeed, we need to update the form as well, I have created a distinct issue about that earlier today: [XWIKI-18837](https://jira.xwiki.org/browse/XWIKI-18837). We have also one for the login screen username / password fields: [XWIKI-18811](https://jira.xwiki.org/browse/XWIKI-18811).

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
```diff
@@ -141,6 +141,27 @@
       <code>require(['jquery', 'xwiki-meta'], function ($, xm) {
   'use strict';
   $(document).ready(function() {
+
+    /** Add autocomplete attributes to ease form filling. */
+    const prefix = 'XWiki.XWikiUsers_0_';
+    var fieldsPurposes = {
+      'first_name': 'given-name',
```
surli, line 148: quite frankly I'm not a big fan of this structure, I'd have preferred something such as: 
```
var fieldsPurposes = [
  {
    'name': 'first_name',
    'purpose': 'given-name'
  },
  ...
]
```
which is more explicit and allows extension IMO. It also helps to understand the code below since you'd use directly `entry.name` and `entry.purpose`.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
```diff
@@ -141,6 +141,27 @@
       <code>require(['jquery', 'xwiki-meta'], function ($, xm) {
   'use strict';
   $(document).ready(function() {
+
+    /** Add autocomplete attributes to ease form filling. */
+    const prefix = 'XWiki.XWikiUsers_0_';
```
surli, line 146: Globally not a big fan of hardcoded stuff like that... my first move would be "what if the page has several user xobjects?" but I guess in that case XWiki only take into account the first one for the sheet?

tmortagne: Actually, having several objects is not really the main risk, if you delete the XWikiUsers object and add a new one the first object number will be 1 and not 0.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
slauriere: Indeed, thanks, I used this structure in the latest commit.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
slauriere: Right, that was a very poor idea, thanks for the review. The latest commit uses instead a jQuery selector based on the input names ending, without any reference to the object(s) identifier, what do you think?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
```diff
@@ -141,6 +141,44 @@
...
+          'purpose': 'street-address'
+        },
+        {
+          'name': 'blog',
+          'purpose': 'url'
+        }
+      ];
+    fieldsPurposes.forEach(function (entry) {
+      var inputField = $('[name$="' + entry.name + '"]');
+      if (inputField.length === 1)
```
surli, line 178: just a style issue, not a big fan of a if without `{` personally, now I don't recall if we have a rule for this or not. Maybe @mflorea would know.

surli (approved):
Just a minor style comment otherwise it's good for me.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserProfileSheet.xml
slauriere: Indeed, I looked into the XWiki JS code base and couldn't find a `if` without brackets except when the conditional statement is in the same line as the `if`, so I added brackets in the latest commit. Should we add a JavaScript section to the [code style page](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/)?

vmassol: Yes that would be nice, but before that we should start a proposal thread on the dev forum and list a few best practices. We could also say that the Java coding style apply and then add exceptions where needed + ofc add js-specific stuff (like the use of JQuery, etc).

For example for me `'[name$="'` should be `'[name$ = "'` for the same reason as in Java (readability). Strangely its done right for the `+` in `name$="' + entry.name + '"]'`(spaces around them).

slauriere: Right, it's close to the Java code style indeed. 

Regarding the addition of spaces in the jQuery selector, I'm wondering if this is what we aim at since:
* The selector is meant to somehow represent part of an HTML element which itself does not typically contain spaces (eg `<input name="XWiki.XWikiUsers_0_firstName"/>` even though spaces around the equal sign [are perfectly allowed](https://dev.w3.org/html5/spec-LC/syntax.html#unquoted)).
* The [jQuery documentation](https://api.jquery.com/category/selectors/) does not include spaces around the selector equal character and most developers don't use spaces for attribute name selectors, it seems.

Wdyt?

surli: > The jQuery documentation does not include spaces around the selector equal character and most developers don't use spaces for attribute name selectors, it seems.

Yeah I don't think it would be a good rule for JS style, clearly if I had to review a selector such as `'[name$ = "'` I would think there's something wrong.

> Yes that would be nice, but before that we should start a proposal thread on the dev forum and list a few best practices.

I think @mflorea started to document some JS best practices, see: https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HJavaScriptBestPractices

vmassol: > The selector is meant to somehow represent part of an HTML element which itself does not typically contain spaces (eg <input name="XWiki.XWikiUsers_0_firstName"/> even though spaces around the equal sign are perfectly allowed).

This is a bad argument IMO. You need to understand the reason for the spaces in code. As I mentioned it's for readability so it's not related to the HTML at all (which BTW is not meant to be read, it's not a general purpose coding language), but to the JS.

vmassol: > Yeah I don't think it would be a good rule for JS style, clearly if I had to review a selector such as '[name$ = "' I would think there's something wrong.

I don't understand the argument. Could you clarify? Thx

vmassol: And not just to me...

vmassol: BTW it's always good to reuse something existing as a base and not reinvent our own. Found https://javascript.info/coding-style which could serve as a base if we don't want to use our java coding style as the base.

vmassol: > The jQuery documentation does not include spaces around the selector equal character and most developers don't use spaces for attribute name selectors, it seems.

Do you know the reason? I can see problems with it but I don't see the advantage of doing that. Do you?

ClemensRobbenhaar: I think it is because these selectors are originally CSS selectors, and it seems whitespace is frowned upon in attribute-like selectors, as whitespace is already used for nesting in CSS. See e.g. examples in the spec: https://www.w3.org/TR/selectors-4/#attribute-substrings 

Also `[name$ ="..."]` is in fact a syntax error (at least jQuery thinks so). The correct whitespacing would be `[name $= "..."]` because both the `$` and the `=` together form the operator `$=`, which is separate from the attribute name.

Then I am slightly afraid of "interesting" browser bugs that might surface because of whitespace in places where it is completely legal by the spec, but uncommon in use.

vmassol: So the rule about spaces around `=` would still be a good one but not in selectors which are special for the reasons mentioned. Fine with me.

slauriere: Thank you for your comments Clemens and Vincent, is there anything needed from me for the PR to go through?

slauriere:
Thank you all for your comments and reviews, is there anything needed from me for the PR to go through?

PR #1667, opened 2021-07-13
XWIKI-18006: Add mechanism to skip header parts and go to main content directly

slauriere:
* In view mode, add link to #xwikicontent
* Make link visible only when it has focus or when it is active, similarly to https://www.w3.org/TR/WCAG20-TECHS/G1.html
* The link uses `btn btn-default` classes
* The link points at `#xwikicontent`, not `#mainContentArea`, since it is meant to skip over the navigation tools including the breadrumb (source : https://www.w3.org/TR/WCAG20-TECHS/G1.html#G1-ex2)


vmassol:
It's good that you explained things in this PR but all these explanations will be lost when the PR is applied. It would be best to move these explanations in the code itself! :)

vmassol:
> In view mode, add link to #xwikicontent

Is it needed only in view mode? What about edit mode for ex?

vmassol:
Do you have a screenshot about how the new link is displayed to see the visual changes?

slauriere:
The latest commit intrudoces a `skipnav.vm` template which is included by `startpage.vm` and `edit.vm`. The `skipnav.vm` code makes the following assumption: all pages have either a `#xwikicontent` block (`view` and `preview` actions), or a `#editMeta` block (`inline` and `edit` actions), or a `#mainContentArea` block (all pages which include `startpage.vm`: `login.vm`, `create.vm`, `copy.vm`, `docdoesnotexist.vm` etc.), or are generated by a template which does not include `startpage.vm` (e.g. `pdf.vm`).

I'm adding a screenshot below illustrating how the link is displayed when it has focus or is active (otherwise it's invisible).

skip-nav

vmassol:
> I'm adding a screenshot below illustrating how the link is displayed when it has focus or is active (otherwise it's invisible).

Thanks for the screenshot. So if you move the mouse over the logo, the button appears? How do you click on the logo? How is the button activated? thx

slauriere:
Actually the button appears when the link contained in the `skip-nav` gets focus (typically via tabulation), not on mouse hover. The behaviour is the same as the one on [this example page](https://www.w3.org/TR/WCAG20-TECHS/G1.html). Clicking on the logo requires to move the focus to a distinct element than the button so that the button disappears (at least on the default installation where the logo's height is somewhat the same as the button one). Once the button has focus, it gets activated by hitting Enter.

vmassol:
> Actually the button appears when the link contained in the skip-nav gets focus (typically via tabulation), not on mouse hover.

ok cool, good that we don't make the wiki less usable when using the mouse, it looks fine to me. So if someone hits "tab" (once or several times), they'll see the button appear (and disappear after clicking more more on tab). Doesn't it feel a bit magical? I'm not sure I would understand why this button appears a user when I hit "tab".

> The behaviour is the same as the one on this example page.

I couldn't find where there was a button and how it was activated.

Thanks

slauriere:
> So if someone hits "tab" (once or several times), they'll see the button appear (and disappear after clicking more more on tab). Doesn't it feel a bit magical? I'm not sure I would understand why this button appears a user when I hit "tab".

It's the intent of [this accessibility technique](https://www.w3.org/TR/WCAG20-TECHS/G1.html) to control the presence of this link via the keyboard: "users navigating via the keyboard include those using techniques that generate keyboard strokes slowly, screen magnification software users, screen reader users working with sighted colleagues, keyboard only users and those navigating using voice recognition software".

It may sound weird since the rule does not seem to be implemented often, but isn't it quite handy on the other hand, for all users?

> I couldn't find where there was a button and how it was activated.

Do you mean that the button does not show up when navigation the page via tabs? Here's another example just in case:

https://www.numerique.gouv.fr/dinum/

BTW the link shows up as a button (`btn btn-default` classes) just to avoid using specific colors and maximize the chances to obtain a good contrast whatever the banner color.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -344,6 +344,9 @@ img=Attached Image
 imgtext=example.jpg
 sign=Signature
 
+### Skip to
+skipTo.content=Skip to Content
```
surli, line 348: The key doesn't really follow the convention: https://dev.xwiki.org/xwiki/bin/view/Community/L10N/Conventions/
Generally speaking the translation key should help identifying where the it's used, which is clearly not the case here.

slauriere: Thank you surli for the review and for the link to L10N conventions, I tried to propose a translation key that is more aligned with the rules, please let me know if you have better suggestions.

surli (approved):

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/skipnav.vm
```diff
@@ -0,0 +1,34 @@
...
+  #if ("$!xcontext.action" == 'view' || "$!xcontext.action" == 'preview')
+    ## Point at #xwikicontent to skip over the navigation tools including the breadrumb
+    #set ($targetNav = '#xwikicontent')
+  #elseif (("$!xcontext.action" == 'inline' || "$!xcontext.action" == 'edit') && !$editConfirmation)
+    ## In edit mode, point at the title field
+    #set ($targetNav = '#editMeta')
+  #else
+    ## Point at '#mainContentArea' which is present in all pages that include "startpage.vm"
+    #set ($targetNav = '#mainContentArea')
+  #end
```
michitux, line 32: I wonder if it wouldn't make sense to make this consistent with the main page region as this if/else seems to be quite fragile.

michitux (approved):
Overall, this looks good to me, I'm just wondering if it wouldn't make sense to make this jump target more consistent and in particular consistent with the main page regions proposed in #2166.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/skipnav.vm
Sereza7: `targetNav` is used as an href. I'm pretty sure we need the **id** of an element to have a proper href.
The id of the main region as introduced in https://github.com/xwiki/xwiki-platform/pull/2166 is **not defined** (no value, or value sensitive to the template). This could be a change I need to make on PR 2166 before we merge it.

Sereza7 (approved):
(Review for May the 30th 2023, nothing major changed in the content of this PR since then)
LGTM
Demo with changes applied on a 14.5 distribution: https://up1.xwikisas.com/#WHEkuSgty1Vy2NrlmLCRDg
Everything seems to be working properly.
I don't have anything to add concerning accessibility. AFAICS this introduces no WCAG violation, respects semantics and attributes.

PR #1668, opened 2021-07-13
XWIKI-17111: User profile sections are using <h1> tag for sub-heading

slauriere:
* Update profile sections to use h2 level headings instead of h1 and h3 instead of h2


surli:
@slauriere could you provide a screenshot before/after your changes to see the difference? Also my only worry here is that I'm wondering if we could have some JS that relies on those h2 tags? Maybe @mflorea would have an idea?

vmassol:
AFAIK this is not related to the user profile and we've used H1 voluntarily. It's even excluded from the WCAG tests. We could decide to change this but it needs a proper discussion IMO.

See https://dev.xwiki.org/xwiki/bin/view/Community/WCAGTesting#HTestExceptions

surli:
> We could decide to change this but it needs a proper discussion IMO.

That's apparently also the opinion of Marius and Anca on the ticket comments: https://jira.xwiki.org/browse/XWIKI-17111. Sounds like it will take a bit more time to be fixed, maybe change the status of this PR to draf @slauriere ?

slauriere:
Indeed @surli I converted it to draft, thank you for your input

Sereza7:
Closing this PR as its issue was solved along [XWIKI-21370](https://jira.xwiki.org/browse/XWIKI-21370) <--jira
PR to fix --> [XWIKI-21370: Header structure uses too many H1](https://github.com/xwiki/xwiki-platform/pull/2448)

PR #1671, opened 2021-07-16
XWIKI-18003: Pencil buttons on profile page are not consistent with the used icon theme and missing contrast

slauriere:
This pull request:
* Adds a pencil icon in profile edit link
* Updates the profile edit link style with btn btn-xs classes and dedicated CSS rules

The contrast has been checked with Firefox accessibility tools (7.29 : 1)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
@@ -286,6 +286,7 @@
             &lt;div class="editProfileCategory"&gt;
               &lt;a title="$escapetool.xml($services.localization.render('platform.core.profile.category.dashboard.edit'))"
                   href="$doc.getURL('edit', 'category=dashboard')" class="btn btn-xs"&gt;
+                &lt;span class="action-icon"&gt;$services.icon.renderHTML('pencil')&lt;/span&gt;
                 &lt;span class='sr-only'&gt;$escapetool.xml($services.localization.render('platform.core.profile.category.dashboard.edit'))&lt;/span&gt;
               &lt;/a&gt;
             &lt;/div&gt;
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
@@ -572,6 +572,7 @@ div.userInfo, div.userPreferences, div.watchlistManagement, div.userDashboard {
 .userInfo h2 {
   font-size: larger;
   font-weight: bolder;
+  margin-top: 10px;
 }
         
 div.userInfo input[type="text"], div.userInfo input[type="password"], div.userInfo textarea, div.userInfo select, div.userPreferences select {
@@ -582,12 +583,21 @@ div.editProfileCategory {
   float:right;
 }
 
-div.editProfileCategory a {
-  display:block;
-  width: 16px;
-  height: 16px;
-  background: url("$xwiki.getSkinFile('icons/silk/pencil.png')") no-repeat;
+div.editProfileCategory {
+  float:right;
+  a.btn-xs {
+    background-color: @xwiki-page-content-bg;
+    border: 1px solid @xwiki-border-color;
+    .action-icon {
+      font-size: larger;
+      color: @btn-default-color;
+    }
+    &amp;:hover, &amp;:active, &amp;:focus {
+      border-color: darken(@dropdown-divider-bg, 10%);
+    }
+  }
 }
+
 
 /* Watchlist */
 
@@ -677,6 +687,9 @@ span#avatarUpload {
 .dashboard .clearfloats:after, .dashboard .clearfloats:before {
   clear: right;
 }</code>
+    </property>
+    <property>
+      <contentType>LESS</contentType>
     </property>
     <property>
       <name>userprofile</name>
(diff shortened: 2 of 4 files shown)


xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
```diff
@@ -283,7 +283,12 @@
       &lt;div class="full column xform"&gt;
         &lt;div class="userDashboard"&gt;
           #if($xcontext.action == 'view' &amp;&amp; $hasEdit)
-            &lt;div class="editProfileCategory"&gt;&lt;a href="$doc.getURL('edit', 'category=dashboard')"&gt;&lt;span class="hidden"&gt;$services.localization.render('platform.core.profile.category.dashboard.edit')&lt;/span&gt;&lt;/a&gt;&lt;/div&gt;
+            &lt;div class="editProfileCategory"&gt;
+              &lt;a href="$doc.getURL('edit', 'category=dashboard')" class="btn btn-xs"&gt;
+                &lt;span class="action-icon"&gt;$services.icon.renderHTML('pencil')&lt;/span&gt;
+                &lt;span class='hidden'&gt;$escapetool.xml($services.localization.render('platform.core.profile.category.dashboard.edit'))&lt;/span&gt;
+              &lt;/a&gt;
+            &lt;/div&gt;
```
vmassol, line 291: Thanks @slauriere for the PR! About the new code layout, did you check the resulting HTML to see if this didn't add new lines and empty spaces? I think it could have been the reason why it was done that way.

I also like the way you did it, it's much simpler to read and I'd love to standardize on this. I think we need a preprocessor to remove empty lines and spaces though. Note: I see it's inside a velocity macro which means it'll remove the leading spaces; there'll still be the extra new lines probably though.

WDYT? Also curious to know what others think about it.

vmassol:
@slauriere thanks! If you could provide a screenshot before/after that would help see what this PR does to apply it. Thx

vmassol:
> @slauriere thanks! If you could provide a screenshot before/after that would help see what this PR does to apply it. Thx

ok, I've found it in the jira issue at https://jira.xwiki.org/secure/attachment/39682/39682_profile.jpg

thanks

vmassol:
I'm not good enough to validate the CSS changes but functionally it's a good change. I'm ok to apply it. Thanks for working on this

xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-ui/src/main/resources/Dashboard/XWikiUserDashboardSheet.xml
slauriere: Thanks @vmassol. I confirm the resulting HTML does not contain new lines nor empty spaces, at least when not using a screen reader, here's a capture in Firefox:

profile-1

However removing the "sr-only" class to simulate what happens on screen readers generates a whitespace (because we have now two visible spans next to another):

profile-2

I don't know how this whitespace is interpreted by screen readers. Since the leading span contains no text, it may be considered as the first whitespace of a text, hence ignored? Not sure. Another approach would consist in using the `aria-label` attribute instead of `sr-only` CSS class to provide the button purpose. Both approaches seem to be fine, I couldn't find whether one is recommanded over the other so far. On [this accessibility recommendation by Font Awesome](https://fontawesome.com/v5.15/how-to-use/on-the-web/other-topics/accessibility), the two spans are not juxtaposed:

```
<i aria-hidden="true" class="fas fa-car" title="Time to destination by car"></i>
<span class="sr-only">Time to destination by car:</span>
<span>4 minutes</span>
```

For insteractive elements such as a link in our case, the Font Awesome documentation suggests using `aria-label` rather than `sr-only`. However [this article](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/) states that "the [aria-label] attribute isn’t announced consistently by web browsers".

Note that using `aria-hidden` seems to be recommended with font icons, I'll create an issue about that.

As for new lines, I have this naive question: even if the code were not in a Velocity macro, can new lines actually get generated by the HTML macro even with `wiki="true"`?

vmassol: I wouldn't trust the HTML view tab in FF because it'll possibly remove whitespace and empty new lines. You should check the raw source I think.

vmassol: > I don't know how this whitespace is interpreted by screen readers. 

Best is to test it I guess :)

vmassol: > As for new lines, I have this naive question: even if the code were not in a Velocity macro, can new lines actually get generated by the HTML macro even with wiki="true"?

AFAIK only the Velocity macro has a filter to remove leading spaces. See https://extensions.xwiki.org/xwiki/bin/view/Extension/Velocity%20Macro#HParametersdefinition

vmassol: Re `aria-label` vs `sr-only`, I'm clueless, I haven't researched the topic. It's important that you make proposals on the forum since we don't have any decisions on anything beyond the automated Dutch WCAG guidelines tests.

surli:
@slauriere you have some conflicts to fix on this one.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -582,13 +583,22 @@ div.editProfileCategory {
   float:right;
 }
 
-div.editProfileCategory a {
-  display:block;
-  width: 16px;
-  height: 16px;
-  background: url("$xwiki.getSkinFile('icons/silk/pencil.png')") no-repeat;
+div.editProfileCategory {
```
surli, line 586: I'm not fluent enough in CSS to review this, might worth an eye from @mflorea

PR #1672, opened 2021-07-19
XWIKI-18009: Notifications macro's buttons missing purpose

slauriere:
* Add i18n labels to the two macro buttons with `sr-only` class so that the purpose of the buttons is visible on screen readers


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
```diff
@@ -422,8 +422,11 @@ require(['jquery', 'XWikiAsyncNotificationsMacro', 'xwiki-events-bridge'], funct
       if (!entry.read) {
         // Style the read button
         readButton.addClass('notification-event-read-button').addClass('btn btn-xs');
-        // Insert the cross icon
-        readButton.html('$services.icon.renderHTML("check")');
+        // Insert the cross icon and an accessible label
+        var readButtonIcon = $('$services.icon.renderHTML("check")');
```
surli, line 426: Sounds like the quotes should be used in the other way: simple quote for velocity to not interprate `check` since it's a constant, and double quotes in javascript. 
```suggestion
        var readButtonIcon = $("$services.icon.renderHTML('check')");
```
You should fix the same way below.

vmassol: Note that Stephane has kept the previous behavior so we have to wonder whether it was working or not before his change.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
```diff
@@ -422,8 +422,11 @@ require(['jquery', 'XWikiAsyncNotificationsMacro', 'xwiki-events-bridge'], funct
...
+        var readButtonLabel= $('&lt;span class="sr-only"/&gt;');
```
vmassol, line 427: what is `sr-only`?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
```diff
@@ -599,8 +602,11 @@ require(['jquery', 'XWikiAsyncNotificationsMacro', 'xwiki-events-bridge'], funct
           entry.find('.notification-content').prepend(readButton);
           // Style the read button
           readButton.addClass('notification-event-read-button').addClass('btn btn-xs');
-          // Insert the cross icon
-          readButton.html('$services.icon.renderHTML("check")');
+          // Insert the cross icon and an accessible label
+          var readButtonIcon = $('$services.icon.renderHTML("check")');
+          var readButtonLabel= $('&lt;span class="sr-only"/&gt;');
+          readButtonLabel.text('$services.localization.render("notifications.macro.markEventAsRead")');
+          readButton.append(readButtonIcon).append(readButtonLabel);
```
vmassol, line 609: Looks like code duplication with above, no? Isn't there a way to prevent the dup?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
```diff
@@ -423,9 +423,9 @@ require(['jquery', 'XWikiAsyncNotificationsMacro', 'xwiki-events-bridge'], funct
         // Style the read button
         readButton.addClass('notification-event-read-button').addClass('btn btn-xs');
         // Insert the cross icon and an accessible label
-        var readButtonIcon = $('$services.icon.renderHTML("check")');
+        var readButtonIcon = $("$escapetool.javascript($services.icon.renderHTML('check'))");
```
vmassol, line 426: I see you've modified as suggested by @surli . Does it mean it wasn't working before your change (since you just reused what was there)?

slauriere: The previous code was working at the cost of an extra string evaluation which was not needed indeed, as pointed out by @surli. This commit fixes this, adding also a JavaScript escape which is needed for escaping quotes possibly present (typically there are some quotes in the HTML icon code, hence a plain `$("$services.icon.renderHTML('check')")` would not work).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
slauriere: `sr-only` is a Bootstrap class standing for "screen reader only". Probably using `aria-label` instead could be considered simpler and possibly a better practice (not sure, currently investigating this), however at the moment we have no control on the attributes set by the Icon Service (I plan to create an issue about that).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-macro/xwiki-platform-notifications-macro-ui/src/main/resources/XWiki/Notifications/Code/Macro/NotificationsMacro.xml
slauriere: It's a duplication indeed because the page defines several JavaScript extensions, in particular `XWikiNotificationsMacro` and `XWikiAsyncNotificationsMacro` with code duplication. Preventing the duplication is probably doable but not trivial (for me) and might be beyond the scope of this issue (since there is more duplicated code than just for this button generation), what do you think?

vmassol: yes I agree :)

surli (approved):

PR #1676, opened 2021-08-12
XWIKI-18892: Properties that does not correspond to XClass properties should not be editable

manuelleduc:
Make the properties that are requested by the LD macro but do not correspond to a document or XClass property not-editable by default.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/WebHome.xml
+++ b/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/WebHome.xml
@@ -59,7 +59,7 @@
     'propertyDescriptors': [
       { 'id': 'wikiprettyname', 'displayer': { 'id': 'link', 'propertyHref': 'wikiprettyname_url' } },
       { 'id': 'owner', 'editable': false},
-      { 'id': 'membershipType', 'sortable': false, 'filterable': false },
+      { 'id': 'membershipType', 'sortable': false, 'filterable': false, 'editable': false },
       { 
         'id': '_actions',
         'displayer': {
--- a/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/test/java/org/xwiki/wiki/WikiManagerTest.java
+++ b/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/test/java/org/xwiki/wiki/WikiManagerTest.java
@@ -87,9 +87,11 @@ class WikiManagerTest extends PageTest
             new JSONObject(htmlDocument.getElementById("wikis").attr("data-config"))
                 .getJSONObject("meta")
                 .getJSONArray("propertyDescriptors");
-        Optional<JSONObject> propertyDescriptor = findObjectById(propertyDescriptors, "owner");
+        Optional<JSONObject> ownerPropertyDescriptor = findObjectById(propertyDescriptors, "owner");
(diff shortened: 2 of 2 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
```diff
@@ -187,6 +187,13 @@ private void addMissingPropertyDescriptors(LiveDataConfiguration config, List<St
                 .map(this::getDefaultPropertyDescriptor).collect(Collectors.toList());
         if (!missingDescriptors.isEmpty()) {
             propertyDescriptors = new ArrayList<>(propertyDescriptors);
+            // By default, a property that is declared in the macro configuration but is not a known XClass property
+            // and does not have a explicit 'editable' configuration is defined as not editable.
+            missingDescriptors.forEach(it -> {
+                if (it.isEditable() == null) {
+                    it.setEditable(false);
```
mflorea, line 194: Should be done from ``getDefaultPropertyDescriptor``. You can rename it to ``getMissingPropertyDescriptor`` if you don't like it. And if you make the missing property not editable then you should maybe also make it not sortable and not filterable by default (this would allow you to remove the ``membershipType`` descriptor from the wiki index).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
```diff
@@ -187,6 +187,13 @@ private void addMissingPropertyDescriptors(LiveDataConfiguration config, List<St
                 .map(this::getDefaultPropertyDescriptor).collect(Collectors.toList());
         if (!missingDescriptors.isEmpty()) {
             propertyDescriptors = new ArrayList<>(propertyDescriptors);
+            // By default, a property that is declared in the macro configuration but is not a known XClass property
+            // and does not have a explicit 'editable' configuration is defined as not editable.
```
mflorea, line 191: I would make this about XClass properties. I'd simply say that without any other information (property descriptor) it's safer to assume that properties with missing descriptors are not editable, not sortable and not filterable, by default.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/DefaultLiveDataConfigurationResolver.java
manuelleduc: I agree to move this into `membershipType`.
I'm less sure regarding sortable and filterable, for instance that makes the logging LD not sortable and filterable anymore. 
Of course we can agree on this and be explain it on the release note but that needs to be discussed :)

mflorea: Then we have an inconsistency. We say that properties without a descriptor are not editable by default but they are sortable and filterable. How do you justify this? I feel it should be all or nothing by default when the descriptor is missing. And since we already have an example where we expect the "missing" properties to be sortable and filterable I think it's best to say that they are also editable by default. Thus we should just fix the ``membershipType`` descriptor on the macro call to mark it as not editable.

manuelleduc: Sounds good. I'll change to PR with a config change in the macro config.

PR #1679, opened 2021-08-12
XWIKI-18904: App within minutes live tables short text field not (easily) searchable

manuelleduc:
- Initialize the matchers with default values when not defined on all cases.
- Introduce a page test of the regression

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
@@ -919,11 +919,15 @@
   #end
 #end
 
-
 #**
- * NOTE: This macro uses variables defined in livetable_filterProperty . It was not meant to be used alone.
- *#
-#macro (livetable_filterStringProperty)
+  * Initialize the match type with a default value when needed.
+  * The default match type depends on the type of the matched property.
+  * Notes:
+  * - This macro expects to find $matchType and $propMetaClass in the caller's scope
+  * - This macro updates the $matchType with the default value
+  * @since 
+  *#
+#macro (livetable_initializeMatchType)
   #if ("$!matchType" == '')
     #if ($propMetaClass.endsWith('ListClass'))
       ## Perform exact matching by default for StaticListClass, DBListClass and DBTreeListClass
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -919,11 +919,15 @@
...
 #**
- * NOTE: This macro uses variables defined in livetable_filterProperty . It was not meant to be used alone.
- *#
-#macro (livetable_filterStringProperty)
+  * Initialize the match type with a default value when needed.
+  * The default match type depends on the type of the matched property.
+  * Notes:
+  * - This macro expects to find $matchType and $propMetaClass in the caller's scope
+  * - This macro updates the $matchType with the default value
+  * @since
```
mflorea, line 928: Missing version.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -942,8 +952,9 @@
   #set ($filterQuery = "#livetable_getFilterQuery(""${safe_tableAlias}.value"" $matchType false $valueCount $paramPrefix)")
   #set ($whereSql = "${whereSql} and ($filterQuery.trim())")
   #foreach ($filterValue in $filterValues)
-    #set ($actualMatchType = $request.getParameterValues("${colname}_match").get($foreach.count - 1))
-    #livetable_addFilterParam($filterValue $actualMatchType $whereParams "${paramPrefix}${foreach.count}")
+    #set ($matchType = $request.getParameterValues("${colname}_match").get($foreach.count - 1))
+    #livetable_initializeMatchType()
+    #livetable_addFilterParam($filterValue $matchType $whereParams "${paramPrefix}${foreach.count}")
```
mflorea, line 957: The match type used to add the parameter value is still not synchronized with the match type used to generated the constraint (where clause) for that parameter. In other words, ``${colname}_match`` seems to influence only the parameter value and not how the where clause is built.

mflorea (changes requested):

mflorea:
The problem was fixed in a different commit.

PR #1686, opened 2021-09-02
XWIKI-16566: No exception mapper found in some conditions

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-16566

- Provide a generic exception mapper for all checked exceptions
- Log the mapped exception in the logs and return a HTTP 500 response


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+
+        // Logs the unmapped exception, with its stacktrace.
+        this.logger.error("A REST endpoint failed with an unmapped exception.", cause);
+
+        // Returns an error response to the client, with some detail about the exception in the response.
+        String message =
+            String.format("No ExceptionMapper was found for [%s]. Cause: [%s].", cause, getRootCauseMessage(cause));
+        return Response.serverError()
+            .entity(message)
+            .type(javax.ws.rs.core.MediaType.TEXT_PLAIN_TYPE).build();
```
surli, line 78: any reason to not import MediaType?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+        if (cause instanceof RuntimeException) {
+            throw (RuntimeException) cause;
+        }
+
+        // Logs the unmapped exception, with its stacktrace.
+        this.logger.error("A REST endpoint failed with an unmapped exception.", cause);
+
+        // Returns an error response to the client, with some detail about the exception in the response.
+        String message =
+            String.format("No ExceptionMapper was found for [%s]. Cause: [%s].", cause, getRootCauseMessage(cause));
```
tmortagne, line 75: I would expect to get the full stack trace in the response, actually. Similar to what you get when an /action/ crashes. I don't see any reason to limit the response content to a one liner.

vmassol: yes +1 for the full stacktrace

manuelleduc: iirc stacktraces are sensitive information that are best kept hidden from the client.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
manuelleduc: Good catch, fixing.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
tmortagne: The client is getting plenty of stack traces currently (failing macros, failing actions, etc.).

This (skipping the stack trace) should be more some option you can enable if you want, but the default should be to print the stack trace IMO.

manuelleduc: ok, I'll update with the full trace in this case.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,81 @@
...
+        // RuntimeException stacktraces are logged by default).
+        if (cause instanceof RuntimeException) {
+            throw (RuntimeException) cause;
+        }
+
+        // Logs the unmapped exception, with its stacktrace.
+        this.logger.error("A REST endpoint failed with an unmapped exception.", cause);
+
+        // Returns an error response to the client, with some detail about the exception in the response.
+        String message = String.format("No ExceptionMapper was found for [%s].%n%s", cause, getStackTrace(cause));
```
tmortagne, line 75: Hmm the format string is strange, you expect a String, a Number and a String but only pass a 2 String as parameter.

tmortagne, line 75: `%n` seems to mean "platform-specific line separator" but you are sending a content to an HTTP client here, and you are not going to read it on the server, so there is no need to be server system dependent. I guess a simple `\n` would do.

I'm wondering if it would not be better to produce this response content in a template. We have tools to produce stack traces (`#displayException` for example) that we use in various places.

manuelleduc:
Please do not merge before https://github.com/xwiki/xwiki-commons/pull/147/files is merged.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,117 @@
...
+        // exceptions using generic types. If the mapped type is RuntimeException (unchecked exception), JaxRsProviders
+        // would re-throw the exception without mapping it (or it would have been handled by another mapper instead of
+        // this one), we reproduce this behavior to stay as close as possible to the default implementation (note:
+        // RuntimeException stacktraces are logged by default).
+        if (cause instanceof RuntimeException) {
+            throw (RuntimeException) cause;
+        }
+
+        // Logs the unmapped exception, with its stacktrace.
+        this.logger.error("A REST endpoint failed with an unmapped exception.", cause);
```
vmassol, line 87: Do we agree that if this code is reached it's that something unexpected happened and we can't recover from it? I'm asking to decide if `error` is good or if `warning` should be used. This means that something needs to be fixed and it's not a usage problem.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,117 @@
...
+
+        // Returns an error response to the client, with some detail about the exception in the response.
+        String message;
+        String templateName = "rest/exception.vm";
+        MediaType textPlainType;
+        try {
+            ScriptContext scriptContext = this.scriptContextManager.getScriptContext();
+            scriptContext.setAttribute("cause", cause, ScriptContext.ENGINE_SCOPE);
+            // Make the media type html by default.
```
vmassol, line 96: This just repeat what's on the line below. Better would be to explain why we want HTML.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,117 @@
...
+            textPlainType = MediaType.TEXT_HTML_TYPE;
+            message = this.templateManager.render(templateName);
+        } catch (Exception e) {
+            this.logger.warn("Failed to render the response using template [{}]. Cause: [{}].", templateName,
+                getRootCauseMessage(e));
+            // Fallback to a formatted string when the template rendering fails.
+            // We use a hardcoded \n because the message is used in an HTTP response.
```
vmassol, line 103: Why do we need a `\n` in an HTTP response?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapperTest.java
```diff
@@ -0,0 +1,144 @@
...
+        verifyNoInteractions(this.contextLocalization);
+    }
+
+    @Test
+    void toResponseUncheckedTemplateRenderError() throws Exception
+    {
+        IOException cause = new IOException("file not found");
+        String expectedMessage = String.format(
+            "No ExceptionMapper was found for [java.io.IOException: file not found].\n%s",
+            getStackTrace(cause));
```
vmassol, line 115: Personally I don't like tests that are too clever. Small code duplication is better in tests in general for me when it makes the test more readable and easier to understand. Here I find it hard to see the full error message we're expecting.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/templates/RestExceptionPageTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Tests of the rest Velocity templates.
+ *
+ * @version $Id$
+ * @since 13.8RC1
+ * @since 13.7.1
+ * @since 13.4.4
+ */
+class RestExceptionPageTest extends PageTest
```
vmassol, line 44: Note that it's `REST` and not `Rest` so all classes should use `REST`. Is it because we've used `Rest` in the past in some class names that you've continued?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
tmortagne: Ending up here definitely means unexpected exception for me. Expected/document REST API behavior (so, something an HTTP client can use) is generally supposed to produce standard HTTP code (401, 404, etc.).

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,117 @@
...
+
+        // Returns an error response to the client, with some detail about the exception in the response.
+        String message;
+        String templateName = "rest/exception.vm";
+        MediaType textPlainType;
+        try {
+            ScriptContext scriptContext = this.scriptContextManager.getScriptContext();
+            scriptContext.setAttribute("cause", cause, ScriptContext.ENGINE_SCOPE);
+            // Make the media type html by default.
+            textPlainType = MediaType.TEXT_HTML_TYPE;
```
tmortagne, line 97: This variable should probably be renamed :)

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/templates/RestExceptionPageTest.java
vmassol: Answering my question: we have `XWikiRestComponent` already for example...

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
tmortagne: It's HTTP response, but it's plain text content and not HTML. But I feel that comment should probably be a bit lower, were you actually use `\n`.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/templates/RestExceptionPageTest.java
vmassol: Hard to be consistent now. Same for `HTTP` (my preference) over `Http`. I thought we agreed to use the uppercase form a long time but either it wasn't the case or we didn't pay attention.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
```diff
@@ -0,0 +1,117 @@
...
+            message = this.templateManager.render(templateName);
+        } catch (Exception e) {
+            this.logger.warn("Failed to render the response using template [{}]. Cause: [{}].", templateName,
+                getRootCauseMessage(e));
+            // Fallback to a formatted string when the template rendering fails.
+            // We use a hardcoded \n because the message is used in an HTTP response.
+            String translation =
+                this.contextLocalization.getTranslationPlain("rest.exception.noMapper", getRootCauseMessage(cause));
```
tmortagne, line 105: Given the error message, I would expect the parameter to be the exception class name and not the root cause message (which is printer later in the stack trace).

manuelleduc (commented):
a

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapper.java
manuelleduc: Yes, exceptions thrown by REST endpoints needs to have an exception mapper, either by inheriting an exception with an exception mapper, or by defining a new one. This logs is only displayed when the exception is not handled properly by a REST endpoint.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/test/java/org/xwiki/rest/internal/exceptions/ExceptionExceptionMapperTest.java
manuelleduc: I get your point. I wrote it that way because I'm did not find an simpler way to assert that the log contains a well-formed stacktrace while keeping the test portable (for instance I'm not sure the stacktrace will stay the same on the CI, or if we update one ouf our test libraries).

PR #1687, opened 2021-09-05
XWIKI-18913: Rights applied for a group are not updated

ClemensRobbenhaar:
XWIKI-18913: Rights applied for a group are not updated after multiple renames of the group page

- store group property as LargeStringValue, not StringValue
- also use ListClass.getStringFromList to join the groups
  (which is more of a theoretical issue as the rights UI
  creates a separate RightsClass object for every group)

Also wrote a unit test to check the fix; however I guess this test can be improved a lot.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
@@ -65,7 +65,7 @@ class RightsManagerTest
     {
         rights = RightsManager.getInstance();
 
-        rightsDocReference = new DocumentReference("xwiki", Collections.singletonList("Some"), "Page");
+        rightsDocReference = new DocumentReference("xwiki", "Some", "Page");
         XWikiDocument rightsDocument = loadTestDocumentFromStore();
 
         rightsDocument.newXObject(MockitoOldcore.GLOBAL_RIGHTS_CLASS, oldcore.getXWikiContext());
@@ -80,10 +80,8 @@ class RightsManagerTest
     @Test
     void replaceUserOrGroupFromAllRights() throws Exception
     {
-        DocumentReference oldGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
-            "GroupBefore");
-        DocumentReference newGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
-            "GroupAfter");
+        DocumentReference oldGroupName = new DocumentReference("xwiki", "XWiki", "GroupBefore");
+        DocumentReference newGroupName = new DocumentReference("xwiki", "XWiki", "GroupAfter");
         String oldValue = "XWiki.GroupBefore,XWiki.Some\\,OtherGroup";
         String expectedValue = "XWiki.Some\\,OtherGroup,XWiki.GroupAfter";
 
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
@@ -1,0 +1,162 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.xpn.xwiki.plugin.rightsmanager;
+
+import java.util.Collections;
+import java.util.Locale;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.test.annotation.AllComponents;
+
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+import com.xpn.xwiki.test.MockitoOldcore;
+import com.xpn.xwiki.test.junit5.mockito.InjectMockitoOldcore;
+import com.xpn.xwiki.test.junit5.mockito.OldcoreTest;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
+
+/**
+ * Test for the RightsManager
+ *
+ * @since 13.8
+ * @version $Id$
+ */
+@OldcoreTest
+@AllComponents
+public class RightsManagerTest
+{
+    @InjectMockitoOldcore
+    MockitoOldcore oldcore;
+
+    /**
+     * instance under test
+     */
+    private RightsManager rights;
+
+    @BeforeEach
+    protected void beforeEach() throws Exception
+    {
+        rights = RightsManager.getInstance();
+
+        doReturn(true).when(oldcore.getSpyXWiki()).checkAccess(any(), any(), any());
+    }
+
+    @Test
+    public void testRenameGroup() throws Exception
+    {
+        DocumentReference oldGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
+            "GroupBefore");
+        DocumentReference newGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
+            "GroupAfter");
+        String oldValue = "XWiki.GroupBefore,XWiki.Some\\,OtherGroup";
+        String expectedValue = "XWiki.Some\\,OtherGroup,XWiki.GroupAfter";
+        // ThreadLocal: only to have this final and use it in the answer
+        final ThreadLocal<String> actualValue = new ThreadLocal<String>();
+
+        XWikiDocument rightsDocument = createMockDocument();
+
+        doReturn(Collections.singletonList(rightsDocument)).when(oldcore.getMockStore()).searchDocuments(any(), any(),
+            eq(oldcore.getXWikiContext()));
+
+        BaseObject rightsObject = Mockito.mock(BaseObject.class);
+        doReturn(Collections.singletonList(rightsObject)).when(rightsDocument)
+            .getXObjects(MockitoOldcore.GLOBAL_RIGHTS_CLASS);
+        doReturn(oldValue).when(rightsObject).getLargeStringValue("groups");
+        doAnswer(new Answer<Void>()
+        {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable
+            {
+                actualValue.set((String) invocation.getArgument(1));
+                return null;
+            }
+        }).when(rightsObject).setLargeStringValue(eq("groups"), any());
+
+        rights.replaceUserOrGroupFromAllRights(oldGroupName, newGroupName, false, oldcore.getXWikiContext());
+
+        assertEquals(expectedValue, actualValue.get());
+    }
+
+    @Test
+    public void testRemoveGroup() throws Exception
+    {
+        String oldValue = "XWiki.SomeGroup,XWiki.GroupToRemove,XWiki.Some\\,OtherGroup";
+        String expectedValue = "XWiki.SomeGroup,XWiki.Some\\,OtherGroup";
+        // ThreadLocal: only to have this final and use it in the answer
+        final ThreadLocal<String> actualValue = new ThreadLocal<String>();
+
+        XWikiDocument rightsDocument = createMockDocument();
+
+        doReturn(Collections.singletonList(rightsDocument)).when(oldcore.getMockStore()).searchDocuments(any(), any(),
+            eq(oldcore.getXWikiContext()));
+
+        BaseObject rightsObject = Mockito.mock(BaseObject.class);
+        doReturn(Collections.singletonList(rightsObject)).when(rightsDocument)
+            .getXObjects(MockitoOldcore.GLOBAL_RIGHTS_CLASS);
+        doReturn(oldValue).when(rightsObject).getLargeStringValue("groups");
+        doReturn("").when(rightsObject).getLargeStringValue("users");
+        doAnswer(new Answer<Void>()
+        {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable
+            {
+                actualValue.set((String) invocation.getArgument(1));
+                return null;
+            }
+        }).when(rightsObject).setLargeStringValue(eq("groups"), any());
+
+        rights.removeUserOrGroupFromAllRights("xwiki", "XWiki", "GroupToRemove", false, oldcore.getXWikiContext());
+
+        assertEquals(expectedValue, actualValue.get());
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        assertEquals(expectedValue, actualValue.get());
+    }
+
+    /**
+     * create a mock document that can be saved via the mock store
+     *
+     * @return the mocked document
+     */
+    private XWikiDocument createMockDocument()
```
tmortagne, line 146: When using `MockitoOldcore` we tend to manipulate `XWikiDocument` instances as we usually would (get the document, modify it, save it, etc.) and not create any mock. It's mocking a memory storage of documents (but it's not interpreting queries, so you do need to mock `searchDocuments`).

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Test for the RightsManager
+ *
+ * @since 13.8
+ * @version $Id$
+ */
+@OldcoreTest
+@AllComponents
+public class RightsManagerTest
```
tmortagne, line 54: With junit5 you can get rid of all the `public` keywords.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+import static org.mockito.Mockito.doReturn;
+
+/**
+ * Test for the RightsManager
+ *
+ * @since 13.8
+ * @version $Id$
+ */
+@OldcoreTest
+@AllComponents
```
tmortagne, line 53: This sounds a lot, I doubt you really need all components. Maybe you should enabled just the few you require.

tmortagne:
The fix looks good, just added some unit tests comments.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @BeforeEach
+    protected void beforeEach() throws Exception
+    {
+        rights = RightsManager.getInstance();
+
+        doReturn(true).when(oldcore.getSpyXWiki()).checkAccess(any(), any(), any());
+    }
+
+    @Test
+    public void testRenameGroup() throws Exception
```
vmassol, line 73: Also please follow the naming best practice; there's no need to prefix with `test` since it's already annotated with `@Test`: https://dev.xwiki.org/xwiki/bin/view/Community/Testing/JavaUnitTesting/#HBestpractices

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * @since 13.8
+ * @version $Id$
+ */
+@OldcoreTest
+@AllComponents
+public class RightsManagerTest
+{
+    @InjectMockitoOldcore
+    MockitoOldcore oldcore;
```
vmassol, line 57: missing `private`

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * instance under test
```
vmassol, line 60: Sam note: should start with uppercase and end with  a dot (if the field was public, checkstyle would complain).

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    private RightsManager rights;
+
+    @BeforeEach
+    protected void beforeEach() throws Exception
```
vmassol, line 65: I don't see any exception thrown, is that normal?

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
+
+/**
+ * Test for the RightsManager
+ *
+ * @since 13.8
```
manuelleduc, line 49: Will this PR be applied only on 13.8 (The fix versions field is empty on https://jira.xwiki.org/browse/XWIKI-18913)? Otherwise some additional `@since` annotations might be missing.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public void testRenameGroup() throws Exception
+    {
+        DocumentReference oldGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
+            "GroupBefore");
+        DocumentReference newGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
+            "GroupAfter");
+        String oldValue = "XWiki.GroupBefore,XWiki.Some\\,OtherGroup";
+        String expectedValue = "XWiki.Some\\,OtherGroup,XWiki.GroupAfter";
+        // ThreadLocal: only to have this final and use it in the answer
+        final ThreadLocal<String> actualValue = new ThreadLocal<String>();
```
vmassol, line 82: I don't understand the comment and why you have to make a value ThreadLocal to be returned.

manuelleduc, line 82: We tend to avoid using the `final` keyword for local variables.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
tmortagne: Actually `@since` is totally useless on test classes IMO.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    {
+        rights = RightsManager.getInstance();
+
+        doReturn(true).when(oldcore.getSpyXWiki()).checkAccess(any(), any(), any());
+    }
+
+    @Test
+    public void testRenameGroup() throws Exception
+    {
+        DocumentReference oldGroupName = new DocumentReference("xwiki", Collections.singletonList("XWiki"),
```
vmassol, line 75: small note: it's not a name but a reference ;)

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+        BaseObject rightsObject = Mockito.mock(BaseObject.class);
+        doReturn(Collections.singletonList(rightsObject)).when(rightsDocument)
+            .getXObjects(MockitoOldcore.GLOBAL_RIGHTS_CLASS);
+        doReturn(oldValue).when(rightsObject).getLargeStringValue("groups");
+        doAnswer(new Answer<Void>()
+        {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable
+            {
+                actualValue.set((String) invocation.getArgument(1));
```
tmortagne, line 98: Seems to me you should use `verify()` API instead of this hack.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+                return null;
+            }
+        }).when(rightsObject).setLargeStringValue(eq("groups"), any());
+
+        rights.replaceUserOrGroupFromAllRights(oldGroupName, newGroupName, false, oldcore.getXWikiContext());
+
+        assertEquals(expectedValue, actualValue.get());
```
vmassol, line 105: This looks misleading to me since actualValue is something computed by the test and not the code under tests.

AFAICS this test only verifies that `setLargeStringValue` is called with first parameter being `groups` and a second parameter (any value). If this is what you want to do, you should probably use a `verify` instead.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
tmortagne: It's because of the mock answer implemented below, but this kind of thhing is supposed to use the `verify()` API instead. Anyway, as I said in another comment, we tend to use real document/object instances with `MockitoOldocore`.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
vmassol: hmm my bad the value passed to `setLargeStringValue` is actually captured and set in another variable (`actualValue)` so it looks ok.

vmassol: To be clear the test wants to verify the value passed to the call to `setLargeStringValue`.

vmassol: BTW since the test exercises `replaceUserOrGroupFromAllRights`, it should probably be named `replaceUserOrGroupFromAllRights` instead of `testRenameGroup`, following our best practice at https://dev.xwiki.org/xwiki/bin/view/Community/Testing/JavaUnitTesting/#HBestpractices

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,162 @@
...
+        {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable
+            {
+                actualValue.set((String) invocation.getArgument(1));
+                return null;
+            }
+        }).when(rightsObject).setLargeStringValue(eq("groups"), any());
+
+        rights.removeUserOrGroupFromAllRights("xwiki", "XWiki", "GroupToRemove", false, oldcore.getXWikiContext());
```
vmassol, line 136: And this test named `removeUserOrGroupFromAllRights`.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
ClemensRobbenhaar: I think it should be merged to all supported branches; I just did not add a proper "fixVersion" to the jira issue ... 
Ok, I will remove the ``@since`` here as it only caused be headaches about which values to set.

ClemensRobbenhaar:
Thanks for the lots of comments in the review! I will look into them this evening, when I have escaped from all meetings.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
ClemensRobbenhaar: I agree in principle, but in practice I know nearly nothing about the class under test and would have to figure out which  dependencies are actual required. Also the actual tests run in less than 8 seconds (stand alone) or even 0.1s if run with the others, compared to the approx 30 seconds needed for the complete ``mvn  -Dtest=RightsManagerTest  test``. If it is acceptable I prefer to keep it as it is.

ClemensRobbenhaar:
Thanks again for all the pointers. I have updates the test, which now looks quite a bit different, making most line comments outdated. I hope it has also been improved sufficiently.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Instance under test.
+     */
+    private RightsManager rights;
+
+    /**
+     * A rights object storing e.g. group information.
+     */
+    private BaseObject rightsObject;
```
tmortagne, line 62: This might be a bit fragile since the document might be cloned in the MockitoOldCore storage in the future (like it is in the real one), never knows. Would be safer to re-get the document from the "store" and access the actually saved object after replaceUserOrGroupFromAllRights/removeUserOrGroupFromAllRights calls.

ClemensRobbenhaar: Updated the tests to always use the document and object from the spyXWiki instance. I hope this is better now.

tmortagne: Yes, looks good.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/plugin/rightsmanager/RightsManagerTest.java
```diff
@@ -0,0 +1,122 @@
...
+     * A reference to the document storing the rights object under test.
+     */
+    private DocumentReference rightsDocReference;
+
+    @BeforeEach
+    void beforeEach() throws Exception
+    {
+        rights = RightsManager.getInstance();
+
+        rightsDocReference = new DocumentReference("xwiki", Collections.singletonList("Some"), "Page");
```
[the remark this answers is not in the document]
ClemensRobbenhaar: Ah, I thought that one is deprecated, or at least "out of style".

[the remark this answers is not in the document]
tmortagne: No reason, it's still a very common use case.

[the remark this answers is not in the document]
ClemensRobbenhaar: Ok, updated the three occurrences of the `DocumentReference` constructor.

ClemensRobbenhaar:
Uh, I botched up by trying to prevent the merge commit that git claimed to need to merge, and instead made a `git rebase master` in the branch before the merge.
That worked on the command line but changed all the commit hashes and said it added all commits from master into the branch.
Merging still worked, but I do not feel like I should push the state of this branch from my local repository to anywhere.

Anyway, the commits from the merged branch can be found e.g. by
 `git log af005ab7a03e91d732686a604f3a~..58d5a61b6858043b35c3dcf4f`

Closing the PR as it has been merged.

PR #1689, opened 2021-09-21
XWIKI-18910: Errors appear in the wiki console after selecting the "Forgot your username or password option", on a Multilingual wiki

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-18910

Proposed solution:
  * Ensure to display the language menu only if $doc is defined in the
    current context.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
@@ -120,7 +120,7 @@
 #end
 
 ## Check that the doc velocity variable is defined: it might not be the case when using some resource handlers.
-#set ($docIsDefined = ("$!doc" != ''))
+#set ($docIsDefined = ($doc != $NULL))
 
 #set($displayContentMenu = true) ## menu on the top of the page with create/edit/more actions
 #set($displayCreateMenu = true) ## create menu
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
@@ -135,7 +135,7 @@
 #set($displayPageHeader = false) ## page header containing the logo and the banner taken from ColorTheme
 #set ($displayQuickSearch = true)
 
-## the language choice only makes sense when a doc is declared.
+## The language choice only makes sense when a doc is declared.
 #set ($displayLanguageChoice = $docIsDefined)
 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
```diff
@@ -119,6 +119,9 @@
  #end
 #end
 
+## Check that the doc velocity variable is defined: it might not be the case when using some resource handlers.
+#set ($docIsDefined = ("$!doc" != ''))
```
tmortagne, line 123: Not sure why you compare it to an empty String, a simple `#if ($doc)` (which check if it's not null) would be enough IMO. Feels more accurate than calling XWikiDocument#toString().

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
```diff
@@ -131,7 +134,9 @@
 #set($displayShortcuts = true)
 #set($displayPageHeader = false) ## page header containing the logo and the banner taken from ColorTheme
 #set ($displayQuickSearch = true)
-#set ($displayLanguageChoice = true)
+
+## the language choice only makes sense when a doc is declared.
```
manuelleduc, line 138: Minor :)
```suggestion
## The language choice only makes sense when a doc is declared.
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/layoutvars.vm
surli: I'm using this pattern when I want a one-liner, it avoids
```
#if ($doc)
  #set ($docIsDefined = true)
#else 
  #set ($docIsDefined = false)
#end 
```
but if you have a better way for that kind of stuff I'm taking it.

surli: hmm maybe I could use `$doc != $NULL`?

tmortagne: That should work, another possibility is `#set ($docIsDefined = !!$doc)` (putting a `!` in front of a variable in a set cause the set to set as value the boolean result of the if and then you invert it again with another `!`).

surli: Hmm I prefer `$doc != $NULL` it's easier to read IMO.

manuelleduc (approved):

PR #1696, opened 2021-10-04
XWIKI-16544: getdocuments and Livetable results filtering improvement

manuelleduc:
- Mark the non-viewable documents as 'obfuscated'
- When the number of results is lower than the limit, removes the obfuscated documents from the response
- Improve Livedata to inform the user that they cannot view the content of a row because they do not have the right


xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/Livetable/Code/Translations.xml
```diff
@@ -0,0 +1,77 @@
...
+  <parent>WebHome</parent>
+  <author>xwiki:XWiki.Admin</author>
+  <contentAuthor>xwiki:XWiki.Admin</contentAuthor>
+  <version>1.1</version>
+  <title>Translations</title>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>plain/1.0</syntaxId>
+  <hidden>true</hidden>
+  <content>livetable.cell.emptyValue=N/A</content>
```
manuelleduc, line 39: New translation key for livetable. They rest of the keys seems to be placed in a deprecated location. I hope it was the right place to add this.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/test/java/org/xwiki/livetable/LiveTableResultsTest.java
```diff
@@ -66,12 +74,20 @@
  * @version $Id$
  */
 @XWikiSyntax20ComponentList
+@ComponentList({
+    // SecurityScriptService
+    SecurityScriptService.class,
+    RightConverter.class,
+    EntityReferenceConverter.class,
```
manuelleduc, line 81: To be moved in its own ComponentList class.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -452,6 +452,13 @@ XWiki.widgets.LiveTable = Class.create({
           }
           container.update(text.escapeHTML());
         }
+
+        // When a cell is part of a non-viewable row, we display a content indicating that the content is not present
+        // because the current user does not have the right to view it.
+        if (!row['doc_viewable'] && !row[column]) {
+          td.update($(table.domNodeName + '-inaccessible-cell').innerHTML + '<sup>*</sup>')
+        }
```
manuelleduc, line 460: @mflorea Does this change looks alright to you? Also, I'm not sure of the best way to test this change, wdyt is the right way to do it?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-script/src/test/java/org/xwiki/security/script/SecurityScriptServiceComponentList.java
```diff
@@ -0,0 +1,57 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Pack of default Component implementations that are needed for {@link SecurityScriptService}.
+ *
+ * @version $Id$
+ * @since 13.9RC1
+ * @since 13.8.1
```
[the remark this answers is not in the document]
manuelleduc: Removed the fix version and the since annotations.

xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -342,7 +342,21 @@
   #foreach($item in $items)
     #gridresult_buildRowJSON($item $rows)
   #end
-  #set($discard = $map.put('rows', $rows))
+  ## When the total number of rows is not bigger than the limit, we remove the obfuscated results from the
```
tmortagne, line 345: Feels like this part (and maybe more I did not chech) should be moved into a macro in some template and used both here and in getdocument.vm.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetable.js
mflorea: You need to use ``fieldName`` instead of ``column``. Then the field value can be boolean (or empty string), so checking for ``!row[fieldName]`` is not right. You need to either check if the row has the key, using ``hasOwnProperty``, or check for ``null`` or ``undefined`` explicitly.

Moreover, if you replace the content of the cell entirely then the execution of the previous code is useless. It doesn't make sense to try to compute the link or the cell text (including special handling of the ``doc.name`` column when the row is not viewable) if in the end you overwrite all that. This code needs a bit more refactoring.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -353,10 +353,26 @@ XWiki.widgets.LiveTable = Class.create({
...
       }
-      // The column's display name to be used when displaying the reponsive version.
+      // The column's display name to be used when displaying the responsive version.
       var displayName = descriptor.displayName || column;
       var fieldName = column.replace(/^doc\./, 'doc_');
-      if (column === '_actions') {
+      // When a cell is part of a non-viewable row (except for the actions that simply stays empty), we display a 
+      // message indicating that the content is not present because the current user does not have the right to view it.
+      if (!row['doc_viewable'] && !row[fieldName]) {
+        const notViewableCellMessage = "$services.localization.render('livetable.cell.emptyValue')";
```
mflorea, line 362:
```suggestion
        const notViewableCellMessage = $jsontool.serialize($services.localization.render('livetable.cell.emptyValue'));
```

In order to make sure the (injected) translation message doesn't break the JavaScript code.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetable.js
```diff
@@ -353,10 +353,26 @@ XWiki.widgets.LiveTable = Class.create({
...
+        var td = new Element('td', {
+          'class': [
+            fieldName,
+            'link' + (descriptor.link || ''),
+            'type' + (descriptor.type || '')
+          ].join(' '),
+          'data-title': displayName
+        });
+        if(column !== '_actions') {
```
mflorea, line 371:
```suggestion
        if (column !== '_actions') {
```

PR #1697, opened 2021-10-08
XWIKI-18758: Replace the Livetable from the Data types page with a Live Data

michitux:
This replaces the Livetable from the Data types page (XWikiClasses) with a Live Data. Further, this contains a few related fixes:

* add a missing dependency to xwiki-platform-localization-macro for the translation macro used in the document.
* Fix the TableLayoutElement PageObject to not wait if no rows can be loaded because the result is empty.

JIRA issue: https://jira.xwiki.org/browse/XWIKI-18758

Before:
Data types - XWiki-18758-Before

After:
Data types - XWiki-18758-After


manuelleduc:
> Delete action icon not in red. Is that normal?

This actions are using the icons for the current icon set. In Font Awesome they are not colored.

> LD headings are not left aligned, is that normal? It would seem better on the left no?

The column labels are slightly moved to the right. When the user hover the column, an icon becomes visible and allows to drag columns to reorganize them. 

> In LT we were careful to not break the text on words AFAIR (and it can be seen on the screenshots above in lots of places like the "r" in "Administrator"). Shouldn't we do the same for LD? (feels nicer)

I agree, that's something we can improve.

> Why is the document name displayed in grey in LD (in the location column)? Visually it feels like it's not clickable.

The style how the html returned by the source is not customized when displayed in LD. We could do the same as in LT and customize the style.

see https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetable.css#L402-L404

```css
.xwiki-livetable .breadcrumb .active a{
  color: $theme.linkColor;
}
```

> I feel it was nicer to have the dates displayed on a single line as in LT but YYMV. The real question is more: why do we use a different layout style than in LT? Was it done voluntarily?

I don't think that's done voluntarily, that's also something we can improve.

PR #1700, opened 2021-10-12

oanalavinia:
* the server will send to the client the start and end offset for an annotation, relatively to the plain content;
* html content (from inside a raw block) is also handled by trying to parse it and get the plain text
* on the client side, a DOM Range corresponding to each annotation is computed using the plain text offset and the annotation is marked on the content

### DIFF
--- a/xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
+++ b/xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
@@ -1097,8 +1097,8 @@ XWiki.Annotation = Class.create({
    * @param endIndex end offset where the range ends
    */
   getDOMRange: function(annotatedElement, startIndex, endIndex) {
-    var startPosition = this.getTextNodeAtPlainTextOffset(annotatedElement, startIndex);
-    var endPosition = this.getTextNodeAtPlainTextOffset(annotatedElement, endIndex);
+    var startPosition = this.getTextNodeAtPlainTextOffset(annotatedElement, startIndex, true);
+    var endPosition = this.getTextNodeAtPlainTextOffset(annotatedElement, endIndex, false);
     var range = new Range();
     range.setStart(startPosition.node, startPosition.offset);
     range.setEnd(endPosition.node, endPosition.offset);
@@ -1112,8 +1112,9 @@ XWiki.Annotation = Class.create({
    *
    * @param parentNode the node where the search is done
    * @param plainTextOffset the offset relative to the full plain content
+   * @param isStart boolean specifying if a start or end offset is targeted
    */
-  getTextNodeAtPlainTextOffset: function(parentNode, plainTextOffset) {
+  getTextNodeAtPlainTextOffset: function(parentNode, plainTextOffset, isStart) {
     var childNodes = parentNode.childNodes;
     var child;
     var parentNodePlainTextLength = 0;
@@ -1123,13 +1124,18 @@ XWiki.Annotation = Class.create({
         var previousSiblingsLength = parentNodePlainTextLength;
         // The spaces are ignored since they were removed as well on the server when the offset was computed.
         parentNodePlainTextLength += child.textContent.replace(/\s/g, '').length;
-        if (plainTextOffset &lt;= parentNodePlainTextLength) {
+        // Consider that an end offset is exclusive.
+        if ((isStart &amp;&amp; plainTextOffset &lt; parentNodePlainTextLength) ||
+            (!isStart &amp;&amp; plainTextOffset &lt;= parentNodePlainTextLength)) {
           // Because plainTextOffset doesn't consider spaces, the real offset of the node needs to be recomputed so that
           // they are included.
-          return {'node': child, 'offset': this.getNodeSpecificOffset(child, plainTextOffset - previousSiblingsLength)};
+          return {
+            'node': child,
+            'offset': this.getNodeSpecificOffset(child, plainTextOffset - previousSiblingsLength, isStart)
+          };
         }
       } else if (child.childNodes.length &gt; 0) {
-        var maybeFoundNode = this.getTextNodeAtPlainTextOffset(child, plainTextOffset - parentNodePlainTextLength);
+        var maybeFoundNode = this.getTextNodeAtPlainTextOffset(child, plainTextOffset - parentNodePlainTextLength, isStart);
         if (maybeFoundNode.node) {
           return maybeFoundNode;
         } else {
@@ -1146,18 +1152,18 @@ XWiki.Annotation = Class.create({
    *
    * @param node a DOM node
    * @param offset the offset relative to the text without whitespaces
+   * @param isStart boolean specifying if a start or end offset is targeted
    */
-  getNodeSpecificOffset: function(node, offset) {
+  getNodeSpecificOffset: function(node, offset, isStart) {
     var nonSpaceCharsLength = 0;
     var chars = Array.from(node.textContent);
     for (let i = 0; i &lt; chars.length; i++) {
-      if (/\s/.test(chars[i])) {
-        continue;
-      } else {
-        if (nonSpaceCharsLength === offset) {
-          return i;
-        }
-        nonSpaceCharsLength++;
+      // Because the end offset is exclusive, it can be a whitespace.
+      if (offset == 0 &amp;&amp; (!isStart || (isStart &amp;&amp; !/\s/.test(chars[i])))) {
+        return i;
+      }
+      if (!/\s/.test(chars[i])) {
+        offset--;
       }
     }
     return chars.length;


xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-core/src/main/java/org/xwiki/annotation/Annotation.java
```diff
@@ -84,6 +84,16 @@
      */
     public static final String TARGET_FIELD = "target";
 
+    /**
+     * The name of the field of this annotation start offset relative to the plain text.
+     */
```
tmortagne, line 89: Don't forget to indicate `@since` when this new API is available before merging it.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-core/src/main/java/org/xwiki/annotation/internal/renderer/AnnotationGeneratorChainingListener.java
```diff
@@ -156,11 +162,18 @@ private void handleRawText(String text)
     @Override
     public void onRawText(String text, Syntax syntax)
     {
+        String textContent = text;
+        if (syntax.getType().equals(SyntaxType.XHTML) || syntax.getType().equals(SyntaxType.HTML)) {
```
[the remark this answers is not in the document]
oanalavinia: I opened a PR for this https://github.com/xwiki/xwiki-rendering/pull/187

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+  /**
+   * For the first and last node, split the nodes at the known offset to not annotate the whole text. Create a new range
+   * with these new nodes.
+   *
+   * @param range the DOM range of the annotated text
+  */
+  fixRangeEnds: function(range) {
+    var strictRange = new Range();
+
+    var newStartNode = range.startContainer.splitText(range.startOffset);
```
mflorea, line 1048: Would be nice to check if the offset is within the text node (not before the first or after the last character), otherwise the split will create useless empty text nodes.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+    strictRange.setStart(newStartNode, 0);
+
+    var newEndNode = range.endContainer.splitText(range.endOffset);
+    strictRange.setEnd(newEndNode, 0);
```
mflorea, line 1052: This means the new range touches ("includes") the text after the split even if that text is not annotated. It would be more semantic to place the range end point before the split:

    strictRange.setEndBefore(newEndNode);

You could even do the same with the start:

    strictRange.setStartBefore(newStartNode);

But note that you have to handle a **tricky** use case: when the initial / input range starts and ends in the **same** text node. In this case, the first split will change the length of the start/end text node and this will break the second split.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+  /**
+   * Create a DOM Range by knowing the start and end index from inside the plain content of the element.
+   *
+   * @param annotatedElement the element from where the range is constructed
+   * @param startIndex start offset where the range begins
+   * @param endIndex end offset where the range end
+  */
+  getDOMRange: function(annotatedElement, startIndex, endIndex) {
+    this.annotatedElementPlainText = "";
+    var startPosition = this.getElementAtOffset(annotatedElement, startIndex);
```
mflorea, line 1066: You don't need to store / keep ``annotatedElementPlainText`` if you remove the length from the ``startIndex`` as you call the method recursively. Same below for ``endIndex``.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+
+  /**
+   * Knowing the offset relative to the full plain content of the root element, get the corresponding child node that contains it and the offset
+   * specific to the new node. The DOM is traversed recursively starting with the root node and the plain content is computed to know when the
+   * wanted element is found.
+   *
+   * @param rootElement the element where the search is done
+   * @param offset the offset relative to the full plain content
+  */
+  getElementAtOffset: function(rootElement, offset) {
```
mflorea, line 1083: "element" has a very precise meaning in the context of DOM processing and it never means "text node". Text nodes and elements are two different types of nodes and you shouldn't mix their semantic.

```suggestion
  getTextNodeAtPlainTextOffset: function(parentNode, plainTextOffset) {
```

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+    var childNodes = rootElement.childNodes;
+    var child;
+    for (let i = 0; i &lt; childNodes.length; i++) {
+      child = childNodes[i];
+      if (child &amp;&amp; child.nodeType == 3) {
+        var initialLength = this.annotatedElementPlainText.length;
+        this.annotatedElementPlainText += child.textContent.replace(/\s/g, '');
```
mflorea, line 1090: You need to count the length of plain text within the given parent as you go and remove it from the plain text offset when calling the function recursively. In any case, the replace above requires a comment as it's not obvious why you need it.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+        if (offset &lt; this.annotatedElementPlainText.length) {
+          var elementSpecificOffset = this.getElementSpecificOffset(child, offset - initialLength);
+          return {"element": child, 'offset': elementSpecificOffset};
+        }
+      }
+      if (child.childNodes.length &gt; 0) {
+        var maybeFoundElement = this.getElementAtOffset(child, offset);
```
mflorea, line 1097: Here you need to remove from the offset the current length of plain text (within the current parent node).

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+   * Get a offset knowing the offset relatively to the plain text of the element.
+   *
+   * @param node a DOM node
+   * @param offset the offset relative to the plain text
+  */
+  getElementSpecificOffset: function(node, offset) {
+    var nonSpaceCharsLength = 0;
+    var chars = Array.from(node.textContent);
+    for (let i = 0; i &lt; chars.length; i++) {
+      if (!/\s/.test(chars[i])) {
```
mflorea, line 1115: I don't understand why you need to ignore spaces.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +977,193 @@ XWiki.Annotation = Class.create({
...
+   *
+   * @param range a DOM Range
+  */
+  getTextNodesInRange: function(range) {
+    var rangeIterator = document.createNodeIterator(
+      range.commonAncestorContainer,
+      NodeFilter.SHOW_TEXT,
+      {
+        acceptNode: function (node) {
+          // Only accept nodes that have content other than whitespace.
```
mflorea, line 1135: Why?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-core/src/main/java/org/xwiki/annotation/internal/renderer/HTMLTextExtracter.java
```diff
@@ -0,0 +1,131 @@
...
+    public String getTextContent(String html)
+    {
+        String fullContent = "";
+        Document htmlDoc = parseHTML(html);
+        XPath xPath = XPathFactory.newInstance().newXPath();
+        try {
+            NodeList textNodes = (NodeList) xPath.compile("//text()").evaluate(htmlDoc, XPathConstants.NODESET);
+            for (int i = 0; i < textNodes.getLength(); i++) {
+                String textContent = textNodes.item(i).getTextContent();
+                fullContent = fullContent.concat(textContent);
```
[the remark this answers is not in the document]
tmortagne: +1, this is kind of the typical `StringBuilder` use case.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
oanalavinia: Since the given offset is relative to the plain content of the text node. What I am trying is to find where that offset is considering also the spaces.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +976,239 @@ XWiki.Annotation = Class.create({
...
+    // Get the Range start and end as text nodes.
+    var startContainer = range.startContainer;
+    if (startContainer.childNodes.length &gt; 0) {
+      startContainer = startContainer.childNodes[range.startOffset];
+    }
+    var endContainer = range.endContainer;
+    if (endContainer.childNodes.length &gt; 0) {
+      // The node at the endOffset position is the boundary point of the Range.
+      endContainer = endContainer.childNodes[range.endOffset - 1];
+    }
```
mflorea, line 1188: The way you compute the start and end container is error-prone. I'd rather use https://developer.mozilla.org/en-US/docs/Web/API/Range/compareBoundaryPoints inside the range iterator loop to check if the current node is inside the range.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +976,239 @@ XWiki.Annotation = Class.create({
...
+    // The split is done only if the offset is not before first or after last character for not creating empty text nodes.
+    if (range.startOffset &gt; 0 &amp;&amp; range.startOffset &lt; (range.startContainer.length - 1)) {
+      newStartNode = range.startContainer.splitText(range.startOffset);
+    }
+    strictRange.setStartBefore(newStartNode);
+
+    var newEndNode = range.endContainer;
+    // The split is done only if the offset is not before first or after last character for not creating empty text nodes.
+    if (endOffset &gt; 0 &amp;&amp; endOffset &lt; (range.endContainer.length - 1)) {
+      newEndNode = range.endContainer.splitText(endOffset + 1);
```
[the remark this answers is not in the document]
oanalavinia: I think I misunderstood the meaning of the `Range.endOffset`. In my case, `range.endOffset` is the index of the last character that should be included in the annotation. If I have `foo` annotated, then `endOffset = 2` and `splitText(2)` will put compute the nodes `fo` and `o` 
Because of this `endOffset + 1` I added also the check of `endOffset < (range.endContainer.length - 1)`

But after checking again the documentation, in the above example the `endOffset` should have been 3. So the problem could be solved  by adding here a `+1` https://github.com/xwiki/xwiki-platform/pull/1700/files#diff-6fae9d69c9935f88c56cf7c9fde0f2f98b2cee09a3810b706400d3c1b007f437R1099  (I think Script.xml diff should be expanded for the link above to work, it's not done by default)
@mflorea WDYT?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -976,28 +976,239 @@ XWiki.Annotation = Class.create({
...
+    return nodes;
+  },
+
+  /**
+   * Remove the wrapper and marker of annotations. The selection highlight is also removed in case the new annotation
+   * was deleted.
+  */
+  removeAnnotationsAndSelectionMarkups: function() {
+    document.querySelectorAll("span.annotation, span.selection-highlight")
+      .forEach(annotationNode =&gt; annotationNode.replaceWith(...annotationNode.childNodes));
```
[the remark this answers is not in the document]
mflorea: @surli see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax .

PR #1703, opened 2021-10-13
XWIKI-19044: Provide real property names in Diff html as data attributes

surli:
## JIRA

https://jira.xwiki.org/browse/XWIKI-19044

## Solution

  * Provide two new data-attributes in diff UI: data-reference when
    starting the diff details of an element, providing the reference of
the element and data-property-name containing the real name of a
property.
  * Display back the serialized reference of the displayed element as
    hint in the diff (was previously commented)

## Screenshots

There's not much to see in the screenshots except for the hints that were previously hidden.

Document diff before:
diff-page-before

Document diff after:
diff-page-after

XClass diff before:
diff-xclass-before
 
XClass diff after:
diff-xclass-after

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/diff_macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/diff_macros.vm
@@ -145,9 +145,6 @@
     <dt id="#diffAnchor($documentReference)" #diffDataReference($documentReference)>
       #diffIcon('file-text' 'change')
       $services.localization.render('web.history.changes.summary.documentProperties')
-      <span class="hint">
-        $escapetool.xml($services.model.serialize($documentReference, 'local'))
-      </span>
     </dt>
     <dd>
       <dl>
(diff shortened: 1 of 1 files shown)


manuelleduc:
Sorry to repeat myself but that's the perfect change to start using [page tests](https://dev.xwiki.org/xwiki/bin/view/Community/Testing/ViewUnitTesting/). 
Also, wdyt of adding some documentation of the newly introduced parameters?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/diff_macros.vm
```diff
@@ -142,11 +142,11 @@
     #if (!$documentReference)
       #set ($documentReference = $newDoc.documentReference)
     #end
-    <dt id="#diffAnchor($documentReference)">
+    <dt id="#diffAnchor($documentReference)" #diffDataReference($documentReference)>
       #diffIcon('file-text' 'change')
       $services.localization.render('web.history.changes.summary.documentProperties')
       <span class="hint">
-        ##$escapetool.xml($services.model.serialize($documentReference, 'local'))
+        $escapetool.xml($services.model.serialize($documentReference, 'local'))
```
mflorea, line 149: I don't see the value in displaying the document reference here. It looks bad and it creates confusion (we don't display document references in the XWiki UI). When the diff is for a single page then it should be obvious that the changed page properties are of that page. When the diff is for multiple pages (like when we compute the changes for an installed XAR extension) each page gets a heading so it should be easy to know whose page properties are shown in the diff. I think we should simply remove the hint.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/diff_macros.vm
```diff
@@ -292,11 +292,11 @@
       #set ($classReference = $services.model.resolveDocument($className))
       #set ($classPropertyReference = $services.model.createEntityReference($propertyDiff.propName, 'CLASS_PROPERTY',
         $classReference))
-      <dt id="#diffAnchor($classPropertyReference)">
+      <dt id="#diffAnchor($classPropertyReference)" #diffDataReference($classPropertyReference)>
         #diffIcon('cube' $propertyDiff.action)
         $propertyDiff.propName
         <span class="hint">
-          ##($escapetool.xml($className))
+          ($escapetool.xml($className))
```
mflorea, line 299: I don't like this either. If need be, we could improve the separation between page properties, object properties and class properties. I'm fine to display the reference on hover (using the ``title`` attribute), but I don't think we should repeat the class reference after each property name.

PR #1704, opened 2021-10-14
[Misc] Add a zip slip check when loading the configurations from a jar

manuelleduc:


xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/PackageMojo.java
```diff
@@ -367,6 +367,15 @@ private void generateConfigurationFiles(File configurationFileTargetDirectory) t
                 if (entry.getName().endsWith(".vm")) {
                     String fileName = entry.getName().replace(".vm", "");
                     File outputFile = new File(configurationFileTargetDirectory, fileName);
+
+                    // Validates that the outFile location is located inside the configuration directory.
```
vmassol, line 371: s/outFile/outputFile

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/test/java/org/xwiki/tool/packager/PackageMojoTest.java
```diff
@@ -57,4 +81,35 @@ public void replaceProperty()
...
+        ReflectionUtils.setFieldValue(mojo, "hibernateConfig", mock(File.class));
+        ReflectionUtils.setFieldValue(mojo, "project", project);
+        ReflectionUtils.setFieldValue(mojo, "repositorySystem", repositorySystem);
+        // Initialize a temporary directory otherwise the configuration files are added to the project folders.
+        ReflectionUtils.setFieldValue(mojo, "outputPackageDirectory", this.tmpDir);
+
+        MojoExecutionException exception = assertThrows(MojoExecutionException.class, mojo::before);
+        assertEquals(IOException.class, exception.getCause().getClass());
+        assertEquals(String.format(
+                "Template [%1$s/webapps/xwiki/WEB-INF/../test] is outside of the configuration file directory "
```
vmassol, line 111: what is what you call the "configuration file directory"? Are you referring to the webapp `WEB-INF`directory?

manuelleduc: > what is what you call the "configuration file directory"? Are you referring to the webapp `WEB-INF`directory?

Yes, I reused the vocabulary used in `generateConfigurationFiles` but a better term could exist :)

> Note: I would create the JAR as part of the test since it's not great to include a binary that we cannot easily recreate and it's easy to crate a jar in the test.

I'll see if I can produce a "malformed" zip programmatically.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/PackageMojo.java
vmassol: ok I've now read https://sonarcloud.io/organizations/xwiki/rules?open=javasecurity%3AS6096&rule_key=javasecurity%3AS6096  and I don't think it applies to us since we control the JAR input. I would remove the unnecessary check which adds code, requires explanation as to why we need it, and costs execution time. Instead I would just add a comment to explain why we don't need a zip slip check, and add an ignore annotation (https://docs.codescan.io/hc/en-us/articles/360012109711-Ignoring-violations)

manuelleduc: ok, I'll update the PR accordingly.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/PackageMojo.java
```diff
@@ -351,6 +351,21 @@ private Artifact resolveArtifactItem(ArtifactItem artifactItem) throws MojoExecu
...
+     * and interpret them before copying the result to {@code configurationFileTargetDirectory}.
+     * <p>
+     * <a 
+     *   href="https://sonarcloud.io/organizations/xwiki/rules?open=javasecurity%3AS6096&rule_key=javasecurity%3AS6096">
+     *   javasecurity:S6096
+     * </a> is ignored because we trust the content of {@code xwiki-platform-tool-configuration-resources} since it is 
+     * produced by maven using source code we own.
+     *
+     * @param configurationFileTargetDirectory the root directory where the configuration files found in {@code
+     *     xwiki-platform-tool-configuration-resources} are copied ({@code "webapps/xwiki/WEB-INF"} by default)
```
vmassol, line 365: Note: We shouldn't put quotes inside `@code` annotation

vmassol: Ok, what I meant by comment was simpler (no need for javadoc especially since it's an impl detail) ;) It was simply a comment where we extract the zip to say that we don't need to verify if there's no relative path to prevent against zip slip attacks since we control the JAR source.

manuelleduc: Thanks :) I'll add an inline comment too, I feel like the "zip slip" keyword is missing somewhere.

PR #1707, opened 2021-10-19
XWIKI-19048: Live data list filter should display and filter by (translated) labels for static list properties

michitux:
This uses the view displayer to populate the metadata that is returned by the REST service.

Jira issue: https://jira.xwiki.org/browse/XWIKI-19048

I would like to get feedback on two questions here:

* The unit test (`StaticListClassPropertyValuesProviderTest`) is not just testing `StaticListClassPropertyValuesProvider` but also its interaction with `ListClass` using an actual `ListClass` object and not a mock. This really helped me during the development to ensure that I'm correctly using the interface and, e.g., handling the XML-encoded result value correctly. However, this is kind of against the idea of an isolated unit test. What do you think, is this kind of test okay or should I change it to, e.g., mock the results of the `ListClass`-methods?
* The implementation of `StaticListClassPropertyValuesProvider::constructPropertyValueForId` is currently quite ugly as it uses the public method `ListClass::displayView`. This involves constructing an additional object with a property just to fulfill the requirements of the parameters and XML-unescaping the result. Its code could be simplified a lot using [`ListClass::getDisplayValue`](https://github.com/xwiki/xwiki-platform/blob/dcfb67ad6b5b78d7d506760ce78d9a3d44b125af/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java#L625), which is currently protected or private, depending on the variant. This method could be used just with the values we already have and a map from values to labels that we can easily obtain and also doesn't require unescaping the result. What do you think about making one of the variants of `ListClass::getDisplayValue` public?

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -321,6 +321,19 @@
                 </item>
               </differences>
             </revapi.differences>
+            <revapi.differences>
+              <justification>Not a breakage, new annotation on existing protected and now public method.</justification>
+              <criticality>allowed</criticality>
+              <ignore>true</ignore>
+              <differences>
+                <item>
+                  <code>java.annotation.added</code>
+                  <old>method java.lang.String com.xpn.xwiki.objects.classes.ListClass::getDisplayValue(java.lang.String, java.lang.String, java.util.Map&lt;java.lang.String, com.xpn.xwiki.objects.classes.ListItem&gt;, com.xpn.xwiki.XWikiContext) @ com.xpn.xwiki.objects.classes.StaticListClass</old>
+                  <new>method java.lang.String com.xpn.xwiki.objects.classes.ListClass::getDisplayValue(java.lang.String, java.lang.String, java.util.Map&lt;java.lang.String, com.xpn.xwiki.objects.classes.ListItem&gt;, com.xpn.xwiki.XWikiContext) @ com.xpn.xwiki.objects.classes.StaticListClass</new>
+                  <annotation>@org.xwiki.stability.Unstable</annotation>
+                </item>
+              </differences>
+            </revapi.differences>
           </analysisConfiguration>
         </configuration>
       </plugin>
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -321,19 +321,6 @@
                 </item>
               </differences>
             </revapi.differences>
-            <revapi.differences>
-              <justification>Not a breakage, new annotation on existing protected and now public method.</justification>
-              <criticality>allowed</criticality>
-              <ignore>true</ignore>
-              <differences>
-                <item>
-                  <code>java.annotation.added</code>
-                  <old>method java.lang.String com.xpn.xwiki.objects.classes.ListClass::getDisplayValue(java.lang.String, java.lang.String, java.util.Map&lt;java.lang.String, com.xpn.xwiki.objects.classes.ListItem&gt;, com.xpn.xwiki.XWikiContext) @ com.xpn.xwiki.objects.classes.StaticListClass</old>
-                  <new>method java.lang.String com.xpn.xwiki.objects.classes.ListClass::getDisplayValue(java.lang.String, java.lang.String, java.util.Map&lt;java.lang.String, com.xpn.xwiki.objects.classes.ListItem&gt;, com.xpn.xwiki.XWikiContext) @ com.xpn.xwiki.objects.classes.StaticListClass</new>
-                  <annotation>@org.xwiki.stability.Unstable</annotation>
-                </item>
-              </differences>
-            </revapi.differences>
           </analysisConfiguration>
         </configuration>
       </plugin>
--- a/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/StaticListClassPropertyValuesProvider.java
+++ b/xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/StaticListClassPropertyValuesProvider.java
@@ -24,10 +24,13 @@ import java.util.List;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import org.apache.commons.text.StringEscapeUtils;
 import org.xwiki.component.annotation.Component;
 import org.xwiki.rest.model.jaxb.PropertyValue;
 import org.xwiki.rest.model.jaxb.PropertyValues;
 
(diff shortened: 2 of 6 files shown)


xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/StaticListClassPropertyValuesProvider.java
```diff
@@ -49,13 +51,64 @@
...
+                && val.getMetaData().get(META_DATA_LABEL).toString().toLowerCase().contains(lowerFilter)))
+            .limit(limit).forEach((result::withPropertyValues));
 
-        allValues.stream().filter(s -> s.toLowerCase().contains(filter.toLowerCase())).limit(limit)
-            .forEach(item -> result.withPropertyValues(new PropertyValue(item)));
         return result;
     }
+
+    /**
+     * {@inheritDoc}
```
surli, line 72: From what I see there's actually no doc in `AbstractClassPropertyValuesProvider` for it, so maybe it would worth it to add some doc there. Moreover since you override the implem it might also worth it to explain why you override it here.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/StaticListClassPropertyValuesProvider.java
```diff
@@ -49,13 +51,64 @@
...
     {
         final PropertyValues result = new PropertyValues();
         List<String> allValues = StaticListClass.getListFromString(propertyDefinition.getValues());
+        Map<String, ListItem> valueMap = StaticListClass.getMapFromString(propertyDefinition.getValues());
+
+        String lowerFilter = filter.toLowerCase();
+
+        allValues.stream().map(id -> constructPropertyValueForId(id, propertyDefinition, valueMap))
+            // Filter both by key and by (translated) label (that's why we construct a PropertyValue first)
+            .filter(val -> val.getValue().toString().toLowerCase().contains(lowerFilter)
```
surli, line 63: You could have used `StringUtils#containsIgnoreCase` to simplify it.

surli (approved):

surli:
> What do you think, is this kind of test okay or should I change it to, e.g., mock the results of the ListClass-methods?

So in general better use mock everywhere you can to only test small things. But do ensure that the expectations of what you mock are also tested. So in your case it would mean mocking ListClass and also ensure that the ListClass behaviour you expect are tested. 
Now there's lots of exceptions to this general rule, like if you can't mock some methods because they are static, and we have lots of our code where it's tested like you did.

> What do you think about making one of the variants of ListClass::getDisplayValue public?

+1 to make it public

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/StaticListClassPropertyValuesProvider.java
michitux: Good catch, I've added a comment.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/classes/StaticListClassPropertyValuesProvider.java
michitux: Didn't know about that (and it hasn't been used before), I've changed the code to use it.

michitux:
Thank you for the review @surli!

I've rebased on master and addressed the comments (in new commits, the existing commits are just rebased). I've also noticed that I missed a `revapi.differences` entry due to the new `@Unstable` annotation, hadn't checked again after adding the annotation. Looking at the code again, I also refactored it a bit to re-use the existing property value from the parent method to avoid that it overrides possibly already existing metadata (no idea if that's possible, in the REST API it's definitely just the query string from the request that is passed as `Object`).

xwiki-platform-core/pom.xml
```diff
@@ -321,6 +321,19 @@
...
+            <revapi.differences>
+              <justification>Not a breakage, new annotation on existing protected and now public method.</justification>
+              <criticality>allowed</criticality>
+              <ignore>true</ignore>
+              <differences>
+                <item>
+                  <code>java.annotation.added</code>
+                  <old>method java.lang.String com.xpn.xwiki.objects.classes.ListClass::getDisplayValue(java.lang.String, java.lang.String, java.util.Map&lt;java.lang.String, com.xpn.xwiki.objects.classes.ListItem&gt;, com.xpn.xwiki.XWikiContext) @ com.xpn.xwiki.objects.classes.StaticListClass</old>
+                  <new>method java.lang.String com.xpn.xwiki.objects.classes.ListClass::getDisplayValue(java.lang.String, java.lang.String, java.util.Map&lt;java.lang.String, com.xpn.xwiki.objects.classes.ListItem&gt;, com.xpn.xwiki.XWikiContext) @ com.xpn.xwiki.objects.classes.StaticListClass</new>
+                  <annotation>@org.xwiki.stability.Unstable</annotation>
```
surli, line 333: Weird that you only get a revapi problem about the annotation and not about the visibility that has changed. Did you try to build `xwiki-platform-oldcore-legacy` with your changes? It's possible that revapi is only executed there to also include legacy code in the API comparison.

tmortagne: > It's possible that revapi is only executed there

Yes, it's only executed in legacy module.

michitux: It is indeed breaking because I missed all the other child classes of `ListClass` as I didn't try building the legacy module. However, making the method public is not breakage. As discussed in the [chat](https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$1637144122457QMuoU:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=nobis.cloud), I've opened https://github.com/xwiki/xwiki-commons/pull/151 to ignore the `@Unstable`-annotations in general. This makes all of these annotations here obsolete. Should I remove them here or should I still add all the necessary annotations for now such that this can be merged?

surli: > Should I remove them here

you should yes.

PR #1709, opened 2021-10-25
XWIKI-18752: Replace the livetable from the restore deleted page form with a Live Data

michitux:
* Change `restore.vm` to use wiki syntax and add the live data macro
* Add an integration test for the undelete action
* Introduce a new page object class `RefactoringStatusPage` class that takes parts of `CopyOrRenameOrDeleteStatusPage`
* Introduce a new page object class `RestoreStatusPage` inheriting from `RefactoringStatusPage`
* Introduce a new page object class `UndeletePage`

JIRA issue: https://jira.xwiki.org/browse/XWIKI-18752

Before:
XWIKI-18752-Before

After:
XWIKI-18752-After


michitux:
This needs some more work from my side, I noticed that I have both waits in the integration test that are not necessary and the page object (most likely) causes long waits due to the use of a list.

PR #1710, opened 2021-11-01

oanalavinia:
* minor changes (comments and annotations updates)
* fixed bug: update the annotations offsets based on the differences between the server and client versions of the document (see comment)


xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -1120,6 +1120,10 @@ XWiki.Annotation = Class.create({
     var parentNodePlainTextLength = 0;
     for (let i = 0; i &lt; childNodes.length; i++) {
       child = childNodes[i];
+      // Skip content which is only needed for editing sections.
+      if (child.className &amp;&amp; child.className.indexOf("edit_section") != -1) {
```
mflorea, line 1124: This shows we have a big problem that we haven't anticipated.. **any** JavaScript code that modifies the document content can potentially break the annotation placement. This happens because:

* on the server side the annotation position is computed based on the plain text after executing the rendering transformations, but **without** the JavaScript execution
* on the client side the annotation position is computed based on the plain text after executing the JavaScript code

If the plain text used on the server is different than the plain text used on the client, the annotations are misplaced. This happens for instance when JavaScript code adds text content to the document content. For plain (content) wiki pages, the JavaScript modifications are very limited (edit section links), but for structured pages they may be consistent (coming from widgets used to display the object properties). For instance ``editableProperty.js`` modifies the page content in view mode to add support for editing the object properties in-place.

We need a **generic** solution for this problem. Fortunately I already implemented something similar for the CKEditor to preserve the user (text) selection between WYSIWYG and Source view. The principle is this:

* You have some content and a segment (range) of it is specified with a start and end offset. In the case of CKEditor this segment (range) is the user text selection, while in your case is the annotated text
* The content changes. In the case of CKEditor the HTML is replaced with wiki syntax (or the other way around). In your case it's the JavaScript code that modifies the content.
* You want to update the segment (range) so that it takes into account the content changes. This means "moving" the start and end offsets up or down depending on the changes (e.g. if content was inserted or removed before the offset).

For this to work you need 3 things:

1. the content before the change (you have it in the annotation REST response)
2. the segment before the change (you have it in the annotation REST response)
3. the content after the change (you take it from the DOM right before marking the annotations)

The algorithm is quite simple:

1. you compute the diff between the two versions of the content
2. you update the segment by increasing or decreasing the start / end offset based on the list of changes (e.g. if 3 characters were added before the start offset then the start offset needs to be increased by 3).

You can check this in https://github.com/xwiki-contrib/application-ckeditor/blob/master/application-ckeditor-plugins/src/main/resources/xwiki-selection/plugin.js#L177 .

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -1860,42 +1861,142 @@ XWiki.Annotation = Class.create({
...
+        var plainTextEndOffset = ann.fields.find(field =&gt; field.name == 'plainTextEndOffset');
+        // Check if the annotation was found.
+        if (!plainTextStartOffset.value || !plainTextEndOffset.value) {
+          return;
+        }
+        plainTextStartOffset.value = offsetUpdater.findOffsetAfterChanges(changes, parseInt(plainTextStartOffset.value));
+        plainTextEndOffset.value = offsetUpdater.findOffsetAfterChanges(changes, parseInt(plainTextEndOffset.value));
+      }
+    });
+    // parse the exception spaces and check where is the current space in that list
```
mflorea, line 1987:
```suggestion

    // parse the exception spaces and check where is the current space in that list
```

mflorea (changes requested):

mflorea (approved):

PR #1711, opened 2021-11-05
XWIKI-19115: Improve extension capability of inplace editing

surli:
* Ensure to expose fakeform
  * Define in configuration which selector is used for the inplace
    editing button
  * Send the JSON answer obtained from the request when sending the
    saved event

### DIFF
--- a/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
+++ b/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
@@ -737,7 +737,7 @@ define('editInPlace', [
         '&lt;div class="inplace-editing-buttons sticky-buttons"/&gt;&lt;/div&gt;').insertAfter(editContent).toggle(!!xwikiDocument);
       var actionButtons = actionButtonsWrapper.children('.sticky-buttons')
         .data('xwikiDocument', xwikiDocument)
-        // expose the fakeform if an extension needs to manipulate it
+        // Expose the fake form if an extension needs to manipulate it.
         .data('fakeForm', fakeForm);
       return loadActionButtons(actionButtons);
     } else {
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
```diff
@@ -768,6 +774,7 @@ define('editInPlace', [
 
   var initTranslateButton = function(xwikiDocument) {
     // Initialize the translate button only if it's visible.
+    // FIXME: this should probably use the translateButtonSelector
     const translateButton = $('#tmTranslate [data-toggle="popover"]').filter(':visible');
```
mflorea, line 778:
```suggestion
    const translateButton = $(config.translateButtonSelector).filter('[data-toggle="popover"]').filter(':visible');
```

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
surli: Note that the current selector is not `#tmTranslate` but `#tmTranslate > a`, that's why I hesitated modifying the selector there. You're sure it would work properly with this?

xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
```diff
@@ -862,6 +869,9 @@ define('editInPlace', [
     insert: function(element) {
       this._getActionButtons().find('.hidden.extra').append(element);
     },
+    append: function(element) {
```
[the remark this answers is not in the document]
surli: To be honest I added that one because I'm using `$('#form').append(node)` and when transforming the code to use either `$('#form')` or `fakeform` I noticed that `insert` does not exist in jquery so felt easier to just create this append method there. So if that's ok to you I might just add a comment to specify that only the first element appending is supported there.

xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml
mflorea: It should work fine because ``$('#tmTranslate > a')`` finds the anchors (it should be just one) and then ``.filter('[data-toggle="popover"]').filter(':visible')`` filters this list of anchors (with just one item) based on whether the anchors have the ``data-toggle`` attribute and are visible. Basically it keeps in the list only the ``'#tmTranslate > a'`` anchors that are visible and have the ``data-toggle="popover"`` attribute, which is what we want.

surli: Thanks for the confirmation

PR #1713, opened 2021-11-10
XWIKI-19125: Introduce a displayedAuthor in XWikiDocument

surli:
## JIRA

https://jira.xwiki.org/browse/XWIKI-19125

## Description

The main goal of this PR is to introduce a `displayedAuthor` in XWikiDocument along with the the `author` and `contentAuthor` that already exists. For the motivation, about introducing this new author, see: https://forum.xwiki.org/t/introduce-a-xwikidocument-displayedauthor/9365/

In order to achieve this goal this PR provides the following:
  - a new DocumentAuthors API wrapping all document authors (metadata author, content author, displayed author and creator) and a DefaultDocumentAuthors internal which contains the setters. All those APIs are using UserReference.
  - a new authors field in XWikiDocument containing a DocumentAuthors instances: all the existing APIs for authors in XWikiDocument are deprecated in favor of get/set Authors
  - a new mapping in hibernate file to store the displayedAuthor reference in XWikiDoc table
  - using of displayedAuthor in XWikiRCSArchive and XWikiDocumentArchive for setting the author
  - using of new API in SaveAction to set the authors
  - using of displayedAuthor in Document#save() 
  - using of displayedAuthor in the headercontent file 
  - minor performance improvment in DocumentDocumentReferenceUserReferenceResolver to use reference dedup 

## Limitations

The following was **not** done:
  - replacement of all authors usage to rely on new API: it would make the PR quickly impossible to read if I start changing all API call right away, so I prefer we agree on that work first. Also note that it should not be a problem for the CI since the XWikiDocument API is never directly accessed in scripts, so we shouldn't get any deprecated warnings. Finally, this work would be a lot easier if we had a UserReference in the context, which should be done first IMO. 
  - moving towards using references for all authors in DB: I only provided a new mapping for the new displayedAuthor, without changing existing mapping for the current authors fields, even though they are deprecated. The reason is that I don't need to provide a migration right now (because displayedAuthor is optional), which would be mandatory to move all existing mappings, so it's simpler to keep those like that for now.


surli:
@tmortagne @vmassol I'd like your opinion on this first commit: this is a WIP so lots of things are still missing to complete the work. As mentioned in my commit message here the idea is to have a dedicated concept to hold the different authors of XWikiDocument to ease maintenance in the future (if we ever want to refactor again those authors), and also to move towards using UserReference for the authors of document. WDYT?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/document/DocumentAuthors.java
```diff
@@ -0,0 +1,60 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.model.document;
```
surli, line 20: Note that I put this new concept in a new package which IMO was missing in model module.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/document/DocumentAuthors.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * The displayed author is the "real" author of a document as "the one who impulses the changes". However this
+     * author is only there for display purpose: it should never be used for security reasons. This author is different
+     * from metadata or content author when the document is saved through a script or a mechanism which delegates the
+     * save of the document.
+     *
+     * @return the reference of the user who impulses the creation of the document, without being the real user used
+     *          in the APIs for saving the changes.
+     */
+    UserReference getDisplayedAuthor();
```
surli, line 59: I hesitated between returning a UserReference or a String here, for handling cases where we want to output a name for the Guest user, but I'm now wondering if it shouldn't be the role of the GuestUserReference to encapsulate that information to have cleaner APIs. WDYT @vmassol?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/document/DocumentDocumentReferenceUserReferenceResolver.java
```diff
@@ -44,18 +45,23 @@
...
+
     @Override
     public UserReference resolve(DocumentReference rawReference, Object... parameters)
     {
         UserReference reference;
         if (rawReference == null) {
             reference = CurrentUserReference.INSTANCE;
         } else {
-            reference = resolveName(rawReference.getName());
+            // small perf improvment to avoid keep duplicated references in memory.
```
surli, line 58: This change is not needed for this work, it's just that I noticed the dedup of references we were doing in XWikiDocument#setAuthorReference and I thought it should be good to dedup them also in this resolver.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/document/DocumentAuthors.java
xwikiorgci: AFAIU `getDisplayedAuthor()` is not about returning the display of the author but something else, and thus we should return a reference. And yes it's the role of GuestUserReference to provide the name (and it does already - through `UserPropertiesResolver`).

Now, I can't say what that something else is since `the user who impulses the creation of the document, without being the real user used in the APIs for saving the changes` is not understandable to me... What does it mean to "impulse the creation of a document". If my boss tells me to create a document in xwiki, does that mean that he's the "display author" (whatever this means)? I think we need a better javadoc and as always, always use examples in the javadoc to make things more clear. BTW from your javadoc, it shouldn't be `getDisplayedAuthor()` but `getImpulseAuthor()`... ;) For me this shows there's a problem in the api name. The fact that it seems hard to document suggests to me there's a problem with this concept ATM and that it might be wrong.

xwikiorgci: >  And yes it's the role of GuestUserReference to provide the name (and it does already - through UserPropertiesResolver).

See https://github.com/xwiki/xwiki-platform/blob/c1bb91f980327e9d7ff05bbc3a7ece78306b1c05/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/GuestConfigurationSource.java#L58

tmortagne:
Even if it's not technically breaking I'm wondering if all that should not be introduced in 14.0 instead of 13.10 since it's a big change conceptually, and it's going to produce an impressive number of deprecation warnings. But maybe 14.0 is a bit late for you @surli ?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Specify the author of the content of the document: this author is only responsible to the content, and not to
+     * other information of the document such as xobjects.
+     *
+     * @param contentAuthor the author of the content of the document.
+     * @return the current instance for Builder pattern.
+     */
+    public DefaultDocumentAuthors setContentAuthor(UserReference contentAuthor)
```
tmortagne, line 65: I'm wondering if `contentAuthor(UserReference contentAuthor)` would not make more sense for what seems to be a fluent API. In my experience, the return of a "set*" when it's not void is generally more the previous value. 

Now, I'm not sure we really need a fluent API here.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
```diff
@@ -0,0 +1,144 @@
...
+            .append(contentAuthor, that.contentAuthor)
+            .append(metadataAuthor, that.metadataAuthor)
+            .append(displayedAuthor, that.displayedAuthor)
+            .isEquals();
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return new HashCodeBuilder(17, 67)
```
tmortagne, line 138: What is the passing custom values to this constructor ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1863,28 +1884,23 @@ private String userReferenceToString(DocumentReference userReference)
...
-
-        this.authorReference = intern(authorReference);
-
-        // Log this since it's probably a mistake so that we find who is doing bad things
-        if (this.authorReference != null && this.authorReference.getName().equals(XWikiRightService.GUEST_USER)) {
-            LOGGER.warn("A reference to XWikiGuest user has been set instead of null. This is probably a mistake.",
-                new Exception("See stack trace"));
-        }
+        UserReference user = this.getUserReferenceDocumentReferenceResolver().resolve(authorReference);
+        this.setAuthors(new DefaultDocumentAuthors(this.getAuthors()).setMetadataAuthor(user));
```
tmortagne, line 1903: Why creating a new DefaultDocumentAuthors for each set instead of just modifying the existing one ?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/document/DocumentAuthors.java
surli: > And yes it's the role of GuestUserReference to provide the name (and it does already - through UserPropertiesResolver).

So the snippet you show does not really answer my question: it always will display "Guest". My question was that, for example if I comment as Guest I have the capability to specify a username that will be stored in the comment and used for the display. IMO it would make sense that somehow the GuestUserReference could contain such information for various usecases when we allow Guest to perform actions. I'm still wondering if it would make sense and could be encapsulated in the serialization of the reference itself, or if it should be something specific for each usecase. 
I'm asking this question since it could make sense that a for a document created by Guest, we actually display the name provided by the user in some field. 

> `the user who impulses the creation of the document, without being the real user used in the APIs for saving the changes` is not understandable to me... 

So to clarify the idea here is that a document can be created or saved by a script that has the rights to do so when the user itself does not necessarily have the rights. So the final user who uses the script is the user "who impulses" the creation or save. Easy example is a the usage of change request: the user who performs the changes is not the one who merges, the first one will be this displayed user. 
I agree that the Javadoc can be improved with this and other examples. Now for discussing the concept itself of adding this new API I suggest to do it in my original proposal on the forum: https://forum.xwiki.org/t/introduce-a-xwikidocument-displayedauthor/9365

surli:
> I'm wondering if all that should not be introduced in 14.0 instead of 13.10 since it's a big change conceptually, and it's going to produce an impressive number of deprecation warnings. But maybe 14.0 is a bit late for you @surli ?

I'm also hesitating about that. It mainly means that ChangeRequest won't be able for sure to depend on 13.x LTS. Not necessarily a big deal, I'm not even sure yet that I won't need new things to be added in 14.x for it.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
surli: > I'm wondering if contentAuthor(UserReference contentAuthor) would not make more sense for what seems to be a fluent API. In my experience, the return of a "set*" when it's not void is generally more the previous value.

I'm personally not a big fan of APIs not starting with `set` for setters, even in fluent APIs. Just because when using auto-completion it's easier to spot what you need to set like that. But probably a matter of taste.

> Now, I'm not sure we really need a fluent API here.

The idea here is to always be able to call a `new DefaultDocumentAuthors().setXXX()` when you need to call `XWikiDocument#setAuthors` to simplify writing it.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
surli: The javadoc suggest to pick randomly chosen hardcoded value for those numbers, if possible different for each classes, so that's what I did.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
surli: Because XWikiDocument contains the interface DocumentAuthor, not the DefaultDocumentAuthor: in the future we can imagine that there's others implementations, and it would still work.

tmortagne: Except that as soon as a set is called you loose the previous custom one and replace it with DefaultDocumentAuthor so you might set DefaultDocumentAuthor from the start and just call set on it.

tmortagne: By the way not exposing any set in DocumentAuthors means that changing one of those authors is quite a pain and close to impossible for an extension without using deprecated APIs since DefaultDocumentAuthors is internal.

surli: Right.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
tmortagne: That would suggest that two object of incompatible type would be used as key in the same Map, quite strange.

vmassol:
I haven't checked the PR itself but the title still mentions "displayedAuthor" and for me it's not about display but about knowing the author that triggered the change. See https://forum.xwiki.org/t/introduce-a-xwikidocument-displayedauthor/9365/8?u=vmassol

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentheader.vm
```diff
@@ -36,7 +36,8 @@
             $xwiki.formatDate($tdoc.date)
           ])
         #else
-          $services.localization.render('core.footer.modification', [$xwiki.getUserName($tdoc.author), $xwiki.formatDate($tdoc.date)])
+          #set ($authorName = $services.user.serialize($tdoc.authors.displayedAuthor))
```
tmortagne, line 39: Wouldn't passing directly `$tdoc.authors.displayedAuthor` to getUserName have the same effect ? I did not check but I think we have a Converter for AuthorReference.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
```diff
@@ -3247,4 +3258,33 @@ public int getLocalReferenceMaxLength()
...
+    /**
+     * Define the author of the document to be displayed, without changing the metadata and content authors.
+     * @param displayedAuthor the author to be displayed.
+     * @since 14.0RC1
+     */
+    @Unstable
+    public void setDisplayedAuthor(UserReference displayedAuthor)
+    {
+        DefaultDocumentAuthors documentAuthors;
+        if (this.doc.getAuthors() != null) {
```
tmortagne, line 3271: I'm really not a fan of how complex it is to set authors in a XWikiDocument.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
```diff
@@ -2543,6 +2550,10 @@ public void save(String comment) throws XWikiException
     public void save(String comment, boolean minorEdit) throws XWikiException
```
tmortagne, line 2550: That's only one entry point to save the document and from what I understand it's not set at all in others, which I don't think is what we want.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
```diff
@@ -220,12 +224,20 @@ public boolean save(XWikiContext context) throws XWikiException
             tdoc.readFromForm(form, context);
         }
 
-        // TODO: handle Author
-        String username = context.getUser();
-        tdoc.setAuthor(username);
+        // TODO: entirely rely on UserReference from the context
```
tmortagne, line 227: Going to be fun finding the right get* name with getUserReference already taken :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentheader.vm
surli: > I did not check but I think we have a Converter for AuthorReference.

I guess you meant `UserReference`. We might have, not sure, but in general I prefer to do the conversion explicitely when it's possible. Do you feel it makes the code harder to read?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
surli: Actually this check should not be needed since I also check it in the setter. Will remove it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
surli: I might miss something there, but from what I recall, you should have the rights to set the authors when using `saveAsAuthor`, which you don't have when you use this API.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/SaveAction.java
surli: To be honest I started looking to perform the refactoring, and I quickly stopped given the nightmare it was between naming stuff,  use UserReference everywhere and initialize it properly where it needs...

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentheader.vm
tmortagne: Yes sorry `UserReference` and yes the point was just to suggest a simpler code.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
tmortagne: My point is that from what I understand if you use #saveAsAuthor the display author is simply not set at all which mean any code which rely on it will have a wrong value.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
tmortagne: My comment is actually more for the fact that you have to create a new DefaultDocumentAuthors (which is an internal class which does not make this whole thing complex but simply impossible for an extension as I already indicated in a previous comment), duplicate the current one except the one value you want to change and set it.

Honestly, I don't see why DocumentAuthors does not simply have setters.

surli: Fixed by adding setters in DocumentAuthors.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/Document.java
surli: Fixed in saveAsAuthor.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
```diff
@@ -0,0 +1,145 @@
...
+            this.effectiveMetadataAuthor = documentAuthors.getEffectiveMetadataAuthor();
+            this.originalMetadataAuthor = documentAuthors.getOriginalMetadataAuthor();
+            this.creator = documentAuthors.getCreator();
+        }
+    }
+
+    @Override
+    public void setContentAuthor(UserReference contentAuthor)
+    {
+        this.contentAuthor = contentAuthor;
```
tmortagne, line 65: One thing which is missing from all the set* methods is to update the XWikiDocument metadataDirty flag.

surli: > One thing which is missing from all the set* methods is to update the XWikiDocument metadataDirty flag.

Yeah I noticed it, and was actually wondering about that. A way for dealing with that would be to introduce a dirty flag in `DefaultDocumentAuthors` and checking for its state when getting the flag value in `XWikiDocument` but wasn't sure if it was needed, or if we could consider that it would be the API user job to also set the flag when using the `DocumentAuthors` setters (in which case I should update the javadoc to indicate it). wdyt?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -635,6 +641,11 @@ private static LinkParser getLinkParser() {
 
     private RenderingContext renderingContext;
 
+    /**
+     * @see #getAuthors()
+     */
+    private DefaultDocumentAuthors authors;
```
tmortagne, line 647: Now that DocumentAuthors is modifiable, it would IMO make more sense to set this field from the start and then always modify it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -9126,4 +9169,71 @@ public int getLocalReferenceMaxLength()
...
+    /**
+     * Allow to set the different authors of the document.
+     * Note that if one of the author changed as part of this call, the {@link #setMetaDataDirty(boolean)} flag is set
+     * to {@code true}.
+     *
+     * @param authors the authors information.
+     * @since 14.0RC1
+     */
+    @Unstable
+    public void setAuthors(DocumentAuthors authors)
```
tmortagne, line 9194: I don't think setAuthors is really needed now that DocumentAuthors has setters.

surli: The immediate UC I can think of for this setter, is if someone needs for some reason another implem of DocumentAuthors, then it's handy to be able to set the instance.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/document/DefaultDocumentAuthors.java
tmortagne: > we could consider that it would be the API user job to also set the flag when using the DocumentAuthors setters (in which case I should update the javadoc to indicate it).

-1, nobody will remember that

What we do usually is embedding the XWikiDocument instance in the modificable object and call set dirty flag through it. See XWikiAttachment#setMetaDataDirty for example.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
surli: Done.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
surli: I kept it, but it only copies the value of the given parameter.

tmortagne: What about a `DocumentAuthors#set(DocumentAuthors)` instead ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -1914,29 +1920,28 @@ public void setAuthor(String author)
 
     /**
      * @since 3.0M3
+     * @deprecated Since 14.0RC1 rely on {@link #getAuthors()}.
      */
     @Override
     public DocumentReference getContentAuthorReference()
     {
-        return this.contentAuthorReference;
+        if (getAuthors() != null && this.getAuthors().getContentAuthor() != null) {
```
tmortagne, line 1928: getAuthors() cannot return null anymore

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -9126,4 +9130,63 @@ public int getLocalReferenceMaxLength()
     {
         return getStore().getLimitSize(this.getXWikiContext(), this.getClass(), "fullName");
     }
+
+    /**
+     * {@inheritDoc}
```
tmortagne, line 9135: Not needed.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentheader.vm
surli: > Wouldn't passing directly $tdoc.authors.displayedAuthor to getUserName have the same effect ? I did not check but I think we have a Converter for AuthorReference.

So for info it does not because `XWiki#getUserName` takes a String as parameter, and the toString of the UserReference is not calling  the UserReference serializer, so we don't produce the proper output. So basically I do need to call the serializer before calling `XWiki#getUserName`

tmortagne: > toString of the UserReference is not calling the UserReference serializer

That's not what is called during a conversion, it should go through UserReferenceConverter#convertToString.

PR #1715, opened 2021-11-16
like_de: add two missing translations

SimpelMe:
- like.button.title.unlike
 - like.configuration.heading

Signed-off-by: Conrad Zelck <git@simpel.cc>


surli:
@SimpelMe thanks for the contribution. For information you can perform translations using the dedicated website on https://l10n.xwiki.org it helps ensuring some checks are performed on the translation.

PR #1718, opened 2021-11-19

oanalavinia:
* fix AnnotationsIT: the js condition to check if an annotation exists was failing when an anotation started at the beginning of the text since `!plainTextStartOffset.value` returns true for `0`

### DIFF
--- a/xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
+++ b/xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
@@ -985,7 +985,8 @@ XWiki.Annotation = Class.create({
     // Check if the annotation was found.
     var plainTextStartOffset = ann.fields.find(field =&gt; field.name == 'plainTextStartOffset');
     var plainTextEndOffset = ann.fields.find(field =&gt; field.name == 'plainTextEndOffset');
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -985,7 +985,8 @@ XWiki.Annotation = Class.create({
     // Check if the annotation was found.
     var plainTextStartOffset = ann.fields.find(field =&gt; field.name == 'plainTextStartOffset');
     var plainTextEndOffset = ann.fields.find(field =&gt; field.name == 'plainTextEndOffset');
-    if (!plainTextStartOffset.value || !plainTextEndOffset.value) {
+    if ((!plainTextStartOffset.value &amp;&amp; plainTextStartOffset.value !== 0) ||
+        (!plainTextEndOffset.value &amp;&amp; plainTextEndOffset.value !== 0)) {
```
surli, line 989: Isn't `plainTextStartOffset.value === "undefined"` when the annotation is not found? Just wondering if the condition couldn't be simplified. At the very least I would create a function for it if we need to use it at different places.

oanalavinia: It would work with `plainTextStartOffset.value === null`, since it is not undefined

PR #1721, opened 2021-12-01
XWIKI-19192: Support creating objects using in-place editable properties

michitux:
* Pass objectPolicy from the data property in all requests of editable properties
* Introduce a new parameter for the `displayObjectProperty` macro to create the object and property if missing
* React on the objectPolicy=updateOrCreate value in display.vm to set the parameter of `displayObjectProperty`
* Add a new page test that tests this behavior of `display.vm`

Jira Issue: https://jira.xwiki.org/browse/XWIKI-19192


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/display_macros.vm
```diff
@@ -17,9 +17,16 @@
 ## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 ## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 ## ---------------------------------------------------------------------------
-#macro (displayObjectProperty $propertyReference $displayMode)
+#macro (displayObjectProperty $propertyReference $displayMode $createIfMissing)
```
surli, line 20: The code looks good, but I'm not sure about the design.
My main concern is that `displayObjectProperty` is now making a bit more stuff than just displaying values, for a very specific UC. So sounds like we almost need a different macro for it (which is almost already the case since you added a new parameter), which would be called only for this UC. Now I don't know much if it would be doable to detect in display.vm if we are in the specific conditions of this UC to call the other macro.

michitux: This is definitely possible, instead of setting the boolean parameter there could be an `#if` with a call to a different macro. Is your suggestion that this extra macro should just create the object and the property if missing (like `#createObjectAndPropertyIfMissing`) or should it replace the existing macro, i.e., basically be a variant of `#displayObjectProperty` like `#displayPotentiallyMissingObjectProperty`?

surli: > Is your suggestion that this extra macro should just create the object and the property if missing (like #createObjectAndPropertyIfMissing)

that sounds better to me yes.

mflorea:
@michitux would be good to give an example of how you plan to use the changes you propose in this PR, because I think you're overlooking something. Currently the usage is something like this:

```
<dt class="editableProperty" data-property="Some.UsefulClass_5_status" data-property-type="object">
  <label for="Some.UsefulClass_5_status">Contributor Type</label>
</dt>
<dd>$doc.display('status', $fifthObject)</dd>
```

I'm particularly interested in what property reference are you planing to use.

michitux:
The intended use is the following:

```html
{{template name="display_macros.vm" /}}

{{velocity}}
{{html clean="false"}}
      <dt#if ($hasEdit) class="editableProperty"#end data-property="NumberedHeadings.Code.NumberedHeadingsClass.status"
      data-property-type="object" data-object-policy="updateOrCreate">
        <label>$escapetool.xml($services.localization.render('numbered.headings.activationUI.label'))</label>
      </dt>
      <dd>#displayObjectProperty('NumberedHeadings.Code.NumberedHeadingsClass.status' 'view' true)</dd>
{{/html}}
{{/velocity}}
```

in the context of the information tab below the content, see also #1725 and the [NCAPP-9](https://github.com/xwiki-contrib/application-numbered-content/compare/NCAPP-9) branch. Basically, the idea is that this allows you to control the heading numbering status which is stored in a static list property in the numbered headings class. Not every page will have such an object in which case the default status shall be displayed and when the user is saving the status, the object shall be created.

mflorea:
@michitux your example works because ``NumberedHeadings.Code.NumberedHeadingsClass.status`` is treated as the ``status`` of the **first** ``NumberedHeadingsClass`` object. It doesn't work if the page already has a ``NumberedHeadingsClass`` object and you want to add a second one using in-place editing. Someone reading the issue description may think that they can write:

```
<dt class="editableProperty" data-property="Some.UsefulClass[5].status" data-property-type="object" data-object-policy="updateOrCreate">
  <label>Contributor Type</label>
</dt>
<dd>#displayObjectProperty('Some.UsefulClass[5].status' 'view' true)</dd>
```

but this won't work. Moreover, with your changes, it will try to create an object of class ``Some.UsefulClass[5]`` which may lead to strange results. So you need to mention clearly, in comments and documentation, that ``data-object-policy="updateOrCreate"`` works only with a very specific case: creating the **first** object. So only when the object number is not specified in the property reference (``data-property``). This is different than using ``objectPolicy`` in a form, such as:

```
<input type="text" name="Some.UsefulClass_5_status" value="test"/>
<input type="hidden" name="objectPolicy" value="updateOrCreate"/>
```

where "it will create it (and pad will null objects if needed)", as per [documentation](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/Architecture/URL%20Architecture/Standard%20URL%20Format/#HParameter%3AobjectPolicy).

michitux:
To be honest, it wasn't clear to me how to specify the object number in the property reference format as the examples I looked at didn't include it. I wouldn't mind updating the code to support that use case with the object number if it is easily possible.

mflorea:
The public ``ObjectReference`` has only an _object name_, that can be anything. The ``BaseObjectReference`` extends ``ObjectReference`` and uses a class reference and an object number to define the object name:

```
<classReference>[<objectNumber>]
Some.UsefulClass[5]
```

> I wouldn't mind updating the code to support that use case with the object number if it is easily possible.

The problem is that we're missing a public API to create an object with a given number. I mean, if you put

```
<dt class="editableProperty" data-property="Some.UsefulClass[5].status" data-property-type="object" data-object-policy="updateOrCreate">
```

then we need an API to create the fifth object of type ``Some.UsefulClass`` (even when there isn't a forth object of type ``Some.UsefulClass``). Now, I'm not sure this is very useful, but the problem is that without this the usage of object policy in in-place creation is not consistent with the usage of object policy in a standard form.

michitux:
There is `com.xpn.xwiki.doc.XWikiDocument#getXObject(org.xwiki.model.reference.EntityReference, int, boolean, com.xpn.xwiki.XWikiContext)` which creates an object with the given number if it doesn't exist. I do not see why we shouldn't expose this to the `Document` API. There is already `com.xpn.xwiki.api.Document#getObject(java.lang.String classname, boolean create)` and `com.xpn.xwiki.api.Document#getObject(java.lang.String classname, int nb)`, so I do not see what speaks against adding `com.xpn.xwiki.api.Document#getObject(java.lang.String classname, int nb, boolean create)`. However, looking a bit further, I think I'm now understanding that we would need `com.xpn.xwiki.doc.XWikiDocument#getXObject(org.xwiki.model.reference.ObjectReference, boolean create, com.xpn.xwiki.XWikiContext)` (and a corresponding API method) instead, as we cannot get a class name and object number from the object reference in the script, is that correct?

I would be much happier if we could implement this without modifying the document, but I'm not sure if this is possible or desirable. The object editor also just creates objects on the document and also simply sets their number [using setNumber()](https://github.com/xwiki/xwiki-platform/blob/ab9ddd1ff8914f818e9b1181dd64fae601628cad/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editobject.vm#L268).

tmortagne:
> I do not see why we shouldn't expose this to the Document API

Most of the APIs exposed in Document have been when needed and there is very often no real thinking on whether or no something that just been introduced in XWikiDocument should be put in Document too.

michitux:
I have extended `XWikiDocument` and the API `Document` to support the use case of a provided object number. Further, I have moved the creation of the object and the property into a separate macro. I am still not really happy with modifying the document but I cannot see any other way to implement this currently. Now it should at least be clear for the user of that macro that it is modifying the document.

michitux:
I've just added another test case to the page test to test with an explicitly specified object number. Further, I have rebased on master and force-pushed to resolve the merge conflict.

mflorea (approved):
Looks good to me.

PR #1724, opened 2021-12-03
XWIKI-19203: The xobjectNumbers from the addedObjects parameter are ignored

michitux:
* Create objects with the correct number
* Add a test case to test adding, updating and removing objects

Jira issue: https://jira.xwiki.org/browse/XWIKI-19203

Note: this depends on #1722 as `getXObject` has been buggy before.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -4099,12 +4099,8 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
             DocumentReference classReference = resolveClassReference(className);
             List<Integer> classIds = objectsToAdd.get(className);
             for (Integer classId : classIds) {
-                BaseObject xObject = getXObject(classReference, classId);
-
                 // we ensure that the object has not been added yet, for example because of the update or create.
-                if (xObject == null) {
-                    newXObject(classReference, context);
-                }
+                getXObject(classReference, classId, true, context);
             }
         }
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4104,12 +4104,8 @@ public void readFromForm(EditForm eform, XWikiContext context) throws XWikiExcep
             DocumentReference classReference = resolveClassReference(className);
             List<Integer> classIds = objectsToAdd.get(className);
             for (Integer classId : classIds) {
-                BaseObject xObject = getXObject(classReference, classId);
-
                 // we ensure that the object has not been added yet, for example because of the update or create.
-                if (xObject == null) {
-                    newXObject(classReference, context);
-                }
+                getXObject(classReference, classId, true, context);
```
tmortagne, line 4108: This looks more like code style than something that has an actual impact on the behavior.

Seems to me, XWIKI-19203 is actually a duplicate of XWIKI-19201, isn't it ?

michitux: > This looks more like code style than something that has an actual impact on the behavior.

No, the previously used code didn't use the `classId` for creating an object at the correct position, instead it created an object with the next available number.

> Seems to me, XWIKI-19203 is actually a duplicate of XWIKI-19201, isn't it ?

No, even though it is similar, it is actually the reverse: while in XWIKI-19201 the problem was that objects were only created if there was no object of the given class at all but objects were created with the given number, here the problem is that while the existence check is okay, the object is not created with the given number.

Note: I've just simplified the test case and rebased my branch on the master branch.

PR #1725, opened 2021-12-03
XWIKI-19171: Add UIXPs for page information in view and edit mode

michitux:
* Add a new UIXP `org.xwiki.platform.panels.documentInformation` in the document information panel
* Move the existing content into UIXs objects, currently all in the same document as the panel

At the moment, this is the first part of XWIKI-19171, as it only adds a UIXP for edit mode. I would still like to get feedback on two issues already:

* Is it okay to place all existing UIX into the document of the panel?
* This changes the output from one definition list into one definition list for each item. Do you have any idea how to fix this or is this not a problem?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/DocumentInformation.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/DocumentInformation.xml
@@ -20,7 +20,7 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
 
-<xwikidoc version="1.3" reference="Panels.DocumentInformation" locale="">
+<xwikidoc version="1.4" reference="Panels.DocumentInformation" locale="">
   <web>Panels</web>
   <name>DocumentInformation</name>
   <language/>
@@ -73,11 +73,11 @@
         <number>8</number>
         <prettyName>Context elements</prettyName>
         <relationalStorage>0</relationalStorage>
-        <separator> </separator>
+        <separator>, </separator>
         <separators>|, </separators>
         <size>5</size>
         <unmodifiable>0</unmodifiable>
-        <values>doc.reference=Document|icon.theme=Icon theme|locale=Language|rendering.defaultsyntax=Default syntax|rendering.restricted=Restricted|rendering.targetsyntax=Target syntax|request.base=Request base URL|request.parameters=Request parameters|request.url=Request URL|request.wiki=Request wiki|user=User|wiki=Wiki</values>
+        <values>action=Action|doc.reference=Document|icon.theme=Icon theme|locale=Language|rendering.defaultsyntax=Default syntax|rendering.restricted=Restricted|rendering.targetsyntax=Target syntax|request.base=Request base URL|request.parameters=Request parameters|request.url=Request URL|request.wiki=Request wiki|user=User|wiki=Wiki</values>
         <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
       </async_context>
       <async_enabled>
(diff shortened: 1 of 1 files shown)


surli:
> Is it okay to place all existing UIX into the document of the panel?

I don't think we have a best practice rule for that (or I forgot about it), but IMO the UIX should be splitted in different pages, or at the very least be in a different page than the UIXP one.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/informationinline.vm
```diff
@@ -44,146 +237,23 @@ $xwiki.jsfx.use('uicomponents/edit/editableProperty.js', {'forceSkinAction': tru
...
-                #set ($discard = $output.add(
-                  "<span class='wikicreatelink skipCreatePagePopup'>" +
-                    $stringtool.trim("#docTranslationLink($supportedLocale $action $fragment)") +
-                  "</span>"
-                ))
-              #end
-            #end
-            <dd data-key="translations">$stringtool.join($output, ', ')</dd>
+        #foreach ($extension in $sortedExtensions)
+          #if ($extension.order < 1000)
```
surli, line 241: You're using 1000 as a limit number between static UIX and dynamic ones? Or I misunderstood something? 
Anyway that doesn't sound good: the order number should only be used for sorting the UIX, so a dynamic extension could be added in the middle of the static ones and vice-versa.

surli:
> This changes the output from one definition list into one definition list for each item. Do you have any idea how to fix this or is this not a problem?

I don't think that's a real problem given the fact that it's to avoid having to define uncleaned html.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/informationinline.vm
michitux: I think you misunderstood something. The value 1000 is the limit between the left and the right column. All UIX (static and dynamic) with order less than 1000 are displayed in the left column and all UIX with a number equal or greater than 1000 are displayed in the right column. The UIX in both columns are sorted according to the order parameter. It is the freedom of the skin to divide the items into more or less columns, that's why the UIXP doesn't have an explicit `column` parameter and instead partitions by order. I should add a comment, though I thought this would also be part of the UIXP documentation.

mflorea:
> Both UIXPs expect UIXs to return ``<dt>/<dd>`` elements without wrapper, which is only possible with uncleaned HTML, to avoid that multiple definition lists are introduced.

Using ``{{html clean="false"}}`` is fine IMO (and actually best practice for developers), since these UIXs output HTML **fragments** which maybe not be valid on their own, but valid once inserted in the target UIXP. It's true that this makes the UIXP more fragile and easily breakable by simple users that experiment with UIXP and want to add something to the Document Information panel / tab.

The alternative, to generate the ``dd`` and ``dt`` from the UIPX based on some parameters given by the UIXs is to limiting for the developer so I'm fine with the current approach.

michitux:
> I don't think we have a best practice rule for that (or I forgot about it), but IMO the UIX should be splitted in different pages, or at the very least be in a different page than the UIXP one.

I have split the UIXs into one UIX per document. I hope the naming is okay, please let me know if not.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/informationinline.vm
michitux: I have added a comment to the source to explain the rationale for splitting by `order` parameter.

michitux: I will define a constant for the number. Just for reference, this splitting into columns has also been discussed in the [forum](https://forum.xwiki.org/t/new-user-interface-extension-point-uixp-for-page-information/9470/2) where I first proposed using a column parameter.

PR #1728, opened 2021-12-10
XWIKI-19193: Objects are not created or removed during preview

michitux:
* Extract a new method `XWikiDocument.readAddedUpdatedAndRemovedObjectsFromForm()` from `readFromForm()`
* Change the existing unit test of `readFromForm()` to test this method instead (as it only tested the extracted code, anyways)
* Use this new method during preview/edit actions instead of `XWikiDocument.readObjectsFromForm()`

Jira issue: https://jira.xwiki.org/browse/XWIKI-19193

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -146,6 +146,7 @@ import org.xwiki.rendering.transformation.TransformationManager;
 import org.xwiki.rendering.util.ErrorBlockGenerator;
 import org.xwiki.security.authorization.ContextualAuthorizationManager;
 import org.xwiki.security.authorization.Right;
+import org.xwiki.stability.Unstable;
 import org.xwiki.store.merge.MergeDocumentResult;
 import org.xwiki.store.merge.MergeManager;
 import org.xwiki.velocity.VelocityContextFactory;
@@ -4093,6 +4094,20 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
         readDocMetaFromForm(eform, context);
         readTranslationMetaFromForm(eform, context);
 
+        readAddedUpdatedAndRemovedObjectsFromForm(eform, context);
+    }
+
+    /**
+     * Adds objects, applies property updates and removes objects as specified in the form.
+     *
+     * @param eform The form from which the values shall be read.
+     * @param context The XWiki context.
+     * @throws XWikiException If an error occurs.
+     * @since 14.0RC1
+     */
+    @Unstable
+    public void readAddedUpdatedAndRemovedObjectsFromForm(EditForm eform, XWikiContext context) throws XWikiException
+    {
         // We add the new objects that have been submitted in the form, before filling them with their values.
         Map<String, List<Integer>> objectsToAdd = eform.getObjectsToAdd();
         for (String className : objectsToAdd.keySet()) {
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -4009,6 +4009,13 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
         }
     }
 
+    /**
+     * Updates properties of existing objects with the values from the given form.
+     *
+     * @param eform The form to read the values from
+     * @param context The context used for getting the classes of objects
+     * @throws XWikiException On errors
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4093,6 +4094,20 @@ public void readFromForm(EditForm eform, XWikiContext context) throws XWikiExcep
         readDocMetaFromForm(eform, context);
         readTranslationMetaFromForm(eform, context);
 
+        readAddedUpdatedAndRemovedObjectsFromForm(eform, context);
+    }
+
+    /**
+     * Adds objects, applies property updates and removes objects as specified in the form.
```
surli, line 4101: So I noticed that we don't have a javadoc for `#readObjectsFromForm`: so right now I have a hard time to understand what's the difference between this new method and `readObjectsFromForm`, it could be good to explain this difference at least in this javadoc, to understand when to use one and when to use the others, since it's not clear right now.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/EditAction.java
```diff
@@ -106,7 +106,7 @@ protected XWikiDocument prepareEditedDocument(XWikiContext context) throws XWiki
 
         // The default values from the template can be overwritten by additional request parameters.
         updateDocumentTitleAndContentFromRequest(editedDocument, context);
-        editedDocument.readObjectsFromForm(editForm, context);
+        editedDocument.readAddedUpdatedAndRemovedObjectsFromForm(editForm, context);
```
surli, line 109: Not sure to get it: thought the bug was in PreviewAction?

michitux: `PreviewAction` inherits from `EditAction` and uses this `prepareEditedDocument` method to get the edited document. The same bug also affects the edit action when you, e.g., return from preview to edit.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
michitux: I agree that this is hard to understand. I didn't want to fundamentally change the behavior of `readObjectsFromForm` by including this additional functionality, in particular as the new method calls either `readObjectsFromForm` or `readObjectsFromFormUpdateOrCreate` depending on the `objectPolicy` request parameter. In my opinion, already `readObjectsFromFormUpdateOrCreate` should have been an extension of `readObjectsFromForm`. If you think the code organization should be changed, please let me know.

I cannot tell you when to use one or the other, in my opinion it should always be `readAddedUpdatedAndRemovedObjectsFromForm` as it is the one method that handles all object updates. I could imagine that there might be scenarios where this is not desirable, e.g., if you rely on the fact that objects you specified for deletion are still there during a preview such that you can display the properties of the objects that would be deleted. However, these examples seem so far-fetched and would require such a lot of custom code, e.g., for re-displaying the form for deleting the objects for example that I find it hard to imagine that somebody relies on this behavior.

surli: > I cannot tell you when to use one or the other

Hmmm I've the feeling that `readAddedUpdatedAndRemovedObjectsFromForm` code was created for some specific requests, e.g. for partial document requests when you only provide a part of the information you want to update, and not all of them. It probably evolved to support more usecases. While on the other side, `readObjectsFromForm` was done to work with the all objects infromation being sent. 
Maybe we should start deprecating `readObjectsFromForm` in favor of the other one, if it supports as well all scenario.

michitux: > Hmmm I've the feeling that `readAddedUpdatedAndRemovedObjectsFromForm` code was created for some specific requests, e.g. for partial document requests when you only provide a part of the information you want to update, and not all of them.

According to the git history, you added that code for adding and removing objects as part of [XWIKI-16210](https://jira.xwiki.org/browse/XWIKI-16210). The code in `readObjectsFromFormUpdateOrCreate` is from [XWIKI-10252](https://jira.xwiki.org/browse/XWIKI-10252). All methods support updating just some objects and do not remove anything if a property is not specified in the request, though `readObjectsFromFormUpdateOrCreate` also cleans-up properties that do not belong to the xclass. In my opinion, we should always process all object information from the form (i.e., also added and removed objects) to avoid issues like the one I'm fixing here but it seems also strange to now merge everything into the one method `readObjectsFromForm` which is the original method for updating objects from the form.

PR #1731, opened 2021-12-12
[Misc] Simplify boolean expressions

Vyom-Yadav:
With Respect to Issue https://github.com/checkstyle/checkstyle/issues/11015
Pull request - https://github.com/checkstyle/checkstyle/pull/11020
The link to failed build- https://app.circleci.com/pipelines/github/checkstyle/checkstyle/10533/workflows/5e688e70-96af-43e5-b25c-af4b52575ea5/jobs/108026

### DIFF
--- a/xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/input/InstanceInputProperties.java
+++ b/xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/input/InstanceInputProperties.java
@@ -67,7 +67,7 @@ public class InstanceInputProperties extends HashMap<String, Object> implements 
     {
         Boolean verbose = (Boolean) get(PROPNAME_VERBOSE);
 
-        return verbose == Boolean.FALSE ? false : true;
+        return verbose != Boolean.FALSE;
     }
 
     @Override
--- a/xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/output/InstanceOutputProperties.java
+++ b/xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/output/InstanceOutputProperties.java
@@ -35,7 +35,7 @@ public class InstanceOutputProperties extends HashMap<String, Object> implements
     {
         Boolean verbose = (Boolean) get(PROPNAME_VERBOSE);
 
-        return verbose == Boolean.FALSE ? false : true;
+        return verbose != Boolean.FALSE;
     }
 
     @Override


xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/input/InstanceInputProperties.java
```diff
@@ -67,7 +67,7 @@ public boolean isVerbose()
     {
         Boolean verbose = (Boolean) get(PROPNAME_VERBOSE);
 
-        return verbose == Boolean.FALSE ? false : true;
+        return verbose != Boolean.FALSE;
```
vmassol, line 70: This looks like it should be the opposite, no?

Like:

```
return verbose;
```

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/output/InstanceOutputProperties.java
```diff
@@ -35,7 +35,7 @@ public boolean isVerbose()
     {
         Boolean verbose = (Boolean) get(PROPNAME_VERBOSE);
 
-        return verbose == Boolean.FALSE ? false : true;
+        return verbose != Boolean.FALSE;
```
vmassol, line 38: Same here.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/input/InstanceInputProperties.java
Vyom-Yadav: This is-
```
if(verbose == Boolean.FALSE) {
    return false;
} else { // verbose != Boolean.FALSE;
    return true;
}
```
so this is equivalent to return `verbose != Boolean.FALSE`, when verbose is not equal to Boolean.FALSE `true` is returned which is being done right now.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/output/InstanceOutputProperties.java
Vyom-Yadav: same as above

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/input/InstanceInputProperties.java
vmassol: if verbose is false then return false
if verbose is true then return true
==> return verbose

Vyom-Yadav: Yes, you are right but I only simplified it to the point Checkstyle cannot detect it, Checkstyle does not detect `verbose != Boolean.FALSE`.

Vyom-Yadav: But if you want, I can simplify it and inline the variable, would that be fine?

vmassol: > But if you want, I can simplify it and inline the variable, would that be fine?

I think that `return verbose;` is simpler to read and understand than `return verbose != Boolean.FALSE;`.

Thanks for your PR

vmassol: Actually `return get(PROPNAME_VERBOSE);` is even simpler.

Vyom-Yadav: Done

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-api/src/main/java/org/xwiki/filter/instance/input/InstanceInputProperties.java
```diff
@@ -65,9 +65,7 @@ public void setEntities(EntityReferenceSet entities)
     @Override
     public boolean isVerbose()
     {
-        Boolean verbose = (Boolean) get(PROPNAME_VERBOSE);
-
-        return verbose == Boolean.FALSE ? false : true;
+        return get(PROPNAME_VERBOSE);
```
tmortagne, line 68: This is not the same thing. If `get(PROPNAME_VERBOSE)` returns null you will end up with a NullPointerException now so this change is wrong.

Vyom-Yadav: @tmortagne I assumed that the return type of `get(PROPNAME_VERBOSE)` is `boolean` and not `Boolean` as it was cast to `Boolean` a line before. I couldn't find the method `get(..)` so wasn't able to verify. But it can be still be simplified to `return verbose != Boolean.FALSE;` so I changed it back to `return verbose != Boolean.FALSE;` to be on the safer side.

Vyom-Yadav: @tmortagne I assumed that the return type of `get(PROPNAME_VERBOSE)` is `boolean` and not `Boolean` as it was cast to `Boolean` a line before. I couldn't find the method `get(..)` so wasn't able to verify. But still, it can be simplified to `return verbose != Boolean.FALSE;` so I reverted back to `return verbose != Boolean.FALSE;` to be on the safer side.

tmortagne: > get(PROPNAME_VERBOSE) is boolean and not Boolean as it was cast to Boolean a line before

It's cast to `Boolean` because it returns an `Object`.

> so I reverted back to return verbose != Boolean.FALSE;

Yes, that should work fine.

tmortagne:
Thanks for the contribution @Vyom-Yadav !

PR #1734, opened 2022-01-01
XWIKI-17568: clean sort parameter value

ClemensRobbenhaar:
if the "sort" parameter does not look like a property name or any other accepted value then just scrap it.

Note: this also removes the `sort` parameter if it is a comma separated list of values, like `doc.fullName, doc.date`. As the `dir` parameter to define the sort direction would only applied to the last value (I think), this is unlikely to be a use case which has been supported. Should I take this into account, anyway? (It will make the regexp to check the parameter somewhat more complicated, however.)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
@@ -81,7 +81,7 @@ $response.setContentType("application/json")
 ##
 #set ($order = "$!request.sort")
 #if (!$regextool.compile('^(doc\.)?\w+$').matcher($order).matches())
-  #set($order = '')
+  #set ($order = '')
 #end
 #set ($orderQueryPart = '')
 #if ($order != '')
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
@@ -81,7 +81,8 @@ $response.setContentType("application/json")
 ##
 #set ($order = "$!request.sort")
 #if (!$regextool.compile('^(doc\.)?\w+$').matcher($order).matches())
-  #set ($order = '')
+  { "error" : "invalid sort parameter" }
+  #stop ("invalid sort parameter [$order] found")
 #end
 #set ($orderQueryPart = '')
 #if ($order != '')
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
@@ -80,12 +80,12 @@ $response.setContentType("application/json")
 ## ORDER
 ##
 #set ($order = "$!request.sort")
-#if (!$regextool.compile('^(doc\.)?\w+$').matcher($order).matches())
-  { "error" : "invalid sort parameter" }
-  #stop ("invalid sort parameter [$order] found")
-#end
 #set ($orderQueryPart = '')
 #if ($order != '')
+  #if (!$regextool.compile('^(doc\.)?\w+$').matcher($order).matches())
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
```diff
@@ -80,6 +80,9 @@ $response.setContentType("application/json")
 ## ORDER
 ##
 #set ($order = "$!request.sort")
+#if (!$regextool.compile('^(doc\.)?\w+$').matcher($order).matches())
+  #set($order = '')
```
mflorea, line 84:
```suggestion
  #set ($order = '')
```

mflorea (approved):

tmortagne:
> As the dir parameter to define the sort direction would only applied to the last value (I think)

It does not feel enough to discard completely the possibility to pass a list of columns to order by, especially since `dir` is optional, if anything it's a limitation that should probably be fixed, but that's not the subject of this issue.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
ClemensRobbenhaar: oh, indeed, fixed

ClemensRobbenhaar:
> Also, I feel like returning an error might be more accurate that just ignore the input to avoid the WTF effect of a not good enough regex which is always possible.

I now changed the behavior to return a JSON object with an error. I do not expect that the calling side expects this ... however when I look around in the current code, this template is mostly called via live data, whose REST endpoint sanitizes the input parameters anyway, so improving this even more seems not to be worth the effort IMO.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
```diff
@@ -80,6 +80,10 @@ $response.setContentType("application/json")
 ## ORDER
 ##
 #set ($order = "$!request.sort")
+#if (!$regextool.compile('^(doc\.)?\w+$').matcher($order).matches())
+  { "error" : "invalid sort parameter" }
+  #stop ("invalid sort parameter [$order] found")
```
michitux, line 85: I'm wondering if logging this input parameter is a good idea (I think this is a debug log), this might allow [log injection](https://owasp.org/www-community/attacks/Log_Injection) unless we already have some code to prevent that. I don't know if we care about this as this is also just a debug log.

tmortagne: > if this indeed goes to the debug log

From what I see https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/DefaultVelocityEngine.java#L286 it seems we don't even log it at all.

michitux:
I've done some debugging related to the question how sorting with multiple columns work as this is something that is indeed supported by Live Data. You can test by going to `<server>/xwiki/bin/delete/Main/WebHome?useLiveData=true`. What I found: there are indeed as many `dir`-parameters as `sort`-parameters. However, `get()` only returns the first value. Therefore, it is perfectly fine to validate the order with this regular expression, to support multiple columns we need to support multiple parameters, not a comma separated list of columns in a single parameter.

michitux (approved):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getdocuments.vm
ClemensRobbenhaar: Sorry, I missed the discussion in the chat, was otp. Should I write a loop over all values of the `sort` request parameter?
(Also I noticed that my check will fail if no `sort` parameter is present, ouch, fixing)

michitux:
Supporting multiple sort values is an improvement that should be separate from this bug fix. You are right regarding the missing parameter, I think this should be fixed and then we can merge this.

ClemensRobbenhaar:
Ok, fixed the "missing sort parameter" check. I will merge this evening unless someone else objects.

PR #1738, opened 2022-01-06
XWIKI-19191: Upgrade to H2 MVStore 2.0.206

manuelleduc:
- Upgrade to H2 MVStore 2.0.206
- Remove the H2 MVStore if its format is v1


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/async/DefaultMentionsBlockingQueueProvider.java
```diff
@@ -78,4 +105,13 @@ public void closeQueue()
             this.mvStore.close();
         }
     }
+
+    private void backupQueueFile(File queueFile) throws IOException
+    {
+        long currentTimeMillis = System.currentTimeMillis();
+        Path backupFile = queueFile.toPath().getParent().resolve(String.format("mvqueue.%d.old", currentTimeMillis));
+        this.logger.info("Backup deprecated v1 h2mvstore file [{}] to [{}].", queueFile.getAbsolutePath(),
```
tmortagne, line 113: I would use a more generic message talking about an "unsupported format" since, as a report shown that, you can have this problem even without an upgrade. It would also be more future-proof, since the same will probably happen when upgrading to another major version.

manuelleduc: Updated https://github.com/xwiki/xwiki-platform/pull/1738/files#diff-14fa7661f76667e4b1cca584edd4b2b140bbe560b25791260d478712de90bc54R113

tmortagne (approved):

PR #1740, opened 2022-01-11
XWIKI-19308: Modularize the redirection logic in XWikiAction

manuelleduc:
- Introduce RedirectionFilter role
- Move the existing redirection logic in a new RedirectClassRedirectionFilter component
- XWikiAction#execute loads and execute the RedirectionFilter components by order of priority
- Make RedirectClassRedirectionFilter with a 500 priority to make it executed early

### DIFF
--- a/xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/web/XWikiActionCompatibilityAspect.aj
+++ b/xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/web/XWikiActionCompatibilityAspect.aj
@@ -17,15 +17,11 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
-package com.xpn.xwiki.internal.redirection;
+package com.xpn.xwiki.web;
 
 import java.io.IOException;
 
-import javax.inject.Named;
-import javax.inject.Singleton;
-
 import org.apache.commons.lang3.StringUtils;
-import org.xwiki.component.annotation.Component;
 import org.xwiki.model.EntityType;
 import org.xwiki.model.reference.DocumentReferenceResolver;
 import org.xwiki.model.reference.EntityReference;
@@ -39,22 +35,42 @@ import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.doc.XWikiDocument;
 import com.xpn.xwiki.internal.mandatory.RedirectClassDocumentInitializer;
 import com.xpn.xwiki.objects.BaseObject;
-import com.xpn.xwiki.redirection.RedirectionFilter;
-import com.xpn.xwiki.web.Utils;
 
 /**
- * Perform a redirection based on the presence of a {@code XWiki.RedirectClass} XClass.
- *
+ * Add a backward compatibility layer to {@link XWikiAction}.
+ * 
  * @version $Id$
- * @since 14.0RC1
  */
-@Component
-@Singleton
-@Named("XWiki.RedirectClass")
-public class RedirectClassRedirectionFilter implements RedirectionFilter
+public privileged aspect XWikiActionCompatibilityAspect
 {
-    @Override
-    public boolean redirect(XWikiContext context) throws XWikiException
+    // handleRedirectObject and handleRedirectObject while previously {@code protected} are now {@code public} due of 
+    // technical limitations of AspectJ.
+    // See https://doanduyhai.wordpress.com/2011/12/12/advanced-aspectj-part-ii-inter-type-declaration/
+    // "If their is one thing to remember from access modifier, it’s that their semantic applies with respect to the
+    // declaring aspect, and not to the target."
+    // They must still be used as if {@code protected}.
+    /**
+     * Indicate if the XWiki.RedirectClass is handled by the action (see handleRedirectObject()).
+     * 
+     * @deprecated since 14.0RC1, see {@link XWikiAction#supportRedirections()}
+     */
+    @Deprecated
+    public boolean XWikiAction.handleRedirectObject = false;
+
+    /**
+     * Redirect the user to an other location if the document holds an XWiki.RedirectClass instance (used when a
+     * document is moved).
+     *
+     * @param context the XWiki context
+     * @return either or not a redirection have been sent
+     * @throws XWikiException if error occurs
+     * @since 8.0RC1
+     * @since 7.4.2
+     * 
+     * @deprecated since 14.0RC1, not used anymore
+     */
+    @Deprecated
+    public boolean XWikiAction.handleRedirectObject(XWikiContext context) throws XWikiException
     {
         WikiReference wikiReference = context.getWikiReference();
 
@@ -86,6 +102,9 @@ public class RedirectClassRedirectionFilter implements RedirectionFilter
             locationReference = entityReference.replaceParent(parentDocument, locationReference);
         }
 
+        // Get the URL corresponding to the location
+        // Note: the anchor part is lost in the process, because it is not sent to the server
+        // (see: http://stackoverflow.com/a/4276491)
         String url = context.getWiki().getURL(locationReference, context.getAction(),
             context.getRequest().getQueryString(), null, context);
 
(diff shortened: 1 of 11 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/redirection/RedirectClassRedirectionFilter.java
```diff
@@ -0,0 +1,102 @@
...
+
+        // If the entity is inside a document, compute the new entity with the new document part.
+        if (entityReference.getType().ordinal() > EntityType.DOCUMENT.ordinal()) {
+            EntityReference parentDocument = entityReference.extractReference(EntityType.DOCUMENT);
+            locationReference = entityReference.replaceParent(parentDocument, locationReference);
+        }
+
+        // Get the URL corresponding to the location
+        // Note: the anchor part is lost in the process, because it is not sent to the server
+        // (see: http://stackoverflow.com/a/4276491)
```
tmortagne, line 89: I know it's a move, but I don't see the point of putting a stackoverflow post as reference... The explanation in the line above is enough.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -1221,4 +1158,27 @@ protected boolean readFromTemplate(XWikiDocument document, String template, XWik
...
+     * the action continues normally.
+     *
+     * @param context the current wiki content
+     * @return {@code true} if a redirection has been performed, {@code false} otherwise
+     * @throws XWikiException in case of error during the execution of a redirection filter
+     */
+    private boolean handleRedirect(XWikiContext context) throws XWikiException
+    {
+        // If no redirection are expected, this step is skipped.
+        if (this.handleRedirectObject) {
```
tmortagne, line 1173: This variable should maybe be deprecated for something a little more generic.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/redirection/RedirectClassRedirectionFilter.java
manuelleduc: Indeed, I haven't checked the moved code closely. I'll remove it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -1221,4 +1158,27 @@ protected boolean readFromTemplate(XWikiDocument document, String template, XWik
...
+            List<RedirectionFilter> filters = Utils.getComponentList(RedirectionFilter.class);
```
tmortagne, line 1174: XWikiActions are component since 13.0 so not need to go through Utils.getComponent*.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/redirection/RedirectClassRedirectionFilter.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Perform a redirection based on the presence of a {@code XWiki.RedirectClass} XClass.
+ *
+ * @version $Id$
+ * @since 14.0RC1
+ */
+@Component
+@Singleton
+public class RedirectClassRedirectionFilter implements RedirectionFilter
```
tmortagne, line 52: I don't think it's right to have an implementation of RedirectionFilter with the hint "default", that's just one source of redirection and it should be associated to something like "XWiki.RedirectClass" for example.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/META-INF/components.txt
```diff
@@ -146,6 +146,7 @@ com.xpn.xwiki.internal.template.TemplateListener
 com.xpn.xwiki.internal.template.VelocityTemplateEvaluator
 com.xpn.xwiki.internal.query.ConfiguredQueryExecutorProvider
 com.xpn.xwiki.internal.query.CurrentLanguageQueryFilter
+500:com.xpn.xwiki.internal.redirection.RedirectClassRedirectionFilter
```
tmortagne, line 149: Why 500 ?

manuelleduc: Arbitrary value to make it executed first and decrease the chances to see regression caused by other components. But might be over-engineered indeed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/redirection/RedirectClassRedirectionFilter.java
manuelleduc: +1

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
manuelleduc: Should it still be used by `RedirectClassRedirectionFilter`, or simply renamed into a more generic `redirectionsAllowed` field for instance?

tmortagne: Note that you have to keep this variable since it's an API, my suggestion was more to deprecate it and introduce one more (or a different system to indicate this information like a method to overwite depending on what you prefer) that would have a more accurate naming.

manuelleduc: I've introduced an `supportRedirections` method instead of `handleRedirectObject` (moved to legacy)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
manuelleduc: Replaced by `this.componentManager.<RedirectionFilter>getInstanceList(RedirectionFilter.class)`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/redirection/RedirectClassRedirectionFilter.java
manuelleduc: `XWiki.RedirectClass` hint added to the component.

xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/web/XWikiActionCompatibilityAspect.aj
```diff
@@ -0,0 +1,120 @@
...
+     * @param context the XWiki context
+     * @return either or not a redirection have been sent
+     * @throws XWikiException if error occurs
+     * @since 8.0RC1
+     * @since 7.4.2
+     * 
+     * @deprecated since 14.0RC1, not used anymore
+     */
+    @Deprecated
+    public boolean XWikiAction.handleRedirectObject(XWikiContext context) throws XWikiException
```
tmortagne, line 73: Shouldn't this method implementation be based on the new components system ? You are afraid that supporting more than XWiki.RedirectClass might be a breakage ?

manuelleduc: This method wasn't used anymore so I simply deprecated it and moved it as-is in legacy. Should I have changed its implementation to lookup for the right `RedirectionFilter` component and call it?

tmortagne: My feeling always is "the less code the better" :)

manuelleduc: I've replaced the content of `handleRedirectObject` with a call to `redirect()`.

PR #1741, opened 2022-01-11
XWIKI-19309: The notification clean link should not navigate to an empty anchor

slauriere:
* Prevent notification clean action event to navigate to distinct URL


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
```diff
@@ -319,6 +319,8 @@ Get the actual notifications.
       .addClass('notification-event-clean')
       .html('$services.icon.renderHTML("trash")&amp;nbsp;$escapetool.xml($escapetool.javascript($services.localization.render('notifications.menu.clear')))')
       .click(function (event) {
+        // Avoid navigating to href="#"
+        event.preventDefault();
```
surli, line 323: wouldn't that be better to avoid that behaviour by using a button instead of a link?

manuelleduc:
I've fixed it by accident in the meantime.

PR #1742, opened 2022-01-14
XWIKI-19306: Can't add annotations on IE 11

oanalavinia:
* adjust code to be IE compatible


surli (approved):
LGTM but far from being a JS expert

mflorea (approved):

PR #1743, opened 2022-01-17
XWIKI-1657: Allow to rename and move attachments

manuelleduc:
Move action on the attachments tab:
image

Move screen
image


xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/MoveAttachmentRequest.java
```diff
@@ -0,0 +1,61 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.attachment;
```
surli, line 20: wouldn't it be better to put it in a `org.xwiki.attachment.refactoring` package? I didn't check the remaining of the PR yet, but a bit a surprise to see that class not in refactoring module, so if it needs to be in attachment module maybe better to emphasize the refactoring in the package?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/AttachmentsListener.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+@Component
+@Singleton
+@Named(HINT)
+public class AttachmentsListener implements EventListener
+{
+    /**
+     * The hint for this class.
+     */
+    public static final String HINT = "Attachment.Code.RedirectAttachmentClass";
```
surli, line 59: Sounds like a hint that shouldn't be change because it's attached to some kind of mechanism? Would be good to document then. I use to use the same hints as the FQN of the class for listeners usually personally.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/AttachmentsManager.java
```diff
@@ -0,0 +1,127 @@
...
+import static org.xwiki.attachment.internal.RedirectAttachmentClassDocumentInitializer.TARGET_LOCATION_FIELD;
+import static org.xwiki.attachment.internal.RedirectAttachmentClassDocumentInitializer.TARGET_NAME_FIELD;
+
+/**
+ * Provide operations to inspect and manipulate attachments.
+ *
+ * @version $Id$
+ * @since 14.0RC1
+ */
+@Component(roles = AttachmentsManager.class)
```
surli, line 50: Wouldn't it worth it to provide an interface for this one? Sounds like the high level operations to perform on attachments here no?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/MoveAttachmentDocumentInitializer.java
```diff
@@ -0,0 +1,115 @@
...
+                    + "{{html clean=\"false\"}}\n"
+                    + "<script id=\"attachment-move-config\" type=\"application/json\">$jsontool.serialize({\n"
+                    + "  'treeWebjar': $!services.webjars.url('org.xwiki.platform:xwiki-platform-tree-webjar', "
+                    + "'require-config.min.js', {'evaluate': true})"
+                    + "})</script>\n"
+                    + "{{/html}}\n"
+                    + "{{/velocity}}");
+                xObject.setStringValue(SCOPE_PROPERTY, "wiki");
+            } catch (XWikiException e) {
+                this.logger.error(
```
surli, line 102: We usually log with using the exception message only in such case. Moreover maybe to reuse the constants for the error log message.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/RedirectAttachmentClassDocumentInitializer.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+@Component
+@Named(RedirectAttachmentClassDocumentInitializer.HINT)
+@Singleton
+public class RedirectAttachmentClassDocumentInitializer extends AbstractMandatoryClassInitializer
+{
+    /**
+     * Hint for this class.
+     */
+    public static final String HINT = "Attachment.Code.RedirectAttachmentClass";
```
surli, line 48: This hint is public, why not using it in the named of the filter and of the listener?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/MoveAttachmentRequest.java
manuelleduc: Yeah, I did use subpackages for internal since their is no other concerns in this module for now. But indeed could be interesting for public classes. I'll move it.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/job/MoveAttachmentJob.java
```diff
@@ -0,0 +1,157 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.attachment.internal.job;
```
surli, line 20: same here I would have "refactoring" somewhere in the package.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/job/MoveAttachmentJob.java
```diff
@@ -0,0 +1,157 @@
...
+    public String getType()
+    {
+        return HINT;
+    }
+
+    @Override
+    protected void process(EntityReference source)
+    {
+        this.progressManager.pushLevelProgress(this);
+        this.request.setId("refactoring", "moveAttachment",
```
surli, line 94: hmmm wondering how much it would be useful to reuse the source in the id? Not sure if it's a practice we do for other refactorings and/or if it might help in case of problem to find back the proper logs / info.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/job/MoveAttachmentJob.java
```diff
@@ -0,0 +1,157 @@
...
+            } else {
+                // "Attachment moved to " + destination
+                String historyMessageSource =
+                    this.contextualLocalizationManager.getTranslationPlain("attachment.job.saveDocument.source",
+                        this.referenceSerializer.serialize(destination));
+                String historyMessageTarget =
+                    this.contextualLocalizationManager.getTranslationPlain("attachment.job.saveDocument.target",
+                        this.referenceSerializer.serialize(source));
+                wiki.saveDocument(sourceDocument, historyMessageSource, this.xcontextProvider.get());
+                wiki.saveDocument(targetDocument, historyMessageTarget, this.xcontextProvider.get());
```
surli, line 132: So in case of exception here, you have no flag informing that the previous save should be rollback? I remember you asked about that, but I guess you could have put a boolean flag and check it in the catch to rollback the previous save no?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/AttachmentsListener.java
manuelleduc: I'm not sure of the documentation you are missing here :)  I'll change the hint to the fqn.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/AttachmentsManager.java
manuelleduc: I could but I'm not sure of why. Just to anticipate eventual overloading of the implementation?

surli: > Just to anticipate eventual overloading of the implementation?

yep

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/MoveAttachmentDocumentInitializer.java
manuelleduc: I'm not sure what the issue here. We have several instances of logging errors on document initializers following the same pattern (for instance, https://github.com/xwiki/xwiki-platform/blob/95062b970dc37c5529ebdbe19c47e38619a48176/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/java/org/xwiki/mentions/internal/descriptors/NotificationAdministrationDocumentInitializer.java#L103)

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/job/MoveAttachmentJob.java
manuelleduc: @tmortagne wdyt? Could be interesting but that's not what we did for the document move for instance.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/MoveAttachmentRequest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * Job request for moving an attachment to a new location. A redirection can be persisted at the old location.
+ *
+ * @version $Id$
+ * @since 14.0RC1
+ */
+@Unstable
+public class MoveAttachmentRequest extends EntityRequest
```
tmortagne, line 33: I'm wondering if this should not go in xwiki-platform-refactoring-api with the other entities refactoring jobs.

manuelleduc: Only the request or also the corresponding job? I'm not sure it's interesting as it add yet another dependency on the refactoring module towards `attachment-api`. That's fine but that will make `refactoring` dependent on a lot of modules.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/job/MoveAttachmentJob.java
manuelleduc: Indeed, I'll add a rollback to put the attachment back in the source if the target save fails.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/MoveAttachmentDocumentInitializer.java
surli: Nevermind mixing up the rule between warn messages and error messages.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/job/MoveAttachmentJob.java
tmortagne: First, I don't understand why you set the id during job execution, that does not really make any sense. The id is supposed to be set when you create the request before executing the job.

> that's not what we did for the document move for instance.

As for having the source in the id, the other refactoring jobs are using the source for the group id but indeed not for the request id from what I can see. It could indeed be interesting for debug to have it, but it means potentially having a path that simply does not work on various systems like Windows because it's too long.

surli: > it means potentially having a path that simply does not work on various systems like Windows because it's too long.

ok I forgot that aspect. So not such a good idea.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/AttachmentsListener.java
surli: > I'm not sure of the documentation you are missing here :)

it was in case the hint was derived from a mechanism involving this class but apparently it's not the case.

manuelleduc:
@surli @tmortagne 
- rollback of the source document in case of failure of the target document save done
- because of the revapi bug on `externalClassExposedInAPI` it is difficult to make the `AttachmentsManager` a public API. Consequently, it makes it difficult to move the job to refactoring, since it dependends on `AttachmentsManager`

I'll cleanup the templates to use `#displayJobStatus` and I'll merge it that's ok.

PR #1746, opened 2022-01-21
XWIKI-19323: Upgrade to Bootstrap Select 1.13.18

michitux:
* Adapt BootstrapSelect page object to no longer rely on removed attribute.
* Make selectors in LocalizationAdministrationSectionPage more robust.
* Improve performance of BootstrapSelect.

This fixes test failures introduced in https://jira.xwiki.org/browse/XWIKI-19323 and improves the performance of the BootstrapSelect page object by reducing the number of calls to the web driver.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
@@ -22,7 +22,6 @@ package org.xwiki.administration.test.po;
 import java.util.Arrays;
 import java.util.List;
 
-import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.FindBy;
 import org.xwiki.test.ui.po.BootstrapSelect;
@@ -52,7 +51,8 @@ public class LocalizationAdministrationSectionPage extends AdministrationSection
         super("Localization");
         waitUntilActionButtonIsLoaded();
         // Wait for asynchronous widgets to be loaded
-        getDriver().waitUntilElementIsVisible(By.xpath("(//div[contains(@class, 'bootstrap-select')])[3]"));
+        getDriver().waitUntilCondition(driver -> multiLingualSelect.isDisplayed() && defaultLanguageSelect.isDisplayed()
+            && supportedLanguagesSelect.isDisplayed());
     }
 
     public void setMultiLingual(boolean isMultiLingual)
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
@@ -35,13 +35,16 @@ import org.xwiki.test.ui.po.BootstrapSelect;
  */
 public class LocalizationAdministrationSectionPage extends AdministrationSectionPage
 {
-    @FindBy(xpath = "(//div[contains(@class, 'bootstrap-select')])[1]")
+    @FindBy(xpath = "//div[contains(@class, 'bootstrap-select')"
+        + " and ./select[@id='XWiki.XWikiPreferences_0_multilingual']]")
     private WebElement multiLingualSelect;
 
-    @FindBy(xpath = "(//div[contains(@class, 'bootstrap-select')])[3]")
+    @FindBy(xpath = "//div[contains(@class, 'bootstrap-select')"
+        + " and ./select[@id='XWiki.XWikiPreferences_0_default_language']]")
     private WebElement defaultLanguageSelect;
 
-    @FindBy(xpath = "(//div[contains(@class, 'bootstrap-select')])[2]")
+    @FindBy(xpath = "//div[contains(@class, 'bootstrap-select')"
+        + " and ../input[@id='XWiki.XWikiPreferences_0_languages']]")
     private WebElement supportedLanguagesSelect;
 
     public LocalizationAdministrationSectionPage()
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
```diff
@@ -35,21 +35,24 @@
...
+        + " and ../input[@id='XWiki.XWikiPreferences_0_languages']]")
     private WebElement supportedLanguagesSelect;
 
     public LocalizationAdministrationSectionPage()
     {
         super("Localization");
         waitUntilActionButtonIsLoaded();
         // Wait for asynchronous widgets to be loaded
-        getDriver().waitUntilElementIsVisible(By.cssSelector(".bootstrap-select"));
+        getDriver().waitUntilElementIsVisible(By.xpath("(//div[contains(@class, 'bootstrap-select')])[3]"));
```
surli, line 55: This "3" means there's several bootstrap select on the same xpath if I'm correct, a bit fragile no? Can't we use a better xpath?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BootstrapSelect.java
```diff
@@ -67,66 +68,79 @@ public void selectByValue(String value)
...
-        if (isMultiple()) {
-            // For each of its option
-            for (WebElement element : menu.findElements(By.tagName("li"))) {
-                // We find the language associated to this item thanks to the attribute "data-original-index"
-                int index = Integer.parseInt(element.getAttribute("data-original-index"));
-                String value = options.get(index);
+        // The list of values is usually small while the list of all options can be huge - use one XPath query per
+        // value.
+        for (String value : values) {
+            assertFalse(value.contains("\""), "Value must not contain a double quote");
```
surli, line 82: Feels a bit weird in PO but I understand why with the following line.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BootstrapSelect.java
```diff
@@ -67,66 +68,79 @@ public void selectByValue(String value)
...
+            WebElement element = this.hiddenSelect.findElement(By.xpath("option[@value = \"" + value + "\"]"));
 
-                // Now, click on the element to select/unselect it if its status is different from what we want
-                if (isElementSelected(element) != values.contains(value)) {
-                    element.findElement(By.tagName("a")).click();
+            if (element.getAttribute("selected") == null) {
+                // FIXME: this is a workaround as our <option>-tag contains child tags which is invalid HTML and not
+                //  supported by WebElement#getText(). The <option>-tag should be fixed.
```
surli, line 87: AFAIU we should actually fix our template to clean the option tag: you should create a ticket for that and probably include the ticket reference here.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
manuelleduc: I agree, does it match `multiLingualSelect`, `defaultLanguageSelect`, or `supportedLanguagesSelect`?

michitux: My understanding of XPaths is that this should match if there are three matching elements. The previous selector already matched after the first select was ready. I am not sure though that this is actually necessary as accessing the individual elements should already wait for them to become available.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BootstrapSelect.java
michitux: Yeah, and there is unfortunately no way to escape quotes in XPath, see [Apostrophe (') in XPath query - StackOverflow](https://stackoverflow.com/questions/1341847/apostrophe-in-xpath-query).

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
surli: Well the elements can be available in the DOM but not visible, I think that's what happened when Guillaume introduced this wait in https://github.com/xwiki/xwiki-platform/commit/b1affcf708dd5511edb54fefd089021e9d191b4a

michitux: Okay, but it doesn't make any sense imho to wait for the first bootstrap-select element to be visible. The first one is the simple one with the static three options, which is much faster than the other two. I know this for sure because I have seen screen recordings of failing tests where just this first select was initialized and the test started interacting with it and then started interacting with the other two when they became available.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BootstrapSelect.java
michitux: Added the ticket and reference.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-pageobjects/src/main/java/org/xwiki/administration/test/po/LocalizationAdministrationSectionPage.java
michitux: I have now added explicit waits for all three elements, tests passing in Firefox and Chrome.

PR #1747, opened 2022-01-21
XWIKI-19343: Add ability to exclude some locations when using the location picker

slauriere:
* Add 'exclusions': "$!options.exclusions" to the map of options passed to the documentTree macro


mflorea (approved):

mflorea:
We need to wait for 14.0RC1 to be released.

PR #1756, opened 2022-01-27
XWIKI-19026: The <head> is full of invalid HTML <link> and <script> elements

michitux:
* XML-escape all URLs in script/link-tags.
* Do not escape URLs during assembly as helper methods add parameters with & as separator.
* Ensure that "\n" is printed after every link/script.
* Cleanup duplicate code.

I have introduced some new helper methods for code deduplication in the parent class which is only relevant for some child classes. I'm wondering if it would be better to move them into separate helper classes. What do you think?

In this code I'm following the principle to XML-escape text the moment it is inserted into an XML tag and not earlier. In my opinion this is the best approach and necessary here as some other methods that are called just add additional parameters using `&` as separator and thus escaping the `&` early just creates the mess we previously had.

Jira issue: https://jira.xwiki.org/browse/XWIKI-19026

### DIFF
--- a/xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
+++ b/xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
@@ -29,6 +29,8 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xwiki.model.reference.DocumentReferenceResolver;
@@ -67,12 +69,26 @@ import com.xpn.xwiki.web.Utils;
 public abstract class AbstractSkinExtensionPlugin extends XWikiDefaultPlugin implements RenderingCacheAware
 {
     /**
+     * The name of the preference (in the configuration file) specifying what is the default value of the defer, in case
+     * nothing is specified in the parameters of this extension.
+     */
+    public static final String DEFER_DEFAULT_PARAM = "xwiki.plugins.skinx.deferred.default";
+
+    /**
      * The URL delimiter part of query parameters.
      */
     protected static final String QUERY_PARAMETER_DELIMITER = "?";
 
+    /**
+     * The separator between parameters.
+     */
+    protected static final String PARAMETER_SEPARATOR = "&";
+
     /** Log object to log messages in this class. */
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractSkinExtensionPlugin.class);
+
+    /** Parameter name force skin action. */
+    private static final String FORCE_SKIN_ACTION = "forceSkinAction";
 
     /** The name of the context key for the list of pulled extensions. */
     protected final String contextKey = this.getClass().getCanonicalName();
@@ -162,6 +178,32 @@ public abstract class AbstractSkinExtensionPlugin extends XWikiDefaultPlugin imp
     public Api getPluginApi(XWikiPluginInterface plugin, XWikiContext context)
     {
         return new SkinExtensionPluginApi((AbstractSkinExtensionPlugin) plugin, context);
+    }
+
+    /**
+     * @param filename The name of the file to get the URL for.
+     * @param context The current request context.
+     * @return The (unescaped) URL of the skin file.
+     * @since 14.0RC1
+     */
+    protected String getSkinFileURL(String filename, XWikiContext context)
+    {
+        boolean forceSkinAction = BooleanUtils.toBoolean((Boolean) getParameter(FORCE_SKIN_ACTION, filename,
+            context));
+
+        StringBuilder url = new StringBuilder(context.getWiki().getSkinFile(filename, forceSkinAction, context));
+        if (forceSkinAction) {
+            String parameters =
+                StringUtils.removeStart(parametersAsQueryString(filename, context), PARAMETER_SEPARATOR);
+            if (!StringUtils.isEmpty(parameters)) {
+                String queryParamDelimiter =
+                    StringUtils.contains(url, QUERY_PARAMETER_DELIMITER) ? PARAMETER_SEPARATOR
+                        : QUERY_PARAMETER_DELIMITER;
+                url.append(queryParamDelimiter).append(parameters);
+            }
+        }
+
+        return url.toString();
     }
 
     private void useResource(String resource, XWikiContext context)
@@ -332,8 +374,8 @@ public abstract class AbstractSkinExtensionPlugin extends XWikiDefaultPlugin imp
     /**
      * This method converts the parameters for an extension to a query string that can be used with
      * {@link com.xpn.xwiki.doc.XWikiDocument#getURL(String, String, String, XWikiContext) getURL()} and printed in the
-     * XHTML result. The parameters separator is the escaped &amp;amp;. The query string already starts with an
-     * &amp;amp; if at least one parameter exists.
+     * XHTML result. The parameters separator is the escaped &amp;. The query string already starts with an
+     * &amp; if at least one parameter exists.
      *
      * @param resource The pulled resource whose parameters should be converted.
      * @param context The current request context.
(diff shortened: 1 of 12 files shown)


xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
```diff
@@ -164,6 +180,32 @@ public Api getPluginApi(XWikiPluginInterface plugin, XWikiContext context)
         return new SkinExtensionPluginApi((AbstractSkinExtensionPlugin) plugin, context);
     }
 
+    /**
+     * @param filename The name of the file to get the URL for.
+     * @param context The current request context.
+     * @return The (unescaped) URL of the skin file.
+     * @since 14.0RC1
```
manuelleduc, line 187: Probably not the right version ;)

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/LinkExtensionPlugin.java
```diff
@@ -69,7 +69,7 @@ public String getLink(String link, XWikiContext context)
             result.append(StringEscapeUtils.escapeXml(entry.getValue().toString()));
             result.append("'");
         }
-        result.append("/>");
+        result.append("/>\n");
```
manuelleduc, line 72: Could it be interesting to have a comment explaining why this line break is needed there?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/stylesheets.vm
```diff
@@ -21,9 +21,9 @@
 
 #macro(concatenateQueryParameter $originalUrl $queryString)
   #if ($originalUrl.contains("?"))
-    $originalUrl&$queryString
+$originalUrl&#38;$queryString
   #else
-    $originalUrl?$queryString
+$originalUrl?$queryString
```
manuelleduc, line 26: Wrong indentation?

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/LinkExtensionPlugin.java
michitux: It is not needed, it is just nicer for the HTML output as there is nothing else adding line breaks between these elements. This is now present in all of these skinx-plugins while previously only some of them had it. I haven't seen any comment in any of these having the line break.

manuelleduc: ok, if it's nothing more than having a nicer output, it's fine to leave it like that.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/stylesheets.vm
michitux: This was an attempt to prevent additional whitespace in the output. Though in fact this is not enough, I should have removed that change again as it doesn't actually fix the issue. What would actually be needed is

```velocity
#if($originalUrl.contains("?"))$originalUrl&#38;$queryString#else$originalUrl?$queryString#end
```

With proper indentation this code leads to

```html
<link href="      /xwiki/bin/skin/skins/flamingo/style.min.css?cache-version=1643188958000&#38;skin=XWiki.DefaultSkin&#38;colorTheme=xwiki%3AFlamingoThemes.Iceberg&#38;colorThemeVersion=1.1
  " rel="stylesheet" type="text/css" media="all" />
<link href="      /xwiki/bin/skin/skins/flamingo/print.min.css?cache-version=1643188958000&#38;skin=XWiki.DefaultSkin&#38;colorTheme=xwiki%3AFlamingoThemes.Iceberg&#38;colorThemeVersion=1.1
  " rel="stylesheet" type="text/css" media="print" />
```

I don't think this is invalid but it definitely doesn't look nice. Note how this macro also introduces a linebreak in the attribute. Even the indentation of the `#if` creates whitespace.

manuelleduc: I'm not sure it's applicable in this case but it's possible to prevent line-breaks by adding a comment at the end of a line.
For instance:
```velocity
$originalUrl&#38;$queryString##
```

michitux: It is working, so we can prevent the line break. All indentation causes spaces to appear, so if we have properly indented code we will have whitespace in the output.

I've just checked, it is definitely valid [with whitespace](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#valid-non-empty-url-potentially-surrounded-by-spaces) so it is on us if we want to have cleaner code or cleaner output.

manuelleduc: I'm leaning toward cleaner output. A comment explaining why we don't follow the usual formatting could be interesting though.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
michitux: Yep, fixed. At least I assume it will be 14.1RC1 as the fix concerns quite a lot of code and nothing is actually broken so I'm not sure this should be cherry-picked to 13.10 or 14.0.

manuelleduc (approved):
Looks good to me. Just maybe, the many small white space changes are not helping the diff readability.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
```diff
@@ -66,14 +68,28 @@
 @SuppressWarnings("deprecation")
 public abstract class AbstractSkinExtensionPlugin extends XWikiDefaultPlugin implements RenderingCacheAware
 {
+    /**
+     * The name of the preference (in the configuration file) specifying what is the default value of the defer, in case
+     * nothing is specified in the parameters of this extension.
+     */
+    public static final String DEFER_DEFAULT_PARAM = "xwiki.plugins.skinx.deferred.default";
```
tmortagne, line 75: Missing `@since`.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
```diff
@@ -66,14 +68,28 @@
...
+
     /**
      * The URL delimiter part of query parameters.
      */
     protected static final String QUERY_PARAMETER_DELIMITER = "?";
```
tmortagne, line 80: Missing `@since`.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
```diff
@@ -66,14 +68,28 @@
...
 
+    /**
+     * The separator between parameters.
+     */
+    protected static final String PARAMETER_SEPARATOR = "&";
```
tmortagne, line 85: Missing `@since`.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
```diff
@@ -445,4 +487,12 @@ protected SkinExtensionAsync getSkinExtensionAsync()
 
         return this.async;
     }
+
+    protected boolean isDefer(String name, XWikiContext context)
```
tmortagne, line 491: Missing `@since`.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-skinx/src/main/java/com/xpn/xwiki/plugin/skinx/AbstractSkinExtensionPlugin.java
tmortagne: Ha, sorry, it was not a new thing...

PR #1757, opened 2022-01-27
XWIKI-19341: commentfield.vm should not manipulate the XWikiComment class

manuelleduc:
Use the edit script service to display the editor instead of XWikiDocument.displayEdit.

Jira issue: https://jira.xwiki.org/browse/XWIKI-19341

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
@@ -26,7 +26,6 @@
 #set ($number = $numbertool.toNumber($request.number).intValue())
 
 #set ($xCommentClass = 'XWiki.XWikiComments')
-#set ($commentClass = $xwiki.getDocument($xCommentClass).getxWikiClass())
 
 ## If a content is provided, we use it. If a comment number is provided, we use the content of the corresponding 
 ## comment. Otherwise, the content is an empty string.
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -22,63 +22,47 @@
...
+  'id': "${commentPrefix}_comment",
+  'name': "${commentPrefix}_comment",
+  'rows': $commentProperty.getProperty('rows').value,
+  'cols': $commentProperty.getProperty('cols').value
+})
+#if ($commentProperty.getPropertyClass().isWysiwyg($xcontext.context))
+  $services.edit.syntaxContent.wysiwyg($value, $doc.syntax, $properties)
+#else
+  $services.edit.syntaxContent.text($value, $doc.syntax, $properties)
+#end
```
manuelleduc, line 66: Opening a PR because I'm wondering if this can be simplified further.

mflorea: I don't think this is a good idea. Any xproperty, including ``comment``, can have a **custom displayer**:

* either coming from the property type (e.g. ``TextArea``), in which case it can be defined in Java, e.g. ``TextAreaClass``, or in a Velocity template such as ``displayer_date.vm`` for Date properties
* or coming from the xclass, e.g. ``XWiki.XWikiComments``, in which case it is defined using the ``customDisplay`` meta property

What you propose is basically to bypass the custom displayer for the ``comment`` property. I don't see the motivation behind this. In any case, such a change deserves a discussion (involving also developers that do heavy customizations on top of XWiki and which rely often on custom displayers).

tmortagne: I, too, am really not a fan of bypassing part of the property display process.

manuelleduc: Indeed I forgot the custom displayers. Thanks for reminding me of them.
In this case I propose to define somewhere (to be determined) a method allowing to use the displayer of a field, while being able to override its value and name.

tmortagne: > while being able to override its value

And being very careful which author is used to execute this displayer ;)

manuelleduc: I've updated the field rendering logic [here](https://github.com/xwiki/xwiki-platform/pull/1757/files#diff-4883aafa61b3e034573bc2799b4cf5f9352db3f540c3c9516282d0aa12c6d558R56-R60). A new 'in-memory' object is created and initialized with the value submitted by the user.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentfield.vm
```diff
@@ -22,63 +22,41 @@
...
-
-#set ($fieldToDisplay = $commentObjClass.get($name))
-
 ## We need to know the type of editor used for comments in order to hide/display the preview button.
 #set ($defaultEditorId = $services.edit.syntaxContent.defaultEditorId)
 <input type='hidden' name='defaultEditorId' value="$escapetool.xml($defaultEditorId)" />
 
-<label for="$name">$services.localization.render('core.viewers.comments.add.comment.label')</label>
+#if ($number && $number > -1)
+  #set ($commentPrefix = "XWiki.XWikiComments_${number}_")
```
tmortagne, line 47: Wasn't the plan to remove any explicit reference to "XWiki.XWikiComments" ?

manuelleduc: Not exactly. The issue title is misleading. The goal is to stop manipulating the `XWiki.XWikiComments` XClass at runtime. The current script add temporarily a field with the desired name to the XClass, to remove it once unused. Which is a terrible solution.

tmortagne:
I went for a much more reusable/java approach to this need.

PR #1760, opened 2022-01-27
XWIKI-19295: Accessing $xcontext.uix requires PR rights

manuelleduc:
Jira issue: https://jira.xwiki.org/browse/XWIKI-19295


xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/api/ContextCompatibilityAspect.aj
```diff
@@ -51,4 +51,19 @@ public privileged aspect ContextCompatibilityAspect
...
+     * Return the current uix object. This method is defined to allow to access this object without programming rights.
+     *
+     * @return the current uix object
+     *
+     * @since 14.1RC1
+     * @since 13.10.3
+     * @deprecated since 14.1RC1, use the "uix" key from the velocity context instead in the UIX templates
+     */
+    @Deprecated
+    public java.lang.Object Context.getUix()
```
manuelleduc, line 65: This will not be deprecated in 13.10.3.

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtension.java
```diff
@@ -158,13 +162,16 @@ private Object before(boolean inline) throws RenderingException
...
         // Remember the previous uix context to restore it
-        Map<String, Object> previousUIXContext = (Map<String, Object>) xcontext.get(WikiUIExtension.CONTEXT_UIX_KEY);
-        // Put the UIX context in the XWiki context
-        xcontext.put(WikiUIExtension.CONTEXT_UIX_KEY, uixContext);
+        Map<String, Object> previousUIXContext = (Map<String, Object>) xcontext.get(CONTEXT_UIX_KEY);
+        // Put the UIX context in the XWiki context. Note that this is deprecated and using the UIX templates is
+        // preferred.
+        xcontext.put(CONTEXT_UIX_KEY, uixContext);
+        // Put the UIX context in the velocity context "uix" key.
+        this.velocityManager.getVelocityContext().put(CONTEXT_UIX_KEY, uixContext);
```
manuelleduc, line 174: This is somehow a breaking change if `uix` was already set in the velocity context.

tmortagne: Variables should be set in the ScriptContext and never in the VelocityContext.

manuelleduc: VelocityContext replaced by ScriptContext. In addition I've fixed the context restoration. Now each context is respectively reset with its own previous values (before my last commit, both context where reset with the same previous value).

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtension.java
```diff
@@ -158,35 +177,40 @@ private Object before(boolean inline) throws RenderingException
...
-        // Put the UIX context in the XWiki context
-        xcontext.put(WikiUIExtension.CONTEXT_UIX_KEY, uixContext);
-
-        return previousUIXContext;
+        this.previousUIXContext = (Map<String, Object>) xcontext.get(CONTEXT_UIX_KEY);
+        // Put the UIX context in the XWiki context. Note that this is deprecated and using the UIX templates is
+        // preferred.
+        xcontext.put(CONTEXT_UIX_KEY, uixContext);
+        // Put the UIX context in the velocity context "uix" key.
+        ScriptContext scriptContext = this.scriptContextManager.getScriptContext();
```
tmortagne, line 189: Better use getCurrentScriptContext() in your use case.

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtension.java
```diff
@@ -78,11 +82,25 @@
...
     /**
      * Parameter manager for this extension.
      */
     private WikiUIExtensionParameters parameters;
 
+    /**
+     * Save the value of "uix" in the {@link XWikiContext} before the initialization, to restore it after the
+     * rendering.
+     */
+    private Map<String, Object> previousUIXContext;
```
tmortagne, line 96: You cannot have a state in WikiUIExtension since the same instance might be used in different threads, that's why before was returning the value to restore, you will just need to return both values to restore.

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtension.java
manuelleduc: But from the javadoc :)

> In doubt always use {@link #getScriptContext()}.

What are the pros of using `getCurrentScriptContext()` in practice?

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtension.java
manuelleduc: Hum indeed. I'll use a inner class to return several values.

xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtension.java
tmortagne: > In doubt always use {https://github.com/link?type=source #getScriptContext()}.

That does not mean getCurrentScriptContext() is totally useless, and the javadoc also explains that it's for your use case (modifying the context).

> What are the pros of using getCurrentScriptContext() in practice?

As indicated in the javadoc `getScriptContext()` does many things you don't need, so it's better for performances.

PR #1762, opened 2022-01-28
XWIKI-19271: Change HTMLConverter to parse and render HTML5

michitux:
* Use HTML5 for inplace editing
* Use HTML in DefaultHTMLConverter
* Configure HTMLCleaner to accept HTML5

Jira issue: https://jira.xwiki.org/browse/XWIKI-19271

This depends on xwiki/xwiki-rendering#195 (XRENDERING-375: Add a parser for HTML 5/XHTML 5).
Further, this also depends on xwiki-contrib/application-ckeditor#46 as without the changes in CKEditor this will break the editing experience.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/script/WysiwygEditorScriptService.java
+++ b/xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/script/WysiwygEditorScriptService.java
@@ -122,6 +122,14 @@ public class WysiwygEditorScriptService implements ScriptService
     }
 
     /**
+     * @return The syntax identifier of the rendered HTML.
+     */
+    public Syntax getHTMLSyntax()
+    {
+        return Syntax.ANNOTATED_HTML_5_0;
+    }
+
+    /**
      * Parses the given HTML fragment and renders the result in annotated XHTML syntax.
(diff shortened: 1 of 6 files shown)


xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/script/WysiwygEditorScriptService.java
```diff
@@ -121,6 +121,14 @@ public boolean isSyntaxSupported(String syntaxId)
         }
     }
 
+    /**
+     * @return The syntax identifier of the rendered HTML.
```
mflorea, line 125: Missing ``@since``.

mflorea (changes requested):
Looks good. Just a missing ``@since``.

xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/script/WysiwygEditorScriptService.java
michitux: I've added `@since` and also `@Unstable` as this is a new public API (at least that's my understanding).

mflorea (approved):

mflorea (approved):

PR #1763, opened 2022-02-01
[Misc] Move old XWikiDocument#readFromTemplate method to legacy

surli:


surli:
@mflorea / @tmortagne creating a PR for this one since I'm not 100% sure: I deprecated `XWikiDocument#readFromTemplate(EditForm)` because there was no usage in platform apparently and it seemed a bit dangerous to me (no check on the template reference rights, contrarily to what we do in XWikiAction). And once this one is deprecated there's no reason to keep the other one already deprecated so I moved both to legacy. 

Do you agree with that move? Not sure if it needs a ticket.

manuelleduc (approved):

tmortagne (approved):

mflorea (approved):

PR #1764, opened 2022-02-02
XWIKI-19378: Asynchronous document static analyzis

manuelleduc:
- Introduce the XWikiTask table in oldcore
- Introduce the APIs to queue and consume tasks in xwiki-platform-index-api
- Replace the use of H2 MVStore with the new API on the mentions

Jira: https://jira.xwiki.org/browse/XWIKI-19378


manuelleduc:
This is currently a draft pull request:
- The JMX operations are not available
- A lot of tests are missing

This Draft PR is currently opened for early review of the architecture.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/TaskConsumer.java
```diff
@@ -17,36 +17,30 @@
...
-     * @since 12.6.1
-     * @since 12.7RC1
+     * @param wikiId the identifier of the wiki where the task is to be executed
+     * @param docName the reference of the document to analyze
+     * @param version the version of the document to analyze
+     * @param author the reference of the user to use to perform the analyzis
+     * @throws IndexException in case of error during the execution of the task
      */
-    void closeQueue();
+    void consume(String wikiId, String docName, Version version, String author) throws IndexException;
```
manuelleduc, line 45: Author might actually be useless and can be derived from the `XWikiDocument` properties

manuelleduc:
Note: With these changes, MentionsIT is failing for a reason that I did not manage to understand so far. Still debugging that.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/TaskManager.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param version the document version
+     * @param kind the kind of task to add
+     * @return a completable future for this task
+     */
+    CompletableFuture<TaskData> replaceTask(String wikiId, long docId, String version, String kind);
+
+    /**
+     * Starts the consumer thread.
+     */
+    void startThread();
```
tmortagne, line 63: Is this public API really needed ? I tend to do this kind of thing in the `initialize()` of the component.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/TaskManager.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Add a task to the queue.
+     *
+     * @param wikiId the wiki containing the document
+     * @param docId the document id
+     * @param version the document version
+     * @param kind the kind of task to add
+     * @return a completable future for this task
+     */
+    CompletableFuture<TaskData> addTask(String wikiId, long docId, String version, String kind);
```
tmortagne, line 47: I'm wondering if "type" would not be better instead of "kind".

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultTasksManager.java
```diff
@@ -0,0 +1,294 @@
...
+    public void dispose()
+    {
+        this.jmxRegistration.unregisterMBean(MBEAN_NAME);
+        this.queue.add(TaskData.STOP);
+    }
+
+    @Override
+    public void startThread()
+    {
+        Thread thread = new Thread(new TasksRunnable());
```
tmortagne, line 164: Feels simpler to just implement Runnable directly in DefaultTasksManager.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/script/MentionsScriptService.java
```diff
@@ -73,13 +73,12 @@ public String getSelfMentionsColor()
...
 
     /**
-     * @see MentionsEventExecutor#getQueueSize()
      * @return the current size of the queue of elements (page, comments...) with mentions to analyze
      * @since 12.6
      */
     public long getQueueSize()
     {
-        return this.eventExecutor.getQueueSize();
+        return this.eventExecutor.getQueueSize("mention");
```
tmortagne, line 81: This value should probably be put in some public const.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/tasks/XWikiTaskId.java
```diff
@@ -0,0 +1,189 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xpn.xwiki.doc.tasks;
+
+import java.io.Serializable;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.suigeneris.jrcs.rcs.Version;
```
tmortagne, line 27: It would be better to stay out of JRCS classes as much as possible (so simply manipulate Strings here for the versions) IMO since we would like to get rid of JRCS (and which dead since more than 10 years).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xwiki.db2.hbm.xml
```diff
@@ -450,4 +450,19 @@
             <generator class="assigned" />
         </id>
     </class>
+
+    <!-- XWikiTask custom mapping-->
+    <class name="com.xpn.xwiki.doc.tasks.XWikiTask" table="xwikitask">
```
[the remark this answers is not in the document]
manuelleduc: Replaced by `XWikiDocumentIndexingTask` for the class, and `xwikidocumentindexingqueue` for the table?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/TaskManager.java
manuelleduc: I agree but if I remove this method from the API, I have to make a cast in `TaskApplicationReadyListener`.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/TaskManager.java
manuelleduc: Renamed!

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/TaskManager.java
tmortagne: I don't understand the relationship. For me if you initialize the thread in DefaultTaskManager#initialize() then TaskApplicationReadyListener becomes totally useless (since it's only point seems to be to start the thread).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xwiki.postgresql.hbm.xml
```diff
@@ -454,4 +454,18 @@
...
+
+    <!-- XWikiDocumentIndexingTask custom mapping-->
+    <class name="com.xpn.xwiki.doc.tasks.XWikiDocumentIndexingTask" table="xwikidocumentindexingqueue">
+        <!-- Stores the queued tasks as well as the running tasks. -->
+        <composite-id name="id" class="com.xpn.xwiki.doc.tasks.XWikiDocumentIndexingTaskId">
+            <key-property name="docId" column="XWT_DOC_ID" type="long" />
+            <key-property name="version" column="XWT_VERSION" type="string" length="10" />
+            <key-property name="type" column="XWT_TYPE" type="string" length="100" />
+            <key-property name="instanceId" column="XWT_INSTANCE_ID" type="string" length="100" />
+        </composite-id>
```
manuelleduc, line 466: @tmortagne I have shortened the size of the string types, otherwise mysql is failing because the size of the index is too large (and I guess this is a problem for postgres too).

manuelleduc: Actually `XWT_VERSION` should be increased to 39 (2 times the length of the largest string representation of a long + 1 char for the dot).
For the other two, 100 is arbitrary but I doubt anybody will need an hint longer than 100 for instance.

tmortagne: >  For the other two, 100 is arbitrary but I doubt anybody will need an hint longer than 100 for instance.

An instance ID is a UUID, right ? I think those have a max size.

manuelleduc: For the default implementation yes (and its 36 chars). But I we can also leave some slack for other implementations if we want to.

PR #1766, opened 2022-02-03
XWIKI-19394: Introduce the notion of cluster identifier

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-19394

### DIFF
--- a/xwiki-platform-core/xwiki-platform-observation/xwiki-platform-observation-remote/src/main/java/org/xwiki/observation/remote/internal/DefaultRemoteObservationManagerConfiguration.java
+++ b/xwiki-platform-core/xwiki-platform-observation/xwiki-platform-observation-remote/src/main/java/org/xwiki/observation/remote/internal/DefaultRemoteObservationManagerConfiguration.java
@@ -95,7 +95,7 @@ public class DefaultRemoteObservationManagerConfiguration implements RemoteObser
         Path observation = getObservationDirectory();
         observation.toFile().mkdir();
         Path idFile = getIdFile();
-        if (!idFile.toFile().exists()) {
+        if (!Files.exists(idFile)) {
             try {
                 String idString = UUID.randomUUID().toString();
                 Files.writeString(idFile, idString);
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-observation/xwiki-platform-observation-remote/src/main/java/org/xwiki/observation/remote/internal/DefaultRemoteObservationManagerConfiguration.java
```diff
@@ -69,4 +82,45 @@ public String getNetworkAdapter()
...
+            } catch (IOException e) {
+                throw new InitializationException(String.format("Failed to read the observation id file [%s]", idFile),
+                    e);
+            }
+        }
+    }
+
+    private Path getIdFile()
+    {
+        return getObservationDirectory().resolve("id");
```
tmortagne, line 119: I'm never a big fan of files without any extension, what about `id.txt` ?

xwiki-platform-core/xwiki-platform-observation/xwiki-platform-observation-remote/src/main/java/org/xwiki/observation/remote/internal/DefaultRemoteObservationManagerConfiguration.java
```diff
@@ -69,4 +82,45 @@ public String getNetworkAdapter()
...
+        return this.id;
+    }
+
+    @Override
+    public void initialize() throws InitializationException
+    {
+        Path observation = getObservationDirectory();
+        observation.toFile().mkdir();
+        Path idFile = getIdFile();
+        if (!idFile.toFile().exists()) {
```
tmortagne, line 98: You don't need to go through the old File API to test that a Path exists :) I think its `Files.exists(idFile)`.

tmortagne (approved):

PR #1773, opened 2022-02-08
[Misc] Fix InformationIT for new parsers

michitux:
Caused by https://jira.xwiki.org/browse/XRENDERING-375


tmortagne:
In such a case, ideally you should have XRENDERING-375 in the commit message, since it's fixed as part of this issue.

michitux:
I've changed the commit message. I didn't do this first, as XRENDERING-375 is not for xwiki-platform and I thought this practice of using the issue in the commit message was just for issues of the same project.

tmortagne:
> I didn't do this first, as XRENDERING-375 is not for xwiki-platform and I thought this practice of using the issue in the commit message was just for issues of the same project.

No, it's find as long as the issue is about a lower level (we often have to commit stuff in platform for big commons changes for example).

PR #1777, opened 2022-02-11

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-19352

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
@@ -1,0 +1,81 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.index.internal;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.index.IndexException;
+import org.xwiki.index.TaskConsumer;
+import org.xwiki.model.reference.DocumentReference;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.DocumentRevisionProvider;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.store.XWikiHibernateStore;
+
+/**
+ * Update the links indexing of a document.
+ *
+ * @version $Id$
+ * @since 14.2RC1
+ */
+@Component
+@Singleton
+@Named(DefaultLinksTaskConsumer.LINKS_TASK_TYPE)
+public class DefaultLinksTaskConsumer implements TaskConsumer
+{
+    /**
+     * Identifier of the links tasks type.
+     */
+    public static final String LINKS_TASK_TYPE = "links";
+
+    @Inject
+    private Provider<XWikiContext> xcontextProvider;
+
+    @Inject
+    private DocumentRevisionProvider documentRevisionProvider;
+
+    @Override
+    public void consume(DocumentReference documentReference, String version) throws IndexException
+    {
+        XWikiContext context = this.xcontextProvider.get();
+        // Skip if backlinks are not supported. Note that this task is not supposed to be queued when backlinks are
+        // not supported.
+        if (context.getWiki().hasBacklinks(context)) {
+            try {
+                XWikiDocument doc = this.documentRevisionProvider.getRevision(documentReference, version);
+
+                XWikiHibernateStore hibernateStore = context.getWiki().getHibernateStore();
+                hibernateStore.beginTransaction(context);
+                hibernateStore.saveLinks(doc, context, false);
+                hibernateStore.endTransaction(context, true);
+            } catch (XWikiException e) {
+                throw new IndexException(
+                    String.format("Failed to updated links for document [%s] version [%s].", documentReference,
+                        version), e);
+            }
+        }
+    }
+}
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
@@ -24,6 +24,7 @@ import javax.inject.Named;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
+import org.slf4j.Logger;
 import org.xwiki.component.annotation.Component;
 import org.xwiki.index.IndexException;
 import org.xwiki.index.TaskConsumer;
@@ -57,6 +58,9 @@ public class DefaultLinksTaskConsumer implements TaskConsumer
     @Inject
     private DocumentRevisionProvider documentRevisionProvider;
 
+    @Inject
+    private Logger logger;
+
     @Override
     public void consume(DocumentReference documentReference, String version) throws IndexException
     {
@@ -66,16 +70,33 @@ public class DefaultLinksTaskConsumer implements TaskConsumer
         if (context.getWiki().hasBacklinks(context)) {
             try {
                 XWikiDocument doc = this.documentRevisionProvider.getRevision(documentReference, version);
-
-                XWikiHibernateStore hibernateStore = context.getWiki().getHibernateStore();
-                hibernateStore.beginTransaction(context);
-                hibernateStore.saveLinks(doc, context, false);
-                hibernateStore.endTransaction(context, true);
+                // Verify that the document has not been removed in the meantime.
+                if (doc != null) {
+                    XWikiHibernateStore hibernateStore = context.getWiki().getHibernateStore();
+                    hibernateStore.beginTransaction(context);
+                    try {
+                        hibernateStore.saveLinks(doc, context, false);
+                    } finally {
+                        endTransaction(context, hibernateStore);
+                    }
+                }
             } catch (XWikiException e) {
                 throw new IndexException(
                     String.format("Failed to updated links for document [%s] version [%s].", documentReference,
                         version), e);
             }
+        } else {
+            this.logger.warn(
+                "Skipping for document [{}] version [{}] because backlinks are not supported in this wiki.",
+                documentReference, version);
+        }
+    }
+
+    private void endTransaction(XWikiContext context, XWikiHibernateStore hibernateStore)
+    {
+        try {
+            hibernateStore.endTransaction(context, true);
+        } catch (Exception e) {
         }
     }
 }
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/listener/LinksUpdateListener.java
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/listener/LinksUpdateListener.java
@@ -21,7 +21,6 @@ package org.xwiki.index.internal.listener;
 
 import javax.inject.Inject;
 import javax.inject.Named;
-import javax.inject.Provider;
 import javax.inject.Singleton;
 
 import org.xwiki.bridge.event.DocumentCreatedEvent;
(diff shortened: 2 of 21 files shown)


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+        // Skip if backlinks are not supported. Note that this task is not supposed to be queued when backlinks are
+        // not supported.
+        if (context.getWiki().hasBacklinks(context)) {
+            try {
+                XWikiDocument doc = this.documentRevisionProvider.getRevision(documentReference, version);
+
+                XWikiHibernateStore hibernateStore = context.getWiki().getHibernateStore();
+                hibernateStore.beginTransaction(context);
+                hibernateStore.saveLinks(doc, context, false);
+                hibernateStore.endTransaction(context, true);
```
manuelleduc, line 73: Sometimes, this piece of code yields the following logs. Afaiu this is not a problem per se, but multiplied by the analysis of every documents on the farm, this is very verbose.
```
2022-02-11 17:50:33,881 [task-manager-consumer] WARN  o.h.e.j.s.SqlExceptionHelper   - SQL Warning Code: -1100, SQLState: 02000 
2022-02-11 17:50:33,881 [task-manager-consumer] WARN  o.h.e.j.s.SqlExceptionHelper   - no data 
2022-02-11 17:50:33,882 [task-manager-consumer] WARN  o.h.e.j.s.SqlExceptionHelper   - SQL Warning Code: -1100, SQLState: 02000 
2022-02-11 17:50:33,882 [task-manager-consumer] WARN  o.h.e.j.s.SqlExceptionHelper   - no data 
```

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R14010XWIKI19352DataMigration.java
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Migrates the links by reindexing all the documents of the farm.
+ *
+ * @version $Id$
+ * @since 14.2RC1
+ */
+// TODO: Implement DataMigration once XWIKI-19399 is fixed.
+@Component
+@Singleton
+@Named(R14010XWIKI19352DataMigration.HINT)
```
manuelleduc, line 53: Wrong version globally, must be 14.2

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xwiki.hbm.xml
```diff
@@ -156,8 +156,12 @@
             <key-property name="link" type="string">
                 <column name="XWL_LINK" index="XWLNK_LINK" length="766" />
             </key-property>
+            <key-property name="type" type="string">
+                <column name="XWL_TYPE" default="'document'" length="768" />
```
tmortagne, line 160: Did you try on MySQL ? AFAIK the whole composite-id must be lower than the max KEY size (that's why "link" is 766, because it shares the key with docId)

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
tmortagne: Note that this is only with HSQLDB and the reason is that HSQLDB report a warning when you do a useless delete (which IMO is not something that worth a WARNING). Those useless deletes come from the fact that saveLinks starts by deleting all existing links for these documents and then insert the current ones.

IMO, this is a mistake on HSQLDB side since there is no reason to complain about this.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xwiki.hbm.xml
manuelleduc: Does that mean that the `composite-id` size is already maxed out?

manuelleduc: The `type` column does not need to be that large, less that 20 chars should be enough but if we don't have 0 size available for the `type` that won't change anything.

tmortagne: > One solution could be to replace the existing composite key (which is apparently not really great performance wide if I trust [stackoverflow.com/questions/8746207/1071-specified-key-was-too-long-max-key-length-is-1000-bytes](https://stackoverflow.com/questions/8746207/1071-specified-key-was-too-long-max-key-length-is-1000-bytes)), with a generated id

That's pretty much what we do in the document table (but with the ID being a hash, which is a bit dangerous, so better choose a safer id generation strategy).

manuelleduc: I've tested on maria, mysql and postgres with the table definition below and on my first tests it seems to be working well. Still need to check what's generated for the existing rows (which is not so important since we recreate all the entries during the migration). And to add a unique constraint on the former composite-id.
```
<class name="com.xpn.xwiki.doc.XWikiLink" table="xwikilinks">
  <id name="id" type="long" unsaved-value="undefined">
  <column name="XWL_ID" not-null="true"/>
  <generator class="assigned"/>
  </id>
  <property name="docId" column="XWL_DOC_ID" type="long"/>
  <property name="link" type="string" column="XWL_LINK" length="766"/>
  <property name="type" type="string" column="XWL_TYPE" length="100" />
  <property name="fullName" type="string" column="XWL_FULLNAME" length="768"/>
  <property name="attachmentName" type="string" column="XWL_ATTACHMENT_NAME" length="768"/>
</class>
```

tmortagne: >   <property name="link" type="string" column="XWL_LINK" length="766"/>

Should be 768, the only reason why it was 766 was because of the composite-id.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R14010XWIKI19352DataMigration.java
surli: Must be 14.2 and also it must follow the format documented there: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/DataMigration.java#L26

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+    private Provider<XWikiContext> xcontextProvider;
+
+    @Inject
+    private DocumentRevisionProvider documentRevisionProvider;
+
+    @Override
+    public void consume(DocumentReference documentReference, String version) throws IndexException
+    {
+        XWikiContext context = this.xcontextProvider.get();
+        // Skip if backlinks are not supported. Note that this task is not supposed to be queued when backlinks are
```
surli, line 64: I guess you added the if to bulletproof then, wondering if you shouldn't put a logger in a else branch then.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+        // not supported.
+        if (context.getWiki().hasBacklinks(context)) {
+            try {
+                XWikiDocument doc = this.documentRevisionProvider.getRevision(documentReference, version);
```
surli, line 68: Since it's asynchronous, it means that in somecases the version of the document might have been deleted, in which case this API returns null if I'm correct. You probably want to handle this.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R14010XWIKI19352DataMigration.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Override
+    public void migrate() throws DataMigrationException
+    {
+        XWikiContext context = this.contextProvider.get();
+
+        try {
+            for (String wikiId : this.wikiDescriptorManager.getAllIds()) {
+                context.setWikiId(wikiId);
```
surli, line 98: Hmmm, @tmortagne would know better than me, but if I'm correct the migration code is triggered independently on each wiki: you shouldn't need to loop yourself on the DB.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
```diff
@@ -90,39 +131,79 @@ public void setFullName(String fullName)
...
-        XWikiLink o = (XWikiLink)obj;
-
-        EqualsBuilder builder = new EqualsBuilder();
-        builder.append(getDocId(), o.getDocId());
-        builder.append(getLink(), o.getLink());
-        builder.append(getFullName(), o.getFullName());
-
-        return builder.isEquals();
+        XWikiLink xWikiLink = (XWikiLink) o;
+        return this.docId == xWikiLink.docId
```
surli, line 185: any reason to not use an equal builder? AFAIR it's part of our best practice.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
```diff
@@ -90,39 +131,79 @@ public void setFullName(String fullName)
...
+            && this.fullName.equals(xWikiLink.fullName)
+            && Objects.equals(this.type, xWikiLink.type)
+            && Objects.equals(this.attachmentName, xWikiLink.attachmentName);
     }
 
     @Override
     public int hashCode()
     {
-        return ("" + getDocId() + this.link).hashCode();
+        return Objects.hash(this.docId, this.link, this.fullName, this.type, this.attachmentName);
```
surli, line 195: Same question for hashcode builder

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R14010XWIKI19352DataMigration.java
tmortagne: Yes, each wiki has its own migration process, so a migration is supposed to only deal with the database.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultLinksTaskConsumer.java
tmortagne: Yes, it could be null in some cases so better be safe.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/listener/LinksUpdateListener.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public LinksUpdateListener()
+    {
+        super("LinksUpdateListener", new DocumentCreatedEvent(), new DocumentUpdatedEvent());
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        XWikiContext context = this.contextProvider.get();
```
tmortagne, line 71: With those events you actually receive the XWikiContext as `data` so we usually cast it instead of using a provider.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
```diff
@@ -33,21 +36,59 @@
...
+    /**
+     * Initializes the docId as well as the link (the used entity), the fullName (the reference of the document using
+     * the entity), and the type of the linked entity.
+     *
+     * @param docId the docId of the document using the entity
+     * @param link the link of the used entity
+     * @param fullName the full name of the document using the entity
+     * @param type the type of the used entity
+     */
+    public XWikiLink(long docId, String link, String fullName, String type)
```
tmortagne, line 86: Missing `@since`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xwiki.hbm.xml
manuelleduc: Does that mean that we can update the link field size to 768 for some db engines?

tmortagne: Yes, my point is that if we get rid of the composite-id constraint, we should definitely migrate to 768.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -5576,8 +5579,8 @@ private void getUniqueLinkedEntityReferences(XDOM dom, EntityType entityType, Se
      */
     public Set<String> getUniqueLinkedPages(XWikiContext context)
     {
-        Set<DocumentReference> references = toDocumentReferenceSet(
-            getUniqueLinkedEntityReferences(context, EntityType.DOCUMENT), getDocumentReference());
+        Set<EntityReference> references = getUniqueLinkedEntityReferences(context, Map.of(EntityType.DOCUMENT,
+            Set.of(ResourceType.SPACE, ResourceType.DOCUMENT, ResourceType.ATTACHMENT, ResourceType.PAGE)));
```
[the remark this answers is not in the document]
manuelleduc: I was planning of systematically run a refactoring on the document holding the attachment on move, instead of keeping self-references in the db.

PR #1790, opened 2022-03-04
XWIKI-19422: Attachment history is lost after rename

manuelleduc:
Duplicate the history when moving the attachment.

https://jira.xwiki.org/browse/XWIKI-19422


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiAttachment.java
```diff
@@ -248,6 +248,25 @@ public XWikiAttachment clone()
...
+     * @param doc the document to attach the clone to
+     * @param context the current context
+     * @return the clone attachment with its updated history
+     * @throws XWikiException in case of error when accessing the attachments content
+     * @throws IOException in case of error when writing the attachments content
+     */
+    public XWikiAttachment clone(String name, XWikiDocument doc, XWikiContext context)
+        throws XWikiException, IOException
+    {
+        XWikiAttachment clone = doc.setAttachment(name, this.getContentInputStream(context), context);
```
tmortagne, line 264: You are loosing a lot of info (mime type, date, author, etc.).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiAttachment.java
```diff
@@ -248,6 +248,25 @@ public XWikiAttachment clone()
...
+     * @return the clone attachment with its updated history
+     * @throws XWikiException in case of error when accessing the attachments content
+     * @throws IOException in case of error when writing the attachments content
+     */
+    public XWikiAttachment clone(String name, XWikiContext context)
+        throws XWikiException, IOException
+    {
+        XWikiAttachment clone = clone();
+        clone.setFilename(name);
+        clone.setContent(this.getContentInputStream(context));
```
manuelleduc, line 265: @tmortagne I think this is required here (but also at https://github.com/xwiki/xwiki-platform/pull/1790/files#diff-d8083a97ce5d1b35703898de56f16ee9deb0724c6aca732a567a769e20127e46R116) because the current implementation on the `clone()` methods does not actually copy the files.
For the double clone of the history, one solution could be to add a `private` `clone()`  method with some parameter(s) to skip the parts that will be overridden.
WDTY?

tmortagne: > the current implementation on the clone() methods does not actually copy the files

It does not really do need as far as I can see since the content is "immutable" in XWikiAttachmentContent (if you try to update the content it will create a new FileItem instance and not modify the target file).

manuelleduc: I tried to understand why the move is not working without `clone.setContent(this.getContentInputStream(context));` but this is apparently beyond my reach.
The best I could to is to define an internal clone with parameters to explicitly skip the clone of the archive and content, and keep the `setContent` call.

PR #1792, opened 2022-03-07

manuelleduc:
Split the migration in two distinct steps:
- remove the xwikilink table in oldcore
- repopulate it with up to date entries in index-api

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/hibernate/R140200000XWIKI19352DataMigration.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/hibernate/R140200000XWIKI19352DataMigration.java
@@ -1,0 +1,79 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/hibernate/R140200000XWIKI19352DataMigration.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public XWikiDBVersion getVersion()
+    {
+        return new XWikiDBVersion(140200000);
+    }
+
+    @Override
+    public String getPreHibernateLiquibaseChangeLog() throws DataMigrationException
+    {
+        return "<changeSet author=\"xwikiorg\" id=\"" + HINT + "\">\n"
```
tmortagne, line 71: Would be safer to check if the table actually exist before doing that (migrations are supposed to be written as much as possible to support re-execution).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/hibernate/R140200000XWIKI19352DataMigration.java
```diff
@@ -0,0 +1,79 @@
...
+            + "  <preConditions onFail=\"CONTINUE\">\n"
```
tmortagne, line 72: I'm not sure "CONTINUE" is the right one here, from what I understand of https://docs.liquibase.com/concepts/changelogs/preconditions.html. I'm a bit worried of  "Execution of the changeset will be attempted again on the next update". Wouldn't `MARK_RAN` make more sense ? Sure nothing was done in the end, but we migrate is finished, there is  no reason to do it again.

PR #1794, opened 2022-03-09
[Misc] Use entities for the wiki macros inserted with the wiki toolbar

surli:


surli:
Close after discussing with @mflorea . More context on this on https://jira.xwiki.org/browse/XWIKI-19503

PR #1797, opened 2022-03-11
XWIKI-18759: Replace the Livetable from the Attachments page with a Live Data

michitux:
* Migrate the all attachments page to Live Data
* TableLayoutElement: fix filtering by empty value
* TableLayoutElement: Add method for sorting by a column
* Add page test for the attachments results
* Move integration test for the attachments to docker

Jira issue: https://jira.xwiki.org/browse/XWIKI-18759

Before:

XWIKI-18759-AllAttachments-with-LT

After:

XWIKI-18759-AllAttachments-with-LD


vmassol:
Cool, thanks @michitux 

It would also be nice to convert the Author column to use user avatar as we do in other LT/LD. I know this is not really the purpose of this PR so that could be done in a different one. OTOH since you also improved other things, it could be part of it too.

vmassol:
Idea: It could be nicer visually if we were able to have a single cell spanning all the columns when there are permission issues with a small sentence directly in that cell. WDYT? Maybe something for the forum to discuss it.

michitux:
I've added user avatars, it now looks as follows:

XWIKI-18759-AllAttachments-with-LD-and-avatars

Similar to https://github.com/xwiki/xwiki-platform/blob/a364a0787d62bec0a218030bbe2d568ef0524505/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getgroupmembers.vm#L45-L46 I've also enabled displaying the user alias for advanced users and admins:

XWIKI-18759-AllAttachments-with-LD-and-avatars-advanced

I think this makes sense as the filter also filters by alias and not by the displayed name.

Regarding the "N/A"-cells this is the default behavior of Live Data and unrelated to this pull request so I think we should open a separate Jira issue for this.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/AllAttachmentsResults.xml
```diff
@@ -67,21 +67,24 @@
...
   #set ($fieldValue = $request.getParameter($fieldName))
+  #set ($matchType = $request.getParameter("${fieldName}_match"))
   #if ("$!fieldValue" != '')
-    #set ($range = $fieldValue.split('-'))
-    #set ($range = [$numbertool.toNumber($range.get(0)).intValue(), $numbertool.toNumber($range.get(1)).intValue()])
+    #if ($matchType != 'empty')
+      #set ($range = $fieldValue.split('-'))
+      #set ($range = [$numbertool.toNumber($range.get(0)).intValue(), $numbertool.toNumber($range.get(1)).intValue()])
+    #else
+      #set ($range = [-2,-1])
```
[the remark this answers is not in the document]
manuelleduc: I'm not 100% sure anymore what was the intent here. This filter is only applied on file size, which are not supposed to be empty.
It's probably a (draft) way I used to create an always false condition, but it is probably something that deserves to be improved.

michitux:
I've rebased the changes and updated the since-versions. Fewer changes to actual production code are necessary now as the attachments tab already uses Live Data. While testing, I discovered that the before/after filters on the date column don't work and opened [XWIKI-21314](https://jira.xwiki.org/browse/XWIKI-21314). This is the same as for the attachments tab and a new feature with Live Data, so I'm not sure whether it's a blocker.

The most controversial change is most likely that I moved `attachment_macros.vm` from `xwiki-platform-flamingo-skin-resources` to `xwiki-platform-web-templates` as otherwise it cannot really be used in `xwiki-platform-index-ui`. The reason for this is that `xwiki-platform-flamingo-skin-resources` depends on `xwiki-platform-index-ui` so introducing a dependency in the other direction creates a cycle. This cycle already existed in theory just not in an actual dependency as only the new page test requires that the template is actually present in the dependencies.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/AllAttachmentsResults.xml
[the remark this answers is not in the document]
michitux: This is no longer part of the pull request as this code has already been merged as part of introducing Live Data in the attachments tab. The merged code there doesn't contain this `#else` branch so with an `empty` simply no filtering happens. Not ideal imho but this pull request also no longer touches this code.

tmortagne:
> xwiki-platform-flamingo-skin-resources depends on xwiki-platform-index-ui

I don't have any judgment regarding the move, but that dependency does not make sense to me. That module should not depend on any XAR extension as, by definition, the XAR cannot be part of the WAR which means the template must be able to work without whatever this XAR provides.

michitux:
I just checked and apparently this dependency has been introduced [here](https://github.com/xwiki/xwiki-platform/commit/08190dd323118e16d833465ba343a4f5732ef9c1#diff-d167e7190750e9e9ccbbfd45455bf4f620b7f29114a181164399d7c1c4c00c27) as part of adding LiveTable to the attachments tab. The rationale is - I think - that it needs an XClass defined in `XWiki.AllAttachments` for the column metadata but not sure. Maybe @oanalavinia can provide more insights.

If we can reverse the dependency (not sure that makes sense either) I also don't need to move the template I think.

oanalavinia:
> The rationale is - I think - that it needs an XClass defined in `XWiki.AllAttachments` for the column metadata but not sure. Maybe @oanalavinia can provide more insights.

I remember I added the dependency for using `XWiki.AllAttachments`. But I think that was before moving to LiveData, since it was needed in [here](https://github.com/xwiki/xwiki-platform/commit/08190dd323118e16d833465ba343a4f5732ef9c1#diff-2394f2f345c0d87650a7e7bc8ee65cff902ccad1725be8d212b103a5eb83404eR261) .
Now I think this dependency can be removed. `XWiki.AllAttachments` is only used for the [className in the liveData](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachment_macros.vm#L269), but I am sure it has any utility with the `vm` defined

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/AllAttachments.xml
```diff
@@ -38,28 +38,35 @@
...
+      {'id': 'view', 'allowProperty': 'acclev'}
+    ]
+  }
 })
-#livetable('allattachments' $columns $columnsProperties $options)
-{{/velocity}}
 
-(% id="inaccessibleDocsMessage" class="hidden" %)(((
-  {{info}}{{translation key="rightsmanager.documentrequireviewrights" /}}{{/info}}
-)))</content>
```
manuelleduc, line 62: Is it expected that this piece of code, along with the corresponding javascript, are moved?

michitux: Yes, Live Data has this integrated. There is a screenshot demonstrating this in this pull request.

manuelleduc: indeed :) All good then! Thanks

manuelleduc (approved):

PR #1798, opened 2022-03-14
XWIKI-19512: Deleted revisions of documents cannot be displayed anymore

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-19512

**WIP**

- PR opened mainly to discuss the naming of the new `DocExistValidator` role and the new components (`ViewDocExistValidator`/`GetDocExistValidator`), as well at the new implementations choices.
- I'm currently writing the test


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -731,45 +732,30 @@ public void execute(XWikiContext context) throws Exception
...
-                        // there is an error we consider that the sheet doesn't exist.
-                        result = true;
-                    }
-                } else {
-                    result = !"1".equals(request.get("disableCheckNotExisting"));
-                }
+        String action = context.getAction();
+        try {
+            DocExistValidator instance = this.componentManager.getInstance(DocExistValidator.class, action);
+            if (instance != null) {
```
surli, line 750: How the instance can be null? If the component is not found it triggers an exception no?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -731,45 +732,30 @@ public void execute(XWikiContext context) throws Exception
...
+                result = instance.docExist(doc, context);
             }
+        } catch (ComponentLookupException e) {
+            // It's not required to have a validator for each action, when not specific component is provided for 
+            // an action, the default false value is returned.
```
surli, line 755: Not sure about this fallback, I'd bulletproof a bit by checking if we're not in get/view and letting the exception propagate in such case.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/web/ViewDocExistValidator.java
```diff
@@ -0,0 +1,97 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xpn.xwiki.internal.web;
```
surli, line 20: New components should not go in that package. It would be cleaner to put them in an `org.xwiki.` package.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/DeletePageOutcomePage.java
```diff
@@ -125,5 +125,22 @@ public DeletePageOutcomePage clickDeleteTerminalPage()
...
+     * @param row the revision row for which to view the deleted page, starting at 1
+     * @return the view page of the deleted page
+     * @since 13.10.4
+     * @since 14.2RC1
+     */
+    public ViewPage clickViewDocument(int row)
+    {
+        WebElement link = getDriver()
+            .findElementWithoutWaiting(
+                By.cssSelector(String.format("table tbody tr:nth-child(%d) > td:nth-child(2) > a", row)));
```
surli, line 142: Not a big fan of such selector... Easy to break if we ever change the UI of the deleted pages. Isn't it possible to introduce an actual CSS class to semantically qualify such link?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
manuelleduc: Why would we hard-code get/view specifically?

surli: I was mentioning those since that's the one for which we do provide components implem in XS: we know that those instances should exist, so if we don't find them it's really an error.

manuelleduc: ok, make sense, thanks.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -731,45 +732,30 @@ public void execute(XWikiContext context) throws Exception
...
-                            ExceptionUtils.getRootCauseMessage(e));
-                        // there is an error we consider that the sheet doesn't exist.
-                        result = true;
-                    }
-                } else {
-                    result = !"1".equals(request.get("disableCheckNotExisting"));
-                }
+        String action = context.getAction();
+        try {
+            DocExistValidator instance = this.componentManager.getInstance(DocExistValidator.class, action);
```
tmortagne, line 749: Throwing exceptions is slow and for all actions other than "view" and "get" you will get an exception so would be better to first call this.componentManager.hasComponent to see if there is actually a component available for this action.

surli: > This also allows differentiating between "there is no DocExistValidator for this action" and "there is one, but it's crashing, so I should be a bit stricter than a debug log".

Note that I also wondered about that and I double checked the interface documentation of ComponentManager it only says that `ComponentLookupException` is thrown if the component is not found. Not when the component fails to initialize. The doc should probably be improved on this aspect.

tmortagne: > The doc should probably be improved on this aspect.

Indeed, that javadoc is wrong. There is an exception if anything prevent to get the component instance.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java
```diff
@@ -743,33 +744,16 @@ public void execute(XWikiContext context) throws Exception
...
-                } else {
-                    result = !"1".equals(request.get("disableCheckNotExisting"));
-                }
+        String action = context.getAction();
+        if (this.componentManager.hasComponent(DocExistValidator.class, action)) {
+            try {
+                result = this.componentManager.<DocExistValidator>getInstance(DocExistValidator.class, action)
+                    .docExist(doc, context);
+            } catch (ComponentLookupException e) {
+                // Should never happen as we check for the component existence before calling getInstance.
```
tmortagne, line 753: As I indicated in my previous comment, this would mean that this component could not be initialized, which can definitely happen and which should not be ignored. The minimum is an error log.

PR #1800, opened 2022-03-18
XWIKI-15776: The groups displayer store values with a coma at the end

surli:
* Refactor ListClass#fromList to take into account multiselect
  * Provide a new ListClass#fromList API with a parameter to filter out
    empty values
  * Provide a new ListClass#getListFromString API with a parameter to
    filter out empty values
  * Use those new APIS in UsersClass/GroupsClass/LevelsClass
  * Add some new tests

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
@@ -27,6 +27,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.ecs.xhtml.input;
@@ -325,6 +326,22 @@ public abstract class ListClass extends PropertyClass
      */
     public static List<String> getListFromString(String value, String separators, boolean withMap)
     {
+        return ListClass.getListFromString(value, separators, withMap, false);
+    }
+
+    /**
+     * @param value the string holding a serialized list
+     * @param separators the separator characters (given as a string) used to delimit the list's items inside the input
+     *            string. These separators can also be present, in escaped ({@value #SEPARATOR_ESCAPE}) form, inside
+     *            list items
+     * @param withMap set to true if the list's values contain map entries (key=value pairs) that should also be parsed.
+     *            Only the keys are extracted from such list items
+     * @param filterEmptyValues {@code true} if the result should not contain any empty values.
+     * @return the list that was stored in the input string
+     */
+    protected static List<String> getListFromString(String value, String separators, boolean withMap,
+        boolean filterEmptyValues)
+    {
         List<String> list = new ArrayList<>();
         if (StringUtils.isEmpty(value)) {
             return list;
@@ -371,7 +388,9 @@ public abstract class ListClass extends PropertyClass
                 // in case of two consecutive identical characters different than a whitespace, then it means
                 // we want to record an empty value.
                 if (!inEscape && currentChar == previousSeparator && !StringUtils.isWhitespace(currentChar + "")) {
-                    list.add("");
+                    if (!filterEmptyValues) {
+                        list.add("");
+                    }
                     previousWasSeparator = false;
                 }
                 previousSeparator = currentChar;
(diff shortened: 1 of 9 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/UsersClass.java
```diff
@@ -210,16 +212,6 @@ public BaseProperty newPropertyfromXML(Element ppcel)
         return selectlist;
     }
 
-    @Override
-    public void fromList(BaseProperty<?> property, List<String> list)
```
tmortagne, line 214: I must be missing something, but don't you want to filter empty values here ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
```diff
@@ -974,10 +1013,36 @@ private option createOption(String value, String display, List<String> selectlis
...
+
+    /**
+     * Set the passed {@link List} into the passed property.
+     *
+     * @param property the property to modify
+     * @param list the list to set
+     * @param filterEmptyValues if {@code true} filter out the empty values from the list.
+     * @since 14.2RC1
+     */
+    protected void fromList(BaseProperty<?> property, List<String> list, boolean filterEmptyValues)
```
tmortagne, line 1027: Maybe this new `filterEmptyValues` concept should be an actual configuration of the property class (that happen to be always true for group/user classes) instead of just a parameter in a protected method.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/UsersClass.java
surli: So in the context of the bug I fixed the call line 169: https://github.com/xwiki/xwiki-platform/pull/1800/files#diff-9db5305f72027b429d4ade2c90c36c1b2b49151d72b07281092732a02f01bc4eR169

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/ListClass.java
surli: Could be a global improvment to make indeed. Feels a bit too much for this PR though.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/UsersClass.java
```diff
@@ -158,13 +165,8 @@ public BaseProperty fromString(String value)
     public BaseProperty fromStringArray(String[] strings)
     {
         List<String> list = new ArrayList<>();
-        for (int i = 0; i < strings.length; i++) {
-            if (!StringUtils.isBlank(strings[i])) {
-                list.add(strings[i]);
-            }
-        }
```
surli, line 165: Wow this is wrong actually... fixing and adding a test to this.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/UsersClass.java
tmortagne: I don't understand, `fromList(BaseProperty<?> property, List<String> list)` does not seem to be reusing  `fromStringArray(String[] strings)`.

surli: That's the contrary: fromStringArray should have used fromList, as in GroupsClass#fromStringArray. So I performed same change in both classes.

tmortagne: My point here is that if someone calls `UsersClass#fromList(myproperty, mylist)`, the empty element from `mylist` won't be filtered because that's the default behavior of `ListClass#fromList(BaseProperty<?> property, List<String> list)` from what I understand.

surli: > My point here is that if someone calls UsersClass#fromList(myproperty, mylist), the empty elements from mylist won't be filtered because that's the default behavior of ListClass#fromList(BaseProperty<?> property, List<String> list) from what I understand.

Yes that's right, I kept same behaviour as before for this call. I'm fixing only the call from #fromStringArray.

tmortagne: But that's wrong. We want to filter empty values in general in GroupClass and UserClass since they don't make sense, it does not make sense to do it only for String arrays and not String lists.

surli: ok I'll put back the override with a filtering of empty values by default then.

PR #1801, opened 2022-03-18
XWIKI-19535: Heading ids are not necessarily unique when using the include or display macro

michitux:
* Adapt heading and image ids when including or displaying pages.

Jira issue: https://jira.xwiki.org/browse/XWIKI-19535


manuelleduc (approved):

michitux:
After feedback from @tmortagne I'll open a new PR in a moment.

PR #1802, opened 2022-03-18

manuelleduc:
Use a listener instead of the data migration feature to re-index the wiki as this is not integrating well with the way we package the jetty-hsqldb distribution.


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/listener/LinksIndexationEventListener.java
```diff
@@ -0,0 +1,141 @@
...
+    public List<Event> getEvents()
+    {
+        return EVENTS;
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        try {
+            XWikiContext context = this.contextProvider.get();
```
manuelleduc, line 100: Note: replace this with a cast of `data`.

PR #1805, opened 2022-03-21

manuelleduc:
Wait for the document analysis tasks to be finished before continuing.

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-data/pom.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-data/pom.xml
@@ -161,6 +161,12 @@
               <artifactId>xwiki-platform-ratings-api</artifactId>
               <version>${project.version}</version>
             </dependency>
+            <!-- Used to process the document analysis tasks (e.g., backlinks indexing). -->
+            <dependency>
+              <groupId>org.xwiki.platform</groupId>
+              <artifactId>xwiki-platform-index-api</artifactId>
+              <version>${project.version}</version>
+            </dependency>
           </dependencies>
           <executions>
             <execution>
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-data/pom.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-data/pom.xml
@@ -161,12 +161,6 @@
               <artifactId>xwiki-platform-ratings-api</artifactId>
               <version>${project.version}</version>
             </dependency>
-            <!-- Used to process the document analysis tasks (e.g., backlinks indexing). -->
-            <dependency>
-              <groupId>org.xwiki.platform</groupId>
-              <artifactId>xwiki-platform-index-api</artifactId>
-              <version>${project.version}</version>
-            </dependency>
           </dependencies>
           <executions>
             <execution>
--- a/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/com/xpn/xwiki/tool/backup/DataMojo.java
+++ b/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/com/xpn/xwiki/tool/backup/DataMojo.java
@@ -93,7 +93,7 @@ public class DataMojo extends AbstractOldCoreMojo
         ComponentManager componentManager =
             (ComponentManager) this.oldCoreHelper.getXWikiContext().get(ComponentManager.class.getName());
         try {
-            // Manually start the consumer threads as it is not automatically started in the packaging setup.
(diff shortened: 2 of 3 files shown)


xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-data/pom.xml
```diff
@@ -161,6 +161,12 @@
               <artifactId>xwiki-platform-ratings-api</artifactId>
               <version>${project.version}</version>
             </dependency>
+            <!-- Used to process the document analysis tasks (e.g., backlinks indexing). -->
+            <dependency>
```
tmortagne, line 165: Not really needed here since it's already a dependency of the plugin.

manuelleduc: Indeed, removing it, thanks.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/com/xpn/xwiki/tool/backup/DataMojo.java
```diff
@@ -71,4 +75,40 @@ public void executeInternal() throws MojoExecutionException
...
+    private void consumeIndexingQueue() throws MojoExecutionException
+    {
+        ComponentManager componentManager =
+            (ComponentManager) this.oldCoreHelper.getXWikiContext().get(ComponentManager.class.getName());
+        try {
+            // Manually start the consumer threads as it is not automatically started in the packaging setup.
+            // This is also convenient as we can be sure that the tasks are all queued before being consumed.
+            TaskManager taskManager = componentManager.getInstance(TaskManager.class);
+            if (taskManager instanceof DefaultTasksManager) {
+                ((DefaultTasksManager) taskManager).startThread();
```
tmortagne, line 100: Why is this needed ? I see that this is done automatically in TaskApplicationReadyListener when receiving ApplicationReadyEvent and ApplicationReadyEvent seems to be sent in OldCoreHelper#create.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/com/xpn/xwiki/tool/backup/DataMojo.java
```diff
@@ -71,4 +75,40 @@ public void executeInternal() throws MojoExecutionException
...
+            }
+
+            getLog().info("Waiting for the document indexing queue to be empty...");
```
tmortagne, line 103: Maybe indicate how big is the queue before we start to wait for it to empty (to give a clue of how long it might take).

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/com/xpn/xwiki/tool/backup/DataMojo.java
manuelleduc: `org.xwiki.index.internal.TaskApplicationReadyListener#initialize` is called but the context provider returns null. 
`org.xwiki.index.internal.TaskApplicationReadyListener#onEvent` is never called, which not be right.

tmortagne: > org.xwiki.index.internal.TaskApplicationReadyListener#initialize is called but the context provider returns null.

This is expected yes, same as in a standard XWiki instance.

>  org.xwiki.index.internal.TaskApplicationReadyListener#onEvent is never called, which not be right.

That's really not expected according to https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/utils/OldCoreHelper.java#L152.

manuelleduc: Alright, debugging.

tmortagne: > That's really not expected according to [https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/utils/OldCoreHelper.java#L152](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/utils/OldCoreHelper.java?rgh-link-date=2022-03-21T15%3A40%3A31Z#L152).

Actually this is ApplicationStartedEvent sorry and indeed we don't seems to send ApplicationReadyEvent.

PR #1806, opened 2022-03-22
XWIKI-19463: Display image ids in the Lightbox

oanalavinia:
* extracted the image id, prioritizing the manually added one, and display it as part of the lightbox description
* provide a button to copy the displayed image id to clipboard

imageIDFull

### DIFF
--- a/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java
+++ b/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java
@@ -100,6 +100,20 @@ class LightboxIT
         // Make sure that the images are displayed.
         lightboxPage.reloadPage();
 
+        // Verify the image popover permalink action.
+        Optional<ImagePopover> imagePopover = lightboxPage.hoverImage(0);
+        if (imagePopover.isPresent()) {
+            ImagePopover currentImagePopover = imagePopover.get();
+            assertTrue(currentImagePopover.isImagePopoverDisplayed());
+
+            WebElement imagePermalinkButton = currentImagePopover.getImagePermalinkButton();
+            String[] elements = imagePermalinkButton.getAttribute("href").split("#");
+            assertEquals("Iimage1.png", elements[elements.length - 1]);
+
+            imagePermalinkButton.click();
+            assertEquals(testUtils.getDriver().getCurrentUrl(), imagePermalinkButton.getAttribute("href"));
+        }
+
         Lightbox lightbox = lightboxPage.openLightboxAtImage(0);
         assertTrue(lightbox.isDisplayed());
 
@@ -126,6 +140,20 @@ class LightboxIT
 
         // Make sure that the images are displayed.
         lightboxPage.reloadPage();
+
+        // Verify the image popover permalink action.
+        Optional<ImagePopover> imagePopover = lightboxPage.hoverImage(0);
+        if (imagePopover.isPresent()) {
+            ImagePopover currentImagePopover = imagePopover.get();
+            assertTrue(currentImagePopover.isImagePopoverDisplayed());
+
+            WebElement imagePermalinkButton = currentImagePopover.getImagePermalinkButton();
+            String[] elements = imagePermalinkButton.getAttribute("href").split("#");
+            assertEquals("manuallyAddedImageId", elements[elements.length - 1]);
+
+            imagePermalinkButton.click();
+            assertEquals(testUtils.getDriver().getCurrentUrl(), imagePermalinkButton.getAttribute("href"));
+        }
 
         Lightbox lightbox = lightboxPage.openLightboxAtImage(0);
         assertTrue(lightbox.isDisplayed());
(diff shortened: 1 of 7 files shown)


vmassol:
Thanks @oanalavinia ! 

This id is somewhat technical and not going to be used that often. Thus I feel that the current way to display it is a bit too visible and proeminent. Is it possible to find a way to display it in a less visible manner? WDYT?

Thanks

michitux:
For me this looks good apart from one point which is that there can be images without ids. Currently, this is for example the case when you use the [box macro](https://extensions.xwiki.org/xwiki/bin/view/Extension/Box%20Macro) with the image attribute. I think the id should only be displayed when it is non-empty.

oanalavinia:
To avoid displaying technical data we could have an action button, on the top left actions group, that would simply copy the id to clipboard
image

vmassol:
> To avoid displaying technical data we could have an action button, on the top left actions group, that would simply copy the id to clipboard
> image

I like it! Thx

mflorea:
BTW, having this in the lightbox actions is fine but I don't think the users should have to open the lightbox to be able to copy the image id. It should (also) be direcly in the image popover. Most of the time I would just hover the image and click on the copy anchor button. And if we do this then I think we should also add the "permalink" button to be able to share externally the link to a given image occurrence (which should be possible for headings also but that's a different topic). So to sum up, I think both the image popover and the lightbox should expose additionally these two actions:

* https://fontawesome.com/v4/icon/link that when clicked simply changes the hash of the current URL to target the image id (i.e. it simply acts as a link to ``#imageId``). This is useful for sharing the link externally
* https://fontawesome.com/v4/icon/anchor that when clicked copies the image id to clipboard. This is useful for creating internal links

vmassol:
> Most of the time I would just hover the image and click on the copy anchor button. 

Sounds good to me. If we do this, I'm not even sure we need the info inside the lightbox.

> And if we do this then I think we should also add the "permalink" button to be able to share externally the link to a given image occurrence (which should be possible for headings also but that's a different topic)

+1

oanalavinia:
Actions group:
newToolbar
Image popover:
imagePopover

mflorea:
@oanalavinia I see you added "copy image id" only to the lightbox. Why not expose it on the image popover also? For the permalink I agree that it could stay only on the image popover, because on the lightbox it has a different meaning (link to the lightbox opened for that image, which is more work to do and hasn't been asked so far). So to be clear I prefer:

* image popover: have both "copy image id" and permalink
* lightbox: have only "copy image id"

While you're at it, I think we need to add a bit of spacing between the actions of the image popover. They look too cramped right now.

oanalavinia:
New toolbar
image

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java
```diff
@@ -100,6 +100,21 @@ void openImageWithoutDescription(TestUtils testUtils, TestReference testReferenc
         // Make sure that the images are displayed.
         lightboxPage.reloadPage();
 
+        // Verify the image ID popover actions.
+        Optional<ImagePopover> imagePopover = lightboxPage.hoverImage(0);
+        if (imagePopover.isPresent()) {
```
mflorea, line 105: Isn't the popover expected? If the purpose of the test is to check the popover actions then the test should ensure that the popover is visible. If it's not then the test should fail.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java
```diff
@@ -126,6 +142,21 @@ void openImageWithCaption(TestUtils testUtils, TestReference testReference, Test
...
```
mflorea, line 147: Same as above, I don't see the need for this ``IF``.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java
```diff
@@ -360,6 +394,34 @@ void verifyPartiallyDisabledLightbox(TestUtils testUtils, TestReference testRefe
...
+        testUtils.createPage(testReference, this.getImageWithoutId(images.get(0)));
+        lightboxPage = new LightboxPage();
+
+        lightboxPage.attachFile(testConfiguration.getBrowser().getTestResourcesPath(), images.get(0));
+
+        // Make sure that the images are displayed.
+        lightboxPage.reloadPage();
+
+        Optional<ImagePopover> imagePopover = lightboxPage.hoverImage(0);
+        if (imagePopover.isPresent()) {
```
mflorea, line 413: Same.

mflorea (commented):
Looks good, besides my comments for the tests.

mflorea (approved):

PR #1809, opened 2022-03-29
XWIKI-19581: Refactor XWIKIDOCUMENTINDEXINGQUEUE to allow null values in the version

manuelleduc:
- Migrate the schema of xwikidocumentindexingqueue to use a native generated id instead of a composite id
- Now that XWT_VERSION is not part of the id, it can be made nullable, allowing Oracle to accept the empty string

https://jira.xwiki.org/browse/XWIKI-19581

If `xwikidocumentindexingqueue` is empty, the table is simply dropped and Hibernate recreates it.
If `xwikidocumentindexingqueue` is not empty, the following steps are executed:
1. `xwikidocumentindexingqueue` is  renamed `xwikidocumentindexingqueuetmp`
2. Hibernate recreates `xwikidocumentindexingqueue`
3. the content of `xwikidocumentindexingqueuetmp` is copied into `xwikidocumentindexingqueue`, and `xwikidocumentindexingqueuetmp` is dropped.


tmortagne:
> xwikidocumentindexingqueue is renamed xwikidocumentindexingqueuetmp

Wouldn't it be simpler to just load the entries in memory, recreated the table and reinsert them ?

manuelleduc:
> > xwikidocumentindexingqueue is renamed xwikidocumentindexingqueuetmp
> 
> Wouldn't it be simpler to just load the entries in memory, recreated the table and reinsert them ?

The would work too, it's just less scalable in case of a lot of rows in `xwikidocumentindexingqueue`. But in practice that shouldn't be the case.

tmortagne:
> it's just less scalable in case of a lot of rows in xwikidocumentindexingqueue

I thought about that, but aren't you going to load all those in the memory queue anyway ?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R140300001XWIKI19571DataMigration.java
```diff
@@ -0,0 +1,185 @@
...
+ * Finally, the temporary table is dropped.
+ *
+ * @version $Id$
+ * @since 14.3RC1
+ * @since 14.2.1
+ */
+@Component
+@Singleton
+@Named(R140300001XWIKI19571DataMigration.HINT)
+public class R140300001XWIKI19571DataMigration extends AbstractHibernateDataMigration
```
manuelleduc, line 61: While this is working correctly with hsqldb, I couldn't test it with Oracle yet. I'll update the corresponding test (and the coverage) once I'll manage to test this.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R140300001XWIKI19571DataMigration.java
```diff
@@ -0,0 +1,184 @@
...
+import org.xwiki.index.internal.TasksStore;
+
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.internal.store.hibernate.HibernateStore;
+import com.xpn.xwiki.store.migration.DataMigrationException;
+import com.xpn.xwiki.store.migration.XWikiDBVersion;
+import com.xpn.xwiki.store.migration.hibernate.AbstractHibernateDataMigration;
+
+/**
+ * Migrate XWikiDocumentIndexingTask by copying the table to a new temporary table. Then, after hibernate migration,
```
tmortagne, line 50: Is this still accurate ?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R140300001XWIKI19571DataMigration.java
```diff
@@ -0,0 +1,184 @@
...
+
+                PersistentClass persistentClass =
+                    this.hibernateStore.getConfigurationMetadata()
+                        .getEntityBinding(XWikiDocumentIndexingTask.class.getName());
+                String tableName = this.hibernateStore.getConfiguredTableName(persistentClass);
+                if (exists(databaseMetaData, tableName)) {
+                    saveTasks(connection, persistentClass, tableName);
+                }
+                return String.format("<changeSet author=\"xwikiorg\" id=\"%s0\">\n"
+                    + "  <preConditions onFail=\"MARK_RAN\">\n"
```
tmortagne, line 105: Since you are already checking if the table exist, I guess you could remove that precondition and instead return the drop only in the previous if ?

PR #1810, opened 2022-03-29
XWIKI-19571: ORA-01400: cannot insert NULL into ("XWIKI"."XWIKIDOCUMENTINDEXINGQUEUE"."XWT_VERSION") when creating an instance on Oracle

manuelleduc:
https://jira.xwiki.org/browse/XWIKI-19571

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultTasksManager.java
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultTasksManager.java
@@ -135,7 +135,7 @@ public class DefaultTasksManager implements TaskManager, Initializable, Disposab
         try {
             try {
                 this.tasksStore.get().addTask(wikiId, xWikiTask);
-            } catch (XWikiException e) {
+            } catch (Exception e) {
                 this.logger.warn(
                     "Failed to add a task for docId [{}], type [{}] and version [{}] in wiki [{}]. This task is queued"
                         + " but will not be will not be restarted if not completed before the server stops."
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/DefaultTasksManager.java
```diff
@@ -135,7 +135,7 @@
         try {
             try {
                 this.tasksStore.get().addTask(wikiId, xWikiTask);
-            } catch (XWikiException e) {
+            } catch (Exception e) {
```
manuelleduc, line 138: This single change should be enough to make Oracle pass (since the task will be added in database). The rest of the PR aims at preventing empty string for the version column by appending a `-` before the version, and removing it afterward.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/internal/TasksStore.java
```diff
@@ -48,6 +50,14 @@
 @Singleton
 public class TasksStore extends XWikiHibernateBaseStore
 {
+    /**
+     * A special character inserted at the beginning of all document versions, to make sure that the versions are never
+     * the empty string, otherwise Oracle fails to persist the task.
+     *
+     * @see <a href="https://jira.xwiki.org/browse/XWIKI-19571">XWIKI-19571</a></a>
+     */
+    private static final String FILLER = "-";
```
tmortagne, line 59: Maybe add a note that it's temporary and will be removed as part of XWIKI-19581 (even if you have a TODO about that below in the code).

PR #1811, opened 2022-03-30
[Misc] Ensure to be able to use Oracle image for docker test

surli:
* Set the missing instructions for allowing to use our custom image
  * Fix the username since TC does not allow to use sys/system user


xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/internal/junit5/database/DatabaseContainerExecutor.java
```diff
@@ -248,14 +249,17 @@ private void mountDatabaseDataIfNeeded(JdbcDatabaseContainer<?> databaseContaine
...
+        String oracleImageFullName;
         if (testConfiguration.getDatabaseTag() != null) {
-            databaseContainer = new OracleContainer(String.format("xwiki/oracle-database:%s",
-                testConfiguration.getDatabaseTag()));
+            oracleImageFullName = String.format("xwiki/oracle-database:%s", testConfiguration.getDatabaseTag());
         } else {
-            databaseContainer = new OracleContainer("xwiki/oracle-database");
+            oracleImageFullName = "xwiki/oracle-database";
         }
+        DockerImageName oracleImage = DockerImageName.parse(oracleImageFullName)
```
vmassol, line 258: I would add a comment here to explain why we want to use our own image and thus have to register it as a substitute

vmassol:
Sounds good!

PR #1815, opened 2022-04-05
XWIKI-18810: Save of the document shouldn't be performed each time an upload is done

surli:
* Provide a new role component for managing temporary attachments
  * Provide implementation based on a map of cache of attachments for
    handling different editions of documents for each user sessions
  * Use and clean the temporary attachments in various actions
  * Improve EditForm to read a list of temporary uploaded attachments,
    to only add those when using save action
  * Small refactoring of DownloadActionTest to remove the old
    MockitoDownloadActionTest
  * Provide new API in WysiwygEditorScriptService to provide an endpoint
    for uploading temporary attachments
  * Small changes in InplaceEditing to send the list of uploaded
    attachments (the inputs read there should be created by the WYSIWYG
editor implementation)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/EditForm.java
```diff
@@ -446,6 +450,20 @@ private void setObjectsToAdd(String[] objectsToAdd)
...
+    private void setTemporaryUploadedFiles(String[] temporaryUploadedFiles)
+    {
+        if (temporaryUploadedFiles == null || temporaryUploadedFiles.length == 0) {
+            this.temporaryUploadedFiles = Collections.emptyList();
+        } else {
+            this.temporaryUploadedFiles = Arrays.asList(temporaryUploadedFiles);
+        }
+    }
+
+    public List<String> getTemporaryUploadedFiles()
```
surli, line 462: Missing doc

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/DownloadActionTest.java
```diff
@@ -768,4 +774,137 @@ void downloadWhenEmptyRange() throws XWikiException, IOException
         verifyOutputExpectations(0, this.fileContent.length);
         verifyResponseExpectations(d.getTime(), this.fileContent.length);
     }
+
+    private static class StubServletOutputStream extends ServletOutputStream
```
manuelleduc, line 778: Out of curiosity, why a stub and not a mock?

surli: So here I just copy/pasted the tests that were in MockitoDownloadActionTest, I fixed some obvious stuff, but I didn't spent much time thinking about that one.

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentManager.java
```diff
@@ -0,0 +1,237 @@
...
+        String sessionId = httpSessionEvent.getSession().getId();
+        if (this.temporaryAttachmentSessionMap.containsKey(sessionId)) {
+            this.temporaryAttachmentSessionMap.get(sessionId).dispose();
+            this.temporaryAttachmentSessionMap.remove(sessionId);
+        }
+    }
+
+    private String getCacheKey(String sessionId, DocumentReference documentReference)
+    {
+        return String.format("%s_%s", sessionId, this.stringEntityReferenceSerializer.serialize(documentReference));
```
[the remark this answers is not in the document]
surli: Note that this is now only used for naming the cache, it's not used to find it.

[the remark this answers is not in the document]
surli: I should rename this method.

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentManager.java
```diff
@@ -0,0 +1,237 @@
...
+
+    private Cache<XWikiAttachment> getOrCreateCache(TemporaryAttachmentSession temporaryAttachmentSession,
+        DocumentReference documentReference) throws CacheException
+    {
+        Cache<XWikiAttachment> result;
+
+        if (!temporaryAttachmentSession.hasOpenEditionSession(documentReference)) {
+            String key = getCacheKey(temporaryAttachmentSession.getSessionId(), documentReference);
+            String configName = String.format("temp.attachment.%s", key);
+            // FIXME: the configuration values should be configurable
```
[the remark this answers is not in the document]
surli: > It is configurable through the infinispan configuration

the problem here is that each cache will have a specific name: is it possible to specify that all caches starting with `temp.attachment` will have a dedicated configuration? I was being adding new properties in xwiki.properties for allowing to set the maxEntries and idleTime default values for newly created cache here.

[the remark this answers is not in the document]
tmortagne: https://infinispan.org/docs/stable/titles/configuring/configuring.html#cache-wildcards-cache-configuration seems to suggest you can indeed indicate a wildcard as cache name.

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentManager.java
```diff
@@ -0,0 +1,261 @@
...
+
+    private Cache<XWikiAttachment> getOrCreateCache(TemporaryAttachmentSession temporaryAttachmentSession,
+        DocumentReference documentReference) throws CacheException
+    {
+        Cache<XWikiAttachment> result;
+
+        if (!temporaryAttachmentSession.hasOpenEditionSession(documentReference)) {
+            String cacheName = getCacheName(temporaryAttachmentSession.getSessionId(), documentReference);
+            result = this.cacheManager.createNewCache(
+                new LRUCacheConfiguration(cacheName, getCacheMaxEntries(), getCacheIdleTime()));
```
michitux, line 157: Why is this a cache with a maximum size? Do you need any caching behavior here that only files that are frequently accessed items are kept? This will cause absolutely weird bugs where users are randomly using temporary attachments if there are too many users uploading temporary attachments at the same time.

michitux: Okay, maybe I misunderstood something here, this is a cache per user and document? If yes, I'm also not sure this is the right approach as I think having such a cache has quite some overhead (but I might be mistaken).

surli: > Okay, maybe I misunderstood something here, this is a cache per user and document?

yes

>  If yes, I'm also not sure this is the right approach as I think having such a cache has quite some overhead (but I might be mistaken).

Yes, this is what we've been discussing with @tmortagne on https://github.com/xwiki/xwiki-platform/commit/1681aa5e88c3d6764be948c3ca5970f47dec0541#r70594779 I'll change that approach.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/store/TemporaryAttachmentManager.java
```diff
@@ -0,0 +1,94 @@
...
+ * Interface for operations related to temporary upload of attachments.
+ * The idea of this API is to allow obtaining directly a temporary {@link XWikiAttachment} from a {@link Part} and to
+ * keep it in cache until it's saved. The cache is managed per user session, and per edited document.
+ *
+ * @version $Id$
+ * @since 14.3RC1
+ */
+@Unstable
+@Role
+public interface TemporaryAttachmentManager
```
tmortagne, line 43: The interface name should maybe make more clear that it's related to the session, to not be missleading.

surli: I put a note in the javadoc for it, but I could add a bit more explanations probably. 
Now for renaming, do you have ideas? Something like `TemporaryAttachmentSessionManager`?

tmortagne: Yes I saw the javadoc, but I think it's not enough. `TemporaryAttachmentSessionManager` sounds good.

PR #1816, opened 2022-04-07
XWIKI-19614: The locale is ignored when indexing documents

manuelleduc:
- Use getDocumentReferenceWithLocale instead of getDocumentReference when:
  - passing the DocumentReference to the task consumers.
  - resolving the previous version of a documentation during mentions analysis
- Reindex the links and attachments they are missing for localized pages

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -146,7 +146,7 @@
                   <code>java.class.removed</code>
                   <old>class org.xwiki.index.migration.R140200001XWIKI19352DataMigration</old>
                   <justification>R140200001XWIKI19352DataMigration replaced by R140300000XWIKI19614DataMigration because a link reindexing is required by XWIKI-19614, making an additional reindexing for migrations from version lower than 14.2RC1 useless.</justification>
-                  <criticality>highlight</criticality>
+                  <criticality>allowed</criticality>
                 </item>
               </differences>
             </revapi.differences>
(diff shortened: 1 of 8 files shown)


xwiki-platform-core/pom.xml
```diff
@@ -138,7 +138,18 @@
...
-            
+
+            <revapi.differences>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.class.removed</code>
+                  <old>class org.xwiki.index.migration.R140200001XWIKI19352DataMigration</old>
+                  <justification>R140200001XWIKI19352DataMigration replaced by R140300000XWIKI19614DataMigration because a link reindexing is required by XWIKI-19614, making an additional reindexing for migrations from version lower than 14.2RC1 useless.</justification>
+                  <criticality>highlight</criticality>
```
tmortagne, line 149: "highlight" is for important breakage, which is not the case here, so it's supposed to be more "documented" AFAIK.

manuelleduc: If I read https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HRevapiIgnores correctly, `documented` in only for unstable code.

>Semantic changes that are potentially breaking in severity(like adding an annotation) but that, after analyzing we consider to not be a breakage.
>
>    Mark the ignore with <criticality>allowed</criticality>

I agree to change to `allowed`.

PR #1819, opened 2022-04-11
XWIKI-19626: Add extension point for inserting content before page header

slauriere:
* Update flamingo/contentview.vm to add UI extension point "org.xwiki.platform.template.content.header.before"

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentview.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentview.vm
@@ -41,6 +41,9 @@ $services.progress.endStep()
 ## ----------------------------
 $services.progress.startStep('Display the title and content')
 <div class="xcontent">
+  #foreach ($uix in $services.uix.getExtensions('org.xwiki.platform.template.content.header.before'))
+    $services.rendering.render($uix.execute(), 'xhtml/1.0')
+  #end
   #template('contentheader.vm')
   #foreach ($uix in $services.uix.getExtensions('org.xwiki.platform.template.content.header.after'))
     $services.rendering.render($uix.execute(), 'xhtml/1.0')


vmassol:
@slauriere hi. I don't recall if this was discussed on the forum. I think it's important since it becomes an API that cannot be broken and every single UIXP makes it a lot more harder to implement a new skin. I'm really worried about the fact that it's going to become near to impossible to create any new skin if we add too many UIXPs and I'd like to know how we address this and what strategy we can have.

I'm also not sure if this UIXP is needed. Maybe it's possible to use CSS to make the after header UIXP be displayed before the header if that's the need? The general idea is: how can we reduce the number of UIXPs needed.

Thanks for the PR!

PS: We've recently switched to HTML5 or XHTML5 (I never recall) , so I wonder if the rendering to `xhtml/1.0` should be changed or not.

slauriere:
Thanks @vmassol for the review. I'm adding also a link to the [before title UIXP](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/BeforeTitle/) which may be relevant as well, even though at first sight I don't known an easy way to achieve the same goal in pure CSS but I'll think more about and will follow up on the forum indeed. I had not considered the difficulty to implement new skins. Also adding a link for the record to [this discussion](https://forum.xwiki.org/t/ui-extension-point-for-document-content-footer/7170) about the content footer UIXP, which might be seen as the counterpart of this one.

michitux:
> We've recently switched to HTML5 or XHTML5

This is nothing recent, we only switched the WYSIWYG editor recently. The skin has been using the HTML 5 doctype since 6.2 milestone 2, see [XWIKI-7552](https://jira.xwiki.org/browse/XWIKI-7552). I think we should have been using `html/5.0` as target for rendering in particular in templates that are part of the Flamingo skin since then.

slauriere:
Good to know about `html/5.0` thx. Meantime I posted a [proposal](https://forum.xwiki.org/t/ui-extension-point-for-the-content-header/10179) on the development forum about introducing a more generic content header UIXP which would make this pull request irrelevant.

slauriere:
I'm closing this pull request since imho it should be replaced by one implementing [this proposal](https://forum.xwiki.org/t/ui-extension-point-for-the-content-header/10179) instead, if we agree on that.

PR #1820, opened 2022-04-11
XWIKI-19627: Expand Lightbox scope to also include document title area

slauriere:
* Target images with selector '.xcontent img' instead of '#xwikicontent img'


slauriere:
Hello @oanalavinia, A short message to ask you for a review of this pull request whenever time permits since it seems I can't request a review from the Reviewers section at the moment.

manuelleduc:
Hi @slauriere. May I as for an example of use case where images from the document title area need to be included in the lightbox by default?

slauriere:
HI @manuelleduc, Thank you for looking into this, I posted a use case on the [forum](https://forum.xwiki.org/t/would-it-make-sense-to-expand-the-scope-of-image-lightbox-to-xcontent-rather-than-xwikicontent/10244).

slauriere:
Closing this pull request after [further discussions](https://forum.xwiki.org/t/would-it-make-sense-to-expand-the-scope-of-image-lightbox-to-xcontent-rather-than-xwikicontent/) on the matter. I closed the corresponding issue XWIKI-19627 as "Won't Fix" and created [XWIKI-19664](https://jira.xwiki.org/browse/XWIKI-19664) instead (proposal) to implement the initially envisioned need.

PR #1826, opened 2022-04-14
XWIKI-19636: The image popup can be hard to follow on large images

oanalavinia:
* open the popover close to the mouse location by adding it to a container that changes position based on mouse events

Small demo:
https://user-images.githubusercontent.com/22794181/163386780-150a0b66-69a8-4579-82a2-efe82c730e23.mp4

### DIFF
--- a/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
+++ b/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
@@ -172,16 +172,20 @@ define('xwiki-lightbox', [
   /*
    * Make sure that the toolbar will remain open also while hovering it, not just the image.
    */
-  var keepToolbarOpenOnHover = function(image, hideTimeout) {
-    hideTimeout = setTimeout(function() {
+  var keepToolbarOpenOnHover = function() {
+    // Hide the image popover after 3 seconds (if the mouse doesn't enter the popover).
+    var hideTimeout = setTimeout(function() {
       $('#imagePopoverContainer').popover('hide');
     }, 3000);
+    // Don't hide the image popover when the mouse is over it.
     $('#imagePopoverContainer .popover').on('mouseenter', function() {
       clearTimeout(hideTimeout);
     });
+    // Hide the image popover when the mouse leaves its area.
     $('#imagePopoverContainer .popover').on('mouseleave', function() {
       $(this).popover('hide');
     });
+    return hideTimeout;
   };
 
   /**
@@ -235,10 +239,10 @@ define('xwiki-lightbox', [
       showTimeout = setTimeout(function() {
         popoverContainer.css({top: e.pageY, left: e.pageX });
         popoverContainer.popover('show');
-      }, 800);
+      }, 400);
     }).on('mouseleave', function() {
       clearTimeout(showTimeout);
-      keepToolbarOpenOnHover($(this), hideTimeout);
+      hideTimeout = keepToolbarOpenOnHover();
     });
   };
 
--- a/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
+++ b/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
@@ -231,7 +231,6 @@ define('xwiki-lightbox', [
     lightboxImages.on('mouseenter', function(e) {
       clearTimeout(hideTimeout);
       popoverContainer.data('target', e.target);
-      $(this).popover('show');
     }).on('mousemove', function(e) {
       popoverContainer.popover('hide');
       // Delay to show the popover until the mouse stops moving.
--- a/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
+++ b/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
@@ -172,15 +172,15 @@ define('xwiki-lightbox', [
   /*
    * Make sure that the toolbar will remain open also while hovering it, not just the image.
    */
-  var keepToolbarOpenOnHover = function(image, timeout) {
-    timeout = setTimeout(function() {
-      image.popover('hide');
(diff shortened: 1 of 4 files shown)


oanalavinia:
The idea is to not have the popover follow the cursor, which could be annoying in some case (e.g. you just go with the mouse over the image, but without the intention to open the lightbox). So for this I added a small timeout to be sure that the image popover was the target.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -172,36 +172,39 @@ define('xwiki-lightbox', [
   /*
    * Make sure that the toolbar will remain open also while hovering it, not just the image.
    */
-  var keepToolbarOpenOnHover = function(image, timeout) {
-    timeout = setTimeout(function() {
-      image.popover('hide');
+  var keepToolbarOpenOnHover = function(image, hideTimeout) {
+    hideTimeout = setTimeout(function() {
```
mflorea, line 176: You get ``hideTimeout`` as parameter and then you overwrite it right away. Doesn't make sense. It means you don't need it as input parameter. Note that this doesn't overwrite the ``hideTimeout`` from ``enableToolbarPopovers``. So something is fishy here.

Also, a comment wouldn't hurt:

```suggestion
    // Hide the image popover after 3 seconds (if the mouse doesn't enter the popover).
    hideTimeout = setTimeout(function() {
```

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -172,36 +172,39 @@ define('xwiki-lightbox', [
...
+      $('#imagePopoverContainer').popover('hide');
     }, 3000);
-    $('.popover').on('mouseenter', function() {
-      clearTimeout(timeout);
+    $('#imagePopoverContainer .popover').on('mouseenter', function() {
```
mflorea, line 179:
```suggestion
    // Don't hide the image popover when the mouse is over it.
    $('#imagePopoverContainer .popover').on('mouseenter', function() {
```

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -172,36 +172,39 @@ define('xwiki-lightbox', [
...
+      clearTimeout(hideTimeout);
     });
-    $('.popover').on('mouseleave', function() {
-      image.popover('hide');
+    $('#imagePopoverContainer .popover').on('mouseleave', function() {
```
mflorea, line 182:
```suggestion
    // Hide the image popover when the mouse leaves its area.
    $('#imagePopoverContainer .popover').on('mouseleave', function() {
```

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -210,20 +217,32 @@ define('xwiki-lightbox', [
...
         $('.popover .copyImageId').attr('data-imageId', imageId);
         $('.popover .copyImageId').parent().removeClass('hidden');
       }
-    }).on('mouseenter', function() {
-      clearTimeout(timeout);
+    });
+    lightboxImages.on('mouseenter', function(e) {
+      clearTimeout(hideTimeout);
+      popoverContainer.data('target', e.target);
       $(this).popover('show');
```
mflorea, line 234: I find it strange that you call the ``.popover()`` function on 3 different things:

* here you call it on the image (``this`` is an image here)
* above you call it on ``$('#imagePopoverContainer')``
* and you also call it on ``$('#imagePopoverContainer .popover')``

Are you sure it works fine in all these cases? Are all these referring to the same popover? I'd be surprised if that is the case.

mflorea: Also, you're showing the popover directly (immediately) on mouse enter and then you hide it directly (immediately) on mouse move. Normally a mouse enter event is always followed by a mouse move (it's really hard to move the mouse on top of some element without moving it over the element). Isn't it point-less to show and then immediately hide the popover?

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -172,36 +172,43 @@ define('xwiki-lightbox', [
...
    */
   var enableToolbarPopovers = function() {
-    var timeout;
-    lightboxImages.popover({
+    var hideTimeout;
+    var showTimeout;
+    // In order to place the popover at cursor location, it would be added to an element that changes its position
+    // based on mouse events.
+    var popoverContainer = $('#imagePopoverContainer');
+    popoverContainer.popover({
```
mflorea, line 200: Don't you need to update the page  objects or the functional tests? I'm thinking at the method that forces the image popover to be shown on hover.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
oanalavinia: Indeed, there is no need to show it on mouseenter, which was anyway misused.
But `$('#imagePopoverContainer .popover')` and `$('#imagePopoverContainer')` are pointing to the same popover, since you can use the element which triggered the popover and the popover itself for referencing the same popover.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
oanalavinia: Forgot it, thanks! The test were not failing since after the upgrade to selenium 4 the workaround is not needed anymore.

PR #1834, opened 2022-05-04
XWIKI-19682: Solr ParseException from SimpleEventQuery when comparing with equality to empty string

michitux:
* Escape empty strings as "" for equality queries
* Match missing values when the value is `null` (previously, it matched the string "null")

Jira issue: https://jira.xwiki.org/browse/XWIKI-19682

### DIFF
--- a/xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-stores/xwiki-platform-eventstream-store-solr/src/main/java/org/xwiki/eventstream/store/solr/internal/SolrEventStore.java
+++ b/xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-stores/xwiki-platform-eventstream-store-solr/src/main/java/org/xwiki/eventstream/store/solr/internal/SolrEventStore.java
@@ -584,7 +584,7 @@ public class SolrEventStore extends AbstractAsynchronousEventStore
         } else {
             builder.append('(');
             builder.append(StringUtils
-                .join(condition.getValues().stream().map(this.utils::toFilterQueryString).iterator(), " OR "));
+                .join(condition.getValues().stream().map(this.utils::toCompleteFilterQueryString).iterator(), " OR "));
             builder.append(')');
 
             return builder.toString();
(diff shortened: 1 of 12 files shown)


tmortagne:
> Escape empty strings as "" in SolrUtils

It does not seems to be the case.

michitux:
> It does not seem to be the case.

It was the case in the first commit, the second commit reverts this as it caused problems with other kinds of queries. In particular, the query that the second commit adds in the test failed with an exception.

xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-stores/xwiki-platform-eventstream-store-solr/src/main/java/org/xwiki/eventstream/store/solr/internal/SolrEventStore.java
```diff
@@ -677,7 +677,16 @@ private String serializeCompareCondition(CompareQueryCondition condition)
 
         switch (condition.getType()) {
             case EQUALS:
-                builder.append(toFilterQueryString(condition.getProperty(), condition.getValue()));
+                String queryString = toFilterQueryString(condition.getProperty(), condition.getValue());
+                if (queryString == null) {
+                    // See https://stackoverflow.com/a/28859224.
```
manuelleduc, line 682: A short summary of the content of the link could be interesting imo.

tmortagne: Yes, I'm not a fan of giving stackoverflow links in comment, I much prefer an explanation and more official references.

manuelleduc: Well, imo having any kind of link is fine, but it not self-sufficient as links can deprecate quite quickly.

michitux:
I've rebased on master and introduced helper methods in `SolrUtils` and replaced almost all existing usages with the new helper methods.

tmortagne (approved):

PR #1838, opened 2022-05-11
[Misc] Always publish local builds to Gradle Enterprise

alextu:


vmassol:
hmmm I'm not sure about that. One issue (and one reason I had turned this off) was that you get a warning if the developer doesn't have the permission to publish (which is the case for all devs by default). Also, I don't think it makes our build more lightweight to do so as it involves an extra step (and also an extra point of failure).

If this is only temporary to gather some data, we could do it (as I did back then) but I'm not sure we want to keep it enabled.

Opinion?

Thx

vmassol:
BTW, a bit off topic ;): currently our devs are still seeing `[INFO] A build scan was not published as you have not authenticated with server 'ge.xwiki.org'.` at the bottom of every build. I remember reporting this and I think you told me it was a bug (since we the publishing feature turned off) and that it would be fixed. Would you have any feedback on this?

vmassol:
@alextu FYI, discussion started at https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$16522742451412OROxj:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=attendees.fosdem.org

alextu:
Sorry for the late response, will follow up on Slack for the reported issue. As for enabling the "always publish", we usually do this for getting significant data as well as debugging issues (failing tests across various builds,...). But it's your call :)

PR #1840, opened 2022-05-12
XWIKI-19707: Add JavaScript API for extending Live Data with menu item and panel

michitux:
* Add an array of extension panel objects to the logic.
* Add a new advanced panel that represents an extension.
* Add menu items for all extension panels.

Jira issue: https://jira.xwiki.org/browse/XWIKI-19707

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
@@ -82,31 +82,10 @@
 
       <li class="dropdown-header">{{ $t('livedata.dropdownMenu.panels') }}</li>
 
-      <!-- Properties Panel -->
-      <li>
-        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, 'propertiesPanel')">
-          <XWikiIcon :icon-descriptor="{name: 'list-bullets'}" /> {{ $t('livedata.dropdownMenu.panels.properties') }}
-        </a>
-      </li>
-
-      <!-- Sort Panel -->
-      <li>
-        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, 'sortPanel')">
-          <XWikiIcon :icon-descriptor="{name: 'table_sort'}"/> {{ $t('livedata.dropdownMenu.panels.sort') }}
-        </a>
-      </li>
-
-      <!-- Filter Panel -->
-      <li>
-        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, 'filterPanel')">
-          <XWikiIcon :icon-descriptor="{name: 'filter'}"/> {{ $t('livedata.dropdownMenu.panels.filter') }}
-        </a>
-      </li>
-
-      <li v-for="extensionPanel in logic.extensionPanels" :key="extensionPanel.id">
-        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, extensionPanel.id)">
-          <XWikiIcon :icon-descriptor="{name: extensionPanel.icon}"/>
-          {{ extensionPanel.name }}
+      <li v-for="panel in logic.panels" :key="panel.id">
+        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, panel.id)">
+          <XWikiIcon :icon-descriptor="{name: panel.icon}"/>
+          {{ panel.name }}
         </a>
       </li>
 
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
@@ -103,6 +103,13 @@
         </a>
       </li>
 
+      <li v-for="extensionPanel in logic.extensionPanels" :key="extensionPanel.id">
+        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, extensionPanel.id)">
+          <XWikiIcon :icon-descriptor="{name: extensionPanel.icon}"/>
+          {{ extensionPanel.name }}
+        </a>
+      </li>
+
     </ul>
 
   </div>
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
@@ -28,15 +28,15 @@
     all the Advanced Panels default behavior
   -->
   <LivedataBaseAdvancedPanel
-      :class="['livedata-advanced-panel-extension', `livedata-advanced-panel-${extension.id}`]"
-      :panel-id="extension.id"
+      :class="['livedata-advanced-panel-extension', `livedata-advanced-panel-${panel.id}`]"
+      :panel-id="panel.id"
       ref="basePanel"
   >
 
     <!-- Provide the panel title and icon to the `header` slot -->
     <template #header>
-      <XWikiIcon :icon-descriptor="{name: extension.icon}"/>
-      {{ extension.title }}
+      <XWikiIcon :icon-descriptor="{name: panel.icon}"/>
+      {{ panel.title }}
     </template>
 
     <!-- Define panel content inside the `body` slot -->
@@ -61,7 +61,7 @@ export default {
 
   inject: ["logic"],
 
-  props: {'extension': Object},
+  props: {'panel': Object},
 
   methods: {
     attachContainer()
@@ -70,16 +70,16 @@ export default {
       if ('bodyContainer' in this.$refs) {
         const bodyContainer = this.$refs.bodyContainer;
         if (!bodyContainer.hasChildNodes()) {
-          bodyContainer.appendChild(this.extension.container);
-        } else if (bodyContainer.firstChild !== this.extension.container) {
-          bodyContainer.firstChild.replaceWith(this.extension.container);
+          bodyContainer.appendChild(this.panel.container);
+        } else if (bodyContainer.firstChild !== this.panel.container) {
+          bodyContainer.firstChild.replaceWith(this.panel.container);
         }
       }
     }
   },
 
   watch: {
-    'extension.container': function () {
+    'panel.container': function () {
       this.attachContainer();
     }
   },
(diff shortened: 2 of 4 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
```diff
@@ -103,6 +103,13 @@
         </a>
       </li>
 
+      <li v-for="extensionPanel in logic.extensionPanels" :key="extensionPanel.id">
+        <a href="#" @click.prevent="logic.uniqueArrayToggle(logic.openedPanels, extensionPanel.id)">
+          <XWikiIcon :icon-descriptor="{name: extensionPanel.icon}"/>
+          {{ extensionPanel.name }}
+        </a>
+      </li>
+
```
manuelleduc, line 112: I wonder if it's interesting to visually separate the 'standard' panels from the one provided by extension.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -1451,6 +1452,23 @@ define('xwiki-livedata', [
...
+     * The panel must have the following attributes:
+     * * id: the id of the panel, must be unique among all panels, also used as suffix of the class on the panel
+     * * name: the name that shall be shown in the menu
+     * * title: the title that shall be displayed in the title bar of the panel
+     * * icon: the name of the icon for the menu and the title of the panel
+     * * bodyElement: the Element that shall be attached to the panel's body, this should contain the main UI
+     *
+     * @param {Object} panel the panel to add
+     */
+    addExtensionPanel(panel)
```
manuelleduc, line 1469: I wonder if `registerExtensionPanel` is not a better name for this. WDYT?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
```diff
@@ -0,0 +1,99 @@
...
+  The LiveDataAdvancedPanelExtension component is used to provide
+  extensions a way to inject a custom panel.
+-->
+<template>
+  <!--
+    Uses the LivedataBaseAdvancedPanel as root element, as it handles for us
+    all the Advanced Panels default behavior
+  -->
+  <LivedataBaseAdvancedPanel
+      :class="['livedata-advanced-panel-extension', 'livedata-advanced-panel-' + extension.id]"
```
manuelleduc, line 31:
```suggestion
      :class="['livedata-advanced-panel-extension', `livedata-advanced-panel-${extension.id}`]"
```

michitux:
Example usage on the wiki index:

```javascript
let extension = {
  id: 'myExtensionpanel', 
  name: "Hello World Extension",
  title: 'Hello World',
  icon: 'camera',
  bodyElement: document.createElement('div')
};
document.getElementById("wikis").__vue__.logic.addExtensionPanel(extension);
extension.bodyElement.innerText = "Hello World!";
```

For an actual usage, we should fix [XWIKI-19635](https://jira.xwiki.org/browse/XWIKI-19635) to allow accessing the logic more easily. Further, we might need an event when the logic has been initialized on a new Live Data instance as I think there is currently no such event.

Note that the whole extension is reactive, you can change the name, title, icon and even the HTML element after adding the extension.

Things I'm wondering:

* Should we make the actual array of extension panels the API or expose just this method that I've added? Exposing the array could expose additional flexibility like inserting an extension only if another one is not already present, re-ordering extensions etc.
* In general, should we introduce some kind of ordering/priority for extension panels and maybe also for the existing panels such that extensions can be inserted in between the existing panels? For the latter, we could add something similar to an extension object for the existing panels but with an extra "type" or so that specifies which panel implementation should be used.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
michitux: I don't think that's interesting from a user's point of view as in general we don't separate between output that has been produced in extensions vs. in core. In fact, I think it could be more interesting to let extensions add panels between the existing ones.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
```diff
@@ -0,0 +1,99 @@
...
+  components: {
+    XWikiIcon,
+    LivedataBaseAdvancedPanel,
+  },
+
+  inject: ["logic"],
+
+  props: {'extension': Object},
+
+  methods: {
```
manuelleduc, line 66: Could be interesting to add a Vue test for this component. Especially because its logic it not really straightforward.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
manuelleduc: Sounds good. Definitely for a next iteration. Having the extension panels at the end of the list is fine for now.

michitux:
If you have any idea how we could reduce the number of nested containers for the extension panel, I would be happy as currently we have:

1. `div.panel-body`
2. `div.extension-body`
3. The container (probably `div`) provided by the extension.
4. The actual content.

In theory, container two and three shouldn't be needed, in particular the second one is really just to have a reference on some element.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
```diff
@@ -103,6 +103,13 @@
         </a>
       </li>
 
+      <li v-for="extensionPanel in logic.extensionPanels" :key="extensionPanel.id">
```
mflorea, line 106: I would have preferred to have ``logic.panels``, initialized by default with the properties, sort and filter panels. Then others could simply call ``logic.addPanel`` to add additional panels.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
```diff
@@ -0,0 +1,99 @@
...
+  <LivedataBaseAdvancedPanel
+      :class="['livedata-advanced-panel-extension', 'livedata-advanced-panel-' + extension.id]"
+      :panel-id="extension.id"
+      ref="basePanel"
+  >
+
+    <!-- Provide the panel title and icon to the `header` slot -->
+    <template #header>
+      <XWikiIcon :icon-descriptor="{name: extension.icon}"/>
+      {{ extension.title }}
```
mflorea, line 39: Do we need both ``name`` and ``title``?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
```diff
@@ -0,0 +1,99 @@
...
+  props: {'extension': Object},
+
+  methods: {
+    attachBodyElement()
+    {
+      // Make sure nothing happens if this method is called several times or at the wrong time.
+      if ('bodyContainer' in this.$refs) {
+        const bodyContainer = this.$refs.bodyContainer;
+        if (!bodyContainer.hasChildNodes()) {
+          bodyContainer.appendChild(this.extension.bodyElement);
```
mflorea, line 73: I find ``bodyElement`` misleading. I'm used to ``container`` or ``wrapper``, or even ``root``. Is there an ``extension.headElement``?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
michitux: The current panels have, e.g., ` Filter... ` as name in the menu and then `Filter` as title. I wanted to give extensions the possibility to do the same.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
michitux: I agree, I'll rename it to `container`. There is no `headElement` as I didn't want to let the extension provide the icon HTML code which could have lead to inconsistencies. In fact, all of the existing panels use the `#header` slot just for an icon and a string, it might make sense to move that template into the `LivedataBaseAdvancedPanel`, in particular if we want to make them more similar to the extensions and provide their data already in a `panels` array (I'm not so sure if this could work with the translations, though).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
michitux: As I've mentioned in the comment above, I also think this could be a good idea. Not doing this now felt a bit safer in terms of not introducing any race conditions or other regressions. The question for me is still what we would store for them, there would definitely not be any HTML element but just a type, the question is if they should still get a name, title and icon? And if yes, how do we get the translated values in there?

manuelleduc:
> If you have any idea how we could reduce the number of nested containers for the extension panel, I would be happy as currently we have:
> 
>     1. `div.panel-body`
> 
>     2. `div.extension-body`
> 
>     3. The container (probably `div`) provided by the extension.
> 
>     4. The actual content.
> 
> 
> In theory, container two and three shouldn't be needed, in particular the second one is really just to have a reference on some element.

Afaik there is no easy way around in vue2.

- `slot` can't have class attribute so we can't flatten levels 1 and 2
- `template` need to have a single root so the div of the slots when using a component cannot be avoided either, so we can't flatten level 2 and 3
- we could possibly merge level 3 and 4 but that does not feel safe

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/panels/LiveDataAdvancedPanelExtension.vue
```diff
@@ -0,0 +1,99 @@
...
+  },
+
+  watch: {
+    'panel.container': function () {
+      this.attachContainer();
+    }
+  },
+
+  mounted()
+  {
```
manuelleduc, line 88: Currently, we place the bracket on the same line as the method name.

PR #1843, opened 2022-05-16
XWIKI-19653: Ability to redirect deleted page link to a new page

oanalavinia:
* added option in the UI to select a new target document and to enable for it the update backlinks and/or the create of an automatic redirect
* set the selected options on the delete request and update AutomaticRedirectCreatorListener and BackLinkUpdaterListener to consider also the DocumentDeletedEvent
* add the possibility to restore a document which was recreated at the original location to include the usecase when a user wants to restore a document after a redirect was added (delete the recreated document and restore the original version)
* add unit tests and functional tests

UI for deleting a page, with and without a target specified:
deletePageWithTarget
deletePageWithoutTarget

UI for restoring a document, with the livetable actions and the confimation modal:
Since before there was only `restore` and `cannot restore`, I added the `replace` action for when a document was recreated at the original location, but with a warning color.
restoreActions
replaceConfirmationModal


xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentDeletedEvent.java
```diff
@@ -42,28 +42,34 @@ public class DocumentDeletedEvent extends AbstractDocumentEvent
...
+
     /**
      * Constructor initializing the event filter with an
      * {@link org.xwiki.observation.event.filter.AlwaysMatchingEventFilter}, meaning that this event will match any
      * other document delete event.
      */
     public DocumentDeletedEvent()
     {
-        
+        this.sourceReference = null;
```
vmassol, line 58: This seems unnecessary

manuelleduc: It is not because `sourceReference` is `final` and need to be initialized explicitly.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentDeletedEvent.java
```diff
@@ -42,28 +42,34 @@ public class DocumentDeletedEvent extends AbstractDocumentEvent
...
      * Constructor initializing the event filter with a {@link org.xwiki.observation.event.filter.FixedNameEventFilter},
      * meaning that this event will match only delete events affecting the same document.
      * 
      * @param documentReference the reference of the document to match
      */
-    public DocumentDeletedEvent(DocumentReference documentReference)    
+    public DocumentDeletedEvent(DocumentReference documentReference)
     {
         super(documentReference);
+        this.sourceReference = documentReference;
```
vmassol, line 70: I think this need some comments since it shouldn't be necessary normally since the ref is passed in the parent constructor and it's supposed to offer a way to get it (I've noticed that it's serialized though which seems weird and would need to be checked for accuracy).

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentDeletedEvent.java
```diff
@@ -72,6 +78,15 @@ public DocumentDeletedEvent(DocumentReference documentReference)
     public DocumentDeletedEvent(EventFilter eventFilter)
     {
         super(eventFilter);
+        this.sourceReference = null;
+    }
+
+    /**
+     * @return the reference of the deleted document
+     */
+    public DocumentReference getSourceReference()
```
vmassol, line 87: This is new public method, so it' missing `@Unstable` and `@since` annotations. Also feels weird, see my previous comment; I'd expect a AbstractDocumentEvent to have a Document source and a getter to get it.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
```diff
@@ -0,0 +1,77 @@
...
+{
+    @BeforeEach
+    void setUp(TestUtils setup)
+    {
+        setup.loginAsSuperAdmin();
+    }
+
+    @Test
+    @Order(1)
+    void restoreAfterDocumentWasReacreated(TestUtils testUtils, TestReference reference) throws Exception
```
vmassol, line 56: typo

vmassol: I don't understand the name of this test as I don't see what "restore after document was recreated" means, and how it's related to deleting.

oanalavinia:
> Thanks @oanalavinia ! Do you have tests for this new redirect on delete feature? I looked quickly but didn't spot on (but I admit I looked very quickly ;)).

Yes, there is https://github.com/xwiki/xwiki-platform/pull/1843/files#diff-6f99d6f68c08c15585b1efaa4931219d51a335ebeebb9261b6eca84e489d2120R652 for the updateLinks and autoRedirec

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
```diff
@@ -0,0 +1,77 @@
...
+import org.xwiki.index.test.po.RestoreDocumentConfirmationModal;
+import org.xwiki.test.docker.junit5.TestReference;
+import org.xwiki.test.docker.junit5.UITest;
+import org.xwiki.test.ui.TestUtils;
+import org.xwiki.test.ui.po.DeletePageConfirmationPage;
+import org.xwiki.test.ui.po.DeletingPage;
+import org.xwiki.test.ui.po.ViewPage;
+
+/**
+ * Tests for Deleted Pages page.
```
vmassol, line 40: Is this the "Deleted Pages" tags in `XWiki.AllDocs`? If so maybe mention it in the javadoc.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
vmassol: Maybe `restoreDocumentAfterDeletionAndRecreation`? (if that's what it does)

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @version $Id$
+ * @since 14.4RC1
+ */
+@UITest
+public class DeletedDocsIT
```
manuelleduc, line 46: `public` is not needed

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
manuelleduc, line 56:
```suggestion
    void restoreAfterDocumentWasRecreated(TestUtils testUtils, TestReference reference) throws Exception
```

manuelleduc: Also, the `throws Exception` is not required.

xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/web/DeleteActionCompatibilityAspect.aj
```diff
@@ -58,6 +58,6 @@ public privileged aspect DeleteActionCompatibilityAspect
     public boolean DeleteAction.deleteToRecycleBin(EntityReference entityReference, XWikiContext context)
         throws XWikiException
     {
-        return this.deleteDocument(entityReference, context, false);
+        return this.deleteDocument(entityReference, context, false, null, false, false);
```
vmassol, line 61: This is sending a warning bell to me as we're not supposed to touch legacy code, and it usually means that some public API was broken.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
```diff
@@ -240,7 +246,14 @@ private boolean deleteDocument(XWikiContext context, boolean shouldSkipRecycleBi
             doesAffectChildren(request, doc.getDocumentReference()) ? doc.getDocumentReference().getLastSpaceReference()
                 : doc.getTranslatedDocument(context).getDocumentReferenceWithLocale();
 
-        return deleteDocument(documentReference, context, shouldSkipRecycleBin);
+        return deleteDocument(documentReference, context, shouldSkipRecycleBin, newTarget, updateLinks, autoRedirect);
+    }
+
+    protected boolean deleteDocument(EntityReference entityReference, XWikiContext context,
```
vmassol, line 252: This makes it a new SPI. Is that absolutely needed?

vmassol: BTW requires a `@since` and I'd say `@Unstable` too

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
oanalavinia: Yes, it's from `XWiki.AllDocs`. I'll mention

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
```diff
@@ -317,4 +339,14 @@ private Job startDeleteJob(EntityReference entityReference, XWikiContext context
             throw new XWikiException(String.format("Failed to schedule the delete job for [%s]", entityReference), e);
         }
     }
+
+    protected DocumentReferenceResolver<String> getCurrentStringDocumentReferenceResolver()
```
vmassol, line 343: Same, this is become visible to users (who extend DeleteAction). Best to avoid it if possible and needs `@since` and `@Unstable`.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-pageobjects/src/main/java/org/xwiki/index/test/po/DeletedDocsPage.java
```diff
@@ -0,0 +1,91 @@
...
+        return this.documentsTrashLivetable.getRowNumberForElement(By.xpath("//a[text()='" + docName + "']"));
+    }
+
+    private WebElement getReplaceAction(String docName)
+    {
+        return this.documentsTrashLivetable.getCell(getRowIndexByDocName(docName), 6)
+            .findElement(By.className("replace"));
+    }
+
+    public Optional<RestoreDocumentConfirmationModal> tryReplaceDoc(String docName)
```
manuelleduc, line 80: Some javadoc would be nice. Also, I don't think the combination of try/catch + Optional is the right pattern. The test code using this method has not idea why the test failed and that's not helping when debugging. I'd be in favor the let the `TimeoutException` propagate (or to use a an alternative to `Optional` that have a field to store the exception, which does not exist by default in java currently).

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-docker/src/test/it/org/xwiki/index/test/ui/docker/DeletedDocsIT.java
oanalavinia: Yes, that's the scope of the test. Updating it

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-pageobjects/src/main/java/org/xwiki/index/test/po/RestoreDocumentConfirmationModal.java
```diff
@@ -0,0 +1,52 @@
...
+{
+    public RestoreDocumentConfirmationModal()
+    {
+        super(By.id("restoreDocumentConfirmModal"));
+        waitUntilReady();
+    }
+
+    private void waitUntilReady()
+    {
+        getDriver().waitUntilElementIsVisible(RestoreDocumentConfirmationModal.this.container,
```
manuelleduc, line 42: Out of curiosity, why `RestoreDocumentConfirmationModal.this.container`? Afaics, `this.container` is enough an less verbose.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1468,7 +1468,7 @@ activitystream.event.deleteComment.rss.body=A comment has been deleted from the
 ### Deleting a page
 core.delete=Delete
 core.delete.title=Delete {0}
-core.delete.backlinksWarning=The following pages contain links to the current page:{0}After deleting this page, those links will point to an empty page.
+core.delete.backlinksInfo=After deleting this page, the {0,choice,0#incoming links|1#incoming link|1<incoming links} will point to an empty page. To avoid this, you can select a new target.
```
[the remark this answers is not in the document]
vmassol: My bad, it's a different key :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1468,7 +1468,7 @@ activitystream.event.deleteComment.rss.body=A comment has been deleted from the
 ### Deleting a page
 core.delete=Delete
 core.delete.title=Delete {0}
-core.delete.backlinksWarning=The following pages contain links to the current page:{0}After deleting this page, those links will point to an empty page.
```
vmassol, line 1471: Has this been deprecated?

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-pageobjects/src/main/java/org/xwiki/index/test/po/RestoreDocumentConfirmationModal.java
oanalavinia: I started with a `waitUntilCondition` and I forgot to update this part, thanks!

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/DeleteRequest.java
```diff
@@ -0,0 +1,162 @@
...
+
+import org.xwiki.model.reference.DocumentReference;
+
+/**
+ * Job request for deleting a page.
+ * 
+ * @version $Id$
+ * @since 14.4RC1
+ */
+public class DeleteRequest extends EntityRequest
```
manuelleduc, line 30: Probably needs to be `@Unstable`

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
oanalavinia: No, I forgot this is the practice for deleted keys. Thanks!

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/delete.js
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+require(['jquery'], function($) {
+    $('#delete #newTarget').on('keyup change', function() {
+        let updateLinksContainer = $("#delete input[name=updateLinks]").closest('dt');
+        let autoRedirectContainer = $("#delete input[name=autoRedirect]").closest('dt');
+        let elements= updateLinksContainer.add(updateLinksContainer.next('dd'))
```
manuelleduc, line 24:
```suggestion
        let elements = updateLinksContainer.add(updateLinksContainer.next('dd'))
```

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/internal/listener/BackLinkUpdaterListenerTest.java
```diff
@@ -19,17 +19,29 @@
  */
 package org.xwiki.refactoring.internal.listener;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
manuelleduc, line 22: Unless I've something miss-configured in my IDE, I find it surprising to see those imports reordered.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
oanalavinia: I added this for https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteSpaceAction.java#L44
But I think I could drop this and just update directly the method call from `DeleteSpaceAction` to use the default values for the extra parameters.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/test/java/org/xwiki/refactoring/internal/listener/BackLinkUpdaterListenerTest.java
oanalavinia: I missed that the imports were rearranges when adding the changes for this file. I'll change them back
But probably it's something from my IDE since I get this quite often

manuelleduc: This should help: https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HConfiguringyourIDEtousetheXWikicodestyle
If it does not let me know :)

oanalavinia: Found the problem. My shortcut for organizing imports (from eclipse) moved it, so it's not related to code style

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-pageobjects/src/main/java/org/xwiki/index/test/po/DeletedDocsPage.java
oanalavinia: Indeed, if the modal is not opened then you need to also know why when debugging. I'll look into it

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/event/DocumentDeletedEvent.java
```diff
@@ -42,28 +42,34 @@ public class DocumentDeletedEvent extends AbstractDocumentEvent
      * changes.
      */
     private static final long serialVersionUID = 1L;
-    
+
+    /**
+     * The reference of the deleted document.
+     */
+    private final DocumentReference sourceReference;
```
tmortagne, line 49: Would be cleaner to introduce that (but under the more generic "documentReference" naming) at AbstractDocumentEvent level IMO, it does not make much sense to have only DocumentDeletedEvent expose this.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/job/DeleteJob.java
```diff
@@ -147,4 +147,14 @@ private void maybeDelete(DocumentReference documentReference)
...
+
+    /**
+     * {@inheritDoc}
+     * 
+     * @see org.xwiki.refactoring.internal.job.AbstractEntityJob#getCommonParent()
+     */
+    public EntityReference getCommonParent()
+    {
+        return getCommonParent(this.request.getEntityReferences());
+    }
```
mflorea, line 159: Why is this overwrite needed? Just to make it public? Then why not return ``super.getCommonParent()``? BTW, I'm not sure the reason you made this public is good. See below.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/listener/AutomaticRedirectCreatorListener.java
```diff
@@ -71,11 +84,34 @@ public void onEvent(Event event, Object source, Object data)
...
                 this.modelBridge.createRedirect(documentRenamedEvent.getSourceReference(),
                     documentRenamedEvent.getTargetReference());
             }
+        } else if (event instanceof DocumentDeletedEvent && this.jobContext.getCurrentJob() instanceof DeleteJob) {
+            DeleteJob job = (DeleteJob) this.jobContext.getCurrentJob();
+            DeleteRequest request = (DeleteRequest) job.getRequest();
+            DocumentDeletedEvent documentDeletedEvent = (DocumentDeletedEvent) event;
+
+            // For case when the delete affects also the child pages, only the root document should have a redirect
+            // created.
```
mflorea, line 98: Why? Note that a delete job request can include a list of documents that have no parent-child relation. You simply want to delete A, B and C, unrelated documents. In that case ``getCommonParent()`` won't return what you expect. So there's no concept of a 'root' document for a delete action.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/listener/BackLinkUpdaterListener.java
```diff
@@ -79,41 +85,77 @@ public class BackLinkUpdaterListener extends AbstractEventListener
...
-                updateBackLinks((DocumentRenamedEvent) event, canEdit, updateLinksOnFarm);
-            }
+    private void maybeUpdateLinksAfterDelete(Event event)
+    {
+        DeleteJob job = (DeleteJob) this.jobContext.getCurrentJob();
+        DeleteRequest request = (DeleteRequest) job.getRequest();
+        Predicate<EntityReference> canEdit = entityReference -> (job.hasAccess(Right.EDIT, entityReference));
+        DocumentDeletedEvent deletedEvent = (DocumentDeletedEvent) event;
+
+        // In case the delete affected the child pages too, only the root document should have the links updated.
```
mflorea, line 116: Why? Again, there's no notion of a "root" document for a delete action. The delete job deletes the **list** of documents specified by the delete job request.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/DeleteRequest.java
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Key of the optional property that indicates whether the document should be send to the recycle bin or removed
+     * permanently.
+     */
+    public static final String SHOULD_SKIP_RECYCLE_BIN = "shouldSkipRecycleBin";
+
+    /**
+     * @see #getNewTarget()
+     */
+    public static final String NEW_TARGET = "newTarget";
```
mflorea, line 41: You should use something more explicit, such as "new **link** target" or "new **backlink** target".

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/DeletePageConfirmationPage.java
```diff
@@ -67,8 +68,80 @@ public boolean isRecycleBinOptionsDisplayed()
         }
     }
 
+    /**
+     * @return the arrow element used for opening or closing the backlinks panel
+     * @since 14.4RC1
+     */
+    public WebElement getBacklinkPanelExpandButton()
```
mflorea, line 75: If this is used only for click, then I'd rather have a "toggle*" or "click*" (high level) method.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/delete.js
```diff
@@ -0,0 +1,33 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+require(['jquery'], function($) {
+    $('#delete #newTarget').on('keyup change', function() {
```
mflorea, line 21: We use 2-space indentation for JavaScript.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/internal/listener/AutomaticRedirectCreatorListener.java
oanalavinia: So the `newBacklinkTarget` should be taken into consideration for all the documents of a delete request? Couldn't this be an unexpected outcome when selecting `affectChildren`? (although an warning can be added to the UI)

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-test/xwiki-platform-index-test-pageobjects/src/main/java/org/xwiki/index/test/po/DeletedDocsPage.java
```diff
@@ -0,0 +1,69 @@
...
+import org.xwiki.test.ui.po.ViewPage;
+
+/**
+ * Represents the actions possible on the Deleted Pages page.
+ *
+ * @version $Id$
+ * @since 14.4.1
+ * @since 14.5RC1
+ */
+@Unstable
```
manuelleduc, line 35: Page Objects are only needed for tests and are not public API.  `@Unstable` is not required.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
manuelleduc: Can this method be considered as deprecated now? If this is the case, it can be moved to legacy (see https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HDeprecatingCode)

oanalavinia: Yes, I think this should be done. I wasn't aware of the process. Thanks!

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -228,7 +231,37 @@
       #end
       </ul>
     #end
-    #set($body = $services.localization.render('core.delete.backlinksWarning', [$message]))
+    #define($body)
+      <p class="noitems">
+        $services.localization.render('core.delete.backlinksInfo', [$doc.backlinks.size()])
```
michitux, line 236: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -228,7 +231,37 @@
...
+      </p>
+      <dl>
+        <dt>
+          <label for="newBacklinkTarget">$services.localization.render('core.delete.backlinkTarget.label')</label>
```
michitux, line 240: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -228,7 +231,37 @@
...
+          <span class="xHint">$services.localization.render('core.delete.backlinkTarget.hint')</span>
```
michitux, line 241: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/refactoring_macros.vm
```diff
@@ -106,6 +106,36 @@
...
+  ## - We don't have management UIs for redirects FTM and reducing the number of redirects make the wiki easier
+  ##   to manage.
+  ## In the future we'll offer a config option to define the default behavior, see
+  ## http://jira.xwiki.org/browse/XWIKI-13384
+  #set ($checked = $request.autoRedirect == 'true')
+  #set ($hidden = $options.hidden)
+  <dt #if ($hidden) class="hidden"#end>
+    <label>
+      <input type="checkbox" name="autoRedirect" value="true" #if ($checked)checked="checked"#end />
+      $services.localization.render($options.label)
```
michitux, line 129: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/refactoring_macros.vm
```diff
@@ -106,6 +106,36 @@
...
+    </label>
+    ## The value submitted when the checkbox is not checked, used to preserve the form state.
+    <input type="hidden" name="autoRedirect" value="false" />
+  </dt>
+  <dd #if ($hidden) class="hidden"#end>
+    <span class="xHint">$services.localization.render($options.hint)</span>
```
michitux, line 135: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -36,7 +36,77 @@
...
+        &lt;/div&gt;
+      &lt;/div&gt;
+    &lt;/div&gt;
+  &lt;/div&gt;
+#end
+#macro(deletedDocsActionsTemplate)
+  &lt;div class="hidden deletedDocsActionsTemplate"&gt;
+    ## Restore document action template.
+    &lt;a href="" class="action restore"
+        title="$services.localization.render('platform.index.trashDocumentsActionsRestoreTooltip')"&gt;
```
michitux, line 75: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -36,7 +36,77 @@
...
+      &lt;span class="action-icon"&gt;$services.icon.renderHTML('refresh')&lt;/span&gt;&lt;span
+        class="action-label"&gt;$escapetool.xml(
+        $services.localization.render('platform.index.trashDocumentsActionsRestoreText'))
+      &lt;/span&gt;
+    &lt;/a&gt;
+    ## Orphaned translations: cannot restore action template.
+    &lt;a href="" class="action cannot-restore" title="$services.localization.render(
```
michitux, line 82: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -36,7 +36,77 @@
...
+        'platform.index.trashDocumentsActionsCannotRestoreCausesOrphanedTranslationTooltip')"&gt;
+      &lt;span class="action-icon"&gt;$services.icon.renderHTML('refresh')&lt;/span&gt;&lt;span
+        class="action-label"&gt;$escapetool.xml(
+        $services.localization.render('platform.index.trashDocumentsActionsCannotRestoreText'))
+      &lt;/span&gt;
+    &lt;/a&gt;
+    ## Replace document action template.
+    &lt;a href="" class="action replace" data-toggle='modal' data-target='#restoreDocumentConfirmModal'
+        title="$services.localization.render('platform.index.trashDocumentsActionsReplaceTooltip')"&gt;
```
michitux, line 91: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-ui/src/main/resources/XWiki/DeletedDocuments.xml
```diff
@@ -36,7 +36,77 @@
...
+      &lt;span class="action-icon"&gt;$services.icon.renderHTML('refresh')&lt;/span&gt;&lt;span
+        class="action-label"&gt;$escapetool.xml(
+        $services.localization.render('platform.index.trashDocumentsActionsReplaceText'))
+      &lt;/span&gt;
+    &lt;/a&gt;
+    ## Delete document action template.
+    &lt;a href="" class="action actiondelete"
+        title="$services.localization.render('platform.index.trashDocumentsActionsDeleteTooltip')"&gt;
```
michitux, line 99: Missing XML-escaping or output syntax parameter.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/delete.js
```diff
@@ -0,0 +1,29 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+require(['jquery'], function($) {
+  $('#delete #newBacklinkTarget').on('keyup change', function() {
```
michitux, line 21: Shouldn't this code wait on document ready or another event to make sure that the elements are actually present in order to avoid creating a race condition?

michitux (changes requested):
I've found quite some unescaped translation strings and another small comment. Further, before merging the since-versions need to be updated.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
```diff
@@ -340,7 +335,13 @@ private Job startDeleteJob(EntityReference entityReference, XWikiContext context
         }
     }
 
-    protected DocumentReferenceResolver<String> getCurrentStringDocumentReferenceResolver()
+    /**
+     * @return the current document reference resolver to be used to resolve a string document into a
+     *         {@link org.xwiki.model.reference.DocumentReference}
+     * @since 14.4.1
+     * @since 14.5RC1
```
mflorea, line 342: ``@since`` makes sense only for public / protected APIs (i.e. that can be used outside of this class).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/DeleteAction.java
```diff
@@ -340,7 +335,13 @@ private Job startDeleteJob(EntityReference entityReference, XWikiContext context
...
+     */
+    private DocumentReferenceResolver<String> getCurrentStringDocumentReferenceResolver()
```
mflorea, line 344: ``DeleteAction`` is a component now, so it shouldn't rely on ``Utils.getComponent()`` anymore. Simply inject the required components.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/DeleteRequest.java
```diff
@@ -78,21 +81,21 @@ public void setShouldSkipRecycleBin(boolean shouldSkipRecycleBin)
     }
 
     /**
-     * @return the document to be used as the new target after the delete
+     * @return the document to be used as the new backlink target after delete
```
mflorea, line 84: There's a problem here. ``DeleteRequest`` extends ``EntityRequest`` which means it targets **multiple** entities. Which is why we have ``getEntityReferences()`` ("the collection of entity references that are targeted by this request"). I think you focused too much on the UI where we delete a single page at once, but at the API level, the delete request can target multiple pages. So I'm not sure it makes sense to have a **single** new backlink target to use for all the deleted pages. It makes more sense to me to have a map ``(deleted entity reference, new backlink reference)``. This way, when you delete multiple pages at once (in the same job, with the same request) you can specify which new backlink reference to use for each of the deleted pages. Moreover, this mapping could be useful also to specify a new backlink target for some of the affected child pages (of course when the delete is done programmatically).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/DeletePageConfirmationPage.java
```diff
@@ -69,37 +69,42 @@ public boolean isRecycleBinOptionsDisplayed()
...
     /**
-     * @return the arrow element used for opening or closing the backlinks panel
-     * @since 14.4RC1
+     * Toggle the backlinks panel
+     *
+     * @since 14.4.1
+     * @since 14.5RC1
      */
-    public WebElement getBacklinkPanelExpandButton()
+    public void toggleBacklinksPanel()
```
mflorea, line 77: "panel" is confusing to me. I know the CSS class contains 'panel' but at the UI level, for the end user, it's not a panel (on the left or right of the page). We usually ca it **pane**. So ``toggleBacklinksPane``.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/DeleteRequest.java
oanalavinia: So for the UI action this would mean that only the parent document will have a new redirect when selecting `affectChildren`. Should this be better highlighted on the backlinks pane?

mflorea (approved):

PR #1844, opened 2022-05-16
XWIKI-15460: Notification filter preferences are not cleaned when a wiki is deleted

manuelleduc:
- Remove all the notification filter preferences from the main wiki after a sub-wiki is deleted
- Update the cache accordingly

Jira: https://jira.xwiki.org/browse/XWIKI-15460


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140400000XWIKI15460DataMigration.java
```diff
@@ -0,0 +1,121 @@
...
+    protected void hibernateMigrate() throws DataMigrationException
+    {
+        XWikiContext context = getXWikiContext();
+        if (!Objects.equals(context.getWikiId(), context.getMainXWiki())) {
+            this.logger.info("Skipping, this migration only applies to the main wiki.");
+            return;
+        }
+
+        int version = this.manager.get().getDBVersion().getVersion();
+        if (version >= 131006000 && version < 140000000) {
```
tmortagne, line 91: I think this kind of check is supposed to go in #shouldExecute instead of #hibernateMigrate.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140400000XWIKI15460DataMigration.java
```diff
@@ -0,0 +1,121 @@
...
+            Set<NotificationFilterPreference> allNotificationFilterPreferences = this.store
+                .getPaginatedFilterPreferences(limit, offset);
+            while (!allNotificationFilterPreferences.isEmpty()) {
+                // We count the deleted filter preferences and adapt the next offset accordingly.
+                int removed = 0;
+                for (NotificationFilterPreference filterPreference : allNotificationFilterPreferences) {
+                    boolean isFromExistingWiki = wikiIds.stream().anyMatch(filterPreference::isFromWiki);
+                    if (!isFromExistingWiki) {
+                        removed++;
+                        this.store.deleteFilterPreference(filterPreference);
```
tmortagne, line 109: Not a fan of deleting each preference one by one, I feel it would be better to gather all the not existing wikis references in the preference you are to load and then just do a single delete (or maybe one per wiki to cleanup if it's too complex).

manuelleduc: I considered this option but I got afraid of the memory usage if the number of filter preferences to retain before delete is large.

tmortagne: You don't need to keep any preferences, just the referenced wikis.

manuelleduc: Ah right, that should work indeed.

manuelleduc: We currently don't have a way to extract the wikiId from a filter preference, but that shouldn't be difficult to add.

tmortagne: Yes, it's not so different from isFromWiki (which you might actually not need anymore unless you are using it somewhere else).

manuelleduc: It's still used for cache invalidation.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140400000XWIKI15460DataMigration.java
```diff
@@ -0,0 +1,131 @@
...
+            Set<String> unknownWikiIds = new HashSet<>();
+
+            int limit = 1000;
+            int offset = 0;
+            Set<NotificationFilterPreference> allNotificationFilterPreferences = this.store
+                .getPaginatedFilterPreferences(limit, offset);
+            while (!allNotificationFilterPreferences.isEmpty()) {
+                // We count the deleted filter preferences and adapt the next offset accordingly.
+                for (NotificationFilterPreference filterPreference : allNotificationFilterPreferences) {
+                    filterPreference.getWikiId().ifPresent(wikiId -> {
```
tmortagne, line 112: This logic does not fully match the one you have in deleteFilterPreference where a preference is deleted if any of the references contains the wiki ID. But maybe it's not possible to have references with different wiki ID ?

manuelleduc: Afaiu wiki/page/pageOnly/user are exclusive. I'll double check though.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140400000XWIKI15460DataMigration.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public String getDescription()
+    {
+        return "Remove the notification filters preferences remaining from removed sub-wikis.";
+    }
+
+    @Override
+    public boolean shouldExecute(XWikiDBVersion startupVersion)
+    {
+        boolean shouldExecute = super.shouldExecute(startupVersion);
```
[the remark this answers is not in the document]
manuelleduc: I agree, I kept it as it does not hurt (and make it somehow more future proof)

manuelleduc:
Tested manually on all supported database successfully.

PR #1848, opened 2022-05-24
XWIKI-18397: Notification filter preferences are never cleaned for deleted users

manuelleduc:
- Listen for deleted users and remove the corresponding notification filter preferences from the store and from the cache
- Update R140500000XWIKI15460DataMigration to also cleanup remaining notification filter preferences for previously removed users

Jira: https://jira.xwiki.org/browse/XWIKI-18397

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/CachedModelBridge.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/CachedModelBridge.java
@@ -149,6 +149,13 @@ public class CachedModelBridge implements ModelBridge, Initializable
     }
 
     @Override
+    public void deleteFilterPreferences(DocumentReference user) throws NotificationException
+    {
+        this.modelBridge.deleteFilterPreferences(user);
+        invalidateUserPreferencesFilters(user);
+    }
+
+    @Override
     public void deleteFilterPreference(WikiReference wikiReference, String filterPreferenceId)
         throws NotificationException
     {
@@ -235,5 +242,15 @@ public class CachedModelBridge implements ModelBridge, Initializable
     {
         this.preferenceFilterCache.values()
             .forEach(set -> set.removeIf(filter -> filter.isFromWiki(wikiReference.getName())));
+    }
+
+    /**
+     * Remove all the {@link  NotificationFilterPreference}s related to the given user form the coche.
+     *
+     * @param user the document reference of the user to remove from the cache
+     */
+    private void invalidateUserPreferencesFilters(DocumentReference user)
+    {
+        this.preferenceFilterCache.remove(user);
     }
 }
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/CachedModelBridge.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/CachedModelBridge.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import javax.inject.Inject;
@@ -36,6 +37,7 @@ import org.xwiki.component.phase.InitializationException;
 import org.xwiki.model.internal.reference.EntityReferenceFactory;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.EntityReference;
+import org.xwiki.model.reference.EntityReferenceSerializer;
 import org.xwiki.model.reference.WikiReference;
 import org.xwiki.notifications.NotificationException;
 import org.xwiki.notifications.NotificationFormat;
@@ -67,6 +69,9 @@ public class CachedModelBridge implements ModelBridge, Initializable
 
     @Inject
     private EntityReferenceFactory referenceFactory;
+
+    @Inject
+    private EntityReferenceSerializer<String> entityReferenceSerializer;
 
     private Map<EntityReference, Set<NotificationFilterPreference>> preferenceFilterCache;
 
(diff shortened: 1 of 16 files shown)


manuelleduc:
Found a missing part to the current logic. When a user is removed:
1. all their preferences should be removed
2. all the users having preferences involving the user (user column) should be removed

While the first point is well handled, the second one is currently not.
Note: The same is probably true for pages and spaces.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/ModelBridge.java
```diff
@@ -89,6 +89,22 @@ default Set<NotificationFilterPreference> getFilterPreferences(WikiReference wik
...
+    /**
+     * Delete all the filter preferences related to a given user.
+     *
+     * @param user the document reference of a wiki user
+     * @throws NotificationException in case of error when deleting the filter preferences
+     * @since 14.5RC1
+     * @since 14.4.1
+     * @since 13.10.7
+     */
+    @Unstable
```
tmortagne, line 101: This class is internal.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/CachedModelBridge.java
```diff
@@ -236,4 +248,19 @@ private void invalidateWikiPreferenceFilters(WikiReference wikiReference)
...
+     * @param user the document reference of the user to remove from the cache
+     */
+    private void invalidateUserPreferencesFilters(DocumentReference user)
+    {
+        String serializedUserReference = this.entityReferenceSerializer.serialize(user);
+        // Remove the filter preferences of the user from the cache.
+        this.preferenceFilterCache.remove(user);
+        // Remove the user from the filter preferences of the other entities of the cache. 
+        this.preferenceFilterCache.values()
+            .forEach(set -> set.removeIf(filter -> Objects.equals(filter.getUser(), serializedUserReference)));
```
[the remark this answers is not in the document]
manuelleduc: Afaik it's always `wiki:XWiki.User`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
```diff
@@ -162,16 +164,18 @@ public List<DefaultNotificationFilterPreference> getPreferencesOfWiki(WikiRefere
      * @since 14.4.1
      * @since 13.10.7
      */
-    public Set<NotificationFilterPreference> getPaginatedFilterPreferences(int limit, int offset)
+    public Set<DefaultNotificationFilterPreference> getPaginatedFilterPreferences(int limit, int offset)
```
surli, line 167: why changing this signature?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
```diff
@@ -220,6 +224,44 @@ public void deleteFilterPreference(DocumentReference user, String filterPreferen
...
+     * @since 14.4.1
+     * @since 13.10.7
+     */
+    public void deleteFilterPreferences(DocumentReference user) throws NotificationException
+    {
+        XWikiContext context = this.contextProvider.get();
+
+        String oriDatabase = context.getWikiId();
+
+        context.setWikiId(context.getMainXWiki());
```
surli, line 242: not sure to understand: aren't the filter preferences stored in the wiki where the user is registered? So in case of a user of a subwiki, shouldn't it be stored in subwiki DB?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
manuelleduc: Because if need to have access to `getOwner` which is only available for `DefaultNotificationFilterPreference`, and not `NotificationFilterPreference`.

manuelleduc: Note that it's documented in the body of the method. Do you think it's interesting to add something in the javadoc too?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
manuelleduc: Hum, looks like a bug indeed, fixing.

manuelleduc: Note that the decision to use the main database for sub-wikis should be based on the `eventstream.usemainstore` property.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
surli: No I  think it's good like that, I'm actually wondering if we shouldn't just put the owner concept in `NotificationFilterPreference` directly. AFAICS we always use a `DefaultNotificationFilterPreference` in the store API in order to set the owner, so it could almost make sense to always have it defined maybe. 
But not something to do now.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
surli: > Note that the decision to use the main database for sub-wikis should be based on the eventstream.usemainstore property.

I wasn't aware about this option, but when I read about it it's only talking about storage of events, it's also used for storage of filter preferences?

PR #1849, opened 2022-05-24
XWIKI-19767: Make the security cache capacity configurable

ClemensRobbenhaar:
the new configuration setting is named:  "security.authorization.cache.capacity"

### DIFF
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
@@ -30,6 +30,7 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
```diff
@@ -117,7 +124,9 @@ private Cache<SecurityCacheEntry> newCache() throws InitializationException
         CacheConfiguration cacheConfig = new CacheConfiguration();
         cacheConfig.setConfigurationId("platform.security.authorization.cache");
         LRUEvictionConfiguration lru = new LRUEvictionConfiguration();
-        lru.setMaxEntries(DEFAULT_CAPACITY);
+        int cacheCapacity = configuration.getProperty("security.authorization.cache.capacity", DEFAULT_CAPACITY);
+        logger.debug("creating security entries cache with capacity of [{}]", cacheCapacity);
+        lru.setMaxEntries(cacheCapacity);
```
surli, line 129: So I might be wrong but AFAIR @tmortagne told me that those infinispan caches are configurable from an XML with their configuration ID. So I don't think such thing is actually needed.

tmortagne: Yes, we usually configure most caches through the Infinispan configuration file (except very old caches like the document cache, which existed long before Infinispan). See https://extensions.xwiki.org/xwiki/bin/view/Extension/Cache%20Module#HInfinispan.

ClemensRobbenhaar:
As I have been informed in the \#xwiki chat, this PR is not necessary, and the configuration should be done in the `cache/infinispan/config.xml` instead.

PR #1850, opened 2022-05-31
XWIKI-19774: Temporary attachments aren't deleted when the session is destroyed

surli:
* Ensure to listen to the session destruction for cleaning temporary
    attachments

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -101,46 +101,23 @@
                  Don't forget to '\'-escape dots and other special characters in this case ;)
 
                  Common justification example:
-
-            <revapi.differences>
-              <justification>This change is necessary to fix bug #1234</justification>
-              <criticality>highlight</criticality>
-              <differences>
-                <item>
-                  <code>java.method.addToInterface</code>
-                  <fullQualifiedClassName>com.acme.ToolInterface</fullyQualifiedClassName>
-                  <new>method void com.acme.ToolInterface::setup()</new>
-                </item>
-                <item>
-                  <code>java.method.removed</code>
-                  <fullQualifiedClassName>com.acme.ToolInterface</fullyQualifiedClassName>
-                  <old>method void com.acme.ToolInterface::initialize()</old>
-                  <justification>...</justification>
-                </item>
-              </differences>
-            </revapi.differences>
-            <revapi.differences>
-              ... other difference group here...
-            </revapi.differences>
-
-                 Single justification example:
-
-            <revapi.differences>
-              <differences>
-                <item>
-                  <code>java.method.removed</code>
-                  <old>method void org.xwiki.diff.xml.XMLDiff::xxx()</old>
-                  <justification>Unstable API added by mistake.</justification>
-                  <criticality>highlight</criticality>
-                </item>
-              </differences>
-            </revapi.differences>
             -->
-            <!-- Difference related to the Security refactoring. Start a new <revapi.differences> entry for other
-                 differences -->
-
-            
-            
+            <revapi.differences>
+              <justification>Unstable API for TemporaryAttachmentSessionManager.</justification>
+              <criticality>allowed</criticality>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.addedToInterface</code>
+                  <new>method void javax.servlet.http.HttpSessionListener::sessionCreated(javax.servlet.http.HttpSessionEvent) @ org.xwiki.store.TemporaryAttachmentSessionsManager</new>
+                </item>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.addedToInterface</code>
+                  <new>method void javax.servlet.http.HttpSessionListener::sessionDestroyed(javax.servlet.http.HttpSessionEvent) @ org.xwiki.store.TemporaryAttachmentSessionsManager</new>
+                </item>
+              </differences>
+            </revapi.differences>
           </analysisConfiguration>
         </configuration>
       </plugin>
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/pom.xml
```diff
@@ -109,12 +109,7 @@
                 <item>
                   <ignore>true</ignore>
                   <code>java.method.addedToInterface</code>
-                  <new>method void javax.servlet.http.HttpSessionListener::sessionCreated(javax.servlet.http.HttpSessionEvent) @ org.xwiki.store.TemporaryAttachmentSessionsManager</new>
-                </item>
-                <item>
-                  <ignore>true</ignore>
-                  <code>java.method.addedToInterface</code>
-                  <new>method void javax.servlet.http.HttpSessionListener::sessionDestroyed(javax.servlet.http.HttpSessionEvent) @ org.xwiki.store.TemporaryAttachmentSessionsManager</new>
+                  <new>method boolean org.xwiki.store.TemporaryAttachmentSessionsManager::removeUploadedAttachments(java.lang.String)</new>
```
manuelleduc, line 112: You can also define a default method for removeUploadedAttachments if you don't want to have a breaking change (which is fine for an unstable interface though).

xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/java/org/xwiki/container/servlet/events/SessionDestroyedEvent.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Event triggered when a {@link javax.servlet.http.HttpSession} is destroyed.
+ * This event is the xwiki event triggered when
+ * {@link javax.servlet.http.HttpSessionListener#sessionDestroyed(HttpSessionEvent)} is called.
+ *
+ *  Note that this event should *not* be serializable: we probably don't want it to be sent remotely.
+ *
+ * The following information are also sent:
+ * <ul>
+ *     <li>source: the identifier of the session {@link HttpSession#getId()}</li>
```
tmortagne, line 37: I really don't see the point, you should simply pass the HttpSession as first argument IMO.

If you really want to do something special with the id you could make it part of the event itself so that someone could listen to the destruction of a specific session for example.

surli: Hmmm I'm following a pattern that I saw a lot in XWiki which is to use a String in source of events and an actual object in data of the event. Maybe I imagined it, but I'm usually following this pattern in the event I'm sending.

tmortagne: >  a pattern that I saw a lot in XWiki

I cannot think of any (but maybe there is), and I really don't see the point. As I said, that String is more the kind of thing we would put in the event itself so that we can filter on it, but I don't remember any duplicates like this in the parameters.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/LegacyActionServlet.java
```diff
@@ -62,6 +63,8 @@ public class LegacyActionServlet extends HttpServlet
 
     private ComponentManager rootComponentManager;
 
+    private ObservationManager observationManager;
```
tmortagne, line 66: You don't seems to need that anymore.

xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/resources/META-INF/components.txt
```diff
@@ -1,3 +1,4 @@
 org.xwiki.container.servlet.internal.CacheControlRequestInitializer
 org.xwiki.container.servlet.internal.DefaultServletContainerInitializer
 900:org.xwiki.container.servlet.internal.LegacyEnvironmentConfiguration
+org.xwiki.container.servlet.internal.HttpSessionManager
```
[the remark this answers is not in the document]
vmassol: note: this is new to me ;) (I've never done that)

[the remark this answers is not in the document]
surli: Same here, I might have done it once or twice but I very often forget to do it.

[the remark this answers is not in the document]
vmassol: Note: it's not that I forget it, but that I don't remember we ever defined it as something we wanted to do so I don't order them by name.

xwiki-platform-core/pom.xml
```diff
@@ -139,8 +139,8 @@
             <!-- Difference related to the Security refactoring. Start a new <revapi.differences> entry for other
                  differences -->
 
-            
-            
+
```
manuelleduc, line 142: leftovers?

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/test/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManagerTest.java
```diff
@@ -120,41 +123,18 @@ void uploadAttachment() throws Exception
             .thenReturn("42");
         when(part.getSize()).thenReturn(41L);
 
+        doAnswer(invocationOnMock -> {
+            TemporaryAttachmentSession temporaryAttachmentSession = invocationOnMock.getArgument(1);
+            assertEquals(sessionId, temporaryAttachmentSession.getSessionId());
+            return null;
+        }).when(httpSession).setAttribute(eq(ATTRIBUTE_KEY), any(TemporaryAttachmentSession.class));
+
```
manuelleduc, line 131: Can this be simplified using argument captor instead https://www.baeldung.com/mockito-argumentcaptor ?

PR #1854, opened 2022-06-14
[Misc] Consider user timezone when formating the image date displayed in lightbox

oanalavinia:


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/pom.xml
```diff
@@ -129,6 +129,11 @@
       <artifactId>moment-jdateformatparser</artifactId>
       <version>1.2.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.webjars.npm</groupId>
+      <artifactId>moment-timezone</artifactId>
+      <version>0.5.34</version>
+    </dependency>
```
vmassol, line 136: why isn't this dep in the lightbox module?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/javascript.vm
```diff
@@ -85,6 +85,7 @@
     'drawer': "#removeJsSuffix($services.webjars.url('drawer', ""js/drawer${jsExtension}""))",
     'deferred': $xwiki.getSkinFile('uicomponents/require/deferred.js'),
     'moment-jdateformatparser': $services.webjars.url('moment-jdateformatparser', 'moment-jdateformatparser.min'),
+    'moment-timezone': $services.webjars.url('org.webjars.npm:moment-timezone', 'builds/moment-timezone-with-data.min'),
```
vmassol, line 88: hmm does it mean we have a hardcoded dependency in flamingo. Imagine that the lightbox modules were in xwiki-contrib, how would you have done it?

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-ui/src/main/resources/XWiki/Lightbox/WebHome.xml
```diff
@@ -469,7 +469,7 @@ bundles=blueimp-gallery,blueimp-helper,blueimp-gallery-fullscreen,blueimp-galler
       &lt;div class="descriptionRow"&gt;
         &lt;span class="publisher"
           aria-label="$escapetool.xml($services.localization.render('image.lightbox.publisher.label'))"&gt;&lt;/span&gt;
-        &lt;span class="date" data-dateFormat="$dateFormat"
+        &lt;span class="date" data-dateFormat="$dateFormat" data-userTimezone="$xwiki.getUserTimeZone()"
```
vmassol, line 472: small note: I'd write `data-userTimeZone` instead. It seems it takes a capital Z (e.g. https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/javascript.vm
vmassol: small open question: what id format should we use for paths? I see some use camel case while others don't. `moment-timezone` or `moment-timeZone`. Same question for `moment-jdateformatparser`.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-ui/src/main/resources/XWiki/Lightbox/WebHome.xml
michitux: Note that uppercase letters have no meaning in HTML attributes (they are case-insensitive). You should use `data-user-time-zone` instead, which can also be requested in `jQuery` using `.data('userTimeZone')` or using `.dataset.userTimeZone` on a plain JavaScript object, see the [documentation for the dataset API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset). Also, this is missing XML escaping for both the date format and the timezone.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -93,8 +94,11 @@ define('xwiki-lightbox-description', [
     }
 
     if (attachmentData.date) {
+      // This should be updated after fixing XWIKI-19808: Proper javascript API to perform Date Formatting.
       var dateFormat = moment().toMomentFormatString($('.lightboxDescription .date').attr('data-dateFormat'));
-      $('.lightboxDescription .date').text(l10n.get('date', moment(attachmentData.date).format(dateFormat)));
+      var timezone = $('.lightboxDescription .date').attr('data-userTimezone');
+      $('.lightboxDescription .date')
+        .text(l10n.get('date', moment(attachmentData.date).tz(timezone).format(dateFormat)));
```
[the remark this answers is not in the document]
michitux: This conversion to moment format strings seems to be from here: https://www.npmjs.com/package/moment-jdateformatparser/v/0.1.2 - though this library seems largely unmaintained and has bugs: https://github.com/MadMG/moment-jdateformatparser/issues/18 A more recent library is available [here](https://www.npmjs.com/package/@rangerrick/moment-javaformat) but I guess due to the API breakages this would cause it isn't possible for us to migrate, we could just use this in addition? Anyways, I think this clearly shows the need for implementing our own date formatting library in JavaScript that encapsulates all this logic ASAP to make sure that we can fix such issues/migrate to fixed libraries easily without having to change every usage.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/javascript.vm
oanalavinia: I am not sure what you mean by hardcoded.
If it were in contrib, then I would have added a separated `require.config({ paths: {'moment-timezone':  ...} })` on my script. But here I added it to the require package, just like it's done for `moment-jdateformatparser`

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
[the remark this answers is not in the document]
oanalavinia: On the lightbox code I think the explicit conversion is not necessarily given that the `dateFormat` attribute should not be empty since a default value is provided here https://github.com/xwiki/xwiki-platform/blob/4ec93d34735e43489e7c10b358be58a082abfd33/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-ui/src/main/resources/XWiki/Lightbox/WebHome.xml#L407
For the LD code I am not sure where is `this.config.dateFormat` initialized.

Probably switching to a new library should be taken into consideration while working on https://jira.xwiki.org/browse/XWIKI-19808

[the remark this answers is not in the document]
michitux: No, this explicit conversion is not for the `null` case. This is for the fact that the placeholders that are supported in the format string in Java aren't the same as the ones in moment. The library `moment-jdateformatparser` provides a conversion between the two format string variants.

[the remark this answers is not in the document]
oanalavinia: Oh, I understood now your question. This conversion is done 2 lines above at https://github.com/xwiki/xwiki-platform/blob/4ec93d34735e43489e7c10b358be58a082abfd33/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js#L98

PR #1857, opened 2022-06-16
XWIKI-19806: BridgeAuthorizationManager errors when deleting a subwiki  which cause errors when creating another subwiki

surli:
* Prevent the DefaultSecurityEntryReader to throw an exception if the
    wiki does not exist anymore: just handle the exception as if the
returned owner is null, as it seems to have been planned for

### DIFF
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-bridge/src/main/java/org/xwiki/security/authorization/internal/DefaultSecurityEntryReader.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-bridge/src/main/java/org/xwiki/security/authorization/internal/DefaultSecurityEntryReader.java
@@ -255,14 +255,18 @@ public class DefaultSecurityEntryReader implements SecurityEntryReader
      * @param wikiReference the wiki to look for owner
      * @return a reference to the owner of the wiki or {@code null}
      */
-    private DocumentReference getWikiOwner(WikiReference wikiReference)
+    private DocumentReference getWikiOwner(WikiReference wikiReference) throws AuthorizationException
     {
         XWikiContext context = this.xcontextProvider.get();
-        String wikiOwner;
+        String wikiOwner = null;
         try {
             wikiOwner = context.getWiki().getWikiOwner(wikiReference.getName(), context);
         } catch (XWikiException e) {
-            wikiOwner = null;
+            if (e.getModule() != XWikiException.MODULE_XWIKI
+                || e.getCode() != XWikiException.ERROR_XWIKI_DOES_NOT_EXIST) {
+                throw new AuthorizationException(wikiReference,
+                    String.format("Could not retrieve the owner of the wiki [%s]", wikiReference.getName()), e);
+            }
         }
 
         if (wikiOwner == null) {
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-bridge/src/main/java/org/xwiki/security/authorization/internal/DefaultSecurityEntryReader.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-bridge/src/main/java/org/xwiki/security/authorization/internal/DefaultSecurityEntryReader.java
@@ -253,17 +253,16 @@ public class DefaultSecurityEntryReader implements SecurityEntryReader
 
     /**
      * @param wikiReference the wiki to look for owner
-     * @return a reference to the owner of the wiki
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-bridge/src/main/java/org/xwiki/security/authorization/internal/DefaultSecurityEntryReader.java
```diff
@@ -253,17 +253,16 @@ private XWikiDocument getDocument(DocumentReference documentReference) throws Au
...
-    private DocumentReference getWikiOwner(WikiReference wikiReference) throws AuthorizationException
+    private DocumentReference getWikiOwner(WikiReference wikiReference)
     {
         XWikiContext context = this.xcontextProvider.get();
         String wikiOwner;
         try {
             wikiOwner = context.getWiki().getWikiOwner(wikiReference.getName(), context);
         } catch (XWikiException e) {
-            throw new AuthorizationException(wikiReference, "Could not retrieve the owner of this wiki", e);
+            wikiOwner = null;
```
tmortagne, line 265: An XWikiException does not always mean that the wiki does not exist, would be more accurate to check if the wiki exist before calling getWikiOwner.

But the best would probably actually to instead use WikiDescriptorManager#getById (after making sure that getById return null and not an exception when the wiki does not exist).

surli: > WikiDescriptorManager#getById (after making sure that getById return null and not an exception when the wiki does not exist).

So it does indeed return null, now it could throw an exception if the document containing the XWikiServerClass cannot be loaded for some reason. Could it create problems if there's wrong rights on this document?

tmortagne: > now it could throw an exception if the document containing the XWikiServerClass cannot be loaded for some reason

And as I said, it's OK, we do want to know when an important document cannot be loaded for an unexpected reason. The problem of your current proposal is that you are hiding those cases.

> Could it create problems if there's wrong rights on this document?

Not sure what you mean.

surli: > >Could it create problems if there's wrong rights on this document?
> Not sure what you mean.

Forget it, was doing a mistake

surli: > But the best would probably actually to instead use WikiDescriptorManager#getById (after making sure that getById return null and not an exception when the wiki does not exist).

So FTR as discussed on the chat this solution cannot be used because the WikiDescriptorCache is also invalidated with a listener and we cannot guarantee which listener is called first.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-bridge/src/main/java/org/xwiki/security/authorization/internal/DefaultSecurityEntryReader.java
```diff
@@ -253,17 +253,20 @@ private XWikiDocument getDocument(DocumentReference documentReference) throws Au
...
     {
         XWikiContext context = this.xcontextProvider.get();
-        String wikiOwner;
+        String wikiOwner = null;
         try {
             wikiOwner = context.getWiki().getWikiOwner(wikiReference.getName(), context);
         } catch (XWikiException e) {
-            throw new AuthorizationException(wikiReference, "Could not retrieve the owner of this wiki", e);
+            if (e.getModule() != XWikiException.MODULE_XWIKI
+                || e.getCode() != XWikiException.ERROR_XWIKI_DOES_NOT_EXIST) {
```
tmortagne, line 266: Isn't this enough ?

surli: you mean only checking the code and not both module and code? Well probably, but it doesn't hurt to check both I guess?

PR #1869, opened 2022-06-23
XWIKI-19879: Update TOC macro DOM root element to make it easier to style

slauriere:
Wrap TOC in a div with class="xtoc"

Does it feel acceptable? Should the "xtoc" value not be hardcoded at that level? Any other recommendation?


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/XWikiTocMacro.java
```diff
@@ -39,11 +47,28 @@
...
         super(XWikiTocMacroParameters.class);
     }
+
+    @Override
+    public List<Block> execute(XWikiTocMacroParameters parameters, String content, MacroTransformationContext context)
+        throws MacroExecutionException
+    {
+        List<Block> blocks = super.execute(parameters, content, context);
+        GroupBlock wrapper = new GroupBlock();
+        wrapper.setParameter(CLASS, "xtoc");
```
vmassol, line 70: Could be nice to have a comment to explain why we need a GroupBlock + class attribute here.

slauriere: Thanks @vmassol. I'm adding a link to the [discussion](https://forum.xwiki.org/t/improving-the-html-structure-of-toc-macro-output/10534/4] where `xtoc` was proposed by @lucaa. I see that the LiveData component uses CSS class `xwiki-livedata`. We have also `xwiki-selectize`... What about `xwiki-toc`?  (it seems the  `x` prefix is mostly for forms)

tmortagne: Yes, this would be a new API.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/XWikiTocMacro.java
```diff
@@ -39,11 +47,28 @@
 @Singleton
 public class XWikiTocMacro extends AbstractTocMacro<XWikiTocMacroParameters>
```
tmortagne, line 48: I'm wondering if it would not be cleaner to add support for a `class` parameter in `TocMacroParameters` (null by default, which does not add any group) and make it `xwiki-toc` (or whatever else we decide) by default in `XWikiTocMacroParameters`.

vmassol: I'm not sure. It would be cleaner in term of cleanly exposing an "API" (since parameters are API). OTOH, it makes it harder to style or have an extension to use it since it's the user who will decide the name of the class to use (i.e. it's no longer static but dynamic).

vmassol: Yes, that's my point: if the user customizes it, it'll break the styling/extension since they won't be able to dynamically get the value easily.

tmortagne: We could perfectly decide in XWikiTocMacroParameters to always make sure the class contains `xwiki-toc` too, it does not really change much to what I proposed. My point was more to make this concept if group around the toc more generic than just in XWikiTocMacro.

tmortagne: > If we always set xwiki-toc

We don't set it we add it if it's not there already, you can have several values.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/XWikiTocMacro.java
slauriere: XWikiTocMacro is located in the generic package `org.xwiki.rendering.internal.macro.toc` but in an xwiki-platform project `xwiki-platform-rendering-macros` so I'm not sure if it is expected to be used within XWiki only or it may be used outside XWiki. I'm asking because this has an impact on the class name we choose. If XWiki only we could go for `xwiki-toc`. Otherwise we may consider `wikitoc`, taking example on the classes used for links (`wikilink`, `wikicreatelink`) etc.  Cc @lucaa with whom I discussed the topic separately (thanks). Wdyt?

slauriere: I've added a comment on why a GroupBlock is needed with a class attribute, using `xwiki-toc` since my understanding is that XWikiTocMacro is meant to be used in XS only (and several XWiki DOM elements such LiveData use the CSS class `xwiki-livedata` naming scheme), as opposed to TocMacro which is generic.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/XWikiTocMacro.java
slauriere: For now I've left the concept of group around the toc at XWikiMacro level only since it covers the use case I'm targeting but I understand a more generic approach could also be relevant, nothing prevents from implementing it at a later stage.

slauriere: `super.execute(parameters, content, context);` returns a list of blocks. In practice I think it contains just a single UnorderedList block that we could use as the main container but the API does not guarantee there is no other block, so my understanding is that we have to introduce a GroupBlock indeed.

michitux (approved):
I'm not really happy with the non-semantic `<div>`-element as there is a really good and semantic element `<nav>` available and by introducing the `<div>` we make it an API. However, I see the complexity of properly labeling the `<nav>` and therefore would suggest to just merge this as-is and maybe add a wrapping `<nav>` at some point that could then also wrap a heading for the table of contents.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-toc/src/main/java/org/xwiki/rendering/internal/macro/toc/XWikiTocMacro.java
michitux: I think it would be quite an API breakage to return something else than just an unordered list (as it would change the resulting HTML) and you could always search for the first unordered list in the returned blocks to catch any potentially wrapping metadata blocks. My understanding was that for some reason you needed another wrapping HTML element for the styling, but if that's not the case maybe just adding the class to the list would be better in terms of both backwards-compatibility and making the potential future addition of a `<nav>`-element easier.

michitux: Note also that any update to the parent class will always happen together with a release of this class so we could always adapt the code when necessary. To be sure we'll notice when this is the case, a test should be added to test the desired behavior (a test would be good, anyways).

vmassol: @slauriere Based on what was said, it's probably best to add the class parameter in `TocTreeBuilder`, on the top level block returned (i.e. in `xwiki-rendering`). It's an internal classes and thus can be modified. We could even pass the class name to the constructor if we want to make it more generic (but I don't think it's needed right now as I don't see any other usage of `TocTreeBuilder`). WDYT?

xwikiorgci: Good remark from @manuelleduc : 

>  the toc macro is overridden by `application-numbered-content-headings-toc`, so be aware that your work on https://github.com/xwiki/xwiki-platform/pull/1869 must also be duplicated on this extension as well if we want our users to be able to use both numbering and the new styling feature

slauriere: @vmassol this sounds good. This means adding `tocBlock.setParameter("class", "wikitoc");` to TocTreeBuilder#generateTree before returning the block, is that correct?

slauriere: I created https://github.com/xwiki/xwiki-rendering/pull/232 in `xwiki-rendering`

michitux:
Now obsolete as xwiki/xwiki-rendering#232 has been merged.

PR #1871, opened 2022-06-23
XWIKI-19756: Add ability to configure globally the default frequency used for email notifications

slauriere:
* Add getInterval methods to interface NotificationEmailUserPreferenceManager
* Implement new methods with Enum generics and refactor getDiffType to
  use the new generic method
* Update NotificationScriptService to save interval at user or wiki
  level
* Add new methods to NotificationPreferenceScriptService
* Update IntervalUserManager to use the preferred interval
* Update NotificationsEmailPreferenceMacro to display the preferred
  interval
* Declare revapi differences in pom

Probably not ready for a merge yet but a review would be great.


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/pom.xml
```diff
@@ -65,4 +65,44 @@
...
+        <groupId>org.revapi</groupId>
+        <artifactId>revapi-maven-plugin</artifactId>
+        <configuration>
+          <analysisConfiguration>
+            <revapi.differences>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.addedToInterface</code>
+                  <new>method org.xwiki.notifications.preferences.email.NotificationEmailDiffType org.xwiki.notifications.preferences.email.NotificationEmailUserPreferenceManager::getDiffType(org.xwiki.model.reference.DocumentReference)</new>
```
manuelleduc, line 80: These breaking changes can be avoided by making the new methods `default` in `NotificationEmailUserPreferenceManager`. 

See https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility

Also, will this work be applied on 14.6RC1+ too?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
```diff
@@ -40,4 +42,31 @@
      * @return the diff type configured for the given user
      */
     NotificationEmailDiffType getDiffType(String userId);
+
+    /**
+     * @param userReference reference of a user
+     * @return the diff type configured for the given user
+     * @since 14.4.2
+     */
+    NotificationEmailDiffType getDiffType(DocumentReference userReference);
```
surli, line 51: Would be better in new APIs to rely on UserReference.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
```diff
@@ -40,4 +42,31 @@
...
+     * @since 14.4.2
+     */
+    NotificationEmailInterval getInterval(String userId);
+
+    /**
+     * @param userReference reference of a user
+     * @return the notification interval configured for the given user
+     * @since 14.4.2
+     */
+    NotificationEmailInterval getInterval(DocumentReference userReference);
```
vmassol, line 71: Some remarks/questions:
1) cannot you use default methods to not break backward compatibility? Including raising a runtime exception if it cannot be implemented with a default (that's unless the feature is critical in which case there's no solution)
2) I don't think it's the best to provide several signatures, we usually try to have only one
3) As Simon said, we use `UserReference` for new APIs. Then we have bridges/APIs to get the current user reference and to convert from string to user reference.
4) We don't break backward compatibility and in the few cases where there's no choice and we still want to do it, we need to ask and have a vote. See https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/pom.xml
```diff
@@ -65,4 +65,44 @@
...
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.addedToInterface</code>
+                  <new>method org.xwiki.notifications.preferences.NotificationEmailInterval org.xwiki.notifications.preferences.email.NotificationEmailUserPreferenceManager::getInterval(org.xwiki.model.reference.DocumentReference)</new>
+                  <justification>This change is necessary to implement XWIKI-19756</justification>
+                </item>
+              </differences>
+            </revapi.differences>
+          </analysisConfiguration>
+        </configuration>
```
vmassol, line 104: Also note that this not setting a criticality (see https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HRevapiIgnores).

But in any case this needs to be discussed/voted if backward compatibility is broken.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
slauriere: I committed a new version using UserReference and with the `default` keyword for new methods but the PR still requires some updates.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
slauriere: Thanks for the review. The latest commit brings the following improvements:
* It uses `default` methods
* It uses `UserReference`

I'll see what I can do with single method signature and backward compatibility. Also working on XWiki.Notifications.Code.NotificationsEmailPreferencesMacro since the one in this PR is wrong.

slauriere: The usage of `UserReference` requires updating also `org.xwiki.notifications.notifiers.internal.email.IntervalUsersManager` to use `UserReference` instead of `DocumentReference`, which has not been done yet.

slauriere: The latest commit restores usage of `DocumentReference` instead of `UserReference` since I noticed this has an impact on `IntervalUserManager` and I'm not sure if we want to update this class as well (might be ok since it's not an interface though) nor how to proceed to make the conversion in that one in case we go for it. I kept only one method signature for `getDiffType(reference)` and `getInterval(reference)`, I updated `NotificationsEmailPreferencesMacro` and made some manual tests.

slauriere: @vmassol Please let me know if you have any advice / recommendation on how to keep improving the PR until passes, but don't hesitate to take over as I'm not sure the next steps are entirely within my reach! Happy to discuss and to try helping though.

slauriere: Just fixed an issue in `NotificationsEmailPreferencesMacro` to make sure the interval select option values are in upper case, since they are in upper case in enum `NotificationEmailInterval`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
```diff
@@ -36,8 +38,31 @@
     NotificationEmailDiffType getDiffType();
 
     /**
-     * @param userId id of a user
+     * @param userReference reference of a user
      * @return the diff type configured for the given user
+     * @since 14.10
      */
-    NotificationEmailDiffType getDiffType(String userId);
+    default NotificationEmailDiffType getDiffType(DocumentReference userReference)
```
surli, line 45: You're still not using the `UserReference` type here... See: https://extensions.xwiki.org/xwiki/bin/view/Extension/User%20Module/API/#HBridge

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
```diff
@@ -36,8 +38,31 @@
     NotificationEmailDiffType getDiffType();
 
     /**
-     * @param userId id of a user
+     * @param userReference reference of a user
      * @return the diff type configured for the given user
+     * @since 14.10
      */
-    NotificationEmailDiffType getDiffType(String userId);
+    default NotificationEmailDiffType getDiffType(UserReference userReference)
```
surli, line 45: Missing `@Unstable` annotations in those new APIs.

slauriere: Right, I added `@Unstable`. Revapi is complaining about `getDiffType(String)` conversion to `getDiffType(UserReference)`, wondering what we should do about it.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/script/NotificationPreferenceScriptService.java
```diff
@@ -292,7 +297,26 @@ public NotificationEmailDiffType getDiffType()
...
+    {
+        return emailUserPreferenceManager.getInterval();
+    }
+
+    /**
+     * @param userId id of a user
+     * @return the notification email interval configured for the given user
+     * @since 14.4.2
+     */
+    public NotificationEmailInterval getInterval(String userId)
```
surli, line 317: Normally you can directly use a `UserReference` as parameter there should be an automatic conversion String -> UserReference.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/script/NotificationPreferenceScriptService.java
```diff
@@ -292,7 +297,26 @@ public NotificationEmailDiffType getDiffType()
      */
     public NotificationEmailDiffType getDiffType(String userId)
     {
-        return emailUserPreferenceManager.getDiffType(userId);
+        return emailUserPreferenceManager.getDiffType(userReferenceResolver.resolve(userId));
+    }
+
+    /**
+     * @return the email notification interval configured for the current user
+     * @since 14.4.2
```
surli, line 305: You forgot to change the since. And also missing Unstable there.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/script/NotificationPreferenceScriptService.java
slauriere: thx done

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/script/NotificationPreferenceScriptService.java
slauriere: thx done

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/internal/email/DefaultNotificationEmailUserPreferenceManager.java
```diff
@@ -77,49 +84,67 @@ public class DefaultNotificationEmailUserPreferenceManager implements Notificati
...
             // Get the config of the user
             DocumentReference emailClassReference = new DocumentReference(EMAIL_PREFERENCES_CLASS,
-                    user.getWikiReference());
-            Object diffType = documentAccessBridge.getProperty(user, emailClassReference, DIFF_TYPE);
-            if (diffType != null && StringUtils.isNotBlank((String) diffType)) {
-                return NotificationEmailDiffType.valueOf((String) diffType);
+                userDocumentReference.getWikiReference());
+            Object value = documentAccessBridge.getProperty(userDocumentReference, emailClassReference, propertyName);
+            if (value != null && StringUtils.isNotBlank((String) value)) {
+                return Enum.valueOf(propertyEnum, ((String) value).toUpperCase());
```
surli, line 122: Minor comment but would be cleaner to have a single return value probably.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/test/java/org/xwiki/notifications/preferences/internal/email/DefaultNotificationEmailUserPreferenceManagerTest.java
```diff
@@ -91,14 +93,16 @@ public void getDiffType() throws Exception
...
 
         // Value from the user's wiki config
         when(documentAccessBridge.getProperty(new DocumentReference("someWiki", getCodeSpace(),
                         "NotificationAdministration"), new DocumentReference("mainWiki",
                         getCodeSpace(), "NotificationEmailPreferenceClass"),
                 "diffType")).thenReturn("NOTHING");
 
-        assertEquals(NotificationEmailDiffType.NOTHING, mocker.getComponentUnderTest().getDiffType(user2Id));
+        assertEquals(NotificationEmailDiffType.NOTHING,
+            mocker.getComponentUnderTest().getDiffType(userReferenceResolver.resolve(user2Id)));
```
surli, line 106: Sounds like you haven't added tests for the new `getInterval`, I understand you're in a rush but have you at least ran the module with quality profile (`mvn clean install -Pquality`) to ensure the coverage is  still good?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsEmailPreferencesMacro.xml
```diff
@@ -440,33 +440,32 @@
...
-              &lt;option value="live" #if($interval == 'live')selected="selected"#end&gt;$escapetool.xml($services.localization.render('XWiki.Notifications.Code.NotificationEmailPreferenceClass_interval_live'))&lt;/option&gt;
+              &lt;option value="HOURLY" #if($interval == 'HOURLY')selected="selected"#end&gt;$escapetool.xml($services.localization.render('XWiki.Notifications.Code.NotificationEmailPreferenceClass_interval_hourly'))&lt;/option&gt;
+              &lt;option value="DAILY"  #if($interval == 'DAILY')selected="selected"#end&gt;$escapetool.xml($services.localization.render('XWiki.Notifications.Code.NotificationEmailPreferenceClass_interval_daily'))&lt;/option&gt;
+              &lt;option value="WEEKLY" #if($interval == 'WEEKLY')selected="selected"#end&gt;$escapetool.xml($services.localization.render('XWiki.Notifications.Code.NotificationEmailPreferenceClass_interval_weekly'))&lt;/option&gt;
+              &lt;option value="LIVE" #if($interval == 'LIVE')selected="selected"#end&gt;$escapetool.xml($services.localization.render('XWiki.Notifications.Code.NotificationEmailPreferenceClass_interval_live'))&lt;/option&gt;
```
surli, line 465: Sounds like it could have been refactored with a foreach loop

surli: Hmmm, actually does it mean there's a bug right now and the interval is not properly selected in the UI? Sounds a bit surprising

slauriere: There was no bug: the interval was stored in lower case in the user profile, and was then converted to upper case by the IntervalUsersManager (reference: [IntervalUsersManager.java#L130](https://github.com/xwiki/xwiki-platform/blob/101becab4069ae9ccdc8c2f2c4edc562645e152b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-api/src/main/java/org/xwiki/notifications/notifiers/internal/email/IntervalUsersManager.java#L130)) but was always retrieved as is from the UI (no case conversion). Now the API `NotificationEmailUserPreferenceManager#getInterval` returns only upper case values, the goal here is to align the case with what the API returns (same approach as what we have with the diffType option values).

surli: ok I understand the idea, but not entirely comfortable with it: it means that an upgraded wiki might have both lower case and upper case stored interval values, and it might possibly break some external code that are reading those values directly from the xobject. 

I agree that the fact the case of the stored data and what's returned by the API should be same, but I'm afraid such change would be possibly breaking and would need some kind of migrations. To be honest I think I'd have been ok with that change if it wouldn't be for 14.10. 

Would need opinion of other committers here, @manuelleduc  @vmassol wdyt?

slauriere: I brought more changes:
- Converted interval value to upper case for existing instances where it would be stored in lower case in an xobject.
- Added foreach loop.
- Added a test case (but cannot run the test suite successfully on my side for now (see error below, any help welcome).
- Removed unncessary test on null value already covered by StringUtils.isNotBlank.

```
org.xwiki.component.manager.ComponentLookupException: Can't find descriptor for the component with type [org.xwiki.model.reference.DocumentReferenceResolver<java.lang.String>] and hint [null]
	at org.xwiki.component.embed.EmbeddableComponentManager.getInstance(EmbeddableComponentManager.java:212)
	at org.xwiki.component.embed.EmbeddableComponentManager.getInstance(EmbeddableComponentManager.java:189)
	at org.xwiki.notifications.preferences.internal.email.DefaultNotificationEmailUserPreferenceManagerTest.setUp(DefaultNotificationEmailUserPreferenceManagerTest.java:62)
```

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/internal/email/DefaultNotificationEmailUserPreferenceManager.java
slauriere: Thanks but not sure what you mean here. Do you mean introducing `T returnValue = propertyDefaultValue;` at the beginning, and reassigning it in `if {...} else if {...} else {...}` blocks with just a final `return` statement, is that right?

surli: Yes that's what I mean.

surli: Note that it's not something we documented as best practice to my knowledge but we generally try to adopt this style when possible.

slauriere: ok thx done this

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/test/java/org/xwiki/notifications/preferences/internal/email/DefaultNotificationEmailUserPreferenceManagerTest.java
slauriere: Sounds good now that you have refactored the test :+1:

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
surli: You should have kept `getDiffType(String)` but mark it as deprecated. And implement it by resolving the String to a UserReference.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/script/NotificationPreferenceScriptService.java
```diff
@@ -286,13 +287,35 @@ public NotificationEmailDiffType getDiffType()
...
+    public NotificationEmailDiffType getDiffType(UserReference user)
+    {
+        return emailUserPreferenceManager.getDiffType(user);
+    }
+
+    /**
+     * @return the email notification interval configured for the current user
+     * @since 14.10
      */
-    public NotificationEmailDiffType getDiffType(String userId)
```
surli, line 293: So same here: you should have kept that one but deprecate it.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
```diff
@@ -38,6 +41,39 @@
     /**
      * @param userId id of a user
      * @return the diff type configured for the given user
+     * @deprecated use {@link #getDiffType(UserReference)} instead
      */
     NotificationEmailDiffType getDiffType(String userId);
```
surli, line 46: You need to also put a `@Deprecated(since="14.10)` annotation on the method.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/internal/email/DefaultNotificationEmailUserPreferenceManager.java
```diff
@@ -77,49 +84,67 @@ public class DefaultNotificationEmailUserPreferenceManager implements Notificati
...
+                if (StringUtils.isNotBlank((String) value)) {
+                    returnValue = Enum.valueOf(propertyEnum, ((String) value).toUpperCase());
+                } else {
+                    // Get the config of the main wiki
+                    WikiReference mainWiki = new WikiReference(wikiDescriptorManager.getMainWikiId());
+                    if (!userDocumentReference.getWikiReference().equals(mainWiki)) {
+                        xwikiPref = new DocumentReference(GLOBAL_PREFERENCES, mainWiki);
+                        emailClassReference = new DocumentReference(EMAIL_PREFERENCES_CLASS, mainWiki);
+                        value = documentAccessBridge.getProperty(xwikiPref, emailClassReference, propertyName);
+                        if (StringUtils.isNotBlank((String) value)) {
```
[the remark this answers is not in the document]
surli: Easiest way in general is just to extract some nested if in a private method: basically put the else in a private method and you should be ok.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/script/NotificationPreferenceScriptService.java
surli: Same here you're missing the `@Deprecated` annotation on the method (besides the javadoc one)

surli: Sounds like you misunderstood me, there's two different things to have: a `@deprecated` javadoc documentation tag, and a Java `@Deprecated` annotation that is used for our tooling. See for example: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiStoreInterface.java#L111

slauriere: Right, sorry for the misunderstanding, I knew there was something wrong, I searched for "@deprecated" lower case only. Fixed it. Many thanks for your guidance.

surli (approved):

manuelleduc (commented):
Is it planned to move the deprecated methods to a legacy extension?

surli:
> Is it planned to move the deprecated methods to a legacy extension?

I'd say better do that when the new API is stable no?

tmortagne:
> Is it planned to move the deprecated methods to a legacy extension?

I don't think that's a good idea, actually. It cannot really work the same as for XWiki Standard where we package the legacy dependencies. If you don't depend on a legacy extension it will never be installed so if you remove anything from it, it's removed and you broke retro compatibility, the fact that you moved it to some other extension that happen to have "legacy" in its name does not really matter.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
```diff
@@ -38,6 +41,40 @@
...
+
+    /**
+     * @param userReference reference of a user
+     * @return the diff type configured for the given user
+     * @since 14.10
+     */
+    @Unstable
+    default NotificationEmailDiffType getDiffType(UserReference userReference)
+    {
+        return getDiffType();
```
michitux, line 57: Is this really a good default implementation? Wouldn't it be better to throw some exception instead of returning a value that is not for the given user to just throw an exception? The same also applies to the other two new methods.

In particular, when you removed the implementation of this method from the default implementation, you would get an endless loop.

michitux (commented):
Apart from the default implementations in the interface, this looks good to me. Are there any integration tests for this feature that could/should be updated?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-preferences/xwiki-platform-notifications-preferences-api/src/main/java/org/xwiki/notifications/preferences/email/NotificationEmailUserPreferenceManager.java
tmortagne: The only point of the default implementation is retro-compatibility (in case this interface is implemented by an extension based on an older version).

> In particular, when you removed the implementation of this method from the default implementation, you would get an endless loop.

But doing that would not really make any sense.

michitux: Okay, I just wasn't sure if it was okay to return possibly incorrect values from the default implementation.

surli:
> Are there any integration tests for this feature that could/should be updated?

I don't recall any integration tests related to this feature as it's difficult to test the email interval scenarios (like testing daily mails vs live emails).

michitux:
> I don't recall any integration tests related to this feature as it's difficult to test the email interval scenarios (like testing daily mails vs live emails).

I meant more for the notification settings.

PR #1873, opened 2022-06-30
XWIKI-19902: Potential error log caused by not existing user during R140401000XWIKI15460 migration

manuelleduc:
- Add a control for the existence of the user's wiki on org.xwiki.user.internal.document.DocumentUserManager.exists(org.xwiki.user.UserReference)
- When removing user's notifications in R140401000XWIKI15460DataMigration, verify if their is still notifications to remove before removing unknown user's notifications

Jira: https://jira.xwiki.org/browse/XWIKI-19902

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140401000XWIKI15460DataMigration.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140401000XWIKI15460DataMigration.java
@@ -26,6 +26,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
@@ -41,6 +42,7 @@ import org.xwiki.notifications.filters.internal.DefaultNotificationFilterPrefere
 import org.xwiki.notifications.filters.internal.NotificationFilterPreferenceConfiguration;
 import org.xwiki.notifications.filters.internal.NotificationFilterPreferenceStore;
 import org.xwiki.stability.Unstable;
+import org.xwiki.user.UserException;
 import org.xwiki.user.UserManager;
 import org.xwiki.user.UserReference;
 import org.xwiki.user.UserReferenceResolver;
@@ -196,16 +198,30 @@ public class R140401000XWIKI15460DataMigration extends AbstractHibernateDataMigr
     }
 
     private void identifyRemovedUsers(Map<String, Boolean> usersStatus,
-        DefaultNotificationFilterPreference filterPreference)
+        DefaultNotificationFilterPreference filterPreference) throws DataMigrationException
     {
         String owner = filterPreference.getOwner();
+
+        // Store the potential exception thrown inside computeIfAbsent, since assigning to a variable is not allowed.
+        AtomicReference<Exception> exception = new AtomicReference<>();
         usersStatus.computeIfAbsent(owner, key -> {
             DocumentReference entityReference = this.resolver.resolve(key);
             UserReference userReference =
                 this.documentReferenceUserReferenceResolver.resolve(entityReference);
 
-            return this.userManager.exists(userReference);
+            try {
+                return this.userManager.exists(userReference);
+            } catch (UserException e) {
+                exception.set(e);
+                return null;
+            }
         });
+
+        // If the exception store has been set, propagate the exception to make the migration fail.
+        if (exception.get() != null) {
+            throw new DataMigrationException(
+                String.format("Failed to identify if the owner of [%s] exists.", filterPreference), exception.get());
+        }
     }
 
     private boolean isMainWiki()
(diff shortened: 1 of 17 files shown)


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/document/DocumentUserManager.java
```diff
@@ -56,24 +59,47 @@ public class DocumentUserManager implements UserManager
...
         DocumentReference userDocumentReference = ((DocumentUserReference) userReference).getReference();
+        String userWikiId = userDocumentReference.getWikiReference().getName();
+        try {
+            if (!this.wikiDescriptorManager.exists(userWikiId)) {
+                result = false;
+            } else {
+                result = exists(xcontext, xwiki, userDocumentReference);
+            }
+        } catch (WikiManagerException e) {
+            this.logger.warn("Failed to determine if wiki [{}] exists. Considering it's not the case. Cause: [{}]",
```
tmortagne, line 85: I'm not a huge fan of this logic, but you did not really introduce this concept and unfortunately #exists was declared as not throwing any exception (which is what should happen here).

surli: I don't know much what could cause such exception in the wiki manager, but for me it makes sense to handle it like that here. Not sure exists should actually throw it: if you have a problem there, it probably means there would be other issues later if you tried to use the user. And it's probably not the job of the consumer of this API to handle the problem caused by the wiki manager. So all in all, it sounds like a good compromise to consider the user just doesn't exist in such case.

manuelleduc: otoh, in case of initialization issue you might wrongfully indicate to the caller that the user does not exist. In the context of `R140401000XWIKI15460` the consequence is the removal of all the user's notification preferences, which is pretty bad.
I think propagating the exception is the safest thing to do here.

tmortagne: The whole point is that an API like this should not lie when it does not know the answer (and `false` is a lie here) because it's a very common cause of cache mess and that's really not the kind of problem you want to have when talking about users.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/UserEventDispatcher.java
```diff
@@ -280,15 +281,15 @@ public void dispatch(Event event)
         }
     }
 
-    private void dispatchInContext(Event event)
+    private void dispatchInContext(Event event) throws UserException
```
tmortagne, line 284: I'm wondering if it's valid to fail the entire pre-filtering if one of the user profile page is broken.

Problem is that it all depend what is the cause:
* general problem: should definitely stop (i.e. not mark the event as pre-filtered)
* only one of the users have a broken document: quite a pain to penalize all following users

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-api/src/main/java/org/xwiki/user/UserException.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @version $Id$
+ * @since 14.6RC1
+ * @since 14.4.3
+ * @since 13.10.8
+ */
+@Unstable
+public class UserException extends Exception
+{
+    private static final long serialVersionUID = 3091220482950706563L;
```
tmortagne, line 35: We generally set `1` (not that it matter that much, since there is almost no chance that this exception will ever be serialized).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/UserEventDispatcher.java
manuelleduc: I agree. My assumption is that the exception will only occur for very bad failures where all users are impacted. Also, if we just ignore "failed" users, we have no easy way to allow user the re-execute the migration.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-api/src/main/java/org/xwiki/user/UserException.java
tmortagne: I don't think it's a written rule, I was just under the impression that it's much more common in our code base, but I may be wrong.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R140401000XWIKI15460DataMigration.java
```diff
@@ -174,7 +172,13 @@ private void internalHibernateMigrate(boolean isMainWiki) throws DataMigrationEx
                 .map(Map.Entry::getKey)
                 .collect(Collectors.toSet());
             for (String userReference : unknownUsers) {
-                this.store.deleteFilterPreferences(this.resolver.resolve(userReference));
+                DocumentReference userDocumentReference = this.resolver.resolve(userReference);
+                // Verify if the user still has some filters preferences to remove, in case they have already all been
+                // cleaned up when removing the filters preferences for the unknown wikis. Without this check, calling
```
tmortagne, line 177: I honestly don't see the point of complexity the code so much for this HSQLDB specific nonsense.

PR #1875, opened 2022-07-04
 XWIKI-19928: Migrate CKEditor to XWiki Platform

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-19928

* CKEditor history preserved. See https://github.com/xwiki/xwiki-platform/pull/1875/commits/48fceaa6b12fb69ae2e66d31ad584635c23aacdc for my actual changes. I first merged ckeditor at the root (ignoring the conflicting files, then moved the directories to `xwiki-platform-ckeditor` and changed what needed to be adapted).

**Other tasks**
* Move the issues in Jira
* Update the documentation


michitux:
Commit 6b1053164386aefc526df7512bc664918aa6849b should also be included in this pull request (probably best to check before the merge if there were any new commits).

manuelleduc:
Note: A xar:format call on xwiki-platform-ckeditor-ui needs to be executed.

manuelleduc:
> Note: A xar:format call on xwiki-platform-ckeditor-ui needs to be executed.

done

pom.xml
```diff
@@ -1276,6 +1275,26 @@
...
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <searchTransitive>true</searchTransitive>
+                  <message>
+                    Do not use the CKEditor extension from contrib. Is it now packaged in xwiki platform.
+                  </message>
+                  <excludes>
+                    <exclude>org.xwiki.contrib:application-ckeditor*:*</exclude>
```
tmortagne, line 1292: Why the `:*` ?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/pom.xml
```diff
@@ -0,0 +1,262 @@
...
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-tree-webjar</artifactId>
+      <version>${project.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <!-- This is used by the resource suggest picker on the link modal. -->
+    <dependency>
+      <groupId>org.webjars.npm</groupId>
+      <artifactId>bootstrap-3-typeahead</artifactId>
+      <version>4.0.2</version>
```
tmortagne, line 67: Maybe we should define the version of that dependency along with the other bootstrap related dependencies in the root platform pom.

manuelleduc: Do we define a variable for every external dependency, even if there is only a single occurrence?

pom.xml
tmortagne: My point is more that it does not seem needed, `org.xwiki.contrib:application-ckeditor*` should be enough, no ? We generally don't indicate the type there.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/pom.xml
tmortagne: Just suggesting to do the same as https://github.com/xwiki/xwiki-platform/blob/master/pom.xml#L427.

PR #1885, opened 2022-07-22
XWIKI-19832: Take into account the request parameters when caching a LESS SSX

slauriere:
This is an attempt to implement XWIKI-19832. Not sure this is the right approach nor if the list of excluded parameters is correct but a review and a recommendation on how to implement the feature is welcome.

Remarks:
* The list of excluded parameters was obtained empirically by observing the existing request parameters when computing the cache key of a LESS resource, there might be other parameters to exclude from the list but I don't know how to get the exhaustive list.
* A possible downside of this approach is that the request parameters are in this case always taken into account when `withContext` is `true`. However a LESS resource is not meant to receive request parameters others than the ones already used for computing the cache key and the excluded ones, is it?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-lesscss/xwiki-platform-lesscss-default/src/main/java/org/xwiki/lesscss/internal/cache/CacheKeyFactory.java
+++ b/xwiki-platform-core/xwiki-platform-lesscss/xwiki-platform-lesscss-default/src/main/java/org/xwiki/lesscss/internal/cache/CacheKeyFactory.java
@@ -19,13 +19,22 @@
  */
 package org.xwiki.lesscss.internal.cache;
 
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
 import javax.inject.Inject;
+import javax.inject.Provider;
 import javax.inject.Singleton;
 
 import org.xwiki.component.annotation.Component;
 import org.xwiki.lesscss.internal.colortheme.ColorThemeReference;
 import org.xwiki.lesscss.internal.skin.SkinReference;
 import org.xwiki.lesscss.resources.LESSResourceReference;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-lesscss/xwiki-platform-lesscss-default/src/main/java/org/xwiki/lesscss/internal/cache/CacheKeyFactory.java
```diff
@@ -62,10 +74,25 @@ public String getCacheKey(LESSResourceReference lessResourceReference, SkinRefer
...
-        
+
         if (withContext) {
             String xcontext = xcontextCacheKeyFactory.getCacheKey();
-            result += CACHE_KEY_SEPARATOR + xcontext.length() + CACHE_KEY_SEPARATOR + xcontext; 
+            result += CACHE_KEY_SEPARATOR + xcontext.length() + CACHE_KEY_SEPARATOR + xcontext;
+
+            /** Also take into account the request parameters, if any, except parameters which are already
+             * taken into account or that are irrelevant. */
+            XWikiContext context = xcontextProvider.get();
```
tmortagne, line 84: An alternative to XWikiContext is to get the request from `org.xwiki.container.Container#getRequest()` and cast it into a `org.xwiki.container.servlet.ServletRequest` to have access to the `HttpServletRequest`.

slauriere: Thank Thomas, I just updated the code accordingly.

PR #1888, opened 2022-08-10
XWIKI-20024: Replace the Livetable from the attachments tab with a LiveData

oanalavinia:
* replace the livetable with a call to the liveData macro and specify that the wiki syntax must be used
* update how filters are applied, since the empty value should be also considered
* update js code to use the liveData api
* lazy load attachments skin extensions inside docextra.vm 
* avoid a cycling dependency by moving AttachmentsPane inside the xwiki-platform-flamingo-skin module, where all the attachments pane resources are
* update AttachmentsPane code to use the LiveData instead of LiveTable
* update liveData filterColumn to support select options with different values from the displayed ones
* create a AttachmentsViewPage to support actions possible on a ViewPage when the attachments tab is active
* update of existing tests


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsjson.vm
```diff
@@ -105,7 +105,9 @@
     #if ($hasAccess)
       #set ($attachment = $doc.getAttachment($attachmentReference.name))
       #set ($authorReference = $services.model.resolveDocument($attachment.author))
+      ## The property "doc.fullName" is needed by LiveData since it is used to identify a live data entry.
```
mflorea, line 108: Well, if you use ``doc.fullName`` then all entries will have the same ID which is obviously not right. The ID should be the attachment name (if all attachments are from the same document). For this you should:
* add an ``id`` field below
* configure the ``entryDescriptor`` in the live data JSON configuration, see https://extensions.xwiki.org/xwiki/bin/view/Extension/Live%20Data%20Macro/#HLiveDataJSON

```
"entryDescriptor": {
  // The property that can be used to identify a live data entry. This is used for entry selection.
  "idProperty": "id",
},
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
```diff
@@ -216,10 +216,13 @@ require(['jquery', 'xwiki-events-bridge'], function($) {
         notification = new XWiki.widgets.Notification(l10n['core.viewers.attachments.delete.inProgress'], 'inprogress');
       },
       success : function() {
-        window[thisLivetableName].refresh();
-        if (thisLivetableName === 'docAttachments') {
-          updateCount();
-        }
+        liveData.updateEntries().then(() => {
+          const liveDataId = liveData.data.id;
+          liveData = $('#' + liveDataId).data('liveData');
```
mflorea, line 221: Is this really needed? I mean, the root element holding the reference to the live data instance shouldn't change when the list of entries is updated, and neither the live data instance (it should be the same instance).

mflorea (changes requested):

mflorea (approved):
Looks good so far.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachment_macros.vm
```diff
@@ -247,37 +246,37 @@
...
-    'url': $attachmentsDoc.getURL('get', "xpage=xpart&vm=attachmentsjson.vm"),
-    'rowCount': 5,
-    'selectedColumn': 'filename',
-    'javascriptName': $javascriptName,
-    'outputOnlyHtml': true
-  })
-  #livetable($livetableId $columns $columnsProperties $options)
-  #deleteAttachmentModal
+    'className': 'XWiki.AllAttachments',
+    "\$doc": "$attachmentsDoc",
```
manuelleduc, line 270: I think `'$doc'` would work as well, without having to escape the `$`.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/skin/test/po/AttachmentsViewPage.java
```diff
@@ -0,0 +1,57 @@
...
+import org.openqa.selenium.By;
+import org.xwiki.test.ui.po.ViewPage;
+
+/**
+ * Represents a page with the attachments pane active.
+ *
+ * @version $Id$
+ * @since 14.7RC1
+ */
+public class AttachmentsViewPage extends ViewPage
```
manuelleduc, line 31: How is this class different from `org.xwiki.test.ui.po.AttachmentsPane`?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/skin/test/po/AttachmentsViewPage.java
```diff
@@ -0,0 +1,57 @@
...
+package org.xwiki.flamingo.skin.test.po;
+
+import org.openqa.selenium.By;
+import org.xwiki.test.ui.po.ViewPage;
+
+/**
+ * Represents a page with the attachments pane active.
+ *
+ * @version $Id$
+ * @since 14.7RC1
```
manuelleduc, line 29: Need to be moved to 14.8RC1 ;)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
```diff
@@ -47,6 +47,7 @@
     $xwiki.ssfx.use('uicomponents/widgets/upload.css', true)
     $xwiki.jsfx.use('js/xwiki/viewers/attachments.js', {'forceSkinAction': true, 'language': ${xcontext.locale}})
     $xwiki.ssfx.use('js/xwiki/viewers/attachments.css', true)
+    $xwiki.jsfx.use('uicomponents/widgets/liveData.js', true)
```
mflorea, line 50: Would be nice to load all the above JavaScript and CSS on demand, lazy, only when the Attachments tab is loaded, using something like https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/display.vm#L49-L50 .

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/skin/test/po/AttachmentsViewPage.java
mflorea: AFAICS it represents a page that has the Attachments tab (``AttachmentsPane``) and knows how to open / activate it, either by mouse click or by keyboard shortcut. You think those could be moved as static methods in ``AttachmentsPane``? I'm not sure it's better.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-test/xwiki-platform-livedata-test-pageobjects/src/main/java/org/xwiki/livedata/test/po/LiveDataElement.java
```diff
@@ -95,6 +95,20 @@ public List<String> getFootnotesText()
...
+     */
+    public boolean isReady()
+    {
+        // See {@link #waitUntilReady()}.
+        String[] classes = getDriver().findElementWithoutWaiting(By.id(this.id)).getAttribute("class").split("\\s+");
+        boolean isVueLoaded = !Arrays.asList(classes).contains("loading");
+        boolean areComponentsLoaded =
+            getDriver().findElement(By.id(this.id)).findElements(By.cssSelector(".xwiki-livedata .loading")).isEmpty();
+
+        return isVueLoaded && areComponentsLoaded;
```
mflorea, line 109: Wouldn't be better to add two methods, reused in both places, in order to prevent code duplication?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/SuggestInputElement.java
```diff
@@ -241,7 +241,7 @@ public SuggestInputElement selectByValue(String value)
      */
     public SuggestInputElement selectByVisibleText(String text)
     {
-        getDriver().findElementWithoutWaiting(
+        getDriver().findElement(
```
mflorea, line 244: Why wait? I think this method is supposed to be called **after** the suggestions are displayed. It assumes you already waited for the suggestions. Maybe the problem is that you didn't wait for the suggestions?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-test/xwiki-platform-livedata-test-pageobjects/src/main/java/org/xwiki/livedata/test/po/TableLayoutElement.java
```diff
@@ -450,7 +450,8 @@ public void filterColumn(String columnLabel, String content, boolean wait)
         if (classes.contains("filter-list")) {
             if (element.getAttribute(CLASS_HTML_ATTRIBUTE).contains("selectized")) {
                 SuggestInputElement suggestInputElement = new SuggestInputElement(element);
-                suggestInputElement.sendKeys(content).selectTypedText();
+                suggestInputElement.clearSelectedSuggestions();
+                suggestInputElement.sendKeys(content).selectByVisibleText(content);
```
mflorea, line 454: I think you're missing a wait for the suggestions to be fetched from the server and displayed.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/skin/test/po/AttachmentsViewPage.java
oanalavinia: I oscillated between having a new class and using static methods, I'm not sure which is the better option

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachment_macros.vm
oanalavinia: I get an `updating entries failed` error from the liveData if I don't use escaping

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
```diff
@@ -55,6 +57,8 @@
       #set ($sortAttachmentsBy = "&sortAttachmentsBy=$escapetool.url(${sortAttachmentsBy})")
     #end
     #set ($discard = $docextras.add(['Attachments', 'attachments', $services.localization.render('docextra.attachments'), $doc.getAttachmentList().size(), "attachmentsinline.vm$!{sortAttachmentsBy}", $services.localization.render('core.shortcuts.view.attachments')]))
+    #getRequiredSkinExtensions($requiredSkinExtensions)
+    #set ($discard = $response.setHeader('X-XWIKI-HTML-HEAD', $requiredSkinExtensions))
```
mflorea, line 61: I don't think this is the right place. ``docextra.vm`` is evaluated on the initial HTML request, so on page load. This needs to be put where the content is loaded asynchronously, after the initial page load, probably in ``attachmentsinline.vm``.

Initially it was hard / complex to load JavaScript / CSS from a content that is fetched and injected dynamically after the page load. The solution was to preload the JavaScript / CSS on page load, which is what ``docextra.vm`` does here: it preloads the resources needed by the Attachments tab (``attachmentsinline.vm``). The downside is that we're loading resources that may not be needed by the user (if the user doesn't open the Attachments tab), thus increasing the page load time. It's better to defer the loading until those resources are needed, i.e. until the Attachments tab is loaded. So the solution is to move the resource includes from ``docextra.vm`` to ``attachmentsinline.vm`` (or some nested template) and to use the ``X-XWIKI-HTML-HEAD`` to load them when the Attachments tab is fetched asynchronously.

oanalavinia: This is already done in `attachmentsinline.vm` https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsinline.vm#L22
Only that if I refresh the page while on the attachments tab, the resources are not loaded if they are not defined also inside  `doextra.vm`

oanalavinia: What I can see while debugging is that here https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js#L1862 the `getHeader('X-XWIKI-HTML-HEAD')` will not contain the required extensions, as if https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsinline.vm#L60 was not done (even if `$requiredSkinExtensions` wasn't empty)

mflorea (approved):

PR #1891, opened 2022-08-17
XWIKI-16192: Refactor links to support properly backlinks in a farm

tmortagne:
Still a work in progress but creating a Pull request since it starting to be close to its final form in terms of design.

Link Store:
* moved the storage of links to the Solr search core (one field to store the actual links and one to help with searching for backlinked entities)
* the links are stored in a way that allows storing any kind of link (not only entity references) but currently only entity reference are stored in practice (storing other kind of links is outside the scope of this issue but wanted to be sure the store allows doing it later without changing anything to the storage format)
* introduce a new LinkStore to access various entities links and backlinks
* deprecated all `XWikiStoreInterface` link related APIs and modified all uses of this API to use the new `LinkStore` instead

Solr extensibility
* new extension point to let any extension be called during the Solr metadata extraction

New reference serialization format including the type
* made a little more official the "<entity type>:<entity reference>" syntax that we are using here and there since a while by supporting it right in the reference resolvers
* allow serializing/parsing parameters for other types than PAGE world entities (currently only `EntityReferenceResolver<String>` and `EntityReferenceSerializer<String>` with `@Named("withparameters")`)

Misc
* improved `XWiki#getDocumentReference(EntityReference, XWikiContext)` to be a reliable way to get a `DocumentReference` from any `EntityReference` (especially for PAGE world references)


xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/listener/MovedAttachmentListener.java
```diff
@@ -94,21 +96,22 @@ public void onEvent(Event event, Object source, Object data)
                 || this.authorization.hasAccess(Right.EDIT, moveAttachmentRequest.getUserReference(), reference))
                 && (!moveAttachmentRequest.isCheckAuthorRights() || this.authorization.hasAccess(Right.EDIT,
                     moveAttachmentRequest.getAuthorReference(), reference)));
-            // TODO: Make sure to apply the refactoring at farm level for attachments as part of XWIKI-8346.
-            updateBackLinks(attachmentMovedEvent, canEdit);
+            try {
+                updateBackLinks(attachmentMovedEvent, canEdit);
+            } catch (RefactoringException e) {
+                this.logger.error("Failed to update links backlinks", e);
```
manuelleduc, line 102: Would it be interesting to log some variables from the context in addition to the exception? I'm thinking of `attachmentMovedEvent` or `moveAttachmentRequest`.

tmortagne: Cannot hurt, I guess. Now the cause of this exception is supposed to be that it failed to load the backlinks for that attachment reference and the cause exception contains the reference.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/test/java/org/xwiki/attachment/internal/listener/MovedAttachmentListenerTest.java
```diff
@@ -95,16 +95,15 @@ void onEventNoUpdateReference()
     }
 
     @Test
-    void onEvent()
+    void onEvent() throws RefactoringException
```
manuelleduc, line 98: It may be just me but I'm always throwing `Exception` and never anything more specific. This makes refactoring easier.

tmortagne: Not sure I'm a big fan of that logic, if you push it you can end up having to put `throws RefactoringException` on every single test. I also like making sure that it's excepted to have those exceptions.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractStringEntityReferenceResolver.java
```diff
@@ -115,6 +115,18 @@ private StringBuilder createStringBuilder(String entityReferenceRepresentation)
     @Override
     public EntityReference resolve(String entityReferenceRepresentation, EntityType type, Object... parameters)
     {
+        // If the type is not provided, try to extract it from the string
+        if (type == null) {
+            Character entityTypeSeparator = getSymbolScheme().getEntityTypeSeparator();
+            if (entityTypeSeparator != null) {
+                type = Enum.valueOf(EntityType.class,
```
manuelleduc, line 122: Feels strange to assign a new value to a parameter.

tmortagne: It's simply that we try to find it if it's null. I guess we could define new variable which by default have the parameters values, but I'm not sure it would really help readability in any way (might make it worst).

surli: No complain from sonar or checkstyle for that? I thought there were checks for those.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractStringEntityReferenceResolver.java
```diff
@@ -115,6 +115,18 @@ private StringBuilder createStringBuilder(String entityReferenceRepresentation)
     @Override
     public EntityReference resolve(String entityReferenceRepresentation, EntityType type, Object... parameters)
     {
+        // If the type is not provided, try to extract it from the string
+        if (type == null) {
+            Character entityTypeSeparator = getSymbolScheme().getEntityTypeSeparator();
+            if (entityTypeSeparator != null) {
```
surli, line 121: No else branch?  no risk of NPE in the remaining of  the method if type remains null? AFAIR it wasn't meant to be nullable is it?

surli: Also not sure why you test for null here since you provided a default method in the interface: the overriding implementation should never return null for this.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/test/java/org/xwiki/model/internal/reference/DefaultStringEntityReferenceSerializerTest.java
```diff
@@ -318,7 +318,7 @@ void serializePageReferencesParameters()
             resolver.resolve("wiki:page1;param1=value1;param2=value2/page2;en_US", EntityType.PAGE);
         assertEquals("wiki:page1;param1=value1;param2=value2/page2;en_US", serializer.serialize(reference));
 
-        reference = resolver.resolve("wiki:pag\\;e1;pa\\;ram1=val\\;ue1;param2=value2/pag\\;e2;en_\\;US", EntityType.PAGE);
-        assertEquals("wiki:pag\\;e1;pa\\;ram1=val\\;ue1;param2=value2/pag\\;e2;en_\\;US", serializer.serialize(reference));
+        reference = resolver.resolve("wiki:pag\\;e1;pa\\;ram1=val\\;ue1;param2=value2/pag\\;e2;en_US", EntityType.PAGE);
+        assertEquals("wiki:pag\\;e1;pa\\;ram1=val\\;ue1;param2=value2/pag\\;e2;en_US", serializer.serialize(reference));
```
surli, line 322: Why did you change that test?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
```diff
@@ -26,7 +26,9 @@
 
 /**
  * Represent the relation between a document using a link to an entity, and the used entity.
+ * @deprecated use TODO instead
  */
+@Deprecated(since = "14.6RC1")
```
surli, line 31: Need to bump the deprecated version.

surli: And the message :)

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/link/LinkStore.java
```diff
@@ -0,0 +1,54 @@
...
+public interface LinkStore
+{
+    // TODO: also introduce an API (#getLinks) to get all types of links and not only entities
+
+    /**
+     * @param reference the reference of the entity containing links
+     * @return the other (DOCUMENT based) entities linked by the passed entity
+     * @throws RefactoringException when failing to load the links
+     */
+    Set<EntityReference> resolveLinkedEntities(EntityReference reference) throws RefactoringException;
```
surli, line 46: Looks like we were using previous "load" instead of "resolve" for same API method. Feels a bit weird to have "resolve" here, but wondering more why it's not "get"?

surli:
I haven't reviewed all but I have 2 global remarks:
  1. You deprecated all APIs allowing to store links in the DB, don't you plan to still allow people to use it in the future by setting a property in the config? It sounds a bit like for notification with pre-filtering / post-filtering, I don't recall that we have deprecated all APIs when we introduced the pre-filtering. So not entirely sure about the move here. Now maybe it's different and we just don't want users to switch back to DB storage for links here?

  2. About your observation regarding `LinkStore` in refactoring: personally I have no problem of having it in the Maven module, however I do find weird that the throwing exception in this API to be a `RefactoringException`. It feels like the UC for using that API are wider than just refactoring.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractStringEntityReferenceResolver.java
tmortagne: Yes, I was surprised too, but neither checkstyle (during the Maven build) nor sonarlint (in Eclipse) are complaining about it for me.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractStringEntityReferenceResolver.java
tmortagne: > No else branch? no risk of NPE in the remaining of the method if type remains null?

Probably, but then it will behave as it used to, I just added an improvement when and entity type separator is provided. The only thing that would make sense in a else is to throw a IllegalArgumentException.

> the overriding implementation should never return null for this.

Technically, nothing prevent an implementation to do `return null`.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/test/java/org/xwiki/model/internal/reference/DefaultStringEntityReferenceSerializerTest.java
tmortagne: I fixed the parsing to have the expected type (Locale) in the parameters, and "en_;US" is an illegal syntax for a locale.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/link/LinkStore.java
tmortagne: It's not `get` because it does more than getting the stored links as is, it also makes sure to return DOCUMENT based references (so it has to resolve anything that is based on PAGE) so that you don't have to worry about that.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractStringEntityReferenceResolver.java
surli: > The only thing that would make sense in a else is to throw a IllegalArgumentException.

Would be better I guess. 

> Technically, nothing prevent an implementation to do return null.

Sure in practice, but the javadoc should be clear that it's not supposed to be nullable IMO.

tmortagne:
> Now maybe it's different and we just don't want users to switch back to DB storage for links here?

Honestly, I don't see the point. The links will keep being stored in the database to avoid breaking old things executing SQL requests mixing the link table and other tables for example, but it does not make that much sense IMO to keep adding/loading hibernate links through the API (it honestly did not make that much sense before either compared to higher level API to get links or the concept of storing links at all IMO).

> I do find weird that the throwing exception in this API to be a RefactoringException. It feels like the UC for using that API are wider than just refactoring

Then it should not be in the refactoring module :)

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractStringEntityReferenceResolver.java
tmortagne: > the javadoc should be clear that it's not supposed to be nullable IMO

Sure why not, but I still prefer to test this use case when it's technically possible :)

surli:
> Honestly, I don't see the point.

Ok so then it makes sense to deprecate those.

> Then it should not be in the refactoring module :)

Yeah I agree, I think my point was more that it's easy to move a class from a module to another in the future (not breaking), but a bit less to change the signature of a method to change the exception  (breaking)

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/SymbolScheme.java
```diff
@@ -38,6 +39,15 @@
      */
     Character getEscapeSymbol();
 
+    /**
+     * @return the separator used to separate the entity type from the reference
+     * @since 14.8RC1
+     */
+    default Character getEntityTypeSeparator()
+    {
+        return ':';
```
manuelleduc, line 48: Is it the same separator as https://github.com/xwiki/xwiki-platform/pull/1891/files#diff-b03a9b20f9c0d2911ca9a8c6cfbcf08b227757ef36e05905dee9615b14a98ca7R92 ?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/SymbolScheme.java
```diff
@@ -94,6 +104,17 @@
      */
     String[] getParameterReplacementSymbols(EntityType type);
 
+    /**
+     * @param parameter the name of the parameter
+     * @param value the String value of the parameter
+     * @return the parameter converted in the expected type
```
manuelleduc, line 110: WDYT of adding examples of expected values for the parameters? I feel like it could help understanding the method.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/SymbolScheme.java
tmortagne: It is the same character. I could indeed reuse the variable from DefaultSymbolScheme even if it might not be clean to have this reference to DefaultSymbolScheme in the interface.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
```diff
@@ -26,7 +26,10 @@
 
 /**
  * Represent the relation between a document using a link to an entity, and the used entity.
+ * 
+ * @deprecated
```
manuelleduc, line 30: missing explanation?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/SymbolScheme.java
tmortagne: Yeah, I guess I could give the Locale example, which is the only use case currently.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/SymbolScheme.java
tmortagne: Ha, I remember now, I did not want to make it public like all the others variables in DefaultSymbolScheme.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentTest.java
```diff
@@ -177,6 +178,14 @@ protected void setUp() throws Exception
         when(this.xWiki.getStore()).thenReturn(xWikiStoreInterface);
         when(this.xWiki.getDocument(any(DocumentReference.class), any())).thenReturn(this.document);
         when(this.xWiki.getDocumentReference(any(XWikiRequest.class), any())).thenReturn(documentReference);
+        when(this.xWiki.getDocumentReference((EntityReference) any(), any())).then(new Answer<DocumentReference>()
```
manuelleduc, line 181: Not sure but you might be able to do the same without the case by passing a class to `any`.
```suggestion
        when(this.xWiki.getDocumentReference(any(EntityReference.class), any())).then(new Answer<DocumentReference>()
```
Also, a matter a taste but the anonymous `Answer` can be converted to a lambda.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/RefactoringException.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public RefactoringException(String message)
+    {
+        super(message);
+    }
+
+    /**
+     * @param cause the cause (which is saved for later retrieval by the {@link #getCause()} method). A {@code null}
+     *            value is permitted, and indicates that the cause is nonexistent or unknown
+     * @since 9.3RC1
```
manuelleduc, line 49: copy and paste?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
tmortagne: There is not exactly any alternative to this class in the new API. The main point of this class was to hold the Hibernate entity.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/RefactoringException.java
tmortagne: I thought I checked, but indeed looks like I missed one.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/doc/XWikiDocumentTest.java
tmortagne: > Not sure but you might be able to do the same without the case by passing a class to any.

Right, probably an old reflex of when any did not take a class. I could have looked at the line just above that one...

> Also, a matter a taste but the anonymous Answer can be converted to a lambda.

Indeed. sonalint usually suggest me this kind of improvement, but strangely it did not match this one.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-default/src/main/java/org/xwiki/refactoring/internal/link/LinksSolrDocumentMetadataExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+import org.xwiki.model.reference.EntityReference;
+import org.xwiki.search.solr.SolrEntityMetadataExtractor;
+import org.xwiki.search.solr.internal.api.FieldUtils;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+/**
+ * @version $Id$
+ * @since 14.8RC1
```
manuelleduc, line 40: You've added an `@since` on some new internal classes, but not all. For instance, https://github.com/xwiki/xwiki-platform/pull/1891/files#diff-96c5df662af3bfa8505fccfa1f7ded90e5ee2dc91e322be308fee2cec1a8e5a4R32

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/AttachmentSolrMetadataExtractor.java
```diff
@@ -92,6 +92,9 @@ public boolean setFieldsInternal(LengthSolrInputDocument solrDocument, EntityRef
 
         setLocaleAndContentFields(attachment, solrDocument);
 
+        // Extract more metadata
```
manuelleduc, line 95: Out of curiosity, are the additional metadata?

tmortagne: That's where we inject the links in the Solr document currently, but the idea was also to provide an extension point for extensions to impact the Solr search code indexing (an example of extension that would benefit from it is https://extensions.xwiki.org/xwiki/bin/view/Extension/Solr%20Security/).

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-core/src/main/resources/conf/managed-schema
```diff
@@ -129,6 +129,8 @@
 
     <!-- XWIKI BEGIN -->
 
+    <!-- The reference of the entity of the form <type>:<standard entity reference>
+    <field name="reference" type="string" indexed="true" stored="true" />
```
manuelleduc, line 133: dead code?

tmortagne: No, bad comment syntax.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/AttachmentSolrMetadataExtractor.java
tmortagne: So it's not anymore where we inject the links, I made that part a bit more core to the Solr indexer. But it's still very useful for extensions, so I will keep it even if it became useless strictly for XWIKI-16192 (OK I'm too lazy to revert it and redo that in a different branch).

tmortagne:
LinkStore now have its own module and LinkException.

xwiki-platform-core/xwiki-platform-link/src/main/java/org/xwiki/link/LinkException.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.xwiki.link;
+
+/**
+ * Any exception raised in refactoring related APIs.
+ *
+ * @version $Id$
+ * @since 14.8RC1
+ */
+public class LinkException extends Exception
```
surli, line 28: Missing Unstable

tmortagne: Indeed.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/AbstractEntityReferenceResolver.java
```diff
@@ -40,7 +40,10 @@
  */
 public abstract class AbstractEntityReferenceResolver
 {
-    protected static final Map<EntityType, EntityType> TYPE_CONVERT_MAPPING = new EnumMap<>(EntityType.class);
+    /**
+     * The mapping between the PAGE and the DOCUMENT based worlds.
+     */
+    public static final Map<EntityType, EntityType> TYPE_CONVERT_MAPPING = new EnumMap<>(EntityType.class);
```
surli, line 46: The name is not very explicit, couldn't it be named `PAGE_TYPE_CONVERT_MAPPING` or something like that?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/DefaultSymbolScheme.java
```diff
@@ -178,14 +188,29 @@ public class DefaultSymbolScheme implements SymbolScheme
...
     {
+        initialize(withParameters);
+    }
+
+    /**
+     * Initialize internal data structures.
+     * 
+     * @param withParameters true if the parameters syntax should be supported for all entity types
+     */
+    public void initialize(boolean withParameters)
```
surli, line 209: This one does not need to be public since it's controlled by the constructor no?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/DefaultLinkStore.java
```diff
@@ -0,0 +1,244 @@
...
+import org.xwiki.search.solr.internal.SolrClientInstance;
+import org.xwiki.search.solr.internal.SolrSearchCoreUtils;
+import org.xwiki.search.solr.internal.api.FieldUtils;
+
+/**
+ * @version $Id$
+ */
+@Component
+@Singleton
+public class DefaultLinkStore implements LinkStore
```
surli, line 58: Shouldn't it be part of a module `link-default` instead of remaining there?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/DefaultLinkStore.java
```diff
@@ -0,0 +1,244 @@
...
+        if (pageBasedReference != null) {
+            filter.append(FieldUtils.LINKS_EXTENDED);
+            filter.append(':');
+            filter.append(this.utils.toCompleteFilterQueryString(this.linkSerializer.serialize(pageBasedReference)));
+        }
+        if (filter.length() > 0) {
+            filter.append(" OR ");
+        }
+        filter.append(FieldUtils.LINKS_EXTENDED);
+        filter.append(':');
```
surli, line 157: Two occurrences of `:` in same file? I'm surprised checkstyle didn't complain.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/DefaultLinkStore.java
surli: Ah no you keep it here because it's tightly linked to the search solr core and those are not stored in another custom core. Makes sense then.

xwiki-platform-core/xwiki-platform-link/src/main/java/org/xwiki/link/LinkStore.java
```diff
@@ -0,0 +1,53 @@
...
+ * Allow accessing the links extracted from various entities.
+ * 
+ * @version $Id$
+ * @since 14.8RC1
+ */
+@Role
+@Unstable
+public interface LinkStore
+{
+    // TODO: also introduce an API (#getLinks) to get all types of links and not only entities
```
surli, line 38: You're planning a future improvment ticket for this?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiStoreInterface.java
```diff
@@ -400,6 +400,11 @@ List<XWikiDocument> searchDocuments(String wheresql, boolean distinctbylanguage,
 
     void deleteLock(XWikiLock lock, XWikiContext context, boolean bTransaction) throws XWikiException;
 
+    /**
+     * @deprecated links indexing move to Solr, see org.xwiki.refactoring.internal.solr.* in
```
surli, line 404: Shouldn't you point to the new xwiki-platform-link module now instead?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
surli: Sounds like you could do point to the new module to look at.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R140300001XWIKI19571DataMigration.java
```diff
@@ -53,10 +53,13 @@
  *
  * @version $Id$
  * @since 14.3RC1
+ * @deprecated links indexing move to Solr, see org.xwiki.refactoring.internal.solr.* in
+ *             xwiki-platform-refactoring-default module
  */
 @Component
 @Singleton
 @Named(R140300001XWIKI19571DataMigration.HINT)
+@Deprecated(since = "14.8RC1")
```
surli, line 62: What does it mean to deprecate a migration?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/DefaultLinkStore.java
tmortagne: Yes, I prefer to have the store and the indexing (the reader and the writer basically) in the same module, easier to test and keep them in sync.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/DefaultLinkStore.java
tmortagne: Checkstyle does not care about characters, it only checks strings.

xwiki-platform-core/xwiki-platform-link/src/main/java/org/xwiki/link/LinkStore.java
tmortagne: Yes. No short term plan to work on it, but I will create issues for all the TODO, but I prefer to keep the TODO to help whoever will work on those new features in the future.

xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-api/src/main/java/org/xwiki/index/migration/R140300001XWIKI19571DataMigration.java
```diff
@@ -53,10 +53,13 @@
  *
  * @version $Id$
  * @since 14.3RC1
+ * @deprecated links indexing move to Solr, see org.xwiki.refactoring.internal.solr.* in
+ *             xwiki-platform-refactoring-default module
```
tmortagne, line 57: Need to update the comment since it's not in xwiki-platform-refactoring-default anymore.

xwiki-platform-core/xwiki-platform-link/src/main/java/org/xwiki/link/LinkException.java
```diff
@@ -0,0 +1,69 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.link;
+
+import org.xwiki.stability.Unstable;
+
+/**
+ * Any exception raised in refactoring related APIs.
```
surli, line 25: Bad copy paste of the javadoc.

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/DeleteRequest.java
```diff
@@ -145,10 +145,12 @@ public void setUpdateLinks(boolean updateLinks)
      * @return {@code true} if the job should update the links that target the old entity reference (before the delete)
      *         from anywhere on the farm, {@code false} if the job should update only the links from the wiki where the
      *         entity was located before the delete
+     * @deprecated not taken into account anymore
      */
+    @Deprecated(since = "14.8RC1")
     public boolean isUpdateLinksOnFarm()
     {
-        return getProperty(UPDATE_LINKS_ON_FARM, false);
```
surli, line 151: I guess it means the `UPDATE_LINKS_ON_FARM` constant should be deprecated and their usages (notably in velocity template I guess) should be removed no?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiStoreInterface.java
tmortagne: For the load*links yes. The message will me a bit different for the write APIs (since there is no write API anymore).

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/SolrLinkSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Help serialize and unserialize links stored in Solr.
+ * 
+ * @version $Id$
+ * @since 14.8RC1
+ */
+@Component(roles = SolrLinkSerializer.class)
+@Singleton
+public class SolrLinkSerializer
```
surli, line 39: Instead of relying on this custom implem, wouldn't it worth it to have it implemented in an EntityReference <-> String converter, so that it could be directly use with SolrUtils to convert the field to the appropriate value?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
tmortagne: But you won't really find anything that you can use for whatever you used to be doing with XWikiLink. The closest alternative to XWikiLink is basically EntityReference.

xwiki-platform-core/xwiki-platform-link/src/main/java/org/xwiki/link/LinkException.java
tmortagne: Indeed.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-core/src/main/resources/conf/managed-schema
```diff
@@ -163,6 +165,10 @@
     <!-- Whether the entity is hidden on not. Only documents can be made hidden explicitly.
          The attachments, objects and object properties are hidden if the document that holds them is hidden. -->
     <field name="hidden" type="boolean" indexed="true" stored="true" />
+    <!-- The references of the resources where the various links found in that entity are leading to -->
+    <field name="links" type="string" indexed="true" stored="true" multiValued="true" />
+    <!-- Contains links plus all the references parents to make easier to search for links from any entity level (per wiki, per space, per document, etc.) -->
+    <field name="links_extended" type="string" indexed="true" stored="false" multiValued="true" />
```
surli, line 171: I don't recall, what does `stored="false"` mean?

xwiki-platform-core/xwiki-platform-refactoring/xwiki-platform-refactoring-api/src/main/java/org/xwiki/refactoring/job/DeleteRequest.java
tmortagne: I deprecated the related APIs, but indeed I forgot the property itself.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/SolrLinkSerializer.java
tmortagne: It's not an EntityReference <-> String converter, but more a SomeFutureLinkType <-> String converter.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-core/src/main/resources/conf/managed-schema
tmortagne: That it's not stored :)

Basically it means you indicate Solr that you will only use it as a search criteria but you never want to retrieve it.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/SolrEntityMetadataExtractor.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * An extension point to be called when an entity is indexed for the Solr search core.
+ * 
+ * @param <E> the type of the parsed entity
+ * @version $Id$
+ * @since 14.8RC1
+ */
+@Role
+@Unstable
+public interface SolrEntityMetadataExtractor<E>
```
surli, line 35: That's a new role you provide, but there's no component implementing it yet? Or I missed it?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
surli: Well you could end up here just because you followed the API that were returning those XWikiLink. That's why I suggest to point out to the new API module. Sounds better than nothing to me.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/DefaultLinkStore.java
surli: right

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/internal/reference/DefaultSymbolScheme.java
surli: Then I'd make it private since the previous initialize was removed anyway.

xwiki-platform-core/xwiki-platform-link/src/main/java/org/xwiki/link/LinkStore.java
surli: Sounds good, wanted to be sure it was not leftover that you were planning to have before merging.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-core/src/main/resources/conf/managed-schema
surli: Ok so it only stores indexes.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/SolrEntityMetadataExtractor.java
tmortagne: You are not missing something. I used to use this new system to implement link indexing from the refactoring module, but link indexing is not part of the core metadata extractor. So it's a new system for extensions which is not use by XWiki Standard, I did not have the heart to remove it (it would be very nice to remove a few hacks in https://extensions.xwiki.org/xwiki/bin/view/Extension/Solr%20Security/).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiLink.java
tmortagne: > you followed the API that were returning those XWikiLink

Those APIs and deprecated too and document alternative usually.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-core/src/main/resources/conf/managed-schema
tmortagne: Yes, not exactly sure what it means exactly internally, but from user point it feels like it.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/metadata/SolrLinkSerializer.java
tmortagne: > here you just need to convert a String to an EntityReference with parameters no?

No. We have resolvers and serializes for that (and that's what SolrLinkSerializer, notice the "withparameters" role hints). The point of SolrLinkSerializer is to serialize a **link** and currently it only supports entity kind of links.

PR #1892, opened 2022-08-17
XWIKI-20071: Cannot use LiveData with LiveTable results for getting results from subwiki

surli:
* Ensure to set the wiki reference for getting the LT results


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataResultsRenderer.java
```diff
@@ -103,6 +107,8 @@ public String getLiveTableResultsFromPage(String page, LiveDataQuery query) thro
                 return document.getRenderedContent(Syntax.PLAIN_1_0, xcontext);
             } catch (Exception e) {
                 throw new RuntimeException(e);
```
surli, line 109: That one is not pretty but not sure which exception I should thrown here in LD.

manuelleduc (approved):

PR #1895, opened 2022-08-19
XWIKI-19993: Allow admins to configure the macros

manuelleduc:
- Introduce getHiddenMacroCategories in RenderingScriptService
- Use getHiddenMacroCategories in the CKEditor macro dialog implementation
- Document rendering.transformation.macro.hiddenCategories in xwiki.properties.vm

Jira: https://jira.xwiki.org/browse/XWIKI-19993

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/MacroService.xml
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/MacroService.xml
@@ -115,7 +115,7 @@
   ## categories.
   ## TODO: Make the list of hidden by default categories configurable from the administration (XWIKI-19993).
   #if(!$services.user.getProperties().displayHiddenDocuments())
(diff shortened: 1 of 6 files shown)


vmassol (approved):
LGTM

tmortagne (approved):

surli (commented):
I haven't checked more how the categories are compared to hide them or not. But here it looks like the API expect a special format for the categories (first character uppercase, remaining lowercase). I hope it's not too fragile and it won't be broken if a user uses "internal" instead of "Internal" for example.

Besides that LGTM.

manuelleduc:
> I haven't checked more how the categories are compared to hide them or not. But here it looks like the API expect a special format for the categories (first character uppercase, remaining lowercase). I hope it's not too fragile and it won't be broken if a user uses "internal" instead of "Internal" for example.

It's based on how the categories are already defined and used. So playing with macro categories case is definitely not advised and would break things (even before this change).

surli:
> So playing with macro categories case is definitely not advised and would break things (even before this change).

Ok then I would explicit that in the comment of the new property.

PR #1898, opened 2022-08-25
XWIKI-20028: Cache issue when mention components need to be reloaded

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20028

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
@@ -662,10 +662,9 @@
       <version>${project.version}</version>
     </dependency>
 
-    <!-- Must be installed as a core extension because xwiki-platform-mentions-default must be installed as a core 
-         extension. Otherwise, the components from mentions-api that depend on components from mentions-default are not
-         reloaded when needed. One concrete consequence being exceptions thrown when accessing a configuration cache in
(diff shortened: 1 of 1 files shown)


xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
```diff
@@ -662,6 +662,18 @@
       <version>${project.version}</version>
     </dependency>
 
+    <!-- Must be installed at the farm level because xwiki-platform-mentions-default must be installed at farm level.
```
tmortagne, line 665: Putting this extension here does not "install it at farm level", it makes it a core extension.

manuelleduc: Comment updated.

xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
```diff
@@ -662,6 +662,18 @@
       <version>${project.version}</version>
     </dependency>
 
+    <!-- Must be installed as a core extension because xwiki-platform-mentions-default must be installed as a core
```
tmortagne, line 665: I would get rid of any reference to "installed" to avoid confusion with what we commonly call installed extensions.

PR #1900, opened 2022-08-26
XWIKI-20099: Move DocumentVersionReference to xwiki-platform-model-api

surli:
* Move DocumentVersionReference to model-api in proper package
  * Create a new legacy module for extension-handler-xar
  * Move the old class in it and mark it deprecated
  * Move old usage of the old class, to use new one
  * Create an aspect for the DocumentUnifiedDiff constructor for
    preserving backward compatibility


surli:
Related proposal: https://forum.xwiki.org/t/move-documentversionreference-to-xwiki-platform-model/10905/

xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-extension-handler-xar/src/main/aspect/org/xwiki/extension/xar/job/diff/DocumentUnifiedDiffAspect.aj
```diff
@@ -0,0 +1,42 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.extension.xar.job.diff;
+
+/**
+ * Add a backward compatibility layer to the {@link DocumentUnifiedDiff} class.
+ *
+ * @version $Id$
```
manuelleduc, line 25: No since?

surli: Not really needed for aspects, is it?

manuelleduc (commented):
Shouldn't `xwiki-platform-legacy-extension-handler-xar` be used in `xwiki-platform-distribution-war-legacydependencies` for instance?

xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-extension-handler-xar/src/main/aspect/org/xwiki/extension/xar/job/diff/DocumentUnifiedDiffAspect.aj
manuelleduc: The aspect is not always introduced on the same version as the deprecation.

surli: Sure, just that I'm not entirely sure about the value it provides. Now I can add it if you prefer.

manuelleduc: Mainly uniformity since it's available on all the other aspects afaics. I'm not sure of the added value I agree.

tmortagne:
`xwiki-platform-distribution-war-legacydependencies` is supposed to list all the legacy artifacts that are supposed to be part of the XWiki Standard WAR.

xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-extension-handler-xar/src/main/aspect/org/xwiki/extension/xar/job/diff/DocumentUnifiedDiffAspect.aj
```diff
@@ -0,0 +1,42 @@
...
+         * Creates a new instance to hold the differences between the specified document versions.
+         *
+         * @param previousReference the reference to the previous version of the document
+         * @param nextReference the reference to the next version of the document
+         * @deprecated now use {@link #DocumentUnifiedDiff(org.xwiki.model.reference.DocumentVersionReference, org.xwiki.model.reference.DocumentVersionReference)}
+         */
+        @Deprecated(since = "14.8RC1")
+        public DocumentUnifiedDiff.new(DocumentVersionReference previousReference, DocumentVersionReference nextReference)
+        {
+            this(new org.xwiki.model.reference.DocumentVersionReference(previousReference, previousReference.getVersion()),
```
tmortagne, line 39: You don't need to pass the version separately, the parameters are copied.

surli: Right I missed the constructor with a single reference, I'll fix that.

xwiki-platform-distribution/xwiki-platform-distribution-war-legacydependencies/pom.xml
```diff
@@ -319,5 +319,17 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-legacy-extension-handler-xar</artifactId>
+      <version>${project.version}</version>
+      <!-- Exclude transitive dependencies which are conflicting with legacy dependencies -->
```
surli, line 326: @tmortagne I spotted that I should exclude oldcore but I didn't perform a deep analysis here: is the build supposed to fail in case I missed some? Or do I need manually to check the deps?

tmortagne: If oldcore is not excluded, we end up with both legacy-oldcore and oldcore in the WAR which should cause the duplicate-finder-maven-plugin plugin to fail the build (because we end up with JARs which contain classes and resources with the same name).

surli: ok, so since the build with quality profile is working there it should be enough then I guess.

manuelleduc (approved):

xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-extension-handler-xar/src/main/java/org/xwiki/extension/xar/job/diff/DocumentVersionReference.java
```diff
@@ -28,7 +28,9 @@
  * 
  * @version $Id$
  * @since 7.0RC1
+ * @deprecated Use {@link org.xwiki.model.reference.DocumentVersionReference} instead.
  */
+@Deprecated(since = "14.8RC1")
 public class DocumentVersionReference extends DocumentReference
```
tmortagne, line 34: Extending `org.xwiki.model.reference.DocumentVersionReference` might make retro compatibility easier. Also make DocumentUnifiedDiff.new implementation a bit simpler.

surli: Might avoid a bit of code duplication.

PR #1901, opened 2022-08-29
XWIKI-20078: Wrong panels preferences displayed and edited in the administration of a page when it's not a top level page

acotiuga:
* use the getSpacePreferenceFor API that takes a spaceReference as a parameter to correctly read preferences for a space of any level

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
@@ -1159,9 +1159,7 @@ public class XWiki extends Api
      * @param preference Preference name
      * @param space The space for which this preference is requested
      * @return The preference for this wiki and the current locale
-     * @deprecated use {@link #getSpacePreferenceFor(String, SpaceReference)} instead
      */
-    @Deprecated(since = "14.7")
     public String getSpacePreferenceFor(String preference, String space)
     {
         return getSpacePreferenceFor(preference, space, "");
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/api/XWiki.java
```diff
@@ -1159,7 +1159,9 @@ public String getSkinPreference(String preference, String defaultValue)
      * @param preference Preference name
      * @param space The space for which this preference is requested
      * @return The preference for this wiki and the current locale
+     * @deprecated use {@link #getSpacePreferenceFor(String, SpaceReference)} instead
      */
+    @Deprecated(since = "14.7")
```
manuelleduc, line 1164: I don't think it's safe to merge this that close to the release of 14.7. I'd suggest to move this to 14.8RC1.
Also, it could be interesting to make a proposal on the forum.
Finally, it's important to remove all use of this method on the codebase, and to move this method to `xwiki-platform-legacy-oldcore` (see https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HDeprecatingCode)

acotiuga: I opened a vote topic https://forum.xwiki.org/t/deprecate-api-xwiki-getspacepreferencefor-methods-using-space-as-string/10912 and as soon as it gets voted, I will update the deprecation version and I will move the methods to the legacy module.
Thanks!

vmassol:
@acotiuga why remove the deprecation? #curious

acotiuga:
@vmassol I created https://jira.xwiki.org/browse/XWIKI-20101 for deprecation removal, since there is more work there while this can be merged quicker and also include in 13.10.9.

PR #1902, opened 2022-08-29
XWIKI-20101: Deprecate XWiki#getSpacePreferenceFor() methods that are using String instead of SpaceReference

acotiuga:


xwiki-platform-core/xwiki-platform-legacy/xwiki-platform-legacy-oldcore/src/main/aspect/com/xpn/xwiki/XWikiCompatibilityAspect.aj
```diff
@@ -193,6 +193,16 @@ public privileged aspect XWikiCompatibilityAspect
         return this.getSpaceCopyright(context);
     }
 
+    /**
+     * @deprecated since 7.4M1, use {@link #getSpacePreference(String, SpaceReference, String, XWikiContext)} instead
+     */
+    @Deprecated
```
manuelleduc, line 199:
```suggestion
    @Deprecated(since = "14.8RC1")
```

manuelleduc (commented):
Shoudn't `getSpacePreferenceFor` also be cleaned up from `layoutvars.vm`?

acotiuga:
@manuelleduc no, because we deprecate only the methods with space as String. The remaining ones are with space as SpaceReference.

tmortagne (approved):

PR #1904, opened 2022-08-30
XWIKI-19893 Allow to set a user when updating the search index

ClemensRobbenhaar:
... to fix issues where the computed title is not indexed.

Note: the admin section to configure that user is also shown in subwikis, but as it is useless there it only shows an explanatory text. I have found no way to not show the section in subwikis (except for adding the new pages in a separate ` xwiki-platform-search-solr-ui-mainwiki` project, which seems overkill to me.


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/DefaultIndexingUserConfig.java
```diff
@@ -0,0 +1,94 @@
...
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+import com.xpn.xwiki.objects.classes.ListClass;
+
+/**
+ * Provides the configured indexing user via information stored in a wiki page.
+ *
+ * @version $Id$
+ * @since 14.8
```
tmortagne, line 45: I guess the target is more 14.8RC1.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/DefaultIndexingUserConfig.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public DocumentReference getIndexingUserReference() throws XWikiException
+    {
+        DocumentReference indexingUser = null;
+        XWikiContext context = contextProvider.get();
+        XWiki wiki = context.getWiki();
+
+        XWikiDocument configPage = wiki.getDocument(new DocumentReference(CONFIG_PAGE,
+            context.getWikiReference()), context);
```
tmortagne, line 78: You said in the description that this is supposed to be configured on the main wiki but here you are loading the configuration in the current wiki (you could have directly passed the `LocalDocumentReference` to `getDocument` by the way, no need to create a `DocumentReference`).

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/XWiki/SolrSearchAdminIndexingUserSheet.xml
```diff
@@ -0,0 +1,324 @@
...
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>== {{translation key="admin.indexinguser" /}} ==
+
+(% class="noitems" %)
+{{translation key="admin.indexinguser.explanation.part1" /}}
+
+(% class="noitems" %)
+{{translation key="admin.indexinguser.explanation.part2" /}}
```
tmortagne, line 45: This should maybe be highlighted a bit more (in a warning macro maybe) as people will be tempted to put the main wiki admin.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/DefaultIndexingUserConfig.java
ClemensRobbenhaar: Ah, I did not know that I can use a `LocalDocument` reference.
The current wiki where the reference is loaded is always the main wiki - that is not decided here, however, but how the `DefaultSolrIndexer` sets up it the ExecutionContext.  
If that setup ever changes then indeed the statement about the main wiki is wrong, but I am happy to update the code to support different indexing users in subwikis then.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/XWiki/SolrSearchAdminIndexingUserSheet.xml
ClemensRobbenhaar: Good idea. I have updated the UI to put the security implications inside a warning box (and also removes the `class="noitems"` form the rest of the text as it looked too small compared to the box.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/DefaultIndexingUserConfig.java
ClemensRobbenhaar: ok, updated

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/DefaultIndexingUserConfig.java
tmortagne: >  The current wiki where the reference is loaded is always the main wiki - that is not decided here, however, but how the DefaultSolrIndexer sets up it the ExecutionContext.

We are not talking about a private method in `DefaultSolrIndexer` here, but an independent component which might be used elsewhere some day without thinking what it does is actually wrong. If the configuration only make sense at main wiki level, `DefaultIndexingUserConfig` should reflect that.

ClemensRobbenhaar: Well, I hoped in the future to support also use cases where the indexing user can be defined separately in a subwiki (it might happen that non-admin users in the main wiki have no access to the subwikis). This seems currently not possible with the way the indexer works (unless the "batches" always belong to one subwiki, or the user is set in a different place).

But anyway, as the case is now, I bulletproofed the code to always fetch the indexing user from the main wiki.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/XWiki/SolrSearchAdminIndexingUserClass.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.1" encoding="UTF-8"?>
```
tmortagne, line 1: I feel it would make more sense to create this class through a mandatory class initializer since you have several Java codes expecting this class to exist, including a mandatory document initializer.

ClemensRobbenhaar: This class is now created by its own initializer.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/IndexingUserConfigurationInitializer.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @version $Id$
+ * @since 14.8M1
+ */
+@Component
+@Named("XWiki.SolrSearchAdminIndexingUser")
+@Singleton
+// initalize after the corresponding class
+@Priority(MandatoryDocumentInitializer.DEFAULT_PRIORITY + 100)
+public class IndexingUserConfigurationInitializer extends AbstractMandatoryDocumentInitializer
```
tmortagne, line 48: Since this document is currently only used on main wiki, the initializer (and the other one I guess) should probably reflect it. You can indicate this in #isMainWikiOnly().

ClemensRobbenhaar: Yes, I could do it, but the UI Section is also show in the subwikis (currently with a "there is nothing to do here, sorry", message) as I could not find a way to only display it in the main wiki, so I thought to have the object there, too, just in case I figure out how to use this in subwikis. (Come to think about it, maybe I should move that UIX to an initializer, too?).
Well, I add the `isMainWikiOnly` in the two initializers here. This can be removed any time the indexing user can be made to work in subwikis.

tmortagne: >  This can be removed any time the indexing user can be made to work in subwikis.

+1

ClemensRobbenhaar: Ok, done.

tmortagne (approved):

PR #1907, opened 2022-08-30
[XWIKI-20109] Notification headers for autoreplies

the-glu:
Hello,

We do have an issue with people in vacations and their autoreply email responding to xwiki automatic notifications.

This PR fixes that by adding common email headers ( RFC 5436, 2.7.1 for 'Auto-Submitted', 'X-Auto-Response-Suppress' used by Exchange servers ) to prevent automatic replies.

It's only set in the notification system. It may be set by default in xwiki since everything is probably generated emails - but I'm not sure to I did it only there :)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/email/AbstractMimeMessageIterator.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/email/AbstractMimeMessageIterator.java
@@ -364,6 +364,8 @@ public abstract class AbstractMimeMessageIterator implements Iterator<MimeMessag
 
             updateFactoryParameters(templateDocumentReference);
             message = this.factory.createMessage(templateDocumentReference, this.factoryParameters);
+            message.setHeader("Auto-Submitted", "auto-generated");
+            message.setHeader("X-Auto-Response-Suppress", "All");
 
             List<EntityEvent> events = new ArrayList<>();
             this.currentEvents.forEach(


surli:
Hi, thanks for the contribution. It looks like an interesting fix but it needs a proper improvment ticket so that we can follow it up. Could you create that on https://jira.xwiki.org please?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/email/AbstractMimeMessageIterator.java
```diff
@@ -364,6 +364,8 @@ public MimeMessage next()
 
             updateFactoryParameters(templateDocumentReference);
             message = this.factory.createMessage(templateDocumentReference, this.factoryParameters);
+            message.setHeader("Auto-Submitted", "auto-generated");
+            message.setHeader("X-Auto-Response-Suppress", "All");
```
tmortagne, line 368: Great to see this feature exist in the protocol.

However, I'm not sure this is the best location, I would expect something like that to be more somewhere in `createMessage` itself but looks like we don't really have any common code between message factories so that would mean adding it to each of them.

I guess another possibility would be to make it the default in ExtendedMimeMessage constructor, but it might be a bit too much.

WDYT @vmassol ?

the-glu:
> Hi, thanks for the contribution. It looks like an interesting fix but it needs a proper improvment ticket so that we can follow it up. Could you create that on https://jira.xwiki.org please?

Sure, it's there: https://jira.xwiki.org/browse/XWIKI-20109

I updated name / commit messages accordingly

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/email/AbstractMimeMessageIterator.java
surli: AFAICS only `Auto-Submitted` is defined in the RFC (https://www.rfc-editor.org/rfc/rfc5436.html#section-2.7.1) the other one is defined by Microsoft according to https://www.arp242.net/autoreply.html. 

Would be good to maybe add some comments to at least explain that. Also reading a bit more the RFC it sounds like the Auto-Submitted field should include the owner-email: 
> The header field "Auto-Submitted: auto-notified" MUST include one or
   both of the following parameters: [owner-email or owner-token]

> I guess another possibility would be to make it the default in ExtendedMimeMessage constructor, but it might be a bit too much.

According to the RFC the header could contain `auto-generated` or `auto-notified`. IMO it makes to always have the header in the emails sent by XWiki, now it's not necessarly always `auto-generated`: it makes sense here for notification, but not necessarily for some other mails (like invitation email for example).

surli: > it would makes sense to use auto-notified for notification system.

the RFC actually says: 
>  Keyword value: auto-notified
   Description: Indicates that a message was generated by a Sieve
      notification system.

we're not in that case since it's not a **Sieve** notification system. So I guess we can just commit to keep auto-generated. I'm not yet entirely sure about putting that to `ExtendedMimeMessage`: what if some extensions / scripts reuse that to send mails that are not generated?
We could maybe go at least for a first version to keep it there only. WDYT?

surli:
Closing in favor of https://github.com/xwiki/xwiki-platform/pull/3802 since I don't have write access on original branch.

PR #1910, opened 2022-09-05
XWIKI-20112: Custom headers are not registered when loaded to soon

oanalavinia:
* move code to plain js, since jquery blocks are executed too late
* recreate the script element to be able to load dynamically added js resources

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
@@ -1835,23 +1835,6 @@ document.observe("xwiki:dom:loaded", function() {
   };
 
   /**
-   * Injects the given stylesheet or JavaScript file into the page head.
-   *
-   * @param {Element} resource stylesheet or JavaScript resource
-   */
-  var loadResource = function(resource) {
-    if (resource.getAttribute('src')) {
-      let script = document.createElement("script");
-      script.setAttribute('defer', '');
-      script.setAttribute('type', 'text/javascript');
-      script.setAttribute('src', resource.getAttribute('src'));
-      document.head.appendChild(script);
-    } else {
-      document.head.appendChild(resource);
-    }
-  };
-
-  /**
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1821,15 +1821,34 @@ document.observe("xwiki:dom:loaded", function() {
...
+   * Injects the given stylesheet or JavaScript file into the page head.
+   *
+   * @param {Element} resource stylesheet or JavaScript resource
+   */
+  var loadResource = function(resource) {
+    if (resource.getAttribute('src')) {
+      let script = document.createElement("script");
+      script.setAttribute('defer', '');
+      script.setAttribute('type', 'text/javascript');
+      script.setAttribute('src', resource.getAttribute('src'));
```
mflorea, line 1847: Have you tried cloning the script element instead. My worry is that you lose some of the attributes from the script element which could be important (the script may read them with https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript ). In any case it needs a comment explaining why we treat ``script`` and ``link`` tags differently.

oanalavinia: No, it's not working with clone either. I'll add a comment
Another way is using `document.createRange().createContextualFragment(resource.outerHTML)`, to preserve the script attributes (see https://ghinda.net/article/script-tags/)

mflorea: Couldn't we use this in ``extendPageHead`` to convert the string HTML to a ``DocumentFragment`` from where to extract the ``script`` and ``link`` tags, instead of using the detached ``DIV`` element? It feels redundant to go from string to DOM (using the ``DIV``) and then back to string (``outerHTML``) and finally back to DOM (``createContextualFragment``), when we could go directly from string to DOM using ``createContextualFragment`` (using the ``HEAD`` element as context).

mflorea (approved):

PR #1912, opened 2022-09-07
XWIKI-19480: Disable Lightbox for profile, activity stream and livetable

oanalavinia:
* disable lightbox for the activity stream column, since it would display the profile picture for each activity done
* disable lightbox for the livetable, since it might contain images not intended as content


vmassol:
It's not needed for LD?

oanalavinia:
Right. Should this be addressed in another issue? Since it's not related to the profile picture issue

vmassol:
Ideally yes but since you've already started combining several commits in the same issue, I had modified the jira issue title.

oanalavinia:
I will update the PR to adapt it to the new jira issue title (to include the livedata as well).
Note that initially I did the modifications for the livetable since it was impacting the profile page as well (the loading bars were displayed inside the page lightbox)

PR #1915, opened 2022-09-09
XWIKI-19480: Lightbox gallery displays many times the profile picture and additional ones

oanalavinia:
XWIKI-20125: Disable lightbox for livetable


manuelleduc:
Is this change tested somewhere? It's interesting to prevent future regression imo.

oanalavinia:
I have a test for content that has the lightbox feature disabled https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-test/xwiki-platform-image-lightbox-test-docker/src/test/it/org/xwiki/image/lightbox/test/ui/LightboxIT.java#L381
Do you think it would still be valuable to also have tests for specific cases?

PR #1916, opened 2022-09-09
[Misc] Trigger an event when an entry is updated on LD

surli:
Trigger an event whenever entries are displayed in a live data. 
This code needs to be better tested, right now I only tested on attachments live data with a JSX containing:
```
require(['jquery'], function ($) {
  var initialize = function () {
    console.log('initialize called');
    $('#docAttachments').find('a.move-attachment').on('click', function (event) {
      event.preventDefault();
      console.log("Property attached!");
    });
  }
  $(document).on('xwiki:livedata:entriesUpdated', initialize);
});
```

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
@@ -583,10 +583,7 @@ define('xwiki-livedata', [
       return this.fetchEntries()
         .then(data => {
           this.data.data = Object.freeze(data);
-          let self = this;
-          Vue.nextTick(function () {
-            self.triggerEvent('entriesUpdated', {});
-          });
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/Logic.js
```diff
@@ -583,6 +583,10 @@ define('xwiki-livedata', [
       return this.fetchEntries()
         .then(data => {
           this.data.data = Object.freeze(data);
+          let self = this;
+          Vue.nextTick(function () {
+            self.triggerEvent('entriesUpdated', {});
+          });
```
manuelleduc, line 589: An alternative is to use an arrow function expression instead of an anonymous function. It's shorter to write. In addition, arrow function does not have their own binding so saving `this` into `self` (which is would call `that` out of habit) is not required.
```suggestion
          Vue.nextTick(() => this.triggerEvent('entriesUpdated', {}));
```

PR #1917, opened 2022-09-11
XWIKI-20112: Custom headers are not registered when loaded to soon

oanalavinia:
* make sure that scripts are executed in the order they are added

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
@@ -1854,9 +1854,10 @@ document.observe("xwiki:dom:loaded", function() {
       }
       var url = resource.getAttribute('href') || resource.getAttribute('src');
       if(loadedResources.indexOf(url) < 0) {
-        if (resource.getAttribute('src')) {
+        if (resource.getAttribute('src') && !resource.getAttribute('async')) {
           // Scripts that are dynamically created and added to the document are async by default, which means there are
           // no guarantees they will execute in the same order they were added.
+          // See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#compatibility_notes
           resource.async = false;
         }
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1854,6 +1854,11 @@ document.observe("xwiki:dom:loaded", function() {
       }
       var url = resource.getAttribute('href') || resource.getAttribute('src');
       if(loadedResources.indexOf(url) < 0) {
+        if (resource.getAttribute('src')) {
+          // Scripts that are dynamically created and added to the document are async by default, which means there are
+          // no guarantees they will execute in the same order they were added.
+          resource.async = false;
```
mflorea, line 1860: I'm not fully sure about this. Normally it's not up to this code to decide how a script is loaded. It should respect whatever is in the received HTML (through the custom HTTP headers). If a script needs to be loaded synchronously then the HTML should indicate this. Are you saying that the same HTML produces a different behavior when parsed on page load vs. when injected after page load?

oanalavinia: I found that scripts that are dynamically added are async by default, which means that the same script can have different results depending on the time is was added if the `async` attribute is not explicitly specified

oanalavinia: In the specific case of `upload.js` and `attachments.js`, which are now not loaded in order due to this problem, both scripts are defined as `defer`. But it seems that the defer attribute has no effect on dynamically added scripts (looking also at questions like https://stackoverflow.com/questions/68795114/execution-order-of-dynamically-added-script-tags-marked-with-defer-attribute) (maybe it has to do with the fact that defer scripts are executed when the DOM is ready, but before `DOMContentLoaded` event)

mflorea (approved):

PR #1926, opened 2022-09-28
XWIKI-20132: Merge conflict hit during flavor upgrade when upgrading from 8.4 with filestore

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20132

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/java/org/xwiki/mentions/internal/descriptors/NotificationAdministrationDocumentInitializer.java
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/java/org/xwiki/mentions/internal/descriptors/NotificationAdministrationDocumentInitializer.java
@@ -22,6 +22,8 @@ package org.xwiki.mentions.internal.descriptors;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.inject.Inject;
 import javax.inject.Named;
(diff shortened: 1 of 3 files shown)


tmortagne:
AFAIR the "startDate" field is a workaround for a problem associated with the old post filtering mode: when you add a new preference you could suddenly end up with old events in your user notifications, so a date was added to also indicate after which date event date this preference applies.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/java/org/xwiki/mentions/internal/descriptors/NotificationAdministrationDocumentInitializer.java
```diff
@@ -92,13 +99,20 @@ public boolean updateDocument(XWikiDocument document)
         }
 
         if (!mentionEventPreferenceAlreadySaved) {
+            String wikiId = document.getDocumentReference().getWikiReference().getName();
+            // It is possible for several document with the "XWiki.Notifications.Code.NotificationAdministration" 
+            // document reference to be initialized separately. We keep a map of the computed dates so that every
```
surli, line 104: > to be initialized separately

Here you mean "separately and in parallel" right? Because if it's called twice sequentially it should not be a problem as there is a check if the xobject is already present or not.

manuelleduc: It is a problem even sequentially if two `XWikiDocument` objects are instantiated with the same document reference. In this case they will be initialized with different `startDate`. This is usually not an issue since document initializers usually use static content (so no conflict in the end). But, it is not true here since a new date is instantiated for each call.

surli (approved):

tmortagne (approved):

PR #1927, opened 2022-09-29
XWIKI-20167: The cursor is not visible after a macro that supports inline is added standalone on chrome

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20167

Result on Chrome:
Peek 2022-09-29 16-19

PS: Working as well on Firefox and Edge.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-macro/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-macro/plugin.js
@@ -447,15 +447,16 @@
           // Remove the element we added after the macro to force the inline rendering.
           var inlineEnforcer = editor.editable().findOne('span#xwiki-macro-inline-enforcer');
           if (inlineEnforcer) {
-            // Place the caret after the inserted inline macro in order to allow the user to continue typing.
-            // We proceed by inserting and non-breakable space after the inline enforcer, before selecting its content.
-            // It is required to be cross-browser compatible. Without these operations, the caret was not visible in
-            // Chrome after the inline macro insertion.
+            // Place the caret after the inserted inline macro in order to allow the user to continue typing. We proceed
+            // by inserting and non-breakable space after the inline enforcer. It is required to be cross-browser
+            // compatible. Without these operations, the caret is not visible in Chrome after the inline macro
+            // insertion.
             var space = new CKEDITOR.dom.text('\u00A0');
             space.insertAfter(inlineEnforcer);
             var range = editor.createRange();
             range.selectNodeContents(space);
-            // TEST with not selected space.
+            // Make the range of length zero (from endOffset to endOffset) so that a caret is displayed after the
+            // space, instead of a selection of the space.
             range.setStartAfter(space, range.endOffset);
             editor.getSelection().selectRanges([range]);
             // Clean up the inline enforcer to avoid duplicates for the next inline macro insertion.
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-macro/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-macro/plugin.js
@@ -447,13 +447,18 @@
           // Remove the element we added after the macro to force the inline rendering.
           var inlineEnforcer = editor.editable().findOne('span#xwiki-macro-inline-enforcer');
           if (inlineEnforcer) {
-            // Adding a zero width space between the widget and inlineEnforcer help webkit based browsers place the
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-macro/plugin.js
```diff
@@ -447,6 +447,11 @@
           // Remove the element we added after the macro to force the inline rendering.
           var inlineEnforcer = editor.editable().findOne('span#xwiki-macro-inline-enforcer');
           if (inlineEnforcer) {
+            // Adding a zero width space between the widget and inlineEnforcer help webkit based browsers place the
+            // caret once inlineEnforcer is removed (without this, the caret is present but not visible). See
+            // https://github.com/ckeditor/ckeditor5/issues/1724 (mentioning CKEditor 5, but also working with CKEditor
+            // 4). 
+            inlineEnforcer.$.insertAdjacentText('beforebegin', '\u200b');
```
mflorea, line 454: Is this zero-width space saved in the generated wiki syntax? If it is then it's not right because the user didn't intend to insert it (didn't ask for it). Moreover, I'm worried that this unexpected, invisible space may cause problems later on. I'd rather insert a non-breaking space and select it so that its gets overwritten as soon as the user types. But even if it stays (e.g. if the user uses the arrow keys) it's less exotic than the zero-width space and it's easier to spot and remove.

manuelleduc: Should have precised it indeed but the zero-width space is not persisted on the wiki syntax.

+1 for the insertion of a space (without selecting it). I feel like 99% of the time users will continue the sentence with a space. The only exception is can imagine if if they users wants to continue with a colon (`:`), in which they will just have to press backspace to remove the space.

Having the space selected by default feels strange imo. Users will see a very tiny selection which feels like a glitch. WDYT?

mflorea: I don't find the selected space that strange, but maybe it's just me. In any case, inserting a non-breaking space after the macro feels safer than an exotic zero-width space.

manuelleduc: After testing ` `, and `&nbsp` instead of `\u200b`, `\u200b` is the only one to provide a nice UX with:
1. a visible caret of Firefox and Chrome
2. a caret located at the right of the mention (and not overlapping it)

mflorea: I assume the above comment is for the case when we don't select the inserted space, be it ``\u200b`` or ``&nbsp``. Selecting the inserted space should move the "caret" on the right of the mention, and it should be visible. So this leaves us two working solutions:

* Insert ``\u200b`` (without selecting it). What I don't like about this is that:
  * the zero-width space is currently removed but I'm not sure it's done by our code so this can change later without us noticing right away, and since this is an exotic character, rarely used, I'm afraid it could lead to unexpected problems
  * the zero-width space currently leads to a different behavior than normal space or non-breaking space but there's no guarantee that in the future the browser / CKEditor won't treat it the same as other space characters, in which case we'll hit the problem again
* Insert ``&nbsp`` and select it. The behavior is not as nice, but not terrible. The advantage I see is that we have control over the fix. The browser must display the selected space at the right of the mention, there's no doubt about this, now and in the future.

Another solution, that I think you didn't try, if you don't like the selected ``&nbsp``, is to move the caret after it. That is, to place the caret in that text node (that has only the ``nbsp``), at the end (``endOffset`` 1).

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-macro/plugin.js
```diff
@@ -448,7 +448,17 @@
...
-            editor.getSelection().selectElement(inlineEnforcer);
+            // We proceed by inserting and non-breakable space after the inline enforcer, before selecting its content.
+            // It is required to be cross-browser compatible. Without these operations, the caret was not visible in
+            // Chrome after the inline macro insertion.
+            var space = new CKEDITOR.dom.text('\u00A0');
+            space.insertAfter(inlineEnforcer);
+            var range = editor.createRange();
+            range.selectNodeContents(space);
+            // TEST with not selected space.
+            range.setStartAfter(space, range.endOffset);
```
manuelleduc, line 459: This is optional and is only needed if we decide that's it better not to have the non-breakable space not selected after insert.

manuelleduc:
With space selected:
Peek 2022-09-29 16-19

With space not selected:
Peek 2022-10-06 10-50

PR #1930, opened 2022-10-03
XWIKI-19702: Attachment selector upload should not save immediately the page

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-19702

- Use the temporary attachment store to save newly uploaded attachments
- Allow users with view right to access the upload button
- List temporary attachments from the current document in the attachment selection modal

TODO:
- [x] Visually distinguish temporary attachment from the others in the selection modal
- [x] Allow temporary attachments to be removed or remove the "delete" button from temporary attachment
- [x] Maybe add a fallback to the legacy behavior when `$services.temporaryAttachments` is missing
- [x] Add missing translations
- [x] take into account `sortAttachmentsBy`
- [x] Add a page test for `AttachmentSelector.xml` (and maybe also a docker test).

Related bug:
- https://jira.xwiki.org/browse/XWIKI-20181

### DIFF
--- a/xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-ui/src/main/resources/XWiki/AttachmentSelector.xml
+++ b/xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-ui/src/main/resources/XWiki/AttachmentSelector.xml
@@ -83,11 +83,9 @@ $xwiki.jsx.use($attachmentPickerDocName)
   #end
   (% class="gallery" %)(((
   ## Only display the upload form if they have edit permission on targetAttachDocument
-  #if ($xwiki.hasAccessLevel('edit',$xcontext.user,${targetAttachDocument.fullName}))
-    #attachmentPicker_displayUploadForm($targetDocument, $targetAttachDocument, $options)
-  #end
+  #attachmentPicker_displayUploadForm($targetDocument, $targetAttachDocument, $options)
   #attachmentPicker_displayAttachmentGalleryEmptyValue($targetDocument, $targetAttachDocument, $options, $currentValue)
-  #set ($sortedAttachments = $collectiontool.sort($targetAttachDocument.getAttachmentList(), "${options.sortAttachmentsBy}") )
+  #set ($sortedAttachments = $services.temporaryAttachments.listAllAttachments($targetAttachDocument))
   #foreach ($attachment in $sortedAttachments)
     #set ($extension = $attachment.getFilename())
     #set ($extension = $extension.substring($mathtool.add($extension.lastIndexOf('.'), 1)).toLowerCase())
@@ -149,8 +147,7 @@ $xwiki.jsx.use($attachmentPickerDocName)
 #macro (attachmentPicker_displayAttachmentDetails $attachment $options)
   #if ($attachment)
     ## Compute the attachment reference because there's no getter.
-    #set ($attachmentReference = $services.model.createAttachmentReference($attachment.document.documentReference,
-      $attachment.filename))
+    #set ($attachmentReference = $services.model.createAttachmentReference($attachment.doc.documentReference, $attachment.filename))
     #set ($attachmentStringReference = $services.rendering.escape($services.model.serialize($attachmentReference, 'default'), 'xwiki/2.1'))
     #if ($attachment.isImage() &amp;&amp; $options.displayImage)
       ## We add the version to the query string in order to invalidate the cache when an image attachment is replaced.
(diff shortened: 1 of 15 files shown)


manuelleduc:
Preview of how the temporary attachments are presented:

image

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/store/TemporaryAttachmentSessionsManager.java
```diff
@@ -57,6 +57,26 @@
...
+     * @return an attachment that is not saved yet but cached and contains the data of the given part
+     * @throws TemporaryAttachmentException if the part size exceeds the maximum upload size, or in case of problem
+     *     when reading the part
+     * @since 14.9RC1
+     */
+    @Unstable
+    default XWikiAttachment uploadAttachment(DocumentReference documentReference, Part part, String filename)
+        throws TemporaryAttachmentException
+    {
+        throw new UnsupportedOperationException();
```
tmortagne, line 77: This basically makes the default implementation totally useless.

manuelleduc: Sometimes, it's possible to return something that makes sense for new methods, the only two other options I have in mind are:
- do nothing
- call `uploadAttachment(DocumentReference documentReference, Part part)

Both of which are not acceptable, so it seems like the best option is to effectively warn the user that the called implementation does not provide an implementation for the new method.

surli: Why not just providing the new method without making it a default and with a revapi documentation?

manuelleduc: Because that's not what we documented (see https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HMakesureyourcodeisbackwardcompatible).

Also, I'm not sure the runtime error would be better with this option.

tmortagne: My point is that implementing that method just to throw an exception indicating that this method is not implemented does not make any kind of sense. All this false implementation does is hiding the fact that you add a breaking change.

> Both of which are not acceptable

What's important is knowing which is worst between a failing call or one of those options.

manuelleduc: Updated according to the new best practices, see:
- https://github.com/xwiki/xwiki-platform/pull/1930/files#diff-2f81214bd08060b9f2f1bdd7cb0dda0e7109ccadbf852cda6ddb864abc29c8e3R139
- https://github.com/xwiki/xwiki-platform/pull/1930/commits/7b74a0986a17a8c81717a11c7212b00a2a29136f#diff-9c6fd1ffb45cadd241e751aadc5b6944a03a59ee8994fafe52c6d9e8c4707ff3R74

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-ui/src/main/resources/XWiki/AttachmentSelector.xml
```diff
@@ -107,12 +110,23 @@ $xwiki.jsx.use($attachmentPickerDocName)
...
-    #set ($cssClass = 'gallery_image')
+  #set ($hasTemporaryAttachment = "$!services.temporaryAttachments" != '')
+  #set ($canEdit = $xwiki.hasAccessLevel('edit', $xcontext.user, ${targetAttachDocument.fullName}))
+  #set ($isTemporaryAttachment = false)
+  #if(!$hasTemporaryAttachment)
+    #set ($canDeleteAttachment = $canEdit)
   #else
-    #set ($cssClass = '')
+    #set ($isTemporaryAttachment = $services.temporaryAttachments.temporaryAttachmentExists($attachment))
+    #set ($canDeleteAttachment = !$isTemporaryAttachment &amp;&amp; $canEdit)
```
surli, line 120: Probably need a TODO comment for future improvment: it means that if you're using the macro in edit mode, then the temp attachments you added will probably be saved to the doc. So it would make sense to allow delete the temp attachment, especially since you should not require edit rights for those.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-ui/src/main/resources/XWiki/AttachmentSelector.xml
```diff
@@ -434,6 +463,68 @@ $xwiki.jsx.use($attachmentPickerDocName)
...
+        this.updateAttachment(response.fileName, response.url);
+        this.dialog.closeDialog();
+      }.bind(this)).fail(function  () {
+        #set ($errorMessage = $escapetool.javascript($services.localization.render('attachment.attachmentSelector.temporaryUpload.failure')))notification.replace(new XWiki.widgets.Notification("$errorMessage", 'error'));
+      });
+    }.bind(this));
+  }
+  
+  function directUploadAttachment(uploadForm) {
+    // FIXME This fails in HTML5, will deal with it later:
```
surli, line 517: what does it mean exactly? UC not supported with html5?

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManager.java
```diff
@@ -97,7 +104,13 @@ public XWikiAttachment uploadAttachment(DocumentReference documentReference, Par
         XWikiContext context = this.contextProvider.get();
         try {
             xWikiAttachment = new XWikiAttachment();
-            xWikiAttachment.setFilename(part.getSubmittedFileName());
+            String actualFilename;
+            if (filename != null) {
```
surli, line 108: Shouldn't we check for blank string too here?

manuelleduc: Indeed, could be safer.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-ui/src/main/resources/XWiki/AttachmentSelector.xml
manuelleduc: This is an old comment I moved when refactoring. Maybe @mflorea knows if this is still accurate.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-ui/src/main/resources/XWiki/AttachmentSelector.xml
manuelleduc: https://jira.xwiki.org/browse/XWIKI-20225 created, adding TODO.

manuelleduc: Done, see https://github.com/xwiki/xwiki-platform/pull/1930/files#diff-e1513599ab698991f6cbba55d38f3f464432ced8d137a668b1f7618c7e747e23R120

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManager.java
manuelleduc: Done https://github.com/xwiki/xwiki-platform/pull/1930/files#diff-f8de10bbbc165fc4673c7bf82e7240394b5f62e332c43d906c5a4536652591c5R109

surli (approved):

PR #1942, opened 2022-10-17
[misc] Simplify token generation in DefaultCSRFToken

manuelleduc:
- Replace several lines of synchronized code by a call to computeIfAbsent on a ConcurrentHashMap
- Code coverage ratio lowered because we do as much with less code

Note: applicable on master/14.4.x/13.10.x


michitux:
Can't you also remove the code before the synchronized block that tries getting the token first?

PR #1943, opened 2022-10-18
XWIKI-13399 delete UI does not have breadcrumbs, so going back to top level is not intuitive

ClemDee:
Added breadcrumbs to `copy.vm`, `delete.vm` and `rename.vm` templates.

For now, the page location displayed in the breadcrumbs is always the original page. This makes sense, because the url always stays to the original page too, as the action is bound to this page.
However, this can be a bit misleading in some situations, like after a rename / move operation, where the breadcrumb is now associated with the old path of the page. Maybe in that case it would be better to modify the breadcrumbs path to the renamed / moved page (even if the template is still bound to the original page location). wdyt?


manuelleduc:
Thanks @ClemDee, could you add some before/after screenshot please?

ClemDee:
## Copy

before:
image
image

after:
image
image


## Rename

before:
image
image

after:
image
image



## Delete

before:
image
image

after:
image
image

surli (approved):

mflorea (approved):

PR #1946, opened 2022-10-25
XWIKI-20289: Deleted terminal documents cannot be viewed

surli:
* Compare PageReferences to check if a revision should be returned
  * Improve XWikiAction to throw a proper XWikiException in case of a missing document revision, instead of storing a null doc in the
context: it should avoid NPE


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/doc/DeletedDocumentRevisionProvider.java
```diff
@@ -58,14 +64,29 @@ public XWikiDocument getRevision(DocumentReference reference, String revision) t
...
         }
 
         return null;
     }
 
+    private boolean areReferencesCompatible(DocumentReference requestedDocRef, DocumentReference deletedDocRef)
+    {
+        boolean result;
+        if (requestedDocRef == null) {
+            result = true;
```
surli, line 78: @tmortagne I kept this behaviour, but it feels a bit weird that we allow it. Our API in DocumentRevisionProvider is used with null references sometimes?

surli: Ok

PR #1951, opened 2022-10-28
XWIKI-20302: No error message in case of file uploads in comment without proper rights

surli:
* Check rights when performing file upload unless a specific attribute asks to skip this check

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/FileUploader.xml
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/FileUploader.xml
@@ -71,7 +71,7 @@
     #sendSuccess($document $fileName)
   #else
     ## Forbidden
-    #sendError(403 'You are not allowed to perform this action.')
+    #sendError(403 "$services.localization.render('ckeditor.upload.error.forbidden')")
   #end
 #end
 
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/FileUploader.xml
```diff
@@ -103,12 +103,17 @@
       #set ($document = $xwiki.getDocument($request.document))
     #end
     #set ($reference = $document.documentReference)
-    #set ($attachment = $services.temporaryAttachments.uploadTemporaryAttachment($reference, 'upload', 
-        $request.filename))
-    #if ($attachment)
-      #sendSuccess($document, $attachment.filename)
+    #set ($skipRightCheck = ($request.getHeader('X-XWiki-Temporary-Attachment-Skip-Right-Check') == 'true'))
```
michitux, line 106: Allowing file uploads without edit rights doesn't seem like a good idea as file uploads could easily lead to DoS attacks by filling the storage (which may be small in the case of cloud instances). I understand you probably need this for change requests but couldn't there at least be a configuration setting to allow this override? Or maybe add some listener that allows change request to explicitly authorize file uploads if the current user is allowed to create a change request for the current document instead of just always allowing file uploads?

surli:
We actually discussed that with @mflorea last week to use a different approach. Discussion available there: https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$1666964768642dPLSQ:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=caubin.fr

PR #1952, opened 2022-10-28
XWIKI-18989: Cannot upload attachments in annotation

surli:
* Implement attaching the temporary attachments to document in TemporaryAttachmentSessionManager
  * Use the new method in AnnotationRestResource after parsing the parameter


michitux:
Have you made sure that this still respects all rights? I'm just wondering because it seems that users with comment rights can add annotations but I don't think every user with comment right should be able to add attachments, and I've just tested this on 13.10.10 and there I indeed get an error when trying to upload an attachment as a user with just comment right.

mflorea (approved):

PR #1957, opened 2022-11-04
XWIKI-20321: Change default right scheme to stop granting all users script rights

surli:


surli:
I think that's all what is needed to fix this, but not sure.

tmortagne:
If we don't want to do anything during upgrade then that's all that is needed I think.

surli:
> If we don't want to do anything during upgrade (which is the case I think)

yes this is what has been voted. 
Thanks for the confirmation I'll merge as it is then.

PR #1959, opened 2022-11-07
XWIKI-20266: Ability to restrict attachment upload based on mime types

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20266

**Noticeable changes**

- [ ] Support for `objectPolicy` configuration in `AdminFieldsDisplaySheet.xml`
- [ ] Introduction of a `xwiki-platform-attachment-validation` module so that the attachment validation interfaces can be defined outside of oldcore
- [ ] New client side event `xwiki:actions:beforeUpload`
- [ ] Usage of always loaded but cached `XWiki.JavaScriptExtension` for the client side attachment validators
- [ ] Updated error management in FileUploader.xml/$services.temporaryAttachments (cc @surli)

**TODO**
- [x] Translations (client side error messages, configuration pages)
- [x] Tests
  - [x] Unit
  - [x] page test
  - [x] docker


xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.properties.vm
```diff
@@ -933,6 +933,23 @@ distribution.automaticStartOnWiki=$xwikiPropertiesAutomaticStartOnWiki
 #-# with programming rights, while this configuration affect any attachment.
 # attachment.download.forceDownload=
 
+#-# [Since 14.10RC1]
+#-# Define the list of allowed attachment mimetypes. By default, this list is empty, but you can specify a 
+#-# comma-separated list of allowed mimetypes. Note that the mimetypes are interpreted as regular expressions. Once this
+#-# list is not empty, any attachment with an unlisted mimetype will be rejected.
+#-# For instance, if you want to only accept plain text and any kind of images, you can set the list to:
+#-# text/plain,image/.*
+# attachment.mimetype.allowList=
```
tmortagne, line 942: The naming is not very consistent with existing properties, `attachment.download.whitelist` too is related to mime types.

Since it's about what you are allowed to upload, wouldn't it make more sense to name it something like `attachment.upload.allowList` (I assume the move from `whitelist` to `allowList` is for "using black vs white is not a great idea" reasons) ?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationException.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package org.xwiki.attachment.validation;
+
+/**
+ * Exception thrown in case of error while validating an attachment.
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+public class AttachmentValidationException extends Exception
```
surli, line 28: Missing unstable.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationException.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.attachment.validation;
+
+/**
+ * Exception thrown in case of error while validating an attachment.
```
surli, line 23: I'd have added a note specifying that this exception class is specifically used to hold message to be displayed on the frontend.

xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.properties.vm
manuelleduc: > (I assume the move from whitelist to allowList is for "using black vs white is not a great idea" reasons)

Exactly :)

I'm fine with `attachment.upload.allowList`, I'll update accordingly (with some delay to let other answer as well).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationScriptService.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Script service for the attachment validation. Provide the attachment validation configuration values.
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+@Component
+@Singleton
+@Named("attachment-validation")
```
surli, line 40: I don't recall having see `-` used in our script services names? We generally use a `.` as separator.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationScriptService.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.script.service.ScriptService;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Script service for the attachment validation. Provide the attachment validation configuration values.
```
surli, line 33: Only configuration values? I'm surprised there's no API to allow check validation of attachment property (e.g. check a mimetype or a size), but maybe it's not needed yet.

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManager.java
```diff
@@ -119,16 +113,24 @@ public XWikiAttachment uploadAttachment(DocumentReference documentReference, Par
             // document since it's a temporary attachment, but it is still useful to have a minimal knowledge of the
             // document it is stored for.
             xWikiAttachment.setDoc(new XWikiDocument(documentReference, documentReference.getLocale()), false);
+            this.attachmentValidator.validateAttachment(xWikiAttachment.getLongSize(), () -> {
+                try {
+                    return Optional.of(xWikiAttachment.getContentInputStream(this.contextProvider.get()));
+                } catch (XWikiException e) {
+                    this.logger.warn("Failed to get the input stream for attachment [{}]. Cause: [{}]", xWikiAttachment,
```
tmortagne, line 120: I find it very strange to basically ignore this problem, I would expect something that serious to fail the validation with a cause.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationScriptService.java
manuelleduc: I've used this name because using `attachment.validation` is trying to find `validation` is `AttachmentScriptService`.
The other option is to add a `get` method in `AttachmentScriptService` with a lookup for the "sub service" as we do for instance in `org.xwiki.export.script.ExportScriptService#get`

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidator.java
```diff
@@ -0,0 +1,61 @@
...
+@Unstable
+public interface AttachmentValidator
+{
+    /**
+     * Check if the part is a valid attachment in the current space.
+     *
+     * @param part the part to validate
+     * @throws AttachmentValidationException in case of error when validating the part
+     */
+    void validateAttachment(Part part) throws AttachmentValidationException;
```
surli, line 48: I think I'd have also provided an API taking an XWikiAttachment for extensions that directly manipulates the attachments (thinking about CR)

manuelleduc: I can't because otherwise we have a circular dependency between this module and oldcore. The alternative is to move this interface in oldcore (but that might lead to other issues).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/AttachmentValidationExceptionMapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Exception mapper for {@link AttachmentValidationException}.
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+@Component
+@Named("org.xwiki.attachment.validation.internal.AttachmentValidationExceptionMapper")
+@Provider
```
surli, line 43: I never used this annotation in a component definition. Out of curiosity why do you need it?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/FileUploader.xml
```diff
@@ -103,12 +103,21 @@
...
+    #elseif ("$!uploadTemporaryAttachmentException" != '')
+      ## The exception is wrapped in an org.apache.velocity.exception.MethodInvocationException
+      #if ($uploadTemporaryAttachmentException.cause.translationKey)
+        #sendError(400 "$services.localization.render($uploadTemporaryAttachmentException.cause.translationKey)")
+      #else
+        #sendError(400 "$services.localization.render('ckeditor.upload.error.emptyReturn')")
+      #end
     #else
-      #sendError(400 "$services.localization.render('ckeditor.upload.error.emptyreturn')")
+      #sendError(400 "$services.localization.render('ckeditor.upload.error.emptyReturn')")
```
tmortagne, line 120: Is this translation key change really expected ?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentMimetypeRestrictionSpaceConfigurationSource.java
```diff
@@ -0,0 +1,84 @@
...
+@Singleton
+@Named(DefaultAttachmentMimetypeRestrictionSpaceConfigurationSource.HINT)
+public class DefaultAttachmentMimetypeRestrictionSpaceConfigurationSource extends AbstractDocumentConfigurationSource
+{
+    /**
+     * Hint of this configuration source.
+     */
+    public static final String HINT = "attachment.mimetypeRestriction.space";
+
+    private static final String SPACE_PREFERENCES = "WebPreferences";
```
surli, line 50: Sounds like a constant that should be available somewhere in XWiki already no?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/FileUploader.xml
manuelleduc: This is an unrelated bug that I've fixed along the way. I should probably commit that separately.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentMimetypeRestrictionSpaceConfigurationSource.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Inject
+    private DocumentAccessBridge documentAccessBridge;
+
+    @Override
+    protected DocumentReference getDocumentReference()
+    {
+        // Note: We would normally use a Reference Resolver here but since the Model module uses the Configuration
+        // module we cannot use one as otherwise we would create a cyclic build dependency...
```
surli, line 59: Not sure to understand this comment: you're not in the configuration module here, you're in a specific module for attachment-validation and I doubt there's a dep between model and this new module...

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+@Component
+@Singleton
+public class DefaultAttachmentValidator implements AttachmentValidator
+{
+    @Inject
+    @Named("readonly")
```
surli, line 64: Using a readonly context for perf reason?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
```diff
@@ -0,0 +1,153 @@
...
+    {
+        long size = part.getSize();
+        // We don't check the mimetype for parts that are not expected to be use as file.
+        if (StringUtils.startsWith(part.getName(), UploadAction.FILE_FIELD_NAME)) {
+            validateAttachment(size, () -> {
+                try {
+                    return Optional.of(part.getInputStream());
+                } catch (IOException e) {
+                    this.logger.warn("Failed to get the input stream for part [{}]. Cause: [{}]", part,
+                        getRootCauseMessage(e));
```
surli, line 84: Add also a logger.debug message with the full stack: it can be quite useful in  case of problems.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
manuelleduc: Mainly because currently there is no reason to do anything else than reading the context in this class.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
```diff
@@ -0,0 +1,153 @@
...
+        String mimeType;
+        try {
+            Optional<InputStream> inputStream = supplier.get();
+            if (inputStream.isEmpty()) {
+                mimeType = "";
+            } else {
+                mimeType = TikaUtils.detect(inputStream.get(), fileName);
+            }
+        } catch (Exception e) {
+            this.logger.warn("Failed to identify the mimetype of [{}]. Cause: [{}]", fileName, getRootCauseMessage(e));
```
surli, line 148: Same here for the logger.debug

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/FileSizeValidation.xml
```diff
@@ -0,0 +1,317 @@
...
+    <property>
+      <code>define('attachment-validation-filesize-messages', {
+  prefix: 'attachment.validation.filesize.',
+  keys: [
+    'errorMessage'
+  ]
+});
+
+require(['jquery', 'xwiki-l10n!attachment-validation-filesize-messages', 'xwiki-events-bridge'], ($, l10n) =&gt; {
+  function formatBytes(bytes) {
```
surli, line 135: Maybe something to provide more globally to XWiki than just here.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationScriptService.java
tmortagne: Yes `.` is a very special separator and it needs to make sense as a sub service (and I guess it does here since we have an `attachment` script service).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/FileSizeValidation.xml
```diff
@@ -0,0 +1,317 @@
...
+    const k = 1024
+    const sizes = ['bytes', 'Kb', 'Mb', 'Gb', 'Tb', 'Pb', 'Eb', 'Zb', 'Yb']
+
+    const i = Math.floor(Math.log(bytes) / Math.log(k))
+
+    return `${parseFloat((bytes / Math.pow(k, i)).toFixed(2))}${sizes[i]}`
+  }
+
+  $(document).on('xwiki:actions:beforeUpload', (event, data) =&gt; {
+    const configElement = document.getElementById('attachment-validation-filesize-configuration');
```
surli, line 150: why using `document.getElementById` and not `$('#..')` here?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/FileSizeValidation.xml
manuelleduc: Since this is the only place requiring this currently I'm not sure this is worth the effort. But +1 to move this somewhere reusable if it becomes useful elsewhere.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/FileSizeValidation.xml
```diff
@@ -0,0 +1,317 @@
...
+    </property>
+    <property>
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content>{{velocity}}
+(% id='attachment-validation-filesize-configuration' class='hidden' %)
```
surli, line 298: I think there was a proposal from @michitux to avoid using wiki syntax in code related pages to avoid future problems with escaping etc. 
I don't see any possible escaping problem with your current code, but maybe better to take the habit to always use plain html for such thing?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidator.java
tmortagne: It should not be too hard for CR to use `#validateAttachment(long attachmentSize, Supplier<Optional<InputStream>> supplier, String filename)`. Even `#validateAttachment(Part)` is actually just a helper here, and I'm wondering if it's really correct to see something specific to the servlets in the AttachmentValidator API.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-upload/plugin.js
```diff
@@ -121,28 +121,34 @@
...
-    // * the loadend event listener is not called if added after the upload request was opened (initialized);
-    //   see https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#monitoring_progress
-    // * another event listener might send the upload request before we overwrite the send function;
-    //   see https://ckeditor.com/docs/ckeditor4/latest/guide/dev_file_upload.html#request-2
+    editor.on('fileUploadRequest', function (event) {
+      var eventFileSelected = $(document).trigger('xwiki:actions:beforeUpload', {
+        file: event.data.fileLoader.file,
+      });
+      // We only queue the file and continue the upload if it is properly validated.
+      if (eventFileSelected.returnValue) {
```
surli, line 131: Where this `returnValue` comes from?  Sounds like it's not a jquery thing, and I don't recall having seen this kind of construction before.

manuelleduc: This is part of the event API. If any of the listener returns false, `returnValue` is false (cc @mflorea if you know a better way to achieve this).

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseAttachmentsResource.java
```diff
@@ -348,25 +360,34 @@ protected Attachments getAttachmentsForDocument(Document doc, int start, int num
...
+            // The doc has been updated during the creation of the attachment, so we need to ensure we answer with the
+            // updated version.
+            Document updatedDoc = xwikiAttachment.getDoc().newDocument(xcontext);
+            Attachment attachment = this.modelFactory.toRestAttachment(this.uriInfo.getBaseUri(),
+                new com.xpn.xwiki.api.Attachment(updatedDoc, xwikiAttachment, this.xcontextProvider.get()),
+                withPrettyNames,
+                false);
+
+            return new AttachmentInfo(attachment, alreadyExisting);
+        } finally {
```
surli, line 384: What's the relation of this try/finally with the new feature?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseAttachmentsResource.java
```diff
@@ -381,6 +402,18 @@ protected XWikiAttachment createOrUpdateAttachment(AttachmentReference attachmen
                 String.format("Failed to create or update the attachment [%s].", attachmentReference), e);
         }
 
+        // We wait for the XWikiAttachment to be loaded before checking the size as accurately checking the size of 
+        // the content input stream is not possible.
+        this.attachmentValidator.validateAttachment(attachment.getLongSize(), () -> {
```
surli, line 407: hmmm so you wouldn't need this if you had the API taking an XWikiAttachment.

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManager.java
```diff
@@ -58,21 +61,18 @@ public class DefaultTemporaryAttachmentSessionsManager implements TemporaryAttac
     @Inject
     private Provider<XWikiContext> contextProvider;
 
+    @Inject
+    private AttachmentValidator attachmentValidator;
```
surli, line 65: Does not need to be in a `Provider` AFAIR the `DefaultTemporaryAttachmentSessionManager` is injected quite early no?

tmortagne:
It's binary compatible from runtime point of view, of course that code using those APIs might receive this new exception and not behave unexpectedly.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
surli: Ok, wondering about the perf impact of using one instead of the other: pretty sure that we could use `readonly` context in many more components. @tmortagne an idea about that?

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManager.java
manuelleduc: This is an annotation provided by jsr311, this is part of its extensibility mechanism. It's used to scan annotations to discover "components".

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
tmortagne: The point of the "readonly" XWikiContext provider is not really performances. As documented on ReadOnlyXWikiContextProvider the only difference is that, if there is no XWikiContext in the current ExecutionContext, you won't get a generated one, so you need to be prepared for the eventuality that it might return null (which I don't think is the case here).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/AttachmentValidationExceptionMapper.java
manuelleduc: This is an annotation provided by jsr311, this is part of its extensibility mechanism. It's used to scan annotations to discover "components".

surli: Ok

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationScriptService.java
manuelleduc: Renamed `attachmentValidation`.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentMimetypeRestrictionSpaceConfigurationSource.java
manuelleduc: Surprisingly, we don't have (yet) this constant defined somewhere reusable.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/resources/BaseAttachmentsResource.java
manuelleduc: The attachment validation requires to have the document where the attachment will be uploaded in the context. I'll add a comment.

xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/filesystem/internal/DefaultTemporaryAttachmentSessionsManager.java
manuelleduc: I did not notice any issue with that yet.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentMimetypeRestrictionSpaceConfigurationSource.java
surli: Hmmm indeed, that's surprising. After second thought, I'm wondering if we shouldn't have just an API returning the DocumentReference of the SpacePreferences storage of a space.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/FileSizeValidation.xml
surli: ok

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/fileupload/FileUploadUtils.java
```diff
@@ -144,4 +153,43 @@ public FileItem createItem(String fieldName, String contentType, boolean isFormF
             }
         }
     }
+
+    private static class PartAttachmentValidationSupplier implements AttachmentValidationSupplier
```
surli, line 157: Really not a fan of this: sounds like something that we might want to be able to reuse, and not only there...

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.InputStream;
+
+/**
+ * Provides the metadata required to validate an attachment .
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+public interface AttachmentValidationSupplier
```
surli, line 30: Missing unstable. To be honest I'm not entirely sure about the name / the location of this API. We could definitely reuse the same interface in TemporaryAttachmentSessionManager.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-default/src/main/java/org/xwiki/attachment/validation/internal/DefaultAttachmentValidator.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+@Component
+@Singleton
+public class DefaultAttachmentValidator implements AttachmentValidator
+{
+    private static final List<String> KNOWN_STEPS =
+        List.of(FileSizeAttachmentValidationStep.HINT, MimetypeAttachmentValidationStep.HINT);
+
+    @Inject
+    private ComponentManager componentManager;
```
surli, line 58: I think you need the context CM here if you want to load components installed by extensions in a subwiki.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidator.java
manuelleduc: Done, see https://github.com/xwiki/xwiki-platform/pull/1959/files#diff-b3291e0da5c23912c8ca043e8ce77635486cf9ef433b28d6a7db23e101252f13

michitux (commented):
I'm wondering about the reasoning for using a list of regular expressions that need to be both valid in Java and in JavaScript (which is not documented, btw.). Wouldn't a single regular expression be enough instead of a list?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/MimetypeValidation.xml
```diff
@@ -0,0 +1,320 @@
...
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <content>{{velocity wiki='false'}}
+<div id='attachment-validation-mimetypes-configuration' class='hidden'>$jsontool.serialize({
+  "allowedMimetypes": $services.attachmentValidation.getAllowedMimetypes(),
+  "blockerMimetypes": $services.attachmentValidation.getBlockerMimetypes()
```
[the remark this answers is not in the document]
manuelleduc: > I'm wondering about the reasoning for using a list of regular expressions that need to be both valid in Java and in JavaScript (which is not documented, btw.). Wouldn't a single regular expression be enough instead of a list?

- Having a list of expression better represent the underlying logic
- Admins that don't know regex, but knows how to define lists in xwiki are still able to enumerate all options.

Another option could be to remove the notion of regex and simply use `*` as a joker, because imo for 99% if the cases, this will be enough.

manuelleduc:
> I'm wondering about the reasoning for using a list of regular expressions that need to be both valid in Java and in JavaScript (which is not documented, btw.). Wouldn't a single regular expression be enough instead of a list?

- Having a list of expression better represent the underlying logic
- Admins that don't know regex, but knows how to define lists in xwiki are still able to enumerate all options.

Another option could be to remove the notion of regex and simply use `*` as a joker, because imo for 99% if the cases, this will be enough.

xwiki-platform-tools/xwiki-platform-tool-configuration-resources/src/main/resources/xwiki.properties.vm
manuelleduc: done

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-ui/src/main/resources/XWiki/Attachment/Validation/Code/PageAdmin/MimetypeValidation.xml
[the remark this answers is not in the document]
manuelleduc: Escaping added.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/attachment/AttachmentAccessWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.xwiki.bridge.attachment;
+
+import java.io.InputStream;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provides the metadata required to validate an attachment.
```
surli, line 28: Sounds like outdated javadoc no? It's not only about validation now.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/attachment/AttachmentAccessWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+@Role
```
surli, line 33: Well it's supposed to be an interface implemented by `XWikiAttachment` so not really a component?

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/attachment/AttachmentAccessWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+@Unstable
+public interface AttachmentAccessWrapper
```
surli, line 35: Wouldn't `AttachmentAccessBridge` would be better, to be aligned with `DocumentAccessBridge`? And actually shouldn't it be `AttachmentModelBridge`?

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/attachment/AttachmentAccessWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @return the attachment size in bytes
+     */
+    long getSize();
+
+    /**
+     * @return the attachment input steam
+     * @throws AttachmentAccessWrapperException in case of error when retrieving the input steam
+     */
+    InputStream getInputStream() throws AttachmentAccessWrapperException;
```
surli, line 46: Why not just throwing an IOException here?

manuelleduc: The goal is to make the exception as specific as possible. But I'd be ok to switch to an IOException.

surli:
> Another option could be to remove the notion of regex and simply use * as a joker, because imo for 99% if the cases, this will be enough.

Maybe something to check on the forum proposal, but I'd be +1 as it would scope a bit more the feature and avoid unintended behaviour with some weird regex

manuelleduc:
> > Another option could be to remove the notion of regex and simply use * as a joker, because imo for 99% if the cases, this will be enough.
> 
> Maybe something to check on the forum proposal, but I'd be +1 as it would scope a bit more the feature and avoid unintended behavior with some weird regex

See the corresponding forum post here: https://forum.xwiki.org/t/mimetype-attachment-restriction/11259/3?u=mleduc

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/attachment/XWikiAttachmentAccessWrapper.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Attachment wrapper for {@link XWikiAttachment}. Provides access to various metadata required to access an attachment.
+ * For instance, its size, name, and an input steam of its content.
+ *
+ * @version $Id$
+ * @since 14.10RC1
+ */
+@Unstable
+public class XWikiAttachmentAccessWrapper implements AttachmentAccessWrapper
```
surli, line 39: Do you need those implementations to be API? Wondering if it shouldn't stay internal.

manuelleduc: @tmortagne wdyt. iirc you mentioned reusing this of replication, and in this case I guess it would be useful to already have a wrapper available.

tmortagne: > you mentioned reusing this for replication

Hmm, does not ring a bell, actually. Do you remember when that was ?

Right now, I'm tempted to agree with @surli.

manuelleduc: > Hmm, does not ring a bell, actually. Do you remember when that was ?

Can't find it, I'm probably mixing things up.

Making those internal then!

PR #1965, opened 2022-11-10
XWIKI-20346: Allow same image be referenced in several sections when splitting document.

AndreasJonsson:


xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/splitter/DocumentSplitterUtils.java
```diff
@@ -84,22 +83,20 @@ public static NamingCriterion getNamingCriterion(String namingCriterionId, Strin
      * 
      * @param sectionDoc the newly created wiki document corresponding to a section of the original office document
      * @param officeDocument the office document being splitted into wiki documents
+     * @param fileMap Map of files from the original document
      * @return the relocated artifacts
      */
-    public static Set<File> relocateArtifacts(WikiDocument sectionDoc, XDOMOfficeDocument officeDocument)
+    public static Set<File> relocateArtifacts(WikiDocument sectionDoc, XDOMOfficeDocument officeDocument,
```
mflorea, line 89: This method doesn't do any relocation. With your changes it's more of a ``getSectionArtifacts``. Also, if you pass ``fileMap`` then you don't need ``officeDocument`` anymore right?

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/splitter/DocumentSplitterUtils.java
```diff
@@ -84,22 +83,20 @@ public static NamingCriterion getNamingCriterion(String namingCriterionId, Strin
...
+        Map<String, File> fileMap)
     {
-        Set<File> artifacts = officeDocument.getArtifactsFiles();
         Set<File> result = new HashSet<>();
         List<ImageBlock> imageBlocks =
             sectionDoc.getXdom().getBlocks(new ClassBlockMatcher(ImageBlock.class), Axes.DESCENDANT);
         if (!imageBlocks.isEmpty()) {
```
mflorea, line 95: This IF is not needed anymore after your changes.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/splitter/DocumentSplitterUtils.java
AndreasJonsson: Yes, the method adds files subdocument. There is no need to remove files from the original document as it will be discarded anyway.
So, the only necessary change to fix the issue is to leave the file in the original set.
But I also moved the map generation outside the method since it makes no sense to reconstruct it for every document.

xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/splitter/DocumentSplitterUtils.java
```diff
@@ -78,16 +78,17 @@ public static NamingCriterion getNamingCriterion(String namingCriterionId, Strin
     }
 
     /**
-     * Move artifacts (i.e. embedded images) from the original office document to a specific wiki document corresponding
-     * to a section. Only the artifacts from that section are moved.
+     * Add artifacts (i.e. embedded images) from the original office document to a specific wiki document corresponding
```
mflorea, line 81: This is misleading because I don't see any ``sectionDoc.add*`` call below (to justify that the artifacts are added to the section document). AFAICS the artifacts are not added (by this method) but returned, so why not call it ``getArtifacts`` or ``getSectionArtifacts``?

AndreasJonsson: That's true.  I changed the name to getSectionArtifacts.

mflorea:
Looks good. There is only one comment left to resolve, the unneeded IF.

PR #1975, opened 2022-11-24
XWIKI-20399: RequestParameterConverter should not always handle errors with ServletResponse

surli:
* Provide an API to allow use RequestParameterConverter and let consumer of the API decides how to handle errors

### DIFF
--- a/xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConversionResult.java
+++ b/xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConversionResult.java
@@ -1,0 +1,78 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.wysiwyg.converter;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.xwiki.stability.Unstable;
+import org.xwiki.wysiwyg.filter.MutableServletRequest;
+
+/**
+ * Simple POJO holding the result of a conversion performed with {@link RequestParameterConverter}.
+ * More specifically this class contains a mutable request, resulting of the conversion, a map of errors that might have
+ * occurred during the conversion for each parameter, and a map of the output of the conversion for each parameter.
+ *
+ * @version $Id$
+ * @since 14.10
+ */
+@Unstable
+public class RequestParameterConversionResult
+{
+    private MutableServletRequest request;
+    private Map<String, Throwable> errors;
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConverter.java
```diff
@@ -40,13 +40,34 @@
...
+     * @param request the request that might contain parameter needing conversion
+     * @return an instance of {@link RequestParameterConversionResult} containing the modified request and the output
+     *         and errors that might have occurred
+     * @throws IOException in case of problem to write an answer in the response
+     * @since 14.10
+     */
+    @Unstable
+    default RequestParameterConversionResult convert(ServletRequest request) throws IOException
+    {
+        return null;
```
manuelleduc, line 71: imo it's better to have either no default implementation, or a exception thrown, instead of a null value.

tmortagne, line 71: Isn't returning null incompatible with the javadoc ?

xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConversionResult.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Default constructor.
+     *
+     * @param request a mutable copy of the original request used for the conversion
+     */
+    public RequestParameterConversionResult(MutableServletRequest request)
+    {
+        this.request = request;
+        this.errors = new LinkedHashMap<>();
```
manuelleduc, line 51: Out of curiosity, why `LinkedHashMap` over `HasMap`? If this is for a specific reason, I think it deserve a documentation.

xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConverter.java
surli: Yes indeed. Will change

xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConversionResult.java
surli: Just to keep same order in case of iterating. It's also easier to test for the future.

xwiki-platform-core/xwiki-platform-wysiwyg/xwiki-platform-wysiwyg-api/src/main/java/org/xwiki/wysiwyg/converter/RequestParameterConverter.java
```diff
@@ -40,13 +40,33 @@
...
+     *
+     * @param request the request that might contain parameter needing conversion
+     * @return an instance of {@link RequestParameterConversionResult} containing the modified request and the output
+     *         and errors that might have occurred
+     * @since 14.10
+     */
+    @Unstable
+    default RequestParameterConversionResult convert(ServletRequest request)
+    {
+        throw new RuntimeException("Not implemented.");
```
manuelleduc, line 70: `UnsupportedOperationException` would be better imo, and that's what currently documented https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HMakesureyourcodeisbackwardcompatible

tmortagne: No, it's not better, and it's not what is documented.

surli: Fixed.

PR #1991, opened 2022-12-02
XWIKI-20205: Clicking the Global Search button also activates the Live Data actions button

raphj:
Jira: https://jira.xwiki.org/browse/XWIKI-20205

This might require Vue tests (said @manuelleduc)


manuelleduc:
The documentation for `aria-expanded` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded

PR #1997, opened 2022-12-06
XWIKI-20426: The backlinks section of the delete UI no longer displayed backlink documents

oanalavinia:
* display the backlinks URL at the beginning of the panel, not after the new target is selected
* deprecate translation key
* delete declaration of $message since it wasn't used anymore

Note that I kept that idea of having a link that redirects the user to a backlinks page rather than displaying the full list on the current panel, which is also consistent with the rename operation https://www.xwiki.org/xwiki/bin/download/Documentation/UserGuide/Features/DocumentLifecycle/WebHome/RenameConfirmFlamingoAdvanced.png?rev=1.2

The backlinks panel when no target is selected:
backlinksPanelWithoutTarget

The backlinks panel with a selected target:
backlinksPanelWithTarget

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
@@ -229,15 +229,6 @@
      <strong>$services.localization.render('core.delete.backlinks')</strong> ($links.size())
     #end
     #define($body)
-      <p class="noitems">
-        #set ($backLinksURL = $escapetool.xml($doc.getURL('view', 'viewer=backlinks')))
-        #set ($backlinksMessage = $escapetool.xml($services.localization.render('core.delete.backlinksMessage', [
-          '__STARTLINK__',
-          $doc.backlinks.size(),
-          '__ENDLINK__'
-        ])).replace('__STARTLINK__', "<a href='$backLinksURL'>").replace('__ENDLINK__', '</a>'))
-        $backlinksMessage
-      </p>
       <dl>
         <dt>
           <label for="newBacklinkTarget">
@@ -258,17 +249,16 @@
         #displayLinksCheckbox({
           'label': 'core.delete.updateLinks.label',
           'hint': 'core.delete.updateLinks.hint',
-          'addBacklinksURL': false,
-          'hidden': true
+          'disabled': true
         })
         #displayAutoRedirectCheckbox({
           'label': 'core.delete.autoRedirect.label',
           'hint': 'core.delete.autoRedirect.hint',
-          'hidden': true
+          'disabled': true
         })
       </dl>
     #end
-    #displayPanel('panel-backlinks', 'panel-default', $heading, $body)
+    $body
     #set($hasInlinks = true)
   #end
 #end
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
@@ -231,8 +231,12 @@
     #define($body)
       <p class="noitems">
         #set ($backLinksURL = $escapetool.xml($doc.getURL('view', 'viewer=backlinks')))
-        $services.localization.render('core.delete.backlinksMessage',
-          ["<a href='$backLinksURL'>", $doc.backlinks.size(), '</a>'])
+        #set ($backlinksMessage = $escapetool.xml($services.localization.render('core.delete.backlinksMessage', [
+          '__STARTLINK__',
+          $doc.backlinks.size(),
+          '__ENDLINK__'
+        ])).replace('__STARTLINK__', "<a href='$backLinksURL'>").replace('__ENDLINK__', '</a>'))
+        $backlinksMessage
       </p>
       <dl>
         <dt>
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
@@ -228,17 +228,11 @@
     #define($heading)
      <strong>$services.localization.render('core.delete.backlinks')</strong> ($links.size())
     #end
-    #define($message)
-      <ul>
-      #foreach($docname in $links)
-        #set($rdoc = $xwiki.getDocument($docname).getTranslatedDocument())
-        <li><a href="$rdoc.getURL('view')">$escapetool.xml($rdoc.getPlainTitle())</a></li>
-      #end
-      </ul>
-    #end
     #define($body)
       <p class="noitems">
-        $escapetool.xml($services.localization.render('core.delete.backlinksInfo', [$doc.backlinks.size()]))
+        #set ($backLinksURL = $escapetool.xml($doc.getURL('view', 'viewer=backlinks')))
+        $services.localization.render('core.delete.backlinksMessage',
+          ["<a href='$backLinksURL'>", $doc.backlinks.size(), '</a>'])
       </p>
       <dl>
         <dt>
@@ -260,6 +254,7 @@
         #displayLinksCheckbox({
           'label': 'core.delete.updateLinks.label',
           'hint': 'core.delete.updateLinks.hint',
+          'addBacklinksURL': false,
           'hidden': true
         })
         #displayAutoRedirectCheckbox({
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
@@ -224,11 +224,9 @@
 #######################################################
 #macro(displayNewTargetOption)
   #set($links = $doc.getBacklinks())
-  #if($links && $links.size() > 0)
-    #set($hasInlinks = true)
-  #end
+  #set ($hasInlinks = $links && $links.size() > 0)
   #define($heading)
-    <strong>$services.localization.render('core.delete.backlinks')</strong> ($links.size())
+    <strong>$escapetool.xml($services.localization.render('core.delete.backlinks'))</strong> ($links.size())
   #end
   #define($body)
     <dl>
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
@@ -123,7 +123,7 @@
 #macro(displayConfirmationPage)
   <form id="delete" class="xform" action="$doc.getURL('delete', "$!{languageparams}")" method="post">
     #getChildren()
-    #getBacklinks()
+    #displayNewTargetOption()
     #getChildren_legacy()
     #displayConfirmMessage()
   </form>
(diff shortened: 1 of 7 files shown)


vmassol:
> Note that I kept that idea of having a link that redirects the user to a backlinks page rather than displaying the full list on the current panel, which is also consistent with the rename operation https://www.xwiki.org/xwiki/bin/download/Documentation/UserGuide/Features/DocumentLifecycle/WebHome/RenameConfirmFlamingoAdvanced.png?rev=1.2

Shouldn't we use the same strategy for all parts of the delete UI? Right now, the Children part of the UI shows a LT.

Also, the rename UI is not using Accordions, which is the main reason IMO that there's a link.

I think we probably need another PR to improve the delete UI to be more similar to the Rename/Move/Copy UIs.

I guess this PR could be a step in that direction.Do you think you would have some time to work on that new PR to improve the delete UI? :) 

Thanks Lavinia!

vmassol:
Thanks for fixing the issue!

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -228,17 +228,11 @@
...
-        #set($rdoc = $xwiki.getDocument($docname).getTranslatedDocument())
-        <li><a href="$rdoc.getURL('view')">$escapetool.xml($rdoc.getPlainTitle())</a></li>
-      #end
-      </ul>
-    #end
     #define($body)
       <p class="noitems">
-        $escapetool.xml($services.localization.render('core.delete.backlinksInfo', [$doc.backlinks.size()]))
+        #set ($backLinksURL = $escapetool.xml($doc.getURL('view', 'viewer=backlinks')))
+        $services.localization.render('core.delete.backlinksMessage',
```
vmassol, line 234: I think you need HTML escape here (`$escapetool.xml(...)`) done on the results of the call to `render()`.

oanalavinia: I'm not sure what escaping should be used in order to not broke the link display 
image

michitux: You need to first use some non-HTML placeholder and then replace that placeholder after the escaping, see, e.g., here: https://github.com/xwiki/xwiki-platform/blob/5a4155e351dc2fa8adf46a4972587e39ce8cd74b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/packagelist.vm#L46-L50

I'm aware this is not nice, we'll work on a better solution.

oanalavinia:
> Note that I kept that idea of having a link that redirects the user to a backlinks page rather than displaying the full list on the current panel, which is also consistent with the rename operation https://www.xwiki.org/xwiki/bin/download/Documentation/UserGuide/Features/DocumentLifecycle/WebHome/RenameConfirmFlamingoAdvanced.png?rev=1.2
> 
> Shouldn't we use the same strategy for all parts of the delete UI? Right now, the Children part of the UI shows a LT.
> 
> Also, the rename UI is not using Accordions, which is the main reason IMO that there's a link.
> 
> I think we probably need another PR to improve the delete UI to be more similar to the Rename/Move/Copy UIs.
> 
> I guess this PR could be a step in that direction.Do you think you would have some time to work on that new PR to improve the delete UI? :)
> 
> Thanks Lavinia!

Yes, I could work on it as part of separate issue

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
michitux: I think this `replaceAll` is not a best practice as this triggers regular expression parsing, this should be `replace`.

oanalavinia: Thanks @michitux ! Updating it

surli (approved):

oanalavinia:
I agree, with this implementation the `automatic redirect` option is linked to the backlinks, even if are independent options. The panel is also hidden if there are no backlinks. But I think it would be better to address it in a separate issue, as improvement

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1483,7 +1483,7 @@ activitystream.event.deleteComment.rss.body=A comment has been deleted from the
 ### Deleting a page
 core.delete=Delete
 core.delete.title=Delete {0}
-core.delete.backlinksInfo=After deleting this page, the {0,choice,0#incoming links|1#incoming link|1<incoming links} will point to an empty page. To avoid this, you can select a new target.
+core.delete.backlinksMessage=After deleting this page, the {0} {1,choice,0#incoming links|1#incoming link|1<incoming links}{2} will point to an empty page. To avoid this, you can select a new target.
```
vmassol, line 1486: Is it normal that there's a space between `{0}` and `{1`?

Also, didn't we have a link before? If so where is the key that was used? Cannot we reuse it?

oanalavinia: We just listed the backlinks [before](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties#L5550), so there is no translation to reuse.
I removed the extra space, it wasn't indeed needed

mflorea:
> Just display the update links and create redirect options independently of the selection of the new target and make the introductory sentence more generic to also include the use case of external links?

@michitux I agree that the current UI is awkward (mostly because it hides the new target selection under the backlinks accordion) but I don't understand your proposal. We cannot update the links or create a redirect without having a new target. The backlinks have to be updated to point to the new target. The created redirect should redirect to the new target. So these two options, update links and create redirect, make sense only when a new target is selected, which is why I think the current behavior of showing them only after a target is selected is OK. If we show them all the time **enabled** then the users will ask "why wasn't the redirect created and the links updated? I checked the corresponding options!" (because you didn't specify a new target...).

I think we should:
* Move New Target, Update Links and Create an Automatic Redirect outside of the Backlinks accordion, preserving the current behavior of showing Update Links and Create an Automatic Redirect only when New Target has a value.
* Use the current text before New Target as the hint for the New Target field, modified to
  * mention external links as @michitux pointed about
  * include a link to check the backlinks (the one from the Update Links hint)
  * and maybe specify the behavior for no value set (basically a merge between the current hint and the text before New Target). My point is that we don't need two hints.
* Alternatively we can show Update Links and Create an Automatic Redirect all the time but **disabled** as long as New Target has no value set. In this case the link to check the backlinks can remain on the Update Links hint.
* Remove the Backlinks accordion.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1483,7 +1483,7 @@ activitystream.event.deleteComment.rss.body=A comment has been deleted from the
 ### Deleting a page
 core.delete=Delete
 core.delete.title=Delete {0}
-core.delete.backlinksInfo=After deleting this page, the {0,choice,0#incoming links|1#incoming link|1<incoming links} will point to an empty page. To avoid this, you can select a new target.
+core.delete.backlinksMessage=After deleting this page, the {0}{1,choice,0#incoming links|1#incoming link|1<incoming links}{2} will point to an empty page. To avoid this, you can select a new target.
```
mflorea, line 1486: This also won't be needed if the hint of Update Links is always visible and has the link to check the backlinks.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -5563,6 +5563,11 @@ core.menu.export.rtf=Export as RTF
 core.menu.export.html=Export as HTML
 core.menu.export.xar=Export as XAR
 
+#######################################
+## until 14.4.8, 14.10.1, 15.0RC1
+#######################################
+core.delete.backlinksInfo=After deleting this page, the {0,choice,0#incoming links|1#incoming link|1<incoming links} will point to an empty page. To avoid this, you can select a new target.
```
mflorea, line 5569: I think we'll have to bring this key back if we show the links to check the backlinks in the Update Links hint. So I'm starting to think that it's best if we do the changes in this PR.

michitux:
> We cannot update the links or create a redirect without having a new target. The backlinks have to be updated to point to the new target. The created redirect should redirect to the new target. So these two options, update links and create redirect, make sense only when a new target is selected, which is why I think the current behavior of showing them only after a target is selected is OK. If we show them all the time **enabled** then the users will ask "why wasn't the redirect created and the links updated? I checked the corresponding options!" (because you didn't specify a new target...).

To me it is just really surprising that a checkbox for creating a redirect appears after selecting the target while I was previously completely unaware that this option existed which is very bad for usability. I would have modeled this dependency rather using form validation on the client and server side, I mean when selecting one of the two options, the target becomes mandatory. I'm not sure disabling the inputs doesn't also create strange situations, like you first select a target, then enable the redirect, then remove the target again. In this situation, from what you describe I would think the redirect option is still selected but now disabled so the user cannot even disable the option anymore.

mflorea:
> To me it is just really surprising that a checkbox for creating a redirect appears after selecting the target while I was previously completely unaware that this option existed which is very bad for usability.

@michitux what is also bad for usability is showing too many configuration options or options that are used only in some cases. In general, I don't think it's bad to hide and show some configuration options when some conditions are met. For instance, for the PDF export we're not showing the Docker settings unless you choose to use Docker for performing the PDF export.

Being unaware that an option exists is not a problem unless you need that option. But if you want to create a redirect then you must have a target for the redirect in mind, or at least you must be aware that you need a redirect target, so seeing the target option should ring a bell. I agree that it's more natural to first ask "Do you want a redirect?" and then "Where to?", but since the "Where to?" part is shared by multiple options (create redirect and update links), we reversed the questions. We first ask "Has this page been replaced by another one?" and if so, "Do you want to create a redirect to the other page?" and "Do you want to update the links to point to the other page?". I don't think it's that bad.

> I would have modeled this dependency rather using form validation on the client and server side, I mean when selecting one of the two options, the target becomes mandatory.

I'm not convinced that this is a better solution. In general I think having less mandatory fields (and relying more on defaults) is better. If I can avoid a mandatory field I will. Moreover, changing the mandatory flag of a field based on the value of a different field is not that common I would say, so I don't have a good feeling about it. Do you have examples where this pattern is used?

> I'm not sure disabling the inputs doesn't also create strange situations, like you first select a target, then enable the redirect, then remove the target again. In this situation, from what you describe I would think the redirect option is still selected but now disabled so the user cannot even disable the option anymore.

Yes, the option would remain selected but disabled, which in technical terms means ignored, but sure, the user may not understand it like this. Now, why would the user want to disable the "Create Redirect" option after explicitly removing (clearing) the target? It should be clear (also from the target hint) that without a target there can't be any redirect.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -217,61 +217,50 @@
...
-        #displayAutoRedirectCheckbox({
-          'label': 'core.delete.autoRedirect.label',
-          'hint': 'core.delete.autoRedirect.hint',
-          'hidden': true
-        })
-      </dl>
-    #end
-    #displayPanel('panel-backlinks', 'panel-default', $heading, $body)
     #set($hasInlinks = true)
   #end
```
mflorea, line 229:
```suggestion
  #set ($hasInlinks = $links && $links.size() > 0)
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -217,61 +217,50 @@
...
+  #define($heading)
+    <strong>$services.localization.render('core.delete.backlinks')</strong> ($links.size())
```
mflorea, line 231: Missing escape.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/delete.js
```diff
@@ -19,11 +19,8 @@
...
   $(document).on('keyup change', '#delete #newBacklinkTarget', function() {
-    let updateLinksContainer = $("#delete input[name=updateLinks]").closest('dt');
-    let autoRedirectContainer = $("#delete input[name=autoRedirect]").closest('dt');
-    let elements = updateLinksContainer.add(updateLinksContainer.next('dd'))
-      .add(autoRedirectContainer).add(autoRedirectContainer.next('dd'));
-
-    elements.toggleClass('hidden', $(this).val() === '');
+    let updateLinksContainer = $("#delete input[name=updateLinks]");
+    let autoRedirectContainer = $("#delete input[name=autoRedirect]");
+    updateLinksContainer.add(autoRedirectContainer).prop('disabled', $(this).val() === '');
```
mflorea, line 24:
```suggestion
    $("#delete input[name=updateLinks], #delete input[name=autoRedirect]").prop('disabled', $(this).val() === '');
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/refactoring_macros.vm
```diff
@@ -85,13 +85,14 @@
 #end
 
 #macro (displayLinksCheckbox $options)
-  #set ($hidden = (!$isAdvancedUser && !$isSuperAdmin) || $options.hidden)
+  #set ($hidden = !$isAdvancedUser && !$isSuperAdmin)
```
mflorea, line 88: I'm not sure this change is needed.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/refactoring_macros.vm
```diff
@@ -122,16 +128,16 @@
   ## In the future we'll offer a config option to define the default behavior, see
   ## http://jira.xwiki.org/browse/XWIKI-13384
   #set ($checked = $request.autoRedirect == 'true')
-  #set ($hidden = $options.hidden)
-  <dt #if ($hidden) class="hidden"#end>
+  <dt>
```
mflorea, line 131: I don't think this change is needed.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/refactoring_macros.vm
```diff
@@ -122,16 +128,16 @@
...
-      <input type="checkbox" name="autoRedirect" value="true" #if ($checked)checked="checked"#end />
+      <input type="checkbox" name="autoRedirect" value="true" #if ($checked)checked="checked"#end
+        #if ($options.disabled)disabled#end/>
       $escapetool.xml($services.localization.render($options.label))
     </label>
     ## The value submitted when the checkbox is not checked, used to preserve the form state.
     <input type="hidden" name="autoRedirect" value="false" />
   </dt>
-  <dd #if ($hidden) class="hidden"#end>
+  <dd>
```
mflorea, line 140: Same, I don't think this change is needed.

mflorea (changes requested):

michitux:
> Being unaware that an option exists is not a problem unless you need that option.

Without being aware of the feature, I wouldn't select a target and thus never discover the option and assume it doesn't exist which I think is a problem.

> I agree that it's more natural to first ask "Do you want a redirect?" and then "Where to?", but since the "Where to?" part is shared by multiple options (create redirect and update links), we reversed the questions.

To me there is no good reason to reverse the questions apart from the fact that is technically easier which is usually not a good indicator of good usability. You could also have the two checkboxes first and when you've selected one of them, the target selection is displayed below the two. Anyways, if you don't agree, I'm fine to leave it as is/whatever you propose.

mflorea (approved):

PR #1998, opened 2022-12-06
XWIKI-20452: Silk icons shouldn't be displayed in lightbox

surli:
* Ensure to ignore the images created for silk icons in lightbox


surli:
@oanalavinia could you confirm that it's ok? The doc mentioned the parent element, so I'm not sure I can do that directly on the img element.

oanalavinia:
@surli Yes, it woks directly on the element. I'll update the documentation to be more precise

oanalavinia:
There was already https://jira.xwiki.org/browse/XWIKI-20126 and I had some work on the issue, I just didn't found the time to finish testing / investigating, but it should help https://github.com/oanalavinia/xwiki-platform/commit/17f1443833635a9efcc4d9242fba6f9d254d62c5 since I think some things are missing from this commit.
I'm not sure about where `data-xwiki-lightbox="false"` should be added, but you should put it for the wiki syntax as well. Also, the LightboxIT should be updated since it used at some point a ``  and will fail now (the above commit fixes it)

PR #1999, opened 2022-12-06
XWIKI-20367: Attachment selector doesn't use the sourceDocument reference

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20367

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image-old/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image-old/plugin.js
@@ -190,14 +190,6 @@
               (newResourceReference.type !== 'url' || newResourceReference.reference.indexOf('://') < 0);
             widget.setData('resourceReference', CKEDITOR.tools.extend(newResourceReference, oldResourceReference));
           }
-        },
-        getBase: function () {
-          var currentInstance = CKEDITOR.currentInstance;
-          var base;
-          if (currentInstance) {
-            base = currentInstance.config.sourceDocument.documentReference;
-          }
-          return base;
         }
       });
       CKEDITOR.plugins.xwikiResource.updateResourcePickerOnFileBrowserSelect(dialogDefinition,
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image-old/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image-old/plugin.js
@@ -190,6 +190,14 @@
               (newResourceReference.type !== 'url' || newResourceReference.reference.indexOf('://') < 0);
             widget.setData('resourceReference', CKEDITOR.tools.extend(newResourceReference, oldResourceReference));
           }
+        },
+        getBase: function () {
+          var currentInstance = CKEDITOR.currentInstance;
+          var base;
+          if (currentInstance) {
+            base = currentInstance.config.sourceDocument.documentReference;
+          }
+          return base;
         }
       });
       CKEDITOR.plugins.xwikiResource.updateResourcePickerOnFileBrowserSelect(dialogDefinition,
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/entityResourcePicker.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/entityResourcePicker.js
@@ -181,7 +181,15 @@ define('entityResourcePicker', [
       }
     });
     handler.open = function(resourceReference) {
-      entityTreePickerHandler.open($resource.convertResourceReferenceToEntityReference(resourceReference));
+      // If the editor instance has a current instance defined, use it as the base entity reference to resolve the 
+      // resource reference.  
+      var baseEntityReference;
+      var currentInstance = CKEDITOR.currentInstance;
+      if (currentInstance) {
+        baseEntityReference = currentInstance.config.sourceDocument.documentReference;
+      }
+      entityTreePickerHandler
+        .open($resource.convertResourceReferenceToEntityReference(resourceReference, baseEntityReference));
     };
     return handler;
   };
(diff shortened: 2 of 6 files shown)


mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/resourcePicker.js
```diff
@@ -169,13 +169,30 @@ define('resourcePicker', [
...
+      // document tree open to the source document of the current editor.
+      // In the case where the picker opens on an attachment, but the current reference is not an attachment,
+      // the reference type is forced to attachments by suffixing it with an '@'.
+      // As a consequence, the document tree opens to the 'Attachments' sub-level of the document tree. 
+      var type = resourceTypeButton.val();
+      var documentReference = CKEDITOR.currentInstance.config.sourceDocument.documentReference;
+      var reference = XWiki.Model.serialize(documentReference);
+      if (type === "attach") {
+        reference += "@";
+      }
```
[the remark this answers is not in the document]
manuelleduc: Thanks for the explanations. I've pushed a new version which imo looks better.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/entityResourcePicker.js
```diff
@@ -181,7 +181,15 @@ define('entityResourcePicker', [
       }
     });
     handler.open = function(resourceReference) {
-      entityTreePickerHandler.open($resource.convertResourceReferenceToEntityReference(resourceReference));
+      // If the editor instance has a current instance defined, use it as the base entity reference to resolve the 
+      // resource reference.  
+      var baseEntityReference;
+      var currentInstance = CKEDITOR.currentInstance;
+      if (currentInstance) {
```
mflorea, line 188: This is the part that I don't like. The resource picker should not be aware of any editor. It's simply a widget that allows you to pick a resource from the wiki. It's not much different, conceptually, than a date picker. It just happens that the CKEditor has the need to select a resource and it uses the resource picker for this. But the Wiki editor could use it too, or some other module. I know it is packaged with CKEditor, but this was needed when the code was in contrib because we were supporting older versions of XWiki so we couldn't have this resource picker in XWiki. Now we can move it in a separate WebJar module if we want.

I would prefer instead to:
* pass the current document as a parameter to the resource picker here https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.10.1/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/plugin.js#L121 , besides the ``resourceTypes`` parameter. I think "current" or "base" makes more sense for the resource picker than "source" (``sourceDocument`` makes sense for the editor)
* store the resource picker parameters (options) as (jQuery) data on the resource picker element, here https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.10.1/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/resourcePicker.js#L47-L50
* take the parameters / options from the resource picker element and pass the "current" / "base" document reference when opening the dedicated pickers here https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.10.1/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/resourcePicker.js#L184 , besides the ``resourceReference``
* use the "current" / "base" document reference in the dedicated pickers here https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.10.1/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/entityResourcePicker.js#L216 , with a fallback to ``XWiki.currentDocument``
* find all occurrences of ``XWiki.currentDocument`` in dedicated resource pickers and replace with the usage of the passed "current" / "base" document reference.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/entityResourcePicker.js
manuelleduc: I've pushed a new version that, I hope, follows your recommendations.
The resources are now selected, and resolve, relatively to the current document of the editor if it exists.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image-old/plugin.js
```diff
@@ -190,6 +190,14 @@
...
             widget.setData('resourceReference', CKEDITOR.tools.extend(newResourceReference, oldResourceReference));
           }
+        },
+        getBase: function () {
+          var currentInstance = CKEDITOR.currentInstance;
+          var base;
+          if (currentInstance) {
+            base = currentInstance.config.sourceDocument.documentReference;
+          }
+          return base;
```
mflorea, line 200: Why not put this directly in ``xwiki-resource`` plugin? I see you duplicate it also in the link plugin, but I'm not sure why.

mflorea (commented):
Looks good, except for the duplicate ``getBase`` implementation that could be put directly in ``xwiki-resource`` plugin, I think.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image-old/plugin.js
manuelleduc: I'm not sure to understand, I thought you said in previous comments that dependencies on CKEditor should be avoided in `xwiki-resource`. Unless you meant that `xwiki-resource` contains  some generic part (that could be extracted in a separate module), and the "glue" part (i.e., `xwiki-resource/plugins.js`). In that case I agree it would be better to move `getBase` here and remove the duplication.

mflorea: > Unless you meant that xwiki-resource contains some generic part (that could be extracted in a separate module), and the "glue" part (i.e., xwiki-resource/plugins.js).

Yes, ``xwiki-resource/plugins.js`` is of course specific to CKEditor, as it defines a CKEditor plugin, so it has to use the CKEditor API. The rest, ``resource*.js`` and ``entityResource*.js``, are generic, defining the "resource picker", that can be used outside of CKEditor. Sorry for the confusion.

manuelleduc: No worry, now that you say it it's kind of obvious :)

mflorea (approved):

PR #2003, opened 2022-12-08
[Misc] Retrieve all from master in 13.10.x

surli:


vmassol:
Small detail: I think this is about translations so maybe update the commit message to: `[Misc] Merge all missing translations from master in 13.10.` ?

So we need to verify that no translation has been reused in master, right?

Or even translation changed to use some new terminology not used on the 13.10.x branch?

surli:
> Small detail: I think this is about translations so maybe update the commit message to: [Misc] Merge all missing translations from master in 13.10. ?

Yes in any case this PR needs editions, changing to draft.

> So we need to verify that no translation has been reused in master, right?

Not sure to understand here. We mostly need to verify that the translation files matches modules that were existing in 13.10.x here. For example the PR currently contains some files related to attachment-validation which is new stuff from 14.x. 

> Or even translation changed to use some new terminology not used on the 13.10.x branch?

Ideally yes, but I don't know how you do it unless you speak many languages :)

vmassol:
> Not sure to understand here. 

Imagine some dev failed to deprecate a translation when adding a new parameter to an existing translation (our best practice). This won't be a problem we notice on master but it would on the 13.10.x branch.

> For example the PR currently contains some files related to attachment-validation which is new stuff from 14.x.

Yes, but these won't cause problems since they're not used.

surli:
> Imagine some dev failed to deprecate a translation when adding a new parameter to an existing translation (our best practice). This won't be a problem we notice on master but it would on the 13.10.x branch.

Well that's nothing new so not sure why you're mentioning it there. Also for me this should be checked / fixed in the commits where the translation parameter has been added to not accept such change, even for master.

> Yes, but these won't cause problems since they're not used.

Strictly speaking no I don't think they would, now it's really not nice to add a hierarchy of folders for modules that don't exist on branches, and possibly error prone (e.g. seeing a folder for xwiki-platform-attachment-validation in stable-13.10.x might give false idea that this module existed in that version)

vmassol:
Another thing to check: the move of deprecated keys that are not deprecated for 13.10.x.

For example, at https://github.com/xwiki/xwiki-platform/pull/2003/files#diff-aa8bc91e65ef54c73cc7f14df2c87641a89f375abfd6b951587f1cc5629e9db3L990-L994

surli:
> Another thing to check: the move of deprecated keys that are not deprecated for 13.10.x.

Agree on the long term view but note that right now this has strictly no impact at all to move them. So it's probably easier to get the translations moved than to commit a change here: we  won't have to handle conflicts for future changes in those translations.

vmassol:
> Well that's nothing new so not sure why you're mentioning it there. 

Because of `This won't be a problem we notice on master but it would on the 13.10.x branch.`.

However, I've checked this PR and I didn't see any such occurence, so good.

vmassol:
> Agree on the long term view but note that right now this has strictly no impact at all to move them. So it's probably easier to get the translations moved than to commit a change here: we won't have to handle conflicts for future changes in those translations.

I don't think we should leave this in. It's bad and nothing is short lived. Any release and the associated sources will live forever (or very very long) and it's not clean to leave wrong things voluntarily in the sources. It's worse than adding new keys that are not used for me.

surli:
> it's not clean to leave wrong things voluntarily in the sources

I agree, but here the examples you gave me are not wrong: those keys are deprecated "until 14.9" or "until 14.6" etc, as it's properly explained in the source. So there's nothing wrong here. 

> It's worse than adding new keys that are not used for me.

I've definitely a different view on this.

vmassol:
> I agree, but here the examples you gave me are not wrong: those keys are deprecated "until 14.9" or "until 14.6" etc, as it's properly explained in the source. So there's nothing wrong here.

Well, any key in that section is supposed to be deprecated.

````
## Deprecated
## Note: each element should be removed when the last branch using it is no longer supported

## Used to indicate where deprecated keys start
#@deprecatedstart
...
```

BTW "until 14.9" si confusing to me. I read it as "these keys are deprecated until 14.9" which obviously is the opposite. Using "since 14.9" would make more sense. If we want to use "until" then we need to add more context, such as: "Used until 14.9". But "since" is better IMO since it matches the keyword we use in java code.

PR #2016, opened 2022-12-15
XWIKI-20479 Planuml fails to generate diagram if the page hierarchy i…

xumix:
…s deep or name too long

### DIFF
--- a/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-temporary/src/main/java/org/xwiki/resource/temporary/internal/DefaultTemporaryResourceStore.java
+++ b/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-temporary/src/main/java/org/xwiki/resource/temporary/internal/DefaultTemporaryResourceStore.java
@@ -31,6 +31,7 @@ import java.util.List;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.xwiki.component.annotation.Component;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-temporary/src/main/java/org/xwiki/resource/temporary/internal/DefaultTemporaryResourceStore.java
```diff
@@ -75,31 +76,35 @@ public File createTemporaryFile(TemporaryResourceReference reference, InputStrea
...
         }
         if (!reference.getParameters().isEmpty()) {
             segments.add(String.valueOf(reference.getParameters().hashCode()));
-            safePathLength++;
         }
-        segments.addAll(reference.getResourcePath());
+
         String path = StringUtils.join(encode(segments), '/');
-        String safePath = StringUtils.join(encode(segments.subList(0, safePathLength)), '/');
+        String md5 = DigestUtils.sha1Hex(path);
```
tmortagne, line 90: I don't mind the use of sha1, but then the variable name is not very accurate :)

xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-temporary/src/main/java/org/xwiki/resource/temporary/internal/DefaultTemporaryResourceStore.java
```diff
@@ -75,31 +76,35 @@ public File createTemporaryFile(TemporaryResourceReference reference, InputStrea
...
+        finalPathSegments.add(String.valueOf(md5.substring(2)));
+
+        finalPathSegments.addAll(reference.getResourcePath());
+
         File rootFolder = this.environment.getTemporaryDirectory();
-        File safeFolder = new File(rootFolder, safePath);
-        File temporaryFile = new File(rootFolder, path);
 
-        // Make sure the resource path is not relative (e.g. "../../../") and tries to get outside of the safe folder.
-        if (!temporaryFile.getAbsolutePath().startsWith(safeFolder.getAbsolutePath())) {
```
tmortagne, line 99: Why have you removed this protection ? It's still possible to have this kind of things in the resource path, so there is no reason to remove that.

xumix: Added safe check

PR #2021, opened 2022-12-20
XWIKI-20475: "Attach" and "Cancel" buttons  are displayed in Attachment Tab when in-place edit mode is cancelled

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20475

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
@@ -53,13 +53,10 @@ viewers.Attachments = Class.create({
   counter : 1,
   /** Constructor. Adds all the JS improvements of the Attachment area. */
   initialize : function() {
-    if ($("attachform")) {
-      // If the upload form is already visible, enhance it.
-      this.prepareForm();
-    } else {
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.js
```diff
@@ -53,13 +53,10 @@ viewers.Attachments = Class.create({
...
   /** Constructor. Adds all the JS improvements of the Attachment area. */
   initialize : function() {
-    if ($("attachform")) {
-      // If the upload form is already visible, enhance it.
-      this.prepareForm();
-    } else {
-      // Otherwise, we wait for a notification for the AJAX loading of the Attachments metadata tab.
-      this.addTabLoadListener();
-    }
+    // Initialize the event listener to prepare the form when it's loaded, or prepare the form straight away.
```
mflorea, line 56: Isn't the issue the fact the in-place editor **re**loads the document extra tabs, [here](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml#L1155-L1164), and the attachments tab load listener is not registered? So I'd make the comment more explicit about the fact that the listener is needed also when the tab is reloaded.

```suggestion
    // Initialize the event listener to prepare the form when the Attachments tab is loaded or reloaded, or prepare the form straight away.
```

PR #2041, opened 2023-01-04
XWIKI-20499: Uniformize the image dimensions on insert/update

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20499

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/plugin.js
@@ -266,23 +266,8 @@ require(['jquery', 'resource', 'resourcePicker'], function ($, $resource) {
         if (onlyTheReferencePart  === true) {
           this.setValue(resourceReference.reference);
         } else {
-          var resourceURL = CKEDITOR.plugins.xwikiResource.getResourceURL(resourceReference,
-            this.getDialog().getParentEditor());
-          var width = this.getDialog().getValueOf(resourcePickerId[0], "width");
-          var height = this.getDialog().getValueOf(resourcePickerId[0], "height");
-
-          if (width || height) {
-            var params = [];
-            if (width) {
-              params.push("width=" + width);
-            }
-            if (height) {
-              params.push("height=" + height);
-            }
-            resourceURL += '&parameters[queryString]=' + encodeURI(params.join(','));
-          }
-          
-          this.setValue(resourceURL);
+          this.setValue(CKEDITOR.plugins.xwikiResource.getResourceURL(resourceReference,
+            this.getDialog().getParentEditor()));
         }
         if (typeof oldCommit === 'function') {
           oldCommit.apply(this, arguments);
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/plugin.js
@@ -268,13 +268,19 @@ require(['jquery', 'resource', 'resourcePicker'], function ($, $resource) {
           var resourceURL = CKEDITOR.plugins.xwikiResource.getResourceURL(resourceReference,
             this.getDialog().getParentEditor());
           var width = this.getDialog().getValueOf(resourcePickerId[0], "width");
-          if (width) {
-            resourceURL = resourceURL + '&width=' + width;
-          }
           var height = this.getDialog().getValueOf(resourcePickerId[0], "height");
-          if (height) {
-            resourceURL = resourceURL + '&height=' + height;
+
+          if (width || height) {
+            var params = [];
+            if (width) {
+              params.push("width=" + width);
+            }
+            if (height) {
+              params.push("height=" + height);
+            }
+            resourceURL += '&parameters[queryString]=' + encodeURI(params.join(','))
           }
+          
           this.setValue(resourceURL);
         }
         if (typeof oldCommit === 'function') {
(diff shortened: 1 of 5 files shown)


michitux:
Have you tested this? I've just performed some small tests and to me this re-creates the original problem of images not preserving their aspect ratio whenever a height is specified. We should not modify the width of an image without also setting its height, this is just a bad idea, so it's either both or none, not just one of them. I recommend reading [this article](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/) on why we should set the `height` and `width` attributes and then still override them using CSS as it is done right now. I see two solutions:

* Explicitly generate a `style` attribute with `aspect-ratio` set to the desired aspect ratio either in CKEditor or during rendering (but then we would need to remove it again during parsing which might be non-trivial). This could also be done just when the aspect ratio is not the native aspect ratio of the image.
* Add a CSS class or a data attribute to images that don't have the native aspect ratio and make them non-responsive based on this attribute.

michitux:
Ideally, we would be setting `aspect-ratio: attr(width) / attr(height)` but unfortunately that's currently not supported by browsers. What browsers currently do from my understanding is that they use this aspect ratio until the image is loaded at which point they will replace that by the actual aspect ratio of the loaded image. We could consider using JavaScript to emulate `aspect-ratio: attr(width) / attr(height)` but I fear this will be difficult with WYSIWYG editing and would create additional attributes in the source.

manuelleduc:
Related PR: https://github.com/xwiki/xwiki-platform/pull/2045

manuelleduc:
Now that #2045 is merged:
- send the width and height to `CKEditor.ResourceDispatcher`
- forward the width and height to the download url on `CKEditor.ResourceDispatcher` when `width` and `height` exists.

michitux:
Note that this won't help much when resizing to sizes larger than the original size. Also, have you tried if this helps with resizing in the editor?

manuelleduc:
> Note that this won't help much when resizing to sizes larger than the original size. Also, have you tried if this helps with resizing in the editor?

- #2045 is enough outside of the editor
- the proposed change is to see the change reflected on the editor
- I haven't tested on resize larger than the original size (trying)

manuelleduc:
> I haven't tested on resize larger than the original size (trying)

I confirm my changes have no effect when the image dimensions are larger than the original ones.

michitux:
With sizes larger than the original, #2045 is not enough even outside the editor as server-side resizing is never upscaling images. What I was wondering about is if it would work when you resize with the mouse by dragging the bottom-right corner of the image (for dimensions smaller than the original).

manuelleduc:
> What I was wondering about is if it would work when you resize with the mouse by dragging the bottom-right corner of the image (for dimensions smaller than the original).

- The server is never queried, and the aspect ratio cannot be changed. The image is just resized client side.
- If after resize, the dimensions are larger than the original, this is not taken into account in the editor, but has an effect on view mode (where the not-scaled image is displayed).

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-resource/plugin.js
```diff
@@ -265,8 +265,17 @@ require(['jquery', 'resource', 'resourcePicker'], function ($, $resource) {
...
+          var resourceURL = CKEDITOR.plugins.xwikiResource.getResourceURL(resourceReference,
+            this.getDialog().getParentEditor());
+          var width = this.getDialog().getValueOf(resourcePickerId[0], "width");
+          if (width) {
+            resourceURL = resourceURL + '&width=' + width;
+          }
+          var height = this.getDialog().getValueOf(resourcePickerId[0], "height");
+          if (height) {
+            resourceURL = resourceURL + '&height=' + height;
+          }
```
mflorea, line 277: This is very hackish. width/height don't make sense for the generic resource plugin. Not all resources have an width/height.

mflorea (changes requested):
The server-side image resize is an optimization (to reduce the downloaded content) that the client side should not be aware of or rely on. So I don't think this is a proper fix. Moreover, as @michitux noted, it doesn't work when the width/height are larger than the original image width/height.

mflorea:
> Ideally, we would be setting aspect-ratio: attr(width) / attr(height) but unfortunately that's currently not supported by browsers.

@michitux AFAIK the issue is not browser support but the fact that ``aspect-ratio`` is used to specify a **preferred** aspect ratio, and on https://w3c.github.io/csswg-drafts/css-sizing-4/#aspect-ratio they say:

> Note: A [preferred aspect ratio](https://w3c.github.io/csswg-drafts/css-sizing-4/#preferred-aspect-ratio) only ever has an effect if at least one of the box’s sizes is [automatic](https://w3c.github.io/csswg-drafts/css-sizing-3/#automatic-size). If neither [width](https://w3c.github.io/csswg-drafts/css-sizing-3/#propdef-width) nor [height](https://w3c.github.io/csswg-drafts/css-sizing-3/#propdef-height) is an automatic size, it can have no effect on its [preferred sizes](https://w3c.github.io/csswg-drafts/css-sizing-3/#preferred-size).

So ``aspect-ratio`` won't help us to preserver the aspect ratio when the image is rendered on small screens.

@manuelleduc I tried a couple of things locally but unfortunately I don't have a solution that:
* respects the specified image width/height on large screens
* keep the aspect ratio derived from the width/height on small screens

The first item can be achieved with what you propose on https://jira.xwiki.org/browse/XWIKI-20499?focusedId=113724&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-113724

```
img:not([height]) {
  height: auto;
} 
```

The issue is that you loose the aspect ratio on small screens. O partial fix is to use on the image:

    object-fit: contain;

but it's not nice.

michitux:
> @michitux AFAIK the issue is not browser support but the fact that `aspect-ratio` is used to specify a **preferred** aspect ratio, and on https://w3c.github.io/csswg-drafts/css-sizing-4/#aspect-ratio they say:
> 
> > Note: A [preferred aspect ratio](https://w3c.github.io/csswg-drafts/css-sizing-4/#preferred-aspect-ratio) only ever has an effect if at least one of the box’s sizes is [automatic](https://w3c.github.io/csswg-drafts/css-sizing-3/#automatic-size). If neither [width](https://w3c.github.io/csswg-drafts/css-sizing-3/#propdef-width) nor [height](https://w3c.github.io/csswg-drafts/css-sizing-3/#propdef-height) is an automatic size, it can have no effect on its [preferred sizes](https://w3c.github.io/csswg-drafts/css-sizing-3/#preferred-size).

But when we set `height: auto`, the height is an automatic size and the aspect ratio should thus be used? So if we computed the correct aspect ratio on the server/in CKEditor (could also be during mouse movement for resizing), shouldn't this work?

PR #2044, opened 2023-01-04
XWIKI-20530 Attachment Selector macro does resolve attachment from different target document

ClemDee:
Jira: https://jira.xwiki.org/browse/XWIKI-20530

The attachment selector was still trying to resolve the attachment from the current document even if the `targetdocname` property was given.


manuelleduc:
The PR looks good. I'd suggest to add a docker test validating the behavior, to validate the fix and prevent future regressions.

mflorea (approved):

PR #2049, opened 2023-01-17
XWIKI-19140: Fix order location actions and breadcrumbs while creating new page

Sereza7:
Jira: https://jira.xwiki.org/browse/XWIKI-19140

- Changes to the locationpicker template: moving the block for the location actions after the breadcrumbs (same as visual) , which allows for the focus to go in the expected order when tabbing.
- The style changes in the .css are there to put the objects back at their original position. 

Screenshot from 2023-01-17 15-38-47

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
@@ -53,7 +53,9 @@
   }
 }
 
-dd .location-actions {
+/* TODO: Update the JS code to use a more specific CSS selector when updating the content of the actual breadcrumb,
+  in order to eventually remove this rule and add the 'breadcrumb' class to the 'location-actions' node. */
+dd > .location-actions {
   /* Retrieve style from the breadcrumb block. Needed for keyboard accessibility reorganization of the
   location-picker component. */
   .breadcrumb;
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
@@ -53,8 +53,8 @@
   }
 }
 
-dd.location-picker-preview{
-  /*Make the location picker display the breadcrumb "block".
-  Useful to separate behaviour from presentation for the breadcrumb class.*/
+dd .location-actions {
+  /* Retrieve style from the breadcrumb block. Needed for keyboard accessibility reorganization of the
+  location-picker component. */
   .breadcrumb;
 }
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
@@ -52,3 +52,9 @@
     }
   }
 }
+
+dd.location-picker-preview{
+  /*Make the location picker display the breadcrumb bg color.
+  Useful to separate behaviour from presentation of the breadcrumb class.*/
+  background-color: @breadcrumb-bg;
+}
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
@@ -54,7 +54,7 @@
 }
 
 dd.location-picker-preview{
-  /*Make the location picker display the breadcrumb bg color.
-  Useful to separate behaviour from presentation of the breadcrumb class.*/
-  background-color: @breadcrumb-bg;
+  /*Make the location picker display the breadcrumb "block".
+  Useful to separate behaviour from presentation for the breadcrumb class.*/
+  .breadcrumb;
 }
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/locationPicker_macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/locationPicker_macros.vm
@@ -65,20 +65,18 @@
       <label>$escapetool.xml($services.localization.render($options.preview.label))</label>
       <span class="xHint">$escapetool.xml($services.localization.render($options.preview.hint))</span>
     </dt>
-    <dd class="location-picker-preview">
+    <dd>
       ## The breadcrumb is updated whenever the user changes the parent page. This causes the breadcrumb element to be
       ## constantly replaced, preventing us from displaying the live validation message after it. In order to overcome
       ## this, we wrap the breadcrumb element in a DIV that remains the same.
       <div class="breadcrumb-container">
         ## Note: We display only the parent reference here. The new document part will be added from JavaScript.
         #hierarchy($options.parent.reference $options)
+        #if ($isDocumentTreeAvailable)
+          #documentPickerModal($options)
+          #locationPickerActions()
+        #end
       </div>
-      #if ($isDocumentTreeAvailable)
-        #documentPickerModal($options)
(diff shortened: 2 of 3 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/locationPicker.css
```diff
@@ -9,10 +9,18 @@
...
+  float: left;
+  width: 100%;
+}
+
 .location-actions {
   float: right;
   /* Use the same padding as the breadcrumb. */
   padding: 8px 15px 0 15px;
+  /*Allows overlapping with the breadcrumb-container.*/
+  margin-left:-70px;
```
michitux, line 23: I think this negative margin is a bad idea for two reasons:

1. The width of the actions is different for advanced vs. non-advanced users (one vs. two icons)
2. The negative margin means that the text of a very long space will be hidden behind the icons instead of continuing in the next line.

I would suggest using a Flexbox layout which should allow creating a layout where the first item fills the available space and the second item has just its natural width easily and without such hacks.

michitux (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/locationPicker.css
Sereza7: I removed the negative margin.
Since there's overlapping of the two children components, I used some absolute positioning instead.

It doesn't seem to me that a Flexbox is useful in this case, is the patch okay without it?

michitux:
Also, a more generic comment, please check [our best practices regarding commit messages](https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HRule:AlwaysputaJIRAissuereferenceincommitmessages). You can leave the PR as-is, we can squash the commits and thereby change the commit message.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/locationPicker.css
michitux: Just to illustrate what I mean, here is a screenshot after manually applying the absolute positioning:

long and wrapping document reference hidden behind the buttons

Before any changes:

long and wrapping document reference
 
I don't insist on using Flexbox, it's just that Flexbox makes exactly this kind of layout very easy, with "this kind of layout" I mean two elements that share a line such that one of them has a fixed width while the other just takes the rest.

Sereza7:
Thanks for the help!
The solution I pushed in the last commit should solve those overlap issues.
This issue was with the breadcrumb-container and the location-actions objects.
The background square (light grey color) was on breadcrumb (child of breadcrumb-container) so I needed to superpose both of the objects. I couldn't get superposition and flexbox together without using a backhanded trick.
So I moved the background to their parent (location-picker-preview) so I could just use the flexbox inside without any issue.


For removing the background in the child, I overwrote the properties defined in style.css instead of editing this file.
I used a hard coded color for the new background, I suppose it's okay since it was the case already on the child object.

Sereza7:
Nevermind, this still has a lot of issues when opening up the edit buttons. More changes needed.

Sereza7:
**PR Changes:**
Fixed the keyboard navigation issue by reorganizing elements in the location picker preview.
This template change had to go with style changes to make sure the new parent class holds all the properties it needs.


The tabulation order (displayed with the Firefox inspector accessibility tool):
Before
TabOrder19140Before
After
tabOrder19140

The change of template hierarchy also created a change in UI: when clicking on the location-action-edit, the un-hidden section is now inside of the light grey box.
Before
BeforePR19140
After
afterPR

It seems to me like this change in UI does not degrade the experience, but I'm looking forward to your feedback about it.

____

Squash commits if merge

michitux:
I'm not really happy with the UI change because this decreases the contrast between the hint and the background. It wasn't an option to introduce a new container for the background that just contains the two elements (path + buttons)?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/locationPicker.css
```diff
@@ -7,23 +7,47 @@
...
+  background-color: transparent;
+  border-radius: 0px;
+  padding: 0;
+  /*Remove the ol left padding*/
+  padding-left: 0;
+}
+
+dd.location-picker-preview{
+  /*Make the location picker display the bg color*/
+  background-color: #f8f8f8;
```
michitux, line 20: The color shouldn't be hardcoded, the background color is defined in the less variable `@breadcrumb-bg` (should be available in `$theme.breadcrumbBg` via Velocity). There are color themes that are adding a border to the breadcrumbs:

https://github.com/xwiki/xwiki-platform/blob/226762d964f8a4ba298d84a3bacad3eb42efed03/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-theme/xwiki-platform-flamingo-theme-bootswatch/src/main/resources/FlamingoThemes/Lumen.xml#L2129

Sereza7:
**PR Changes:**

- Switched the 'location-edit' element to a new block (see comments)
- Added a LESS rule to shift all of the 'breadcrumb' class rules to a new class 'location-picker-preview'.
- Overwrite to 'none' more fields on the now repetitive 'breadcrumb' element

<details>
  <summary> 

**Overview using the Iceberg color theme:**</summary>

19140iceberg100
</details>

<details>
  <summary>

**Overview using the Lumen color theme:**</summary>

19140Lumen100
</details>

____
**Comments:**
Currently, the 'location-edit' is in a new 'dd' element. I'm not sure of what is the best practice in this case:
- Keep two dds one after another in the dl.
- Put the 'location-edit' at the end of the dl without anything more.
- Create a hidden dt to describe the last dd.
The idea of moving 'location-edit' out of the dl currently returns an error when validating the form.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/locationPicker_macros.vm
```diff
@@ -65,18 +65,20 @@
       <label>$escapetool.xml($services.localization.render($options.preview.label))</label>
       <span class="xHint">$escapetool.xml($services.localization.render($options.preview.hint))</span>
     </dt>
-    <dd>
-      #if ($isDocumentTreeAvailable)
-        #documentPickerModal($options)
-        #locationPickerActions()
-      #end
+    <dd class="location-picker-preview">
```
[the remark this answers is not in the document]
Sereza7: Hello! 
I tested out your solution and it's working properly on the default color theme (Iceberg). However, as you can see under there, it doesn't work with the Lumen color theme, that defines custom border properties to the `breadcrumb` class. 

Moreover, as for implementation, I can't just use the `breadcrumb` class on the `location-actions` because this class is used by a script to dynamically change the content of the breadcrumb depending on the form. That's why I added all properties of breadcrumb to a new css class using less (so the script wouldn't target the new block).

```
dd .location-actions {
  /* Retrieve style from the breadcrumb block. Needed for keyboard accessibility reorganization of the 
  location-picker component.*/
  .breadcrumb;
}
```


_Using Iceberg color theme_
19140IcebergSplitBreadcrumbs


_Using Lumen color theme_
19140LumenTestbreadcrumbssplit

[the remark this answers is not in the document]
Sereza7: Sorry, my previous comment is not relevant.
I added a couple rules for both of the children and this should be enough to correct the border issues (except if the border-radius gets changed itself in a color theme).
The commit to come solves those border problems, however there's still an addition to the `breadcrumb.less` file.

19140LumenCorrected

Sereza7:
**PR Changes:**
Added changes proposed here: https://github.com/xwiki/xwiki-platform/pull/2049#discussion_r1091994209

**View**
_With Iceberg color theme_
19140IcebergFinal


_With Lumen color theme_
19140LumenFinal

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/breadcrumbs.less
```diff
@@ -52,3 +52,9 @@
     }
   }
 }
+
+dd .location-actions {
+  /* Retrieve style from the breadcrumb block. Needed for keyboard accessibility reorganization of the
+  location-picker component. */
+  .breadcrumb;
```
mflorea, line 59: The best would be to have this inside ``locationPicker.css`` but we don't support LESS there ATM so we're left with two options:

* keep this (temporarily) inside ``breadcrumb.less``
* update the JavaScript code to use a more specific CSS selector when updating the actual breadcrumb.

I'm not sure what's best. I guess the first option is acceptable for now.

mflorea (changes requested):

Sereza7:
**PR Changes:**

- Updated to take into account Marius' comments.
- Added a TODO on the .less for https://github.com/xwiki/xwiki-platform/pull/2049#discussion_r1094224712
- Improved specificity of selectors for style.

PR #2051, opened 2023-01-18
XWIKI-13786-2: Accessibility problem with rights management's checkboxes states

Sereza7:
Jira: https://jira.xwiki.org/browse/XWIKI-13786

Continuation of: https://github.com/xwiki/xwiki-platform/pull/542
This old PR was small and the sole file it concerned moved since then, creating a merge issue.

I reused the lines from the above PR to solve the initial issue and added more accessibility support (mentioned by the issue reporter on jira) for the Rights table: keyboard navigation and interactivity.


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -55,6 +55,13 @@ window.MSCheckbox = Class.create({
     ];
     this.labels = ['','',''];
 
+    // Alts are the messages describing the checkbox images
+    this.alts = [
+      '$escapetool.javascript($services.localization.render("none"))',
+      '$escapetool.javascript($services.localization.render("allow_1"))',
+      '$escapetool.javascript($services.localization.render("allow_0"))'
```
vmassol, line 62: the keys are missing some namespacing. See https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HTranslationBestPractices

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -89,6 +99,8 @@ window.MSCheckbox = Class.create({
       delete this.table.fetchedRows[this.idx];
     }
     this.draw(this.state);
+    /*After changing the image through an interaction, put focus back on it.*/
```
vmassol, line 102: This is explaining what it does but not why. In comments, it's always better to explain why we do something (more useful when another coder reads it or when you read it back a few years after :)).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -205,10 +217,21 @@ window.MSCheckbox = Class.create({
...
+      if(event.key == "Enter"){
+        clickHandler()
+      }
+    }
+  },
 
   attachEvents: function()
   {
     Event.observe(this.domNode, 'click', this.createClickHandler(this));
+    Event.observe(this.domNode, 'keydown', this.createKeyboardHandler(this));   /*Added for accessibility through keyboard*/
```
vmassol, line 234: This is a good comment (it explains why). I'd place it like this:

```
// Added for accessibility through keyboard
Event.observe(this.domNode, 'keydown', this.createKeyboardHandler(this)); 
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -609,4 +632,4 @@ function setGuestExtendedRights(self)
       }
     }
   }
-}
+}
```
vmassol, line 635: the last new line was removed. Was this intentional?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -71,6 +78,9 @@ window.MSCheckbox = Class.create({
     //add new image
     var img = document.createElement('img');
     img.src = this.images[state];
+    img.alt = this.domNode.getAttribute('data-title') + " " + this.alts[state];
```
michitux, line 81: Combining several translation messages in a fixed order is a bad practice and incompatible with some languages. If you need to combine messages, one of them should have a placeholder for the other such that the translator can determine its position. See also https://dev.xwiki.org/xwiki/bin/view/Community/L10N/Conventions/#HRulesfortranslatingcompositeandlongmessages

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -71,6 +78,9 @@ window.MSCheckbox = Class.create({
...
+    img.tabIndex = 0;
```
michitux, line 82: Isn't it a bad accessibility practice to add focus to elements that don't support it natively? Wouldn't it be better to wrap the image in a button?

michitux (commented):
This whole class that tries to add button-like behavior to an image looks like a huge hack imho. At least, I think an element that supports proper keyboard and mouse interaction natively should be added as a wrapper instead of adding more hacks to it to make it somehow accessible. Possibly, it might also make sense to just deprecate this whole class and replace it by something more modern.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
Sereza7: My bad, I had trouble figuring out what this was about.
I understood why it happened (text mode edits in Linux) and fixed it.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
Sereza7: I decided to use the 'platform.core.rightsManagement.' prefix/namespace. 
I'm not sure I understood how I'm supposed to extend the translations.
Of what I understood, since these string are used in the core project, all I need to do is add them directly in the ApplicationResources.properties file. There is nothing to do with L10N.

`platform.core.rightsManagement.allowed=Allowed
platform.core.rightsManagement.denied=Denied
platform.core.rightsManagement.inherit=Inherit`

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
Sereza7: I wrapped the image in a button and moved most of the behavior to this button. :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -55,28 +64,51 @@ window.MSCheckbox = Class.create({
...
+    
+    //Update the description of the button for accessibility.
+    require(['xwiki-l10n!my-translation-keys'], function(l10n) {
+      var alts = [
+        l10n['inherit'],
+        l10n['allowed'],
+        l10n['denied']
+      ];
+      img.alt = alts[this.state];
+      this.button.title = alts[state];
```
michitux, line 103: I wonder if it is enough to indicate the state. On the one hand, it could be annoying for screen reader users to have more text, on the other hand, how do they know what clicking on the button does? Do you know about best practices for this situation?

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2109,6 +2109,9 @@ platform.core.rightsManagement.editRightsForSpace=Editing access rights for spac
 platform.core.rightsManagement.ajaxFailure=An error occurred while communicating with the server. Please check that the server is accessible, and you have the proper rights to perform the requested action.
 platform.core.rightsManagement.saveFailure=An exception occurred while trying to save the current modifications. Please check if you have the proper rights to perform these modifications.
 platform.core.rightsManagement.saveComment={0} {1} right for {2}
+platform.core.rightsManagement.allowed=Allowed
+platform.core.rightsManagement.denied=Denied
+platform.core.rightsManagement.inherit=Inherit
```
michitux, line 2114: This is more complex than inherit. For example, when you give one group view right on a document, the view right is denied for all other groups, the value of the parent has no influence here. It is just not set I would say.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
michitux: You don't need to do anything for Weblate, that's automatic.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
manuelleduc: wdyt of `undefined` for the "not set values"?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.css
```diff
@@ -50,6 +50,10 @@
 #usersandgroupstable.rights .rights {
   width: 15% !important;
 }
+#usersandgroupstable button.rights-edit {
```
manuelleduc, line 53: If this is done for accessibility reason. Can you add a comment explaining what's the impact of this block on accessibility?

manuelleduc (commented):
General remark: it is often useful for reviewers to have a comment explaining what changes with the PR (and if the change is visual, having before/after screenshot helps a lot - or video recordings)

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -55,28 +64,51 @@ window.MSCheckbox = Class.create({
...
+    this.button.className = "rights-edit";
+    this.button.onclick = this.createClickHandler(this);
+    
+    var img = document.createElement("img");
+    
+    this.button.appendChild(img);
+    
+    this.label = document.createElement('span');
+    this.label.hidden = true;
+    this.label.appendChild(document.createTextNode(this.labels[this.state]));
```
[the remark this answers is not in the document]
Sereza7: The label element was present in the structure, however it was never used.
I changed the way the table was generated, so in the case one wanted to use it someday, I had to also change the way the label element was instantiated (and updated).

We should be able to remove it without any changes, it remains hidden.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
michitux: I think the cleanest thing would be to make the items of the table select inputs as this really models what this is about. The problem is that styling of select elements is extremely limited in HTML and we would be forced to put plain text values in there which would be a significant change and not so nice for the design I fear.

What you suggest sounds good, I think putting the current state first should allow screen reader users to quickly switch to the next item in the table after they heard the first word, but maybe it could be good to confirm this with an actual screen reader user (if this is what they would prefer or if there should, e.g., be some instructions that describe the buttons generically at the start of the table and just the current state on the actual buttons).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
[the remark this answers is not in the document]
michitux: I would actually be inclined to just remove these label elements as with the alt text and title they are quite redundant though it could be considered that they are part of the API of XWiki which shouldn't be changed. Any other opinions on this?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
michitux: Btw., I think the respective reference documentation for accessibility of this kind of button is https://www.w3.org/WAI/tutorials/images/functional/#example-5-image-used-in-a-button which says that the `alt` attribute should be used to "convey the purpose of the button".

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -93,13 +94,13 @@ window.MSCheckbox = Class.create({
...
-    require(['xwiki-l10n!my-translation-keys'], function(l10n) {
+    require(['xwiki-l10n!users-and-group-translation-keys'], function(l10n) {
       var alts = [
-        l10n['inherit'],
+        l10n['undefined'],
         l10n['allowed'],
         l10n['denied']
       ];
-      img.alt = alts[this.state];
+      img.alt = alts[this.state] + " . " + l10n.get('changeStateTo', alts[this.nextState()]);
```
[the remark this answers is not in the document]
Sereza7: I meant for the dot to just be a separator between two different semantic units juxtaposed (current state & purpose of the button).
Okay, got it. I changed the translation in the three states to just contain one string that also defines the next state (since this should not change, I guess it can be hard written in the translations. **Should I add a commentary for these entries in the application resource or is that good as is?**
There's a couple needs/preferences for the translation that might get lost otherwise, but I navigated through the resource bundle and could hardly see any comment except those for delimiting sections.
Something like:
### Needs to define current state and then purpose of the button. Should start with the current state name for easier navigation with screen readers.

Sereza7:
PR Changes:
Fixed alt text on image and title of button. The new alt text is not composed. Most of the complexity is shifted to the L10N entry itself. The new alt text is composed of the button status followed by its purpose (switching status).
The third status has been changed to 'undefined' instead of 'inherit'. 
Fixed a lot of issues with L10N usage in the .js file.

This screenshot displays the result on a local instance of WCAG2.1AA validation using the Wave extension for Firefox.
WaveAnalysis
We can see that this validation does not display any violation in the rights edit table, and that the button titles are properly set.



For reference, here is the tabbing order through the table (displayed with Firefox Inspector accessibility tool):
tabbingOrderRights

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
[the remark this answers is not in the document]
michitux: We rarely add comments, yes, but it would be a good idea to change that. Comments above the translation string are displayed in Weblate afaik so this should work.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -55,34 +64,62 @@ window.MSCheckbox = Class.create({
     ];
     this.labels = ['','',''];
 
+    
+    
+    this.button = document.createElement("button");
+    this.button.className = "rights-edit";
+    this.button.onclick = this.createClickHandler(this);
```
michitux, line 71: This looks like a quite old way to add event listener, wouldn't it be better to use [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)? I don't know what the best practices are, but may feeling is that the current way is not a best practice.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -2109,6 +2109,9 @@ platform.core.rightsManagement.editRightsForSpace=Editing access rights for spac
 platform.core.rightsManagement.ajaxFailure=An error occurred while communicating with the server. Please check that the server is accessible, and you have the proper rights to perform the requested action.
 platform.core.rightsManagement.saveFailure=An exception occurred while trying to save the current modifications. Please check if you have the proper rights to perform these modifications.
 platform.core.rightsManagement.saveComment={0} {1} right for {2}
+platform.core.rightsManagement.allowed=Allowed is the current state. Switch to denied.
+platform.core.rightsManagement.denied=Denied is the current state. Switch to undefined.
+platform.core.rightsManagement.undefined=Undefined is the current state. Switch to allowed.
```
michitux, line 2114: What do you think about something like "Allowed. Click to deny."/"Denied. Click to switch to undefined."/"Undefined. Click to allow." I'm not sure about the exact wording, either, but I think it could be interesting to have a shorter text when possible.

michitux (commented):
I have still some small comments but overall I think this looks good now.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -55,34 +64,62 @@ window.MSCheckbox = Class.create({
     ];
     this.labels = ['','',''];
 
+    
+    
+    this.button = document.createElement("button");
+    this.button.className = "rights-edit";
+    Event.observe(this.button, 'click', this.createClickHandler(this));
```
michitux, line 71: Is there any reason for using Prototype here? We're moving away from Prototype so I don't think it is a good idea to expand it's usage unless it is strictly required (see also, e.g., https://jira.xwiki.org/browse/XWIKI-17645).

Sereza7:
**PR Changes:**

Updated L10N key en values to shorter strings. 
Added resource comments to emphasize the importance of keeping the two elements in the key.
Fixed the listener instantiation.

___
Might need more changes to resolve [this conversation](https://github.com/xwiki/xwiki-platform/pull/2051#discussion_r1086906531).
The problem is that the former structure of the table cell had the possibility to keep and update a label. However, this label was not used at all. The question at hand is whether or not to remove this (unused) label.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
Sereza7: Thanks! I made the changes too quickly, I reverted it back to the way they added the listener before this PR, and didn't realize it was still different from the way you linked [above](https://github.com/xwiki/xwiki-platform/pull/2051#discussion_r1088768833).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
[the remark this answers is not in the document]
mflorea: I don't think this label was ever used, so I'm in favor or removing it.

Sereza7:
**PR Changes:**

- Removed the (unused) label from the rights-edit. See https://github.com/xwiki/xwiki-platform/pull/2051#discussion_r1086354379
- Fixed a comment in the css
- Fixed a bad entry in the applicationResources

**View:**

No changes since https://github.com/xwiki/xwiki-platform/pull/2051#issuecomment-1406249803

mflorea (approved):

PR #2052, opened 2023-01-18
[Misc] Add missing dependency in tool packager plugin

surli:
* Tool packager plugin seemed to miss a dependency to xwiki-platform-url-default which is needed


surli:
Closing in favor of https://github.com/xwiki/xwiki-platform/pull/2062

PR #2053, opened 2023-01-19
XWIKI-20549: Provide a new script service API to check trustfulness of an URI

michitux:
* Add missing quoting during URI parsing to avoid parse errors.

This fixes support for redirects to URLs that contain illegal characters like spaces in query strings.  However, this also changes how newlines are handled and requires for security that the parsed URI is used for redirect which could lead to unexpected results.

Jira issue: https://jira.xwiki.org/browse/XWIKI-20549

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
@@ -25,6 +25,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
 import java.util.Locale;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.servlet.ServletOutputStream;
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiServletResponse.java
```diff
@@ -75,14 +84,42 @@ public void sendRedirect(String redirect) throws IOException
...
                         redirect);
                 } else {
-                    this.response.sendRedirect(redirect);
+                    this.response.sendRedirect(uri.toString());
+                }
+            }
+        }
+    }
+
+    private URI parseURI(String location)
```
surli, line 93: Feels like this should also be part of an API: the reason is that I need to also parse a String to URI in the new URLSecurityScriptService, so would be useful to use the same code in all places where we have this need.

michitux: Using this code in `URLSecurityScriptService` is a very bad idea as there will be no guarantee that the original string representation is safe when the parsed URI is safe. I know examples where this is not the case, i.e., where the original string results in a different host in a browser than what the URI has as host.

surli: > Using this code in URLSecurityScriptService is a very bad idea as there will be no guarantee that the original string representation is safe when the parsed URI is safe. I know examples where this is not the case, i.e., where the original string results in a different host in a browser than what the URI has as host.

Then it means the script service is useless if we cannot check the safeness of the given URL

surli:
Closing in favor of #2054

PR #2056, opened 2023-01-20
XWIKI-20538: attachments are not returned by solr when the locale does not match the default locale

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20538

Attachments are linked with one of the locale of a page. But in the search, we restrict the locale to the current one.
Consequently, in mulilingual wikis, chances are the attachment search is partial.

The current solution is to include `fq=locale:*` to return the attachments regardless of their locale.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfig.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfig.xml
@@ -905,6 +905,7 @@ qf=doccontent^2 doccontentraw</query>
     </property>
     <property>
       <query>fq=type:ATTACHMENT
+fq=locale:*
 qf=filename</query>
     </property>
     <property>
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfig.xml
```diff
@@ -1066,6 +1067,7 @@ qf=filename</query>
     </property>
     <property>
       <query>fq=type:ATTACHMENT
+fq=locale:*
```
manuelleduc, line 1070: I'm less confident about that one. I think it means returning indexed content for all locales. But it's possible another property can be exploited to use the locale of the attachment itself (cc @mflorea)

mflorea (commented):
I don't like adding ``locale:*`` everywhere, because this means "give me all results that **have** a locale with any value" while in these cases we don't care about locale at all, i.e. "give me **all** the results, with or without a locale". Sure, in our case attachments are normally indexed with a locale, but semantically speaking I find it better to execute the query without any locale filtering. So I think we need to be able to say to the Solr search suggest service "don't add the default locale filtering".

PR #2061, opened 2023-01-24
XWIKI-20369: Use correct ellipsis for wysiwyg.admin.fontNames.add.hint

wthrajat:
- This time only English, French and German files are changed.
- Replaced 4 dots with an ellipsis in the key ```core.delete.warningExtensions.tree.paginationNode``` keeping the rules of ellipsis in Eng, French and German in mind
- Comments have not been changed
- Email domain has not been changed


michitux:
But now this doesn't fix anymore what the originally reported problem in XWIKI-20369 was?

wthrajat:
@michitux Ahh,There were three key pairs which had ``..`` (2 dots) which were left. Fixed them in the new commit.
Now it does fix what was asked in https://jira.xwiki.org/browse/XWIKI-20369 (including replacement of many 3 dots with a proper ellipsis) I guess

PR #2062, opened 2023-01-25
XWIKI-20578: tool packager plugin ouputs lots of errors related to missing URLConfiguration component

surli:
* Use a Provider<URLConfiguration> in InternalSkinManager to avoid dependency issues with tool packager plugin
  * Add a new ComponentManagerInitializer in the tool packager plugin to handle unregistering the stuff we don't need for packaging

Note that I only tested this by building  `xwiki-platform-distribution-flavor-data-hsqldb` so I might miss impacts on the other mojo than packager.

### DIFF
--- a/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/internal/PackagerComponentManagerInitializer.java
+++ b/xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/internal/PackagerComponentManagerInitializer.java
@@ -23,24 +23,30 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.xwiki.component.annotation.Component;
-import org.xwiki.context.ExecutionContext;
-import org.xwiki.context.ExecutionContextException;
+import org.xwiki.component.manager.ComponentManager;
+import org.xwiki.component.manager.ComponentManagerInitializer;
 import org.xwiki.context.ExecutionContextInitializer;
+import org.xwiki.observation.EventListener;
 
 /**
- * Cancel ThreadClassloaderExecutionContextInitializer to not mess with the Maven classloader.
- * 
+ * Dedicated component manager initializer for the packager plugin.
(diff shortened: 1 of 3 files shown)


surli:
Also note that it doesn't prevent some warnings when building `flavor-data-hsqldb` such as: 
> [WARNING] Failed to evaluate the Velocity code from the query [select doc.fullName, [space.name](http://space.name/) from XWikiDocument doc, XWikiSpace space where [doc.space](http://doc.space/) = space.reference and space.parent is null and doc.translation = 0 and [doc.name](http://doc.name/) = 'WebHome' order by [space.name](http://space.name/)]. Root cause is [ComponentLookupException: Can't find descriptor for the component with type [interface org.xwiki.url.URLConfiguration] and hint [default]]. Continuing with the raw query.

xwiki-platform-tools/xwiki-platform-tool-packager-plugin/src/main/java/org/xwiki/tool/packager/internal/PackagerComponentManagerInitializer.java
```diff
@@ -23,24 +23,30 @@
...
     @Override
-    public void initialize(ExecutionContext context) throws ExecutionContextException
+    public void initialize(ComponentManager componentManager)
     {
-        // Cancel
+        // We don't want the solr indexer to be triggered whenever a document is imported.
+        componentManager.unregisterComponent(EventListener.class, "solr.update");
+        // We don't need the untyped event listener and it requires specific dependencies.
+        componentManager.unregisterComponent(EventListener.class, "Untyped Event Listener");
+        componentManager.unregisterComponent(ExecutionContextInitializer.class, "threadclassloader");
```
tmortagne, line 50: Should probably add the comment that used to be in MavenBuildThreadClassloaderExecutionContextInitializer javadoc above that line.

surli: Indeed

PR #2064, opened 2023-01-25
XRENDERING-701: Introduce a "source" parameter in the content macro

tmortagne:
* add support for script and verious entities implementations of MacroContentWikiSourceFactory
* refactor source API introduced in XWIKI-20462 to reuse the new rendering one
* move EmailAddressObfuscator and GeneralMailConfiguration in an -API module so that they can be used without depending on xwiki-platform-oldcore

Depends on https://github.com/xwiki/xwiki-rendering/pull/243


michitux:
I'm absolutely +1 for the script source. However, I'm hesitant about supporting also all the other sources in the content macro. The reason for this is that using any of them will always produce a security issue unless you know exactly what you're doing.

tmortagne:
I actually implemented the various sources as a generic source for various macros, but I agree with you that in the specific case of the content macro it's most probably a good way to introduce security problems.

Now it could make sense for others, like the context macro (in which I also plan to add support for macro content wiki sources) so I would not like to completely remove those sources. One possibility would be to forbid some way entities sources for the content macro.

tmortagne:
I decided to postpone support for entities sources in the content macro to later, as I don't really have the need for now.

PR #2066, opened 2023-01-26
XWIKI-20575: Impossible to search pages created without locale

surli:
* Create a SolrSchemaUtils to provide APIs that might be needed in migrations
  * Reuse SolrSchemaUtils in the AbstractSolrCoreInitializer
  * Check in the migration if the field title__ exists for creating or not the dynamic field missing

WIP: Missing documentation and test when the schema contains the appropriate field.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
@@ -47,6 +47,7 @@ import org.apache.solr.schema.StrField;
 import org.apache.solr.schema.TextField;
 import org.xwiki.component.descriptor.ComponentDescriptor;
 import org.xwiki.search.solr.internal.DefaultSolrUtils;
+import org.xwiki.search.solr.internal.SolrSchemaUtils;
 
 /**
  * Base helper class to implement {@link SolrCoreInitializer}.
@@ -140,6 +141,9 @@ public abstract class AbstractSolrCoreInitializer implements SolrCoreInitializer
     @Inject
     protected ComponentDescriptor<SolrCoreInitializer> descriptor;
 
+    @Inject
+    protected SolrSchemaUtils solrSchemaUtils;
+
     protected SolrClient client;
 
     protected Map<String, FieldTypeRepresentation> types;
@@ -150,6 +154,7 @@ public abstract class AbstractSolrCoreInitializer implements SolrCoreInitializer
     public void initialize(SolrClient client) throws SolrException
     {
         this.client = client;
+        this.solrSchemaUtils.setClient(getCoreName(), this.client);
 
         // Make sure the base schema (mostly default types) in in sync with this version of XWiki
         initializeBaseSchema();
@@ -183,21 +188,7 @@ public abstract class AbstractSolrCoreInitializer implements SolrCoreInitializer
 
     protected Map<String, Map<String, Object>> getFields(boolean force) throws SolrException
     {
-        if (this.fields == null || force) {
-            SchemaResponse.FieldsResponse response;
-            try {
-                response = new SchemaRequest.Fields().process(this.client);
-            } catch (Exception e) {
-                throw new SolrException("Failed to get the list of fields", e);
-            }
-
-            Map<String, Map<String, Object>> map = new HashMap<>(response.getFields().size());
-
-            response.getFields().forEach(e -> map.put((String) e.get(SOLR_FIELD_NAME), e));
-
-            this.fields = map;
-        }
-
+        this.fields = this.solrSchemaUtils.getFields(getCoreName(), force);
         return this.fields;
     }
 
@@ -871,13 +862,9 @@ public abstract class AbstractSolrCoreInitializer implements SolrCoreInitializer
      */
     protected void setField(String name, String type, boolean dynamic, Object... attributes) throws SolrException
     {
-        Map<String, Object> fieldAttributes = new HashMap<>();
-        fieldAttributes.put(SOLR_FIELD_NAME, name);
-        fieldAttributes.put(FieldType.TYPE, type);
-
-        MapUtils.putAll(fieldAttributes, attributes);
-
-        setField(fieldAttributes, dynamic);
+        this.solrSchemaUtils.setField(getCoreName(), name, type, dynamic, attributes);
+        // Ensure to keep internal fields updated
+        getFields(false);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
```diff
@@ -890,28 +877,10 @@ protected void setField(String name, String type, boolean dynamic, Object... att
...
-            }
-
-            // Add it to the cache
-            this.fields.put(name, fieldAttributes);
-        } catch (Exception e) {
-            throw new SolrException("Failed to set a field in the Solr core", e);
-        }
+        this.solrSchemaUtils.setField(getCoreName(), fieldAttributes, dynamic);
+        // Ensure to keep internal fields updated
+        getFields(false);
```
tmortagne, line 883: This call seems useless. The new field is already added to the cache by #setField.

surli: Well not really anymore: `SolrSchemaUtils#setField` does add the new field to the `SolrSchemaUtils.SolrSchema#fields` but in AbstractSolrCoreInitializer, `fields` is a copy of  `SolrSchemaUtils.SolrSchema#fields` and not a reference to it. See: https://github.com/xwiki/xwiki-platform/pull/2066/files#diff-4b1b8c4ad281f72b8ea3f2415bf4111077639746b3ab520f7ba134958173c4b0R101

I hesitated between having a pointer or a copy to it, but it felt safer to return a copy to avoid any concurrency issue: by documenting that I want also to prevent any inconsistency in case someone calls commit, which empty the field map.

tmortagne: > avoid any concurrency issue

Well this copy is really error-prone and, to me, all it takes to avoid concurrency issues with the fields is to store a ConcurrentHAshMap instead of a HashMap. You don't even need to have a `fields` field on `AbstractSolrCoreInitializer` side.

surli: > You don't even need to have a fields field on AbstractSolrCoreInitializer side.

problem is that it's protected right now, so it's an API. 

> all it takes to avoid concurrency issues with the fields is to store a ConcurrentHashMap instead of a HashMap

yes I thought about that (was my first implementation actually), but in case of a call to `SolrSchemaUtils#commit` it means `AbstractSolrCoreInitializer#fields` will keep pointing an outdated map. Now maybe it's less error prone.

surli: So I changed my mind and I now use a ConcurrentHashMap. Not perfect, but I'm planning to improve all this in the future.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/internal/migration/R141004000XWIKI20575DataMigration.java
```diff
@@ -92,9 +104,20 @@ public XWikiDBVersion getVersion()
...
         try {
-            this.solrInstance.deleteByQuery("*:*");
-            this.solrInstance.commit();
-        } catch (SolrServerException | IOException e) {
+            Map<String, Map<String, Object>> fields =
+                this.solrSchemaUtils.getFields(SolrClientInstance.CORE_NAME, true);
+
+            // We only check for the presence of the title__ field, which should be representative of the dynamic field
+            if (!fields.containsKey("title__")) {
+                this.solrInstance.deleteByQuery("*:*");
```
surli, line 112: I have a doubt about that logic: if it's a remote solr then the schema would be updated by the new config file, but I'd still need to perform the reindex no? Meaning that I need to move this call outside the if?

tmortagne: Yes, I agree.

surli: I fixed it.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
```diff
@@ -1031,12 +996,7 @@ protected void setFieldType(FieldTypeDefinition definition, boolean add) throws
...
-        try {
-            this.client.commit();
-        } catch (Exception e) {
-            throw new SolrException("Failed to commit", e);
-        }
-
+        this.solrSchemaUtils.commit(getCoreName());
         // Reset the cache
         this.types = null;
         this.fields = null;
```
tmortagne, line 1002: This reset is not really needed anymore since you always call `solrSchemaUtils.getFields` (so nothing really check if it's null anymore).

surli: it avoids mistakes in case an inherited class use directly the field: it ensures that the field won't contain the previous map

surli: Ideally we should make this field private, but that would be an API change... wdyt?

tmortagne: > but that would be an API change

Yes, I almost suggested to deleted that field actually, but then I remembered that this class was public API so that would be an API breakage, unfortunately.

Anyway, I think we need to provide a totally new helper and deprecate all that in the long run.

surli: > Anyway, I think we need to provide a totally new helper and deprecate all that in the long run.

Agreed.

surli: See: https://jira.xwiki.org/browse/XWIKI-20581

PR #2067, opened 2023-01-27
XWIKI-20583: Provide a macro for sanitizing URLs in templates

surli:
* Add the new macro in macros.vm
  * Provide a new test in MacrosTest for covering it

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
@@ -306,11 +306,10 @@ class MacrosTest extends PageTest
         SecurityScriptService securityScriptService = mock(SecurityScriptService.class);
         this.oldcore.getMocker().registerComponent(ScriptService.class, "security", securityScriptService);
         URLSecurityScriptService urlSecurityScriptService = mock(URLSecurityScriptService.class);
-        this.oldcore.getMocker().registerComponent(ScriptService.class, "security.url", urlSecurityScriptService);
         when(securityScriptService.get("url")).thenReturn(urlSecurityScriptService);
 
-        String htmlElement = "input";
-        String attributeName = "value";
+        String htmlElement = "form";
+        String attributeName = "action";
         String url = "javascript:alert('foobar')";
         String fallbackUrl = "xwiki/bin/view/Sandbox";
         String script = "#getSanitizedURLAttributeValue($myHtmlElement,$myAttributeName,$myUrl,$myFallback,"
@@ -348,7 +347,7 @@ class MacrosTest extends PageTest
         out = new StringWriter();
         this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
         assertEquals("", velocityContext.get("sanitizedResult"));
-        // This should not have been called this time.
+        // Since the URLSecurityScriptService returns false for both URLs the HTMLScriptService should not be called.
         verify(htmlScriptService, times(2)).isAttributeSafe(htmlElement, attributeName, fallBackUri.toString());
 
         // Fourth check: the given URL is valid and there's no fallback
@@ -365,7 +364,7 @@ class MacrosTest extends PageTest
         this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
         assertEquals("http://xwiki.org/?query=foo&#38;bar=%20space%4F#éâ", velocityContext.get("sanitizedResult"));
 
-        // Fourth check: the given URL is not valid and there's no fallback
+        // Fifth check: the given URL is not valid and there's no fallback
         when(urlSecurityScriptService.parseToSafeURI(url)).thenReturn(null);
         out = new StringWriter();
         this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
@@ -21,19 +21,25 @@ package org.xwiki.web;
 
 import java.io.StringReader;
 import java.io.StringWriter;
+import java.net.URI;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.velocity.VelocityContext;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.script.ModelScriptService;
+import org.xwiki.script.service.ScriptService;
+import org.xwiki.security.script.SecurityScriptService;
 import org.xwiki.test.annotation.ComponentList;
 import org.xwiki.test.page.PageTest;
+import org.xwiki.url.script.URLSecurityScriptService;
 import org.xwiki.velocity.VelocityManager;
 import org.xwiki.velocity.internal.XWikiDateTool;
+import org.xwiki.xml.html.script.HTMLScriptService;
 
 import static java.util.Arrays.asList;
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
```diff
@@ -287,4 +297,78 @@ void displayUser() throws Exception
...
+    @Test
+    void getSanitizedURLAttributeValue() throws Exception
+    {
+        HTMLScriptService htmlScriptService = mock(HTMLScriptService.class);
+        this.oldcore.getMocker().registerComponent(ScriptService.class, "html", htmlScriptService);
+        SecurityScriptService securityScriptService = mock(SecurityScriptService.class);
+        this.oldcore.getMocker().registerComponent(ScriptService.class, "security", securityScriptService);
+        URLSecurityScriptService urlSecurityScriptService = mock(URLSecurityScriptService.class);
+        this.oldcore.getMocker().registerComponent(ScriptService.class, "security.url", urlSecurityScriptService);
+        when(securityScriptService.get("url")).thenReturn(urlSecurityScriptService);
```
manuelleduc, line 310: you probably don't event need to register "security.url" as it is revolved by `thenReturn` instead of a component manager inside the `get`

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
```diff
@@ -287,4 +297,78 @@ void displayUser() throws Exception
...
+
+        // Second check: given URL is not secure according to security.url, so we fallback but the HTMLScriptService
+        // does not allow the fallback URL.
+        when(htmlScriptService.isAttributeSafe(htmlElement, attributeName, fallBackUri.toString())).thenReturn(false);
+        out = new StringWriter();
+        this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
+        verify(htmlScriptService, times(2)).isAttributeSafe(htmlElement, attributeName, fallBackUri.toString());
+        assertEquals("", velocityContext.get("sanitizedResult"));
+
+        // Third check: both the given URL and the fallback URL are not secure according to security.url
```
manuelleduc, line 346: Unless the checks are dependent of each other, this feel like each check should be in its own test method.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
surli: right

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
surli: Old debate of structure of tests :) Here my argument to have a single method is that old my checks share the same fixture, and also the current structure of the test class is basically one method = one macro being tested. So I decided to follow the same.

manuelleduc: The only pro I see to a methods splitting is that you get more failures at once (and finer grained), but I won't fight for this :)

michitux: You can get the same effect by using `assertAll`, see https://junit.org/junit5/docs/current/user-guide/#writing-tests-assertions.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
```diff
@@ -287,4 +297,78 @@ void displayUser() throws Exception
...
+
+        URI xwikiOrgURI = new URI(url);
+        when(urlSecurityScriptService.parseToSafeURI(url)).thenReturn(xwikiOrgURI);
+        when(htmlScriptService.isAttributeSafe(htmlElement, attributeName, xwikiOrgURI.toString())).thenReturn(true);
+
+        out = new StringWriter();
+        this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
+        assertEquals("http://xwiki.org/?query=foo&#38;bar=%20space%4F#éâ", velocityContext.get("sanitizedResult"));
+
+        // Fourth check: the given URL is not valid and there's no fallback
```
michitux, line 368: Shouldn't this be "fifth" check?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
```diff
@@ -287,4 +297,78 @@ void displayUser() throws Exception
...
+        this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
+        verify(htmlScriptService, times(2)).isAttributeSafe(htmlElement, attributeName, fallBackUri.toString());
+        assertEquals("", velocityContext.get("sanitizedResult"));
+
+        // Third check: both the given URL and the fallback URL are not secure according to security.url
+        when(urlSecurityScriptService.parseToSafeURI(fallbackUrl)).thenReturn(null);
+        out = new StringWriter();
+        this.velocityManager.evaluate(out, "getSanitizedURLAttributeValue", new StringReader(script));
+        assertEquals("", velocityContext.get("sanitizedResult"));
+        // This should not have been called this time.
```
michitux, line 351: I don't understand this comment.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
```diff
@@ -287,4 +297,78 @@ void displayUser() throws Exception
...
+        HTMLScriptService htmlScriptService = mock(HTMLScriptService.class);
+        this.oldcore.getMocker().registerComponent(ScriptService.class, "html", htmlScriptService);
+        SecurityScriptService securityScriptService = mock(SecurityScriptService.class);
+        this.oldcore.getMocker().registerComponent(ScriptService.class, "security", securityScriptService);
+        URLSecurityScriptService urlSecurityScriptService = mock(URLSecurityScriptService.class);
+        this.oldcore.getMocker().registerComponent(ScriptService.class, "security.url", urlSecurityScriptService);
+        when(securityScriptService.get("url")).thenReturn(urlSecurityScriptService);
+
+        String htmlElement = "input";
+        String attributeName = "value";
```
michitux, line 313: I mean you mock everything here so it doesn't matter, but for the sake of keeping things realistic, I would have used something like `form` and `action` or `a` and `href` as it doesn't make sense to use this macro for the value of an input element.

michitux (approved):
Looks good in general, just some minor comments on the test.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/MacrosTest.java
surli: Sure I could change that, here I was thinking about `<input type="button" value="foo"/>` when I wrote the test.

michitux: Okay, but even for buttons, this is the label that is displayed on the button, so there is absolutely no need to sanitize anything here. The HTML sanitizer explicitly treats `value` as a safe attribute, so it wouldn't return `false` regardless which value you provide.

surli: Yes you're right, I was planning to change it after reading your comment anyway

PR #2068, opened 2023-01-27
XWIKI-20585: Warn the user when the dimensions are larger than the selected image

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20585


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/Translations.xml
```diff
@@ -215,7 +215,7 @@ imageEditor.modal.loadFail.message=Failed to retrieve the image editor form.
 imageEditor.modal.title=Edit Image
 imageEditor.modal.insertButton=Insert
 imageEditor.modal.initialization.fail=Failed to retrieve the image selection form.
-
+imageEditor.modal.outscaleWarning=One of the dimensions is larger than the original image ({0}). This currently results in undefined behavior. It is not recommended to do this unless you are sure you understand the consequences.
```
mflorea, line 218: "undefined behavior" sounds a bit strange to me. Maybe "This can lead to unexpected results, especially on small screens.".

mflorea (approved):
Looks good otherwise.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/Translations.xml
manuelleduc: +1 for using "unexpected results" instead of "undefined behavior" but I don't get the "on small screens" part.

mflorea: > I don't get the "on small screens" part

I thought outscaling creates problems more often on small screens. If you have a large screen you may not notice the problem, but of course it depends how much you outscale. You can drop it if you think it doesn't make sense.

PR #2069, opened 2023-01-28
XWIKI-17588: Display some text when there's no tag for a page

wthrajat:
Summary:
When there are no tags:

When user **has edit rights**, ```Tags [+]``` is displayed.
When user **does NOT have edit rights**, nothing is displayed at all.

When there are tags, it is displayed to all type of users irrespective of their edit rights.
(But user with edit rights can see the ```+``` sign as well :)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
@@ -138,7 +138,8 @@
     #end
     #if($hasedit || !$tags.isEmpty())
         $services.localization.render('core.tags.list.label')
-        #foreach($tag in $tags)
+    #end
+    #foreach($tag in $tags)
         #displayTag($tag)
     #end
     #if($hasedit && $xwiki.tag)
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
@@ -152,7 +152,7 @@
             #else
                 #displayAddForm()
             #end
-        </div>##
+        </div>
     #end
 </div>
     #if($tagErrorMessage != '')
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
@@ -136,20 +136,13 @@
         #set($hasTagsPlugin = false)
         #set($tags = $doc.getTagList())
     #end
-    ##   #if($!tags.size() == 0)
-    ##     No tags
-    ##   #else
-    $services.localization.render('core.tags.list.label')
-    #foreach($tag in $tags)
+    #if($hasedit || !$tags.isEmpty())
+        $services.localization.render('core.tags.list.label')
+        #foreach($tag in $tags)
         #displayTag($tag)
     #end
-    ##   #end
-    #if($hasedit)
-        #if($xwiki.tag)
-          <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-        #else
-            ## TODO
-        #end
+    #if($hasedit && $xwiki.tag)
+        <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
     #end
 </div>
     #if($tagErrorMessage != '')
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
@@ -143,7 +143,16 @@
         #displayTag($tag)
     #end
     #if($hasedit && $xwiki.tag)
-        <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
+        <div class="tag-tool tag-add">##
+            #if("$!{request.showTagAddForm}" == '')
+                <a href="$doc.getURL('view', 
+                    "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" 
+                    title="$services.localization.render('core.tags.add.tooltip')" 
+                    rel="nofollow">[+]</a>##
+            #else
+                #displayAddForm()
+            #end
+        </div>##
     #end
 </div>
     #if($tagErrorMessage != '')
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
@@ -87,7 +87,7 @@ class DocumentTagsTest extends PageTest
         // Remove extra spaces to make it easy to assert the result below.
         String result = templateManager.render("documentTags.vm").trim().replaceAll("\\s+", " ");
 
-        // Verify that the generated HTML matches the expectation:
+        // Verify that the generated HTML matches the expectations:
         // - The tag label is not displayed since there is/are no tag(s)
         // - No tag is listed after the tag label
         // - No "+" link is displayed since the user doesn't have edit rights
@@ -104,7 +104,7 @@ class DocumentTagsTest extends PageTest
         // Remove extra spaces to make it easy to assert the result below.
         String result = templateManager.render("documentTags.vm").trim().replaceAll("\\s+", " ");
 
-        // Verify that the generated HTML matches the expectation:
+        // Verify that the generated HTML matches the expectations:
         // - The tag label is displayed
         // - No tag is listed after the tag label
         // - The "+" link is displayed since the user has edit rights
@@ -128,7 +128,7 @@ class DocumentTagsTest extends PageTest
         // Remove extra spaces to make it easy to assert the result below.
         String result = templateManager.render("documentTags.vm").trim().replaceAll("\\s+", " ");
 
-        // Verify that the generated HTML matches the expectation:
+        // Verify that the generated HTML matches the expectations:
         // - The tag label is displayed
         // - The tags after the tag label
         // - No "+" link is displayed since the user doesn't have edit rights
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
@@ -109,8 +109,8 @@ class DocumentTagsTest extends PageTest
         // - No tag is listed after the tag label
         // - The "+" link is displayed since the user has edit rights
         assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label "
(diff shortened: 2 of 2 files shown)


vaibhav-init:
@wthrajat  have you self-tested this?

michitux:
Have you executed the tests, in particular `DocumentTagsTest`? It should be sufficient to build `xwiki-platform-web-templates` with Maven, ideally with the "quality" profile (`-Pquality`). I think at least one of the tests should fail now and needs to be corrected for the changed template.

wthrajat:
2023-02-01-15 23 09-screenshot
Test is done and build is successful now.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -90,7 +90,7 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndNoTags() throws Exceptio
         // - The tag label is displayed
```
michitux, line 90: This is not true anymore, please update this comment.

michitux (changes requested):
This looks good to me and I think this matches what was discussed on the previous pull requests #1559 and #1579. I have some minor comments regarding the tests but apart from that this looks good.

The only thing to improve would be to provide before/after screenshots but I think it should be fine.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -90,7 +90,7 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndNoTags() throws Exceptio
...
         // - No tag is listed after the tag label
         // - No "+" link is displayed since the user doesn't have edit rights
-        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label </div>"));
+        assertThat(result, matchesPattern("<div class=\"doc-tags\" id=\"xdocTags\"> </div>"));
```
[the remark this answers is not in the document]
wthrajat: In my opinion, the reason for using ```matchesPattern``` instead of ```containsString``` is to check the exact format of the string. ```containsString``` only checks if a substring is present in the result, but it doesn't verify if it's in the correct format or if there are no extra or missing spaces/characters. By using a pattern, you can check for exact string format and avoid false negatives due to differences in formatting.

wthrajat:
**BEFORE** <hr>
**When no edit rights (we had problem with this only)**
2023-02-01-16 20 08-screenshot

**AFTER** <hr>

**1) When has edit rights**
no tag, has edit
tag, has edit

**2) When no edit rights**
no tag, no edit
tag, no edit

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
wthrajat: Done with the latest commit.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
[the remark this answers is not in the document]
michitux: Okay, I think you're right that we're in fact matching the whole string here, not a substring. But in this case it is even easier: Just replace it by an `assertEquals`
```suggestion
        assertEquals("<div class=\"doc-tags\" id=\"xdocTags\"> </div>", result);
```

You'll also need to add an appropriate import:
```java
import static org.junit.jupiter.api.Assertions.assertEquals;
```

There are other places where the pattern makes sense as it contains a match for an appropriate URL for example, but I think here it's just useless.

michitux (approved):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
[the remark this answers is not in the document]
wthrajat: Applied ```assertEquals``` :)

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -43,6 +43,7 @@
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.matchesPattern;
 import static org.mockito.Mockito.when;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
michitux, line 46: Imports should be ordered alphabetically within their group. In general, you should use an IDE, ideally Eclipse or IntelliJ and [configure it to use XWiki's code style](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HConfiguringyourIDEtousetheXWikicodestyle).
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.mockito.Mockito.when;
```

michitux (approved):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
```diff
@@ -136,20 +136,14 @@
...
     #end
-    ##   #end
-    #if($hasedit)
-        #if($xwiki.tag)
-          <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
-        #else
-            ## TODO
-        #end
+    #if($hasedit && $xwiki.tag)
+        <div class="tag-tool tag-add">#if("$!{request.showTagAddForm}" == '')<a href="$doc.getURL('view', "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" title="$services.localization.render('core.tags.add.tooltip')" rel="nofollow">[+]</a>#else #displayAddForm()#end</div>
```
manuelleduc, line 146: This line is too long and looks easy to split to make it more readable. Don't forget to use `##` at the end of lines to prevents unwanted line break in the resulting html.

manuelleduc (approved):
Approved with two remarks:
- the line too long code style issue I raised in the html
- while fixing broken tests is a good (and mandatory) first step, it would be nice to introduce new tests to cover the various new cases introduced by this work

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
michitux: I agree, but imho it's fine to keep it as-is as this pull request didn't change anything in this line (apart from the indentation). So, yes, would be a nice improvement but imho not necessary as part of this pull request (but it also doesn't hurt).

manuelleduc: I agree, not mandatory at all.
I mainly raised this comment because I had a hard time understanding if something had changed, because of the too long line.
So changing this is mainly to be nice to future maintainers of this code :)

wthrajat:
@michitux, @manuelleduc:

1. The long line has been splitted
2. Fixed another broken test case which resulted due to point no. 1
3. Added another test case (there were 3 test cases out of total 4)
4. ```mvn clean install -Pquality``` checked successfully with the latest commit
5. Tags are still working as like shown in the screenshots

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -144,4 +144,41 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndTags() throws Exception
...
+
+    @Test
+    void displayTagsWhenEditRightsAndTagPluginAvailableAndTags() throws Exception
+    {
+        // Give edit rights ($hasEdit = true)
+        when(this.oldcore.getMockContextualAuthorizationManager().hasAccess(Right.EDIT)).thenReturn(true);
+
+        // Add tags to the current document
+        XWikiDocument currentDocument = this.context.getDoc();
+        BaseObject bo = new BaseObject();
```
manuelleduc, line 156:
```suggestion
        BaseObject baseObject = new BaseObject();
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
```diff
@@ -143,7 +143,16 @@
...
+        <div class="tag-tool tag-add">##
+            #if("$!{request.showTagAddForm}" == '')
+                <a href="$doc.getURL('view', 
+                    "showTagAddForm=true&amp;$!{escapetool.url(${request.queryString})}#${tagsId}")" 
+                    title="$services.localization.render('core.tags.add.tooltip')" 
+                    rel="nofollow">[+]</a>##
+            #else
+                #displayAddForm()
+            #end
+        </div>##
```
manuelleduc, line 155: That last one is not mandatory, it wasn't here before the code style improvement.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -103,31 +104,31 @@ void displayTagsWhenEditRightsAndTagPluginAvailableAndNoTags() throws Exception
...
-        // Verify that the generated HTML matches the expectation:
+        // Verify that the generated HTML matches the expectations:
         // - The tag label is displayed
         // - No tag is listed after the tag label
         // - The "+" link is displayed since the user has edit rights
         assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label "
-            + "<div class=\"tag-tool tag-add\"><a href=\"\\E.*\"\\Q title=\"core.tags.add.tooltip\" "
-            + "rel=\"nofollow\">[+]</a></div> </div>\\E"));
+            + "<div class=\"tag-tool tag-add\"> <a href=\"/xwiki/bin/view/space/page?showTagAddForm=true#xdocTags\" title=\"core.tags.add.tooltip\" "
+            + "rel=\"nofollow\">[+]</a> </div> </div>"));
```
michitux, line 113: This shows that you've introduced some spaces. I would prefer if we could avoid that as I think spaces between the elements could be visible (haven't checked, though). Maybe you could check the Velocity and introduce some comments or put the line breaks at other places to avoid them?

Also, now that you've put the full URL in the string, I think you could/should use `assertEquals` instead of a regular expression.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
```diff
@@ -143,4 +144,41 @@ void displayTagsWhenNoEditRightsAndTagPluginAvailableAndTags() throws Exception
...
+
+        // Verify that the generated HTML matches the expectations:
+        // - The tag label is displayed
+        // - The tags after the tag label
+        // - The "+" link is displayed since the user has edit rights
+        assertThat(result, matchesPattern("\\Q<div class=\"doc-tags\" id=\"xdocTags\"> core.tags.list.label "
+        + "<span class=\"tag-wrapper\"> <span class=\"tag\">"
+        + "<a href=\"/xwiki/bin/view/Main/Tags?do=viewTag&amp;tag=tag1\">tag1</a></span> "
+        + "<span class=\"separator\">[</span>"
+        + "<a href=\"/xwiki/bin/view/space/page?xpage=documentTags&amp;xaction=delete&amp;tag=tag1&amp;form_token=&amp;xredirect=%2Fxwiki%2Fbin%2Fview%2Fspace%2Fpage%23xdocTags\" "
```
michitux, line 174: I'm wondering if it really makes sense to assert the full URL or if this wouldn't be a good place to actually use the fact that you're in a regular expression and just, e.g., match `[^"]*` inside the URL (at least after some point, you could still match, e.g., the start of at least the first tag's URL).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/test/java/org/xwiki/web/DocumentTagsTest.java
wthrajat: I checked and found out that these extra 2 spaces (they weren't there before), was the result of splitting the ```#if($hasedit && $xwiki.tag)``` even though I have put ```##``` at the end of HTML tags as suggested by @manuelleduc, I am not sure why the 2 extra space is being generated in the velocity.
The velocity needs to be changed now. Doing it...

manuelleduc: @wthrajat did you try by adding some more `##` on other newly introduced linebreaks? The nice thing with the page tests is that you can quickly iterate by editing the velocity and running the tests.

PR #2073, opened 2023-02-02
XWIKI-20602: Server-side image resizing does not preserver image ratio in case of outscaling

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20602

### DIFF
--- a/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/main/java/com/xpn/xwiki/plugin/image/ImagePlugin.java
+++ b/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/main/java/com/xpn/xwiki/plugin/image/ImagePlugin.java
@@ -367,33 +367,53 @@ public class ImagePlugin extends XWikiDefaultPlugin
     private int[] reduceImageDimensions(int currentWidth, int currentHeight, int requestedWidth, int requestedHeight,
         boolean keepAspectRatio)
     {
-        double aspectRatio = (double) currentWidth / (double) currentHeight;
+        int[] dimensions;
+        // Keep the aspect ratio when requested or width or height are missing.
+        if (keepAspectRatio || requestedWidth <= 0 || requestedHeight <= 0) {
+            dimensions = reduceImageDimensions(currentWidth, currentHeight, requestedWidth, requestedHeight);
+        } else if (requestedWidth > currentWidth || requestedHeight > currentHeight) {
+            dimensions = reduceImageDimensions(requestedWidth, requestedHeight, currentWidth, currentHeight);
+        } else {
+            dimensions = new int[] { requestedWidth, requestedHeight };
+        }
 
+        return dimensions;
+    }
+
+    /**
+     * Reduce the image of dimension {@code currentWidth*currentHeight} so that it fits a box of dimension
+     * {@code requestedWidth*requestedHeight} while preserving its current dimensions. If {@code requestedWidth} or
+     * {@code requestedHeight} is 0, negative, or larger than respectively {@code currentWidth} and
+     * {@code currentHeight}, it is ignored.
+     *
+     * @param currentWidth the current width of the image
+     * @param currentHeight the current height of the image
+     * @param requestedWidth the requested width of the image
+     * @param requestedHeight the requested height of the image
+     * @return the new dimensions of the image, preserving to current aspect ratio.
+     */
+    private int[] reduceImageDimensions(int currentWidth, int currentHeight,
+        int requestedWidth, int requestedHeight)
+    {
         int width = currentWidth;
         int height = currentHeight;
 
-        // Keep the aspect ratio when requested or width or height are missing.
-        if (keepAspectRatio || requestedWidth <= 0 || requestedHeight <= 0) {
-            // Ignore the width if it is not given or too large, i.e., larger than the current width or larger than the
-            // width derived from the requested height.
-            if (requestedWidth <= 0 || requestedWidth >= currentWidth
-                || (requestedHeight > 0 && requestedWidth > (int) (requestedHeight * aspectRatio)))
-            {
-                // Ignore the requested width. Check the requested height.
-                if (requestedHeight > 0 && requestedHeight < currentHeight) {
-                    // Reduce the height, keeping aspect ratio.
-                    width = (int) (requestedHeight * aspectRatio);
-                    height = requestedHeight;
-                }
-            } else {
-                // Ignore the requested height. Reduce the width, keeping aspect ratio.
-                width = requestedWidth;
-                height = (int) (requestedWidth / aspectRatio);
+        double aspectRatio = (double) currentWidth / (double) currentHeight;
+        // Ignore the width if it is not given or too large, i.e., larger than the current width or larger than the
+        // width derived from the requested height.
+        if (requestedWidth <= 0 || requestedWidth >= currentWidth
+            || (requestedHeight > 0 && requestedWidth > (int) (requestedHeight * aspectRatio)))
+        {
+            // Ignore the requested width. Check the requested height.
+            if (requestedHeight > 0 && requestedHeight < currentHeight) {
+                // Reduce the height, keeping aspect ratio.
+                width = (int) (requestedHeight * aspectRatio);
+                height = requestedHeight;
             }
         } else {
-            // Reduce both width and height, possibly loosing aspect ratio.
-            width = Math.min(requestedWidth, currentWidth);
-            height = Math.min(requestedHeight, currentHeight);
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/test/java/com/xpn/xwiki/plugin/image/ImagePluginTest.java
```diff
@@ -242,7 +242,9 @@ void scaling(int width, int height, boolean keepAspectRatio, int expectedWidth,
 
         this.plugin.downloadAttachment(attachment, xcontext);
         if (expectedWidth == image.getWidth(null) && expectedHeight == image.getHeight(null)) {
-            verify(this.imageProcessor, never()).scaleImage(any(), anyInt(), anyInt());
+            // If the returned image has the same dimensions as the original, it shouldn't be because the requested 
+            // dimensions are the same as the original.
+            verify(this.imageProcessor, never()).scaleImage(any(), eq(expectedWidth), eq(expectedHeight));
```
michitux, line 247: Why did you change this? Now the assertion won't trigger when the `scaleImage` method is called with wrong dimensions.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/main/java/com/xpn/xwiki/plugin/image/ImagePlugin.java
```diff
@@ -367,33 +367,85 @@ private XWikiAttachment shrinkImage(XWikiAttachment attachment, int requestedWid
...
+        int requestedHeight)
+    {
+        int width = currentWidth;
+        int height = currentHeight;
+        double requestedVerticalRatio = (double) requestedWidth / (double) requestedHeight;
+        double requestedHorizontalRatio = (double) requestedHeight / (double) requestedWidth;
+
+        double originalVerticalRatio = (double) currentWidth / (double) currentHeight;
+
+        if (requestedHorizontalRatio > originalVerticalRatio) {
```
michitux, line 445: It is not clear to me why this is the correct condition. I think this either needs a comment or the code should be simplified. For example, you could first calculate the width and if it is too large, reset it and calculate the height instead.

michitux (changes requested):

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/test/java/com/xpn/xwiki/plugin/image/ImagePluginTest.java
manuelleduc: Because it was too restrictive. With the `"800, 600, false, 400, 300"` case, even though we initially don't request for the same dimensions as the original image, we end up with a call to `scaleImage` returning the same dimensions as the original image because the requested dimensions are larger than the original image, but with the same ratio.
Now we check that `scaleImage` is never called with the exact initial dimensions only.

michitux: I don't understand this argument. The check is executed when the **expected** (and not the **requested**) width and height match the image's width and height.

michitux: The important part here is to assert that the algorithm indeed calculated the expected width and height and thus skipped the call. If you only assert that there was no call with the expected width and height, you haven't asserted anything as when the algorithm is wrong, it will trigger a call with different values and this isn't noticed by this assert.

manuelleduc: yes, and in the case of "800, 600, false, 400, 300" the expected width and height match the image's width and height (400x300) and it is nonetheless expected for `scaleImage` to be called.

michitux: Why is it expected for `scaleImage` to be called? The code will always skip the call to `scaleImage` when the calculated dimensions match the image:

https://github.com/xwiki/xwiki-platform/blob/df79bd8c5ed4abba20689bafaf5e38cae0157689/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/main/java/com/xpn/xwiki/plugin/image/ImagePlugin.java#L322-L332

Why is it expected that `scaleImage` is called? And with which dimensions? I mean if you expect it to be called with `400` and `300`, then this assertion cannot pass. And if you expect it to be called with different values, then you should adjust the expected values, as obviously this is wrong then.

manuelleduc: That was indeed a but on my code :)

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/main/java/com/xpn/xwiki/plugin/image/ImagePlugin.java
```diff
@@ -367,33 +367,53 @@ private XWikiAttachment shrinkImage(XWikiAttachment attachment, int requestedWid
     private int[] reduceImageDimensions(int currentWidth, int currentHeight, int requestedWidth, int requestedHeight,
         boolean keepAspectRatio)
     {
-        double aspectRatio = (double) currentWidth / (double) currentHeight;
+        int[] dimensions;
+        // Keep the aspect ratio when requested or width or height are missing.
+        if (keepAspectRatio || requestedWidth <= 0 || requestedHeight <= 0) {
+            dimensions = reduceImageDimensions(currentWidth, currentHeight, requestedWidth, requestedHeight);
+        } else if (requestedWidth > currentWidth || requestedHeight > currentHeight) {
+            dimensions = reduceImageDimensions(requestedWidth, requestedHeight, currentWidth, currentHeight);
```
michitux, line 375: I think this is quite nice but a comment to explain why this is done and why this works would be nice.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/test/java/com/xpn/xwiki/plugin/image/ImagePluginTest.java
```diff
@@ -183,10 +183,10 @@ void cacheOfScaledAttachment() throws Exception
 
     @ParameterizedTest
     @CsvSource({
-        "100, 500, false, 100, 300",
-        "500, 100, false, 400, 100",
+        "100, 500, false, 60, 300",
+        "500, 100, false, 400, 80",
+        "800, 600, false, 400, 300",
```
michitux, line 188: Maybe it would also be interesting to add another case where both dimensions are larger than the input but don't match the input aspect ratio?

michitux (approved):
I think this is much better now but an additional test and a comment to explain the trick would be nice.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/test/java/com/xpn/xwiki/plugin/image/ImagePluginTest.java
manuelleduc: done

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-processing/xwiki-platform-image-processing-plugin/src/main/java/com/xpn/xwiki/plugin/image/ImagePlugin.java
manuelleduc: I've improved to doc of `reduceImageDimensions`, let me know if that does not make things better.

PR #2074, opened 2023-02-03
XWIKI-20615: $response.sendRedirect repair too much

surli:
* Avoid double-encoding when trying to repair a URI by escaping % that belongs to percent encoded byte and replacing them afterwards
  * Add new various test cases related to reparation of URI with %

### DIFF
--- a/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
+++ b/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
@@ -37,6 +37,15 @@ import org.xwiki.stability.Unstable;
 public interface URLSecurityManager
 {
     /**
+     * Dedicated string used to escape {@code %} character.
+     * @see #parseToSafeURI(String)
+     * @since 15.1RC1
+     * @since 14.10.5
+     */
+    @Unstable
+    String PERCENT_ESCAPE = "__XWIKI_URL_SECURITY_PERCENT__";
+
+    /**
      * Constant to be used in {@link org.xwiki.context.ExecutionContext} with the value {@code "true"} to bypass a
      * check of {@link #isDomainTrusted(URL)}.
      */
@@ -82,6 +91,10 @@ public interface URLSecurityManager
      * This method throws a {@link SecurityException} if the parsed URI is not safe to use according to
      * {@link #isURITrusted(URI)}. It might also throw a {@link URISyntaxException} if the parameter cannot be properly
      * parsed.
+     * Note that this method might try to "repair" URI that are not parsed correctly by {@link URI#URI(String)}
+     * (e.g. serialized uri containing spaces). As part of that repair we need to perform escaping of the {@code %}
+     * character and we use {@link #PERCENT_ESCAPE} value as replacement string: the chain contained in
+     * {@link #PERCENT_ESCAPE} is then a protected token that should never be present in your serialized URI.
      *
      * @param serializedURI a string representing a URI that needs to be parsed.
      * @return a URI safe to use
--- a/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
+++ b/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
@@ -37,15 +37,6 @@ import org.xwiki.stability.Unstable;
 public interface URLSecurityManager
 {
     /**
-     * Dedicated string used to escape {@code %} character.
-     * @see #parseToSafeURI(String)
-     * @since 15.1RC1
-     * @since 14.10.5
-     */
-    @Unstable
-    String PERCENT_ESCAPE = "__XWIKI_URL_SECURITY_PERCENT__";
-
-    /**
      * Constant to be used in {@link org.xwiki.context.ExecutionContext} with the value {@code "true"} to bypass a
      * check of {@link #isDomainTrusted(URL)}.
      */
@@ -92,9 +83,7 @@ public interface URLSecurityManager
      * {@link #isURITrusted(URI)}. It might also throw a {@link URISyntaxException} if the parameter cannot be properly
      * parsed.
      * Note that this method might try to "repair" URI that are not parsed correctly by {@link URI#URI(String)}
-     * (e.g. serialized uri containing spaces). As part of that repair we need to perform escaping of the {@code %}
-     * character and we use {@link #PERCENT_ESCAPE} value as replacement string: the chain contained in
-     * {@link #PERCENT_ESCAPE} is then a protected token that should never be present in your serialized URI.
+     * (e.g. serialized uri containing spaces).
      *
      * @param serializedURI a string representing a URI that needs to be parsed.
      * @return a URI safe to use
--- a/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
+++ b/xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
@@ -59,6 +59,8 @@ import com.xpn.xwiki.XWikiContext;
 public class DefaultURLSecurityManager implements URLSecurityManager
 {
     private static final char DOT = '.';
+    private static final char PERCENT = '%';
+    private static final String PERCENT_ESCAPE = "__XWIKI_PERCENT__";
 
     // Regular expression taken from https://www.rfc-editor.org/rfc/rfc3986#appendix-B.
     private static final Pattern URI_PATTERN =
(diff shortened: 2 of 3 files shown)


xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -223,10 +225,16 @@ public URI parseToSafeURI(String serializedURI) throws URISyntaxException, Secur
...
+                String path = replaceUnquotedPercent(matcher.group(5));
+                String query = replaceUnquotedPercent(matcher.group(7));
+                String fragment = replaceUnquotedPercent(matcher.group(9));
+                // This constructor automatically encodes all % characters, that's why we replaced them all first by
+                // a chain if they belonged to a percent-encoded byte: if they don't, they will be encoded and that's
+                // correct. By doing so we avoid any double-encoding.
                 uri = new URI(scheme, authority, path, query, fragment);
+                // the URI should be parsed again after properly replacing the escape chain by the % character since
+                // it won't be encoded anymore with the single argument constructor.
+                uri = new URI(uri.toString().replaceAll(PERCENT_ESCAPE, "%"));
```
michitux, line 237: Wrong replace-method (should be `replace`, not `replaceAll`).

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -223,10 +225,16 @@ public URI parseToSafeURI(String serializedURI) throws URISyntaxException, Secur
             if (matcher.matches()) {
                 String scheme = matcher.group(2);
                 String authority = matcher.group(4);
```
michitux, line 227: I'm wondering if we should do something with the authority, too, according to https://url.spec.whatwg.org/#hosts-(domains-and-ip-addresses) there could be percent-encoding in the host, too.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -59,6 +59,8 @@
 public class DefaultURLSecurityManager implements URLSecurityManager
 {
     private static final char DOT = '.';
+    private static final char PERCENT = '%';
+    private static final String PERCENT_ESCAPE = "__XWIKI_PERCENT__";
```
michitux, line 63: I wonder if it's okay to assume that this pattern won't be in any URL parameter as the URL can contain user input. What would be safe, imho, is to have a pattern with, e.g., `0__` at the end and to increment that `0` until the pattern isn't part of the input anymore.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/test/java/org/xwiki/url/internal/DefaultURLSecurityManagerTest.java
```diff
@@ -464,5 +464,31 @@ void parseToSafeURI() throws URISyntaxException
 
         location = "foo://www.x wiki.org/";
         assertParseToSafeThrowSecurity(location, "foo://www.x%20wiki.org/");
+
+        location = "/xwiki/bin/edit/GroupeInformatique/Comptes%20Rendus/03%20f%C3%A9vrier%202023";
```
tmortagne, line 468: Maybe use something more common than this very french example :)

surli: Well French is handy to have plenty of accents :) but ok

michitux (changes requested):
I wonder if it wouldn't have been cleaner/easier to understand to do the initial replacement with a regular expression that matches directly all `%`  that are followed by two quoted characters like `%(?=[0-9A-Fa-f]{2})` but okay, I think that's a matter of style. Apart from that a few small comments.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: I hesitated to add it. I'll change that and cover it with test.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
surli: Frankly the chance that this pattern is used somewhere are very low. I'm really not in favor of adding another layer of complexity here. Now we should probably have some documentation in the API about that escape String, and it should be part of the API.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
```diff
@@ -82,6 +91,10 @@ default boolean isURITrusted(URI uri)
      * This method throws a {@link SecurityException} if the parsed URI is not safe to use according to
      * {@link #isURITrusted(URI)}. It might also throw a {@link URISyntaxException} if the parameter cannot be properly
      * parsed.
+     * Note that this method might try to "repair" URI that are not parsed correctly by {@link URI#URI(String)}
+     * (e.g. serialized uri containing spaces). As part of that repair we need to perform escaping of the {@code %}
+     * character and we use {@link #PERCENT_ESCAPE} value as replacement string: the chain contained in
+     * {@link #PERCENT_ESCAPE} is then a protected token that should never be present in your serialized URI.
```
surli, line 97: Note that I even hesitated to add a check in the beginning of the implem to immediately return an error if the token is used. But it feels a bit too much.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
```diff
@@ -36,6 +36,15 @@
...
 public interface URLSecurityManager
 {
+    /**
+     * Dedicated string used to escape {@code %} character.
+     * @see #parseToSafeURI(String)
+     * @since 15.1RC1
+     * @since 14.10.5
+     */
+    @Unstable
+    String PERCENT_ESCAPE = "__XWIKI_URL_SECURITY_PERCENT__";
```
michitux, line 46: I'm not sure this should be part of the API as it's really an implementation detail that we need to use a placeholder because of the not-so-nice Java URI implementation.

michitux (approved):
Looks good, I'm not sure about the API part but as we have `URI` in the API and this won't change it probably makes sense to also add the constant.

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-api/src/main/java/org/xwiki/url/URLSecurityManager.java
surli: I agree that it's a technical detail, but making it an API is a good way to ensure we won't change it by accident and that it's publicly documented for any consumer of the API.

tmortagne: > but making it an API is a good way to ensure we won't change it by accident

Changing it is not a problem as long as the tests are passing. I agree with @michitux that it's not a good idea to make this an API.

surli: > Changing it is not a problem as long as the tests are passing.

I don't agree, the purpose here to inform external consumer of the API about the token that should not be used in an URI: we don't have any test about that right now. As I mentioned in a comment, I'm hesitating to actually directly throw an error if such token is found. Do you think it's a better idea? If I use that protection, I could indeed avoid making that an API and enforce the behaviour with a test.

tmortagne: > Do you think it's a better idea?

The best would be to support inputs with such values, but you decided it does not worth it because "the chance that this pattern is used somewhere are very low". Making this hack public make things worst because I'm sure one say we will want to improve this.

In the meantime, yes, producing an error for an input you don't support, which is the subject here, is always more clear.

surli: > In the meantime, yes, producing an error for an input you don't support, which is the subject here, is always more clear.

ok will do that

xwiki-platform-core/xwiki-platform-url/xwiki-platform-url-default/src/main/java/org/xwiki/url/internal/DefaultURLSecurityManager.java
```diff
@@ -214,6 +220,11 @@ public boolean isURITrusted(URI uri)
     public URI parseToSafeURI(String serializedURI) throws URISyntaxException, SecurityException
     {
         URI uri;
+        if (serializedURI.contains(PERCENT_ESCAPE)) {
+            throw new IllegalArgumentException(
+                String.format("The given uri [%s] contains the string [%s] which is used internally "
+                + "for performing escaping operations. Please use another marker.", serializedURI, PERCENT_ESCAPE));
```
michitux, line 226: Wouldn't it make more sense to only perform this check when we're actually using this pattern, i.e., in the catch clause below?

surli: Right, it avoid a limitation for URI that we can parse with URI single constructor...

PR #2077, opened 2023-02-06
XWIKI-20290: Set correct content type in rendercontent.vm

michitux:
* Set the content type to text/plain when the output syntax is not HTML
* Add a page test

It would be possible to move the code to `contentvars.vm`, it just seemed strange to set the content type in this template (on the other hand, the rendered document could do the same). Any opinions?

Jira issue: https://jira.xwiki.org/browse/XWIKI-20290


mflorea (approved):
I think it makes more sense to set the response content type where we actually return / print the response, and that is in ``rendercontent.vm``. ``contentvars.vm`` just computes the rendered content but the result could be used for something else (not just for printing on the response).

PR #2083, opened 2023-02-10
XWIKI-17325  The page title field cannot be made mandatory at space level

acotiuga:
* we should check first if the current space is expecting the title to be mandatory
* using #getSpacePreference(String) will eventually fallback on a #getXWikiPreference() method


vmassol:
Sounds good to me. 

Notes:
* We'll need to ask ourselves if we want to have this property honored at the doc level too or if space is enough. FTM this seems enough until someone needs it at the page level... 
* Ideally we would need a test somewhere to prove that the fix works and will continue to work in the future. But I don't want to block the PR for this.

Thanks Alex.

PR #2084, opened 2023-02-10
XWIKI-12870  Empty Select on property of Database List linked with an integer property

acotiuga:
* allow the id and value fields to point to a property that can be stored in any table, not only in the StringProperty one

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/objects/classes/ImplicitlyAllowedValuesDBListQueryBuilder.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/objects/classes/ImplicitlyAllowedValuesDBListQueryBuilder.java
@@ -26,9 +26,12 @@ import java.util.Map;
 
 import javax.inject.Inject;
 import javax.inject.Named;
+import javax.inject.Provider;
 import javax.inject.Singleton;
 
+import org.slf4j.Logger;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.model.reference.DocumentReferenceResolver;
 import org.xwiki.query.Query;
 import org.xwiki.query.QueryBuilder;
 import org.xwiki.query.QueryException;
@@ -36,8 +39,13 @@ import org.xwiki.query.QueryFilter;
 import org.xwiki.query.QueryManager;
 import org.xwiki.text.StringUtils;
 
+import com.xpn.xwiki.XWiki;
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.objects.classes.BaseClass;
 import com.xpn.xwiki.objects.classes.DBListClass;
 import com.xpn.xwiki.objects.classes.DBTreeListClass;
+import com.xpn.xwiki.objects.classes.PropertyClass;
 
 /**
  * Builds a query from the meta data of a Database List property.
@@ -80,6 +88,16 @@ public class ImplicitlyAllowedValuesDBListQueryBuilder implements QueryBuilder<D
     private static final String TEMPLATE_NAME = "templateName";
 
     private static final String COLUMN_SEPARATOR = ", ";
+
+    @Inject
+    @Named("current")
+    private DocumentReferenceResolver<String> documentReferenceResolver;
+
+    @Inject
+    private Logger logger;
+
+    @Inject
+    private Provider<XWikiContext> contextProvider;
 
     @Inject
     private QueryManager queryManager;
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/objects/classes/ImplicitlyAllowedValuesDBListQueryBuilder.java
```diff
@@ -156,8 +174,17 @@ private void addFieldToQuery(String fieldName, String fieldAlias, boolean hasCla
         } else if (!hasClassName) {
             selectClause.add(DOC_PREFIX + fieldName);
         } else {
+            PropertyClass pc = null;
+            try {
+                pc = (PropertyClass) contextProvider.get().getWiki()
+                    .getXClass(documentReferenceResolver.resolve((String) parameters.get(CLASS_NAME)),
```
mflorea, line 180: What is the "current" document in this case? The class reference is probably local, so the wiki part needs to be added. I see above:

    query.setWiki(spec.wiki);

and

    spec.wiki = dbListClass.getOwnerDocument().getDocumentReference().getWikiReference().getName();

So the query is not executed on the current wiki but on the wiki of the xclass that has the DBList property. Maybe you can replace the ``boolean hasClassName`` parameter with the full spec ``DBListQuerySpec`` in order to have quick access to ``hasClassName``, ``className`` and ``wiki``.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/objects/classes/ImplicitlyAllowedValuesDBListQueryBuilder.java
```diff
@@ -156,8 +174,17 @@ private void addFieldToQuery(String fieldName, String fieldAlias, boolean hasCla
...
+                        contextProvider.get()).get(fieldName);
+            } catch (XWikiException e) {
+                logger.error("Failed to get property [{}] for class [{}]", fieldName, parameters.get(CLASS_NAME), e);
```
mflorea, line 183: If you can continue after such an exception (I'm not sure you can) then you should log an warning message instead, with the root cause message. Otherwise you should simply throw an exception. See also https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices .

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/internal/objects/classes/ImplicitlyAllowedValuesDBListQueryBuilderTest.java
```diff
@@ -301,12 +329,34 @@ public void buildWithParent() throws Exception
...
+        parameters.put("className", "XWiki.XWikiUsers");
+
+        documentReferenceResolver = mocker.getInstance(DocumentReferenceResolver.TYPE_STRING, "current");
+        DocumentReference usersClassReference = new DocumentReference("wiki", "XWiki", "XWikiUsers");
+        when(documentReferenceResolver.resolve((String) parameters.get("className"))).thenReturn(usersClassReference);
+
+        usersClass = mock(BaseClass.class);
+        when(xwiki.getXClass(usersClassReference, xcontext)).thenReturn(usersClass);
+        propertyClass = mock(PropertyClass.class);
+        when(usersClass.get("")).thenReturn(propertyClass);
```
mflorea, line 360: The empty field name looks strange.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/internal/objects/classes/ImplicitlyAllowedValuesDBListQueryBuilderTest.java
```diff
@@ -301,12 +329,34 @@ public void buildWithParent() throws Exception
...
+        this.dbListClass.setClassname("XWiki.XWikiUsers");
         assertQuery("select distinct doc.fullName as unfilterable0, idProp.value, valueProp.value, parentProp.value"
             + " from XWikiDocument as doc, BaseObject as obj, StringProperty as idProp, StringProperty as valueProp,"
             + " StringProperty as parentProp where doc.fullName = obj.name and obj.className = :className and"
             + " doc.fullName <> :templateName and obj.id = idProp.id.id and idProp.id.name = :idProp and"
             + " obj.id = valueProp.id.id and valueProp.id.name = :valueProp and obj.id = parentProp.id.id and"
             + " parentProp.id.name = :parentProp");
     }
+
+    private void setPropertyType() throws Exception
```
mflorea, line 341: You made the current tests pass, but you haven't added a test for the new use case that you added support for. Maybe this method could have some parameters to control the property type (``StringProperty`` vs. say ``IntegerProperty``).

mflorea (changes requested):

PR #2092, opened 2023-02-21
XWIKI-20541: Integrate a WCAG 2.1 validation tool in XWiki build/CI 

Sereza7:
**Jira:**  https://jira.xwiki.org/browse/XWIKI-20541
**PR Changes:**

- Integrated Axe core to the XWiki build and in the CI pipeline
- Removed the old Dutch Web Guidelines tests

Design proposal here: https://design.xwiki.org/xwiki/bin/view/Proposal/WCAG21validationinbuild
Forum topic here: https://forum.xwiki.org/t/wcag-2-1-validation-proposal/11792
Example of an accessibility warning report generated while building xwiki-platform-menu-test-docker with the parameter xwiki.test.ui.wcag set to true:
[wcagWarnings.txt](https://github.com/xwiki/xwiki-platform/files/10795333/wcagWarnings.txt)


xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/pom.xml
```diff
@@ -156,6 +156,12 @@
       <groupId>commons-httpclient</groupId>
       <artifactId>commons-httpclient</artifactId>
     </dependency>
+    <!-- Used by the wcag validation-->
+    <dependency>
```
tmortagne, line 160: This dependency is already coming with xwiki-platform-test-ui AFAICS.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -26,24 +26,31 @@
...
 import org.xwiki.test.ui.po.editor.WikiEditPage;
 
+import com.deque.html.axecore.results.Results;
+import com.deque.html.axecore.selenium.AxeBuilder;
+
 /**
  * Represents the common actions possible on all Pages.
  * 
- * @version $Id$
+Added  * @version $Id$
```
tmortagne, line 53: Looks like a mistake.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/pom.xml
```diff
@@ -64,13 +64,19 @@
     <dependency>
       <groupId>io.github.bonigarcia</groupId>
       <artifactId>webdrivermanager</artifactId>
-      <version>5.1.1</version>
+      <version>5.3.1</version>
```
vmassol, line 67: was this needed or is it something separate?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -118,6 +125,38 @@ public BasePage()
...
+                    wcagContext.addWcagTime(deltaTime);
+                }
+            }catch (Exception e) {
+                LOGGER.error("Exception when calling axe core validation on the BasePage.");
+                e.printStackTrace();
+            }
+        }
+        else{
+            LOGGER.info("[{} : {}] This combination of URL:class was already WCAG checked.", this.getPageURL(), this.getClass());
+        }
```
vmassol, line 159: should be put in a separate method to be more clear about what it's about.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -118,6 +125,38 @@ public BasePage()
     {
         super();
         waitUntilPageIsReady();
+
+        Logger LOGGER = LoggerFactory.getLogger(BasePage.class);
```
vmassol, line 129: only needed if WCAG validation is enabled.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -118,6 +125,38 @@ public BasePage()
...
+        WCAGContext wcagContext = getUtil().getWcagContext();
+        if (!wcagContext.isWcagEnabled()) {
+            return;
+        }
+        if (wcagContext.checkAccessibility(this.getPageURL(), this.getClass())) {
+            /* Perform accessibility checks on the element when instanciated if the wcag is activated in build. */
+            long startTime = System.currentTimeMillis();
+            XWikiWebDriver driver = this.getDriver();
+            try {
+                if (driver == null) {
```
vmassol, line 139: can this happen?

vmassol:
Thanks @Sereza7 . Does it work? :) Could you show some results of the build when executing it?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/pom.xml
Sereza7: This dependency is used in the `com.deque.html.axe-core` package too, and leaving the version to 5.1.1 fails the build.

The build logs without this version change:

>  [INFO] --- maven-enforcer-plugin:1.4.1:enforce (enforce-upper-bounds) @ xwiki-platform-test-ui ---
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for io.github.bonigarcia:webdrivermanager:5.1.1 paths to dependency are:
+-org.xwiki.platform:xwiki-platform-test-ui:15.1-SNAPSHOT
  +-io.github.bonigarcia:webdrivermanager:5.1.1
and
+-org.xwiki.platform:xwiki-platform-test-ui:15.1-SNAPSHOT
  +-com.deque.html.axe-core:selenium:4.6.0
    +-io.github.bonigarcia:webdrivermanager:5.3.1
]

Sereza7:
> Thanks @Sereza7 . Does it work? :) Could you show some results of the build when executing it?

Example of an accessibility warning report generated while building xwiki-platform-menu-test-docker with the parameter xwiki.test.ui.wcag set to true:

`mvn clean verify -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false -Pintegration-tests -Dxwiki.test.ui.wcag=true > wcagAccessibilityLogs.log`


[wcagAccessibilityLogs.log](https://github.com/xwiki/xwiki-platform/files/10801862/wcagAccessibilityLogs.log)
[wcagWarnings.txt](https://github.com/xwiki/xwiki-platform/files/10795333/wcagWarnings.txt)

vmassol:
> Example of an accessibility warning report generated while building xwiki-platform-menu-test-docker with the parameter xwiki.test.ui.wcag set to true:

Sounds good.

I think we should log in the console only when `xwiki.test.debug` is `true` for:
* `10:06:09.981 [main] INFO  org.xwiki.test.ui.po.BasePage - [http://host.testcontainers.internal:8080/xwiki/bin/view/Applications/WebHome : class org.xwiki.test.ui.po.ViewPage] Found [2] WCAG violations.``
* `10:06:22.061 [main] INFO  org.xwiki.test.ui.po.BasePage - [http://host.testcontainers.internal:8080/xwiki/bin/view/Applications/WebHome : class org.xwiki.test.ui.po.ViewPage] This combination of URL:class was already WCAG checked.`

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
```diff
@@ -0,0 +1,295 @@
...
+    // WCAG 2.2 Level A & AA Rules
+    entry("target-size", false)
+  );
+  /**
+   *
+   */
+  private static final List<String> VALIDATE_TAGS = Arrays.asList("wcag2a", "wcag2aa", "wcag21a", "wcag21aa");
+  private static final List<String> DISABLED_RULES = Arrays.asList();
+
+  private static class WCAGTestResults {
```
michitux, line 131: A JavaDoc comment that explains the purpose of this class would be nice.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
```diff
@@ -0,0 +1,295 @@
...
+        this.warnReport = AxeReporter.getAxeResultString();
+      }
+    }
+    public String getFailReport() {
+      return this.failReport;
+    }
+    public String getWarnReport() {
+      return this.warnReport;
+    }
+    protected boolean failIsEmpty() {
```
michitux, line 164: I think it is a good practice to have the verb first, i.e., `isFailsEmpty` (but not sure, this is really just a feeling).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
```diff
@@ -0,0 +1,295 @@
...
+      return (this.failReport.equals(""));
```
michitux, line 165: The `()` around the return value aren't necessary and should be omitted as there is no good reason to have them.

michitux (changes requested):
From the many formatting problems I spotted here I assume that you're either not using an IDE or you haven't properly configured your IDE to follow XWiki code style. We're currently not enforcing code style using CheckStyle on test code (which is a mistake imho) but we should still follow the rules as much as possible so please make sure you're using a properly configured IDE.

Overall, this looks okay, but for me it is not clear how the enabled (or disabled?) rules were chosen. 

Also, from what I understood, this doesn't seem to fail any test so we won't notice the failures unless we look at the results. With this approach, it seems hard to follow the progress we make and to catch regressions. What's the plan how we see the results? Will they be available on CI?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
Sereza7: I shouldn't happen, since the getDriver returns a driver that's always initialized during construction of PersistentTestContext. The instantiation of the driver happens either in the PersistentTestContext constructor or beforehand.
The driver being null would surely make the test framework fail before reaching this point anyways.

I'm removing this clause for the next commit :+1:

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/AllTests.java
```diff
@@ -53,9 +53,13 @@ public static Test suite() throws Exception
         addTest(suite, DefaultValidationTest.suite(DefaultValidationTest.class, html5Validator),
             DefaultValidationTest.class);
 
+        /*
+        Removed Dutch Guidelines validation tests for https://jira.xwiki.org/browse/XWIKI-20541 .
+
         CustomDutchWebGuidelinesValidator dwgValidator = new CustomDutchWebGuidelinesValidator();
         addTest(suite, CustomDutchWebGuidelinesValidationTest.suite(CustomDutchWebGuidelinesValidationTest.class,
             dwgValidator), DefaultValidationTest.class);
+        */
```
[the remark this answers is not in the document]
Sereza7: This is the last part of this issue. 
https://jira.xwiki.org/browse/XWIKI-20541
The Dutch Web Guidelines Validator was used to check for a set of outdated accessibility standards based on WCAG 2.0 .
I'm removing the concerned files appropriately in the upcoming commit.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
```diff
@@ -1427,6 +1432,46 @@ public String getSecretToken()
...
+    {
+        return this.wcagContext;
+    }
+
+    public void writeWCAGResults(Logger logger) throws IOException{
+        WCAGContext wcagContext = getWcagContext();
+        if (wcagContext.isWcagEnabled()) {
+            float totalTime = (float) wcagContext.wcagTimer / 1000;
+            logger.info("Time spent on WCAG validation [{}] (in s)", totalTime);
+            File wcagDir = new File("target/wcag");
```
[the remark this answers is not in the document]
Sereza7: For next commit, I set this value as a static final variable in TestUtils (and built it using `File.separator` instead of just a `/` ).
This should be better but I'm not sure this solves this issue completely, I'm not sure of what the best practice here would be.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
```diff
@@ -0,0 +1,295 @@
...
+  /** List of rules to fail on.
+   * Failing rules will fail the test suite, non failing rules will only return a warning if violated.
+   * Those rules are identified by ID, and a full list of ids can be found here:
+   * <a href="https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md#deprecated-rules">...</a>
+   * Only rules for validating up to WCAG level AA 2.1 are specified here.
+   * Rules not defined here will keep their default behavior and fail the test-suite on violation.
+   * There is a parameter for rules *reviewOnFail* in axe-core, however it's not possible to change it using the
+   * current selenium integration library.
+   **/
+  private static final Map<String, Boolean> FAILS_ON_RULE = Map.<String, Boolean>ofEntries(
```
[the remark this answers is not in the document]
Sereza7: This was highlighted by Intellij as a code quality warning:
`Vararg method call with 50+ poly arguments may cause compilation and analysis slowdown `

I'm omitting it for the next commit :+1:

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
```diff
@@ -0,0 +1,295 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static java.util.Map.entry;
+
+
+public class WCAGContext{
+
+  /** List of rules to fail on.
```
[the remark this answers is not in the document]
Sereza7: I needed to update this doc, I'm adding precisions to try and make those points clearer :ok_hand:

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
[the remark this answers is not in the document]
michitux: I've just had a look and found https://github.com/xwiki/xwiki-platform/blob/e44c811a38d395a58722110e914ee8a1e6dfb00c/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/internal/junit5/DockerTestUtils.java#L273 and https://github.com/xwiki/xwiki-platform/blob/a25ac4c3dfacd781d6ed1a86d40ca517a7da26e6/xwiki-platform-core/pom.xml#L340 used at https://github.com/xwiki/xwiki-platform/blob/8793d8438ebfb7cfe40d9b4ce31d9f791f563187/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestDebugger.java#L63 - I think you could/should do something similar for the WCAG results.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
```diff
@@ -0,0 +1,295 @@
...
+      axeBuilder.disableRules(DISABLED_RULES);
+    }
+    return axeBuilder;
+  }
+
+  /**
+   * Checks if there's a need to perform an accessibility analysis of a basePage.
+   *
+   */
+  public boolean checkAccessibility(String url, Class pageClass) {
```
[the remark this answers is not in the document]
Sereza7: I renamed it `isAlreadyInCache` and changed a bit it's behavior, thanks for the tip, it was a really bad name.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -118,6 +125,38 @@ public BasePage()
...
+                    throw new NullPointerException("The provided webDriver is null.");
```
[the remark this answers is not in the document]
Sereza7: https://github.com/xwiki/xwiki-platform/pull/2092#discussion_r1114285333
I removed this throw :+1:

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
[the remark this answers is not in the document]
michitux: Oh, okay, I didn't know that one, I'm normally just seeing the warning the other way around. You can also keep it and just add a comment that explains why it is there.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WCAGContext.java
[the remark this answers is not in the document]
vmassol: Note that `isCached()` is also simple and effective ;)

xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards/src/test/it/org/xwiki/test/webstandards/AllTests.java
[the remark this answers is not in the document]
Sereza7: Okay, I'll leave them for now and try to focus making sure the properties they cover are made failing in the new test asap. This should be easy but some rules might not match exactly and catch violations that didn't get caught before, so it's not just a matter of switching the FAILS_ON_RULE entries to true.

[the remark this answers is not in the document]
michitux: I'm not saying that the new tests need to be equivalent, it's just that we actually have failures of these validation tests from time to time (e.g., they found an input without a label recently) and I think we shouldn't loose this basic level of checks as long as the framework isn't actually active in a way that we notice (i.e., that lead to actual failures that are visible in the CI).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
```diff
@@ -1427,6 +1432,46 @@ public String getSecretToken()
...
+                if (!wcagDir.exists()) {
+                    Files.createDirectory(wcagDir.toPath());
+                }
+                String outputName = "wcagFails";
+                File failsFile = new File(wcagDir, outputName);
+                WCAGContext.writeWCAGReportToFile(failsFile, wcagContext.buildFailsReport());
+            }
+            // Assert the results of the different WCAG Checks are all empty
+            assertFalse(wcagContext.hasWCAGFails(), "There are some accessibility issues in the test suite." +
+              "See target/wcag/wcagFails for more details.");
```
michitux, line 1471: To me it seems unexpected that the code to write the results contains the assertion that fails the build. Also, wouldn't it be nicer to include at least the description of the first failure in the assertion such that it is displayed in Jenkins and whoever sees this doesn't always need to look at the output files?

Sereza7:
**PR Changes:**
* Fixed checkstyle requirements.
* New class XWikiCustomAxeReporter to contain most of the string building for the report, and allow generating a report more adapted to our use case.
* Changed log info to debug.
* Removed hardcoded file location.
* Added back the DutchWebGuidelines (to keep until some fails are triggerable)
* Added the first rule failed details to the error message when asserting there's no failing violation.

Example of an accessibility warning report generated while building xwiki-platform-menu-test-docker with the parameter xwiki.test.ui.wcag set to true. For the sake of variety, one rule has been set as failing:
WcagContext.FAILS_ON_RULE -> entry("list", true)

mvn clean verify -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false -Pintegration-tests -Dxwiki.test.ui.wcag=true > wcagAccessibilityLogs.log

[wcagAccessibilityLogs.log](https://github.com/xwiki/xwiki-platform/files/10815444/wcagAccessibilityLogs.log)
[wcagFails.txt](https://github.com/xwiki/xwiki-platform/files/10815447/wcagFails.txt)
[wcagWarnings.txt](https://github.com/xwiki/xwiki-platform/files/10815448/wcagWarnings.txt)

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Stores a human-readable summary of the violations contained in scanResults in axeResultString.
+     *
+     * @param testMethodName from the current test context
+     * @param pageClassName  name of the class of the page to analyze
+     * @param url            of the page to analyze
+     * @param scanResults    the list of rules breached during validation.
+     * @return true if there is any rule breached in scanResults
+     */
+    static boolean getReadableAxeResults(String testMethodName, String pageClassName,
```
michitux, line 81: From what I can see, the result is never used when returning `false` so I wonder why this method doesn't simply return a `String` (or `Optional<String>`) instead of this strange design with a static result string.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,413 @@
...
+                the default behavior will be to add it to the fails.
+                In order to resolve these test-suite fails quickly, set them as "false" in FAILS_ON_RULE.
+                 */
+                .collect(Collectors.toList());
+            if (XWikiCustomAxeReporter.getReadableAxeResults(testMethodName, pageClassName, url, failingViolations))
+            {
+                this.failReport = XWikiCustomAxeReporter.getAxeResultString();
+                if (shortFailReport.isEmpty())
+                {
+                    /* Only store the first failed rule to display in the error logs. */
```
michitux, line 172: I know this was my suggestion, we could imho also just display all failed rules if this requires extra work to get the first result. My intention was primarily to display something informative when a test fails such that the developer doesn't first need to get the log to see the reason for the failure. My understanding is that we will only enable rules for which we don't have known failures so in this case this also shouldn't flood the logs unless a regression is introduced (or a new test is added, but then it should be fixed immediately).

michitux (changes requested):
Some new comments, but this looks already much better.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
```diff
@@ -0,0 +1,120 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+/*
+ * Alternatively, at your choice, the contents of this file may be used under the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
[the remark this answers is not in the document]
Sereza7: Should I just copy the copyright notice from here and add a link to this file?
https://github.com/dequelabs/axe-core-maven-html/blob/develop/selenium/src/main/java/com/deque/html/axecore/selenium/AxeReporter.java

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
Sereza7: It can return false, if the report asked for is "empty".
I however still set up conditions against empty reports so I'll remove this structure and make sure it still runs as expected.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
Sereza7: The thing is, for less memory use, I process reports as soon as they are generated so I only keep the string we want. This is less flexible than keeping everything in memory and waiting for the end to generate reports.
This makes it a bit weird to save only the first rule break, since it's often only a part of one validation check.
It'd be easier to display the first report fully, however it could get quite lengthy if a few rules fail at once.
This shouldn't happen too often, so I'll remove this static variable and go with the easy solution

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
michitux: Yes, I understand that, but in this case we don't use the string it built so instead of returning a boolean, it could just return the string and when the report is "empty" it could return either an empty string, `null` or an empty `Optional<String>` if you want to be more explicit. That way, it wouldn't need this really ugly structure with the static variables (which are completely incompatible with multi-threading for example - I know this is not relevant here, but still, it's just not nice).

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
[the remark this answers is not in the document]
michitux: Yes, I think you should copy the copyright notice and include a reference to the source. I think in the end, in particular with the refactoring I suggested, the similarity to the original won't be that big and I think there is not much creativity and thus not much to copyright in this code. But still, it is good to properly attribute the source.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
michitux: Sounds good, we can change it also later if we should discover that it is indeed flooding the log.

Sereza7:
**PR Changes:**
* Removed useless complexity from XWikiCustomAxeReporter
* Fixed licencing on XWikiCustomAxeReporter
* Changed the error message to display more (simpler/cleaner implementation)
* Fixed @since on files
* Fixed logging for basePage, now following best practice with a static logger.

**Example:**
Accessibility warning report generated while building xwiki-platform-menu-test-docker with the parameter xwiki.test.ui.wcag set to true. For the sake of variety, one rule has been set as failing:
WcagContext.FAILS_ON_RULE -> entry("list", true)

mvn clean verify -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false -Pintegration-tests -Dxwiki.test.ui.wcag=true > wcagAccessibilityLogs.log

[wcagAccessibilityLogs.log](https://github.com/xwiki/xwiki-platform/files/10822793/wcagAccessibilityLogs.log)

[wcagFails.txt](https://github.com/xwiki/xwiki-platform/files/10822796/wcagFails.txt)
[wcagWarnings.txt](https://github.com/xwiki/xwiki-platform/files/10822797/wcagWarnings.txt)

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-docker/src/main/java/org/xwiki/test/docker/junit5/TestConfiguration.java
```diff
@@ -484,6 +492,22 @@ public void setVNC(boolean vnc)
         this.vnc = vnc;
     }
 
+    /**
+     * @return true if WCAG rules should be checked.
+     */
+    public boolean isWcagEnabled()
```
vmassol, line 498: Missing `@since` annotation

vmassol: Small preference to use WCAG and not `Wcag` since it's an acronym. So either `isWCAG()` or `isWCAGEnabled()`. See also `getExtraJARs()`, `getJDBCDriverVersion()`, etc.

vmassol: note that `isWCAG()` might be more consistent with the other `isVerbose()`, `isOffice()`, etc.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
```diff
@@ -214,6 +219,8 @@
     /** Cached secret token. TODO cache for each user. */
     private String secretToken = null;
 
+    /** Cached accessibility results. */
```
vmassol, line 222: We use multiline format in general.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
```diff
@@ -1427,6 +1434,73 @@ public String getSecretToken()
         return this.secretToken;
     }
 
+    /**
+     * Get the WCAG context for the test suite.
+     *
+     * @return WCAG Context
+     */
+    public WcagContext getWcagContext()
```
vmassol, line 1442: `WCAGContext`

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
```diff
@@ -1427,6 +1434,73 @@ public String getSecretToken()
...
+    {
+        return this.wcagContext;
+    }
+
+    /**
+     * Writes the wcag validation results to custom reports.
+     * @throws IOException can be thrown when the directory creation fails.
+     */
+
```
vmassol, line 1451: extra new line

vmassol: missing `@since`

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/TestUtils.java
```diff
@@ -1427,6 +1434,73 @@ public String getSecretToken()
...
+        String testClassesDirectory;
+        String mavenBuildDir = System.getProperty("maven.build.dir");
+        if (mavenBuildDir == null) {
+            testClassesDirectory = "target/wcag-reports";
+        } else {
+            testClassesDirectory = String.format("%s/wcag-reports", mavenBuildDir);
+        }
+        return testClassesDirectory;
+    }
+
```
vmassol, line 1503: I'd move everything related to WCAG inside another class.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+    private boolean wcagEnabled = true;
+    private long wcagTimer;
+    private final Map<String, ArrayList<String> > wcagValidationCache = new HashMap<>();
+    private String testClassName;
+    private String testMethodName;
+
+    /**
+     * @return the first non-empty fail report from all the validations.
+     */
+    public String getFirstWcagFail()
```
vmassol, line 214: `getFirstWCAGFailure()`? The javadoc doesn't say what it does if there's no failure. Not sure that returning an empty string is great. I'd prefer null.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+    {
+        String report = "";
+        for (WCAGTestResults wcagResult : wcagResults)
+        {
+            if (!wcagResult.isFailEmpty())
+            {
```
vmassol, line 220: wrong code style

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+                report = wcagResult.getFailReport();
+                break;
+            }
+        }
+        return report;
+    }
+
+    /**
+     * Sets the current test class name.
+     * @param testClassName new value to be set.
```
vmassol, line 230: missing `the`. See code style at https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HJavadocBestPractices

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+     */
+    public void setTestClassName(String testClassName)
+    {
+        this.testClassName = testClassName;
+    }
+    protected String getTestClassName()
```
vmassol, line 236: new line missing.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+     */
+    public static void writeWCAGReportToFile(final File outputFile, final Object output)
+    {
+        try (Writer writer =
+                     new BufferedWriter(
+                             new OutputStreamWriter(
+                                     new FileOutputStream(outputFile), StandardCharsets.UTF_8))) {
+            writer.write(output.toString());
+        } catch (IOException e) {
+            e.printStackTrace();
```
vmassol, line 409: This looks wrong

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+     * Writes a WCAG report to a file for proper accessibility warnings and fails examination.
+     * @param outputFile file to write the object to.
+     * @param output object to keep in memory.
+     */
+    public static void writeWCAGReportToFile(final File outputFile, final Object output)
+    {
+        try (Writer writer =
+                     new BufferedWriter(
+                             new OutputStreamWriter(
+                                     new FileOutputStream(outputFile), StandardCharsets.UTF_8))) {
```
vmassol, line 406: bad code style

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/WcagContext.java
```diff
@@ -0,0 +1,412 @@
...
+     * @return a readable report of wcag violations (non failing) found so far.
+     */
+    public String buildWarningsReport()
+    {
+        StringBuilder mergedReport = new StringBuilder();
+        boolean warningsEmpty = true;
+        for (WCAGTestResults result : this.wcagResults) {
+            if (!result.isWarnEmpty()) {
+                if (warningsEmpty) {
+                    mergedReport.append(String.format("WCAG warnings in the test class %s:", getTestClassName()));
```
vmassol, line 385: Not using our logging best practice of using `[...]`. See https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
```diff
@@ -0,0 +1,128 @@
...
+ * distribute or in any file that contains substantial portions of this source
+ * code.
+ */
+/*
+    Source:
+    com.deque.html.axecore.selenium.AxeReporter
+    https://github.com/dequelabs/axe-core-maven-html/blob/develop/selenium/src/main/java/com/deque/html/axecore/selenium
+    /AxeReporter.java
+
+ */
```
vmassol, line 37: Weird comment. What does it mean?

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Creates a human-readable report from an axe-core scan result.
+ * @since 15.2RC1
+ * @version $Id$
+ */
+public final class XWikiCustomAxeReporter
+{
+    private XWikiCustomAxeReporter()
+    {
```
vmassol, line 53: not passing checkstykle normally. Should have a comment to explain why it's empty.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -118,6 +126,10 @@ public BasePage()
     {
         super();
         waitUntilPageIsReady();
+
+        WcagContext wcagContext = getUtil().getWcagContext();
```
vmassol, line 130: missing comment and should have a check, there'esno need to gather wcag context if wcag is not enabled

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -693,6 +705,60 @@ private void useShortcutKeyAndLoads(CharSequence... shortcut)
...
+     */
+    protected void validateWCAG(WcagContext wcagContext, boolean checkCache)
+    {
+        if (!wcagContext.isWcagEnabled())
+        {
+            /*
+            Block wcag validation if it is not enabled, in all cases.
+             */
+            return;
+        }
```
vmassol, line 732: to be moved up there

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
```diff
@@ -693,6 +705,60 @@ private void useShortcutKeyAndLoads(CharSequence... shortcut)
...
+
+        long startTime = System.currentTimeMillis();
+        if (!checkCache || wcagContext.isNotCached(this.getPageURL(), this.getClass().getName())) {
+            /*
+            Performs accessibility validation on the element when instanced if
+            the page is not in the WCAGContext cache.
+             */
```
vmassol, line 739: should use `//`comment-style

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/XWikiCustomAxeReporter.java
Sereza7: This class was heavily inspired from the axe-core native final class, it's a way to quote the source. Now I'm not sure we have that much from the original source anymore though. 
I'm making it more explicit for the next commit :+1:

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
Sereza7: On this basePage, from what I know, we need to check wcagContext to see if wcag is enabled.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java
Sereza7: https://github.com/xwiki/xwiki-platform/pull/2092#discussion_r1115895145
I agreed with @michitux 's opinion, I think it's more important to include the check in the function for explicit validation calls in the UITests. 
However I did not give a wide enough scope (were set as `protected` instead of `public`).

Sereza7:
**PR Changes:**
* Corrected some checkstyle issues.
* Added/ fixed incomplete documentation
* Fixed presentation of the report.
* Moved most of the methods from TestUtils to WCAGUTils.

**Example:**
Accessibility warning report generated while building xwiki-platform-menu-test-docker with the parameterxwiki.test.ui.wcagset to true. For the sake of variety, one rule has been set as failing:
WcagContext.FAILS_ON_RULE -> entry("image-alt", true)

>    mvn clean verify -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false -Pintegration-tests -Dxwiki.test.ui.wcag=true > wcagAccessibility.log

[wcagAccessibility.log](https://github.com/xwiki/xwiki-platform/files/10882896/wcagAccessibility.log)

[wcagFails.txt](https://github.com/xwiki/xwiki-platform/files/10882897/wcagFails.txt)
[wcagWarnings.txt](https://github.com/xwiki/xwiki-platform/files/10882898/wcagWarnings.txt)

PR #2094, opened 2023-02-23
[Misc] Fix checkstyle issues in xwiki-platform-test-ui module

surli:
:warning: Work In Progress

The idea of this PR is to clean up all checkstyle (or at least the easy ones) we have in test-ui module. 
Note that in some cases it means refactoring a bit some page objects since it's cleaner to fix some issues like that. 

Full list of issues is available on this pad: https://cryptpad.fr/code/#/2/code/view/qwo61L8FTjvbVMyDDuDdmzNv1HQqt3Sw9Q2YkQmdG0c/

I'm removing the issues I believe are fixed in the pad whenever I'm working on a file, but I didn't re-built the module yet as I'm expecting new ones to appear times to finish the work. So I'll refresh the pad only when all already present are done. 

Opening this PR so people can review what I'm doing all along the process: I'm expecting this to last a while, especially since I'm working on it when I have not enough brain power for more intensive tasks.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
+++ b/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
@@ -39,6 +39,8 @@ public class CommentForm extends BaseElement
      */
     private By containerLocator;
 
+    private final By textareaLocator = By.tagName("textarea");
+
     /**
      * Creates a new form instance.
      * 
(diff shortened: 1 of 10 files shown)


xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentForm.java
```diff
@@ -39,6 +39,8 @@ public class CommentForm extends BaseElement
      */
     private By containerLocator;
 
+    private final By textareaLocator = By.tagName("textarea");
```
tmortagne, line 42: I'm wondering if it would not make more sense to make this a static since it's always the same thing from what I understand, as any other cons variable.

PR #2098, opened 2023-03-01
XWIKI-19450: Several nav tags are used without aria-label

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-19450
**PR Changes:**
* Added labels to the <nav> elements in the XWiki interface
* Added labels to the role="navigation" elements in the XWiki interface
(*Added role="navigation" to the vertical custom menus)
* Added translation entries for these labels

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
@@ -77,6 +77,8 @@ livedata.selection.infoBar.selectedCount={0} entries selected
 livedata.selection.infoBar.allSelected=All entries selected
 livedata.selection.infoBar.allSelectedBut=All entries selected but {0}
 
+livedata.pagination.label=Data page details
+livedata.pagination.index.label=Data page selection
 livedata.pagination.currentEntries=Entries {0} - {1} out of {2}
 livedata.pagination.pageSize=per page of
 livedata.pagination.page=page
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
@@ -72,12 +72,13 @@
 
     <!--
       The actual pagination widget
-      It dislays the the available pages numbers, and change to them on click.
+      It displays the the available pages numbers, and change to them on click.
       Not all page numbers are show depending of the `pagination.maxShownPages`
       property in the Livedata meta config.
       Arrows can be shown to go to first, last, previous, next page.
     -->
-    <nav class="pagination-indexes">
+    <nav class="pagination-indexes"
+      :aria-label="$t('livedata.pagination.index.label')">
       {{ $t('livedata.pagination.page') }}
       <!--
         Go to First Page button
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
@@ -28,11 +28,12 @@
 -->
 <template>
   <!-- Pagination -->
-  <nav class="livedata-pagination">
+  <nav class="livedata-pagination"
+    :aria-label="$t('livedata.pagination.label')">
 
     <!--
       Display the pagination current entry range
-      Can be shown / hiden by the `pagination.showEntryRange` property
+      Can be shown / hidden by the `pagination.showEntryRange` property
       in the Livedata meta config
     -->
     <span
--- a/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
+++ b/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
@@ -1171,7 +1171,7 @@ $menuTemplateDoc.content.replace('{', '')
   #if ($stringtool.isBlank("$!id"))
     #set ($id = $wikimacro.context.getXDOM().getIdGenerator().generateUniqueId("M", "GeneratedMenuId"))
   #end
-  (% role="navigation" class="menu-horizontal-toggle" %)(((
+  (% role="navigation" class="menu-horizontal-toggle" aria-label="$services.localization.render('rendering.macro.menu.horizontal.label')" %)(((
     (% class="navbar-header" %)(((
       {{html}}
         &lt;button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#$!{escapetool.xml($id)}" aria-expanded="false"&gt;
--- a/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
+++ b/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
@@ -1187,7 +1187,7 @@ $menuTemplateDoc.content.replace('{', '')
     )))
   )))
 #else
-  (% #if ("$!id" != '') id="${services.rendering.escape($id, 'xwiki/2.1')}"#end class="menu menu-${services.rendering.escape($!type, 'xwiki/2.1')}" %)(((
+  (% role="navigation" #if ("$!id" != '') id="${services.rendering.escape($id, 'xwiki/2.1')}"#end class="menu menu-${services.rendering.escape($!type, 'xwiki/2.1')}" aria-label="$services.localization.render('rendering.macro.menu.vertical.label')" %)(((
     {{wikimacrocontent/}}
   )))
 #end
--- a/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
+++ b/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
@@ -84,6 +84,9 @@ rendering.macro.menu.parameter.type.name=Type
 rendering.macro.menu.parameter.type.description=The optional menu type determines the menu appearance and behaviour. The supported values are: horizontal (default value) and vertical.
 rendering.macro.menu.content.description=Define the menu structure using wiki syntax. Each menu item should be a list item and should contain the menu item label or link. You can use nested lists for sub-menu items.
 
+# Accessibility labels
+rendering.macro.menu.horizontal.label=Horizontal custom menu
+
 # Menu WebHome
 menu.description=This is a simple application that helps you create navigation menus to be displayed either horizontally as a top bar after the page header or vertically in a side panel.</content>
   <object>
--- a/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
+++ b/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
@@ -86,6 +86,7 @@ rendering.macro.menu.content.description=Define the menu structure using wiki sy
 
 # Accessibility labels
 rendering.macro.menu.horizontal.label=Horizontal custom menu
+rendering.macro.menu.horizontal.label=Vertical custom menu
 
 # Menu WebHome
 menu.description=This is a simple application that helps you create navigation menus to be displayed either horizontally as a top bar after the page header or vertically in a side panel.</content>
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/drawer.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/drawer.vm
@@ -21,7 +21,8 @@
 ##
 ## Display the drawer
 ##
-<nav class="drawer-nav" id="tmDrawer">
+<nav class="drawer-nav" id="tmDrawer"
+aria-label="$escapetool.xml($services.localization.render('core.menu.drawer.label'))">
   <ul class="drawer-menu">
     ##
     ## Drawer header (with elements concerning the current user: profile, login, logout, register, etc...)
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_macros.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_macros.vm
@@ -31,7 +31,8 @@
 ###    Toplevel Menu
 ###
 #macro(xwikitopmenustart)
-  <nav class="navbar navbar-default actionmenu">
+  <nav class="navbar navbar-default actionmenu"
+  aria-label="$services.localization.render('core.menu.navbar.label')">
     <div class="container-fluid">
       ## Brand and toggle get grouped for better mobile display
       <div class="navbar-header">
@@ -163,14 +164,14 @@
 ###    Toplevel Menu entry separator
 ###
 #macro(xwikitopmenuseparator)
-  <li class="divider" role="separator"></li>
+  <li class="divider" role="separator" aria-hidden="true"></li>
 #end
 
 ###
 ###    Menu submenu separator
 ###
 #macro(submenuseparator)
-  <li class="divider" role="separator"></li>
+  <li class="divider" role="separator" aria-hidden="true"></li>
 #end
 
 ###
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
@@ -954,6 +954,7 @@ core.minoredit.hide=Hide minor edits
 ### top menu
 core.menu.main.title=General Actions:
 core.menu.content.title=Page Actions:
+core.menu.navbar.label=Top level bar
 core.menu.goto.wiki=Go to Wiki
 core.menu.goto.space=Go to Space
 core.menu.goto.page=Go to Page
@@ -977,6 +978,7 @@ core.menu.edit.class=Class
 core.menu.edit.rights=Access Rights
 core.menu.edit.currentEditor=Edit{0}
 core.menu.drawer=Drawer
+core.menu.drawer.label=Top drawer menu
 core.menu.view.source=Source
 core.menu.view.comments=Comments
 core.menu.view.attachments=Attachments


xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -1171,7 +1171,7 @@ $menuTemplateDoc.content.replace('{', '')
   #if ($stringtool.isBlank("$!id"))
     #set ($id = $wikimacro.context.getXDOM().getIdGenerator().generateUniqueId("M", "GeneratedMenuId"))
   #end
-  (% role="navigation" class="menu-horizontal-toggle" %)(((
+  (% role="navigation" class="menu-horizontal-toggle" aria-label="$services.localization.render('rendering.macro.menu.horizontal.label')" %)(((
```
michitux, line 1174: Missing escaping for XWiki syntax.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -1187,7 +1187,7 @@ $menuTemplateDoc.content.replace('{', '')
     )))
   )))
 #else
-  (% #if ("$!id" != '') id="${services.rendering.escape($id, 'xwiki/2.1')}"#end class="menu menu-${services.rendering.escape($!type, 'xwiki/2.1')}" %)(((
+  (% role="navigation" #if ("$!id" != '') id="${services.rendering.escape($id, 'xwiki/2.1')}"#end class="menu menu-${services.rendering.escape($!type, 'xwiki/2.1')}" aria-label="$services.localization.render('rendering.macro.menu.vertical.label')" %)(((
```
michitux, line 1190: Missing escaping for XWiki syntax.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
```diff
@@ -84,6 +84,10 @@ rendering.macro.menu.parameter.type.name=Type
 rendering.macro.menu.parameter.type.description=The optional menu type determines the menu appearance and behaviour. The supported values are: horizontal (default value) and vertical.
 rendering.macro.menu.content.description=Define the menu structure using wiki syntax. Each menu item should be a list item and should contain the menu item label or link. You can use nested lists for sub-menu items.
 
+# Accessibility labels
+rendering.macro.menu.horizontal.label=Horizontal custom menu
+rendering.macro.menu.horizontal.label=Vertical custom menu
```
michitux, line 89: Wouldn't it be better to just use this as a fallback and let the user provide a label to the macro? Also, when the menu macro is used twice, the labels should definitely be different.

michitux (changes requested):

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
Sereza7: I'm worried that the user might not give a name descriptive enough to the menus, so I can't just use it as is.
I could however add a parameter in those entries to contain the name of the menu itself, this would make the labels different from one menu to the other (except for vertical menus where the admin can use the panel multiple times :thinking: ) and give a bit of extra info without relying too much on admin's practices when it comes to naming custom menus.
e.g.:
`rendering.macro.menu.horizontal.label=Horizontal custom menu: {0}`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
```diff
@@ -71,12 +72,13 @@
...
       The actual pagination widget
-      It dislays the the available pages numbers, and change to them on click.
+      It displays the the available pages numbers, and change to them on click.
       Not all page numbers are show depending of the `pagination.maxShownPages`
       property in the Livedata meta config.
       Arrows can be shown to go to first, last, previous, next page.
     -->
-    <nav class="pagination-indexes">
+    <nav class="pagination-indexes"
+      :aria-label="$t('livedata.pagination.index.label')">
```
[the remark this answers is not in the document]
Sereza7: We can nest nav elements, there's no real incorrect way to use navs, as long as they describe a section with links.
> The [nav](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-nav-element.html#the-nav-element) element [represents](http://www.w3.org/TR/2011/WD-html5-20110525/rendering.html#represents) a section of a page that links to other pages or to parts within the page: a section with navigation links.

Nesting nav elements: https://stackoverflow.com/questions/6986612/in-html5-can-you-nest-nav-elements
Nav element def: https://www.w3.org/TR/2011/WD-html5-author-20110809/the-nav-element.html

___
The usage here is a bit of a stretch since the parent nav does not contain any other links. I think this would be better for users  to remove this nav.
There is some info relevant to navigation in the parent, even if not links, and removing the nested nav would just force users to navigate in the nav-related infos before accessing the links.

___
_Note_: 
As stated in the nav [definition](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-nav-element.html
), 
> Not all groups of links on a page need to be in a nav element — the element is primarily intended for sections that consist of major navigation blocks. In particular, it is common for footers to have a short list of links to various pages of a site, such as the terms of service, the home page, and a copyright page. The footer element alone is sufficient for such cases; while a nav element can be used in such cases, it is usually unnecessary.

Our use of the nav landmark here is not mandatory, however I think it's still a good idea to keep it because often livedata tables are central to the content of the page and it's nice to access their nav quickly. 
This nav landmark could be removed once we use another landmark close enough (such as a `section` for the liveData object).

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
michitux: For example, on XWiki.org I think this menu macro is used for the main menu at the top. I think it would be strange to have the label "Horizontal custom menu" on this main navigation menu, regardless if there is also another label behind it.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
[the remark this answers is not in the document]
michitux: I agree that having some nav element in the LiveData pagination is probably fine, but I think we should reduce it to one.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
Sereza7: @michitux What label do you think would be more appropriate for this kind of navigation element?
Is it the "custom" which is inappropriate because this is not customizable for the user but only for admins?
Maybe "horizontal" is too close to the implementation and should be replaced by "top" (what the user sees).
I think keeping "menu" is relevant since this element is part of the menu macro.

Some possible alternatives: 
* "Top menu bar"
* "Top navigation menu"
* "Top navigation bar"


For reference:
https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/SecondGenerationWiki/
19450ExampleMenu2
Iceberg with top menu:
19450ExampleMenu1

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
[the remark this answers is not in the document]
michitux: Another problem here is the label in combination with multiple Live Data elements on a single page. For example, consider the case of opening the attachments tab on a document that already contains a Live Data (I think there are also documents with multiple Live Data elements in the content, I just don't have any example right now).

Sereza7:
**PR Changes:**
* Added escapes
* Removed a nested nav-tag in livedata page selection
* Added individual nav labels for menus (vertical and horizontal)
* Added individual nav labels for livedata tables, based on the class fetched.


_Notes:_ 
* Individual nav labels for menus should never repeat because they are based on the XPage where the menu content is defined. These menus are made so that either vertical or horizontal can't be used twice in the interface.
* Individual nav labels **might** repeat because they are based on the class fetched for the query. If two livetables fetch the same class, their labels will be equal.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
```diff
@@ -28,11 +28,12 @@
 -->
 <template>
   <!-- Pagination -->
-  <nav class="livedata-pagination">
+  <nav class="livedata-pagination"
+    :aria-label="$t('livedata.pagination.label', [this.data.query.source.className])">
```
michitux, line 32: There is no reason why a source should have a `className`, this is specific to the live table source and also there it isn't mandatory.

Sereza7:
**PR Changes:**
* Changed the label creation parameter: from query source class to the id of the panel.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -77,6 +77,8 @@ livedata.selection.infoBar.selectedCount={0} entries selected
 livedata.selection.infoBar.allSelected=All entries selected
 livedata.selection.infoBar.allSelectedBut=All entries selected but {0}
 
+livedata.pagination.label={0} live table page info
+livedata.pagination.label.empty=Live table page info
```
michitux, line 81: Why does this talk about "Live table" and not "Live data"? Is that consistent with the rest of the UI?

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/UIExtensionSheet.xml
```diff
@@ -56,12 +56,16 @@
...
+    label=""$doc.plainTitle""}}{{include
+    reference=""${escapetool.d}uix.doc.documentReference""
+    /}}{{/menu}}
     ${escapetool.h}panelfooter()
     {{/velocity}}")
   #set ($menuBar = "{{velocity}}
     ${escapetool.h}if (${escapetool.d}xwiki.hasAccessLevel('view', ${escapetool.d}xcontext.user, ${escapetool.d}uix.doc.documentReference))
-      {{menu type=""horizontal fixedWidth"" id=""menu-horizontal-$doc.fullName.replaceAll('[\s|,|.|:|\[|\]|\[\/\]|\[\\\]|=|@|#]', '-')""}}{{include reference=""${escapetool.d}uix.doc.documentReference"" /}}{{/menu}}
+      {{menu type=""horizontal fixedWidth""
+      id=""menu-horizontal-$doc.fullName.replaceAll('[\s|,|.|:|\[|\]|\[\/\]|\[\\\]|=|@|#]', '-')"" label=""$doc.plainTitle""}}{{include reference=""${escapetool.d}uix.doc.documentReference"" /}}{{/menu}}
```
michitux, line 68: You shouldn't use `$doc.plainTitle` without escaping, this can contain arbitrary text. Also this `fullName.replaceAll` looks quite dangerous to miss stuff.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
Sereza7: That's a bad mistake, fixing it, thanks.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/UIExtensionSheet.xml
michitux: Yeah, as I've said, it's probably more or less fine, and in particular fine from a security point of view, but when you add escaping to the new stuff, I think it would be nice if you could escape this (for XWiki syntax), too.

michitux: Actually, I think I forgot to submit one comment, so what I wanted to add is that the id attribute in HTML5 allows basically everything as long as it is non-empty and no ascii whitespace, so the important part is to ensure it is properly escaped. Citing [the specification](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute):

> When specified on [HTML elements](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements), the [id](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute) attribute value must be unique amongst all the [IDs](https://dom.spec.whatwg.org/#concept-id) in the element's [tree](https://dom.spec.whatwg.org/#concept-tree) and must contain at least one character. The value must not contain any [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -1382,4 +1389,83 @@ $menuTemplateDoc.content.replace('{', '')
...
+        <size>60</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </type>
+    </class>
+    <property>
+      <defaultValue/>
+    </property>
+    <property>
+      <description>Optional menu label that will be set on the HTML element that wraps the menu. You can use this label in JavaScript code to label the menu DOM or in functional tests to assert the menu name.</description>
```
michitux, line 1459: This description shouldn't describe how this can be used in the code but what kind of value the user should set when using the menu macro. This is meant for end-users, not developers (it will be displayed, e.g., in the WYSIWYG editor when you insert the menu macro). Also, you need to add a translation value for this (can be the same as this text, it's just so translators will translate it).

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -1382,4 +1389,83 @@ $menuTemplateDoc.content.replace('{', '')
...
+    </property>
+    <property>
+      <mandatory>0</mandatory>
+    </property>
+    <property>
+      <name>label</name>
```
michitux, line 1465: To be consistent with the other parameters, I would suggest to name this "Label".

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
```diff
@@ -84,6 +84,10 @@ rendering.macro.menu.parameter.type.name=Type
 rendering.macro.menu.parameter.type.description=The optional menu type determines the menu appearance and behaviour. The supported values are: horizontal (default value) and vertical.
 rendering.macro.menu.content.description=Define the menu structure using wiki syntax. Each menu item should be a list item and should contain the menu item label or link. You can use nested lists for sub-menu items.
 
+# Accessibility labels
+rendering.macro.menu.horizontal.label=Top navigation menu: {0}
+rendering.macro.menu.vertical.label=Side navigation menu: {0}
```
michitux, line 89: I'm not sure this `rendering.macro` prefix should be used here, this is specific for the macro and parameter names and descriptions. See also https://dev.xwiki.org/xwiki/bin/view/Community/L10N/Conventions/#HKeyformat for the general naming rules.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/UIExtensionSheet.xml
```diff
@@ -56,12 +56,17 @@
   ## Prepare the extension point map.
   #set ($menuPanel = "{{velocity}}
     ${escapetool.h}panelheader(""$doc.plainTitle"")
-    {{menu type=""vertical collapsible open""}}{{include reference=""${escapetool.d}uix.doc.documentReference"" /}}{{/menu}}
+    {{menu type=""vertical collapsible open""
+    label=""$escapetool.velocity($doc.plainTitle)""}}{{include
```
michitux, line 60: Why do you use `$escapetool.velocity` here? While this might indeed be necessary, this also needs escaping for XWiki syntax.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/UIExtensionSheet.xml
```diff
@@ -56,12 +56,17 @@
...
+    reference=""${escapetool.d}uix.doc.documentReference""
+    /}}{{/menu}}
     ${escapetool.h}panelfooter()
     {{/velocity}}")
   #set ($menuBar = "{{velocity}}
     ${escapetool.h}if (${escapetool.d}xwiki.hasAccessLevel('view', ${escapetool.d}xcontext.user, ${escapetool.d}uix.doc.documentReference))
-      {{menu type=""horizontal fixedWidth"" id=""menu-horizontal-$doc.fullName.replaceAll('[\s|,|.|:|\[|\]|\[\/\]|\[\\\]|=|@|#]', '-')""}}{{include reference=""${escapetool.d}uix.doc.documentReference"" /}}{{/menu}}
+      {{menu type=""horizontal fixedWidth""
+       id=""menu-horizontal-$doc.fullName.replaceAll('[\s|,|.|:|\[|\]|\[\/\]|\[\\\]|=|@|#]', '-')""
+       label=""$escapetool.velocity($doc.plainTitle)""}}{{include reference=""${escapetool.d}uix.doc.documentReference"" /}}{{/menu}}
```
michitux, line 69: Same as above, Velocity escaping is most likely not enough.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -1159,19 +1159,25 @@ $menuTemplateDoc.content.replace('{', '')
...
 #set ($colorTheme = $xwiki.getUserPreference('colorTheme'))
 #if ("$!colorTheme" != '')
   ## Make sure we use an absolute reference (see XWIKI-9672)
   #set ($colorTheme = $services.model.resolveDocument($colorTheme, $doc.documentReference))
 #end
 #set ($discard = $xwiki.ssx.use("$xcontext.macro.doc.prefixedFullName", {'colorTheme': $colorTheme}))
 #set ($discard = $xwiki.jsx.use("$xcontext.macro.doc.prefixedFullName"))
+## Make sure the label is non-empty as otherwise the aria-label doesn't work.
+#if ($stringtool.isBlank("$!label"))
+  #set ($label = $wikimacro.context.getXDOM().getIdGenerator().generateUniqueId("MenuLabel"))
```
michitux, line 1172: While using the id generator sounds good to get a unique label, I'm wondering if we shouldn't use something more readable/pronounceable here (or maybe just a number?) instead of "IMenuLabel" - "Top navigation menu: IMenuLabel" just doesn't sound good. Also, I'm not sure you can rely on the id generator always being present, so maybe have a fallback for when the call fails.

Also, as there can be just one top navigation menu, wouldn't it be better to just omit the label if there is none provided instead of appending this auto-generated string?

michitux (changes requested):

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
michitux: > I'm worried that the user might not give a name descriptive enough to the menus, so I can't just use it as is.

I don't agree with this. The description of the label parameter should provide enough guidance to pick a suitable label. Also, I'm not sure how much value a prefix "Vertical navigation menu" provides to the user of a screen reader. I have the impression this might just be annoying, in particular if what comes afterwards doesn't expect the prefix as the admin won't be aware of the prefix (the description of the parameter doesn't even mention it). So you might end up with "Top navigation menu: Top navigation".

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
Sereza7: AFAIK the other parameters of the menu macro are named 'id' and 'type', is there something I don't get, or should all of these be capitalized?

michitux: Hm, maybe I'm mixing something here and this is indeed the internal, technical name in which case lowercase would be fine (and expected). However, you should provide a capitalized label in the translations (the other parameters have it there).

vmassol:
Remaining to do before applying the PR:
* Reply to Michael's questions at https://github.com/xwiki/xwiki-platform/pull/2098/files#r1143471743
* Send proposal about adding a "label" parameter to macros generating UI content (navigation or other), e.g. the gallery macro, and also discuss the differences between "label", "title" and "alt". Discuss using "label" vs "name" too.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
Sereza7: > While using the id generator sounds good to get a unique label, I'm wondering if we shouldn't use something more readable/pronounceable here (or maybe just a number?) instead of "IMenuLabel" 

Yesterday with @vmassol  we checked the idGenerator methods, and using 
.generateUniqueId('Menu ', '') 
should give IDs like: Menu 1, Menu 2, Menu 3,...
> Also, I'm not sure you can rely on the id generator always being present, so maybe have a fallback for when the call fails.

The id generator was already used a few lines under there: https://github.com/xwiki/xwiki-platform/blob/cbe530642d81c313d385b3f8401c0056250e1f96/xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml#L1177 without a fallback so I think it's okay to use it there too.
> Also, as there can be just one top navigation menu, wouldn't it be better to just omit the label if there is none provided instead of appending this auto-generated string?


Navigation blocks don't *need* to have a unique label in all cases:
> If a page includes more than one navigation landmark, each should have a unique label.

from [ARIA Patterns](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/navigation.html)
However we have another navigation landmark on all pages, which is the [navbar](https://github.com/xwiki/xwiki-platform/blob/649d442f6879127cc57c4169a48aaf51fb2eedcf/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_macros.vm#L34). Moreover, at least on view pages there should soon be additional navs in the context of [XWIKI-20696](https://jira.xwiki.org/browse/XWIKI-20696).

michitux: I think you misunderstood my comment regarding the omission of the label. What I meant is to omit the user-provided label and instead just use, e.g., "Top navigation menu" as label instead of "Top navigation menu: Menu 2". Now if there can indeed be several top navigation menus, this doesn't make that much sense.

> .generateUniqueId('Menu ', '') should give IDs like: Menu 1, Menu 2, Menu 3,...

I think this will rather give an `IllegalArgumentException` as `StringUtils.isAlpha(prefix)` will be `false` as the provided prefix contains a space. You could try `.generateUniqueId('Menu', '')` to get output like `Menu`, `Menu-1`, `Menu-2` etc.

vmassol: >  You could try .generateUniqueId('Menu', '')

Yes, this is what we had discussed actually. Lucas made a mistake in his reply.

vmassol: Sounds ok. 

We might want to improve the id generation implementation to start at "2" when a suffix is added (to get `Menu` and `Menu-2`), and possibly also add a new API signature to force using the suffix (so that we can have `Menu-1` and `Menu-2`). IMO this can be done separately from this PR.

Note that IMO, you also need to improve the Menu application so that it uses the `label` parameter by default / allow the user to set it (it could be an advanced feature without a specific UI for it and that advanced users could do by editing using the object editor for example, but documented).

michitux: > We might want to improve the id generation implementation to start at "2" when a suffix is added (to get `Menu` and `Menu-2`)

That's not possible/a really breaking change as it will invalidate all links that point to ids (e.g., from headings) with the `-1`-suffix.

vmassol: Yes, it would need to be a different generator or a feature flag for the old generator (a new boolean parameter).

Sereza7:
**PR Changes: **
* Fixed the ID generation for horizontal header menu
* Changed an imprecise translation

Created a [forum post about a standardized parameter "label" for macros generating UI content](https://forum.xwiki.org/t/label-parameter-for-macros-generating-ui-content/12300).

Sereza7:
**Merge conflict solving notes:**
There was a choice to make when it came to the role (from ARIA) of the menu elements: navigation or menu/menubar.

> When including a "menu" for site navigation, do not use the menu role. Rather, for the main site navigation use the native HTML [<nav>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) element or simply a list of links. The menu role should be reserved for composite widgets requiring focus management. See [ARIA practices for disclosure navigation](https://www.w3.org/TR/wai-aria-practices-1.2/examples/disclosure/disclosure-navigation.html) for an explanation and additional examples.

from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/UIExtensionSheet.xml
```diff
@@ -56,12 +56,17 @@
   ## Prepare the extension point map.
   #set ($menuPanel = "{{velocity}}
     ${escapetool.h}panelheader(""$doc.plainTitle"")
-    {{menu type=""vertical collapsible open""}}{{include reference=""${escapetool.d}uix.doc.documentReference"" /}}{{/menu}}
+    {{menu type=""vertical collapsible open""
+      label="$escapetool.velocity($services.rendering.escape($doc.plainTitle, 'xwiki/2.1'))"}}{{include
```
michitux, line 60: Shouldn't this be double quotes (I mean `""`) as otherwise this would end the string that starts at `{{velocity}}`?

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuTranslations.xml
```diff
@@ -80,6 +80,8 @@ rendering.macro.menu.name=Menu
 rendering.macro.menu.description=Displays a menu created using simple wiki syntax (nested lists and links).
 rendering.macro.menu.parameter.id.name=Id
 rendering.macro.menu.parameter.id.description=Optional menu identifier that will be set on the HTML element that wraps the menu. You can use this identifier in JavaScript code to access the menu DOM or in functional tests to assert the menu items.
+rendering.macro.menu.parameter.label.name=Label
+rendering.macro.menu.parameter.label.description=Optional menu label used to describe the content of the menu.
```
michitux, line 84: I would add more context, in particular that this label is important for accessibility as otherwise I think admins won't understand why they should set this label.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -77,6 +77,8 @@ livedata.selection.infoBar.selectedCount={0} entries selected
 livedata.selection.infoBar.allSelected=All entries selected
 livedata.selection.infoBar.allSelectedBut=All entries selected but {0}
 
+livedata.pagination.label={0} live data page info
```
michitux, line 80: I'm not sure about this string. First, is "page info" really how you call this? Not rather something like pagination/pagination control? Second, while the term "live data" might be well-known to regular editors of XWiki, I don't think it is used anywhere in the UI. Therefore, somebody visiting the wiki might not be aware what "live data" is. Third, I'm not sure it is a good idea to start with the id, which might be a technical name and might not be properly pronounced by a TTS engine.

I'm not sure we have a non-technical name how we call "live data" in the UI. So what about something like `Pagination controls for {0}`? Or is it a good practice not to start labels with the same string so users can more quickly differentiate between different labels? I think it could also make sense to add a comment for translators what the parameter content is.

michitux (commented):
In general this looks good, just some small comments.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
Sereza7: > Third, I'm not sure it is a good idea to start with the id, which might be a technical name and might not be properly pronounced by a TTS engine.

The id is defined by the creator of the table. From what I saw in the few LiveTables we have by default in XWiki, it was properly set and quite easy to read. E.g. menu, childrenIndex, ...
To make things cleaner we could create a `label` parameter for the LiveTables, but I think using ids is good enough here.

>I'm not sure we have a non-technical name how we call "live data" in the UI. So what about something like Pagination controls for {0}?

We do use it in the UI only for class names :+1: 
`Pagination controls for {0}` is a way better value than before, thanks! :)
I think I'd make it a bit more precise so that the user knows that they are going to read through a table entry representation of an object, and not a table with the object itself:
`Pagination controls for the {0} table`

> Or is it a good practice not to start labels with the same string so users can more quickly differentiate between different labels? I think it could also make sense to add a comment for translators what the parameter content is.

I don't think this label will be read out too often, so it's okay to have the "specific" part of the label at the end: 
it would have been less okay if we had this read out on every entry in the table. In most cases one page will only contain one table, and even in niche cases, I don't see it being more than two or three tables.

PR #2099, opened 2023-03-01
XWIKI-20694: Mention notifications are talking about AWM in case of mention in an xobject

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20694


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm
```diff
@@ -82,8 +82,9 @@
         #set ($descriptionKey = 'mentions.event.mention.description.comment')
       #elseif($location == "ANNOTATION")
         #set ($descriptionKey = 'mentions.event.mention.description.annotation')
-      #elseif($location == "AWM_FIELD")
-        #set ($descriptionKey = 'mentions.event.mention.description.awmField')
+      #elseif($location == "AWM_FIELD" || $location == "TEXT_FIELD")
+        ## TODO: AWM_FIELD is deprecated since 14.4.8, 14.10.6, and 15.2RC1
```
surli, line 86: Why a TODO? We'll need to support it anyway so not much to do...

surli (approved):
LGTM not sure why it's a draft. Also don't you have an integration test checking the content of the notif?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm
manuelleduc: I need to rephrase it. The todo is mainly to remind me to check if `$location == "AWM_FIELD"` can be moved to legacy (but looks tricky).

manuelleduc:
> LGTM not sure why it's a draft. Also don't you have an integration test checking the content of the notif?

It's a draft because:
- some tests are missing
- I need to move the deprecated enum value to legacy

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/ApplicationResources.properties
```diff
@@ -29,10 +29,25 @@ mentions.event.mention.description=mentioned you on page
 mentions.event.mention.description.document=mentioned you on page
 mentions.event.mention.description.comment=mentioned you on a comment on page
 mentions.event.mention.description.annotation=mentioned you on an annotation on page
-mentions.event.mention.description.awmField=mentioned you on a field on the AWM document
+mentions.event.mention.description.textField=mentioned you on a field on document
```
michitux, line 32: I think there are parameters missing in these translations, you cannot assume that a sentence structure "[User] ... [Document]" is possible in all languages. While this may be hard to change for existing translations, I think at least no new keys with this problem should be introduced.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm
tmortagne: You would not have needed all those `ifs` if you had generated the translation keys with the location value directly instead of doing this mapping.

manuelleduc: At the time, it was not obvious that all sentences would have the same structure. But in hindsight, I agree with you.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/ApplicationResources.properties
manuelleduc: The `[Document]` would be easy to fix. For the `[User]` it needs to be checked since for now [`displayEventDetail`](https://github.com/manuelleduc/xwiki-platform/blob/0dd5eec56bceeee20b0366b92a1c89f094ea32d2/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm#L36) is taking care of inserting the User link.

surli: yes right now it's the standard way of displaying notifications with the existing macros, I don't think we should fix this here as this targets LTS. We'd need to rework all those macros with new ones probably.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/templates/mentions/mention.vm
tmortagne: Now since the content of those keys is not great in terms of "translability" as reported by @michitux maybe it's a good excuse to refactor this logic and generate new keys (based on the location value this time).

manuelleduc: @tmortagne Actually I'm not 100% sure to understand what you mean by `generate new keys (based on the location value this time).` Do you mean using the `choice` feature of message format combined with the index of the enum values?

tmortagne: I mean

```
"mentions.event.mention.description.${location}"
```

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/main/resources/ApplicationResources.properties
```diff
@@ -25,14 +25,34 @@ mentions.mention.event.description=Mentions
...
+## Note: each element should be removed when the last branch using it is no longer supported
+###############################################################################
+
+## Used to indicate where deprecated keys start
+#@deprecatedstart
+
+#######################################
+## until 14.10.7, 15.2RC1 
+#######################################
+mentions.event.mention.description=mentioned you on page {0}
```
surli, line 51: From what I see in the diff the previous translation did not have an argument: 
> mentions.event.mention.description=mentioned you on page

manuelleduc: done

PR #2100, opened 2023-03-01
XWIKI-20637: Image overlaps lists when using start alignment and a mix with text wrap and the start alignment

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20637


manuelleduc:
An con the `overflow: auto` is that the part of a list going below the image is not indented to the left anymore.

Before:

image

After:

image

michitux:
Have you checked what happens when you have actual content like long text in the list? Or maybe something like an image that could exceed the width of the document?

manuelleduc:
> Have you checked what happens when you have actual content like long text in the list? Or maybe something like an image that could exceed the width of the document?

Not yet, that's the kind of things I need to consider.

PR #2103, opened 2023-03-03
XWIKI-20702: Allow to chose the fallback version in case of conflict when performing merge

surli:
* Provide a new configuration option for allowing to fallback on another version
  * Use that new configuration in all merge operations
  * Change the signature of the BaseProperty#merge method to be able to reference the configuration instance

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -133,7 +133,7 @@
                   <ignore>true</ignore>
                   <code>java.method.addedToInterface</code>
                   <new>method org.xwiki.store.merge.MergeManagerResult&lt;com.xpn.xwiki.objects.ElementInterface, java.lang.Object&gt; com.xpn.xwiki.objects.ElementInterface::merge(com.xpn.xwiki.objects.ElementInterface, com.xpn.xwiki.objects.ElementInterface, com.xpn.xwiki.doc.merge.MergeConfiguration, com.xpn.xwiki.XWikiContext)</new>
-                  <justification>This method replaces the now deprecated old ElementInterface#merge method.</justification>
+                  <justification>Change needed to provide a merge operation taking into account the configuration: we cannot provide a default method as it would wrongly give the feeling that the operation succeeded.</justification>
                 </item>
               </differences>
             </revapi.differences>
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -126,8 +126,30 @@
                  Single justification example:
             -->
             
-            
-            
+            <revapi.differences>
+              <justification>The MergeConfiguration should always have been specified for calling this
+                method, it became obvious when adding configurations we need to reuse.</justification>
+              <criticality>highlight</criticality>
+              <differences>
+                <item>
+                  <code>java.method.numberOfParametersChanged</code>
+                  <old>method void com.xpn.xwiki.objects.BaseProperty&lt;R extends org.xwiki.model.reference.EntityReference&gt;::mergeValue(java.lang.Object, java.lang.Object, com.xpn.xwiki.doc.merge.MergeResult)</old>
+                  <new>method void com.xpn.xwiki.objects.BaseProperty&lt;R extends org.xwiki.model.reference.EntityReference&gt;::mergeValue(java.lang.Object, java.lang.Object, com.xpn.xwiki.doc.merge.MergeConfiguration, com.xpn.xwiki.doc.merge.MergeResult)</new>
+                </item>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.numberOfParametersChanged</code>
+                  <old>method void com.xpn.xwiki.objects.LargeStringProperty::mergeValue(java.lang.Object, java.lang.Object, com.xpn.xwiki.doc.merge.MergeResult)</old>
+                  <new>method void com.xpn.xwiki.objects.LargeStringProperty::mergeValue(java.lang.Object, java.lang.Object, com.xpn.xwiki.doc.merge.MergeConfiguration, com.xpn.xwiki.doc.merge.MergeResult)</new>
+                </item>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.numberOfParametersChanged</code>
+                  <old>method void com.xpn.xwiki.objects.ListProperty::mergeValue(java.lang.Object, java.lang.Object, com.xpn.xwiki.doc.merge.MergeResult)</old>
+                  <new>method void com.xpn.xwiki.objects.ListProperty::mergeValue(java.lang.Object, java.lang.Object, com.xpn.xwiki.doc.merge.MergeConfiguration, com.xpn.xwiki.doc.merge.MergeResult)</new>
+                </item>
+              </differences>
+            </revapi.differences>
           </analysisConfiguration>
         </configuration>
       </plugin>
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -126,14 +126,13 @@
                  Single justification example:
             -->
             <revapi.differences>
-              <justification>Needed changes for properly handling merge with fallback version.</justification>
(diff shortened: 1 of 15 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/BaseProperty.java
```diff
@@ -261,7 +265,8 @@ public void merge(ElementInterface previousElement, ElementInterface newElement,
      * @param mergeResult merge report
      * @since 3.2M1
```
tmortagne, line 266: The since is not accurate anymore.

surli: > There was really no way to keep the old one, just not used ?

No I could probably keep the old one and make it deprecated. Wasn't sure it worth the effort as I don't see it much used in XS so I don't expect it to have been used. But if you feel we should keep it I can.

tmortagne: Better keep APIs, just in case, you never know what one might be doing in some extension and the promise is supposed to be that it's not going to be broken.

surli: Done with the new changes, but since I tried not using a deprecated class in argument, the changes became much more complex... sorry for that...

surli:
crap... I just realized that I lost changes introduced for XWIKI-20698: MergeResult is missing information about location of conflicts

xwiki-platform-core/pom.xml
```diff
@@ -125,9 +125,18 @@
...
-            
+            <revapi.differences>
+              <justification>Needed changes for properly handling merge with fallback version.</justification>
+              <criticality>highlight</criticality>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.addedToInterface</code>
+                  <new>method org.xwiki.store.merge.MergeManagerResult&lt;com.xpn.xwiki.objects.ElementInterface, java.lang.Object&gt; com.xpn.xwiki.objects.ElementInterface::merge(com.xpn.xwiki.objects.ElementInterface, com.xpn.xwiki.objects.ElementInterface, com.xpn.xwiki.doc.merge.MergeConfiguration, com.xpn.xwiki.XWikiContext)</new>
+                  <justification>This method replaces the now deprecated old ElementInterface#merge method.</justification>
```
tmortagne, line 136: This is not the expected information, IMO. The point here is to explain why we are OK to break this API. So this should indicate why we don't think providing a default implementation is required (as it seems technically doable from what I see, it would just return a `MergeManagerResult` without Conflict, but it's less of a problem than a failing merge call).

surli: > it seems technically doable from what I see

yes it's technically doable, but I didn't want to do it as it wouldn't respect the contract of the config in case of the config is set to modify the current instance: calling it wouldn't have any effect, and the returned result would be wrong. So IMO it's better to just not implement it. 
I guess I can provide those info here.

tmortagne: My main point is not that we should provide a default implementation (I don't think anyone implements this interface directly) but that what is indicated is not what is needed.

vmassol: Yes, we're never supposed to break backward compatibility. When we think we have no choice, this requires a VOTE, to make sure committers review it and agree (maybe some ideas will emerge to avoid the breakage). At the very least the VOTE needs to explain why it's ok to break it and why the impact is very limited.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/BaseCollection.java
```diff
@@ -821,67 +822,92 @@ public void setDocumentReference(DocumentReference reference)
...
                     if (propertyResult.equals(previousProperty)) {
                         // Let some automatic migration take care of that modification between DB and new
-                        safeput(diff.getPropName(),
+                        modifiableResult.safeput(diff.getPropName(),
                             configuration.isProvidedVersionsModifiables() ? newProperty : newProperty.clone());
                         mergeResult.setModified(true);
                     } else if (!propertyResult.equals(newProperty)) {
                         // Try to apply 3 ways merge on the property
-                        mergeField(propertyResult, previousProperty, newProperty, configuration, context, mergeResult);
+                        // FIXME: we should deprecate mergeField and rewrite it properly, but it's lot of work
```
manuelleduc, line 881:
```suggestion
                        // FIXME: we should deprecate mergeField and rewrite it properly, but it's a lot of work
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/ElementInterface.java
```diff
@@ -68,10 +70,32 @@
      * @param context the XWiki context
      * @param mergeResult the merge report
      * @since 3.2M1
+     * @deprecated Since 14.10.7,15.2RC1 use
```
vmassol, line 73: Usually I don't repeat the version since it's now in the `@Deprecated` annotation. I would just write:

`@deprecated use {@link #merge(ElementInterface, ElementInterface, MergeConfiguration, XWikiContext)}.``

WDYT?

surli: Fine, I took the habit to specify the version before we had it in the annotation.

tmortagne: > WDYT?

That's actually our current rule (but I would not be surprised to see that I did the same mistake by reflex).

vmassol: Yeah, we don't really say it explicitly , but we have https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HDeprecation (I'll update it to mention it)

xwiki-platform-core/pom.xml
surli: So I changed the revapi message, let me know if that's ok.

tmortagne: Looks better yes. I guess I would have added a note indicating that the chance that this interface is directly implemented are close to 0 since it would not work in most cases.

xwiki-platform-core/pom.xml
```diff
@@ -125,9 +125,18 @@
...
-            
+            <revapi.differences>
+              <justification>Needed changes for properly handling merge with fallback version.</justification>
+              <criticality>highlight</criticality>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.method.addedToInterface</code>
+                  <new>method org.xwiki.store.merge.MergeManagerResult&lt;com.xpn.xwiki.objects.ElementInterface, java.lang.Object&gt; com.xpn.xwiki.objects.ElementInterface::merge(com.xpn.xwiki.objects.ElementInterface, com.xpn.xwiki.objects.ElementInterface, com.xpn.xwiki.doc.merge.MergeConfiguration, com.xpn.xwiki.XWikiContext)</new>
+                  <justification>Change needed to provide a merge operation taking into account the configuration: we cannot provide a default method as it would wrongly give the feeling that the operation succeeded.</justification>
```
vmassol, line 136: I believe we need to explain why we're ok to break it. And the main reason is because this is an API that's unlikely to be extended.

surli: Yes that's also @tmortagne feedback in the thread above. I'll add a note.

PR #2104, opened 2023-03-06
XWIKI-19383: Display a title on createlink

Sereza5:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-19383

# Changes

## Description

* Added translated wanted link title generators for documents and pages, for spaces, and a
  fallback for any other kind of reference, implementing the `WantedLinkTitleGenerator` role added
  in xwiki/xwiki-rendering#256.
* Added the wanted link title translation keys to the rendering module's own resource bundle.
* Added a title to the wanted links rendered by the page displayer and by the App Within Minutes
  class editor sheet, which both hand-write their own `wikicreatelink` markup.

## Clarifications

* **xwiki/xwiki-rendering#256 has to be merged first.** This PR only compiles against the
  `WantedLinkTitleGenerator` role it adds, so the `Quality / Analyze` check fails with
  `cannot find symbol: class WantedLinkTitleGenerator` until that PR is merged and its snapshot is
  published. That is expected, and it is why this PR is still a draft.
* The translation keys live in `xwiki-platform-rendering-xwiki`'s own
  `ApplicationResources.properties`, next to the sibling `rendering.*` keys, rather than in
  oldcore. `RootClassLoaderTranslationBundle` merges every `ApplicationResources.properties` on the
  classpath, so the `.vm` template and the XAR sheet resolve them identically.
* The class editor sheet uses its own `platform.appwithinminutes.classEditorCreatePageTitle` rather
  than borrowing a `rendering.xwiki.*` key from another extension.
* Attachment wanted links are **not** covered. No renderer emits `wikicreatelink` for an attachment
  reference — `AttachmentXHTMLLinkTypeRenderer` extends `AbstractXHTMLLinkTypeRenderer` and only
  ever emits `wikiattachmentlink` — so a generator for them would be unreachable. Happy to wire
  that up, as its own issue.

# Screenshots & Video

See below

# Executed Tests

```
mvn clean install -Plegacy,quality \
  -pl xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki,\
xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui,\
xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates
```

`BUILD SUCCESS` on the three modules: 124 tests in `xwiki-platform-rendering-xwiki`, 82 in
`xwiki-platform-web-templates`, 5 in `xwiki-platform-appwithinminutes-ui` including `xar:verify`,
0 Checkstyle violations, all coverage checks met, Revapi clean. Built against a local
`xwiki-rendering` 18.8.0-SNAPSHOT carrying xwiki/xwiki-rendering#256, since it is not merged yet.

New tests: `XWikiDocumentWantedLinkTitleGeneratorTest` with 4 cases — untyped reference, typed
reference, nested page, page reference — plus tests for the space and default generators, and a
`ClassEditSheetPageTest` case asserting the create title lands on the anchor. Each was checked
against the pre-fix code to confirm it fails there rather than merely passing.

The titles those tests pin down, since there are no screenshots:

| Reference | Rendered title |
|---|---|
| `[[Space.UnknownPage]]` | `Create page: UnknownPage` |
| `[[Foo>>doc:.Fa\.Fi\.Foo.WebHome]]` | `Create page: Foo` |
| a space reference | `Create space: <name>` |
| anything else | `Create reference: <reference>` |

Unrelated caveat: a commons SNAPSHOT from 2026-09-06 pulls surefire 3.6.0, which fails every module
with `ClassNotFoundException: org.xwiki.test.CaptureConsoleRunListener` before any test runs, on a
clean master checkout too. The runs above used `-Dsurefire.version=3.5.6`.

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * None — depends on a new API added in master only.


Sereza7:
**PR Changes:**
* Solved the mismatch in hints for the URITitleGenerator implementation

**NOTE:** This commit needs to be merged with/after https://github.com/xwiki/xwiki-rendering/pull/256

**View:**
19383-InitialPRState

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/displayer_page.vm
```diff
@@ -57,7 +57,8 @@
   #else
     #set ($title = $escapetool.xml($pageDoc.plainTitle))
     #if ($pageDoc.isNew())
-      <span class="wikicreatelink"><a href="$pageDoc.getURL('create')">$title</a></span>
+      <span class="wikicreatelink"><a href="$pageDoc.getURL('create')"
+      title="$services.localization.render('core.create.inline.label', [$title])">$title</a></span>
```
manuelleduc, line 61: Wrong formatting + escaping missing.
```suggestion
      <span class="wikicreatelink">
        <a href="$pageDoc.getURL('create')" title="$services.localization.render('core.create.inline.label', [$title])">$title</a>
      </span>
```

Sereza7:
**PR Changes:**
* Fixed escaping on title.
* Properly formatted changes.

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
```diff
@@ -256,7 +256,10 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
     #set ($action = 'create')
     #set ($discard = $params.put('parent', $doc.fullName))
   #end
-  &lt;span class="$class"&gt;&lt;a href="$escapetool.xml($xwiki.getURL($reference, $action, $escapetool.url($params)))"
+  &lt;span class="$class"&gt;
+  &lt;a href="$escapetool.xml($xwiki.getURL($reference, $action, $escapetool.url($params)))"
```
mflorea, line 260: The span and anchor were put on the same line most probably to avoid white space. I don't see any reason to change this line, as it fits the 120 characters limit.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+import org.xwiki.model.reference.PageAttachmentReferenceResolver;
+import org.xwiki.rendering.listener.reference.ResourceReference;
+import org.xwiki.localization.ContextualLocalizationManager;
+import org.xwiki.rendering.renderer.reference.link.URITitleGenerator;
+
+/**
+ * Generate link titles for ATTACH URIs.
+ *
+ * @version $Id$
+ * @since 15.2RC1
```
mflorea, line 37: This needs to be updated.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+    @Named("current")
+    private PageAttachmentReferenceResolver<String> currentAttachmentReferenceResolver;
+
+    @Override
+    public String generateCreateTitle(ResourceReference reference)
+    {
+        PageAttachmentReference attachmentReference =
+            this.currentAttachmentReferenceResolver.resolve(reference.getReference());
+        return this.contextLocalization.getTranslationPlain("core.create.inline.label",
+         attachmentReference.getName());
```
mflorea, line 59: I don't understand, you're returning "Create page: attachmentName"? Doesn't make any sense. Also bad formatting.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/displayer_page.vm
```diff
@@ -55,9 +55,14 @@
     ## The view displayer generates HTML so we need to escape the XML special characters from the page reference.
     $escapetool.xml($pageReference)
   #else
-    #set ($title = $escapetool.xml($pageDoc.plainTitle))
+    #set ($title = $escapetool.velocity($pageDoc.plainTitle))
```
mflorea, line 58: This is wrong. You need XML escaping because you're printing the value below in HTML.

mflorea (changes requested):
This PR looks unfinished.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
Sereza7: :+1: for the formatting.
When used on a rendered link, it seems to me like it works properly, the reference name is the name of the page we're looking to create. What should I use in its stead?

For reference, I just checked again this code in action:
> [[Dead link>>The dead page that should be created]]

19383-titleGenerator

vmassol:
Please add to the jira issue the exact WCAG problem (link to it) + mention if it's auto detected by Axe core or not, and if not, do we need to define a best practice for developers to avoid regressing on this (or other use cases) in the future.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+import org.xwiki.localization.ContextualLocalizationManager;
+import org.xwiki.rendering.renderer.reference.link.URITitleGenerator;
+
+/**
+ * Generate link titles for document URIs.
+ *
+ * @version $Id$
+ * @since 15.3RC1
+ */
+@Component(hints = {"doc", "page"})
```
vmassol, line 39: Not needed.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    @Inject
+    @Named("current")
+    private DocumentReferenceResolver<String> currentDocumentReferenceResolver;
+
+    @Override
+    public String generateCreateTitle(ResourceReference reference)
+    {
+        DocumentReference documentReference =
+            this.currentDocumentReferenceResolver.resolve(reference.getReference());
```
vmassol, line 57: This will only work for `DocumentResourceReference` and for other types, it'll cause a bug. For example `SpaceResourceReference` references won't be resolved correctly.

michitux (commented):
There is one missing escaping (see comment). I'm not 100% sure regarding the different resolver implementations but it looks good.

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/ClassEditSheet.xml
```diff
@@ -256,9 +256,13 @@ xcontext.put('propertyCustomDisplayer', new PropertyCustomDisplayer(xcontext))
     #set ($class = 'wikicreatelink')
     #set ($action = 'create')
     #set ($discard = $params.put('parent', $doc.fullName))
+    #set ($title = $services.localization.render('core.create.inline.label',[$reference.name]))
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2104/commits/45a7315dbb73d755e991db33c3ad681c20022d3d

Sereza5:
Rebased onto current master and reworked. The description is the state at open time; this covers what
changed since. The branch was 4733 commits behind with seven merge-from-master commits and one
unrelated XWIKI-19450 commit, so I replanted the effective diff rather than replay them. `@since`
tags moved to `18.8.0RC1`.

## A user-visible bug on the primary path

`XWikiDocumentWantedLinkTitleGenerator` named the wrong thing on **both** reachable branches:

| Input | Before | Now |
|---|---|---|
| `[[Space.UnknownPage]]` | `Create page: Space.UnknownPage` | `Create page: UnknownPage` |
| `[[Foo>>doc:.Fa\.Fi\.Foo.WebHome]]` | `Create page: WebHome` | `Create page: Foo` |

- The `isTyped()` guard could never pass: parsers call `setTyped(false)` for any reference without an
  explicit `type:` prefix, and the generator is already selected *by* scheme `doc`/`page` — so the
  `else` branch was the only path for ordinary wiki syntax. `XWikiSpaceWantedLinkTitleGenerator` has
  no such guard, which is what suggested it was unintentional.
- `CurrentStringDocumentReferenceResolver` does no nested-page collapsing, so `A.B.WebHome` gives
  name `WebHome`. That is what CKEditor emits for a non-existent nested page — see
  `createLinkForNotExistingPage` in the CKEditor `LinkIT`.

Fixed by dropping the guard and naming the document the way the default link label format does,
reusing `%np` from `XWikiLinkLabelGenerator.getNestedPageLabel`. Before, a link's visible label and
its tooltip disagreed about the page name.

## Other fixes

- **Two keys were referenced but never defined**: `core.create.inline.label` exists nowhere in the
  repo, and `rendering.xwiki.wantedLink.document.label` was never added — so the ticket's primary
  case rendered a raw key as the tooltip. Keys used and defined now match exactly.
- **Keys moved out of oldcore** into `xwiki-platform-rendering-xwiki`'s own bundle, which already
  carries the sibling `rendering.*` keys. Done before 18.8.0 ships deliberately: once released the
  convention is to deprecate and re-key rather than move.
- **The AWM sheet had been rendering a `rendering.xwiki.*` key**; it now has its own
  `platform.appwithinminutes.classEditorCreatePageTitle`.
- **`XWikiAttachmentWantedLinkTitleGenerator` was unreachable** and is removed with its key and
  `components.txt` entry — `wikicreatelink` is emitted only by
  `DocumentXHTMLLinkTypeRenderer.beginInternalLink`, and the attachment renderers never emit it.
  Happy to wire attachment wanted links up instead, as a separate issue.
- **Double escaping** in `displayer_page.vm`: `$title` was already escaped, then escaped again, so
  `&` came out as `&amp;amp;`.
- Removed a duplicate `xwiki-platform-localization-api` dependency; moved the class editor title onto
  the anchor for the accessibility reason in xwiki/xwiki-rendering#256.

## Tests

The four generators shipped with no tests, which is what let the naming bug through. Added tests for
the document, space and default generators plus a `ClassEditSheetPageTest` case. Each was checked
against the pre-fix code to confirm it fails there — the generator tests with
`Create page: Space.UnknownPage` and `Create page: WebHome`.

`displayer_page.vm` still has no page test; `displayPageReference` needs a document with a Page-type
property, which is a larger fixture. Not in this batch.

## Why CI is red

`Quality / Analyze` fails to compile with `cannot find symbol: class WantedLinkTitleGenerator` in the
three generators — nine errors, all the same missing interface. That interface is added by
xwiki/xwiki-rendering#256, which is not merged, so the published
`xwiki-rendering-api:18.8.0-SNAPSHOT` that CI resolves does not contain it. I checked: the file does
not exist on xwiki-rendering master, only on that PR's branch.

Nothing to fix on this side. The job dies at `Build`, so `Analyze` never runs and there is no second
problem hidden underneath, and it will stay red on every push until #256 merges and its snapshot
publishes. That is why this PR is still a draft.

**Merge order:** xwiki/xwiki-rendering#256 goes first; this only compiles against it.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
Sereza5: Kept. These hints are what the renderer looks a generator up by, so this is the same decision as your
comment on the by-scheme lookup — let's keep it in one place, on the PR that merges first. I am
replying there: https://github.com/xwiki/xwiki-rendering/pull/256#discussion_r1152126160

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/renderer/XWikiDocumentURITitleGenerator.java
Sereza5: Fixed in 0f664d6b7df, and thank you — it was real, and worse than the comment says. `space` now has
its own generator and `page` resolves through `currentPageReferenceResolver`, but the class was also
naming the wrong thing on both of its remaining paths:

| Input | Before | Now |
|---|---|---|
| `[[Space.UnknownPage]]` | `Create page: Space.UnknownPage` | `Create page: UnknownPage` |
| `[[Foo>>doc:.Fa\.Fi\.Foo.WebHome]]` | `Create page: WebHome` | `Create page: Foo` |

The `isTyped()` guard could never pass, since parsers call `setTyped(false)` for any reference
without an explicit `type:` prefix while the generator is already selected by scheme — so the `else`
branch, using the whole reference, was the only live path for ordinary wiki syntax. And
`CurrentStringDocumentReferenceResolver` does no nested-page collapsing, so `A.B.WebHome` gave
`WebHome`, which is the form CKEditor emits for a non-existent nested page.

Both are fixed by dropping the guard and reusing the `%np` rule from
`XWikiLinkLabelGenerator.getNestedPageLabel`. There are now four tests on this class, each checked
against the pre-fix code to confirm it fails there.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/displayer_page.vm
Sereza5: Fixed, though not with the suggested snippet.

It now renders `rendering.xwiki.wantedLink.page.label` with the raw `$pageDoc.plainTitle`, escaped
once, on a single line so no whitespace lands inside the anchor text.

Sereza5:
Leaving this PR as a draft until https://github.com/xwiki/xwiki-rendering/pull/256 is merged.

PR #2106, opened 2023-03-07
XWIKI-20696: Accessibility best-practice: use landmarks in view pages

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20696

**Forum discussion:** https://forum.xwiki.org/t/accessibility-best-practice-landmarks/12051

**PR Changes:**
* Added header and footer landmarks.
* Added main landmark for view pages.
* Added navigation and complementary landmarks for panels (with aria-labels)
* Added landmarks for contentheader, contentfooter, extradoc (comments, attachment, ...) and breadcrumbs.

***Note:*** The examples under have been taken without the `xwiki-platform-panels-ui` xar updated. That's why all the panels are landmarked as `Complementary`.

**View:**
Landmarks are highlighted with https://addons.mozilla.org/en-US/firefox/addon/landmarks/ .
Dashed boxes are `guessed` landmarks. Those are not read by a default configuration NVDA. However they might show up with some tools. These guessed landmarks are based on `class` (and not `role` or element nature).
e.g. there's three element with the class `nav` in the Application panel and the tool guesses that there are three navs here.
It would be a good practice to avoid using such classes.
See https://github.com/matatk/landmarks/blob/1ef7cbd194ecbfc8516c40961b90965c71dbe2f6/src/assemble/_landmarksFinder.js#L451-L477 for examples of ids/classes to avoid.

*Before*
19450-individualLabels
*After*
20696-afterInitialCommit

**Video demo:**
It's easier to navigate through the page using a screen reader. With NVDA, the D (and shift D) key press sends to the next landmark.

https://user-images.githubusercontent.com/28761965/223481055-e1348b86-dab8-4d82-97bb-1aa3a062303c.mp4


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/hierarchy_macros.vm
```diff
@@ -146,7 +146,10 @@
       #set ($breadCrumbParameters = "$breadCrumbParameters data-${option.toLowerCase()}='$escapetool.xml($options.get($option))'")
     #end
   #end
-  <ol #if("$!options.id" != '')id="$options.id"#end class="$breadCrumbClassNames" $breadCrumbParameters><li
+  ## Set as a navigation role because it contains mostly links.
+  <nav aria-label="#if("$!options.id" != '')$options.id#end $breadCrumbClassNames">
```
mflorea, line 150: I find it strange to use the id and class names as the value for the ``aria-label``. I thought the practice is to either use a human readable text with ``aria-label`` (which is not the case here) or use ``aria-labelledby`` to target an existing element.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -819,22 +819,36 @@ $fn
...
 #end
 
-#macro(largepanelheader $title)
-#panelheader($title, true, false)
+## Changes the semantic role of the panel.
+#macro(navigationpanelheader $title)
+  #panelheader($title, false, false, true)
+#end
+
+#macro(largenavigationpanelheader $title)
```
mflorea, line 850: For new macro names we should use camel case.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/hierarchy_macros.vm
Sereza7: Thanks! I'll look for a solution and fix this soon. 
I really messed up this PR state, so if it's okay I'll close it and start a new fresh PR from https://github.com/Sereza7/xwiki-platform/tree/XWIKI-20696
Right now because of my mess ups, it's almost impossible to review the PR anymore and I'm pretty sure I can't fix this myself :/

Sereza7: Using the id makes sure we have a unique aria-label, which is what we're aiming for. I'm working on an implementation of a label parameter. However, default hierarchies (without a label) will have the same label, which could be problematic in some cases. 
> If a page includes more than one navigation landmark, each should have a unique label.

https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/navigation.html

manuelleduc: > Right now because of my mess ups, it's almost impossible to review the PR anymore and I'm pretty sure I can't fix this myself :/

Something like the following should set the PR in a correct state:
```bash
$ git fetch
$ git rebase origin/master ## possibly fixing some conflicts
$ git push NAME_OF_REMOTE NAME_OF_LOCAL_BRANCH:NAME_OF_REMOTE_BRANCH --force-with-lease
```

Sereza7: Okay, thanks!

Sereza7: @manuelleduc Unfortunately, I can't reopen this PR, I created https://github.com/xwiki/xwiki-platform/pull/2133 instead and just after that saw your message.
This helped me "save" [this other PR](https://github.com/xwiki/xwiki-platform/pull/2107) though, thanks again!

PR #2108, opened 2023-03-08
[Misc] Document that 0 disables the limit in Query

raphj:


xwiki-platform-core/xwiki-platform-query/xwiki-platform-query-manager/src/main/java/org/xwiki/query/Query.java
```diff
@@ -175,7 +175,7 @@ default QueryParameter bindValue(String var)
     Query setOffset(int offset);
 
     /**
-     * @return limit the limit of result list to set ({@code execute().size() <= limit})
+     * @return the limit of result list to set ({@code execute().size() <= limit}). 0 for no limit.
```
surli, line 178: Well 0 means that there's no defined limit in the query: I guess there could always been a default limit provided by the implementation of the query executor.

tmortagne: By the way, it's not really 0 in practice right now, but anything <= 0.

Virad01 (approved):

PR #2110, opened 2023-03-08
XWIKI-18859: On the login page, the form controls receive focus before the banner elements

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-18859
**PR Changes:**
* Removed tab indexes from form elements.
* Added a semantic landmark for easier navigation from/to the main content (the form).
* Changed the focus at load from a js script block to an HTML5 attribute

**View:**
The tab order on a xwiki distrib as displayed by the Firefox accessibility tool.
*Before:*
18859-TabOrder

*After:*
18859-tabIndexAfterChanges

___
*Notes:* 
* The js block replaced was created in 2008 when the autofocus attribute from HTML5 had barely any support on Web browsers.
 See this commit: https://github.com/xwiki/xwiki-platform/commit/f7bc659c5e159c61e422ee32cb3bb2fb41d604b1
* The PR as of now adds one action when filling out the form with keyboard interactions: pass the link `Forgot your password?` by pressing Tab one more time than what used to be (ie. password right after username in the tab order).


michitux:
I think not having the password as next element after the username is really bad UX. When I fill a login form (manually), I really expect that password is one tab away from username. Could we maybe move both links after the login form, i.e., after the "Log-in" button? That's, e.g., how [GitHub](https://github.com/login) has it in tab-order (and HTML element order) - they still display the link above the password input using some clever relative positioning which works with one link but not with two as in our case I fear.

Sereza7:
The first link `Forgot your username?` is before in tab-order the username input as of now, so we could do the same as GitHub to obtain a similar solution.
However if we do that, we'd have one recover link before the inputs, and the other one after inputs.

I'll try this solution out :+1:

michitux:
What I meant to say but didn't say clearly enough is that for me it would actually also be okay to have both links displayed below the button. The important part for me is just that they are in the tab order not between the elements you regularly use for login as this really breaks the expected experience.

Sereza7:
**PR Changes:**
* Changed recovery links position to be displayed after the submit form button.
_Note:_ The username and password fields are now (as initially) one after the other in the tab-order.

**View:**
18859-ChangedRecoveryLinks
The tab order, DOM order and visual order of the elements all align.
Both of the recovery links are now displayed under the `Log-in` button.

michitux:
I think the layout looks strange now, the alignment to the right just seems wrong. Maybe both in a line? Not sure, would be good to have some more input from others.

Virad01 (approved):

Sereza7:
**PR Changes:**
* Updated style according to forum feedback.

**View:**
18859-UpdatedView
18859-tablistUpdated

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/login.vm
```diff
@@ -58,48 +59,46 @@
...
+  <dt class="submitSection">
+    <input type="submit" class="btn btn-primary col-xs-12" value="$services.localization.render('login')"/>
+
+    #set ($forgotUsernameURL = $services.security.authentication.getAuthenticationURL('retrieveusername', $NULL))
+    #if("$!forgotUsernameURL" != "")
+      <span class="xAdditional"><a href="$forgotUsernameURL">$services.localization.render('xe.admin.forgotUsername.loginMessage')</a></span>
+    #end
+    #set ($resetPasswordUrl = $services.security.authentication.getAuthenticationURL('resetpassword', $NULL))
+    #if("$!resetPasswordUrl" != '')
+      <span class="xAdditional recoverPassword"><a href="$resetPasswordUrl" >$services.localization.render('xe.admin.passwordReset.loginMessage')</a></span>
```
tmortagne, line 95: The result of this translation should probably be XML escaped.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/login.vm
```diff
@@ -58,48 +59,46 @@
...
       $supplementaryForm
     </div>
   #end
-  <dt><input type="submit" class="btn btn-primary col-xs-12" value="$services.localization.render('login')" tabindex="400"/></dt>
+  <dt class="submitSection">
+    <input type="submit" class="btn btn-primary col-xs-12" value="$services.localization.render('login')"/>
+
+    #set ($forgotUsernameURL = $services.security.authentication.getAuthenticationURL('retrieveusername', $NULL))
+    #if("$!forgotUsernameURL" != "")
+      <span class="xAdditional"><a href="$forgotUsernameURL">$services.localization.render('xe.admin.forgotUsername.loginMessage')</a></span>
```
tmortagne, line 91: The result of this translation should probably be XML escaped.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/login.vm
```diff
@@ -58,48 +59,46 @@
...
   <dd class="input-group">
     <span class="input-group-addon">$services.icon.renderHTML('lock')</span>
-    <input type="password" id="j_password" size="60" name="j_password" value="" tabindex="200"/>
+    <input type="password" id="j_password" size="60" name="j_password" value="" />
   </dd>
   <dt></dt>
   <dd>
     <label class="xHint" for="rememberme">
-      <input id="rememberme" type="checkbox" name="j_rememberme" value="true" tabindex="300"/> $services.localization.render('remembermeonthiscomp')
+      <input id="rememberme" type="checkbox" name="j_rememberme" value="true" /> $services.localization.render('remembermeonthiscomp')
```
mflorea, line 78: I know you only removed the tab index but it may be worth adding XML escaping too.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/login.vm
mflorea: Same for all the above.

Sereza7:
**PR Changes:**
* Escaped translated strings.

**No style change**

mflorea (approved):

PR #2111, opened 2023-03-10
XWIKI-20724: Allow extensions to add parameters to the html conversion request

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20724

Required for https://jira.xwiki.org/browse/NCAPP-68

Example of use by an extension:

```javascript
$(document).on('xwiki:ckeditor:convertHTML', function(event, data) {
    data.example = 1;
});
```

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-source/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-source/plugin.js
@@ -32,7 +32,7 @@
       // The values of this object are then used as parameters on the html convertion request.
       var extendedParams = {};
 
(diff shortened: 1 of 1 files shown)


mflorea (approved):
Looks good, my only problem is that by listening to ``xwiki:ckeditor:convertHTML`` you tie yourself to CKEditor which is just one possible WYSIWYG editor implementation. An XWiki extension can provide a different WYSIWYG editor that the administrator can configure in the administration as the default editor. The question is basically if this need to add a parameter to the HTML conversion is specific to the CKEditor implementation or not. I'd say that the HTML conversion is generic and other WYSIWYG editors will need it as long as:
* we store the content using wiki syntax (and not HTML) (I don't see us changing this anytime soon)
* the conversion to wiki syntax is done on the server side (I also don't see us duplicating the rendering code on the client side anytime soon)
With this in mind may it would be good to make the event generic ``xwiki:wysiwyg:convertHTML``?

PR #2112, opened 2023-03-10
XWIKI-20722: JSX Objects are not loaded in the Rendered Diff view

oanalavinia:
* inject required skin extensions

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesinline.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesinline.vm
@@ -17,6 +17,10 @@
 ## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 ## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 ## ---------------------------------------------------------------------------
+#if ($xcontext.action == 'get')
+  #template('display_macros.vm')
+  #initRequiredSkinExtensions()
+#end
 #set($showactions = 0)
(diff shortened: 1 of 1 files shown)


mflorea (approved):
Looks good. I'd improve the comment, since it's not only about JSX (there's SSX too and any other skin extension).

surli:
> @surli do you foresee any problem with this for your work on Change Request?

Fine with me, I'm not using this template.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesinline.vm
```diff
@@ -89,3 +93,8 @@
 #set ($nextNewdocVersion = $newdoc.documentArchive.getNextVersion($newdoc.RCSVersion))
 #template('changesdoc.vm')
 </div>
+## Pass the required skin extensions to the client side to be injected automatically in the HTML head by xwiki.js.
+#if ($xcontext.action == 'get')
+  #getRequiredSkinExtensions($requiredSkinExtensions)
+  #set ($discard = $response.setHeader('X-XWIKI-HTML-HEAD', $requiredSkinExtensions))
```
tmortagne, line 99: Not a blocker for this PR, but I have the feeling that we should automate this more instead of having it in various tamplates. All it takes is to push an async context when an ajax request starts and automatically "augment" the response with the required HTTP headers when we are done executing it.

oanalavinia: Indeed. I'll create a jira issue for this

PR #2125, opened 2023-03-22
XWIKI-18004: Alert menu cannot be activated with keyboard + hover message not accessible with keyboard

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-18004

**PR Changes:**
* Changed the dropdown toggle to a button. 
* Removed superfluous markup from the element.
* Updated css rules on hover and focus

**Notes:**
* The dropdown toggle used to be a link but the correct use of links is for navigation. Here we want to trigger a change of interface (and not move to somewhere else in the interface), so we should use a button. Buttons are properly accessible even without any text content.
* Because of a bootstrap dependency, there is still major accessibility violations inside the popup menu that should be reported one by one and addressed.  

**View:**
https://up1.xwikisas.com/#DbzK6mOlPPHzA-k9nquRcA

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsTrayPage.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-test/xwiki-platform-notifications-test-pageobjects/src/main/java/org/xwiki/platform/notifications/test/po/NotificationsTrayPage.java
@@ -46,7 +46,7 @@ public class NotificationsTrayPage extends ViewPage
     @FindBy(css = "li#tmNotifications div.notifications-area")
     private WebElement notificationsArea;
 
-    @FindBy(css = "li#tmNotifications a[title='Notifications']")
+    @FindBy(css = "li#tmNotifications button[title='Notifications']")
     private WebElement watchListButton;
 
     @FindBy(css = "li#tmNotifications")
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/action-menus.less
```diff
@@ -151,6 +151,28 @@
   }
 }
 
+#tmNotifications {
```
mflorea, line 154: The ``#tmNotifications`` HTML element is generated by a UIX defined in a wiki page, ``XWiki.Alerts.Code.AlertsMenuUIX``, so I'd prefer to keep the styles in the same place (to be packaged / installed / uninstalled together), e.g. using a SSX object on that page with LESS enabled.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/action-menus.less
Sereza7: As there is already style related to the notification header in `XWiki.Notifications.Code.NotificationsDisplayerUIX`, I'll change the CSS SSX in here to a LESS and move my new style here :)
Note: There's also the javascript dealing with the behavior of `#tmNotifications` here

mflorea (approved):

PR #2129, opened 2023-03-23
[Misc] Fix URL handling pageReady.js

michitux:
* Make sure all URLs are absolute
* Handle fetch requests with Request parameter


mflorea (approved):
Looks good. Thanks for taking care of this.

PR #2130, opened 2023-03-27
XWIKI-20762: Large images are overlapped by the PDF footer

slauriere:
- In xhtml2fo XSL, update property "block-progression-dimension.maximum" from fixed value to dynamic value taking into account page header and footer heights, in the context of an A4 page only.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xhtml2fo.xsl
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xhtml2fo.xsl
@@ -34,6 +34,8 @@ WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING O
     <!-- page size -->
     <xsl:param name="page-width">auto</xsl:param>
     <xsl:param name="page-height">auto</xsl:param>
+    <!-- a4 page height is used to compute the maximal height of images in template "process-img" -->
+    <xsl:param name="a4-page-height">11.7in</xsl:param>
     <xsl:param name="page-margin-top">1in</xsl:param>
     <xsl:param name="page-margin-bottom">1in</xsl:param>
     <xsl:param name="page-margin-left">1in</xsl:param>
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/xhtml2fo.xsl
```diff
@@ -2077,7 +2079,11 @@ WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING O
         <xsl:attribute name="inline-progression-dimension.maximum">100%</xsl:attribute>
         <!-- Min and max height -->
         <xsl:attribute name="block-progression-dimension.minimum">auto</xsl:attribute>
-        <xsl:attribute name="block-progression-dimension.maximum">700px</xsl:attribute>
+        <!-- Limit height to A4 page height minus the header and footer height minus 1in to leave space below header
+        and above footer. -->
+        <xsl:attribute name="block-progression-dimension.maximum">
+            <xsl:value-of select="$a4-page-height"/> - <xsl:value-of select="$page-margin-top"/> - <xsl:value-of select="$page-margin-bottom"/> - 1in
```
michitux, line 2085: Why don't you use `$page-height` here?

mflorea: @michitux if you want a specific page size you have to set the ``page-width`` and ``page-height`` parameters in a custom ``xhtml2fo.xsl``. The solution for me to set ``page-width`` and ``page-height`` to A4 in the default ``xhtml2fo.xsl`` and use them to compute the footer space. Anyone that wants a different size will modify ``page-width`` and ``page-height`` in their custom ``xhtml2fo.xsl`` and the footer space will be updated automatically.

vmassol: Indeed, it doesn't make sense to use `11.7in`  here and then `page-height` of `auto`. They need to be consistent.

mflorea: FTR, the page width/heigth is set in https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/fop-config.xml#L35 when left as auto in ``xhtml2fo.xsl``.

lucaa: Maybe one of the things we can do is to set the page size in xhtml2fo instead (or in addition to) of the fopconfig. The result XWiki would achieve is the same (A4 exported by default) but the values would be better localized, stored in the file that relies on them.

In this case, the calculation for the image size would be able to use the pageheight instead of a new variable.

vmassol: @lucaa Yep, agreed, this is what we discussed at https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$16800103991469afCmu:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=mtrx.nz

PR #2131, opened 2023-03-28
XWIKI-20781: Livedata colors are hard coded

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20781

**PR Changes:**
* Replaced hard coded values with color theme values, trying to match the closest: value-wise (considering the Iceberg color theme) and then semantically-wise.
* Replaced some semi-transparent elements with full colors (from the theme), some uses of transparency seemed superfluous

**View:**
20781ViewAnnotated

**Note:**
 Those changes will only change code quality and not accessibility of the interface as long as https://jira.xwiki.org/browse/XWIKI-20680 is not closed.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataDropdownMenu.vue
@@ -129,13 +129,20 @@ export default {
 
 <style>
 
-.livedata-dropdown-menu .btn-default {
-  background-color: #f8f8f8;
-  background-image: none;
-  border-color: #e5e5e5;
-  box-shadow: none;
-  color: #333333;
-  text-shadow: none;
+.livedata-dropdown-menu {
+  // Similar to .flat-buttons(), but there's no access
+  .btn-default {
+    background-color: @breadcrumb-bg;
+    background-image: none;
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -715,7 +715,7 @@
       <dependency>
         <groupId>io.sf.carte</groupId>
         <artifactId>css4j</artifactId>
-        <version>3.9.1</version>
+        <version>4.0</version>
```
manuelleduc, line 718: I guess this is not part of the PR?

Sereza7: When I tried to build (to take screenshots and update), I had an issue, and did a fetch - rebase - push force with lease.
However the two last commits in the history of the PR as of now are still here...

Sereza7:
**View:**
Theme Name | Before |  After
:---------:|:-------------------------:|:-------------------------:
Iceberg | 20781-IcebergBefore | 20781-IcebergAfter
Darkly | 20781-DarklyBefore | 20781-DarklyAfter
Garden | 20781-GardenBefore | 20781-GardenAfter
Kitty | 20781-KittyBefore | 20781-KittyAfter

As we can see on these examples, this PR does not change radically the style of LiveData.
The slight variations are "smart" in that they follow the theme conventions: for Iceberg, Garden and Kitty, we can see that the links on the right of each entry ( Edit & Delete ) are a bit darker after the fix, while they are a bit lighter on Darkly, which is a dark theme. In those 4 cases, it fixes contrasts issues on the table.

PR #2136, opened 2023-03-30
XRENDERING-655: Add the ability to style image captions

manuelleduc:
Jira: https://jira.xwiki.org/browse/XRENDERING-655
Rendering PR: https://github.com/xwiki/xwiki-rendering/pull/258
Forum proposal: https://forum.xwiki.org/t/inject-chain-listeners-from-extensions/12291


xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-style/xwiki-platform-image-style-api/src/main/java/org/xwiki/image/style/internal/rendering/CaptionedImageParseListenerProvider.java
```diff
@@ -0,0 +1,155 @@
...
+ * @version $Id$
+ * @since 15.3RC1
+ * @since 14.10.8
+ */
+@Component
+@Singleton
+@Named("captionedImageParse")
+public class CaptionedImageParseListenerProvider implements ListenerProvider
+{
+    private static final List<Syntax> ACCEPTED_SYNTAX = List.of(XWIKI_2_0, XWIKI_2_1);
```
tmortagne, line 63: A Set might be better since the point is to use `#contains` (but not sure for such a small collection honestly).

manuelleduc: +1 because it's conceptually more correct. Though I agree it won't make much difference in practice.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-image/plugin.js
```diff
@@ -229,14 +238,24 @@
...
          */
         function removeAttribute(widget, key) {
+          widget.element.removeAttribute(key);
           widget.parts.image.removeAttribute(key);
           widget.wrapper.removeAttribute(key);
         }
 
+        function updateFigureDimensions(widget) {
+          if (widget.data.hasCaption) {
+            setAttribute(widget, 'style', "width: " + widget.data.width + "px");
```
michitux, line 248: Won't this override any existing styles on the figure?

manuelleduc: Good catch, thanks. I'll apply the same merge strategy as in the listeners.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-style/xwiki-platform-image-style-api/src/main/java/org/xwiki/image/style/internal/rendering/CaptionedImageRenderListenerProvider.java
```diff
@@ -0,0 +1,117 @@
...
+         */
+        protected InternalChainingListener(ListenerChain listenerChain)
+        {
+            setListenerChain(listenerChain);
+        }
+
+        @Override
+        public void beginFigure(Map<String, String> parameters)
+        {
+            this.cleanedUpParameters = new HashMap<>(parameters);
```
michitux, line 94: I think it would be nicer to use `LinkedHashMap` here to preserve parameter order (this is also nicer for some integration tests).

PR #2139, opened 2023-03-31
XWIKI-20548: Allow choosing the authenticator at runtime

tmortagne:
* introduce a role to expose an authenticator as component
* refactor XWiki#getAuthService() to rely on components by default
* add a section in the authentication administration to
  * show information about the current authenticator
  * allow changing it (when it's not an authenticator set through xwiki.cfg or injected dynamically)
* as a bonus of being configured at wiki level: it's possible to use different auth services on different wikis (unconfigured subwiki inherit main wiki configuration which inherit xwiki.properties configuration)

A backport of the API is provided through https://github.com/xwiki-contrib/authservice-backport because existing authenticators needs to implement this new role, and they are obviously not going to upgrade to 15.3 anytime soon.

Before someone ask: the java part is not in xwiki-platform-security-authentication because this is just here to expose the current API as component, and ideally one day we want to design an entirely new authentication API without any dependency on oldcore classes which and, this time, will be located in by xwiki-platform-security-authentication-api.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -171,6 +171,7 @@ import org.xwiki.resource.entity.EntityResourceReference;
 import org.xwiki.script.ScriptContextManager;
 import org.xwiki.security.authorization.AuthorizationManager;
 import org.xwiki.security.authorization.Right;
+import org.xwiki.security.authservice.internal.AuthServiceManager;
 import org.xwiki.skin.Resource;
 import org.xwiki.skin.Skin;
 import org.xwiki.skin.SkinManager;
@@ -339,6 +340,8 @@ public class XWiki implements EventListener
     private XWikiPluginManager pluginManager;
 
     private XWikiAuthService authService;
+
+    private AuthServiceManager authServices;
 
     private XWikiRightService rightService;
 
@@ -847,6 +850,15 @@ public class XWiki implements EventListener
         }
 
         return this.referenceUpdater;
+    }
+
+    private AuthServiceManager getAuthServiceManager()
+    {
+        if (this.authServices == null) {
+            this.authServices = Utils.getComponent(AuthServiceManager.class);
+        }
+
+        return this.authServices;
     }
 
     private String localizePlainOrKey(String key, Object... parameters)
@@ -5969,7 +5981,7 @@ public class XWiki implements EventListener
             if (isLDAP()) {
                 authClass = "com.xpn.xwiki.user.impl.LDAP.XWikiLDAPAuthServiceImpl";
             } else {
-                authClass = "com.xpn.xwiki.user.impl.xwiki.XWikiAuthServiceImpl";
+                return null;
             }
         }
 
@@ -5997,10 +6009,15 @@ public class XWiki implements EventListener
                 try {
                     Class<? extends XWikiAuthService> authClass = getAuthServiceClass();
 
-                    setAuthService(authClass);
+                    if (authClass != null) {
+                        // Remember the authenticator instance corresponding to the configured class
+                        setAuthService(authClass);
+                    } else {
+                        // Search for the configured component based auth service
+                        return getAuthServiceManager().getAuthService();
+                    }
                 } catch (Exception e) {
-                    LOGGER.warn("Failed to get the configured AuthService class, fallbacking on standard authenticator",
-                        e);
+                    LOGGER.error("Failed to get the configured AuthService, fallbacking on standard authenticator", e);
 
                     this.authService = new XWikiAuthServiceImpl();
 
@@ -6012,6 +6029,16 @@ public class XWiki implements EventListener
 
             return this.authService;
         }
+    }
+
+    /**
+     * @return true if the auth service is a component, false when it's based on the old xwiki.cfg's authclass property
+     *         or if it's directly injected through {@link #setAuthService(Class)}
+     * @since 15.3RC1
+     */
+    public boolean isAuthServiceComponent()
+    {
+        return this.authService == null;
     }
 
     private void setAuthService(Class<? extends XWikiAuthService> authClass)
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/XWikiAuthServiceComponent.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/XWikiAuthServiceComponent.java
@@ -1,0 +1,45 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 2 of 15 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -5997,10 +6009,15 @@ public XWikiAuthService getAuthService()
...
+                        // Remember the authenticator instance corresponding to the configured class
+                        setAuthService(authClass);
+                    } else {
+                        // Search for the configured component based auth service
+                        return getAuthServiceManager().getAuthService();
+                    }
                 } catch (Exception e) {
-                    LOGGER.warn("Failed to get the configured AuthService class, fallbacking on standard authenticator",
-                        e);
+                    LOGGER.error("Failed to get the configured AuthService, fallbacking on standard authenticator", e);
```
manuelleduc, line 6020: What the reason of the change of log level? Was it too low previously?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/XWikiAuthServiceComponent.java
```diff
@@ -0,0 +1,45 @@
...
+ * @version $Id$
+ * @since 15.3RC1
+ */
+@Role
+@Unstable
+public interface XWikiAuthServiceComponent extends XWikiAuthService
+{
+    // Name provided as translation "security.authservice.<id>.name"
+
+    // Description provided as translation "security.authservice.<id>.description"
```
manuelleduc, line 39: Can't this be in some javadoc instead of inline comments?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-ui/src/main/resources/XWiki/Authentication/Administration.xml
```diff
@@ -36,7 +36,73 @@
   <minorEdit>false</minorEdit>
   <syntaxId>xwiki/2.1</syntaxId>
   <hidden>true</hidden>
-  <content>{{velocity}}
+  <content>== {{translation key="security.authservice.admin.title"/}}
```
manuelleduc, line 39: missing closing `==` ? + I think the use of the translation macro in heading is causing a webstandard issue

manuelleduc: iirc because the computation of the heading id is done before the computation of the translation macro, leading to an empty id, which is raised as an error by webstandard.

manuelleduc: see https://github.com/xwiki/xwiki-platform/commit/6ed8244be36cee8caad055fe123b796d7337491a#diff-14c2728e791b12a9ea983da33040f3b43edd6bf16f4b5943d69b5901e5fa0638R106 for instance

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/internal/AuthServiceConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    public static final LocalDocumentReference DOC_REFERENCE = new LocalDocumentReference(SPACES, "Configuration");
+
+    /**
+     * The name of the property containing the identifier of the authenticator in the xwiki.properties file.
+     */
+    public static final String CONFIGURATION_INSTANCE_PROPERTY =
+        "security.authentication." + AuthServiceConfigurationClassInitializer.FIELD_SERVICE;
+
+    private class ServiceCacheEntry
```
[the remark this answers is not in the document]
manuelleduc: I don't get how you do this distinction.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/internal/StandardXWikiAuthServiceComponent.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public String getId()
+    {
+        return ID;
+    }
+
+    @Override
+    public void initialize() throws InitializationException
+    {
+        this.service = new XWikiAuthServiceImpl();
```
manuelleduc, line 64: Is the initialize required here (instead of instantiating `XWikiAuthServiceImpl` line 53)?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-ui/src/main/resources/XWiki/Authentication/Administration.xml
```diff
@@ -36,7 +36,75 @@
...
+    $authService.class.name
+  #end
+{{/box}}
+
+#if ($authServiceType == 'component')
+  #set ($authServices = $services.security.authService.getAuthServices())
+  #if ($authServices.size() &gt; 1)
+    {{html}}
+      &lt;form action="$xwiki.relativeRequestURL" method="post"&gt;
+        &lt;input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /&gt;;
```
manuelleduc, line 80: No need to xml escape here?

manuelleduc: Superflous `;`?
```suggestion
      &lt;form action="$xwiki.relativeRequestURL" method="post"&gt;
        &lt;input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /&gt;
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/internal/AuthServiceConfiguration.java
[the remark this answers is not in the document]
tmortagne: * "we did not cache the value for this wiki yet": the cache does not contain any ServiceCacheEntry for a given wiki
* "this wiki does not have any configured authservice": the cache contains a ServiceCacheEntry for a given wiki, but its value is null

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/internal/StandardXWikiAuthServiceComponent.java
tmortagne: No, not really. I think I used to have more stuff in this initialize and forgot to cleanup.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/security/authservice/internal/AuthServiceConfiguration.java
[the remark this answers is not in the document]
manuelleduc: A right, got it. Unless it is just me, wdyt of expliciting that in the javadoc?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authentication/xwiki-platform-security-authentication-ui/src/main/resources/XWiki/Authentication/Administration.xml
tmortagne: > No need to xml escape here?

Not with the current token format, I guess we could argue that it's more future-proof to escape anyway but note that our code is full of that exact same code.

> Superflous ;?

Looks like it indeed, not sure how it ended up there.

PR #2141, opened 2023-04-03
XWIKI-20812 Removed few quick code smells

SahibYar:
I forked a project today as I wanted to remove few quick code smells, which I found while skimming through different files for understanding.

Following code smells were removed. 
1. Merged few if conditions with `&&` condition
2. Changed the visibility of `public` constructors to `protected` for abstract classes. 
3. Removed few unused parameters.


xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacroRefactoring.java
```diff
@@ -85,22 +85,22 @@ public class IncludeMacroRefactoring implements MacroRefactoring
 
     @Override
     public Optional<MacroBlock> replaceReference(MacroBlock macroBlock, DocumentReference currentDocumentReference,
-        DocumentReference sourceReference, DocumentReference targetReference, boolean relative)
```
SahibYar, line 88: removed unused parameter.

PR #2145, opened 2023-04-07
XCOMMONS-2507: Add a priority concept in components

tmortagne:
* add priority support to wiki components

Depends on https://github.com/xwiki/xwiki-commons/pull/265.


pom.xml
```diff
@@ -111,7 +111,7 @@
     <pgsql.version>42.6.0</pgsql.version>
     <oracle.groupId>com.oracle.database.jdbc</oracle.groupId>
     <oracle.artifactId>ojdbc8</oracle.artifactId>
-    <oracle.version>21.9.0.0</oracle.version>
+    <oracle.version>23.2.0.0</oracle.version>
```
surli, line 114: Shouldn't be part of this PR I think

tmortagne: Yes, I was in the wrong branch. It's in master too.

PR #2149, opened 2023-04-13
XWIKI-16150: Make livetable suggest filters accessible

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-16150
**PR Changes:**
* Added a title to the select elements for filters
* Changed the structure of the date filter element to make it match the combobox pattern.
* Added a description to the date filter element.

_Note:_ The dropdown element on the date filter is from Bootstrap. It would need to pretty much get fully remade in order to pass accessibility standards. Instead, I chose to just make it easier to avoid using this dropdown by providing a proper description of the date format expected in the input field of the combobox.

**View:**
The titles are displayed as tooltips:
16150-PreviewTitles


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -309,8 +309,11 @@
       #set ($filterType = 'text')
     #end
   #end
+  #set ($selectTitle = $escapetool.xml($services.localization.render('platform.livetable.filtersTitle',
```
michitux, line 312: The variable name seems wrong as this is not just used for selects, maybe just `$filterTitle`?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -46,8 +46,8 @@
         var input = $(element);
         var hidden = input.prev('input[type="hidden"]');
         var dateFormat = moment().toMomentFormatString(input.attr('data-dateformat'));
-
         input.daterangepicker({
+          parentEl: "#"+this.parentElement.id,
```
michitux, line 50: Doesn't this id need escaping to be used in the selector?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetabledate.js
```diff
@@ -105,10 +108,22 @@
           }
         });
 
+        input.on('showCalendar.daterangepicker', function(event) {
+          event.target.next('.daterangepicker').setAttribute('style',
+            'display:block; width: max-content;');
```
michitux, line 113: I wonder if this style shouldn't be better placed in some CSS file (but I haven't checked if there is a suitable one including a suitable selector).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3011,6 +3011,7 @@ platform.livetable.asyncActionInProgress=In progress...
 platform.livetable.asyncActionDone=Done
 platform.livetable.asyncActionFailed=Failed
 platform.livetable.filtersTitle=Filter for the {0} column
+platform.livetable.filter.date.format.description=date format: {0} - {0}
```
michitux, line 3014: Is it usual to have such a short description? I would have rather expected something like "Date range filter, enter the date range in the format {0} or use the date range picker.". Also, I would suggest to pass the whole date range format as a single parameter instead of having `{0} - {0}` in the translation as this will confuse translators and translators might mess up the format.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -1522,7 +1522,8 @@ $Msz MB##
     #foreach ($column in $columns)
       #set ($columnProperties = $columnsProperties.get($column))
       ## Note: Even though the column is hidden, we still need a (pottentially hidden) TD to generate valid HTML
-      <td class="xwiki-livetable-display-header-filter #if ("$!columnProperties.type" == 'hidden')hidden#end">
+      <td id="xwiki-livetable-${htmlLiveTableId}-header-filter-${foreach.count}"
```
michitux, line 1525: Is this id really required (I couldn't find where it is used)?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -352,19 +355,23 @@
...
-      #if ("$!columnProperties.size" != '')size="$!escapetool.xml($columnProperties.size)"#end
-      title="$escapetool.xml($services.localization.render('platform.livetable.filtersTitle',
-      [$services.localization.render("${transprefix}$column")]))" />
+    <div role="combobox" id="xwiki-livetable-${htmlLiveTableId}-filter-date-combobox-${foreach.count}"
+      class="xwiki-livetable-display-header-filter-date-combobox" aria-expanded="false">
+      <input id="xwiki-livetable-${htmlLiveTableId}-filter-${foreach.count}" type="text"
+        data-type="date" data-dateformat="$escapetool.xml($dateFormat)"
+        #if ("$!columnProperties.size" != '')size="$!escapetool.xml($columnProperties.size)"#end
+        title="$!selectTitle"
+        aria-description="$!escapetool.xml($services.localization.render('platform.livetable.filter.date.format.description',[$escapetool.xml($dateFormat)]))"/>
```
michitux, line 364: The inner escaping around `$dateFormat` should be removed, or the date format will be escaped twice.

michitux (changes requested):
Overall this looks good but there are some minor issues.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/pull/2149/commits/588a4e553cec58dfd235a1e66c4c0eea3a138c70

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetabledate.js
Sereza7: I'm not sure I understand your point. We get the id from the parentElement. If this id was not escaped before it'd break beforehand. E.g. there can't be any spaces in this id since it's already used as an id.
I think this'd cause double escaping.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetabledate.js
Sereza7: Blocks move a lot here and are generated using bootstrap 3.
If I remember well, I tried just adding a class with style attached to it, but the style would get taken over by bootstrap generating a 'style' attribute inline over it.
This is the only way I found to change this couple of properties from the bootstrap default (that needed to be overridden for accessibility) on this element.
The other style properties I could just do it with regular selectors: see [xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetable.css](https://github.com/xwiki/xwiki-platform/pull/2149/files#diff-8eaa7402bfc75704d30fcf9128bf89987aeff1cc3a38eb20e66ea83ed6b84c53).

____

The time I spent on understanding this and the bad looks of this code is also why I decided to not update the whole bootstrap modal for accessibility. It'd probably be faster coding it from scratch again :D
As of now, Assistive Techs (AT) users have a proper way to select date ranges, even if it's not as nice as the one regular users get.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetabledate.js
michitux: You're passing this as selector in jQuery here I think. The id may contain, e.g., `.`. If you don't escape this, the selector won't match the id but something else. There seems to be https://api.jquery.com/jQuery.escapeSelector/ for this purpose.

michitux (approved):
Looks good apart from the escaping in the jQuery selector.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/table/livetabledate.js
Sereza7: Thanks a lot! 
It's clearer to me now :)
This same structure is also found [here in the xwiki-platform repository](https://github.com/xwiki/xwiki-platform/blob/6006c3c4ae08e592d12de3230a7cc5cb06dd1f1a/xwiki-platform-core/xwiki-platform-edit/xwiki-platform-edit-ui/src/main/resources/XWiki/InplaceEditing.xml#L407).
Fixed in https://github.com/xwiki/xwiki-platform/pull/2149/commits/3e6c82bd0a4ae2ce86c77400708688b72c955618

Sereza7:
Fixed this last comment in https://github.com/xwiki/xwiki-platform/pull/2149/commits/3e6c82bd0a4ae2ce86c77400708688b72c955618 :)

PR #2150, opened 2023-04-13
XWIKI-20907: Introduce the notion of required rights

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20907

- Allow the activation of the required rights
- Allows the definition of the required rights for a document
- Take the required rights into account during right check

**Notes**
- The required rights are current not activated by default

**Future work**
- provide a dashboard to help the activation of required rights on new documents
- activate the required rights on all pages provided by XWiki Standard
- allow to configure different strictness level


xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/model/document/RequiredRights.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @param newRights the new rights to set
+     */
+    void setRights(Set<Right> newRights);
+
+    /**
+     * @param right the right to check
+     * @return {@code true} of the right is part of the required rights, {@code false} otherwise
+     */
+    boolean has(Right right);
```
tmortagne, line 50: Is this really needed with #getRigths ?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
```diff
@@ -75,8 +75,10 @@
   ## EDIT menu
   ##
   #define ($editMenu)
-    #if ($displayEditMenu && $hasEdit && !$isReadOnly)
+    #if($displayEditMenu && $hasEdit && !$isReadOnly)
```
tmortagne, line 78: Looks like a mistake.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/model/internal/document/DefaultRequiredRights.java
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return getRights().contains(right);
+    }
+
+    @Override
+    public void setRights(Set<Right> newRights)
+    {
+        this.requiredRights.clear();
+        this.requiredRights.addAll(newRights);
+        this.document.setMetaDataDirty(true);
```
tmortagne, line 80: Ideally, this should be called only if the rights actually changed.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/model/document/RequiredRights.java
manuelleduc: Strictly, no. But it's making it easier to interact with required rights.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/handler/packager/DefaultXWikiDocumentMerger.java
```diff
@@ -309,6 +313,15 @@ private XWikiDocument askDocumentToSave(XWikiDocument currentDocument, XWikiDocu
             }
         }
 
+        if (mergedDocument != null) {
+            Set<Right> rights = new HashSet<>(mergedDocument.getRequiredRights().getRights());
+            if (currentDocument != null) {
+                rights.addAll(currentDocument.getRequiredRights().getRights());
```
surli, line 319: Not sure to understand this logic? If it's merged already why adding those? Else if it's not merged why don't you consider the next document?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/EditForm.java
```diff
@@ -474,6 +500,26 @@ public List<String> getTemporaryUploadedFiles()
         return temporaryUploadedFiles;
     }
 
+    /**
+     * @return the value of the {@code activateRequiredRights} request parameter
+     * @since 15.4RC1
+     */
+    @Unstable
+    public String getActivateRequiredRights()
```
surli, line 508: Would make more sense for this to return a boolean value and to check there that it's equals to 1 no?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -4158,6 +4164,27 @@ public void readDocMetaFromForm(EditForm eform, XWikiContext context) throws XWi
         if (eform.getHidden() != null) {
             setHidden("1".equals(eform.getHidden()));
         }
+
+        if (eform.getActivateRequiredRights() != null) {
+            if (Objects.equals(eform.getActivateRequiredRights(), "1")) {
+                this.requiredRightsActivated = true;
+                setMetaDataDirty(true);
```
surli, line 4171: The metadata dirty flag should be changed only if the required right flag has changed no?

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-events/xwiki-platform-filter-event-model/src/main/java/org/xwiki/filter/event/model/WikiDocumentFilter.java
```diff
@@ -180,6 +180,18 @@
      */
     String PARAMETER_REVISION_ORIGINALMETADATA_AUTHOR = "revision_original_author";
 
+    /**
+     * @type {@link String}
+     * @since 15.4RC1
+     */
+    String PARAMETER_REVISION_REQUIRED_RIGHTS = "revision_required_rights";
```
tmortagne, line 187: Should probably have `@Unstable`.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-events/xwiki-platform-filter-event-model/src/main/java/org/xwiki/filter/event/model/WikiDocumentFilter.java
```diff
@@ -180,6 +180,18 @@
...
+
+    /**
+     * @type {@link String}
+     * @since 15.4RC1
+     */
+    String PARAMETER_REVISION_REQUIRED_RIGHTS_ACTIVATED = "revision_required_rights_activated";
```
tmortagne, line 193: Should probably have `@Unstable`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -9499,6 +9529,27 @@ public DocumentAuthors getAuthors()
...
+    }
+
+    /**
+     * Set {@link #isMetaDataDirty} if the required rights are changed.
+     *
+     * @param requiredRights the required rights to set
+     * @since 15.3RC1
+     */
+    @Unstable
+    public void setRequiredRights(RequiredRights requiredRights)
```
tmortagne, line 9545: It's not very clear to me how someone can modify the required rights since there is no public setters for them (`DefaultRequiredRights` is internal).

Why don't you simply follow the same logic as DocumentAuthors since it's a very similar use case?

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/DocumentModelBridge.java
```diff
@@ -149,6 +150,16 @@ default DocumentAuthors getAuthors()
...
     }
 
+    /**
+     * @return the required rights of the document
+     * @since 15.4RC1
+     */
+    @Unstable
+    default RequiredRights getRequiredRights()
+    {
+        return null;
```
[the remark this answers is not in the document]
manuelleduc: I've considered it but if we make the assumption that all up to date implementations of `DocumentModelBridge` would never return null, this is complexfying the use of this method for not much added value.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-streams/xwiki-platform-filter-stream-xar/src/main/java/org/xwiki/filter/xar/internal/input/XARInputFilterStream.java
```diff
@@ -36,16 +36,28 @@
...
-import org.xwiki.filter.xar.internal.XARFilterUtils;
 import org.xwiki.filter.xml.input.SourceInputSource;
 import org.xwiki.model.reference.EntityReference;
 
+import static org.xwiki.filter.xar.internal.XARFilterUtils.ROLEHINT_11;
+import static org.xwiki.filter.xar.internal.XARFilterUtils.ROLEHINT_12;
+import static org.xwiki.filter.xar.internal.XARFilterUtils.ROLEHINT_13;
+import static org.xwiki.filter.xar.internal.XARFilterUtils.ROLEHINT_14;
+import static org.xwiki.filter.xar.internal.XARFilterUtils.ROLEHINT_15;
+import static org.xwiki.filter.xar.internal.XARFilterUtils.ROLEHINT_16;
```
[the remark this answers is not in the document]
manuelleduc: I've changed it to improve the readability of the `@Component` below. But I'm find with changing it to the code below if you feel like having a single import is important.

```java
@Component(hints = {
    XARFilterUtils.ROLEHINT_16,
    XARFilterUtils.ROLEHINT_15,
    XARFilterUtils.ROLEHINT_14,
    XARFilterUtils.ROLEHINT_13,
    XARFilterUtils.ROLEHINT_12,
    XARFilterUtils.ROLEHINT_11
})
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
manuelleduc: Because `DocumentAuthors` is only relying on fields of XWikiDocument itself. Whereas, I need to access values of an external table.

tmortagne: > Because DocumentAuthors is only relying on fields of XWikiDocument itself. Whereas, I need to access values of an external table.

I don't see the relationship with making RequiredRights readonly.

In any case, the current API makes it impossible to set required rights from what I understand.

tmortagne: OK, this whole conversation does not make any sense because I missed RequiredRights#setRights.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/bridge/DocumentModelBridge.java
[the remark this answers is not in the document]
surli: You cannot provide an empty required rights instead of null then? like a stub implem?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/model/internal/document/DefaultRequiredRights.java
```diff
@@ -0,0 +1,126 @@
...
+    {
+        return this.document.isRequiredRightsActivated();
+    }
+
+    @Override
+    public void setRights(Set<Right> newRights)
+    {
+        boolean hasChanged = !Objects.equals(newRights, this.requiredRights);
+        if (hasChanged) {
+            this.requiredRights.clear();
```
tmortagne, line 87: This logic is not going to work so well from concurrency point of view. The safest IMO would be to create a new HashSet and change `this.requiredRights` with it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/EditForm.java
```diff
@@ -461,6 +474,24 @@ private void setTemporaryUploadedFiles(String[] temporaryUploadedFiles)
         }
     }
 
+    private void setActivateRequiredRights(String activateRequiredRights)
+    {
+        if (activateRequiredRights != null) {
+            this.activateRequiredRights = Objects.equals(activateRequiredRights, "1");
+        } else {
+            this.activateRequiredRights = null;
```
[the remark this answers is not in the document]
tmortagne: If you need a distinction, it seems to me a Boolean would make more sense than a String (null, `true`, `false`).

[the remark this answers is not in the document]
manuelleduc: `this.activateRequiredRights` is a `Boolean`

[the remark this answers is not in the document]
tmortagne: Right, sorry. I thought method this was a setter, but it's actually a converter.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/java/org/xwiki/extension/xar/internal/handler/packager/DefaultXWikiDocumentMerger.java
manuelleduc: Merge moved to `org.xwiki.store.merge.internal.DefaultMergeManager#mergeDocument`

xwiki-platform-core/pom.xml
```diff
@@ -229,6 +229,7 @@
     <module>xwiki-platform-release</module>
     <module>xwiki-platform-rendering</module>
     <module>xwiki-platform-repository</module>
+    <module>xwiki-platform-requiredrights</module>
```
surli, line 232: Sorry if it has been discussed in the past, but feels a bit weird that it has its own root module: I'd have expected either to be in the security-authorization module, or if it really needs its own module, to be a submodule of security.

xwiki-platform-core/xwiki-platform-bridge/src/main/java/org/xwiki/model/document/RequiredRights.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @return {@code true} when the required rights are activated for the document
```
surli, line 53: Would be to good to explain in the javadoc of the interface what's an "activated" right. It's not obvious when reading it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateStore.java
```diff
@@ -1898,6 +1929,33 @@ private void saveAttachmentList(XWikiDocument doc, XWikiContext context) throws
         }
     }
 
+    private void saveRequiredRights(XWikiDocument doc, XWikiContext inputxcontext) throws XWikiException
+    {
+        XWikiContext context = getExecutionXContext(inputxcontext, true);
+
+        // Extract the links
```
surli, line 1936: Sounds like copy/pasted comments

xwiki-platform-core/pom.xml
manuelleduc: Is has not been discussed. I've placed it here but I'm not against something like `xwiki-platform-security-requiredrights`

michitux (changes requested):
I think the script services need to be fundamentally reconsidered as otherwise, required rights don't provide any protection for script to programming right escalation.

Further, `saveAsAuthor()` needs to be updated to make sure that a script cannot just save another document to escape from script to programming right again.

When `AuthorExecutor` is used, in particular updated required rights may be lost.

I wonder how this whole concept of required rights interferes with asynchronous execution. This involves several questions:
1. Are required rights correctly considered when, e.g., the async macro is used?
2. The async context stores right evaluations and their results. This seems incompatible with this special context property to disable required rights.

When a document contains an XObject like a UIX that has wiki scope, the code for loading them currently checks if the last author has admin right on the wiki. Even if we don't introduce a way to limit this with required rights, we should keep the current system extensible to allow this. With the APIs that are introduced here, this would require a lot of changes, again. As a lot of code assumes that wiki admin right implies script right, this might also be an easy way to work around required rights.

To me there are two different dimensions to consider when thinking about script/programming/admin rights:

1. The user scope where the object is used. For example, is this UIX just for its author or for all users?
2. The page/space/wiki scope: Is this JSX just for the current document or for the whole wiki?

I could for example imagine that we could allow users to write JSX that are just executed for them even on the whole wiki without any security impact. Of course, we could say that such a JSX just doesn't require script right in this case. However, we have a real issue with admin right as having admin right on a document is not the same as having admin right on the wiki but we have [existing](https://github.com/xwiki/xwiki-platform/blob/301a1e0143eebb8dfc3f860c2c504847f7855421/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm#L196) [checks](https://github.com/xwiki/xwiki-platform/blob/e3b462dcdd5949282c673b9b73a7fe82e6b371aa/xwiki-platform-core/xwiki-platform-uiextension/xwiki-platform-uiextension-api/src/main/java/org/xwiki/uiextension/internal/WikiUIExtensionComponentBuilder.java#L105) for admin right on the wiki that are checking for admin right of the last author which should be restrictable with required rights imho. Therefore, I think we need a way to store required rights for different scopes - to cover everything probably page, space and wiki.

I also don't like this hack with the context property for skipping required rights. Ideally, I would like to have a new API to query document/document author rights, i.e., the rights that the last content/effective metadata author has, as restricted by required rights. The contextual authorization manager would use this for script and programming right. The standard authorization manager, on the other hand, won't consider required rights. However, the problem is that this is incompatible with the existing uses of the (standard) authorization manager, e.g., for JSX or UIX. So maybe the best option would be to introduce two new authorization managers: one for user rights (without required rights) and one for document rights (with required rights). We could migrate the APIs to use `UserReference` at the same time which would justify the introduction of a new API imho.

In the context of the security cache, it seems unfortunate that it stores required rights in every access entry even though the required rights don't depend on the user for which rights are queried. I wonder if the required rights couldn't be stored as `SecurityRule` instance on the documents. Unfortunately, this won't solve the need for having security access entries both with and without required rights applied. Maybe the best idea in the end is to keep required rights out of the security cache/just in the security rules (?) and to just apply the required rights in the new document rights authorization manager?

PR #2153, opened 2023-04-17
XWIKI-17034: Allow to define different grouping strategy for notifications

surli:
The purpose of this PR is to provide the full architecture for having different strategies for grouping notifications depending on users' choices and on the kind of output target (e.g. email, alert, rss, etc).
It contains the new roles for defining a grouping strategy and for loading the strategy component depending on the user and the target. By default the grouping strategy is the one that was already existing and based on on the similarity algorithm. Another strategy is already provided only based on the type of events.

In terms of architectural change the idea is that now the API would always return by default a `List<Event>` instead of a `List<CompositeEvent>` and depending on the purpose of this list of events, a different grouping strategy would be used: e.g. if it's for displaying the notifications in the alert area it would use the strategy for alert, and if it's for email notifications it would use the email grouping strategy. 
Those strategies are handled with a new xobject stored as a user preference. 

I finally decided to keep an API that would return a `List<CompositeEvent>` in `ParametrizedNotificationManager` without making it deprecated and that would ensure to return the maximum number of composite events, respecting the requested number of events in `NotificationParameters#expectedCount`: the idea is to keep using that API for now in the UI of XS, so that the user would keep obtain 20 composite events in the UI, and note 20 events  that could be grouped in 2 or 10 composite events, which wouldn't be nice. 
Then I also provided distinct caches to ensure to be able to cache either
the composite events, or the events, and to distinguish their count.

I also decided to deprecate `NotificationManager` in favor of `ParametrizedNotificationManager`: despite the bad name, this API is far more easy to use and to maintain on the long run.


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-api/src/main/java/org/xwiki/notifications/CompositeEvent.java
```diff
@@ -84,21 +84,36 @@ public List<String> getEventIds()
...
         }
         similarityBetweenEvents = similarity;
+        this.add(event);
+    }
+
+    /**
+     * Add an event to the composite event and order it based on the dates.
+     *
+     * @param event the event to add
+     */
```
tmortagne, line 104: Missing `@since`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/NotificationManager.java
```diff
@@ -108,13 +123,37 @@ List<CompositeEvent> getEvents(String userId, NotificationFormat format, int exp
...
+     * @param untilDate do not return events happened after this date
+     * @param untilDateIncluded true if the passed {@code untilDate} should be included
+     * @param fromDate do not return events happened before this date
+     * @param blackList list of ids of blacklisted events to not return (to not get already known events again)
+     * @return the matching events for the user, could be less than expectedCount but not more
+     * @throws NotificationException if error happens
+     * @since 15.4RC1
+     */
+    @Unstable
+    default List<Event> getRawEvents(String userId, NotificationFormat format, int expectedCount, Date untilDate,
```
tmortagne, line 151: I'm wondering if it's not a good occasion to pass some kind of configuration instead of so many parameters, since it's a lot more future-proof (if we need more later).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/ParametrizedNotificationManager.java
```diff
@@ -19,11 +19,14 @@
  */
```
tmortagne, line 19: I noticed only now that there is a typo in the name of that API, but it's a bit late now...

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/NotificationManager.java
surli: Using NotificationParameters maybe?

tmortagne: I don't know it very well, but if it already carries the same information, then why not yes.

surli: Yes it does, I'm not a huge fan of it, because it doesn't really have accessor: the fields are public. I'm wondering if we shouldn't break it once for all to set accessors...

tmortagne: > it doesn't really have accessor: the fields are public

Indeed, strange one.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/NotificationManager.java
```diff
@@ -46,6 +48,7 @@
      * @throws NotificationException if error happens
      * @since 10.1RC1
      */
+    @Deprecated(since = "15.4RC1")
```
tmortagne, line 51: Not needed, you deprecated the whole interface.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-sources/src/main/java/org/xwiki/notifications/sources/NotificationManager.java
```diff
@@ -74,6 +77,7 @@ List<CompositeEvent> getEvents(String userId, int expectedCount, Date untilDate,
...
```
tmortagne, line 80: Not needed, you deprecated the whole interface.

PR #2154, opened 2023-04-18
XWIKI-20853: Page index display the effective metadata author and not the original metadata author

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-20853

  * Use the original metadata author for the information displayed as author in the documents LT
  * Solution inspired from what we already have to displayed the original metadata author in the contentheader
  * Specific handling for displaying unknown user in case of guest user as there's a test to cover this


xwiki-platform-core/xwiki-platform-livetable/xwiki-platform-livetable-ui/src/main/resources/XWiki/LiveTableResultsMacros.xml
```diff
@@ -439,14 +439,24 @@
...
-      #set($discard = $row.put('doc_author_url', $xwiki.getURL($translatedDoc.author)))
       #set($discard = $row.put('doc_date', $xwiki.formatDate($translatedDoc.date)))
       #set($discard = $row.put('doc_title', $translatedDoc.plainTitle))
       #set($rawTitle = $translatedDoc.title)
       #if($rawTitle != $row['doc_title'])
         #set($discard = $row.put('doc_title_raw', $rawTitle))
       #end
-      #set($discard = $row.put('doc_author', $xwiki.getPlainUserName($translatedDoc.authorReference)))
+      ## Note that this serialization step is mandatory and we cannot count on the UserReferenceConverter here
+      ## because UserReference#toString does not rely on the serializer so it's not the expected format for
```
tmortagne, line 449: I don't understand the relationship with UserReference#toString, why don't you simply pass the reference to #getPlainUserName (and it will be converted) ?

surli: Hmmm I put the comment we already had in contentheader.vm but there I was using `getUserName` which takes a String as parameter and here I'm using `getPlainUserName` which takes a DocumentReference so indeed I might not need this.

tmortagne: Even if it was taking a String, you would have a real convertion to String and not a toString() call

mflorea (approved):

manuelleduc (approved):
I'm never against a page test for this kind of change. Besides, it looks good, thanks!

PR #2156, opened 2023-04-24
XWIKI-20868: UserReference cannot be converted to DocumentReference in Velocity

michitux:
* Add support for converting UserReference to DocumentReference
* Add a unit test

Jira issue: https://jira.xwiki.org/browse/XWIKI-20868

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/converter/AbstractUserReferenceConverter.java
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/converter/AbstractUserReferenceConverter.java
@@ -24,6 +24,8 @@ import java.lang.reflect.Type;
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import org.xwiki.component.util.ReflectionUtils;
+import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.properties.converter.AbstractConverter;
 import org.xwiki.user.UserReference;
 import org.xwiki.user.UserReferenceResolver;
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/converter/AbstractUserReferenceConverter.java
```diff
@@ -55,6 +61,24 @@ protected UserReference convertToType(Type targetType, Object value)
...
 
+    @Override
+    @SuppressWarnings("unchecked")
+    public <G> G convert(Type targetType, Object sourceValue)
+    {
+        G result;
+
+        // Special handling for target type DocumentReference to support converting UserReference to DocumentReference.
+        if (targetType != null && ReflectionUtils.getTypeClass(targetType).equals(DocumentReference.class)
+            && sourceValue instanceof UserReference)
```
tmortagne, line 72: It's extremely unlikely that this converter is called if none of the source or the target is a UserReference, but it does not harm I guess.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/converter/AbstractUserReferenceConverter.java
```diff
@@ -55,6 +61,24 @@ protected UserReference convertToType(Type targetType, Object value)
...
     }
 
+    @Override
+    @SuppressWarnings("unchecked")
+    public <G> G convert(Type targetType, Object sourceValue)
+    {
+        G result;
+
+        // Special handling for target type DocumentReference to support converting UserReference to DocumentReference.
+        if (targetType != null && ReflectionUtils.getTypeClass(targetType).equals(DocumentReference.class)
```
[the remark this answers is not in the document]
michitux: Seems so, it works and a short search confirms this.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-default/src/main/java/org/xwiki/user/internal/converter/AbstractUserReferenceConverter.java
michitux: I prefer leaving this as it avoids that the cast in the next line fails.

PR #2157, opened 2023-04-25
XWIKI-20843: Application Panel "More Applications" dropdown is not accessible

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20843
**PR Changes:**
* Replaced the anchor for _more applications_ in the Applications panel with a button.
* Updated style
* Added semantics to the disclosure

**View:**
Before PR:
20843-Before
After PR:
20843-After
The button has the same style when the disclosure is closed.

**Demo:** (made before the last changes on style, but still very similar even on style)
https://up1.xwikisas.com/#NdrhBiSPlXXJNnwWHWfPtA

**Notes:**
* This section is visible only by admins and not regular users.
* For regular users, the section appears and then gets hidden. It was not very noticeable but with a blue button it's way easier to notice this behavior. I couldn't find the code responsible for hiding the section so I didn't fix this. 
https://up1.xwikisas.com/#504lX2oFtkkPm8M9w-L7XQ

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
@@ -240,11 +240,11 @@
 &lt;/ul&gt;
 &lt;ul class="applicationsPanel applicationsPanelMoreList nav nav-pills nav-stacked"&gt;
   &lt;li&gt;
-    &lt;a href="$xwiki.getURL($services.model.createDocumentReference('', 'XWiki', 'XWikiPreferences'), 'admin',
-        'editor=globaladmin&amp;section=XWiki.AddExtensions')" class="applicationPanelMoreButton" title="$services.localization.render('panels.applications.more')"&gt;
+    &lt;button class="applicationPanelMoreButton btn btn-default"
+      title="$services.localization.render('panels.applications.more')" aria-expanded='false' &gt;
       &lt;span class="application-img"&gt;$services.icon.renderHTML('add')&lt;/span&gt;
       &lt;span class="application-label"&gt;$services.localization.render('panels.applications.more')&lt;/span&gt;
-    &lt;/a&gt;
+    &lt;/button&gt;
     &lt;div class="applicationPanelMoreContainer hidden"&gt;
       &lt;ul class="nav nav-pills nav-stacked"&gt;
         #displayAppPanelEntries('org.xwiki.platform.panels.Applications.more')
@@ -366,6 +366,7 @@ panels.MoreApplicationsButtonListener = Class.create({
     var moreAppsCount = this.container.select('.application-label').length;
     if (moreAppsCount &gt; 0) {
       this.container.id = 'applicationPanelMoreContainer' + Math.floor(Math.random()*101);
+      this.button.setAttribute('aria-controls',this.container.id);
       this.isToggling = false;
       this.button.observe('click', function(e) {
         e.stop();
@@ -389,6 +390,7 @@ panels.MoreApplicationsButtonListener = Class.create({
         duration: 0.1,
         afterFinish: function() {
           this.isToggling = false;
+          this.button.setAttribute('aria-expanded','true');
         }.bind(this)
       });
     } else {
@@ -399,6 +401,7 @@ panels.MoreApplicationsButtonListener = Class.create({
         afterFinish: function() {
           this.container.addClassName('hidden');
           this.isToggling = false;
+          this.button.setAttribute('aria-expanded','false');
         }.bind(this)
       });
     }
@@ -613,20 +616,6 @@ require(['scriptaculous/effects'], function() {
   padding: 0;
 }
 
-.panel-width-Small .applicationPanelMoreContainer ul {
-  padding: 0;
-  margin: 0;
-  background-color: $theme.pageContentBackgroundColor;
-}
-
-.panel-width-Small .applicationPanelMoreContainer ul li:hover {
-  background-color: $theme.highlightColor;
-}
-
-.panel-width-Small .applicationPanelMoreContainer ul li:hover a {
-  color: $theme.textPrimaryColor;
-}
-
 .panel-width-Small .applicationPanelMoreContainer > ul > li > a {
   justify-content: center;
 }
@@ -693,15 +682,6 @@ require(['scriptaculous/effects'], function() {
 }
 
 /* for colibri */
-
-.skin-colibri .applicationPanelMoreContainer ul {
-  margin-left: 0;
-}
-
-.skin-colibri .panel-width-Medium .applicationPanelMoreContainer ul,
-.skin-colibri .panel-width-Large .applicationPanelMoreContainer ul {
-  margin-left: 20px;
-}
 
 .skin-colibri .panel-width-Small .applicationsPanel {
   margin: 0;
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
@@ -562,7 +562,7 @@ require(['scriptaculous/effects'], function() {
 .applicationPanelMoreButton .application-label{
   white-space: normal;
   text-align: left;
-  display: block;
+  display: inline;
   margin-left: 20px;
 }
 
@@ -628,6 +628,11 @@ require(['scriptaculous/effects'], function() {
 
 .panel-width-Small .applicationPanelMoreContainer > ul > li > a {
   justify-content: center;
+}
+
+.panel-width-Small .applicationPanelMoreContainer ul {
+  padding: 0;
+  margin: 0;
 }
 
 .panel-width-Small .application-img {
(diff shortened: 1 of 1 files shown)


Sereza7:
@michitux 
> I think the button is quite a style change and imho it has way too much visual weight. 

Okay for the style change, I'll reduce the weight 'artificially' with specific CSS rules to make it more similar to a link

> Wouldn't this kind of element be a perfect application of the [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) HTML element which should have very good built-in accessibility that wouldn't even need the JavaScript code for the toggle?

As seen in this [detailed blogpost from an accessibility specialist](https://www.scottohara.me//blog/2022/09/12/details-summary.html) published in September 2022, there is lots of pros to using details, but also lots of cons especially concerning assistive tech support. I checked out [the ARIA Authoring Practice Guide](https://www.w3.org/WAI/ARIA/apg/patterns/) and they don't mention \<details> in any of their patterns.

> 
> Learn how the \<summary> element is exposed to AT:
> 
> Depending on the screen reader and browser pairing you may be using, the \<summary> element would have been exposed in one of the following ways:
> 
> "Disclosure Triangle" with Narrator, VoiceOver and TalkBack when paired with Edge/Chrome.
> "Button" with NVDA when paired with Firefox, Edge or Chrome.
> "Summary" with VoiceOver when paired with Firefox or Safari.
> Bugged behavior or no role announced with TalkBack with Firefox or iOS VO with Safari, respectively.

From [Scott O'Hara's blog](https://www.scottohara.me//blog/2022/09/12/details-summary.html)

So I decided to go with the recommended solution from ARIA APG and [just use a button to toggle the dropdown](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-faq/). Using this standard pattern ensures that we shouldn't be too affected by changes in the ARIA specs and shouldn't have too much technical debt.

I can't say for certain what solution will create the least technical debt (maybe in a couple months all web browsers are going to have good support for \<details> and it'll be the new recommendation!... but it's been already a few years :S). It seems to me that using a button should be fine on short and middle time scale.
I think both would be correct on a WCAG standpoint (as in, we do fulfill our part of the job in making xwiki content accessible), but one is definitely better for assistive technologies users.

At last, if the button pattern does become obsolete on the long term, it shouldn't be too difficult to replace.

michitux:
Okay, I didn't expect these problems with `<details>`. I agree with using a button to toggle the links.

Regarding the appearance, I'm not totally opposed to making this a button visually, but then please a button more like the "Create" button in the content menu and not a blue button.

You could also open a proposal to get some more opinions.

Sereza7:
@michitux 
I created a forum proposal for this topic :) 
https://forum.xwiki.org/t/ui-application-panel-dropdown-toggler/12602

Sereza7:
## PR Changes
* Updated style to keep the link presentation

## View
20843-NonAccessibilityIcebergView
20843-AccesssibleIcebergView


## TODO

Clean up the javascript extension, there is some code that's left unused as of now...

Sereza7:
## PR Changes
* Updated style
* Removed unused style classes

## View
Iceberg - Collapsed menu (hover)
20843-IcebergCollapsed
Iceberg - Expanded menu 
20843-IcebergExpanded
Darkly - Expanded menu
20843-Darkly

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -553,13 +550,34 @@ require(['scriptaculous/effects'], function() {
...
+}
+
+.applicationPanelMoreButton:hover {
+  background-color: $theme.highlightColor;
 }
 
-.applicationPanelMoreContainer li a {
-  font-size: 90%;
-  line-height: 16px;
+.applicationPanelMoreButton .application-label{
```
michitux, line 567: I think we normally put a space before `{`.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -553,13 +550,34 @@ require(['scriptaculous/effects'], function() {
...
+  display: inline;
+  margin-left: 20px;
+}
+
+.panel-width-Small .applicationPanelMoreButton {
+  flex-direction: column;
+  padding: 0;
+}
+
+.panel-width-Small .applicationPanelMoreButton .application-label{
```
michitux, line 579: Same as above, missing space before `{`.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -232,21 +232,22 @@
 #panelheader($services.localization.render('panels.applications.title'))
 
 {{html}}
-&lt;ul class="applicationsPanel nav nav-pills nav-stacked"&gt;
```
michitux, line 235: Why did you remove these classes?

michitux (changes requested):
Some small comments. Is it correct that this shouldn't change the appearance now, at least not when hovering/focusing the button?

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/pull/2157/commits/96471a12f760ccb90ac33bfa00f3c66e98b97aef

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/pull/2157/commits/96471a12f760ccb90ac33bfa00f3c66e98b97aef

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
Sereza7: Because removing them did not change significantly the presentation, and they brought an unexpected style from bootstrap 3 that would make the color match badly on some color themes. 
On Darkly, the child anchors were just white instead of being green like any other anchor in the panels, and almost anywhere on the page vv
20843-removeBootstrapNav
Moreover I think this reduces the dependency on Bootstrap 3, which is a good thing if we want to move on from it someday. :)

Sereza7:
@michitux The button had a background style change for :hover, however I did forget about the :focus pseudo class.
This is fixed in https://github.com/xwiki/xwiki-platform/pull/2157/commits/01170f804ad3240cb42327161f4b2fe3eccd5310 .
Here is an updated demo of this PR: https://up1.xwikisas.com/#Z82BiviqdKeiKakqsd625w

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
michitux: Looking at this screenshot, the indentation of the "More applications" seems changed. Previously, this was aligned with the other entries like "Dashboard". Why has this alignment changed?

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -232,21 +232,22 @@
...
   #set ($configDoc = $xwiki.getDocument($services.model.createDocumentReference('', 'PanelsCode',
     'ApplicationsPanelConfiguration')))
   #displayAppPanelEntries('org.xwiki.platform.panels.Applications' $configDoc)
 &lt;/ul&gt;
-&lt;ul class="applicationsPanel applicationsPanelMoreList nav nav-pills nav-stacked"&gt;
+&lt;ul class="applicationsPanel applicationsPanelMoreList nav"&gt;
   &lt;li&gt;
-    &lt;a href="$xwiki.getURL($services.model.createDocumentReference('', 'XWiki', 'XWikiPreferences'), 'admin',
-        'editor=globaladmin&amp;section=XWiki.AddExtensions')" class="applicationPanelMoreButton" title="$services.localization.render('panels.applications.more')"&gt;
+    #set ($panelId = $doc.getXDOM().getIdGenerator().generateUniqueId('applicationPanelMoreContainer',''))
```
michitux, line 243: I hadn't noticed that before. This seems wrong. `$doc.getXDOM()` returns a clone of a cached XDOM instance, I don't think it makes sense to use its id generator.

Sereza7: @michitux 
I'm not sure this is possible to keep the XDOM object through different calls of this panel.
I can't find how to access a proper IdGenerator here.
Instead, we could append a random alphanumeric string at the end of the string to make it 'unique'.

Change proposed in https://github.com/xwiki/xwiki-platform/pull/2157/commits/926c9109b2023237c74aa4d740fe489dcdd7d2e3
20843-UniqueID
Note that this is very similar to the solution used before (10 alphanumeric instead of a number from 1 to 100, but the concept is the same).

PR #2151, opened 2023-04-14
[Misc] Fixed formFields structure in functional tests

Sereza7:
* The HTML wasn't valid. The region was opened with \<dd> and closed with \<dl>.

**WCAG report that highlights this typo in tests:** 

> WCAG warnings in the test class [org.xwiki.export.pdf.test.ui.AllITs]:
> Validation in the test method [formFields]
> Check for [org.xwiki.test.ui.po.ViewPage] at [http://xwikiweb:8080/xwiki/bin/view/PDFExportIT/FormFields].
> \<dt> and \<dd> elements must be contained by a \<dl>
> Description: Ensures \<dt> and \<dd> elements are contained by a \<dl>
> Help URL: https://dequeuniversity.com/rules/axe/4.6/dlitem?application=axeAPI
> Impact: serious
> Tags: cat.structure, wcag2a, wcag131
> HTML element: 
> 	\<dd>
> 	\</dd>
> Selector: [dd:nth-child(1)]
> Fix any of the following:
>   Description list item does not have a \<dl> parent element

PR #2158, opened 2023-04-25
[Misc]: Fixing export PDF functional test

Sereza7:
Fixing regression introduced with https://github.com/xwiki/xwiki-platform/pull/2151
Testing logs (after PR): https://up1.xwikisas.com/#LGC2mQlGjQ5xIGzu9cUj8w

### DIFF
--- a/xwiki-platform-core/xwiki-platform-export/xwiki-platform-export-pdf/xwiki-platform-export-pdf-test/xwiki-platform-export-pdf-test-docker/src/test/it/org/xwiki/export/pdf/test/ui/PDFExportIT.java
+++ b/xwiki-platform-core/xwiki-platform-export/xwiki-platform-export-pdf/xwiki-platform-export-pdf-test/xwiki-platform-export-pdf-test-docker/src/test/it/org/xwiki/export/pdf/test/ui/PDFExportIT.java
@@ -607,9 +607,10 @@ class PDFExportIT
             assertEquals(2, pdf.getNumberOfPages());
 
             String content = pdf.getTextFromPage(1);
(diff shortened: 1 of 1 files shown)


michitux:
Do you have any explanation why the expected result should change?

Sereza7:
> Do you have any explanation why the expected result should change?

The fact that the result changed is explainable: Before, because of the error in the content, the different fields were all independant. So we had Title->value for each field. However now they are all part of the same dl so I guess all values are displayed and all labels next? I'm not sure this makes sense and/or this changes anything on the display (it looks bad semantically though).

Here are the displays of the page without and with the fix from the PR. We can see that the first one lacks in structure somewhere and that the displays for both of those should be different.

miscDDinsteadOfDL
FixedDDinsteadOfDL

Sereza7:
Here are the results of the exports (before and after PR):
[FormFieldsWithoutPR.pdf](https://github.com/xwiki/xwiki-platform/files/11323764/FormFieldsWithoutPR.pdf)
[FormFieldsWithPR.pdf](https://github.com/xwiki/xwiki-platform/files/11323763/FormFieldsWithPR.pdf)

mflorea:
> The fact that the result changed is explainable: Before, because of the error in the content, the different fields were all independant. So we had Title->value for each field. However now they are all part of the same dl so I guess all values are displayed and all labels next? I'm not sure this makes sense and/or this changes anything on the display (it looks bad semantically though).

It has to do with:
* the way the definition list is translated into PDF (by Chrome at least)
* the way the ``PDFTextStripper`` (from pdfbox) extracts the text from the PDF page

We can't do much about it (besides providing a valid HTML of course) than checking what ``PDFTextStripper`` gives us and assuming that is the right expected text.

mflorea (approved):

PR #2160, opened 2023-04-26
XWIKI-20870: Add support for the anchor in EntityResourceReference and EntityResourceReference resolvers

tmortagne:
* add the concept of anchor to EntityResourceReference
* add the support for extracting the anchor in EntityResourceReference resolvers

### DIFF
--- a/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-default/src/main/java/org/xwiki/resource/entity/EntityResourceReference.java
+++ b/xwiki-platform-core/xwiki-platform-resource/xwiki-platform-resource-default/src/main/java/org/xwiki/resource/entity/EntityResourceReference.java
@@ -27,6 +27,7 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.xwiki.model.reference.EntityReference;
 import org.xwiki.resource.AbstractResourceReference;
 import org.xwiki.resource.ResourceType;
+import org.xwiki.stability.Unstable;
 import org.xwiki.text.XWikiToStringBuilder;
 
 /**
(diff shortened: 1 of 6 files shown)


vmassol:
In term of terminology I don't think it's an anchor but a fragment. See https://docs.oracle.com/javase/7/docs/api/java/net/URL.html and https://www.ietf.org/rfc/rfc2396.txt

michitux:
The modern URL standard that is used by browsers agrees, this is called "fragment": https://url.spec.whatwg.org/#concept-url-fragment

vmassol (approved):
Apart from the terminology, for which we need to decide between "anchor" and "fragment" , it sounds good. Thanks.

tmortagne:
> In term of terminology I don't think it's an anchor but a fragment. See [docs.oracle.com/javase/7/docs/api/java/net/URL.html](https://docs.oracle.com/javase/7/docs/api/java/net/URL.html) and [ietf.org/rfc/rfc2396.txt](https://www.ietf.org/rfc/rfc2396.txt)

In the URL specification yes here it's only for  EntityResourceReference which is not a URL but an entity so I kept the same naming that we use in in other areas.

vmassol:
It's true that we have used "anchor" in the past for https://github.com/xwiki/xwiki-rendering/blob/26bc8eb7f28a47775fe092edd46547efb384498c/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/reference/DocumentResourceReference.java#L74 (doesn't mean it's right)

tmortagne:
> It's true that we have used "anchor" in the past for [xwiki/xwiki-rendering@26bc8eb/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/reference/DocumentResourceReference.java#L74](https://github.com/xwiki/xwiki-rendering/blob/26bc8eb7f28a47775fe092edd46547efb384498c/xwiki-rendering-api/src/main/java/org/xwiki/rendering/listener/reference/DocumentResourceReference.java#L74)

That's also the naming we use in the URL factory and most other places that maybe manipulate this (XWiki#getURL, etc.).

vmassol:
ok, fine with me in the end (mostly for consistency and historical reasons).

michitux:
From what I saw, the term anchor was used indeed in HTML 4 but doesn't seem to be used in the modern HTML standard. But :+1: to keep this consistent if we already use the term.

PR #2142, opened 2023-04-05
XWIKI-16140: Make quick search suggestions accessible

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-16140
**PR Changes:**
- Set the search form as the quick search modal parent. 
- Changed the clickable divs in the quick search menu to buttons. 
- Increased contrast on a .css muted-text (hide quick search)

**View:**
The DOM tree before (right) and after (left) this PR
16140-UpdatedDOM
Showcase of keyboard navigation:
https://up1.xwikisas.com/#1X2CBROxBl50VKazoiTyeg

PR #2163, opened 2023-04-28
XWIKI-20733: Top navigation search section has two tab indexes

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20733
**PR Changes:**
* Added the aria attributes where they are needed, according to [this disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/)
* Changed the DOM order of the form elements [to be more consistent with the rest of the right side navbar](https://up1.xwikisas.com/#8JYJK7Wzh50u3RZr3d5VnA).
* Set the type of the text input to hidden when it's hidden visually.
* Updated the style of the quick search form

**Demo:** https://up1.xwikisas.com/#dADjEwhfi470Drx2G8Fcgg
**Note:** Some of those changes are beyond what the ticket described, but they concern accessibility of the search bar. They do not overlap with [XWIKI-16140](https://jira.xwiki.org/browse/XWIKI-16140) but are complementary.


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
```diff
@@ -246,12 +265,13 @@
...
-        &lt;label class="hidden" for="headerglobalsearchinput"&gt;$services.localization.render('search')&lt;/label&gt;
-        &lt;input type="text" name="text" placeholder="$services.localization.render('panels.search.inputText')" id="headerglobalsearchinput" /&gt;
-        &lt;button type="submit" class="btn" title="$services.localization.render('search')"&gt;$services.icon.renderHTML('search')&lt;/button&gt;
-      &lt;/form&gt;
+      &lt;form class='navbar-form globalsearch globalsearch-close form-inline' id='globalsearch' action='$searchURL'
+        role='search'&gt;
+        &lt;label class='hidden' for='headerglobalsearchinput'&gt;$services.localization.render('search')&lt;/label&gt;
+        &lt;button type='submit' class='btn' title="$services.localization.render('search')"
+        aria-expanded='false' aria-controls='headerglobalsearchinput'&gt;$services.icon.renderHTML('search')&lt;/button&gt;
+        &lt;input type='hidden' name='text' placeholder="$services.localization.render('panels.search.inputText')" id='headerglobalsearchinput' /&gt;
```
michitux, line 273: Is it really a good idea to put the input behind the button in the HTML and thus also in the tab order? So the logic is that the other elements are also reversed in tab order so this should be reversed, too? I'm not sure I'm convinced by this, I would rather prefer undoing the reversal of the other elements. Also, blind people won't notice the general reversal of the UI elements and might still expect the submit button after the input.

Sereza7: I see two usability shortcuts here:
1. The submit button should be put after the inputs in the form, so that after filling up the form, the user can just tab and submit it.
2. The dropdown activation button should be put before the dropdown content, so that after opening the dropdown, the user can quickly tab to the content.

In our situation, both function are upheld by the same element:
When the dropdown is closed: onclick -> open dropdown
When the dropdown is opened: onclick -> if the input is empty: close dropdown if the input is full: submit form.

I decided to change the tab order since:
1. It aligns with the other elements on this line of controls: the DOM goes from right to left.
2. It favors the dropdown interaction.

I think keeping the dropdown interaction is more relevant since:
1. Submitting the form can be done by pressing the return key from the input 
2. When we are editing the input, we have: 
  a. The submit form right before
  b. The search suggestions right after
I think we want the search suggestions close to the input, and in order to quickly navigate to the *first* suggestion, it would be better to have them after the input.
_Note :_ This wasn't a question before https://github.com/xwiki/xwiki-platform/pull/2142 since the search suggestion were not there in the DOM.

One solution to solve this duality problem would be to add a separate element for both the dropdown and the submit interactions. However this would probably mean a UI change.

michitux: Hm, yes, with the search suggestions it is indeed not that easy. I mean the search suggestions could possibly (haven't checked) be between the input and the submit button and could be dismissed with escape to reach the submit button more easily but I understand your point. For this "separate toggle and submit" we could indeed have a separate submit button that is revealed together with the input while hiding the open button. However, this makes everything more complex.

My concerns just come from the fact that it seems unnatural to reverse the element order in CSS and I have the feeling we're adding technical debt here.

Anyways, there seem to be good reasons to go with what you're suggesting here. We could re-visit this decision when we address the unnatural tab order in the right part of the top navigation.

PR #2165, opened 2023-05-03
XWIKI-20844: Implement an accessible pattern for document extra tabs

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20844
**PR Changes:**
* Removed All tabs from the tabindex. Now only the active tab can be focused using Tab.
* Added arrow navigation through the tabs
* Minimal changes to the visual UI

**Demo:** https://up1.xwikisas.com/#V1vdhvIRmRdOUH3EVW6XHA

**Notes:**
* Focus management was complicated. I could manage to keep focus close to objects, but the AJAX requests lost it. I couldn't find a solution to refocus the tab properly after loading its panel with AJAX. It works perfectly when there's no loading though.
* A lot of style had to be added since it comes from a bootstrap definition. Might be possible to refactor it better in the LESS
* We had a discussion on how to implement these changes over at https://forum.xwiki.org/t/javascript-xwiki-augmentation/12417


michitux (changes requested):
Apart from the small changes I requested, this looks good to me.

I see that there is a comment in `layout.less` about `.xwikitabbar` that says "Used by: AllDocs, DocExtra". Have you checked how AllDocs ("Page Index" in the drawer at the right) looks with your changes?

Sereza7:
It does break AllDocs:
20844-breakAllDocs
Setting to draft until solved.

Sereza7:
## PR Changes

* Updated PR to take into account the AllDocs page
* Factorized some style
* Added a forgotten aria-labelledby

## Video demo
https://up1.xwikisas.com/#NtC7sfU9J3ezCL8DN5MaOQ
## TODO
Check if there's tests about the AllDocs page and make sure they don't fail.

michitux (changes requested):
Two more comments where I'm wondering if additional CSS is required for the page index. Also, I found more code that uses `.xwikitabbar` with a list both in XWiki itself and in an extension:

* https://github.com/xwiki/xwiki-platform/blob/b320fb7914887e05a460808f9161c660dc0278f1/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/CreateWiki.xml#L255
* https://github.com/xwikisas/macro-pdfviewer/blob/4ea0048aec249c94c5b4d465e5c7278ff7791378/api/src/main/resources/templates/pdfviewer/pdfviewer.vm#L155

I fear both will be broken after this change. Maybe you could keep the CSS backwards compatible for them? Or even go back to a list? I understand that this is not the ideal solution according to this [Stack Overflow discussion](https://stackoverflow.com/questions/69756534/how-to-fix-role-tablist-role-tab-axe-violations) but maybe it is the more backwards-compatible option?

Sereza7:
Not passing a mvn build of NestedNavigationIT **as of now**, because of the master build fail
https://ci.xwiki.org/job/XWiki/job/xwiki-platform/job/master/6737/testReport/junit/org.xwiki.flamingo.test.docker/AllIT$NestedNavigationIT/Platform_Builds___main___integration_tests___IT_for_xwiki_platform_core_xwiki_platform_flamingo_xwiki_platform_flamingo_skin_xwiki_platform_flamingo_skin_test_xwiki_platform_flamingo_skin_test_docker___Build_for_IT_for_xwiki_platform_core_xwiki_platform_flamingo_xwiki_platform_flamingo_skin_xwiki_platform_flamingo_skin_test_xwiki_platform_flamingo_skin_test_docker___docExtraLoadingFromTabClicks_TestUtils__TestReference_/

Sereza7:
> Two more comments where I'm wondering if additional CSS is required for the page index.

Fixed an imprecision in https://github.com/xwiki/xwiki-platform/pull/2165/commits/b37deed193f33b1d3cc99342eff79b83eafeb0f5 .
AFAIK it looked okay with manual tests.
> Also, I found more code that uses .xwikitabbar with a list both in XWiki itself and in an extension:
> * [xwiki-platform/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/CreateWiki.xml](https://github.com/xwiki/xwiki-platform/blob/b320fb7914887e05a460808f9161c660dc0278f1/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/CreateWiki.xml#L255) Line 255 in [b320fb7](https://github.com/xwiki/xwiki-platform/commit/b320fb7914887e05a460808f9161c660dc0278f1)
>      &lt;ul class="xwikitabbar"&gt; 

Addressed this issue in https://github.com/xwiki/xwiki-platform/pull/2165/commits/8131492b349b6ff1e5c070e44dc9756193189e60/8131492b349b6ff1e5c070e44dc9756193189e60 . The fix is not perfect but this is Admin only UI so accessibility is not as paramount.
> * https://github.com/xwikisas/macro-pdfviewer/blob/4ea0048aec249c94c5b4d465e5c7278ff7791378/api/src/main/resources/templates/pdfviewer/pdfviewer.vm#L155

**TODO: Can open a ticket on the extension if we agree with the current solution**
____

Didn't spot either of those when checking myself: did not check the double quote use of the html class and did not check the extensions.
What is the way you recommend to search through the extensions? I can't imagine one needs to search through every single one individually.
____
> I fear both will be broken after this change. 
> Maybe you could keep the CSS backwards compatible for them? 

:+1:  Addressed in https://github.com/xwiki/xwiki-platform/pull/2165/commits/ece13973311bf8494e2d3328c53da94bebaf1361
> Or even go back to a list?  I understand that this is not the ideal solution according to this [Stack Overflow discussion](https://stackoverflow.com/questions/69756534/how-to-fix-role-tablist-role-tab-axe-violations) but maybe it is the more backwards-compatible option?

:-1: I don't think it's worth making the structure heavier and less understandable just for the sake of an eventual backward compatibility. There's a high chance custom code would not work anymore even if we keep the \<li> elements.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
```diff
@@ -324,17 +324,44 @@
 // Tabbar =====================================================================
 
 .xwikitabbar { // Used by: AllDocs, DocExtra
-  .nav;
-  .nav-tabs;
+  display: flex;
+  border-bottom: 1px solid @nav-tabs-border-color;
   margin: 0;
+  & > button, & > a{
```
michitux, line 330:
```suggestion
  & > button, & > a {
```

michitux:
> **What is the way you recommend to search through the extensions? I can't imagine one needs to search through every single one individually.**

Just use the search on GitHub, either globally or limit it to organizations like `xwikisas` and `xwiki-contrib`, this should cover most extensions.

Is my understanding correct that the display should now be unchanged for extensions using the `xwikitabbar` class?

Apart from the space in the CSS, this looks good now.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
```diff
@@ -91,23 +91,24 @@
 #if($docextras.size() > 0)
   <div id="xwikidatacontents">
     <div class="floatcontainer" id="docExtraTabs">
-      <ul class="xwikitabbar" id="docExtrasTabsUl">
+      <div id='tablistlabel' class='hidden'>$escapetool.xml($services.localization.render('docextra.tablist.label'))</div>
```
michitux, line 94: Why are you using a `hidden` class here and not just the [hidden attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)?

Sereza7:
@michitux 
> Just use the search on GitHub, either globally or limit it to organizations like xwikisas and xwiki-contrib, this should cover most extensions.

Thanks! I didn't use it before but it's really useful to check out public extension source :)

> Is my understanding correct that the display should now be unchanged for extensions using the xwikitabbar class?

See the [changes I made to the CreateWiki file](https://github.com/xwiki/xwiki-platform/pull/2165/files#diff-044b7721066e9933c59dccb98f1cfae4ad4931eed13d097f5d6f6b986083ea93 ). This will change the way the xwikitabbar is displayed, but if its structure is not updated the new accessibility features won't work.
In order to make sure I didn't break the visuals, I tested the tab funcitonality from the Pro PDF Viewer you linked above:
20844-noChangeFormerUI
It seems to me like the UI is not changed and the extension is fine. I didn't change much style whatsoever in this PR (just looked back at the files to check that).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/print.css
```diff
@@ -38,7 +38,7 @@
   background-color: white;
 }
 
-.xwikitabbar li.active {
+.xwikitabbar button.active {
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2165/commits/ece13973311bf8494e2d3328c53da94bebaf1361

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
Sereza7: This class was already used extensively in this file, notably on https://github.com/xwiki/xwiki-platform/blob/e7b6d35fbb7dd55af958496a0e6c19e344411817/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm#L111 
that was written in 2014.

___
In regards to the discussion about it on https://github.com/xwiki/xwiki-platform/pull/2182#discussion_r1268224725 , I decided to change this use of "hidden" to the `sr-only` class. Addressed in https://github.com/xwiki/xwiki-platform/pull/2165/commits/19b1cf391baa5ea4920bd157ee146f138d3e550c

michitux:
> This will change the way the xwikitabbar is displayed, but if its structure is not updated the new accessibility features won't work.

That change how the tabs are doesn't look good from what I can see. In your screenshot it looks like there is too much margin on the left and the line and margin below the tabs is missing.

> **TODO: Can open a ticket on the extension if we agree with the current solution**

This extension (PDF viewer macro) supports XWiki 13.10, it will probably upgrade to 14.10 in half a year and then XWiki 15.10 in another year, so yes, in about 1.5 years the extension can be updated to use the new accessibility features. Until then, it will need to use something that works both with the old and the new styles. For this reason, please keep them backwards compatible with the old HTML structure.

Further, the change to the HTML structure of the tab bar also breaks the [sunken flamingo theme](https://extensions.xwiki.org/xwiki/bin/view/Extension/SunkenFlamingoTheme/) as it defines styles targeting `.xwikitabbar > li.active > a, .xwikitabbar > li.active > a:hover, .xwikitabbar > li.active > a:focus`.

So in other words, I think this is really a case where we should consider this tab bar HTML structure an API in the front-end. I would definitely like to get another opinion if it is okay to break, if this shouldn't even have a vote on the forum to decide that this API breakage is okay.

@mflorea maybe we could have your opinion the change of the HTML structure/design changes that affect the PDF viewer pro macro?

michitux (commented):
In the [Bootstrap 5 documentation](https://getbootstrap.com/docs/5.0/components/navs-tabs/#regarding-accessibility) it is mentioned that in the case that the tabs are regular links that navigate to other pages (as it is the case on the page index), a `navigation` role should be used and `tablist` or `tabpanel` should not be used. Unfortunately, the link to the [WAI ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices/#tabpanel) in that document is broken and I couldn't find the exact content they're referring to.

Sereza7:
> That change how the tabs are doesn't look good from what I can see. In your screenshot it looks like there is too much margin on the left and the line and margin below the tabs is missing.

I could see the issue with the left margin (addressed in https://github.com/xwiki/xwiki-platform/pull/2165/commits/5caca1a0b845112d303958ac156f8d75108d8cb5 ), but not with the line and margin below the tabs. Here is a screenshot of the reference I took before doing the changes:
20844-defaultView
Here is what this PR looks like after the small padding fix I added in https://github.com/xwiki/xwiki-platform/pull/2165/commits/5caca1a0b845112d303958ac156f8d75108d8cb5 : 
20844-AfterFix

>  This extension (PDF viewer macro) supports XWiki 13.10, it will probably upgrade to 14.10 in half a year and then XWiki 15.10 in another year, so yes, in about 1.5 years the extension can be updated to use the new accessibility features. Until then, it will need to use something that works both with the old and the new styles. For this reason, please keep them backwards compatible with the old HTML structure.

As things are done in this PR right now, it will have no or minimal impact on the backward HTML. This also means that the accessibility caveats will stay as long as the old HTML structure is kept (from what I can remember, the HTML that was used was not precise enough to convey very accessible content). It shouldn't be an issue for backward compatibility.

> In the [Bootstrap 5 documentation](https://getbootstrap.com/docs/5.0/components/navs-tabs/#regarding-accessibility) it is mentioned that in the case that the tabs are regular links that navigate to other pages (as it is the case on the page index), a navigation role should be used and tablist or tabpanel should not be used. Unfortunately, the link to the [WAI ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices/#tabpanel) in that document is broken and I couldn't find the exact content they're referring to.

I can't find anything about the navigation role on [the tab pattern proposed by WAI-ARIA](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) either. I don't think it's wrong to make some tabs wrapped in a navigation landmark. In most cases people should use a menu pattern (with a menubar role) instead, but I think our use is appropriate.

Moreover, I don't want to change too much the page index page, as this PR aims at fixing accessibility issues on the document extra tabs. Of course, replacing the tab structure from the index would be the perfect solution, but this takes time and introduces complexity in this PR that is not needed to solve the issue at hand. 

____

> So in other words, I think this is really a case where we should consider this tab bar HTML structure an API in the front-end. I would definitely like to get another opinion if it is okay to break, if this shouldn't even have a vote on the forum to decide that this API breakage is okay.
@mflorea maybe we could have your opinion the change of the HTML structure/design changes that affect the PDF viewer pro macro?

Summary of the discussion so far: in order to improve accessibility of the tabs, we need to change their HTML structure. However, this structure is used in a few places in xwiki-platform and extensions. Is it okay to partially break backward compatibility for it?

mflorea:
> 👎 I don't think it's worth making the structure heavier and less understandable just for the sake of an eventual backward compatibility. There's a high chance custom code would not work anymore even if we keep the <li> elements.

@Sereza7 I don't agree with this. Backward compatibility is very important. More important than having the optimal HTML structure. End users don't care about the HTML structure. But they will complain if an extension stops working after an upgrade. What's the impact for the end user of making the structure heavier and less understandable?

I see two options:

* either try to make the tabs accessible with as little structural changes as possible (e.g. only replace the anchors with buttons and add some HTML attributes)
* or define a new UI component (i.e. an "accessible tab list") that uses a different HTML structure, different CSS classes and styles and start migrating existing code to the new UI component; XWiki extensions will be migrated slowly, but what's important is that they will continue to work.

@michitux WDYT?

Sereza7:
Since we didn't reach an agreement on the solution proposed here and it didn't get updated in a while, I'll close the PR.

PR #2166, opened 2023-05-03
XWIKI-18838: Use ARIA landmarks for main page regions

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-18838
**PR Changes:**
* Added main landmarks on pages.

**Notes:**
* As of now, using the preview will create an accessibility violation (two `main` landmarks on the page, one inside an iframe in the other). AFAIK this is only on a few pages accessible by admins (e.g. color theme editing) so it's okay.
* The element that was used to be the `main` landmark depends on the templates, since [we don't want to include breadcrumbs in it](https://forum.xwiki.org/t/accessibility-best-practice-landmarks/12051/9). Typically, it was either: set on the mainContentArea if it doesn't start with a "top breadcrumb", or set on a new DOM element that contains the title and the content of the page (without the breadcrumb).
* This PR is very much related to https://github.com/xwiki/xwiki-platform/pull/2133 . The following views are made without the changes in the PR for XWIKI-20696. My solution for XWIKI-20696 as of now contains changes to the hierarchy macro and makes all breadcrumbs `nav` landmarks.

**View:**
Those screenshots were made with the landmark extension for Firefox.
You should look for the `Main` landmark indicated with a pink square on every screenshot
18838-accessDenied
___
18838-login
___
18838-editInline
___
18838-edit
___
18838-DoesNotExist
___
18838-FlamingoDelete
___
18838-FlamingoCreate
___
18838-flamingoCopy
___
18838-contentView
___
18838-moveAttachment


Sereza7:
### Ready for review
**PR Changes:**
* Updated the test resources in `xwiki-platform-tool-standards-validator`

**Note:** Tests on the `xwiki-platform-tool-standards-validator` module after changes [passed](https://up1.xwikisas.com/#RY7BYLLfd_ijp74PbDlFkg).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/delete.vm
```diff
@@ -43,13 +43,15 @@
...
     #template("hierarchy.vm")
+    <main>
     #define($title)<a href="$doc.getURL('view')">$!escapetool.xml($doc.plainTitle)</a>#end
     #set($titleToDisplay = $services.localization.render('core.delete.title', [$title]))
     <div class="xcontent">
       #template('contentheader.vm')
       $deletecontent
       <div class="clearfloats"></div>
     </div> ## xcontent
+    </main>
```
manuelleduc, line 54:
```suggestion
    <main>
      #define($title)<a href="$doc.getURL('view')">$!escapetool.xml($doc.plainTitle)</a>#end
      #set($titleToDisplay = $services.localization.render('core.delete.title', [$title]))
      <div class="xcontent">
        #template('contentheader.vm')
        $deletecontent
        <div class="clearfloats"></div>
      </div> ## xcontent
    </main>
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docdoesnotexist.vm
```diff
@@ -29,6 +29,7 @@
   <div class="main">
   <div id="mainContentArea">
   #template("hierarchy.vm")
+  <main>
```
manuelleduc, line 32: inner indentation missing

Sereza7: I didn't put inner indentation on some files because:
- It's already a complex PR to review, we don't need to view the 20 lines where I just added one indentation.
- The indentation policy is already loose here (cf L29-30 on this document)
- We don't lose much information. Here, the level of the L31 is weird, but except this line, it doesn't matter if it's properly indented.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editinline.vm
```diff
@@ -49,6 +50,9 @@ $xwiki.ssfx.use('uicomponents/widgets/fullScreen.css', true)##
   </div>
 #end
 #template("hierarchy.vm")
+#if ($allowDocEdit)
```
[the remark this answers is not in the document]
Sereza7: 18838-editInlineBlocked
It works properly without it. It was just a safeguard around a behaviour I didn't understand:
18838-editInlineBlocked
On the screenshot, the generated mainContentArea has the `main` type on the left, despite it being a editinline. I can only guess that the root of the tree is generated in this case using another template. I couldn't figure out which one yet. Maybe it's warning -> message.vm .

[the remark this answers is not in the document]
Sereza7: With this safeguard, when we generate the warning, that already has a main in it, we don't generate a main elsewhere.

michitux:
While I agree that it might be preferable to not to have the breadcrumbs in the main content I wonder if it is really a good idea to have on some pages `<div id="mainContentArea">` while on others it is `<main id="mainContentArea">`. This creates quite an inconsistency that could lead to strange problems. Btw., I found this code which I think needs to be updated and also nicely highlights the problem with this inconsistency:

https://github.com/xwiki/xwiki-platform/blob/f472a032c0a4f6bc38ffcb1510dfd57ecb97e4cf/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/BasePage.java#L603

Also, see #1667, I think it would be nice to have a consistent jump target for the skip header link. Maybe we should re-consider the decision to (not) include the breadcrumbs in the main content area.

michitux:
As I've just found it, here is another comment saying that breadcrumbs shouldn't be part of the main content: https://github.com/w3c/aria-practices/issues/1469#issuecomment-664149633.

Sereza7:
## PR Updates
* Removed the breadcrumbs from main on every template.
* Improved comments
* Homogenized changes

## View
18838-delete
18838-notExisting
18838-editInline
18838-home
18838-copy

michitux (approved):
Looks good to me, but there is a conflict that prevents merging.

PR #2170, opened 2023-05-10
XWIKI-20825: Edit autosave frequency contrast issue

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20825

**PR Changes:**
* Removed the use of opacity
* Added a new class to the parent element instead.
* Updated style using theme for the new class.

**View:** 
20825-HighContrast

**Video demo:** https://up1.xwikisas.com/#31sM6F-oM1I2dW67EIaV3g

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -7,7 +7,7 @@
   .buttons {
     margin: 0;
     padding: 1em;
-    #autosaveControl.disabled > label.frequency {
+    #autosaveControl.autosaveDisabled > label.frequency {
       .text-muted;
     }
   }
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -7,6 +7,9 @@
   .buttons {
     margin: 0;
     padding: 1em;
+    #autosaveControl.disabled{
+      .text-muted;
+    }
   }
 
   &.sticky-buttons-fixed {
@@ -148,3 +151,5 @@ textarea#content {
 .editableProperty-editor {
   margin-top: @button-spacing;
 }
+
+
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
@@ -7,7 +7,7 @@
   .buttons {
     margin: 0;
     padding: 1em;
-    #autosaveControl.disabled {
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/edit.less
```diff
@@ -7,6 +7,9 @@
   .buttons {
     margin: 0;
     padding: 1em;
+    #autosaveControl.disabled{
```
surli, line 10: Missing a space before { (I don't know if we have a checkstyle for those, but doesn't look neat :)

Sereza7: :+1: Addressed in [a7ba0e4](https://github.com/xwiki/xwiki-platform/pull/2170/commits/a7ba0e4f83468e4216e074f6611cf3029cc4fefb)

surli (approved):

michitux:
Why did you move the style to the whole option? I think this creates bad usability. Having the whole text in lighter color creates the impression that the whole option cannot be used. Previously, only the text "every X minutes" was grayed out as setting it isn't useful when the option is disabled but the text "Autosave" was displayed in the regular text color as this is a regular option that can be enabled.

This separation in words probably also causes translation issues and might thus be questionable, but I think rather than making the whole text gray, we should think about either a) not making anything gray or b) splitting this into two clearly distinct options (like "Autosave" and "Autosave interval"). But that's a separate issue.

For now, I suggest to restore the previous behavior of just changing the color of the text around the input and not the "Autosave" text.

Sereza7:
## PR Change
* Restored the previous color behavior


## Answer to @michitux 
:+1: I reverted the changes accordingly. I think the best solution would be to have the frequency be shown only when it's relevant. This would reduce cluttering here. For now I just updated the selector to change the style of the frequency option only and reported this UI issue in a [new ticket](https://jira.xwiki.org/browse/XWIKI-20953).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/autosave.js
```diff
@@ -114,11 +112,11 @@ editors.AutoSave = Class.create({
     this.setTimeUnit();
     frequencyLabel.appendChild(document.createTextNode(" "));
     frequencyLabel.appendChild(this.timeUnit);
-    if (!this.options.enabled) {
-      frequencyLabel.setOpacity(this.options.disabledOpacity);
-    }
     // A paragraph containing the whole thing
     var container = new Element('div', {"id": "autosaveControl"});
+    if (!this.options.enabled) {
+      container.addClassName('disabled');
```
michitux, line 118: I wonder if this class name isn't too generic and could conflict with other uses. For example, I found this (deprecated) style that adds an opacity on any element with class `disabled`: https://github.com/xwiki/xwiki-platform/blob/bc6fb452ef10414c1756b2b547e8129ec1e35c9d/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/dataeditors.css#L190

I wonder if it wouldn't be enough to match on the checkbox state in CSS. The following selector seems to work in a simple test: `#autosaveControl:has(input:checked) label.frequency`.

michitux: Just noticed that this doesn't work in Firefox as it doesn't support `:has`. :disappointed: So we need the class I fear. What about making the class more specific like `autosaveDisabled`?

michitux: > Updated in the latest commit

Is it possible that forgot to push this commit?

PR #2174, opened 2023-05-12
XWIKI-20613: Tags are not spaced well when user does not have edit rights

wthrajat:
- [Jira Link here](https://jira.xwiki.org/browse/XWIKI-20613)

This PR adds a background color (`@breadcrumb-bg`) to the tags present in a wiki page to be able to better differentiate between different tags.

**Screenshots:**

- Before this PR (very difficult to figure out that `abstract theorem` is a single tag and NOT two different tags)
[
2023-05-12-19 50 49-screenshot
](url)

- After this PR (tags are easily differentiable all thanks to the background color) 
2023-05-12-19 51 59-screenshot
2023-05-12-19 49 56-screenshot


michitux:
The spacing looks a bit off in the screenshot, when comparing `monkey` and `probability`, there seems to be space inside the gray area after `monkey` but not after `probability` in the version without edit right. In the version with edit right, I'm also wondering if the "x" and thus the extra space for it should really be inside the background.

PR #2177, opened 2023-05-15
XWIKI-20903: Add support for configuring an empty line placeholder

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-20903

This PR adds a placeholder text to new lines in the WYSIWYG editor. 
The plugin comes with a default configuration that can be overridden by other plugins. (e.g. advertising features to the user)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
@@ -24,26 +24,6 @@
 
   CKEDITOR.plugins.add('xwiki-focusedplaceholder', {
 
-    /**
-     * Styles that would be applied to the editor by the placeholder text when visible.
-     *
-     * @property {String}
-     */
-    placeholderStyle: '[' + ATTRIBUTE_NAME + ']::before {' +
-      'position: absolute;' +
-      'opacity: .8;' +
-      'color: #aaa;' +
-      'content: attr( ' + ATTRIBUTE_NAME + ' );' +
-      '}' +
-      '.cke_wysiwyg_div[' + ATTRIBUTE_NAME + ']::before {' +
-      'margin-top: 1em;' +
-      '}',
-
-    onLoad: function () {
-      // Adding the style only when the editor is loaded prevents placeholders to appear in view mode
-      CKEDITOR.addCss(this.placeholderStyle);
-    },
-
     beforeInit: function (editor) {
       // Default plugin configuration, to be overriden by other plugins
       editor.config["xwiki-focusedplaceholder"] = {
@@ -118,6 +98,20 @@
     },
 
     init: function (editor) {
+
+      // Filter out placeholders when saving
+      var filterPlaceholders = {
+        elements: {
+          '^': function (element) {
+            if (element.attributes[ATTRIBUTE_NAME] !== undefined) {
+              delete element.attributes[ATTRIBUTE_NAME];
+            }
+          },
+        }
+      };
+      var htmlFilter = editor.dataProcessor && editor.dataProcessor.htmlFilter;
+      htmlFilter.addRules(filterPlaceholders);
+
 
       // Create a TextWatcher instance to detect changes in the document
       var placeholderTextWatcher = new CKEDITOR.plugins.textWatcher(editor, updatePlaceholder);
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
@@ -24,41 +24,31 @@
 
   CKEDITOR.plugins.add('xwiki-focusedplaceholder', {
 
+    /**
+     * Styles that would be applied to the editor by the placeholder text when visible.
+     *
+     * @property {String}
+     */
+    placeholderStyle: '[' + ATTRIBUTE_NAME + ']::before {' +
+      'position: absolute;' +
+      'opacity: .8;' +
+      'color: #aaa;' +
+      'content: attr( ' + ATTRIBUTE_NAME + ' );' +
+      '}' +
+      '.cke_wysiwyg_div[' + ATTRIBUTE_NAME + ']::before {' +
+      'margin-top: 1em;' +
+      '}',
+
+    onLoad: function () {
+      // Adding the style only when the editor is loaded prevents placeholders to appear in view mode
+      CKEDITOR.addCss(this.placeholderStyle);
+    },
+
     beforeInit: function (editor) {
       // Default plugin configuration, to be overriden by other plugins
       editor.config["xwiki-focusedplaceholder"] = {
-        // Default placeholder for content sectioning, text and table HTML tags
-        placeholder: {
-          body: "xwiki-focusedplaceholder.placeholder.body", // Body
-          address: "xwiki-focusedplaceholder.placeholder.address", // Address
-          aside: "xwiki-focusedplaceholder.placeholder.aside", // Aside
-          footer: "xwiki-focusedplaceholder.placeholder.footer", // Footer
-          header: "xwiki-focusedplaceholder.placeholder.header", // Header
-          h1: "xwiki-focusedplaceholder.placeholder.h1", // Heading 1
-          h2: "xwiki-focusedplaceholder.placeholder.h2", // Heading 2
-          h3: "xwiki-focusedplaceholder.placeholder.h3", // Heading 3
-          h4: "xwiki-focusedplaceholder.placeholder.h4", // Heading 4
-          h5: "xwiki-focusedplaceholder.placeholder.h5", // Heading 5
-          h6: "xwiki-focusedplaceholder.placeholder.h6", // Heading 6
-          main: "xwiki-focusedplaceholder.placeholder.main", // Main content
-          nav: "xwiki-focusedplaceholder.placeholder.nav", // Navigation
-          section: "xwiki-focusedplaceholder.placeholder.section", // Section
-          blockquote: "xwiki-focusedplaceholder.placeholder.blockquote", // Quote
-          dl: "xwiki-focusedplaceholder.placeholder.dl", // Definition list
-          dt: "xwiki-focusedplaceholder.placeholder.dt", // Definition entry
-          dd: "xwiki-focusedplaceholder.placeholder.dd", // Description
-          figcaption: "xwiki-focusedplaceholder.placeholder.figcaption", // Caption
-          figure: "xwiki-focusedplaceholder.placeholder.figure", // Figure
-          ol: "xwiki-focusedplaceholder.placeholder.ol", // Ordered list
-          ul: "xwiki-focusedplaceholder.placeholder.ul", // Unordered list
-          menu: "xwiki-focusedplaceholder.placeholder.menu", // Menu (unordered list)
-          li: "xwiki-focusedplaceholder.placeholder.li", // List item
-          p: "xwiki-focusedplaceholder.placeholder.p", // Paragraph
-          pre: "xwiki-focusedplaceholder.placeholder.pre", // Preformated
-          caption: "xwiki-focusedplaceholder.placeholder.caption", // Title
-          th: "xwiki-focusedplaceholder.placeholder.th", // Header
-          td: "xwiki-focusedplaceholder.placeholder.td", // Cell
-        },
+        // Left empty to be configured in the next instruction
+        placeholder: {},
         /* Empty inline tags (or nodeNames) that usually do not appear on screen should be ignored by default
          * because they might get propagated to new lines by CKEditor*/
         ignoreIfEmpty: [
@@ -87,6 +77,44 @@
           "ins",
         ],
       };
+
+      // Add generic default placeholders for content sectioning, text and table HTML tags
+      [
+        "body",
+        "address",
+        "aside",
+        "footer",
+        "header",
+        "h1",
+        "h2",
+        "h3",
+        "h4",
+        "h5",
+        "h6",
+        "main",
+        "nav",
+        "section",
+        "blockquote",
+        "dl",
+        "dt",
+        "dd",
+        "figcaption",
+        "figure",
+        "ol",
+        "ul",
+        "menu",
+        "li",
+        "p",
+        "pre",
+        "caption",
+        "th",
+        "td",
+      ].forEach(function (tagName) {
+        editor
+          .config["xwiki-focusedplaceholder"]
+          .placeholder[tagName] = "xwiki-focusedplaceholder.placeholder." + tagName;
+      });
+
     },
 
     init: function (editor) {
@@ -101,34 +129,34 @@
         // Attach the textWatcher to the editor
         placeholderTextWatcher.attach();
 
-        /* The placeholder should update when the content is changed via menus.
-         * snapshots are created when that occurs*/
+        // The placeholder should update when the content is changed via menus.
+        // snapshots are created when that occurs
         editor.on("saveSnapshot", function () {
           placeholderTextWatcher.check(false);
         });
-      });
 
+      });
 
       // Using contentDom event in case the DOM Tree gets recreated by CKEditor (e.g. mode change)
       editor.on("contentDom", function () {
-
+        var editable = editor.editable();
         // The placeholder should update when the user clicks somewhere in the document.
-        editor.document.$.addEventListener("click",
+        editable.attachListener(editable, "click",
           function () {
             placeholderTextWatcher.check(false);
           });
 
         // The placeholder should update when the editor gains focus.
-        editor.document.$.addEventListener("focus",
+        editable.attachListener(editable, "focus",
           function () {
             placeholderTextWatcher.check(false);
           });
 
-        /* The placeholder should disappear when the editor loses focus.
-         * (The empty document placeholder might appear) */
-        editor.document.$.addEventListener("blur",
+        // The placeholder should disappear when the editor loses focus.
+        // (The empty document placeholder might appear)
+        editable.attachListener(editable, "blur",
           function () {
-            removeAllPlaceholders(editor.document.$);
+            removeAllPlaceholders(editor.editable().$);
           });
 
       });
@@ -208,7 +236,7 @@
        */
       function updatePlaceholder(range) {
         // Clear previous placeholders
-        removeAllPlaceholders(range.document.$);
+        removeAllPlaceholders(range.root.$);
 
         // No placeholder when a selection is occuring
         if (!range.collapsed) {
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
@@ -17,252 +17,239 @@
  * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
-(function() {
-    'use strict';
+(function () {
+  'use strict';
 
-    var ATTRIBUTE_NAME = 'data-xwiki-focusedplaceholder';
+  var ATTRIBUTE_NAME = 'data-xwiki-focusedplaceholder';
 
-    CKEDITOR.plugins.add('xwiki-focusedplaceholder', {
+  CKEDITOR.plugins.add('xwiki-focusedplaceholder', {
 
-        /**
-         * Styles that would be applied to the editor by the placeholder text when visible.
-         *
-         * @property {String}
-         */
-        placeholderStyle: '[' + ATTRIBUTE_NAME + ']::before {' +
-            'position: absolute;' +
-            'opacity: .8;' +
-            'color: #aaa;' +
-            'content: attr( ' + ATTRIBUTE_NAME + ' );' +
-            '}' +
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+         */
+        placeholderStyle: '[' + ATTRIBUTE_NAME + ']::before {' +
+            'position: absolute;' +
+            'opacity: .8;' +
+            'color: #aaa;' +
+            'content: attr( ' + ATTRIBUTE_NAME + ' );' +
+            '}' +
+            '.cke_wysiwyg_div[' + ATTRIBUTE_NAME + ']::before {' +
+            'margin-top: 1em;' +
+            '}',
```
mflorea, line 40: I'm wondering if it's not better to put the styles in https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/ContentSheet.xml#L202 (which holds custom styles that apply to the edited content). They wont be packaged with the plugin code but at least we don't mix JavaScript and CSS. WDYT?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+(function() {
+    'use strict';
```
mflorea, line 21: So far we used 2 spaces indentation in our JavaScript code. Would be good to follow the same style, for consistency.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+                    figure: "xwiki-focusedplaceholder.placeholder.figure", // Figure
+                    ol: "xwiki-focusedplaceholder.placeholder.ol", // Ordered list
+                    ul: "xwiki-focusedplaceholder.placeholder.ul", // Unordered list
+                    menu: "xwiki-focusedplaceholder.placeholder.menu", // Menu (unordered list)
+                    li: "xwiki-focusedplaceholder.placeholder.li", // List item
+                    p: "xwiki-focusedplaceholder.placeholder.p", // Paragraph
+                    pre: "xwiki-focusedplaceholder.placeholder.pre", // Preformated
+                    caption: "xwiki-focusedplaceholder.placeholder.caption", // Title
+                    th: "xwiki-focusedplaceholder.placeholder.th", // Header
+                    td: "xwiki-focusedplaceholder.placeholder.td", // Cell
```
mflorea, line 81: I wouldn't make this default configuration explicit. Instead, given a ``tagName`` I would:
```
var placeholder = editor.config.focusedplaceholder.placeholder[tagName];
if (typeof placeholder !== 'string') {
  placeholder = editor.localization.get('xwiki-focusedplaceholder.placeholder.' + tagName);
}
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+                    "del",
+                    "ins",
+                ],
+            };
+        },
+
+        init: function(editor) {
+
+            var focusedplaceholder = editor.plugins["xwiki-focusedplaceholder"];
+            focusedplaceholder.editor = editor;
```
mflorea, line 116: This is not right. Plugin instances are shared between editor instances. Edit ``/xwiki/bin/view/Help/Applications/Movies/Meet John Doe`` which has 2 editor instances and do this in the JavaScript console:
```
CKEDITOR.instances['Help.Applications.Movies.Code.MoviesClass_0_longText1'].plugins['xwiki-macro'].foo = 'bar'
CKEDITOR.instances['content'].plugins['xwiki-macro'].foo
```

If you need a state you should save it on the editor instance, not on the plugin instance.

Also, the preferred style is to use camelcase for variable names, so ``focusedPlaceholder``.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+        isEmpty: function(element) {
+
+            // Consider the length of a text node as its number of children
+            if (element.nodeName === "#text") {
+                return !Boolean(element.length);
+            }
+            var i;
+            for (i = 0; i < element.childNodes.length; i++) {
+                var child = element.childNodes[i];
+                // Child is not empty if it doesn't have a nodeName
```
mflorea, line 175: Which nodes don't have a node name? Can you give an example?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+         *
+         * @param {string} tag - Lowercase tagName
+         * @return {string}
+         */
+        getPlaceholderContent: function(tag) {
+
+            var translationKey = this.editor.config.focusedplaceholder.placeholder[tag];
+
+            if (translationKey === null) {
+                translationKey = this.editor.config.focusedplaceholder.placeholder._fallback;
```
surli, line 257: The fallback value is currently empty...

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+            // No placeholder when a selection is occuring
+            if (!range.collapsed) {
+                return null;
+            }
+
+            // Find the matching ancestor
+            var ancestor = this.getFirstConfiguredAncestor(range.getCommonAncestor().$);
+            if (ancestor === null) {
+                ancestor = range.getCommonAncestor().$;
+            }
```
mflorea, line 238: This needs a comment. It's not clear to me why you need the fallback.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
douakli: I understand the idea, but I'm not sure how to properly handle non existing translation keys.
The localization.get call would fail silently.

rondales (approved):

rondales (approved):

rondales (approved):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
...
+            CKEDITOR.addCss(this.placeholderStyle);
+        },
+
+        beforeInit: function(editor) {
+            // Default plugin configuration, to be overriden by other plugins
+            editor.config.focusedplaceholder = {
+                // Default placeholder for content sectioning, text and table HTML tags
+                placeholder: {
+                    _fallback: "",
+                    body: "xwiki-focusedplaceholder.placeholder.body", // Body
```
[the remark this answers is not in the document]
douakli: I don't think the user can generate new `<body>` tags either, however the user can move the caret to a position where it is only within a body tag and not inside a p tag. If we remove placeholders for the body tags, no placeholder will appear when the user erases the whole document for instance.

In the same way, I think that the other tags may appear in a document even if not directly created by CKEditor. The XWiki Syntax provides means to insert html code in the document, a user or extension might introduce the html tags you are suggesting to remove in the document.
Is there a significant overhead if we configure them in case they might be used?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
douakli: Thank you, the fallback was indeed not used.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
douakli: I agree, the style is hard to read in Javascript context, I moved it in the ContentSheet.
Thank you.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
douakli: I did that so that the different functions would have access to the configuration. I didn't know the plugin instances were shared, thank you!
I moved the functions definitions in the init method so the editor would be in their scope.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
douakli: Currently, the body placeholder is configured by default so an ancestor will be found.
But if the configuration changes, this is not guaranteed.
I replaced the replacement value by an empty return because if no configured ancestor is found no placeholder will be inserted anyways.
I also added a comment as you requested.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
douakli: Ahah, I don't think I can give an example, I started with `tagName` so this was a check for non element nodes.
I removed the unnecessary check, thank you!

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,255 @@
...
+      /* Using contentDom event in case the DOM Tree gets recreated by CKEditor (e.g. mode change)*/
+      editor.on("contentDom", function () {
+
+        /* The placeholder should update when the user clicks somewhere in the document. */
+        editor.document.$.addEventListener("click",
+          function () {
+            placeholderTextWatcher.check(false);
+          });
+
+        /* The placeholder should update when the editor gains focus. */
```
manuelleduc, line 124: We usually use single line comments (`//`) for this kind of short comments.
```suggestion
        // The placeholder should update when the editor gains focus.
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,255 @@
...
+        document.querySelectorAll("[" + ATTRIBUTE_NAME + "]").forEach(function (match) {
+          match.removeAttribute(ATTRIBUTE_NAME);
+        });
+      }
+
+      /**
+       * Checks if an element should be considered empty
+       *
+       * @param {Object} element - the DOM Element
+       * @return {bool}
```
manuelleduc, line 154: missing return comment

manuelleduc:
Thanks for this work @douakli !
I would be interested to see some screenshots/recording of the feature :)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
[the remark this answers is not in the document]
Sereza7: > I don't think the user can generate new <body> tags either, however the user can move the caret to a position where it is only within a body tag and not inside a p tag.

20903-body
Even if the user gets out of the \<p> , the \<body> is way up here and not something you get close to when editing.
What you actually edit with CKEditor AFAIK is not the layout, but the page content and its title. So everything around (menus, header, footer, side panels, ...) is outside of the editing scope. For some tag names it's debatable if we can include it in the user content (aside especially), but for the few I mentioned above, if it's possible to use them in the editor, it's a major ticket we need to raise. 

> In the same way, I think that the other tags may appear in a document even if not directly created by CKEditor. The XWiki Syntax provides means to insert html code in the document, a user or extension might introduce the html tags you are suggesting to remove in the document.

If they can insert a \<body>, I'd consider it a bug. \<body> and the tag names mentioned above are a bit special, as they should only be used once in a HTML document. If we can use them in user generated content, do we have safeguards to avoid using them multiple times, and what if we already use them outside of the interface?
I don't think they are necessary to write down some page content, so I'm pretty sure CKEditor doesn't give an easy way to use them.

> Is there a significant overhead if we configure them in case they might be used?

Each tag element would need a translation for each language. It's not that much but it adds up quickly for translation contributors :)
Moreover, I think it's unclear to not make a distinction between the tag names that can get generated and those which can't. This might generate more work down the line. If I stumbled across this a few months ago, I might have thought that all tag names could be generated by CKEditor and searched more to see if I should report the potential issue mentioned above.
I think a comment explaining why you're putting every tag name (or not) would help.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,255 @@
...
+      };
+    },
+
+    init: function (editor) {
+
+      // Create a TextWatcher instance to detect changes in the document
+      var placeholderTextWatcher = new CKEDITOR.plugins.textWatcher(editor,
+        function (range) {
+          updatePlaceholder(range);
+        });
```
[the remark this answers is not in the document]
douakli: This was needed as the updatePlaceholder used to be a member method using the `this` keyword which would refer to the textWatcher instance and not the plugin instance when passed directly.
This is no longer the case, so your suggestion works, thank you!

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
mflorea: Why would the translations be missing? That would be a bug no? If you really want to bulletproof the code then you can check if the output of ``localization.get`` is the translation key, which means the translation is missing.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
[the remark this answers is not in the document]
douakli: > Even if the user gets out of the `<p>` , the `<body>` is way up here and not something you get close to when editing.

body-placeholder
The `<body>` is not that far away when using the iframe editor (Which an user can access in advanced mode by clicking the WYSIWYG Editor button from the dropdown edit menu. Apparently the placeholder currently updates before the `<p>` tag is inserted, I will try to find a better event to update the placeholder on to take the `<p>` insertion into account.

Thank you, that also lead me to notice that I was adding event listeners on the DOM Document, which is not what we want for the inline editor, I will find the correct element to add the listeners to in the ckeditor documentation.

> Each tag element would need a translation for each language. It's not that much but it adds up quickly for translation contributors :)

That is something I did not think about, indeed. :sweat_smile:
I can fix the update for late automatic `<p>` insertions, remove the `body` tag from the configured elements
and add a comment explaining that not all the tags can be created by ckeditor but that the caret might be moved into them. What do you think? Should we discuss it with other committers on the XWiki matrix channel?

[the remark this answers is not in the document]
mflorea: > A somewhat relevant reference: https://ckeditor.com/docs/ckeditor5/latest/features/html/general-html-support.html#enabling-all-html-features

@Sereza7 we're using CKEditor 4 not 5.

> Even if the user gets out of the <p> , the <body> is way up here and not something you get close to when editing.

In-place editing is not the only usage of CKEditor in XWiki. You're forgetting the **standalone** WYSIWYG edit mode (currently used when creating a new wiki page) and the Text Area object properties that are editable with the CKEditor. In these cases the **editable** element is the body element, because the edited content is loaded in an iframe.

> If they can insert a <body>, I'd consider it a bug. <body> and the tag names mentioned above are a bit special, as they should only be used once in a HTML document.

The placeholders @douakli is adding are not about content generation. First you need to understand that WYSIWYG edit mode and CKEditor are separate things. The WYSIWYG edit mode is **using** the CKEditor, but CKEditor is not limited to the WYSIWYG edit mode. CKEditor is a generic rich text editing widget and so it needs to support many uses cases. Nothing prevents me from activating the CKEditor in-line (non-iframe based) on a footer, header or main element. Even if we don't have this use case right now there's no reason to limit this just because the WYSIWYG edit mode doesn't need it. Again, the CKEditor Integration module provides an API for extension developers to use CKEditor in XWiki, outside of the WYSIWYG edit mode.

> Each tag element would need a translation for each language. It's not that much but it adds up quickly for translation contributors :)

I don't think the number of translation messages is a problem.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -17,252 +17,236 @@
...
+          });
+
+        // The placeholder should update when the editor gains focus.
+        editor.document.$.addEventListener("focus",
+          function () {
+            placeholderTextWatcher.check(false);
+          });
+
+        /* The placeholder should disappear when the editor loses focus.
+         * (The empty document placeholder might appear) */
```
mflorea, line 128:
```suggestion
        // The placeholder should disappear when the editor loses focus.
        // (The empty document placeholder might appear)
```

douakli:
> I would be interested to see some screenshots/recording of the feature :)

Here is a quick preview of the feature, I didn't go into all the types of tags but I think it gives an idea of how it behaves :)
PluginPreview

mflorea (commented):
Looks good.

rondales (commented):
#2169

manuelleduc:
> > I would be interested to see some screenshots/recording of the feature :)
> 
> Here is a quick preview of the feature, I didn't go into all the types of tags but I think it gives an idea of how it behaves :) PluginPreview

Out of curiosity,

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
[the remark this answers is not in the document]
Sereza7: Thanks for the explanation and the demo! I did not understand our use of CKEditor enough.
It's much clearer now. :slightly_smiling_face:

rondales (approved):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-focusedplaceholder/plugin.js
```diff
@@ -0,0 +1,268 @@
+/*
```
rondales, line 1: True

mflorea (approved):

douakli:
I want to thank everyone for the reviews!

@mflorea pushed the changes in master, hence this PR is no longer needed

PR #2182, opened 2023-05-17
XWIKI-20891: Rich select input is not accessible

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20891
## PR Changes
* Added keyboard support for the rich select input
* Added proper attributes to the rich select input
* Updated style: 
  * selected and hovered colors now use the new colortheme variables (and are different one from the other!)
  * The input now uses a flex layout instead of hacks with float.
## Notes
* This PR adds a lot of event listeners.
* Some specific interactions with the option filter make the jquery selectors for moving around the options quite complex.
* The up and down arrows are not implemented in [the default grouped listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/). I decided to add a behavior to them. As of the start of this PR, they allow to navigate quickly from one category of options to the other. This might not be the best way, but it's clearer semantically. Going up or down a line is doing a +2 /-2 in the option index, which is not really that interesting but could be easier to use since it matches with the visuals. For screen reader users, this is similar to content navigation: up and down to move line (category) and right/left to move letter (option).
## Video demo
https://up1.xwikisas.com/#NKIaIoFqGoP6AhU6bwPLPQ
(Reference of the same interface before the PR for comparison: https://up1.xwikisas.com/#mXWDrjTBkKmNfGSFjbHPnw)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -2623,40 +2623,47 @@ Recursive title display detected!##
...
+    <span id="${escapedId}-listbox-description" class="sr-only">
+      $escapetool.xml($services.localization.render('web.widgets.select.listbox.description'))
+    </span>
+    <div id="${escapedId}-listbox" role='listbox' class='xwiki-select-options' tabindex='0'
+      aria-labelledby="${escapedId}-listbox-label" aria-describedby="${escapedId}-listbox-description"
+      aria-activedescendant="${fieldName}_0_option">
+      #set ($count = 0)
+      #foreach ($category in $options)
+        <ul id="${escapedId}-group${count}" role='group' class="xwiki-select-category">
+          <label for="${escapedId}-group${count}">
```
michitux, line 2638: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#technical_summary), `<label>` isn't allowed as child of `<ul>`.

michitux (changes requested):
Overall this looks good (I haven't checked every detail, though), but the HTML structure doesn't work, you cannot have `<label>` as child of `<ul>`. I think using a definition list might work better.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
Sereza7: Adressed in https://github.com/xwiki/xwiki-platform/pull/2182/commits/69fe0619c9b32bf77e7bd6466e8dd6ba56549648

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -2623,40 +2623,47 @@ Recursive title display detected!##
       <input type="text" id="${escapedId}-filter" class="xwiki-select-filter"
         placeholder="$escapetool.xml($services.localization.render('web.widgets.select.filter.placeholder'))" />
     </p>
-    <div class="xwiki-select-options">
-      <ul>
-        #set ($count = 0)
-        #foreach ($category in $options)
-          <li class="xwiki-select-category">$category.name (<span class="xwiki-select-category-count">$category.options.size()</span>)
```
michitux, line 2630: I'm wondering why you removed this outer list. Having a list of a list seems logical here. You could still have the label as first child of the list item.

michitux (approved):
Looks good overall, there is just one HTML change I'm wondering about.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -2623,40 +2623,47 @@ Recursive title display detected!##
...
+    <span id="${escapedId}-listbox-description" class="sr-only">
+      $escapetool.xml($services.localization.render('web.widgets.select.listbox.description'))
+    </span>
+    <div id="${escapedId}-listbox" role='listbox' class='xwiki-select-options' tabindex='0'
+      aria-labelledby="${escapedId}-listbox-label" aria-describedby="${escapedId}-listbox-description"
+      aria-activedescendant="${fieldName}_0_option">
+      #set ($count = 0)
+      #foreach ($category in $options)
+        <label for="${escapedId}-group${count}">
+          $category.name (<span class="xwiki-select-category-count">$category.options.size()</span>)
```
michitux, line 2638: I've just checked and [the HTML standard](https://html.spec.whatwg.org/#the-label-element) says that the `for` attribute must point to an id of a [labelable element](https://html.spec.whatwg.org/#category-label). Unfortunately, elements with role `group` do not seem to be in that category.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
michitux: See my comment below regarding the label implementation. My point was more about backwards-compatibility, e.g., with custom CSS. Now searching GitHub, I couldn't find any custom code re-using these classes so its probably not such an issue here. I just think that if the structure somehow makes sense it might be worth not changing it too much.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2182/commits/bcd341777cc54503603a51be015dfd329085ae57

I decided to follow the same structure as the one from the codepen at https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/ .

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -2642,40 +2642,47 @@ Recursive title display detected!##
...
-                    <label for="${fieldName}_${count}">$escapetool.xml($option.name)</label>
-                    #if ("$!option.description" != '')
-                      <p class="xHint">$escapetool.xml($option.description)</p>
-                    #end
-                  </div>
-                </li>
-                #set ($count = $count + 1)
-              #end
-            </ul>
+    <span id="${escapedId}-listbox-label" class="sr-only">
```
michitux, line 2645: Why is this `sr-only` while in the very similar similar document extra tabs, the label and the description are hidden?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -2642,40 +2642,47 @@ Recursive title display detected!##
...
+      $escapetool.xml($services.localization.render('web.widgets.select.listbox.label', [$fieldName]))
+    </span>
+    <span id="${escapedId}-listbox-description" class="sr-only">
```
michitux, line 2648: Same here, why is this sr-only while in the very similar document extra tabs, the label and the description are hidden?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/select/select.js
```diff
@@ -51,11 +59,189 @@ require(['jquery'], function($) {
...
+      self.changeHighlight(option);
+      // Update the new selected option
       option.addClass('xwiki-select-option-selected');
+      option.attr('aria-selected', 'true');
       self.triggerSelectionChange();
     };
 
+    /**
+     * Highlights the option given as a parameter.
+     * @since 15.5RC1
```
michitux, line 81: The since-versions here and below need to be updated.

michitux (approved):
Looks still good and now better with the "label". I have just some doubts regarding the `sr-only` class for label/description and the since-versions need to be updated.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
Sereza7: Thanks for the question. I'm writing a proposal for HTML/CSS recommendations so that such an inconsistency won't happen anymore.
Note that sr-only only hides visually and hidden also hides semantically. However,  when referenced with aria-labelledby or aria-describedby, the content is still found and read out even if the node is hidden.
I think that using `sr-only` is still preferable, I don't know if all web browser and assistive techs will consider the above feature as a feature or a bug that needs to be fixed.
I'll fix the PR for XWIKI-20844 where I hid document extra tabs labels with 'hidden'.

Sereza7: Here is the proposal for code style: https://forum.xwiki.org/t/css-code-style-hiding-elements/12861

Sereza7: Here is the commit on the PR for XWIKI-20844: https://github.com/xwiki/xwiki-platform/pull/2165/commits/19b1cf391baa5ea4920bd157ee146f138d3e550c

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
Sereza7: See https://github.com/xwiki/xwiki-platform/pull/2182#discussion_r1272206878

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/select/select.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2182/commits/00b30139f8961bb7646c0f6595d6d25eda2a0d69

Sereza7:
Needs an approval/answer from @michitux on https://github.com/xwiki/xwiki-platform/pull/2182#discussion_r1272206878 before merging.
Otherwise, all comments have been addressed.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
michitux: I've just looked at the [official documentation of aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) and found

> If the interface is such that it is not possible to have a visible label on the screen, authors SHOULD use [aria-label](https://w3c.github.io/aria/#aria-label) and SHOULD NOT use [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby).

which to me clearly indicates that in this case we should prefer `aria-label` over `aria-labelledby`. Any reason why you think `aria-labelledby` is preferable here? Also, the fact that `aria-labelledby` and `aria-describedby` can reference hidden content including HTML elements with `hidden` attribute is included in the official [ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_with_aria-labelledby) so it would be very strange if this feature was removed one day. On the other hand, using `sr-only` could lead to content being read twice by screen readers.

Sereza7: > Any reason why you think aria-labelledby is preferable here?

I preferred this solution because the label and description (or hint :upside_down_face: ) fields are handled the same way. Unfortunately, [aria-description is not a thing](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) until ARIA 1.3 gets recommended (the project is not close to finishing, so probably in at least a couple years). I like this similarity to make code easier to understand (I don't think it's important enough to make a proposal out of this though ), but I won't die on this hill, let me know what you think we should do here.


> Also, the fact that aria-labelledby and aria-describedby can reference hidden content including HTML elements with hidden attribute is included in the official [ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_with_aria-labelledby) so it would be very strange if this feature was removed one day.

You got a point. Addressed in https://github.com/xwiki/xwiki-platform/pull/2182/commits/b7aa61e742929b75751607d787824903576bdf25
I updated https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/ to highlight this case :)

PR #2184, opened 2023-05-22
XWIKI-20922: Menu Application layout problem when icons used at top level of horizontal menu

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20922
## PR Changes:
* Fixed the imprecision in the Javascript that caused the regression in the case there's multiple children elements in the content.
* Fixed/cleaned up style for the drop-downs

## View
For the sake of easily viewing, the 'display=none' rule on the dropdowns has been deactivated here in Firefox.
20922-ViewFixed


xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -214,10 +214,16 @@ require(['jquery','xwiki-l10n!menu-ui-translation-keys'], function($, l10n) {
...
-    dropDownHeader.append(dropDownButton);
-    $(this).prepend(dropDownHeader);
-    $(dropDownHeader).next().addClass('xDropdown-menu');
+      let dropDownContent = $(this).contents();
+      // We put all the content of the entry in the header,
+      // except for the last one which is the content of the dropdown. This dropdown stays where it is.
+      for (let index = 0; index &lt; dropDownContent.length -1 ; index++) {
+      let item = dropDownContent[index];
+      dropDownHeader.append(item);
+      }
```
manuelleduc, line 223:
```suggestion
      for (let index = 0; index &lt; dropDownContent.length -1 ; index++) {
        let item = dropDownContent[index];
        dropDownHeader.append(item);
      }
```

Sereza7:
Passed tests locally with 
`mvn clean install -Pquality,integration-tests,docker -f xwiki-platform-core/xwiki-platform-menu -Dxwiki.test.ui.wcag=true -U`.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -809,10 +815,12 @@ require(['jquery','xwiki-l10n!menu-ui-translation-keys'], function($, l10n) {
         li {
           /* Text inside menu */
           color: @dropdown-link-color;
+          display: flex;
+          align-items: center;
+          padding: 3px 20px;
           /* Links inside menu */
           a {
             display: block;
-            padding: 3px 20px;
```
michitux, line 815: Why is moving the padding necessary? Doesn't this make the links smaller and thus harder to click?

michitux (commented):
Looks good to me for the JavaScript logic, I'm less sure about the impact of the CSS changes, see also my comment.

michitux:
In the screenshot, the vertical alignment of the open/close icons looks wrong (too low compared to the text).

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
Sereza7: This ticket was raised because I didn't take into account the fact that users could put multiple elements in the top level menu, I want to make sure there is no issues if submenus are used the same way. If we assume that there is just one element per submenus (anchor or a text title for the subsection), this change is not needed.

Because of the padding on the links, we can't use multiple elements in a submenu without messing up the presentation.
Double margins, no margins, etc...
20922-BothPaddings

Icons can be used **if they are contained by the anchor** but that's it (Note: I'm pretty sure this ticket would have not been raised if the icons were part of the anchors in the top menu).
It makes the style much more difficult because we have to take into account a lot of cases: anchor by itself, no anchor, anchor with icon before, anchor wiht icon after, all of those with and without a flyout submenu (with the carret to keyboard control it).

Is it worth the resource to add event listeners on all the submenu items that redirect the _click_ event to the first link in their content? Should we just assume that there is only one item in a submenu line, and be done with these style complexities.

Sereza7: Extra notes: The events for the menu/submenu flying out are on the hover of the container element. So the links being small would not impact all steps of navigation but only the last one (aka. click the anchor to reach the page, and not all the steps to get this anchor visible and/or in focus).

The :hover background is only set for the link, so it's not looking the best in some interactions.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
```diff
@@ -851,12 +858,6 @@ require(['jquery','xwiki-l10n!menu-ui-translation-keys'], function($, l10n) {
               padding: 0;
               display: inherit;
             }
-            /* Place the arrow on the right */
```
Sereza7, line 854: This block of style is not used anymore, it can be dropped without consequences.

xwiki-platform-core/xwiki-platform-menu/xwiki-platform-menu-ui/src/main/resources/Menu/MenuMacro.xml
Sereza7: Created [a separate ticket](https://jira.xwiki.org/browse/XWIKI-20930) to extend on this reflection and avoid blocking the resolution of the actual issue at stakes here.

Sereza7:
## PR Changes:
* Formatting
* Removed style changes for [XWIKI-20930](https://jira.xwiki.org/browse/XWIKI-20930) 
## Updated view:
(same as before, display=none on the drop-down menu has been deactivated to view the content in the menu)
20922-FixedFinalFinalFinal

manuelleduc (approved):

PR #2185, opened 2023-05-23
XWIKI-19982: Add ability to add a macro by using the keyboard

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-19982

This PR introduces a "{" shortcut to quickly insert Macros from the WYSIWYG editor.
It also skips the macro parameters dialog when no mandatory fields are present to reduce interactions (The user can still edit them afterwards).

autoMacro

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
@@ -28,16 +28,16 @@ ul.macro-list > li.selected {
 }
 
 ul.macro-list .macro-name,
-li.auto-macro-list .macro-name,
+li.macro-suggestion .macro-name,
 ul.macro-parameters .macro-parameter-name {
   color: #333;
   color: $theme.textColor;
 }
 
 ul.macro-list .macro-description,
-li.auto-macro-list .macro-description,
+li.macro-suggestion .macro-description,
 ul.macro-list .macro-extension,
-li.auto-macro-list .macro-extension,
+li.macro-suggestion .macro-extension,
 .macro-editor .macro-description,
 ul.macro-parameters .macro-parameter-description,
 ul.macro-parameters > li.more,
@@ -48,13 +48,14 @@ ul.macro-parameter-group-members > li.more {
   font-size: smaller;
 }
 
-li.auto-macro-list .macro-description {
+li.macro-suggestion .macro-description {
+  /* We do not want the description to affect the width of the macro suggestion menu*/
   width: 0;
   min-width: 100%;
 }
 
 ul.macro-list .macro-extension,
-li.auto-macro-list .macro-extension {
+li.macro-suggestion .macro-extension {
   margin-right: 1em;
 }
 
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
@@ -28,16 +28,13 @@ ul.macro-list > li.selected {
 }
 
 ul.macro-list .macro-name,
-li.macro-suggestion .macro-name,
 ul.macro-parameters .macro-parameter-name {
   color: #333;
   color: $theme.textColor;
 }
 
 ul.macro-list .macro-description,
-li.macro-suggestion .macro-description,
 ul.macro-list .macro-extension,
-li.macro-suggestion .macro-extension,
 .macro-editor .macro-description,
 ul.macro-parameters .macro-parameter-description,
 ul.macro-parameters > li.more,
@@ -48,14 +45,7 @@ ul.macro-parameter-group-members > li.more {
   font-size: smaller;
 }
 
-li.macro-suggestion .macro-description {
-  /* We do not want the description to affect the width of the macro suggestion menu*/
-  width: 0;
-  min-width: 100%;
-}
-
-ul.macro-list .macro-extension,
-li.macro-suggestion .macro-extension {
+ul.macro-list .macro-extension {
   margin-right: 1em;
 }
 
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
@@ -28,13 +28,16 @@ ul.macro-list > li.selected {
 }
 
 ul.macro-list .macro-name,
+li.auto-macro-list .macro-name,
 ul.macro-parameters .macro-parameter-name {
   color: #333;
   color: $theme.textColor;
 }
 
 ul.macro-list .macro-description,
+li.auto-macro-list .macro-description,
 ul.macro-list .macro-extension,
+li.auto-macro-list .macro-extension,
 .macro-editor .macro-description,
 ul.macro-parameters .macro-parameter-description,
 ul.macro-parameters > li.more,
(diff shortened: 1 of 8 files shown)


vmassol:
Thanks for working on this. This looks very nice. Do we have a forum proposal for this?

I'm asking since it's not obvious to me that we need this seen that we're working on quick actions. And in any case we need proposals. Thanks again!

douakli:
Hi Vincent,

> Do we have a forum proposal for this?

We do not have a specific forum proposal for this, I worked on this as part of the Quick Actions proposal since the Quick Actions API is not ready yet, this code can be partly reused for integration in Quick Actions.

One of the possible integration for macros we discussed was using one similar to links, where one of the Quick Actions available is "Insert a Macro" which redirects to a new `{` shortcut used to insert macros.

vmassol:
@douakli thanks for the info. Yes I'm aware you're working on Quick Actions with Marius :) I'm just mentioning that the way we work is by sending proposals. So it would be good to have one for quick actions (if not already done) and one for the topic of this PR (which in addition wasn't planned for this roadmap). See also https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HGeneralDevelopmentFlow

mflorea:
@vmassol this is part of the Quick Actions and there will be a proposal for it.

vmassol:
> @vmassol this is part of the Quick Actions and there will be a proposal for it.

Ok thanks. I was surprised to see a PR before a proposal :)

mflorea:
That's because @douakli is too fast :)

douakli:
I marked this PR as draft because I will be working on integrating it with Quick Actions

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
```diff
@@ -45,8 +48,13 @@ ul.macro-parameter-group-members > li.more {
   font-size: smaller;
 }
 
+li.auto-macro-list .macro-description {
+  width: 0;
+  min-width: 100%;
```
mflorea, line 53: Would be good to add a comment explaining what you want to do this this.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/plugin.js
```diff
@@ -365,6 +365,133 @@
...
+              macros.forEach(function (macro) {
+                // Search macros by name
+                if (macro.name.toLowerCase().includes(opts.query.toLowerCase())) {
+
+                  // Minimal item object
+                  var item = {
+                    id: macro.id.id,
+                    label: macro.name,
+                    description: macro.description,
+                    installed: true,
```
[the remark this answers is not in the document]
douakli: I forgot to remove this key, thank you.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroWizard.css
mflorea: This might not be needed with https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/EditSheet.xml#L822 .

douakli: Indeed, I didn't notice you had added this rule with Quick Actions.

Some of the badges wrap due to the addition of that rule. Do you think it's okay?
image

Maybe the max width rule should be specific to Quick Actions?

douakli:
Hi @mflorea, I pushed the changes to
* Adapt group names to macros
* Avoid double macro entries
* Add badges to Quick Actions
* Remove macro shortcut
* Add installation notification and placeholder widget

I noticed a (I believe) serious issue introduced by this PR, that I'll be investigating tomorrow:
When calling the macroService getMacros function, the select button in the macro modal gets disabled.

image

douakli:
> I noticed a (I believe) serious issue introduced by this PR, that I'll be investigating tomorrow:
When calling the macroService getMacros function, the select button in the macro modal gets disabled.

It looked like a race condition when opening the macro selection modal for the first time. Calling the macroService's getMacros() during initialization caches the response. Which seems to highlight that it is possible for the service to answer before the MacroSelectorAPI has been fully initialized. I decided to stop troubleshooting the race condition and instead introduced a separate cache for Quick Actions because I couldn't find solution to the race condition itself.

mflorea:
> When calling the macroService getMacros function, the select button in the macro modal gets disabled.

Is it disabled only **initially** (because the ``ready`` event is either for fired or not caught) or does it stay disabled when you **click** on a macro (i.e. when you select a macro, because the ``change`` event is not fired or not caught)? In any case I don't see how calling ``macroService.getMacros()`` before the macro selector is initialized can break the macro selector. Even if the list of macros is cached, ``getMacros()`` still returns a promise (deferred object) so the code handling the returned list of macros is still asynchronous (i.e. it doesn't matter if it takes 3 seconds because a real HTTP request was made or 1 millisecond because of the cache returned by the promise).

> Which seems to highlight that it is possible for the service to answer before the MacroSelectorAPI has been fully initialized.

Yes, but they are separate modules. I really don't see how the macro service can influence the macro selector. The macro service returns a promise (deferred), which makes it async. From the point of view of the macro selector it doesn't matter how the list of macros is obtained by the macro service. The macro service could prefetch the list of macros when the editor is loading. This shouldn't affect the macro selector.

>  I decided to stop troubleshooting the race condition and instead introduced a separate cache for Quick Actions because I couldn't find solution to the race condition itself.

I'm not convinced. I'd like to understand the problem before going on with this workaround.

mflorea:
``.data('macroSelectorAPI')`` is set only by ``xwikiMacroSelector()`` which is called in two places only:
* when loading the macro selector, after adding the event listeners for ``ready`` and ``change``
* on click on the macro selector

We can exclude the second, because the selector button is disabled, so the only place is when the macro selector is loaded. Thus I don't see how ``.data('macroSelectorAPI')`` could be set before loading the macro selector.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/plugin.js
```diff
@@ -619,9 +545,110 @@
...
+                order: 9000 + macroGroupsCount * 10
+              });
+              macroGroupsCount++;
+            }
+
+            // Add relevant iconClass
+            attributes.iconClass = categoriesIcons[(categories[0] && categories[0].id)] ||
+                categoriesIcons._fallback;
+
+            // Add badge for each category exceot the first one
```
mflorea, line 607:
```suggestion
            // Add badge for each category except the first one.
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/plugin.js
```diff
@@ -619,9 +545,110 @@
...
+            categories.map(function (category, i) {
+              attributes["badge" + i] = category.label || category;
+            });
```
mflorea, line 610: I don't see the need for ``map``. Using ``forEach`` makes more sense in this case I think. Also, where exactly do you exclude the first category, as the comment implies?

I guess using ``"badge" + i`` helps with the template, but how do you know how many badges there are? Also, why the fallback on ``category``? Isn't ``category`` and object? Maybe fallback on category id?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
```diff
@@ -30,7 +30,12 @@
     constructor(config) {
       this.config = Object.assign({
         // Default values for quick action / group fields.
-        defaultAction: {group: '', id: '', name: '', description: '', iconClass: '', iconURL: '', shortcut: ''},
+        defaultAction: {group: '', id: '', name: '', description: '',
+          iconClass: '', iconURL: '', shortcut: '',
+          //Default value for badges
+          ...Array(3).fill().map(function (_, i) {return i;}).reduce(function (acc, element)
```
mflorea, line 36: I take it that there can be at most 3 badges, right? A comment would help.

PR #2183, opened 2023-05-21
Update stop_xwiki.bat

amanna13:
Fixed typos. Changed enviroment to environment on line 22.

PR #2186, opened 2023-05-23
Done the changes as it was previously.

amanna13:


surli:
Please don't open multiple pull requests unless you have a good reason to do so. You already have  #2183 opened for this. Also the changes you proposed here are wrong. If you want to contribute check https://dev.xwiki.org/xwiki/bin/view/Onboarding/ and come talk to us instead of opening random pull requests. Thanks.

amanna13:
Thank you for your feedback and guidance. 
I apologize for opening multiple pull requests for the same changes. I will make sure to adhere to the project's guidelines moving forward.

PR #2187, opened 2023-05-24
XWIKI-20892: Search form facets are not properly labelled

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20892

## PR Changes

* Added a label to the facet inputs.
* Added appropriate attributes to the label and input elements

## Extra changes related to the ticket
Those changes were not reported in the ticket itself, however they fix other small accessibility issues on the Solr search page, so I decided to include them here. Comments on the original ticket report with more details 
* Removed a harmful alt on the user facet image.
* Changed the DOM order of the facet entry -- replaced the floating layout by a flex display.
* Added a label for the search input form at the top of the page.

## View
**No change to view**:
20892-ViewAfter

## Tests
Successfully passed `mvn clean install -Pquality,integration-tests,docker -f xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/ -Dxwiki.test.ui.wcag=true`.

## Notes

* In order to keep the "Open in a new tab" functionality of the anchor, I chose not to replace it with a label like what I originally thought would be the good solution. [It's not possible to place an anchor inside a label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) and [it's not recommended to place a label inside of an anchor](https://stackoverflow.com/questions/21635431/wrap-inputs-label-in-hyperlink-anchor-tag). So I decided to duplicate the content in the anchor to put it in the label and give the label a `sr-only` (does not take visual space but is still active) class. See the ticket comments for more details.
* For the label of the search input form, this didn't trigger a WCAG violation, but it's a best-practice reported by axe-core to ensure all form inputs are properly labelled, not just with a title that might not be used by every assistive techs. So I just put the content of the title attribute we already used, inside a `label` element.
* Some accessibility issues with the mediatype facet have not been addressed here, but I raised a ticket to handle them separately: https://jira.xwiki.org/browse/XWIKI-20931


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
```diff
@@ -198,14 +198,33 @@
     }
   };
 
-  var addFacetValueCheckbox = function() {
-    var checkBox = $(document.createElement('input')).attr('type', 'checkbox');
+  var addFacetValueCheckbox = function(index) {
+    // Create an id unique to the facet value.
+    let facetContainer = $(this).parents('.search-facet').first();
+    let checkBoxId = facetContainer.attr('data-name') + '-' + index.toString() + '-checkbox';
```
michitux, line 204: What about checking if an element with the id exists already, just to be sure?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -67,8 +67,11 @@
     &lt;/div&gt;
     &lt;div class="col-xs-12 col-sm-6"&gt;
       &lt;div class="input-group"&gt;
-        &lt;input type="search" name="text" class="form-control withTip useTitleAsTip"
+        &lt;input id='search-page-bar-input' type="search" name="text" class="form-control withTip useTitleAsTip"
```
michitux, line 70: Do we have any policy how to name such ids? Also, do we really need an additional label? From my understanding of https://www.w3.org/TR/WCAG20-TECHS/H65.html this shouldn't be needed.

michitux (commented):
Overall, this looks good but I have some comments/questions.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
```diff
@@ -198,14 +198,33 @@
...
+    let checkBox = $(document.createElement('input')).attr('type', 'checkbox');
+    checkBox.attr('id', checkBoxId);
+
+    // Initialize the checkbox label.
+    let checkBoxLabel = $(document.createElement('label'));
+    checkBoxLabel.attr('for', checkBoxId);
+    checkBoxLabel.addClass('sr-only');
+    // We duplicate the content of the anchor in the facet value instead of having it be a label
+    // because we want to keep the use case "Open in a new page" possible.
+    $(this).contents().clone().appendTo(checkBoxLabel);
```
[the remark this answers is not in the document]
michitux: According to https://dequeuniversity.com/rules/axe/4.6/label-title-only, using `aria-labelledby` should be sufficient.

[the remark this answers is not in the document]
Sereza7: > Also, did you consider using a title attribute instead of the label? I'm not sure this would be better, I'm just wondering if it is really good to duplicate the link content.

See https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#accessiblenameguidancebyrole
They recommend adding a `label` to `type=checkbox` inputs without text content.

____
> I assume it is not sufficient to use an aria-labelledby attribute with a reference to the link? [...] 
According to https://dequeuniversity.com/rules/axe/4.6/label-title-only, using aria-labelledby should be sufficient.

The label is an HTML functionality, it's always better to use raw HTML than ARIA:
> If you can use a native HTML element [[HTML51](https://www.w3.org/TR/using-aria/#bib-html51)] or attribute with the semantics and behavior you require already built in, instead of re-purposing an element and adding an ARIA role, state or property to make it accessible, then do so.

From https://www.w3.org/TR/using-aria/#rule1

So if it's not too much trouble, I'd rather use a `label` than `aria-labelledby`. Usually in XWiki, it's more annoying because we need to generate unique ids, and it's not straightforward to get them.

___
> I'm just wondering if it is really good to duplicate the link content.

I wondered too, but it seems to me like it's necessary if we want the "false label" to be an anchor (and keep its 'Open in another tab' functionality)

See the first comment I added on the PR opening post (not clearly worded though x) ) :

> In order to keep the "Open in a new tab" functionality of the anchor, I chose not to replace it with a label like what I originally thought would be the good solution. [It's not possible to place an anchor inside a label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) and [it's not recommended to place a label inside of an anchor](https://stackoverflow.com/questions/21635431/wrap-inputs-label-in-hyperlink-anchor-tag). So I decided to duplicate the content in the anchor to put it in the label and give the label a sr-only (does not take visual space but is still active) class. See the ticket comments for more details.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
Sereza7: > Also, do we really need an additional label?

It's not "needed", but [recommended by WAI ARIA](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#accessiblenameguidancebyrole) to get label for those elements.

And also recommended by HTML:
> Relying on the [title](https://html.spec.whatwg.org/multipage/dom.html#attr-title) attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

From https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute


Note that the title attribute is nice, but some AT don't support it properly, so the rule of thumb I kept is: have at least one kind of 'real' label (label, aria-labelledby, aria-label) and if you don't mind the hover popup, add a title with the same content.
A couple of sources to support this: 
* https://adrianroselli.com/2020/01/my-priority-of-methods-for-labeling-a-control.html
* https://www.tpgi.com/using-the-html-title-attribute/

Sereza7: > Do we have any policy how to name such ids?

I just figured such an id would be okay since it's quite specific and found only once on the page when testing. I don't know of any name policy for ids.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
Sereza7: Note: If the facetcontainer data-name is not unique, the setSearchUIState bugs out anyways.

Sereza7: The `index` part is unique to the group of checkboxes in the same container (generated by the jquery `each` on [L252](https://github.com/xwiki/xwiki-platform/blob/f86702aba120ce73d099b8dd3385d1d293084e45/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml#L252)).

So the id generated should be unique anyways.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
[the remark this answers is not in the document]
michitux: Have you checked how this is read in a screen reader? Won't the screen reader read the label twice now?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
michitux: > Note that the title attribute is nice, but some AT don't support it properly, so the rule of thumb I kept is: have at least one kind of 'real' label (label, aria-labelledby, aria-label) and if you don't mind the hover popup, add a title with the same content. A couple of sources to support this:
> 
> * https://adrianroselli.com/2020/01/my-priority-of-methods-for-labeling-a-control.html
> * https://www.tpgi.com/using-the-html-title-attribute/

At least the second resource doesn't support this. The hidden label element has the exactly same accessibility problems, actually, additionally, it cannot even be accessed with a pointing device, so the accessibility is even worse. From the second link:

> For a label on a control that has no visible text label: `<input type="text" title="name">`
> Screen readers users will have access to the control label, as the title attribute is mapped to the accessible name property in accessibility APIs (when a text label using the label element is not supplied ). Many other users will not. Recommend including a visible text label whenever possible.

So it explicitly states that screen reader users have access to the title. What you're proposing here isn't better in any way from what I understand as it is even less visible than the title attribute.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrUserFacet.xml
```diff
@@ -42,7 +42,7 @@
   #set ($escapedUserName = $escapetool.xml($xwiki.getPlainUserName($userReference)))
   #getUserAvatarURL($userReference $avatarURL 30)
   &lt;span class="user"&gt;##
-    &lt;img class="user-avatar" src="$escapetool.xml($avatarURL.url)" alt="" /&gt;##
+    &lt;img class='user-avatar' src="$escapetool.xml($avatarURL.url)" alt='' /&gt;##
```
michitux, line 45: This change looks unrelated.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -67,8 +67,11 @@
...
           title="$services.localization.render('search.page.bar.query.title')" value="$escapetool.xml($text)"/&gt;
+        &lt;label class="sr-only" for='search-page-bar-input'&gt;
+          $services.localization.render('search.page.bar.query.title')
+        &lt;/label&gt;
```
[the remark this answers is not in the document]
michitux: To summarize this, I think we should either let the browser use the title as fallback or add a shorter label that is different from the title.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrUserFacet.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2187/commits/347fb7f510c1139738e466662fa735b9d5db43f0

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2187/commits/7c4005a71dc38a98189bd6d511099567b7bbc96a

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2187/commits/7c4005a71dc38a98189bd6d511099567b7bbc96a
See https://github.com/xwiki/xwiki-platform/pull/2187#discussion_r1235402944

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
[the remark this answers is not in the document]
Sereza7: > Why is this label after the input?

It's a mistake, the label--input order is reversed for the checkboxes (label after the input), which explains why I was doing it this way by default in this PR. Addressed in https://github.com/xwiki/xwiki-platform/pull/2187/commits/a49534d6132a986d32404626e9521553c820bdd3.

> Also, I'm wondering if "Enter your search query" is a good label. The whole point of "Rule 4: Avoid Browser Fallback" in https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#cardinalru is "Because the purpose of these attributes is not naming, their content typically yields low quality accessible names that are not effective.". If we just re-use the title, it seems there is little value added.

Here, I think the title content we use is already short and good for a label. 
It would be read by NVDA as "Enter your search query; edit; <value>" where <value> is nothing when first loading the page. I didn't test it myself, but checked [this page about support of input[type="search"] by screen readers](https://a11ysupport.io/tests/tech__html__input__input-search#assertion-html-input(type-search)_element-provide_shortcuts-).


We have the hand over the value that we use in the label and I think this value used in the title is of good enough quality, at least in english. Note that there's a lot of places where we already use title and label indiscriminately. We discussed it a bit with Vincent in the past, and the difference seemed too subtle to matter. So far from what I saw in xwiki-platform titles are appropriate names. If needed, a change to the title value can be done, since there is usually no strong constraints on it (we're just adding the strong constraint 'be a proper accessible name' to a loosely defined set of constraints that pretty much match our new constraint: be short, be understandable, be to the point, ...).

The value added by this change is minimal, but it ensures that all browser/AT combos react the way we want it to. The title is a common fallback but it might not be perfectly supported at all times, and labels are a way more standard alternative to naming inputs.

Sereza7:
## PR Updates:
* Replaced the duplicated label with an `aria-labelledby` attribute on the input
* Reverted unrelated changes
* Changed wrong order for an input--label pair

## Tests
Successfully passed `mvn clean install -f xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/ -Pdocker,integration-tests,quality -Dxwiki.test.ui.wcag=true -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false`.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
```diff
@@ -198,14 +198,21 @@
...
 
-  var addFacetValueCheckbox = function() {
-    var checkBox = $(document.createElement('input')).attr('type', 'checkbox');
+  var addFacetValueCheckbox = function(index) {
+    // Create an id unique to the facet value.
+    let facetContainer = $(this).parents('.search-facet').first();
+    $(this).attr('id', facetContainer.attr('data-name') + '-' + index.toString());
+    // Initialize the checkbox.
+    let checkBox = $(document.createElement('input')).attr('type', 'checkbox');
+    checkBox.attr('aria-labelledby',$(this).attr('id') );
```
michitux, line 207:
```suggestion
    checkBox.attr('aria-labelledby', $(this).attr('id'));
```

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
```diff
@@ -198,14 +198,21 @@
...
     checkBox.prop('checked', $(this).hasClass('selected'));
     // Add the 'checked' attribute so that it can be easily located with CSS.
     checkBox.prop('checked') &amp;&amp; checkBox.attr('checked', 'checked');
     checkBox.on('click', this.click.bind(this));
-    // Remove the 'selected' class because the selected state is marked using the check box.
-    $(this).removeClass('selected').before(checkBox);
+    // Remove the 'selected' class because the selected state is already marked using the check box.
+    $(this).removeClass('selected')
+    // Add the checkbox to the DOM
+    $(this).before(checkBox)
```
michitux, line 215: This change doesn't seem to do anything. What about just reverting it?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -67,7 +67,10 @@
     &lt;/div&gt;
     &lt;div class="col-xs-12 col-sm-6"&gt;
       &lt;div class="input-group"&gt;
-        &lt;input type="search" name="text" class="form-control withTip useTitleAsTip"
+        &lt;label class="sr-only" for='search-page-bar-input'&gt;
```
michitux, line 70: Minor point, but it seems inconsistent to use `'` and `"` for different attributes of the same element.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
```diff
@@ -67,7 +67,10 @@
...
+        $services.localization.render('search.page.bar.query.title')
+        &lt;/label&gt;
+        &lt;input id='search-page-bar-input' type="search" name="text" class="form-control withTip useTitleAsTip"
```
michitux, line 73: Same here, it seems inconsistent to use `'` and `"` for different attributes of the same element.

michitux (approved):
Some minor suggestions but overall this looks good now.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
Sereza7: I think it helped made the code clearer, and it was close enough to code I modified here.

When I first went through the code, I couldn't spot at a glance where the element was added to the DOM, since this was done on the end of a line that was said to do something else.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
Sereza7: Yeah, I should have used single quotes for the class. I think in some cases my editor autocompletes attributes with double quotes, I need to see how to change this :)
Addressed in https://github.com/xwiki/xwiki-platform/pull/2187/commits/63d5756b9f44d528d90c23acc755f423acceb08a

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2187/commits/013015e347f52b417dc472e6eef7802c2e25d65e

PR #2188, opened 2023-05-24
XWIKI-20932: Images pasted from clipboard are attached with 'grafik.png' name on Firefox with German language

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20932

Currently, we decide if we are in the presence of a pasted image based on its name (i.e., `image.png`) which proved to be inaccurate and locale dependent.

Instead, I propose to toggle a `duringPaste` flag based on two listeners, `beforePaste` and `afterPaste`, respectively turning setting the flag `true` and `false`.
We append a timestamp and a random integer to the filename when  `duringPaste` is `true`.

Note: Not fully cleaned up, I'd like to have some opinion of the approach before.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-upload/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-upload/plugin.js
@@ -59,15 +59,15 @@
         return startMarker + image + stopMarker;
       });
 
-      // Chrome uses 'image.png', or a localized form (e.g., 'graphik.png' in German) as file name when pasting images,
-      // instead of leaving the file name empty, and this prevents us from using the configured default file name.
-      // Keeping the same name would override again and again the same attachment, which is problematic when more than
-      // one image is pasted to the same page.
-      // Instead, we integrate the timestamp and a random number between the filename and the file extension.
-      // For instance, 'image.png' becomes 'image-1684887709-691.png'
-      // See https://github.com/ckeditor/ckeditor-dev/issues/664
+      // Browses uses 'image.png', or a localized form (e.g., 'graphik.png' in German Firefox) as file name when pasting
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-upload/plugin.js
```diff
@@ -59,16 +59,36 @@
...
+      
+      var duringPaste = false;
+
+      editor.on('beforePaste', function () {
+        duringPaste = true;
+      });
+
+      editor.on('afterPaste', function () {
+        duringPaste = false;
+      });
```
manuelleduc, line 79: Add a comment to explain what's going on here

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/resources/xwiki-upload/plugin.js
```diff
@@ -59,16 +59,36 @@
...
+        if (duringPaste) {
+          var splitName = file.name.split(".");
+          // Note: pop is stateful and needs to be executed before join.
+          var extension = splitName.pop();
+          var fileName = splitName.join(".");
+          var timestamp = Date.now();
+          var min = 100;
+          var max = 999;
+          var random = Math.floor(Math.random() * (max - min + 1)) + min;
+          name = fileName + '-' + timestamp + '-' + random + '.' + extension;
```
manuelleduc, line 91: more to a method and or explain what's done here

mflorea:
@manuelleduc this change in behaviour is significant so I think it deserves a vote.

manuelleduc:
> @manuelleduc this change in behaviour is significant so I think it deserves a vote. Putting that aside, the proposed change looks good.

Good idea, here is the proposal https://forum.xwiki.org/t/name-of-images-pasted-in-ckeditor/12581

PR #2190, opened 2023-05-25
XWIKI-20913: More actions dropdown structure is not accessible

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20913

## PR Changes

* Replaced the <ul> with a more semantically precise structure based on a <dl>
* Updated style

## Notes

* In xwiki-platform, the submenuheader velocity macro is only used for this dropdown. Those changes might create issues for extensions relying on this macro.
* Added a condition on menu items to not add the class atttribute if it's empty (looks less cluttered in the HTML markup).

## View

On the left, the dropdown before the changes (example from xwiki.org), on the right, the dropdown after the changes.
We can see on this screenshot that the HTML structure is radically different, while the visuals are very similar.
20913-ComparisonStructure

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
@@ -164,9 +164,17 @@
           <span class="sr-only">$services.localization.render('core.menu.toggleDropdown')</span>
         </button>
       #end
-      <ul class="dropdown-menu #if ("$!actionUrl" == '')dropdown-menu-right#end">
-        $menuContent
-      </ul>
+      #if ("$!actionUrl" == '')
+        ## We use another component for the more action menu, since it has
+        ## a more complex structure with submenus.
+        <dl class='dropdown-menu dropdown-menu-right'>
+          $menuContent
+        </dl>
+      #else
+        <ul class='dropdown-menu'>
+          $menuContent
+        </ul>
+      #end
     #end
   </div>
 #end
@@ -220,8 +228,8 @@
   #setVariable("$count", $count)
 #end
 
-#macro (submenuheader $titleKey)
-  <li class="dropdown-header">$services.localization.render($titleKey)</li>
+#macro (submenuheader $titleKey, $needsSeparator)
+  <dt class="dropdown-header#if($needsSeparator) separator-above#end">$services.localization.render($titleKey)</dt>
 #end
 
 #**
@@ -402,7 +410,9 @@
       #set ($adminActions = "#displayAdminActions()")
       #if ($stringtool.isNotBlank("$!adminActions"))
         #submenuheader('core.menu.actions.main')
-        $adminActions
+        <dd><ul>
+          $adminActions
+        </ul></dd>
         #set ($needsSeparator = true)
       #end
     #end
@@ -410,11 +420,10 @@
     #if ($displayMoreActionsMenu)
       #set ($moreActions = "#displayMoreActions()")
       #if ($stringtool.isNotBlank("$!moreActions"))
-        #if ($needsSeparator)
-          #submenuseparator()
-        #end
-        #submenuheader('core.menu.actions.others')
-        $moreActions
+        #submenuheader('core.menu.actions.others', $needsSeparator)
+        <dd><ul>
+          $moreActions
+        </ul></dd>
         #set ($needsSeparator = true)
       #end
       ## ----------------------------
@@ -423,20 +432,19 @@
       #if($xwiki.hasAccessLevel('view') && $displayShortcuts)
         #set ($viewersContent = "#template('shortcuts.vm')")
         ## Only display a separator if at least one menu entry has been displayed before
-        #if ($needsSeparator)
-          #submenuseparator()
-        #end
-        #submenuheader('core.menu.actions.viewers')
+        #submenuheader('core.menu.actions.viewers', $needsSeparator)
         ##
         ## VIEW SOURCE
         ##
         #if ("$!request.rev" != '')
           #set ($revisionParameter = "&rev=$request.rev")
         #end
-        #submenuitem($doc.getURL('view', "viewer=code$!{revisionParameter}") $services.localization.render('core.menu.view.source') 'tmViewSource', '', 'search', $extraAttributes)
-        $viewersContent
-        ## Display Viewers menu UIX..
-        #displaySecureUIX('org.xwiki.platform.template.menu.viewers', [])
+        <dd><ul>
+          #submenuitem($doc.getURL('view', "viewer=code$!{revisionParameter}") $services.localization.render('core.menu.view.source') 'tmViewSource', '', 'search', $extraAttributes)
+          $viewersContent
+          ## Display Viewers menu UIX..
+          #displaySecureUIX('org.xwiki.platform.template.menu.viewers', [])
+        </ul></dd>
         #set ($needsSeparator = true)
       #end
     #end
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
```diff
@@ -220,8 +228,8 @@
   #setVariable("$count", $count)
 #end
 
-#macro (submenuheader $titleKey)
-  <li class="dropdown-header">$services.localization.render($titleKey)</li>
+#macro (submenuheader $titleKey, $needsSeparator)
```
manuelleduc, line 231: The comma is not required between velocity macro params.
```suggestion
#macro (submenuheader $titleKey $needsSeparator)
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
```diff
@@ -402,19 +410,20 @@
       #set ($adminActions = "#displayAdminActions()")
       #if ($stringtool.isNotBlank("$!adminActions"))
         #submenuheader('core.menu.actions.main')
-        $adminActions
+        <dd><ul>
+          $adminActions
+        </ul></dd>
```
manuelleduc, line 415:
```suggestion
        <dd>
          <ul>
            $adminActions
          </ul>
        </dd>
```

michitux:
I'm wondering if this whole `$needsSeparator` logic is required at all. From what I can see, the purpose is just to not add a separator above the first `dt`. Wouldn't it be simpler to remove the separator with a `:first-child`-selector in CSS?

Sereza7:
## PR Changes
* Replaced a style class with a CSS selector
* Fixed formatting to be consistent with https://github.com/xwiki/xwiki-platform/pull/2190#discussion_r1211756253

No additional changes to style and behavior

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
```diff
@@ -400,48 +408,50 @@
...
-        $viewersContent
-        ## Display Viewers menu UIX..
-        #displaySecureUIX('org.xwiki.platform.template.menu.viewers', [])
-        #set ($needsSeparator = true)
+        <dd>
+          <ul>
+            #submenuitem($doc.getURL('view', "viewer=code$!{revisionParameter}") $services.localization.render('core.menu.view.source') 'tmViewSource', '', 'search', $extraAttributes)
+            $viewersContent
+            ## Display Viewers menu UIX..
+            #displaySecureUIX('org.xwiki.platform.template.menu.viewers', [])
```
michitux, line 452: I've just had a look at this macro's code and unfortunately, it seems like this code could call `#submenuseparator()`. Is this a problem with these changes or would this still work?

michitux: I'm not sure it is good to just remove the separator logic. Are you sure this isn't needed? Also, what happens if a UIX just calls `#submenuheader` (which I assume is the idea when using a separator)? Won't this break the HTML? On the other hand, https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/Viewers%20Menu/ doesn't even mention the parameter. However, this very same macro is used in https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/Edit%20Actions where it matters I think.

Sereza7:
## PR Updates
* Reverted the suppression of the separator logic
* Created a new macro to contain the change to the separator needed for the action menu.
* Updated the edit menu to use this macro
* Changed the LESS for separators to display *under* groups instead of *above* titles, so that structures without dt can still have separators.

## View
20913-EditAfter

## Note
With things as of now, every use of the velocity macro `#displaySecureUIX` supposes that the menu follows the new structure.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
```diff
@@ -168,9 +168,11 @@
           <span class="sr-only">$services.localization.render('core.menu.toggleDropdown')</span>
         </button>
       #end
-      <ul class="dropdown-menu #if ("$!actionUrl" == '')dropdown-menu-right#end">
+      ## We use another component for the more action menu, since it has
+      ## a more complex structure with submenus.
+      <dl class='dropdown-menu dropdown-menu-right'>
```
michitux, line 173: Why did you remove this check on the action URL?

michitux (commented):
Maybe I missed something, but won't this cause invalid HTML because you have `dd` children as first child of the `dl`?

Sereza7:
This will cause invalid HTML 5: https://stackoverflow.com/questions/3275054/does-definition-list-dl-require-that-each-dd-will-have-dt-tag

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/menus_content.vm
Sereza7: Long story short, I made some changes and 'reverted' them, but left a diff afterwards.
This class changes the alignement of the dropdown relative to the button, it was not critical but I didn't notice it when comparing them on: https://github.com/xwiki/xwiki-platform/pull/2190#issuecomment-1653152654
Addressed in https://github.com/xwiki/xwiki-platform/pull/2190/commits/09d9f3b1600f44cfb67bcc5e7335a57b55837889

Sereza7:
## PR Changes
* Fixed a regression on style
* Added a description term for the edit menu
* Escaped properly all the submenu headers

## View
20913-descriptionterm
## Note
I decided to go with a single \<dt> for the whole edit menu. This is enough for making valid HTML5 and all the \<dd> can use the same description anyways. The point @michitux raised in https://github.com/xwiki/xwiki-platform/pull/2190#pullrequestreview-1549552547 was addressed in https://github.com/xwiki/xwiki-platform/pull/2190/commits/433f9b1613ac7f13e309e580fa0245056193a296

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -976,6 +976,7 @@ core.menu.translate=Translate
 core.menu.translate.hint=Translate this page in {0}
 core.menu.translate.details=You''re viewing the {0} translation of this page (its original version) because this page has not been translated in {1} (the current locale) yet.
 core.menu.edit=Edit
+core.menu.edit.dropdown.hint=Select an edit mode to use.
```
michitux, line 979: I wonder if something like "Edit modes" wouldn't be more consistent with the short submenu headings in the more actions menu (that just has "Manage", "Actions", "Viewers").

michitux (approved):
Looks good now, I just wonder if the hint for the edit menu shouldn't be shorter.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
Sereza7: You're right, and the use of a verb here is not very appropriate for this kind of label. :+1: 
Addressed in https://github.com/xwiki/xwiki-platform/pull/2190/commits/16fc7ba863d7c5652e50a7a4ddbb3f6e8d9ca15e

PR #2194, opened 2023-05-31
XWIKI-20919: The document is deleted when it has only one version if a revision deletion action is triggered with a non-existing version

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-20919

Note that it targets LTS. 

  * Add checks in XWiki#deleteDocumentVersions and XWikiDocumentArchive#removeVersions to perform removal is the range of version is not empty and to avoid reseting the entire archive in such case
  * Perform some refactoring in XWikiDocumentArchive to document properly getNodes and make it more clear on the meaning of its arguments. Rename #getNearestFullVersion also since the name is wrong: that method does not return the nearest full version, it returns the next one

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
@@ -33,6 +33,7 @@ import java.util.TreeSet;
 import org.suigeneris.jrcs.rcs.Version;
 import org.suigeneris.jrcs.util.ToString;
 import org.xwiki.model.reference.WikiReference;
+import org.xwiki.stability.Unstable;
 
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
```diff
@@ -453,30 +478,47 @@ public Version getPrevVersion(Version ver)
...
+     */
+    @Unstable
+    public Version getNextFullVersion(Version ver)
     {
         if (this.fullVersions.contains(ver)) {
             return ver;
         }
         SortedSet<Version> headSet = this.fullVersions.headSet(ver);
-        return (headSet.size() == 0) ? null : headSet.last();
+        return (headSet.isEmpty()) ? null : headSet.last();
```
manuelleduc, line 507:
```suggestion
        return headSet.isEmpty() ? null : headSet.last();
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
```diff
@@ -453,30 +478,47 @@ public Version getPrevVersion(Version ver)
...
-     * @return nearest version which contain full information (not patch)
+     * @deprecated This method has been deprecated as its name is error-prone, and it has never done what's documented:
+     * it always returned the next full version, and never the nearest one. So we renamed it to
+     * {@link #getNextFullVersion(Version)}.
+     *
+     * @param ver - for what version find next full one
+     * @return next version which contain full information (not patch)
      */
+    @Deprecated(since = "14.10.12,15.5RC1")
     public Version getNearestFullVersion(Version ver)
```
manuelleduc, line 489: Is this method still used somewhere?

surli: To my knowledge no but was a bit lazy to create the aspect just for it...

manuelleduc: yeah, I can related :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
```diff
@@ -453,30 +478,47 @@ public Version getPrevVersion(Version ver)
     }
 
     /**
-     * @param ver - for what version find nearest
-     * @return nearest version which contain full information (not patch)
+     * @deprecated This method has been deprecated as its name is error-prone, and it has never done what's documented:
```
tmortagne, line 481: `@deprecated` is generally the last part of the javadoc

tmortagne (approved):

manuelleduc (approved):

PR #2197, opened 2023-06-02
XWIKI-20755: Livedata filter inputs don't have labels

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20755

## PR Changes
* Added translated labels to the filter inputs.
* Fixed unit test for FilterList.spec.js that it was breaking
* Updated the translation promises in Logic.js

## Tests
Passes the tests: `mvn clean install -Pquality,integration-tests,docker -f xwiki-platform-core/xwiki-platform-livedata/ -Dxwiki.test.ui.wcag=true`

## View
No visual UI change
20755-Demo
We can see that the filters of the livedata on this page are now properly labelled.

## Notes
I had trouble updating the unit tests for FilterList.vue . Making them pass without hacks took me time since I didn't know well the vue templating langage.


michitux (commented):
I don't think these labels are useful. I think the labels should rather be something like `Filter for the "{0}" column.` where the parameter is the column name. Or would this be too much redundancy as screen readers probably already provide column names?

Sereza7:
@michitux 
See [this short discussion on stackoverflow](https://stackoverflow.com/questions/65847633/nvda-says-table-columns-name-an-extra-time-when-a-link-is-present).

Checking behaviour on the couple of most used screen readers:
* JAWS 

> If a table has been properly marked up with column and row headers, when moving left and right in a table, you would hear the item in each column at the top of the column and then the contents of the current cell. When moving up and down in a table, you hear the item at the beginning of each row and then the contents of the current cell.

from https://www.freedomscientific.com/SurfsUp/Tables.htm
(JAWS is not free to use, and it's a hassle already to open up a windows VM just for screen reading, so I didn't test it out myself)
* NVDA
Navigate with keys, and in a similar fashion, it will read out the content of the column header when moving.
Test demo https://up1.xwikisas.com/#l1XOVVUqfjwiy-KFPsZfTg

____
Having an input in a td is a pretty niche structure, I couldn't find patterns or discussions about it and how to make it accessible.
The issue I see so far is:
1. when going through the page with Tab (interactive elements only): the input is selected, only its label is read out
2. when going through the page with other means (especially table navigation with NVDA: T -> Ctrl+Alt+Arrows ): the table header is read out, then the content of the table cell (here, just the input)

Either we add the column name in the cell description, and 1. is perfect but 2. has the column name repeated twice.
Or we don't add it and 2. is perfect but 1. is imprecise.

Since those two solutions seem similar in effectiveness, I'd advocate for the simplest solution (leaving the column name out of the input label) and leave it at that. More precision will add complexity to the table and produce a bit more technical debt.

I think actual screen reader users are not navigating through a new page with Tab by default. Maybe some power users that expect to find the filters after X tabs, but I doubt a precise title would be useful to them. Moreover, power users wouldn't have trouble changing their reading mode to find the table cell above the input and get the info they want.

When making the demo for NVDA, I just realized that the button names are read out in the header titles, making them really long and hard to understand. I'm reporting it as a new issue since the actual value of the table header has not much to do with the WCAG violation this PR fixes.

PR #2198, opened 2023-06-02
XRENDERING-670: Image captions cannot be used in combination with links on images

michitux:
* Handle images that are wrapped inside links for figure styles.
* Refactor the code to move common code inside the chaining listeners to a base class.
* Add a functional test to verify that image captions and links can actually be used together via CKEditor.

Jira issue: https://jira.xwiki.org/browse/XRENDERING-670


xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-style/xwiki-platform-image-style-api/src/main/java/org/xwiki/image/style/internal/rendering/AbstractCaptionedImageChainingListener.java
```diff
@@ -0,0 +1,112 @@
...
+        Map<String, String> imageParameters);
+
+    private QueueListener.Event getPreviousFigureEvent()
+    {
+        QueueListener.Event result = null;
+
+        QueueListener.Event candidate = getPreviousEvents().peekLast();
+        if (candidate != null && candidate.eventType == EventType.BEGIN_LINK) {
+            result = candidate;
+            candidate = getPreviousEvents().get(getPreviousEvents().size() - 2);
```
michitux, line 104: This together with the increased lookahead (2 vs. 1) is the main change, the rest is refactoring.

PR #2200, opened 2023-06-05
XWIKI-18001: Provide a textual description of the video on the home page for accessibility + provide title

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-18001
## PR Changes
* Added a label to the video iframe
* Added a summary and a transcript as the video iframe description. 

## Tests
Passed `mvn clean install -Pquality,integration-tests,docker -f xwiki-platform-distribution/xwiki-platform-distribution-flavor/ -Dxwiki.test.ui.wcag=true`

## View
No changes to the view

## Note
I decided to add proper semantics inside the transcript that would otherwise be more difficult to understand

### DIFF
--- a/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-common/src/main/resources/Main/WebHome.xml
+++ b/xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-common/src/main/resources/Main/WebHome.xml
@@ -44,7 +44,71 @@
 ## The workaround is to make sure the submitted content (the macro markers) don't match the preview HTML.
 ## See XWIKI-14818: Preview triggers ERR_BLOCKED_BY_XSS_AUDITOR in Chrome
 #set ($iframe = 'iframe')
-&lt;$iframe src="https://www.youtube.com/embed/9QTWrZ7OfzI" title="YouTube video of XWiki" role="presentation" allowfullscreen&gt;&lt;/$iframe&gt;
+&lt;$iframe src="https://www.youtube.com/embed/9QTWrZ7OfzI" title="YouTube video of XWiki" allowfullscreen
+  aria-labelledby='xwikipresentationvideolabel' aria-describedby='xwikipresentationvideodescription'&gt;&lt;/$iframe&gt;
+&lt;div id='xwikipresentationvideolabel' class='sr-only'&gt;Video presentation of XWiki&lt;/div&gt;
+&lt;dl id='xwikipresentationvideodescription' class='sr-only'&gt;
+  &lt;dt&gt;
+    Summary:
+  &lt;/dt&gt;
+  &lt;dd&gt;
+    XWiki is a flexible and reliable Enterprise Wiki, allowing organizations of all sizes to better collaborate and exchange information in a timely and cost-effective manner.
+    XWiki makes use of all the features you would expect from an Enterprise Wiki. Ergonomic WYSIWYG Editing, Page linking, History and rollback, User and rights management, File attachments,  Macros, and advanced search using the Apache SOLR engine, just to name a few.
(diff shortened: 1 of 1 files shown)


michitux:
I'm not really convinced and I think this needs a broader discussion. First of all, this "transcript" is not a transcript of the audio but the content that is visible in the video. I think this needs to be clear. Also, I'm not sure how the summary relates to the transcript, as the summary doesn't summarize the transcript in my opinion. As we already perform changes around this video, I think we should take the occasion to get rid of the dependency on YouTube (but again, something to be discussed).

Also, this content seems to be mainly an advertisement for XWiki SAS. This is okay according to the [governance](https://dev.xwiki.org/xwiki/bin/view/Community/Governance) I think, but I find it questionable to have this content in a place where it would need to be translated by community contributors.

vmassol:
I agree that we should discuss getting rid of the youtube video for several reasons:
* https://jira.xwiki.org/browse/XWIKI-14858 (the home page is complex and scary when editing it by a new user)
* https://jira.xwiki.org/browse/XWIKI-15194
* And of course the WCAG issue at https://jira.xwiki.org/browse/XWIKI-18001

> This is okay according to the [governance](https://dev.xwiki.org/xwiki/bin/view/Community/Governance) I think

Actually I don't think it is. The places where the TSC is allowed to perform advertising are clearly defined and the XS home page is not listed AFAICS.

TBH I thought that the video was not related to XWiki SAS. I had not watched it for the past 10+ years and just watched it now and indeed it's related to XWiki SAS, especially with the xwiki.com link at the end (but also showing some XWiki SAS projects).

So +1 to discuss it on the xwiki.org forum and to contact XWiki SAS to discuss it with them. I'll start the process.

vmassol:
@michitux ah yes indeed. This text was added not for the home page (which predates the governance) but for this VOTE: http://markmail.org/message/mcbcvd4i5cgoebh5

The consequence of the VOTE was this rewrite: https://dev.xwiki.org/xwiki/bin/view/Community/Governance?viewer=changes&rev1=4.3&rev2=5.2&

Back at the time, I think we forgot about the home page video and personally I don't remember considering it an advertising for XWiki SAS back then (I think I had forgotten about its content and thought it was a generic video about XWiki features... ;)).

Still, you're right that the governance update almost allows it (it's not really non-intrusive IMO ;)).

Sereza7:
> First of all, this "transcript" is not a transcript of the audio but the content that is visible in the video.

Transcript is a general term, it means converting X -> text . Usually when transcribing a video, most of the focus is set on audio -> text (useful for deaf/impaired-audition people, it can be read and/or transformed into captions for subtitles). Here, there's no audio, but all the info is conveyed through a video so we do video(visual) -> text (useful for blind/impaired-vision people, it can be read out loud by screen readers).

> Also, I'm not sure how the summary relates to the transcript, as the summary doesn't summarize the transcript in my opinion.

This is the description of the Youtube video I just reused there. I feel like it's a bit shorter than the actual transcript and contain most of the main ideas. It's still not that short but I figured it'd be nice to have two alternatives to the video: the transcript which tries to copy the content in the video so that an assistive tool user wouldn't miss anything, and a summary which is actually a reinterpretation (!= content) of the video that's made to better fit the text format. 
I won't die on this hill though, one could be enough :+1: 

> And of course the WCAG issue at https://jira.xwiki.org/browse/XWIKI-18001

There's actually more WCAG violations related to the Youtube embed (some of which we don't really have a hand on I think), but by adding a transcript they'd become less important.

> I don't remember considering it an advertising for XWiki SAS back then

XWiki SAS is not mentionned once in the video, but it's implicitly said (-> I had to include it in the transcript for it to make sense), and the link at the end is xwiki.com ...

vmassol:
Done in [XWIKI-21028](https://jira.xwiki.org/browse/XWIKI-21028)

PR #2204, opened 2023-06-08
XWIKI-10752: "More applications" content is not readable in condensed display

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-10752

 
## PR Changes
* Reduced scope of the media query selectors in charge for this presentation bug
* Added extra style to display the inner elements nicely whatever the size of the column

## Video demo
https://up1.xwikisas.com/#mpc2g5T1yZzarhjFXvLGDw

## Note
BFD

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
@@ -659,7 +659,7 @@ require(['scriptaculous/effects'], function() {
 .panel-width-Large .applicationsPanel &gt; li &gt; a:focus,
 .panel-width-Large .applicationsPanelMoreList &gt; li &gt; a:hover,
 .panel-width-Large .applicationsPanelMoreList &gt; li &gt; a:focus,
-.panel-width-Large .applicationPanelMoreContainer ul &gt; li &gt; a:hover
+.panel-width-Large .applicationPanelMoreContainer ul &gt; li &gt; a:hover,
 .panel-width-Large .applicationPanelMoreContainer ul &gt; li &gt; a:focus {
   background-color: transparent;
   text-decoration: underline;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -649,15 +659,15 @@ require(['scriptaculous/effects'], function() {
 .panel-width-Large .applicationsPanel &gt; li &gt; a:focus,
 .panel-width-Large .applicationsPanelMoreList &gt; li &gt; a:hover,
 .panel-width-Large .applicationsPanelMoreList &gt; li &gt; a:focus,
-.panel-width-Large .applicationPanelMoreContainer ul &gt; li &gt; a:hover 
+.panel-width-Large .applicationPanelMoreContainer ul &gt; li &gt; a:hover
```
michitux, line 662: Why is there no `,` at the end of the line?

michitux (approved):
Looks good to me, one small comment.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
Sereza7: The diff is weird :thinking: , it looks like they weren't any beforehand too, just a space.
Maybe it was a bug but it got never reported because I don't think many people use large panels, and it's not an UI breaking bug, just a couple of colors (maybe) looking weird on anchors. I don't think we have a manual test covering this.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2204/commits/aaf04ab737288a251968f37ab39fe816f8dc8000

PR #2205, opened 2023-06-08
XWIKI-20699: Annotations are pointing to the old document after move

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20699

**TODO**
- [x] unit tests
- [x] integration tests
- [x] Update the `@since`


xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/R150500000XWIKI20699DataMigration.java
```diff
@@ -0,0 +1,91 @@
...
+ * targeting the document holding the annotation XObject.
+ *
+ * @version $Id$
+ * @since 15.5
+ * @since 14.10.14
+ */
+@Component
+@Named("R150500000XWIKI20699")
+@Singleton
+public class R150500000XWIKI20699DataMigration extends AbstractDocumentsMigration
```
manuelleduc, line 49: `150500000` replaced by `141014000` when cherry picking to 14.10.14

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
```diff
@@ -65,9 +62,7 @@ protected void hibernateMigrate() throws DataMigrationException
     {
         List<XWikiDocument> documents = selectDocuments()
             .stream()
-            .map(this::convert)
-            .filter(Optional::isPresent)
-            .map(Optional::get)
+            .map(documentReference -> new XWikiDocument(this.documentReferenceResolver.resolve(documentReference)))
```
manuelleduc, line 65: Not sure if this change might deserve its own issue.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
```diff
@@ -130,8 +131,7 @@ public void addAnnotation(String target, Annotation annotation) throws IOService
             EntityReference annotationClassReference = this.configuration.getAnnotationClassReference();
             annotationClassReference =
                 annotationClassReference.removeParent(annotationClassReference.extractReference(EntityType.WIKI));
-            int id = document.createXObject(annotationClassReference, deprecatedContext);
-            BaseObject object = document.getXObject(this.configuration.getAnnotationClassReference(), id);
+            BaseObject object = document.newXObject(annotationClassReference, deprecatedContext);
```
surli, line 134: `deprecatedContext` is a weird variable name?

tmortagne: That code is old enough to still have the illusions that the XWikiContext would be replaced in a short future :)

manuelleduc: yes, I think I can rename it to `context` safely.

tmortagne: I tend to prefer the name xcontext on my side for the XWikiContext (a little more explicit and more consistent with script variable names), but there is no official rule.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumer.java
```diff
@@ -0,0 +1,103 @@
...
+            XWikiContext context = this.contextProvider.get();
+            task(context.getWiki().getDocument(documentReference, context));
+        } catch (XWikiException e) {
+            throw new IndexException(String.format("Failed to resolve document [%s]", documentReference), e);
+        }
+    }
+
+    private void task(XWikiDocument document) throws IndexException
+    {
+        BaseObject xObject = document.getXObject(XWikiDocument.COMMENTSCLASS_REFERENCE);
```
surli, line 82: You could have multiple xobjects in the doc no? Or I'm missing something obvious here?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
```diff
@@ -200,12 +205,12 @@ public Collection<Annotation> getAnnotations(String target) throws IOServiceExce
                 return Collections.emptySet();
             }
             for (BaseObject object : objects) {
-                // if it's not on the required target, ignore it
-                if (object == null || !localTargetId.equals(object.getStringValue(Annotation.TARGET_FIELD))) {
-                    continue;
-                }
                 // use the object number as annotation id
-                result.add(loadAnnotationFromObject(object, deprecatedContext));
+                if (object != null && List.of(localTargetId, "")
```
tmortagne, line 209: Creating a new List just to check if `object.getStringValue(Annotation.TARGET_FIELD)` is equals to `localTargetId` or empty is not great (for performance and readability IMO).

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
```diff
@@ -385,10 +392,15 @@ protected Annotation loadAnnotationFromObject(BaseObject object, XWikiContext de
...
-                    annotation.set(propName, ((BaseProperty) object.get(propName)).getValue());
+                    Object value = ((BaseProperty) object.get(propName)).getValue();
+                    if (Objects.equals(propName, "target") && StringUtils.isEmpty(String.valueOf(value))) {
+                        value = target;
+                    }
+                    annotation.set(propName, value);
                 } catch (XWikiException e) {
-                    this.logger.warn("Unable to get property " + propName + " from object " + object.getClassName()
-                        + "[" + object.getNumber() + "]. Will not be saved in the annotation.", e);
+                    this.logger.warn(String.format(
```
tmortagne, line 401: You don't need String.format with a logger.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumer.java
```diff
@@ -0,0 +1,103 @@
...
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+
+/**
+ * Replace the target of annotation referencing themselves to avoid having outdated targets in case of document move or
+ * copy.
+ *
+ * @version $Id$
+ * @see <a href="https://jira.xwiki.org/browse/XWIKI-20699">XWIKI-20699</a>
+ * @since 15.5
```
tmortagne, line 50: Isn't it a big short for 15.5 ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
tmortagne: Well I think what is needed is a much bigger change than that actually as loading all the impacted XWikiDocument before handling them it's quite horrible for the memory. It should instead, for each entry of `selectDocuments()`, load the  XWikiDocument and call `logBeforeQueuingTask`.

manuelleduc: https://jira.xwiki.org/browse/XWIKI-21089 created for this, I'll revert here and open a dedicated PR.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
manuelleduc: What would be a suitable alternative for you?

tmortagne: Actually check "if `object.getStringValue(Annotation.TARGET_FIELD)` is equals to `localTargetId` or empty ".

manuelleduc: ok, I find it more verbose and less readable but ymmv. As I'm fine with both options anyway, I'll change that.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
```diff
@@ -200,12 +207,13 @@ public Collection<Annotation> getAnnotations(String target) throws IOServiceExce
...
             for (BaseObject object : objects) {
-                // if it's not on the required target, ignore it
-                if (object == null || !localTargetId.equals(object.getStringValue(Annotation.TARGET_FIELD))) {
-                    continue;
-                }
                 // use the object number as annotation id
-                result.add(loadAnnotationFromObject(object, deprecatedContext));
+                if (object != null) {
+                    String targetField = object.getStringValue(Annotation.TARGET_FIELD);
+                    if (Objects.equals(localTargetId, targetField) || Objects.equals("", targetField)) {
```
surli, line 213: I'm failing to remember exactly, but if I'm correct the check with empty string is for backward compatibility no? It might worth a comment here.

michitux, line 213: Shouldn't annotations with empty target only be loaded if the target is actually the document (i.e., its content) and not an XObject property reference? With the current code, annotations that are for the content will also be loaded for every object property.

Wouldn't it be safer to use [`StringUtils.isBlank(targetField)`](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#isBlank%28java.lang.String%29) to also catch, e.g., `null` values or user-edited objects where the user unintentionally inserted a space?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
```diff
@@ -240,11 +248,12 @@ public Annotation getAnnotation(String target, String annotationID) throws IOSer
             BaseObject object =
                 document.getXObject(this.configuration.getAnnotationClassReference(),
                     Integer.valueOf(annotationID));
-            if (object == null || !localTargetId.equals(object.getStringValue(Annotation.TARGET_FIELD))) {
+            String targetField = object == null ? null : object.getStringValue(Annotation.TARGET_FIELD);
+            if (object == null || !(Objects.equals(localTargetId, targetField) || Objects.equals("", targetField))) {
```
michitux, line 252: Same here regarding `isBlank` and matching the empty string only for the document/its content and not all object properties.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumer.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Inject
+    private TypedStringEntityReferenceResolver referenceResolver;
+
+    @Override
+    public void consume(DocumentReference documentReference, String version) throws IndexException
+    {
+        try {
+            XWikiContext context = this.contextProvider.get();
+            task(context.getWiki().getDocument(documentReference, context));
```
michitux, line 75: Wouldn't it be good to check here if the document exists? I guess nothing should happen if it doesn't exist but still it could be more obvious.

michitux (changes requested):
I think there is a regression in the logic for determining if an annotation is applicable for an object property - it will now also load annotations for the content instead of just annotations for the object property.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
manuelleduc: > Shouldn't annotations with empty target only be loaded if the target is actually the document (i.e., its content) and not an XObject property reference? With the current code, annotations that are for the content will also be loaded for every object property.

I'm probably missing something here. What's the usecase you have in mind?

michitux: According to the code and [the forum discussion](https://forum.xwiki.org/t/annotations-and-the-target-property/12187/3), the target is either a) a document reference, in which case the annotation is for the document's content or b) an object property reference, in which case the annotations is for an object property. My understanding is that for the first case, you're now using the empty string as reference. In the second case, the original target reference is still stored and annotations still break after moving, but my understanding is that we have currently no UI for that, anyways, so at least with the standard UI that won't be visible. However, when whatever code would load annotations for an object property, your changed code would now not only return annotations for that object property but also for the document's content, which, even if unused currently, is a regression.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/R160000000XWIKI20699DataMigration.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public String getDescription()
+    {
+        return "Queue documents containing annotations with a non-empty target for cleanup";
+    }
+
+    @Override
+    public XWikiDBVersion getVersion()
+    {
+        return new XWikiDBVersion(160000000);
```
surli, line 63: We need to be careful, that one already exists in master: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R160000000XWIKI17243DataMigration.java#L122

manuelleduc: Updated to `180600000` (been a while).
Also, I plan to update this is to the relevant one for each backport. Meaning that it's possible for the migration to be run several times for the same instance for some incremental upgrade scenarios.
But, the second run shouldn't have any side effect except a slightly slower than needed database upgrade step.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumer.java
manuelleduc: done.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
manuelleduc: Logic updated and moved to `org.xwiki.annotation.io.internal.DefaultIOService#matchesTarget`

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
manuelleduc: Should be fixed by  `org.xwiki.annotation.io.internal.DefaultIOService#matchesTarget`

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumer.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private boolean task(XWikiDocument document, BaseObject xObject)
+    {
+        String target = xObject.getStringValue(Annotation.TARGET_FIELD);
+        boolean updated = false;
+        if (!StringUtils.isEmpty(target)) {
+            EntityReference resolve = this.referenceResolver.resolve(target, EntityType.DOCUMENT);
+            if (Objects.equals(resolve, document.getDocumentReference())) {
+                xObject.setStringValue(Annotation.TARGET_FIELD, "");
```
michitux, line 122: Shouldn't the document be cloned at some point before it is modified?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/R180600000XWIKI20699DataMigration.java
```diff
@@ -0,0 +1,97 @@
...
+        XWiki wiki = getXWikiContext().getWiki();
+
+        try {
+            return wiki.getStore().getQueryManager()
+                .createQuery("SELECT distinct doc.fullName "
+                    + "FROM XWikiDocument doc, BaseObject as obj, StringProperty as prop "
+                    + "where doc.fullName = obj.name and obj.className = 'XWiki.XWikiComments' "
+                    + "and obj.id = prop.id.id "
+                    + "and prop.id.name='target'"
+                    + "and prop.value <> ''", Query.HQL)
```
michitux: Does this work on Oracle? Shouldn't we also test if the value is not null?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
michitux: Isn't this supposed to be fixed by this PR?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
```diff
@@ -151,7 +153,12 @@ public void addAnnotation(String target, Annotation annotation) throws IOService
             if (targetReference.getType() == EntityType.OBJECT_PROPERTY
                 || targetReference.getType() == EntityType.DOCUMENT)
             {
-                object.set(Annotation.TARGET_FIELD, this.localSerializer.serialize(targetReference), deprecatedContext);
+                // We only store the target if it is not pointing to the document containing the object.
+                // This makes it easier to have a valid annotation reference in case of page move/copy.
+                if (!Objects.equals(targetReference, object.getDocumentReference())) {
```
michitux, line 177: Does this equality work when the annotation points to a translation to the document? It seems that annotations don't store the locale the annotation is on, though - I just tested locally, and the target value doesn't have any locale parameter.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/test/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumerTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Test of {@link AnnotationInternalTargetFixTaskConsumer}.
+ *
+ * @version $Id$
+ * @since 17.10.11
+ * @since 18.4.3
+ * @since 18.6.0RC1
+ */
+@ComponentTest
```
michitux: Feels like an `OldcoreTest` would require a less fragile test setup and would probably also reduce the test setup code. We have `UpdatedDocumentMentionsAnalyzerTest` that shows how to run exactly the mandatory document initializer that is needed here to have the XClass initialized.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/test/java/org/xwiki/annotation/io/internal/DefaultIOServiceTest.java
```diff
@@ -0,0 +1,309 @@
...
+
+/**
+ * Test of {@link DefaultIOService}.
+ *
+ * @version $Id$
+ * @since 17.10.11
+ * @since 18.4.3
+ * @since 18.6.0RC1
+ */
+@ComponentTest
```
michitux: Same comment as above, I think an `OldcoreTest` would require less setup and in particular a less fragile setup.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-test/xwiki-platform-annotation-test-docker/src/test/it/org/xwiki/annotation/test/ui/AnnotationsIT.java
```diff
@@ -80,6 +84,12 @@ void setUp(TestUtils setup, TestReference testReference)
         setup.deletePage(testReference);
 
         setup.createUser(USER_NAME, USER_PASS, "", "");
+
+        // The user is advanced to simplify the move operations.
+        Map<String, Object> userProperties = new HashMap<>();
+        userProperties.put("usertype", "Advanced");
+        setup.updateObject("XWiki", "UserAnnotation", "XWiki.XWikiUsers", 0, userProperties);
+
         setup.login(USER_NAME, USER_PASS);
```
michitux:
```suggestion
        setup.createUserAndLogin(USER_NAME, USER_PASS, "usertype", "Advanced");
```

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-test/xwiki-platform-annotation-test-docker/src/test/it/org/xwiki/annotation/test/ui/AnnotationsIT.java
```diff
@@ -175,9 +185,44 @@ void showAnnotationsByClickingOnAQuote(TestUtils setup, TestReference testRefere
...
+        annotatableViewPage.addAnnotation(ANNOTATED_TEXT_1, ANNOTATION_TEXT_1);
+        annotatableViewPage.addAnnotation(ANNOTATED_TEXT_2, ANNOTATION_TEXT_2);
+        annotatableViewPage.addAnnotation(ANNOTATED_TEXT_3, ANNOTATION_TEXT_3);
+        annotatableViewPage.addAnnotation(ANNOTATED_TEXT_4, ANNOTATION_TEXT_4);
+
+        // Move the page to check that the annotations are still displayed afterward.
+        RenamePage renamePage = annotatableViewPage.getWrappedViewPage().rename();
+        renamePage.getDocumentPicker().setName("NewName");
+        CopyOrRenameOrDeleteStatusPage renameStatusPage = renamePage.clickRenameButton();
+        assertEquals("Done.", renameStatusPage.getInfoMessage());
```
michitux, line 198: Shouldn't this first wait for the rename to be done? I mean, shouldn't this call `renameStatusPage.waitUntilFinished()`?

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-test/xwiki-platform-appwithinminutes-test-docker/src/test/it/org/xwiki/appwithinminutes/test/ui/UserAndGroupClassFieldSubWikiIT.java
michitux, line 50: What's the relationship between this test and the original issue? I don't see the relationship.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-streams/xwiki-platform-filter-stream-xar/src/main/java/org/xwiki/filter/xar/internal/input/XARInputFilterStream.java
michitux, line 107: What's the relationship between this change the PR? Feels very wrong to have this change in this PR.

xwiki-platform-core/xwiki-platform-node/src/main/node/dev/config/src/vitest-config.ts
michitux, line 32: Same comment, this change shouldn't be in this PR.

XWIKI-20699-critical-review.md
```diff
@@ -0,0 +1,94 @@
+# Critical review of the XWIKI-20699 finalization work
```
michitux: While interesting, I don't think this file should be part of the PR.

XWIKI-20699-critical-review.md
```diff
@@ -0,0 +1,94 @@
...
+### L1 — Weak assertions in the adapted unit test
+- `verify(document).setAuthor(anyString())` — I stub `getUser()` to a concrete value, so I could
+  assert the exact string. `anyString()` is looser than necessary.
+- `when(document.clone()).thenReturn(document)` returns the *same* mock, so the test does not
+  actually prove that the *clone* (rather than the cached instance) is the object saved — which is
+  the whole reason master introduced the clone. The test passes but does not guard the intent.
+
+### L2 — Stale `@version $Id: …$` on the renamed migration file
+`R180600000XWIKI20699DataMigration.java` keeps the old `$Id: 3bfa9eff…$` from before the rename.
+Cosmetic (these keywords are effectively dead post-SVN), but inconsistent with a fresh file.
```
michitux: Wrong, they are not dead. They are set by git.

XWIKI-20699-review.md
```diff
@@ -0,0 +1,59 @@
+# PR #2205 — XWIKI-20699 finalization
```
michitux: As noted already on the other file, this needs to be removed from the PR.

michitux (changes requested):
This PR seems to be in bad state regarding the merge, it contains completely unrelated changes and Markdown files from some coding agents. This needs a thorough cleanup. Further, as mentioned in the comments, I think the tests need to be improved. I have also a question regarding the locale, but from a quick check it seems we don't distinguish them right now which feels strange. We should still check if this is only in storage or also in memory. Further, as mentioned in a comment, I fear the migration breaks on Oracle DB/will find all documents on Oracle DB.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-streams/xwiki-platform-filter-stream-xar/src/main/java/org/xwiki/filter/xar/internal/input/XARInputFilterStream.java
michitux: Maybe GitHub is also wrongly displaying this just for me?

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/R180600000XWIKI20699DataMigration.java
tmortagne: Yes, note that an empty string basically does not exist in Oracle: it will be stored as a null value.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-streams/xwiki-platform-filter-stream-xar/src/main/java/org/xwiki/filter/xar/internal/input/XARInputFilterStream.java
tmortagne: It's not just you, at least.

I noticed this pull request is in conflict, maybe github got lost a bit and resolving the conflict would help.

manuelleduc: thanks for the review @michitux and @tmortagne 
I've fixed the most obvious unrelated changes for now (not sure what that comes fro, but probably a mistake on my end). I'll address the rest of the observations later.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumer.java
manuelleduc: done

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/R180600000XWIKI20699DataMigration.java
manuelleduc: Replaced with `and length(prop.value) > 0`

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/DefaultIOService.java
manuelleduc: Yes, this is https://jira.xwiki.org/browse/XWIKI-21960

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/test/java/org/xwiki/annotation/io/internal/migration/hibernate/AnnotationInternalTargetFixTaskConsumerTest.java
manuelleduc: test migrated to oldcore, thanks for the suggestion

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/test/java/org/xwiki/annotation/io/internal/DefaultIOServiceTest.java
manuelleduc: test migrated to oldcore, thanks for the suggestion

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-test/xwiki-platform-annotation-test-docker/src/test/it/org/xwiki/annotation/test/ui/AnnotationsIT.java
manuelleduc: I introduced `org.xwiki.test.ui.po.RenamePage#setNewTitle` instead, to avoid the need to be advanced just to set the page title.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-test/xwiki-platform-annotation-test-docker/src/test/it/org/xwiki/annotation/test/ui/AnnotationsIT.java
manuelleduc: good point, wait added.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-io/src/main/java/org/xwiki/annotation/io/internal/migration/hibernate/R180600000XWIKI20699DataMigration.java
tmortagne: > Replaced with and length(prop.value) > 0

Is this something you found elsewhere ? I'm just wondering if Oracle will be a big fan of `length(null)`.

manuelleduc: I found a bunch of existing examples, for instance https://github.com/xwiki/xwiki-platform/blob/560de1b05a78c0227b4396cd6ccc4d8af6da9772/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/test/java/org/xwiki/notifications/filters/internal/livedata/custom/NotificationCustomFiltersQueryHelperTest.java#L280

PR #2210, opened 2023-06-13
Update Gradle Enterprise Maven extensions

jprinet:


vmassol:
Thanks! Could you do the same for xwiki-commons and xwiki-rendering? :)

jprinet:
Those are already up-to-date 👍

PR #2217, opened 2023-06-14
XWIKI-20990: User profile select for showing hidden elements lacks an accessible name

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20990

## PR Changes
* Bound the label and description to the selects using aria-describedby and label->for
* Updated labels in the notification preference tab

## View
No view changes

## Notes
I decided to include changes to the notification preference section in this PR since: 
* those changes fix similar issue to the ones reported in the aforementionned issue.
* this PR is straightforward and relatively small even with those changes
* the notification preference section accessibility issues were not reported by automated testing

## Tests
Passed successfully:
* mvn clean install -f xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/ -Pdocker,integration-tests,quality -Dxwiki.test.ui.wcag=true
* mvn clean install -f xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-test/ -Pdocker,integration-tests,quality -Dxwiki.test.ui.wcag=true

Tested the changes on a local distribution, the reported WCAG violations were fixed and the changes did not bring any more WCAG violation.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
@@ -212,8 +212,11 @@
   #set ($defaultValueHint = "($escapetool.xml($services.localization.render('userprofile.default')))")
   #if ($isEdit)
     #set ($escapedId = $escapetool.xml("XWiki.XWikiUsers_0_${fieldName}"))
-    #set ($escapedHintId = $escapedSelectId + "_hint")
-    &lt;select name="$escapedId" id="$escapedId" size="1" aria-describedby="$escapedHintId"&gt;
+    &lt;select name="$escapedId" id="$escapedId" size="1"
+      #if ($services.localization.get("XWiki.XWikiUsers_${fieldName}.hint"))
+        #set ($escapedHintId = $escapedSelectId + "_hint")
+        aria-describedby="$escapedHintId"
+      #end&gt;
       #if ($defaultValue)
         &lt;option value=""&gt;---&lt;/option&gt;
       #end


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsApplicationsPreferencesMacro.xml
```diff
@@ -1044,7 +1044,12 @@ require(['jquery', 'xwiki-meta', 'ApplicationWidget', 'xwiki-bootstrap-switch',
       #end
       ## Collapse button
       &lt;td class="notificationCollapseButtonCell"&gt;
-        &lt;button type="button" class="btn btn-default btn-sm collapseButton"&gt;$services.icon.renderHTML('caret-down')&lt;/button&gt;
+        &lt;button type="button" class="btn btn-default btn-sm collapseButton"&gt;
+          $services.icon.renderHTML('caret-down')
+          &lt;label class='sr-only'&gt;
+            $escapetool.xml($services.localization.render('notifications.settings.applications.table.dropdown.toggle.label',[$type.applicationName]))
+          &lt;/label&gt;
```
michitux, line 1051: The use of `<label>` is not allowed here, button elements don't allow interactive content and `<label>` is interactive content.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -212,7 +212,8 @@
   #set ($defaultValueHint = "($escapetool.xml($services.localization.render('userprofile.default')))")
   #if ($isEdit)
     #set ($escapedId = $escapetool.xml("XWiki.XWikiUsers_0_${fieldName}"))
-    &lt;select name="$escapedId" id="$escapedId" size="1"&gt;
+    #set ($escapedDescriptionId = $escapedSelectId + "_description")
+    &lt;select name="$escapedId" id="$escapedId" size="1" aria-describedby="$escapedDescriptionId"&gt;
```
michitux, line 216: Is there any problem if the hint doesn't exist? I've searched a bit and found https://stackoverflow.com/questions/67797543/aria-labelledby-is-not-valid-508-compliance which suggests it might be a problem. Also, shouldn't this then also be changed more systematically for other displayers and AWM? Also, it might make sense to update https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/VerticalForms/ if `aria-describedby` should always be used when there is a hint.

michitux (changes requested):

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsApplicationsPreferencesMacro.xml
Sereza7: Source: https://html.spec.whatwg.org/multipage/dom.html#interactive-content

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: AFAIK, the displaySelectField velocity macro is only used in the displayField where the description is defined.

Aria-describedby is completely optional here as of now, I used it here since it made sense and it wasn't much trouble, but I don't think it's that important and that we should systematically update it.
___
There's https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html going for it.
But automatic testing tools don't warn about missing descriptions.
If we have a good enough label, we can go without a description for most elements (sometimes the behavior is unexpected or complex, that's when aria-describedby shines).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsApplicationsPreferencesMacro.xml
```diff
@@ -1044,7 +1045,12 @@ require(['jquery', 'xwiki-meta', 'ApplicationWidget', 'xwiki-bootstrap-switch',
       #end
       ## Collapse button
       &lt;td class="notificationCollapseButtonCell"&gt;
-        &lt;button type="button" class="btn btn-default btn-sm collapseButton"&gt;$services.icon.renderHTML('caret-down')&lt;/button&gt;
+        &lt;button id="notifications-applications-preferences-macro-$appCount-button" type="button" class="btn btn-default btn-sm collapseButton"&gt;
+          $services.icon.renderHTML('caret-down')
+        &lt;/button&gt;
+        &lt;label for="notifications-applications-preferences-macro-$appCount-button" class='sr-only'&gt;
+        $escapetool.xml($services.localization.render('notifications.settings.applications.table.dropdown.toggle.label',[$type.applicationName]))
+        &lt;/label&gt;
```
michitux, line 1053: Why don't you just put the content inside the button in a `<span>`? I don't think it's a best practice to associate a label with a button.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
michitux: No, from my understanding of the code there is an `#if ($services.localization.get($hintKey)` before displaying the description so the description might or might not exist.

I'm not saying that we should make a description mandatory, just that it might make sense to update our HTML examples and the standard code for generating such forms in AWM to include the `aria-describedby` attribute when there is a hint.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,10 +248,12 @@
...
 #macro (displayField $fieldName $label $defaultValue)
   #set ($hintKey = "XWiki.XWikiUsers_${fieldName}.hint")
+  #set ($escapedSelectId = $escapetool.xml("XWiki.XWikiUsers_0_${fieldName}"))
+  #set ($escapedDescriptionId = $escapedSelectId + "_description")
   &lt;dt&gt;
-    &lt;label&gt;$!escapetool.xml($services.localization.render($label))&lt;/label&gt;
+    &lt;label for="$escapedSelectId"&gt;$!escapetool.xml($services.localization.render($label))&lt;/label&gt;
     #if ($services.localization.get($hintKey))
-      &lt;span class="xHint"&gt;$!escapetool.xml($services.localization.render($hintKey))&lt;/span&gt;
+    &lt;span id="$escapedDescriptionId" class='xHint'&gt;$!escapetool.xml($services.localization.render($hintKey))&lt;/span&gt;
```
michitux, line 256: Why did you remove the indentation here? I'm not sure we have a policy for this, but to me it seems like the rest of the code is indented when it makes sense from a Velocity point of view. Here the removed indentation makes it easy to miss the `#if`.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsApplicationsPreferencesMacro.xml
Sereza7: I think keeping the translation as "[...].dropdown.toggle.label" is correct. This could need to change with this change though.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2217/commits/344f06dcf537054a0e718adbd66fdaf09482d199

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsApplicationsPreferencesMacro.xml
```diff
@@ -1044,7 +1045,12 @@ require(['jquery', 'xwiki-meta', 'ApplicationWidget', 'xwiki-bootstrap-switch',
       #end
       ## Collapse button
       &lt;td class="notificationCollapseButtonCell"&gt;
-        &lt;button type="button" class="btn btn-default btn-sm collapseButton"&gt;$services.icon.renderHTML('caret-down')&lt;/button&gt;
+        &lt;button id="notifications-applications-preferences-macro-$appCount-button" type="button" class="btn btn-default btn-sm collapseButton"&gt;
```
michitux, line 1048: Now that there is no more label, you can remove the id again and also the `$appCount` variable.

Sereza7: :face_with_spiral_eyes: Trying to follow with fixes too fast, I should take my time and actually do something correctly...

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsAutoWatchPreferencesMacro.xml
```diff
@@ -410,11 +410,11 @@
 {{html clean="false"}}
   &lt;div class="xform"&gt;
     &lt;dl&gt;
-      &lt;dt&gt;&lt;label&gt;$escapetool.xml($services.localization.render('notifications.settings.watch.autowatchmode'))&lt;/label&gt;
-      &lt;p class="xHint"&gt;$escapetool.xml($services.localization.render('notifications.settings.watch.autowatchmode.hint'))&lt;/p&gt;
+      &lt;dt&gt;&lt;label for='preference_notification_autowatch'&gt;$escapetool.xml($services.localization.render('notifications.settings.watch.autowatchmode'))&lt;/label&gt;
+      &lt;p id='preference_notification_autowatch_description' class="xHint"&gt;$escapetool.xml($services.localization.render('notifications.settings.watch.autowatchmode.hint'))&lt;/p&gt;
```
michitux, line 414: Just seeing this again, I'm wondering: do we have any existing uses of this "description" suffix? In most cases we call this "hint" in XWiki, so maybe an id ending with `_hint` would be more consistent.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/pull/2217/commits/3d7d018870c801dbb1cdc741fcf5582a990f16b7

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsApplicationsPreferencesMacro.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2217/commits/3f4fdf1ad81aecc0b7af41299cb41b6c3ca60ac6

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsAutoWatchPreferencesMacro.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2217/commits/a9e4001ccb02b456060af4bf4d70b414e1bb0ce8

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
michitux: According to [this Stack Overflow post](https://stackoverflow.com/questions/68386577/can-aria-describedby-or-aria-details-reference-an-element-inside-of-an-iframe), pointing to a non-existing element can cause a validation errors. As I've said before `displayField` only prints the hint when there is actually one so it is not guaranteed that the hint element exists.

michitux:
Looks good apart from the question about the references to non-existing hints.

Sereza7:
This PR also fixes https://jira.xwiki.org/browse/XWIKI-20989

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: I understand now, thanks for the explanation :)
Addressed in https://github.com/xwiki/xwiki-platform/pull/2217/commits/6026b98b02184157f85bcf9cbf4e62db9c5eae4d

Sereza7:
## PR Changes
* Removed the aria attribute aria-describedby when the description/xHint span is not generated.

This PR is ready for review

PR #2218, opened 2023-06-15
XWIKI-21012: Rights are messed up after unregistering a right/uninstalling the like application

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-21012

  * Ensure to not mess up the list of rights used for computing ordinals by setting null in it when unregistering
  * Reuse the empty places in the list of rights for preventing reaching the limit of 64 rights
  * Add new unit test for covering previous bug and for emphasizing the 64 rights limitation

### DIFF
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
@@ -28,7 +28,9 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.xwiki.model.EntityType;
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
```diff
@@ -344,7 +368,9 @@ private Set<Right> cloneImpliedRights(Set<Right> impliedRights)
      */
     public static List<Right> values()
     {
-        return UNMODIFIABLE_VALUES;
+        return VALUES.stream()
+            .filter(Objects::nonNull)
+            .collect(Collectors.toUnmodifiableList());
```
michitux, line 373: Are you sure that callers don't rely on positions in the returned list?

surli: Hmmm I cannot guarantee it for usages of this API outside of XS. And for XS itself it's used internally by `Right#size` which is used by `RightSet` and `RightMap`. So probably a mistake.

surli: So I see various options: 
  1. We juste return the unmodifiable list of values. The problem is that external callers might not check for null and get a NPE because of that.
  2. We return a list of values without null, and we fix  `#size` to rely on `VALUES`. The problem is that it create a bit of inconsistency as `#size` won't return the same size as this.
  3. We return a list of values without null, we introduce a new API that returns the null values, and we use that one when needing to check the size in XS for RightSet/RightMap. 

Personally I think 3 is the safest.

michitux: Maybe we should decouple the concept of "size" from the concept of "maximum value of a right" to make things clearer for callers.

surli: > Maybe we should decouple the concept of "size" from the concept of "maximum value of a right" to make things clearer for callers.

Well I thought about adding a new API just for the size but it feels more limiting than entirely returning the actual list of values with null elements.

PR #2221, opened 2023-06-15
XWIKI-21002: Page picker macro has no label

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-21002

## PR Changes
* Added a title to the different select elements if not defined beforehand.

## Tests
Passed:
* `mvn clean install -f xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/ -Pdocker,integration-tests,quality -Dxwiki.test.ui.wcag=true`

## Notes
I decided to go further than the reported issue, and fix every similar use of the suggest widget that would be the cause of a violation.

The localization keys and values have been chosen according to the information on [the Suggest Widget documentation page](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/AutoSuggestWidget/).

When testing with Axe devtool (browser tester based on axe-core), there is still a best practice reported:
> Ensure that every form element has a visible label and is not solely labeled using hidden labels, or the title or aria-describedby attributes.

https://dequeuniversity.com/rules/axe/4.6/label-title-only

I suppose this is okay to not fulfill it since this is only a best practice (not a WCAG violation), and trying to do so would bring about a long discussion: is such a change in the UI even positive (this could cause clutter)?


xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1376,6 +1376,11 @@ core.widgets.suggest.showResults=Go to search page\u2026
 core.widgets.suggest.valuePrefix=Value:
 core.widgets.suggest.transportError=Failed to retrieve suggestions:
 core.widgets.suggest.hide=hide suggestions
+core.widgets.suggest.attachmentPicker.title=Pick attachments.
+core.widgets.suggest.classPicker.title=Pick classes.
+core.widgets.suggest.pagePicker.title=Pick pages.
+core.widgets.suggest.userPicker.title=Pick users.
+core.widgets.suggest.xpropertyPicker.title=Pick property values.
```
michitux, line 1383: I'm wondering about the single vs. multiple select use case. As far as I understand, all of them support single or multiple selection. I think the label should reflect this.

michitux (commented):
I'm wondering if it isn't rather the job of the caller to provide a proper form label. Any examples where this is actually used and an issue? I mean the test where this fails is a rather constructed example as I think these pickers shouldn't be used without a surrounding form that provides labels.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
Sereza7: :+1: 
Fixed in https://github.com/xwiki/xwiki-platform/pull/2221/commits/dbfb5ddb85158d89a0730f12c2f1c57a15459357

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -2423,6 +2423,11 @@ $string##
 #macro (userPicker_input $multiSelect $parameters)
   #if ($multiSelect)
     #set ($discard = $parameters.put('multiple', 'multiple'))
+    #set ($discard = $parameters.putIfAbsent('title',
+      "$services.localization.render('core.widgets.suggest.userPicker.multi.title')"))
```
mflorea, line 2427:
```suggestion
      $services.localization.render('core.widgets.suggest.userPicker.multi.title')))
```

Same below.

mflorea (approved):
Looks good. Thanks.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1376,6 +1376,16 @@ core.widgets.suggest.showResults=Go to search page\u2026
 core.widgets.suggest.valuePrefix=Value:
 core.widgets.suggest.transportError=Failed to retrieve suggestions:
 core.widgets.suggest.hide=hide suggestions
+core.widgets.suggest.attachmentPicker.single.title=Pick an attachment.
+core.widgets.suggest.attachmentPicker.multi.title=Pick attachments.
+core.widgets.suggest.classPicker.single.title=Pick a class.
+core.widgets.suggest.classPicker.multi.title=Pick classes.
+core.widgets.suggest.pagePicker.single.title=Pick a page.
+core.widgets.suggest.pagePicker.multi.title=Pick pages.
+core.widgets.suggest.userPicker.single.title=Pick an user.
```
michitux, line 1385:
```suggestion
core.widgets.suggest.userPicker.single.title=Pick a user.
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/pull/2221/commits/200a0e6772b6bf275389cfccc2c1ba0158815fb5

michitux (approved):
I meant more that the user of the picker would add an actual `<label>` element to label the picker. Anyways, adding the title won't hurt and the label would override it, anyways.

PR #2075, opened 2023-02-04
XWIKI-18007: Drawer menu improvements for accessibility

wthrajat:
This PR partly fix the issue https://jira.xwiki.org/browse/XWIKI-18007
The drawer menu can now be closed by pressing the ```ESC``` key (this works for both the cases, 1) if you opened the drawer using keyboard 2) if you opened the drawer by clicking on it)

Earlier there was no such option to close the drawer menu through keyboard.

PR #2225, opened 2023-06-16
XWIKI-18007: Drawer menu improvements for accessibility

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-18007

## PR Changes
* Changed the drawer menu position, so it gets right after its toggler.
* Updated the drawer controls to close when focusing out and pressing escape
* Moved the script from javascript.vm to flamingo.js
* Added one style rule in drawer.less to hide elements from the tab order when the drawer is closed.

## Video demo
https://up1.xwikisas.com/#BrnQdLpraMVHfWyF0be0og
## Notes
This PR takes elements from https://github.com/xwiki/xwiki-platform/pull/2075 and adds upon them more changes to fix the ticket completely.

## Tests
`mvn clean install -f xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/ -Pdocker,integration-tests,quality` failed only on NavigationIT.navigateWithKeyboardShortcuts:142 and NavigationIT.docExtraLoadingFromTabClicks:164, which are build fails as of the 16th of June.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
@@ -51,3 +51,39 @@ require(['jquery', 'bootstrap'], function($) {
     translateButton.attr('title', translateButton.attr('data-title')).popover();
   });
 });
+
+require(['jquery', 'iscroll', 'drawer'], function($, IScroll) {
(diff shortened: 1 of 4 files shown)


michitux (changes requested):
While the code looks okay, I don't agree with the solution for moving focus out of the drawer. I think it is bad usability to close the drawer when the focus moves out of it. Imho, moving focus shouldn't close the drawer but rather focus should stay inside the drawer. Further, but less important, I think it would be good to add an explicit close button. It could, e.g., be positioned at the top left of the drawer outside the actual drawer area.

Sereza7:
In regards to https://jira.xwiki.org/browse/XWIKI-21037, it seems like keeping the dependency on the `drawer` webjar is not a good idea. Current web features should allow to make a clean drawer without much trouble.

Closing this PR because the solution does not seem appropriate to the situation.

PR #2226, opened 2023-06-19
XWIKI-21034: Write functional tests and Page Object for Quick Actions

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-21034

This pull request adds a a page object and a first functional test for Quick Actions.


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/QuickActionDropdown.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    }
+    
+    /**
+     * Submits the selected Quick Action.
+     */
+    public void submit()
+    {
+        if (this.getDropdown() != null) {
```
mflorea, line 122: This IF doesn't make sense because if the test calls ``submit`` then it means it expects the drop down to be displayed.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/QuickActionDropdown.java
```diff
@@ -0,0 +1,149 @@
...
+            return null;
+        }
+    }
+    
+    
+    /**
+     * Open the Quick Action drop-down.
+     * 
+     */
+    public void open()
```
mflorea, line 79: I'm wondering if it's not better to have a ``CKEditor.openQuickActions()`` method instead. This would allow you to do the wait in the ``QuickActionDropdown`` constructor and to also get a reference to the underlying ``WebElement`` instance there.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/QuickActionDropdown.java
mflorea, line 1: The plural ``Actions`` sounds better to me, as this is a drop down that lists quick actions.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/QuickActionDropdown.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+    
+    /**
+     * Finds the element enclosing given text.
+     * @param tag - search query
+     * @param content - the text to be found in the element
+     * @return the corresponding element
+     */
+    public boolean findElementsWithContent(String tag, String content)
```
mflorea, line 134: It seems this method looks in the edited content, so it doesn't seem related to the quick actions drop down page object. Also the method name suggests it return an element, but in fact it returns a boolean for which ``hasElementsWithContent`` is a better name. Then I'm not sure you need this method when you have ``RichTextAreaElement#getContent()``.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionIT.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+/**
+ * All functional tests for Quick Actions.
+ *
+ * @version $Id$
+ * @since 15.5RC1
+ */
+@UITest
+public class QuickActionIT
```
mflorea, line 41:
```suggestion
public class QuickActionsIT 
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionIT.java
```diff
@@ -0,0 +1,81 @@
...
+@UITest
+public class QuickActionIT 
+{
+    
+    /**
+     * Test string that will be inserted to check formatting.
+     */
+    private static final String TEST_TEXT = "Hello, world!";
+
+    @BeforeEach
```
mflorea, line 49: The browser is not closed after each test so the user remains logged in. Thus you need to login the user only once, at the start.

```suggestion
    @BeforeAll
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionIT.java
```diff
@@ -0,0 +1,81 @@
...
+        // Run the first quick action
+        QuickActionDropdown dropdown = new QuickActionDropdown(editor);
+        
+        dropdown.open();
+        dropdown.sendKeys("heading 1");
+        dropdown.submit();
+        
+        editor.getRichTextArea().sendKeys(TEST_TEXT);
+        assert dropdown.findElementsWithContent("h1", TEST_TEXT);
+        editPage.clickSaveAndView();
```
mflorea, line 79:
```suggestion
        WYSIWYGEditPage editPage = WYSIWYGEditPage.gotoPage(testReference);
        CKEditor editor = new CKEditor("content").waitToLoad();
        RichTextAreaElement textArea = editor.getRichTextArea();
        
        textArea.sendKeys("/hea");
        editor.getQuickActionsDropDown().waitForQuickActionSelected("Heading 1");
        textArea.sendKeys(Keys.ENTER, "TEST_TEXT");
        assertEquals("...", textArea.getContent());
```

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionIT.java
```diff
@@ -0,0 +1,81 @@
...
+import org.xwiki.test.docker.junit5.UITest;
+import org.xwiki.test.ui.TestUtils;
+import org.xwiki.test.ui.po.editor.WYSIWYGEditPage;
+
+
+/**
+ * All functional tests for Quick Actions.
+ *
+ * @version $Id$
+ * @since 15.5RC1
```
manuelleduc, line 38: Note: 15.5RC1 has been released, the `@since` will need to be updated to 15.5 or 15.6RC1 before merging.

douakli: Thank you!

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/QuickActionDropdown.java
douakli: I moved it as you suggested, but I'm not sure it is necessary since as you mentioned, the user does not wait for the dropdown to show

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/QuickActionsDropdown.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    
+    /**
+     * Waits for the given Quick Action to be selected.
+     * @param label - The name of the action
+     */
+    public void waitForQuickActionSelected(String label)
+    {
+        WebElement item = getDriver().findElement(By.className("cke_autocomplete_selected"));
+        item.findElement(By.xpath("//*[text() = '" + label + "']"));
```
mflorea, line 52: Could be written in a single line.

```suggestion
        getDriver().findElement(By.xpath("//*[contains(@class, 'cke_autocomplete_selected')]//*[. = '" + label + "']"));
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
```diff
@@ -0,0 +1,173 @@
...
+        textArea.sendKeys("hea");
+        qa.waitForQuickActionSelected("Heading 1");
+        textArea.sendKeys(Keys.ENTER);
+        qa.waitForQuickActionSubmitted();
+        
+        // Write some text
+        textArea.sendKeys(TEST_TEXT);
+        
+        assert textArea.getContent().contains("<h1>" + TEST_TEXT + "<br></h1>");
+        editPage.clickSaveAndView();
```
mflorea, line 91: Does it bring any value to save the page? Are you testing something with this? (maybe it helps with debugging?)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+    
+    @Test
+    @Order(2)
+    void heading2(TestUtils setup, TestReference testReference) throws Exception
```
mflorea, line 96: I think it was enough to write a single heading test (the one that is more complex). But let's keep them.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
```diff
@@ -0,0 +1,173 @@
...
+
+        // Switch to another style
+        QuickActionsDropdown qa = editor.openQuickActionsDropdown();
+        textArea.sendKeys("hea");
+        qa.waitForQuickActionSelected("Heading 1");
+        textArea.sendKeys(Keys.ENTER);
+        qa.waitForQuickActionSubmitted();
+        
+        // Write some text
+        textArea.sendKeys(TEST_TEXT);
```
mflorea, line 161: You could assert here the HTML and drop the first heading test. It's fine to have a "story", i.e. to test multiple things in functional tests (compared to unit tests, which are less expensive).

mflorea (approved):
Looks very good. Let me know if you have something more to add (we should merge tomorrow).

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
douakli: I'm saving only to avoid the confirmation dialog before leaving the page when starting the next test. There probably is a better way to handle that indeed.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
douakli: Oh okay, you did say to test every manually defined action so I went for it :)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
mflorea: Ah, I see, this deserves a comment. Now, instead of saving you can also cancel :) which is less expensive so normally faster. You can also handle the confirmation modal with Selenium but it's probably easier to just cancel the edit.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/pom.xml
```diff
@@ -58,6 +58,20 @@
...
+      <artifactId>xwiki-platform-mentions-ui</artifactId>
+      <version>${project.version}</version>
+      <type>xar</type>
+    </dependency>
+    <dependency>
+      <!-- We need this to test the code snippet Quick Action. -->
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-rendering-macros</artifactId>
+      <version>${project.version}</version>
+      <type>pom</type>
```
mflorea, line 73: Depending on the POM looks strange. Don't you need to depend on ``xwiki-platform-rendering-macro-code`` instead?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/CKEditor.java
```diff
@@ -155,13 +155,24 @@ private void internalClickImageButton()
...
+    
+    /**
+     * Clicks close on a CKEditor dialog.
+     */
+    public void cancelDialog() 
+    {
+        WebElement cancelButton = getDriver().findElement(By.className("cke_dialog_ui_button_cancel"));
+        getDriver().waitUntilElementIsEnabled(cancelButton);
+        cancelButton.click();
+        getDriver().waitUntilElementDisappears(By.className("cke_dialog_contents"));
```
mflorea, line 176: These two methods fit better in a ``CKEditorDialog`` page object.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/MacroDialogEditModal.java
```diff
@@ -37,7 +37,8 @@ public class MacroDialogEditModal extends BaseElement
      */
     public MacroDialogEditModal waitUntilReady()
     {
-        getDriver().waitUntilElementIsVisible(By.cssSelector(".gadget-editor-modal .macro-name"));
+        getDriver().waitUntilElementIsVisible(
+                By.xpath("//*[contains(@class, 'editor-modal')]//*[contains(@class, 'macro-name')]"));
```
mflorea, line 41: Would be good to add a comment explaining why we use the suffix ``editor-modal`` instead of the full class name. And you can use a CSS selector, ``[attr*=value]``, see https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors .

    [class*='-editor-modal'] .macro-name

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/MacroDialogEditModal.java
```diff
@@ -46,6 +47,7 @@ public MacroDialogEditModal waitUntilReady()
      */
     public void clickSubmit()
     {
-        getDriver().findElement(By.cssSelector(".gadget-editor-modal .modal-footer .btn-primary")).click();
+        getDriver().findElement(
+                By.xpath("//*[contains(@class, 'editor-modal')]//*[contains(@class, 'modal-footer')]//*[contains(@class, 'btn-primary')]")).click();
```
mflorea, line 51: Same, the CSS selector might be simpler. XPath is useful when you need to match the content of an HTML element, or its parent or a sibling.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/RichTextAreaElement.java
```diff
@@ -155,4 +156,18 @@ private WebElement getActiveElement()
         getDriver().switchTo().frame(this.iframe);
         return getDriver().switchTo().activeElement();
     }
+    
+    /**
+     * Waits until a macro appears in the document.
+     */
+    public void waitUntilMacroAppears()
```
mflorea, line 163: I find this too specific. I'd rather have a more generic "wait until the content is re-loaded".

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/RichTextAreaElement.java
```diff
@@ -155,4 +156,18 @@ private WebElement getActiveElement()
...
+    {
+        try {
+            getDriver().waitUntilElementDisappears(this.iframe, By.tagName("iframe"));;
```
mflorea, line 166: I don't understand this line. You pass the iframe as the parent and you're looking for a child element with name "iframe"? Doesn't make sense, because the iframe element doesn't have child nodes. There's also a double ``;`` at the end.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
```diff
@@ -39,7 +43,9 @@
  * @version $Id$
  * @since 15.5
  */
-@UITest
+@UITest(extraJARs = {
+    "org.xwiki.platform:xwiki-platform-search-solr-query",
```
mflorea, line 47: This needs a comment.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
```diff
@@ -70,25 +88,35 @@ private static void createAndLoginStandardUser(TestUtils setup)
...
+        
+        // Switch back to paragraph
+        textArea.sendKeys("/");
+        qa = new AutocompleteDropdown();
+        textArea.sendKeys("parag");
+        qa.waitForItemSelected("Paragraph");
+        textArea.sendKeys(Keys.ENTER);
+        qa.waitForItemSubmitted();
+        
+        assert textArea.getContent().contains("<p>" + TEST_TEXT + "<br></p>");
```
mflorea, line 119: I think we prefer the JUnit assertions (``Assertions.assertTrue``) because the behaviour of ``assert`` can be disabled, thus making the tests pass.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/pom.xml
tmortagne: Yes. The point of depending on a pom is to get the dependencies of a module without the module itself, but it does not seem to be something needed here.

That being said, it seems strange to need the code macro for a test. If you added it because it's used by a XAR you are testing then it should be added as dependency of that XAR instead.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/QuickActionsIT.java
tmortagne: In general `assert` is generally more something mean to be used at runtime, as a checkpoint (but that's not really something we do in XWiki).

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/RichTextAreaElement.java
douakli: I'm not sure how to tell that the content has been reloaded. The content refreshes but I believe the ckeditor widget take some time to appear?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/RichTextAreaElement.java
douakli: Thank you, I'm not even sure how the test can pass despite that mistake

douakli: Correction, the test fails when I remove the parent, I'm confused

michitux: It passes because it waits for the iframe to not have a child with tag `iframe`, a condition that is instantly true.

mflorea (commented):
Looks good, with some small comments.

mflorea (approved):

PR #2228, opened 2023-06-20
XWIKI-18508: Access (view) denied randomly on pages and solr search

michitux:
* Don't fail the security cache loading when an entry cannot be inserted into the security cache.
* Bulletproof the security cache against adding non-group/user entries as parents when a group/user is required.
* Mark the main wiki entry as user to respect this model that user parents are actually user entries.

Jira issue: https://jira.xwiki.org/browse/XWIKI-18508

This PR also includes the commits from #2223 as I wouldn't dare applying this without these fixes and the changes have quite some conflicts. If you want to review this in isolation, I would suggest to just look at the most recent commit. I'll rebase this once #2223 has been merged.


manuelleduc (commented):
Thanks for working on this. I did not find any code style related issues. For the rest, I don't know this part of the code well enough to have something interesting to say.

tmortagne (approved):

michitux:
I've merged this manually, GitHub somehow didn't recognize my force push after the rebase.

PR #2232, opened 2023-06-21
XWIKI-20945: Display macro fails to execute when no reference is specified

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20945

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -174,6 +174,19 @@
                 </item>
               </differences>
             </revapi.differences>
+            <revapi.differences>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.annotation.added</code>
+                  <old>method void org.xwiki.rendering.macro.display.DisplayMacroParameters::setReference(java.lang.String)</old>
+                  <new>method void org.xwiki.rendering.macro.display.DisplayMacroParameters::setReference(java.lang.String)</new>
+                  <annotation>@org.xwiki.properties.annotation.PropertyMandatory</annotation>
+                  <justification>Reference property of the display macro made explicitly mandatory.</justification>
+                  <criticality>allowed</criticality>
+                </item>
+              </differences>
+            </revapi.differences>
           </analysisConfiguration>
         </configuration>
       </plugin>
--- a/xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-macro/src/main/java/org/xwiki/rendering/macro/display/DisplayMacroParameters.java
+++ b/xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-macro/src/main/java/org/xwiki/rendering/macro/display/DisplayMacroParameters.java
@@ -28,6 +28,7 @@ import org.xwiki.properties.annotation.PropertyDisplayHidden;
 import org.xwiki.properties.annotation.PropertyDisplayType;
 import org.xwiki.properties.annotation.PropertyFeature;
 import org.xwiki.properties.annotation.PropertyGroup;
+import org.xwiki.properties.annotation.PropertyMandatory;
 import org.xwiki.properties.annotation.PropertyName;
 
 /**
@@ -66,6 +67,7 @@ public class DisplayMacroParameters
     @PropertyGroup("stringReference")
     @PropertyFeature("reference")
     @PropertyDisplayType(EntityReferenceString.class)
+    @PropertyMandatory
     public void setReference(String reference)
     {
         this.reference = reference;


xwiki-platform-core/pom.xml
```diff
@@ -174,6 +174,19 @@
...
               </differences>
             </revapi.differences>
+            <revapi.differences>
+              <differences>
+                <item>
+                  <ignore>true</ignore>
+                  <code>java.annotation.added</code>
+                  <old>method void org.xwiki.rendering.macro.display.DisplayMacroParameters::setReference(java.lang.String)</old>
+                  <new>method void org.xwiki.rendering.macro.display.DisplayMacroParameters::setReference(java.lang.String)</new>
+                  <annotation>@org.xwiki.properties.annotation.PropertyMandatory</annotation>
```
vmassol, line 184: We may want to change the following instead: https://github.com/xwiki/xwiki-commons/blob/4da483054e412d315dce36d34c24fd5ef454ae59/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/revapi.json#L59-L72

Do you see any case where adding this annotation would be a breakage?

vmassol:
Also shouldn't the macro code be modified to not check anymore the reference value? If it's set as mandatory I guess that the underlying framework will perform the check, right?

xwiki-platform-core/pom.xml
manuelleduc: That's a good point. I think it has already been discussed but I don't recall the conclusion. I don't think it would hurt.

manuelleduc:
> Also shouldn't the macro code be modified to not check anymore the reference value? If it's set as mandatory I guess that the underlying framework will perform the check, right?

Yes, I'll check if it's still possible to reach this part of the code with a `null` reference.

vmassol:
> > Also shouldn't the macro code be modified to not check anymore the reference value? If it's set as mandatory I guess that the underlying framework will perform the check, right?
> 
> Yes, I'll check if it's still possible to reach this part of the code with a `null` reference.

yeah, I don't recall if the mandatory annotation is used only in the WYSIWYG editor or if it's used in the generic macro manager code. It's very possible it's not and that we still need the check and thus have the error reported in the screenshot of  https://jira.xwiki.org/browse/XWIKI-20945 (hence my question there).

manuelleduc:
> yeah, I don't recall if the mandatory annotation is used only in the WYSIWYG editor as validation or if it's used in the generic macro manager code

The error is generated from here https://github.com/xwiki/xwiki-rendering/blob/ad08d973d7c1242ac72f16a26d021473493e70f7/xwiki-rendering-transformations/xwiki-rendering-transformation-macro/src/main/java/org/xwiki/rendering/internal/transformation/macro/MacroTransformation.java#L303
And the initial exception thrown here https://github.com/xwiki/xwiki-commons/blob/7ac887f71da70e3ad087ed7c5a2fd9397a92f66e/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanManager.java#L183

vmassol:
ok so if I understand your reply, it means we can remove the check for null, right?

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-macro/src/main/java/org/xwiki/rendering/macro/display/DisplayMacroParameters.java
```diff
@@ -66,6 +67,7 @@ public class DisplayMacroParameters
     @PropertyGroup("stringReference")
     @PropertyFeature("reference")
     @PropertyDisplayType(EntityReferenceString.class)
+    @PropertyMandatory
```
tmortagne, line 70: You cannot do that, since it's not that simple. There is a reason why this was not done like this: there are several way to provide the reference like setting the page (or the document when including legacy API).

tmortagne (changes requested):

xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-macro/src/main/java/org/xwiki/rendering/macro/display/DisplayMacroParameters.java
vmassol: ah right, I forgot this... good catch Thomas

vmassol: We should add a comment in the file to explain it.

vmassol: We would need a mandatory group annotation to say we need one of the listed option to be set.

tmortagne: Yes, the right way to fix this is to add the concept of mandatory to @PropertyFeature.

PR #2238, opened 2023-06-28
XWIKI-21083: Tag suggest feature is broken

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-21083

## PR Changes
* Now using  `javascript:void(0)` instead of `null` for an empty URL
* (extra Accessibility) Updated the DOM to properly generate the suggest element
* (extra Accessibility) Updated style to avoid visual changes related to the DOM changes

## Video demo
Used keyboard only to choose tags. It'd be similar with a mouse.
https://up1.xwikisas.com/#0ig--RpKFzgBJ3ES2Yd6Ew

## Notes
It's not a good thing to have an anchor with an empty URL, so there is no perfectly clean solution to make those and I chose to go with the `javascript:void(0)` which has some impact on performance, very minimal but still here.
The better technical solution would have been to replace the anchor with a button in the case there's no URL, however it would have brought a lot of changes in the way the suggest currently works and would have probably needed an overhaul of this UI component.
Since this issue is a huge blocker, I decided to go with a correct solution here instead of the technically perfect one.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
@@ -109,9 +109,11 @@
 ## Note that the form_token parameter needs to be kept before the xredirect parameter since the JS code might replace the latter
 ## All that would need a cleaner fix in the javascript of tags.
 <form action="$doc.getURL('view', "xpage=documentTags&amp;xaction=add&amp;form_token=$!{escapetool.url($services.csrf.token)}&amp;xredirect=${xredirect}")" method="post" class="tag-add-form">
-  <div>
+  <div id="tag-add-form-suggest">
     <label for="tag">$services.localization.render('core.tags.add.label')<br/>
-      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label><br/>
+      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label>
+  </div>
+  <div>
     <span class="buttonwrapper"><input class="button button-add-tag" type="submit" value="$services.localization.render('core.tags.add.submit')"/></span>
     <span class="buttonwrapper"><a class="button button-add-tag-cancel" href="$doc.getURL('view', "#${tagsId}")">$services.localization.render('core.tags.add.cancel')</a></span>
   </div>
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
@@ -111,7 +111,8 @@
 <form action="$doc.getURL('view', "xpage=documentTags&amp;xaction=add&amp;form_token=$!{escapetool.url($services.csrf.token)}&amp;xredirect=${xredirect}")" method="post" class="tag-add-form">
   <div id="tag-add-form-suggest">
     <label for="tag">$services.localization.render('core.tags.add.label')<br/>
-      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label>
+      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/>
+    </label>
   </div>
   <div>
     <span class="buttonwrapper"><input class="button button-add-tag" type="submit" value="$services.localization.render('core.tags.add.submit')"/></span>
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/suggest.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/suggest.js
@@ -730,7 +730,9 @@ var XWiki = (function(XWiki){
         noHighlight: true, // we do the highlighting ourselves
         containerTagName: 'a'
       });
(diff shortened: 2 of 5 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/suggest.js
```diff
@@ -730,7 +730,10 @@ var XWiki = (function(XWiki){
         noHighlight: true, // we do the highlighting ourselves
         containerTagName: 'a'
       });
-      item.containerElement.setAttribute('href', arr[i].url);
+      if (arr[i].url == null)  {
+        item.containerElement.setAttribute('href', 'javascript:void(0)');
+      }
+      else item.containerElement.setAttribute('href', arr[i].url);
```
mflorea, line 736: Strange formatting. Also, you should always wrap if/else blocks in ``{...}``. But in this case you can simplify it to:

```suggestion
      item.containerElement.setAttribute('href', arr[i].url || 'javascript:void(0)');
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/viewers/tags.js
```diff
@@ -169,6 +169,7 @@ viewers.Tags = Class.create({
 
     new XWiki.widgets.Suggest(form.down('input[type=text]'), {
       script: "${xwiki.getURL('Main.WebHome', 'view', 'xpage=suggest&classname=XWiki.TagClass&fieldname=tags&firCol=-&secCol=-')}&",
+      parentContainer:"tag-add-form-suggest",
```
mflorea, line 172: Deserves a comment explaining why the default value is not good enough.

mflorea (changes requested):

mflorea:
> I chose to go with the javascript:void(0) which has some impact on performance, very minimal but still here.

Can you detail this part. What's the performance impact?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/viewers/tags.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2238/commits/800ad72f3ccb496ceaa30ddab845f4c8e6ae1af7

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
```diff
@@ -109,9 +109,12 @@
 ## Note that the form_token parameter needs to be kept before the xredirect parameter since the JS code might replace the latter
 ## All that would need a cleaner fix in the javascript of tags.
 <form action="$doc.getURL('view', "xpage=documentTags&amp;xaction=add&amp;form_token=$!{escapetool.url($services.csrf.token)}&amp;xredirect=${xredirect}")" method="post" class="tag-add-form">
-  <div>
+  <div id="tag-add-form-suggest">
     <label for="tag">$services.localization.render('core.tags.add.label')<br/>
-      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/></label><br/>
+      <input class="input-tag" type="text" id="tag" name="tag" autocomplete="off"/>
+    </label>
```
manuelleduc, line 115: Is the removal of the trailing `<br/>` intentional?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/suggest.js
```diff
@@ -730,7 +730,7 @@ var XWiki = (function(XWiki){
         noHighlight: true, // we do the highlighting ourselves
         containerTagName: 'a'
       });
-      item.containerElement.setAttribute('href', arr[i].url);
+      item.containerElement.setAttribute('href', arr[i].url || 'javascript:void(0)');
```
manuelleduc, line 733: Maybe add a comment explaining why the `'javascript:void(0)'` alternative is required for empty urls.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/documentTags.vm
Sereza7: Yes :)
It wasn't needed with the other changes to style. Semantically it did not hold much importance.

A source about the semantics of line breaks, and its conclusion:
https://www.joedolson.com/2007/01/is-a-br-tag-semantic/

> I don’t think that a line break has semantic value. However, there are situations where a line break is required because custom dictates it. Other methods of formatting may have their value, but the use of an occasional line break isn’t going to make or break a site’s quality code.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/suggest.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2238/commits/bca95f38b237e50d19004759e8abd6e06b4f6996

manuelleduc (approved):

Sereza7:
Note: this blocker should be to be cherry-picked for LTS

PR #2240, opened 2023-06-29
XWIKI-21089: AbstractDocumentsMigration memory footprint is too large

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21089


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
```diff
@@ -63,28 +58,13 @@ public abstract class AbstractDocumentsMigration extends AbstractHibernateDataMi
...
-    {
-        XWikiContext context = getXWikiContext();
-        try {
-            return Optional.of(
-                context.getWiki().getDocument(this.documentReferenceResolver.resolve(documentReference), context));
-        } catch (XWikiException e) {
-            this.logger.error("Failed to resolve [{}]", documentReference, e);
-            return Optional.empty();
+        for (String documentReference : selectedDocuments) {
+            XWikiDocument document = new XWikiDocument(this.documentReferenceResolver.resolve(documentReference));
```
tmortagne, line 65: Hmm I did not notice that in your previous version but `this.documentReferenceResolver` is the default `DocumentReferenceResolver`, which means that the missing wiki part is always going to be resolved as "xwiki" so this code does not really work in subwikis.

manuelleduc: Even when the wiki in the context is a subwiki?

tmortagne: > Even when the wiki in the context is a subwiki?

The default resolver does not care about that, you need the "current" resolver to take the context into account.

manuelleduc: See https://github.com/xwiki/xwiki-platform/pull/2241 for the bugfix of that part.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
```diff
@@ -65,28 +80,15 @@ public abstract class AbstractDocumentsMigration extends AbstractHibernateDataMi
...
-        try {
-            return Optional.of(
-                context.getWiki().getDocument(this.documentReferenceResolver.resolve(documentReference), context));
-        } catch (XWikiException e) {
-            this.logger.error("Failed to resolve [{}]", documentReference, e);
-            return Optional.empty();
+        for (ReferenceWithLocale referenceWithLocale : selectedDocuments) {
+            XWikiDocument document =
+                new XWikiDocument(this.documentReferenceResolver.resolve(referenceWithLocale.reference),
+                    toLocale(referenceWithLocale.locale, ROOT));
```
tmortagne, line 89: Note that `toLocale` might fail here when the document in the database has an unsupported locale (it's unlikely, but it did happen not long ago) and from what I understand it would crash the rest of the migration.

manuelleduc: Can `toLocale` with a fallback fail?

manuelleduc: For instance, the following returns `true`

```java
assertEquals(Locale.CANADA, LocaleUtils.toLocale("aaaaaa", Locale.CANADA));
```

tmortagne: Indeed, it cannot fail if you pass a fallback. But it does not make sense to fallback on ROOT, if that document has an unsupported locale it should be skipped.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/R150000000XWIKI20285DataMigration.java
```diff
@@ -73,22 +76,26 @@ protected String getTaskType()
...
                 .stream()
+                .map(array -> new ReferenceWithLocale(String.valueOf(array[0]), String.valueOf(array[1])))
                 // Exclude document that are provided by extensions, because they are fixed using the usual xar upgrade
                 // mechanism.
-                .filter(documentReference -> !this.installedXARs.isExtensionDocument(
-                    this.documentReferenceResolver.resolve(documentReference)))
+                .filter(documentReference -> {
+                    DocumentReference resolve = this.documentReferenceResolver.resolve(documentReference.reference);
+                    return !this.installedXARs.isExtensionDocument(
+                        new DocumentReference(resolve, toLocale(documentReference.locale, ROOT)));
```
tmortagne, line 97: Same potentially unsupported locale problem as the one discussed above.

tmortagne: Actually, I'm wondering if it wouldn't be cleaner to make `selectDocuments` return a List<DocumentReference>, that would avoid having to resolve the reference twice in the case of R150000000XWIKI20285DataMigration (and it makes more sense as an API, even if AbstractDocumentsMigration is internal).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
```diff
@@ -98,25 +104,37 @@ private Optional<XWikiDocument> convert(String documentReference)
...
     }
 
     /**
      * Prints an info logs with an individual document and well as its queued task.
      *
-     * @param document a individual document that will be queued
+     * @param reference a unique document reference that will be queued
      */
-    protected void logBeforeQueuingTask(XWikiDocument document)
+    protected void logBeforeQueuingTask(ReferenceWithLocale reference)
```
tmortagne, line 124: A since would be nice now that the signature changed.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
```diff
@@ -98,25 +104,37 @@ private Optional<XWikiDocument> convert(String documentReference)
...
-    protected abstract List<String> selectDocuments() throws DataMigrationException;
+    protected abstract List<ReferenceWithLocale> selectDocuments() throws DataMigrationException;
 
     /**
      * Prints an info log with the number of queued documents and the type of the task.
      *
      * @param documents the full list of documents that will be queued
      */
-    protected void logBeforeQueuingTasks(List<XWikiDocument> documents)
+    protected void logBeforeQueuingTasks(List<ReferenceWithLocale> documents)
```
tmortagne, line 114: A new `@since` could be interesting since, while the signature did not technically change, what is expected the List to contain did.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/AbstractDocumentsMigration.java
```diff
@@ -98,25 +104,37 @@ private Optional<XWikiDocument> convert(String documentReference)
     /**
      * @return the list of document ids to migrate
      */
-    protected abstract List<String> selectDocuments() throws DataMigrationException;
+    protected abstract List<ReferenceWithLocale> selectDocuments() throws DataMigrationException;
```
tmortagne, line 107: A new @since could be interesting since, while the signature did not technically change, what is expected the List to contain did.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/migration/R150000000XWIKI20285DataMigration.java
```diff
@@ -73,22 +75,33 @@ protected String getTaskType()
...
                 .stream()
+
                 // Exclude document that are provided by extensions, because they are fixed using the usual xar upgrade
                 // mechanism.
-                .filter(documentReference -> !this.installedXARs.isExtensionDocument(
-                    this.documentReferenceResolver.resolve(documentReference)))
+                .map(array -> resolveDocumentReference(String.valueOf(array[0]), String.valueOf(array[1])))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .map(documentReference -> {
```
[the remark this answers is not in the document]
manuelleduc: Indeed... I also slightly improved the map above, to avoid having to  do the ugly `filter(Optional::isPresent).map(Optional::get)` chaining

tmortagne (approved):

PR #2243, opened 2023-07-05
XWIKI-21103: Add support for executing actions on the page tree from a context menu

surli:
* Provide actions in the context for pages
WIP

### DIFF
--- a/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/DocumentTreeMacros.xml
+++ b/xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/DocumentTreeMacros.xml
@@ -20,7 +20,7 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-index/xwiki-platform-index-tree/xwiki-platform-index-tree-macro/src/main/resources/XWiki/DocumentTreeMacros.xml
```diff
@@ -878,41 +886,65 @@
...
+      'action': 'openLink',
+      'parameters': {
+        'urlProperty': 'editURL'
+      }
     },
-    'cut': {
-      'separator_before': true,
-      'label': 'Cut',
-      'icon': 'fa fa-scissors'
+    'changeTitle': {
```
surli, line 905: This is a mock operation in the context menu right now, it's not yet implemented.

surli:
Right now moving a node inside the same parent triggers a move, which is wrong and needs to be fixed. Also the panel edition from administration is broken (it shouldn't trigger the move operation in case of d&d)

surli:
There's a bug when trying to move a terminal page to the root space, it should not be authorized.

surli:
Also the menu is wrongly enabling the create child page on a terminal page.

PR #2244, opened 2023-07-05
XWIKI-21108: "Border" of the page content area in edit mode is not of…

lucaa:
… the same color as in view mode

* Reset the color of the border of the content area in edit mode to xwiki-border-color instead of input-border


Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-21108

LGTM :+1:

PR #2246, opened 2023-07-11
XWIKI-21154: Lack of space in the bottom of the login form

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-21154

## PR Change
* Added a gap in between the login form submit button and the recovery anchors

## Note
I made this gap 1em large, but this is completely arbitrary.

## Visuals
18859-2-After
vvv with the box model highlighted vvv
18859-2-After2


michitux:
As this is a user-visible change, I think it would be better to create a Jira issue for it.

PR #2251, opened 2023-07-13
XWIKI-21136: On upgrade, the old version of an extension is not marked as not installed on solr

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21136

### DIFF
--- a/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/internal/ExtensionIndexSolrUtil.java
+++ b/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/internal/ExtensionIndexSolrUtil.java
@@ -38,7 +38,6 @@ import org.xwiki.extension.repository.InstalledExtensionRepository;
 import org.xwiki.search.solr.SolrUtils;
 
 import static org.xwiki.extension.InstalledExtension.FIELD_INSTALLED_NAMESPACES;
-import static org.xwiki.search.solr.SolrUtils.ATOMIC_UPDATE_MODIFIER_REMOVE;
 import static org.xwiki.search.solr.SolrUtils.ATOMIC_UPDATE_MODIFIER_SET;
 
 /**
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/internal/ExtensionInstallListener.java
```diff
@@ -72,15 +74,32 @@ public void onEvent(Event event, Object source, Object data)
...
         } catch (Exception e) {
             this.logger.error("Failed to update the local extension store", e);
         }
     }
 
+    private void onUpgraded(ExtensionId extensionId, Collection<InstalledExtension> previousExtensions,
+        String namespace) throws SolrServerException, IOException
+    {
+        for (InstalledExtension previousExtension : previousExtensions) {
+            if (previousExtension.getNamespaces().contains(namespace)) {
```
tmortagne, line 94: I don't understand why you check this:
* the extension cannot exist in the same namespace if it's listed as previous extension
* this logic does not work with the root (`null`) namespace anyway, since the list is empty in that case

manuelleduc: Removed, in indeed realized in the meantime that I was causing NPEs

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/internal/ExtensionIndexSolrUtil.java
```diff
@@ -0,0 +1,174 @@
...
+            }
+            this.utils.setAtomic(ATOMIC_UPDATE_MODIFIER_SET, InstalledExtension.FIELD_INSTALLED_NAMESPACES,
+                installedNamespaces, document);
+
+            // We can already set those extensions as "incompatible" with those namespaces
+            this.utils.set(ExtensionIndexSolrCoreInitializer.SOLR_FIELD_INCOMPATIBLE_NAMESPACES, installedNamespaces,
+                document);
+        } else {
+            this.utils.setAtomic(ATOMIC_UPDATE_MODIFIER_REMOVE, InstalledExtension.FIELD_INSTALLED_NAMESPACES,
+                List.of(), document);
```
manuelleduc, line 91: Check if this new else branch is relevant and does not have bad side effects.

PR #2258, opened 2023-07-18
XWIKI-20938: Add ability to insert an image using the keyboard

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-20938

This pull request adds a new Quick Action to quickly upload or search for and insert an image into the WYSIWYG editor.

preview

### DIFF
--- a/xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
+++ b/xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
@@ -29,8 +29,11 @@ define('xwiki-attachment-picker-translation-keys', {
 });
 
 define('xwiki-attachment-picker',
-  ['jquery', 'xwiki-attachments-icon', 'xwiki-l10n!xwiki-attachment-picker-translation-keys'],
-  function ($, attachmentsIcon, translations) {
+  ['jquery',
+  'xwiki-attachments-icon',
+  'xwiki-l10n!xwiki-attachment-picker-translation-keys',
+  'xwiki-attachment-picker-solr-search'],
+  function ($, attachmentsIcon, translations, SolrSearch) {
     'use strict';
 
     const ATTACHMENT_PICKER_INITIALIZED_CLASS = 'initialized';
@@ -49,108 +52,6 @@ define('xwiki-attachment-picker',
         clearTimeout(timeout);
         timeout = setTimeout(() => callback.apply(this, args), delay);
       };
-    }
-
-    /**
-     * Handle the interaction with the solr search endpoint.
-     */
-    class SolrSearch {
-      constructor(options) {
-        const doc = new XWiki.Document(XWiki.Model.resolve('XWiki.SuggestSolrService', XWiki.EntityType.DOCUMENT));
-        this.sugestSolrServiceURL = doc.getURL('get');
-        this.options = options || {};
-        this.limit = options.limit || 20;
-        this.target = options.target;
-        this.solrOptions = this.options.solrOptions || {};
-      }
-
-      search(query, isGlobal) {
-        const {currentWiki, currentSpace, currentPage} = this.resolveTargetFqs();
-        const localDocumentOnly = this.searchSolr(query, Object.assign({}, this.solrOptions, {
-          fqs: [
-            `wiki:${currentWiki}`,
-            `space:"${currentSpace}"`,
-            `name:"${currentPage}"`
-          ]
-        }));
-
-        // TODO: Make this algorithm configurable through an macro parameter.
-        let values;
-        if (isGlobal) {
-          const globalSearch = this.searchSolr(query, this.solrOptions);
-          values = [localDocumentOnly, globalSearch];
-        } else {
-          values = [localDocumentOnly];
-        }
-        return Promise.all(values).then(results => {
-          /*
-           * Take the attachments from the current document first. Then take the attachments from the global search that
-           *  are not from the current document (to avoid duplicates). At the end, only keep a fixed number of 
-           * attachments (this.nb).
-           */
-          const localAttachments = results[0];
-          localAttachments.forEach(attachment => attachment.isLocal = true);
-          const localAttachmentsId = localAttachments.map(attachment => attachment.id);
-
-          // Check if global attachments where request before trying to merge them while removing the duplicates.
-          let globalAttachments = [];
-          if (results[1]) {
-            globalAttachments = results[1].filter(it => !localAttachmentsId.includes(it.id));
-          }
-          globalAttachments.forEach(attachment => attachment.isLocal = false);
-          return localAttachments.concat(globalAttachments).slice(0, this.limit);
-        });
-      }
-
-      resolveTargetFqs() {
-        let currentWiki = XWiki.currentWiki;
-        let currentSpace = XWiki.currentSpace;
-        let currentPage = XWiki.currentPage;
-        if (this.target) {
-          const target = new XWiki.Document(XWiki.Model.resolve(this.target, XWiki.EntityType.DOCUMENT));
-          if (target.wiki) {
-            currentWiki = target.wiki;
-          }
-          currentSpace = target.space;
-          currentPage = target.page;
-        }
-        return {currentWiki, currentSpace, currentPage};
-      }
-
-      searchSolr(input, options) {
-        options = options || {};
-        const optionsFqs = options.fqs || [];
-        const types = options.filter;
-        var typesFqs;
-        if (types !== undefined && types !== '') {
-          typesFqs = ['mimetype:(' + types.split(",").map(type => "(" + type + ")").join(" OR ") + ')'];
-        } else {
-          typesFqs = [];
-        }
-        const computedFqs = ['type:ATTACHMENT'].concat(optionsFqs).concat(typesFqs);
-        const query = computedFqs.map((fq) => 'fq=' + fq).join('\n');
-
-        return new Promise((resolve, reject) => {
-          // TODO: handle more kind of scopes
-          if (input === '') {
-            // Replace the empty string with '*', matching everything. 
-            input = '*';
-          }
-          $.getJSON(this.sugestSolrServiceURL, {
-            'query': query,
-            'nb': this.limit,
-            'media': 'json',
-            'input': input
-          })
-            .done(function (response) {
-              resolve(response);
-            })
-            .fail(function (_jqxhr, textStatus, error) {
-              reject(`${textStatus} ${error}`);
-            });
-
-        });
-      }
     }
 
     /**
--- a/xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/solrSearch.js
+++ b/xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/solrSearch.js
@@ -1,0 +1,135 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+define('xwiki-attachment-picker-solr-search', ['jquery'], function($) {
+      /**
+     * Handle the interaction with the solr search endpoint.
+     */
+    class SolrSearch {
+      constructor(options) {
+        const doc = new XWiki.Document(XWiki.Model.resolve('XWiki.SuggestSolrService', XWiki.EntityType.DOCUMENT));
+        this.sugestSolrServiceURL = doc.getURL('get');
+        this.options = options || {};
+        this.limit = options.limit || 20;
+        this.target = options.target;
+        this.solrOptions = this.options.solrOptions || {};
+      }
+
+      search(query, isGlobal) {
+        const {currentWiki, currentSpace, currentPage} = this.resolveTargetFqs();
+        const localDocumentOnly = this.searchSolr(query, Object.assign({}, this.solrOptions, {
+          fqs: [
+            `wiki:${currentWiki}`,
+            `space:"${currentSpace}"`,
+            `name:"${currentPage}"`
+          ]
+        }));
+
+        // TODO: Make this algorithm configurable through an macro parameter.
+        let values;
+        if (isGlobal) {
+          const globalSearch = this.searchSolr(query, this.solrOptions);
+          values = [localDocumentOnly, globalSearch];
+        } else {
+          values = [localDocumentOnly];
+        }
+        return Promise.all(values).then(results => {
+          /*
+           * Take the attachments from the current document first. Then take the attachments from the global search that
+           *  are not from the current document (to avoid duplicates). At the end, only keep a fixed number of
+           * attachments (this.nb).
+           */
+          const localAttachments = results[0];
+          localAttachments.forEach(attachment => attachment.isLocal = true);
+          const localAttachmentsId = localAttachments.map(attachment => attachment.id);
+
+          // Check if global attachments where request before trying to merge them while removing the duplicates.
+          let globalAttachments = [];
+          if (results[1]) {
+            globalAttachments = results[1].filter(it => !localAttachmentsId.includes(it.id));
+          }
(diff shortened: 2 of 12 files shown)


vmassol:
Looks good, thanks. Question: why, when selecting the image quick action, the prompt changes to `img:`?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
```diff
@@ -128,14 +106,24 @@ define('xwiki-attachment-picker',
...
+          computedParams.push('sort=' + options.sort);
+        }
+
+        const query = computedParams.join('\n');
 
         return new Promise((resolve, reject) => {
           // TODO: handle more kind of scopes
           if (input === '') {
-            // Replace the empty string with '*', matching everything. 
+            // Replace the empty string with '*', matching everything.
```
manuelleduc, line 122: In general, we tend to avoid minor changes like this in a PR as it make it slightly more difficult to reviews ;)
But they are definitely nice to have in a separate misc commit.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
```diff
@@ -28,30 +28,8 @@ define('xwiki-attachment-picker-translation-keys', {
...
-    function debounce(callback, delay) {
-      let timeout;
-      return function (...args) {
-        clearTimeout(timeout);
-        timeout = setTimeout(() => callback.apply(this, args), delay);
-      };
-    }
-
-    /**
+define('xwiki-solr-search', ['jquery'], function($) {
```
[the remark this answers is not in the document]
manuelleduc: Or `xwiki-attachment-picker-solr-search`, and in this case it could stay here, as iiuc you extracted this code to reuse independently

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/imageWizard.js
```diff
@@ -48,6 +48,11 @@ define('imageWizard', ['imageSelector', 'imageEditor'], function(imageSelector,
     if (CKEDITOR.currentInstance) {
       params.currentDocument = CKEDITOR.currentInstance.config.sourceDocument.documentReference;
     }
+
+    if (params.setImageData) {
+      return $.Deferred().resolve(params.setImageData);
```
manuelleduc, line 53: Is it mandatory to use jquery deferred here? Javascript no provides a promise api (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -100,13 +101,283 @@
       }, this);
     },
     init: function(editor) {
+
+      // Register the the /img:: autocomplete
+      editor.config.mentions = editor.config.mentions || [];
```
manuelleduc, line 106: not sure to understand here, you mention img autocomplete in the comment but manipulate mentions just below.

manuelleduc: ah, unless 'mentions' is the name of the feature in CKEditor? In this case might be good to explain it in the comments as it might confuse somebody else :)

douakli:
> Looks good, thanks. Question: why, when selecting the image quick action, the prompt changes to `img:`?

We thought that we might end up adding too many shortcuts if we do one specific for each kind of autocomplete.
We already have one for links and mentions, now the /. And we are going to add image and icon which are going to be implemented similarly. So we went for a trigger that wouldn't be typed in without the goal of starting a search. I do not think we expect a double column in the Quick Action query or that it would provide any result anyway, using it to define kind of sub actions is a way to achieve the goal of not interfering with common usage.

I'm open to suggestions, of course.

Thanks.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
douakli: Thank you, I'll remove the change, my editor removed the trailing space without me noticing :)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -100,13 +101,283 @@
...
+              // Show the wildcard query's results
+              callback(wildcardItems);
+
+              return wildcardAttachments;
+            });
+
+            // Display all the results once the solr requests are finished
+            Promise.all([
+              // Display the specific query results higher in the list
+              localQuery, globalQuery, wildcardQuery
```
manuelleduc, line 240: Is there a risk that this Promise.all is called before the one above?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -100,13 +101,283 @@
...
+            ]).then(function (values) {
+
+              const finalItems = [...topItems];
```
manuelleduc, line 243: Note that is assignment is creating an entirely new "cloned" array. Very small perf improvement but you could just assign topItems as the inital array afaics (since it is not manipulated in the rest of the method)

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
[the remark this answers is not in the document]
douakli: It would make sense to move it to a new `xwiki-platform-attachment-webjar`. I'm not sure I should prioritize this as I never done it and it would take me some time to figure out how to do. Is it okay to move it in the next release if I don't find the time to do it now?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/solrSearch.js
```diff
@@ -0,0 +1,135 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+define('xwiki-solr-search', ['jquery'], function($) {
```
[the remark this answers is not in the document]
douakli: Yes, I forgot to remove it from attachmentGalleryPicker.js, thank you.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
```diff
@@ -344,8 +344,9 @@
           // Get the text before the caret.
           var left = text.slice(0, offset),
               // Will look for the marker followed by text.
-              match = left.match(new RegExp(escapeRegExp(config.marker) + '.{0,30}$'));
-          if (match) {
+              match = left.match(new RegExp(escapeRegExp(config.marker) + '.{0,30}$')),
+              subMatch = left.match(new RegExp(escapeRegExp(config.marker) + '.{0,30}::\\S{0,30}$'));
```
manuelleduc, line 348: Might be interesting to develop a bit what's going on in the comments. Looks magic.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/imageWizard.js
douakli: I believe it was until the changes Marius made to the CKEditor integration to allow for newest javascript features. I implemented that part before he did the change. jquery deferred is used in other ckeditor plugins we defined. We could replace it indeed. I'll change it here.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
```diff
@@ -619,7 +620,7 @@
           name: editor.lang.common.image,
           iconClass: 'fa fa-image',
           description: editor.localization.get('xwiki-slash.action.img.hint'),
-          command: 'image'
+          outputHTML: '/img::'
```
manuelleduc, line 623: should the 'comment' key be removed here?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/XWiki/SuggestSolrMacros.xml
```diff
@@ -148,8 +148,17 @@
     #set ($limit = 3)
   #end
   #set ($discard = $query.setLimit($limit))
+  #set ($macro.results = $query.execute()[0].results)
+  #foreach ($result in $macro.results)
+    #set ($resultReference = $services.solr.resolve($result))
+    #set ($options = {
+      'excludeSelf': true,
+      'limit': 4
+    })
```
manuelleduc, line 157: Looks unused

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
douakli: I can see how this is confusing, the ckeditor mentions plugin is based on the autocomplete plugin, so technically it is also an autocomplete. I can change it to mention to avoid confusion

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
douakli: I'm not sure I understand the question, can you elaborate?
The Promise.all will be called after the one above, but if you're talking about the callbacks I'm using the results of the one above in this one. (The wildcardQuery results appear, but at the end of the list)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
douakli: Indeed, I'll change it, thank you.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
douakli: I'm making sure the dropdown doesn't show when we add `::` in a slash query. I will add comments, thanks.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
manuelleduc: Yes, but it is possible for the resolve of the first one to be called after the resolve of the second one.
Meaning that the wildcards would appear one top by accident.

douakli: From my understanding of Promise.all, since I'm storing the newly generated Promise in the variable wildcardQuery and reusing it in the second Promise.all, it guarantees that the second call resolves after the first call.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/attachmentService.js
```diff
@@ -0,0 +1,68 @@
...
+    return new Promise(function (resolve) {
+
+      var cached = attachmentsByQuery[(isGlobal ? "global" : "local")][query];
+
+      if (cached && !force) {
+        return resolve(cached);
+      }
+
+      const solrSearch = new SolrSearch({
+          limit: 100,
```
manuelleduc, line 49: Do we really need to fetch that many attachments?
We are calling `getAttachments` four times on each (debounced) keystroke, meaning a potential 400 entries (some of the as duplicates).
WDYT of lowering this to say, 40? I doubt anybody would go that low in the list.
Future improvement could be to lazily show additional results when the user reaches the end of the list.
And also to show a message indicating if this is the end of the list, or if there is additional results.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/attachmentService.js
```diff
@@ -0,0 +1,68 @@
...
+          target: XWiki.Model.serialize(editor.config.sourceDocument.documentReference),
+          solrOptions: {
+            sort: "attdate_sort " + sort
+          }
+      });
```
manuelleduc, line 54:
```suggestion
      const solrSearch = new SolrSearch({
        limit: 100,
        target: XWiki.Model.serialize(editor.config.sourceDocument.documentReference),
        solrOptions: {
          sort: "attdate_sort " + sort
        }
      });
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -100,13 +101,284 @@
...
+        minChars: 0,
+        itemsLimit: 5,
+        itemTemplate: '<li data-id="{id}">' +
+
+          // Allow different display for non image entries.
+          '<div class="ckeditor-autocomplete-item-head {extraClass}">' +
+
+            '<span class="ckeditor-autocomplete-item-{type}-wrapper">' +
+              '<span class="{iconClass}" aria-hidden="true"></span>' +
+              '<img src="{imgSrc}" aria-hidden="true"></img>' +
```
manuelleduc, line 120:
```suggestion
              '<img src="{imgSrc}" aria-hidden="true"/>' +
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -100,13 +101,284 @@
...
+                    );
+                    return Promise.reject();
+                  });
+                });
+              });
+
+            });
+
+          return "";
+          }
```
manuelleduc, line 353:
```suggestion
            return "";
          }
```

The method above could benefit from some auto-format, the indentation if off and some lines are too long for instance.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/XWiki/SuggestSolrMacros.xml
```diff
@@ -148,8 +148,13 @@
     #set ($limit = 3)
   #end
   #set ($discard = $query.setLimit($limit))
+  #set ($macro.results = $query.execute()[0].results)
+  #foreach ($result in $macro.results)
+    #set ($resultReference = $services.solr.resolve($result))
+    #set ($result.location = "#hierarchy($resultReference {'local': true, 'excludeSelf': true, 'plain': true})")
+  #end
```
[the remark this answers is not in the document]
douakli: Thank you Michael.
That's a good catch, I did not realize the performance implications when this was written.
This should have been implemented as a new service similar to `xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/LinkSuggestions.xml`, and also probably only for the results that are actually being displayed rather than all the documents.

PR #2259, opened 2023-07-18
XWIKI-21039: Don't dispose security cache entries hierarchically when the cache is full

michitux:
* Add a reference map to store entries that are needed for the hierarchy but that are no longer in the cache.
* Use weak references for children to allow children to just disappear and to allow the GC to remove the parts of the hierarchy that are no longer referenced by the cache.
* Don't trigger hierarchical cache invalidation when an entry is removed from the cache.
* Fix adding shadow parents in the security cache test.
* Modify the tests to expect the new behavior of not triggering hierarchical removal when entries are removed from the cache.

Jira issue: https://jira.xwiki.org/browse/XWIKI-21039

### DIFF
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
@@ -453,12 +453,12 @@ public class DefaultSecurityCache implements SecurityCache, Initializable
         public void dispose()
         {
             if (!disposed) {
+                DefaultSecurityCache.this.cache.remove(getKey());
+                DefaultSecurityCache.this.internalEntries.remove(getKey());
                 disposed = true;
 
                 disconnectFromParents();
                 disposeChildren();
-                DefaultSecurityCache.this.cache.remove(getKey());
-                DefaultSecurityCache.this.internalEntries.remove(getKey());
             }
         }
 
@@ -659,6 +659,14 @@ public class DefaultSecurityCache implements SecurityCache, Initializable
         return getInternal(getShadowEntryKey(userReference, wiki));
     }
 
+    /**
+     * Get a security cache entry from the cache or the internal map. In the latter case, the entry is re-inserted
+     * into the cache. This method can be called without locking, it uses the read lock internally.
+     *
+     * @param key the key of the entry to retrieve
+     * @throws IllegalStateException if the entry has been disposed (this should never happen)
+     * @return the entry corresponding to the given key, null if none is available in the cache
+     */
     private SecurityCacheEntry getInternal(String key)
     {
         readLock.lock();
@@ -678,6 +686,12 @@ public class DefaultSecurityCache implements SecurityCache, Initializable
                     this.cache.set(key, result);
                 }
             }
+
+            if (result != null && result.disposed) {
+                throw new IllegalCacheStateException(
+                    String.format("Entry [%s] has been disposed without being removed from the cache.", result));
+            }
+
             return result;
         } finally {
             readLock.unlock();
@@ -728,23 +742,21 @@ public class DefaultSecurityCache implements SecurityCache, Initializable
     /**
      * Add a new entry in the cache and prevent cache container deadlock (in cooperation with the entry dispose method)
      * in case adding the entry cause this same entry to be evicted.
-     * 
+     *
      * @param key the key of the entry to be added.
      * @param entry the entry to add.
-     * @throws ConflictingInsertionException when the entry have been disposed while being added, the full load should
-     *             be retried.
      */
-    private void addEntry(String key, SecurityCacheEntry entry) throws ConflictingInsertionException
+    private void addEntry(String key, SecurityCacheEntry entry)
     {
         cache.set(key, entry);
         this.internalEntries.put(key, entry);
+
         if (entry.disposed) {
-            // XWIKI-13746: The added entry have been disposed while being added, meaning that the eviction
-            // triggered by adding the entry has hit the entry itself, so remove it and fail.
-            // TODO: this should be impossible now.
-            throw new ConflictingInsertionException(String.format(
-                "The cache entry [%s] with key [%s] has been disposed by another thread while being added.", entry,
-                key));
+            // This should never happen as entries cannot be disposed while being added to the cache as both
+            // operations require the write lock. However, if it happens, there is a serious bug in the code so
+            // better fail with an exception.
+            throw new IllegalCacheStateException(
+                String.format("Entry [%s] has been disposed while being added to the cache.", entry));
         }
     }
 
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
@@ -23,8 +23,10 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.HashSet;
+import java.util.WeakHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -33,10 +35,11 @@ import javax.inject.Inject;
 import javax.inject.Singleton;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections4.map.AbstractReferenceMap;
+import org.apache.commons.collections4.map.ReferenceMap;
 import org.slf4j.Logger;
 import org.xwiki.cache.Cache;
 import org.xwiki.cache.CacheManager;
-import org.xwiki.cache.DisposableCacheValue;
 import org.xwiki.cache.config.CacheConfiguration;
 import org.xwiki.cache.eviction.EntryEvictionConfiguration;
 import org.xwiki.cache.eviction.LRUEvictionConfiguration;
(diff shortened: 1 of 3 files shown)


michitux:
This will break the [security inspector application](https://extensions.xwiki.org/xwiki/bin/view/Extension/Security%20Inspector%20Application/) in unexpected ways as soon as the cache is full and discards any entries. Also, the inspector may initially display entries that are not in the cache if they can still be reached through parent/child relationships and won't be notified when they are discarded by the GC. It would be nice if we could use an actual cache that supports similar inspection for the reference map. This would actually be possible if we directly used Caffeine for caching which [supports](https://github.com/ben-manes/caffeine/wiki/Eviction#reference-based) this. Unfortunately, we use Infinispan which is based on Caffeine but doesn't expose this feature from what I found (I searched and couldn't find anything). I unfortunately also cannot see how we can fix the security inspector application without using something like Caffeine. The security cache inspector is hardly used and primarily makes sense when the cache is small and thus didn't discard any entries so I don't think this is a big issue and I'll just document it if you agree.

We also loose some observability of the cache size as the size of the reference map won't be reported using JMX. Again, this could be fixed by using Caffeine which supports JMX. I think this is not a big issue as we use weak references, so everything that is not in the actual cache is cleared on every (full) GC run (as confirmed by the unit test) and the only thing that remains are internal entries of the hierarchy. The number of internal entries not in the cache should be small compared to the number of entries in the cache unless you have, e.g., a lot more groups than users or entries that are deep in a space hierarchy where nothing else is cached on that hierarchy, which should, again, be rare.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
```diff
@@ -131,12 +141,14 @@ private Cache<SecurityCacheEntry> newCache() throws InitializationException
     public void initialize() throws InitializationException
     {
         cache = newCache();
+        this.internalEntries = new ReferenceMap<>(AbstractReferenceMap.ReferenceStrength.HARD,
+            AbstractReferenceMap.ReferenceStrength.WEAK);
```
surli, line 145: Isn't this an open door for memory leak: I mean if the cache reach a given size, entries are automatically removed, but the map will still keep the entries no? So isn't there a risk for this map to grow indefinitely? 

You said in the javadoc:
> When an entry is removed from the cache, the GC will also remove it from the internal entries unless it is still referenced as a parent somewhere else.

So it means it could be the case no? Unless you assume that the risk is very low that a lot of references are kept without being in the cache anymore?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/test/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCacheTest.java
```diff
@@ -813,20 +815,26 @@ void testCacheEvictedEntries() throws Exception
...
+            {
+                this.cache.remove(entry.getKey());
             }
         }
 
-        removerTest(entries, new Remover()
+        // For a GC to try emptying the reference map as much as possible to purge the docRef/userRef access entry.
+        // All "internal" entries should be kept as they should be referenced through parents of the access entries
+        // still in the cache.
+        System.gc();
```
manuelleduc, line 831:
```suggestion
        // Force a GC to try emptying the reference map as much as possible to purge the docRef/userRef access entry.
        // All "internal" entries should be kept as they should be referenced through parents of the access entries
        // still in the cache.
        System.gc();
```

Also, possibly because I don't know well this part of the code but I don't get from the comments why calling the gc is relevant. Asking as it is quite unusual to explicitly call it.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
```diff
@@ -664,10 +655,30 @@ private SecurityCacheEntry getEntry(UserSecurityReference userReference, Securit
      *         cache.
      */
     private SecurityCacheEntry getShadowEntry(SecurityReference userReference, SecurityReference wiki)
+    {
+        return getInternal(getShadowEntryKey(userReference, wiki));
+    }
+
+    private SecurityCacheEntry getInternal(String key)
```
manuelleduc, line 662: Could be interesting to have an explanation of why getInternal is required instead of a simple get.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/test/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCacheTest.java
surli: > I don't get from the comments why calling the gc is relevant

Well it's explained in the comment, it's to trigger the emptying of the `ReferenceMap`, I didn't know about such map implementation, but it's a map that aims at getting its elements removed thanks to the GC, see: https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/ReferenceMap.html

michitux: Yeah, so basically this map stores only weak references to the values (as I configured it that way) so values will be garbage collected if there are no other strong references to them. The garbage collector notifies the map which values have been garbage collected so the map can properly remove them.

manuelleduc: > Well it's explained in the comment, it's to trigger the emptying of the `ReferenceMap`

ok, but why do you need to explicitly empty it at this exact point in time.
Also, garbage collection the whole VM just to clean a single collection feels super heavy.

Repeating: I don't know the domain so I might be asking silly questions :)

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
```diff
@@ -725,19 +732,15 @@ private boolean isAlreadyInserted(String key, SecurityEntry entry, Collection<Gr
...
-                    key));
-            }
-        } finally {
-            newEntry = null;
+        cache.set(key, entry);
+        this.internalEntries.put(key, entry);
+        if (entry.disposed) {
+            // XWIKI-13746: The added entry have been disposed while being added, meaning that the eviction
+            // triggered by adding the entry has hit the entry itself, so remove it and fail.
+            // TODO: this should be impossible now.
```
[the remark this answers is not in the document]
michitux: Good point, so I think it might make sense to keep an exception just in case, but maybe additionally log an error. The reason why this is impossible is that disposing now only happens with an explicit removal operation and this cannot happen in parallel. The cache also cannot trigger this anymore so there is no unknown cache operation that could trigger this. Maybe we could add some more checks in some places to ensure that disposed entries are really never returned anywhere.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/test/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCacheTest.java
manuelleduc: Sorry, I'd missed it was for a test... Perfectly fine here of course :)

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
michitux: I recommend reading the documentation of the [security inspector application](https://extensions.xwiki.org/xwiki/bin/view/Extension/Security+Inspector+Application) as I think it is the best description that we have of the contents of the security cache. Ignoring some special cases, we have the following cases:

1. Security access entries. They store the outcome of a right computation for a specific user/document combination. Those are never the parent of another entry so they should always be collected by the GC when they are disposed from the actual cache.
2. Entries that are representing regular documents, spaces or wikis. They store the access rules that are associated with the document/space/wiki. They can be the parent of another document, space, access entry, user or group. In any case, they must have some descendant that is actually in the size-limited cache and the number of documents above them is limited by the depth of the hierarchy. Further, their number is naturally limited by the number of documents in the XWiki installation. If you have deep hierarchies and the cache is just storing an access entry, you could indeed have, say, 5 times as many entries in the internal map than in the size-constrained cache. That said, any computation of a security access entry accesses the whole hierarchy and thus tries their re-insertion in the size-constrained cache. Further, in the UI, we frequently display the whole path to a document and thus check the whole hierarchy. So I think the chances that you have a long hierarchy without many entries in the size-constrained cache are low.
3. Entries that are representing users or groups. These are basically the same as documents but additionally they have parent groups as parents. Again, I don't expect huge hierarchies there. Also, we don't store all members of a group so even huge groups are not an issue.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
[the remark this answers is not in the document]
michitux: I've added a dedicated, unchecked exception for that case now so we noticed clearly if this happens for whatever reasons we didn't imagine (the exception should be catched and logged as an error in [`DefaultAuthorizationManager`](https://github.com/xwiki/xwiki-platform/blob/4a7bf51c92a63aea65cb51078142bba0bf0a9bfe/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/DefaultAuthorizationManager.java#L121)).

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/cache/internal/DefaultSecurityCache.java
michitux: To provide some context why all of this work is necessary and why having some more entries in memory might be good:

Imagine that we have a wiki that has primarily public access. Ideally, almost all security access entries for public access and all security access entries for author right checks should be in the security cache. Therefore, only these are accessed and nothing of the hierarchy behind them. Imagine now that the size of the security cache is not enough to store all of them, so we just store the most popular ones. Now the cache needs to decide which ones to keep. The cache actually has access frequency data, at least as an approximate sketch. So it can make informed decisions that these entries that are representing regular documents are never accessed and it will remove them. Currently, this will also remove the frequently accessed security access items that have them as parents. With this change, instead, these "internal" entries will remain in the internal map and the frequently accessed security access items can remain in the cache as the hierarchy is intact (if the document is modified, it needs to be removed from the cache including all its children, for that purpose we need to be able to find it).

So in this scenario we will indeed store more than originally configured, but I think this is actually good.

surli (approved):

PR #2261, opened 2023-07-19
XWIKI-20867: Low contrast on Iceberg validation message

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20867
## PR Changes
* Removed the specific value for `@btn-success-bg` on the Iceberg color theme and used this value for `@brand-success` instead
* Fixed the skinless theme too
## View
20867-AfterFix

## Note
This is not the only place where this low contrast was triggering WCAG violation in automated tests.


Sereza7:
@michitux 
Good question, it took me a while to find back why I did this and I didn't write anything about it anywhere ^^'

I removed the values for `@btn-success-bg` when the color theme is set as Iceberg (Iceberg.xml > L893), since the default value for `@btn-success-bg` defined in the bootstrap-3.4.1 `variables.less` is `@brand-success`.
Since now those two values have the same 

The removal on line 251 of variablesInit.vm is to fix a typo I introduced a few months ago. The correct value was already on line 253 and I kept it until now. I removed it in https://github.com/xwiki/xwiki-platform/pull/2261/commits/da2dadec6f05e7ee008f4339c6a56972af3c262b, and it's still working as expected because bootstrap is still loaded even without a colortheme (and same reason as Iceberg).

## Note

The color of the popup message 'Saved' on the bottom of the screen is changed by this PR, which is correct too because it's white on green, and didn't have enough contrast either. It was following the old theme color `notificationSuccessColor` which is mapped to `@brand-success`. It didn't change color with the previous changes to `@btn-success-bg`.
I suppose we could change the `@btn-danger-bg` in the same way too, but so far I didn't notice somewhere where it would be a contrast issue yet. It would just make more sense for code quality and clearness to just have a list with all four main brand colors (pirmary, success, warning, danger) changed the same way.

Sereza7:
> Another comment, after reading [this blog article about color in UI design](https://www.learnui.design/blog/color-in-ui-design-a-practical-framework.html), I wondered if it wouldn't make sense to increase the saturation of this color. And indeed, I think it looks nicer (greener) when increasing the saturation, e.g., to #077537.

:+1: and I checked, this also improves contrast on white :) (so probably improves for all shades of light grey)


> Now of course it makes sense to keep this consistent with @brand-success, so maybe we should change @brand-success - could you maybe test how that would look on the button?

Here are two screens with both values changed.
20867-IncreasedSaturationButton
20867-SavedFlat
20867-IncreasedSaturationText

It's nice on the button and the popup 'Saved' text, but to be honest I could not see the difference on the text targeted by this PR.

michitux:
> It's nice on the button and the popup 'Saved' text, but to be honest I could not see the difference on the text targeted by this PR.

At least on my screens I see a difference, it definitely looks greener and is less easily confused with gray (but I also have screens that have somewhat good color reproduction).

Sereza7:
We can do cosmetic changes without a Jira according to [the Development practices](https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices).
So I integrate it to this PR with https://github.com/xwiki/xwiki-platform/pull/2261/commits/124fd902bb43983d49001958860a2c41d0bd38ad


PS: Thanks for sharing the article about hue, it was really interesting and it now feels like I should have followed what it describes a long time ago. It also makes understanding contrast values easier.

PR #2262, opened 2023-07-20
XWIKI-20843: Application Panel "More Applications" dropdown is not accessible

Sereza7:
**Jira:** https://jira.xwiki.org/browse/XWIKI-20843

## PR Changes
* Fixed style
* Fixed escaping of `>` characters in the xml
## Note
This is a fix PR to avoid regression on UI introduced in https://github.com/xwiki/xwiki-platform/pull/2157

## View
This view was made with accessibility features turned on, only links are underlined.
20843-2-Fix


manuelleduc (approved):

michitux (changes requested):
In the before screenshot of #2157, there is a clearly visible margin above the "more applications" link. This margin was good imho as this is something different than yet another application. Now this margin seems gone.

manuelleduc:
Before fix

image

After fix:

image

14.10.14:

image

surli:
Converted to draft: this PR needs to reapply first https://github.com/xwiki/xwiki-platform/commit/d2ef20d37833e26590c15ebed747bf8c2e2cd54b

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -551,18 +551,14 @@ require(['scriptaculous/effects'], function() {
...
   color: $theme.linkColor;
   box-shadow: none;
   background-color: transparent;
   width: 100%;
   display: flex;
-}
-
-.applicationPanelMoreButton:hover,
-.applicationPanelMoreButton:focus {
-  background-color: $theme.highlightColor;
```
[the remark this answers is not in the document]
Sereza7: Here is a demo without focus or hover styles: https://up1.xwikisas.com/#VquQmLMqMHKMMTPBfCwugw
I agree that we should at least have an indicator on hover (on focus, I guess the blue border could be enough).
I propose to only highlight the button on hover. Not on focus, so that it doesn't stay grey after being clicked. 
```
.applicationPanelMoreButton:hover {
  background-color: $theme.highlightColor;
}
```
It's a middle ground that sounds correct to me. Here is a demo with just the 'hover' change of background (and a small border radius to make it look less raw): https://up1.xwikisas.com/#QgoJqFfsWK6PEqc-Xh_5cw

What do you think of this @manuelleduc ?

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -551,18 +551,14 @@ require(['scriptaculous/effects'], function() {
...
+  text-align: left;
```
Sereza7, line 561:
```suggestion
  text-align: left;
}
.applicationPanelMoreButton:hover
  background-color: $theme.highlightColor;
```

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
```diff
@@ -551,18 +551,14 @@ require(['scriptaculous/effects'], function() {
 }
 
 .applicationPanelMoreButton {
-  padding: 10px 15px 10px 0;
+  padding: 10px 0 0 0;
   border: 0;
```
Sereza7, line 555: Related to https://github.com/xwiki/xwiki-platform/pull/2262/files#r1272322198.
This does not matter much and can be ignored or dismissed without a discussion.
```suggestion
  border: 0;
  border-radius: 0.2em;
```

Sereza7:
Answer to @surli https://github.com/xwiki/xwiki-platform/pull/2262#issuecomment-1645235069
I'm not sure I properly reapplied https://github.com/xwiki/xwiki-platform/commit/383ff2e0ee6c53d8fe515c5f217cb9cbe80119fa , I was expecting to have one more commit to show my operations, but I think all the changes got added to https://github.com/xwiki/xwiki-platform/pull/2262/commits/52a6698dae6709d6d882dd5e299309f84dab83b0 instead.

I'll wait for your answer to open this draft back to 'ready for review' if you think things are okay as it is.

surli:
> if you think things are okay as it is.

I think it's ok: we'll squash the commits and use a proper commit message to inform it applies back the changes from https://github.com/xwiki/xwiki-platform/commit/383ff2e0ee6c53d8fe515c5f217cb9cbe80119fa with other changes too.

michitux (changes requested):
From what I can see, your last commit removes the space before the more applications dropdown again. While padding might not be good here for the focus box, maybe we could use a margin?

Also, the more I think about this solution the less convinced I am by the solution to use a button but style it as a link. I think opinions diverge where [some](https://www.a11y-101.com/design/button-vs-link) think you shouldn't restyle buttons as links while [others](https://yatil.de/blog/buttons-vs-links) think this is okay. The thing is just that here we're somehow doing a lot of styling to make this button look like a link. I'm wondering if it wouldn't be better to just go back to a link and just add the aria attributes.

Sereza7:
@michitux 

> From what I can see, your last commit removes the space before the more applications dropdown again. While padding might not be good here for the focus box, maybe we could use a margin?

From what I understood, we want the entry to be aligned with the other items, and therefore no margin.
I think it looks better with everything aligned rather than the last item indented like this:
beforeFix

> Also, the more I think about this solution the less convinced I am by the solution to use a button but style it as a link.

I didn't like it much at first either but we had [a discussion on the forum](https://forum.xwiki.org/t/ui-application-panel-dropdown-toggler/12602) with Vincent and Marius and the conclusion was to use this "false link" ...

> I think opinions diverge where [some](https://www.a11y-101.com/design/button-vs-link) think you shouldn't restyle buttons as links while [others](https://yatil.de/blog/buttons-vs-links) think this is okay.

Thanks for the links. I didn't know the first one (unfortunately he seems to not post much -> https://www.nils-lauk.de/ ). I did know the second one and it's a good source for A11y opinions :+1: .

> I'm wondering if it wouldn't be better to just go back to a link and just add the aria attributes.

Semantically, I don't think it matches what a link is supposed to be, so I'd be against this (see the discussion linked above for arguments).

michitux:
> From what I understood, we want the entry to be aligned with the other items, and therefore no margin.
> I think it looks better with everything aligned rather than the last item indented like this:

This is not about indentation. This is about vertical spacing. Please compare the before screenshot to the current state and make sure there are no visual changes. Before your changes, there was a clear space that separated the "more applications" link vertically from the other items as they are also semantically very different. This space should remain and it should be a clear space before the link/button, not a space inside the button.

> Semantically, I don't think it matches what a link is supposed to be, so I'd be against this (see the discussion linked above for arguments).

With a button, you have still the problem that it looks like a link so users might still expect link behavior. Anyways, if you can fix the visuals to be mostly the same as before (maybe minus the hover behavior), I'm not against a button I'm just wondering if the time and code complexity for "fixing" (or breaking, depends on the standpoint) the visual appearance is worth the semantic improvement it brings.

Sereza7:
Here is the state after https://github.com/xwiki/xwiki-platform/pull/2262/commits/44fc553892601ae85d24dd22e813628c7255aa97:

20843-fixfix2final

Sereza7:
Answer to @michitux 's previous comment:
1. From what I can see, this issue only happens on my distrib: on the panel page when minimification is turned off => see https://up1.xwikisas.com/#EQhqkEu5lL4xRXrrvxw3-Q . This is probably an issue with some minified cache that doesn't get overwritten. Everything is fine when I deactivate the use of minified files.
2. Addressed in https://github.com/xwiki/xwiki-platform/pull/2262/commits/330694de9a49a3b5561f4885b55bd03a5d61aa32 . I removed the inline styling on line 386, but it was necessary to avoid style flickering at the first opening (for a fraction of time, the object is open, and the Blinddown animation hasn't started. I reverted this change by adding back the inline styling.
3. I think this is close enough to the actual style, and this has the advantage of highlighting the difference with the other links. Moreover, turning the whole style extension in a LESS just to get the new color theme variable `@link-hover-color` (which is just a `darken(@link-color, 15%)` with LESS, but support for the CSS `color-mix` function is not that good yet). I don't think it's worth going through all these changes and hacks. About the rounded corners, I made a proposition [earlier on this PR](https://github.com/xwiki/xwiki-platform/pull/2262#discussion_r1272332743).
4. (Small panels) Some of it came from the fact I didn't update the small-panel style to take into account the last changes, and some came from my screen resolution and not testing with multiple zoom levels. Addressed in  https://github.com/xwiki/xwiki-platform/pull/2262/commits/7efaf852bccc355e309c3831caf5871f1756fe83 See https://up1.xwikisas.com/#rhqE-YlUQrcZVvtjvIN3og for a quick video demo of the updated UI. (note that 2. wasn't fixed yet in this video)

Sereza7:
> According to the video, the icons for the two items in "More applications" are still on the left while the item text is centered. This just looks untidy. In my screenshot, you can see that before your changes the triangular icons were above the two items. I don't insist on getting exactly the same style but I think we need either: 
> 1. Icons on the left and text left-aligned with proper spacing between the icon and the text, or
> 1. Icons on the top with centered text

I prefer option 1 since the carets are pointing right and it would look weird if they pointed at nothing. I don't think down-carets would look much better either. Addressed in https://github.com/xwiki/xwiki-platform/pull/2262/commits/f1fcfe4200c51ef9b754dd82b5024faee9123eeb

> It seems super strange to me to have a hover-style that is used exclusively on this element.

I fixed it so that it would be mostly consistent with the hover styles of the links next to it. Everything but the color is changed. However, to get the proper color we would need to access it via @link-hover-color). I don't think it's worth parsing a LESS file only for a color variation only an admin will see when hovering one button. We can access the regular link color since it's an old theme variable ($theme.LinkColor), and those can be accessed from CSS, but AFAIK @link-hover-color cannot be accessed if we don't get a LESS. Let me know if you think it should be changed.

Here is a demo of the latest version:
https://up1.xwikisas.com/#bPDL_aKDgFp68ftq7m2JHg

> That rounded corners proposal is not using the same radius as the other rounded corners we have. See also the extensive discussion in https://forum.xwiki.org/t/rounder-corners-everywhere-minimalist-skin-3/12919 - I think it is important that we keep things consistent as otherwise we're adding technical debt as we'll need to change all these places when we want to use a different border radius.

It probably is not relevant anymore, closing it.

> Maybe we can re-vive the [forum discussion](https://forum.xwiki.org/t/ui-application-panel-dropdown-toggler/12602) to discuss the hover-behavior?

Going through a forum discussion takes time from everyone and slows down the process. I think that the stakes here are pretty low and don't need a forum discussion. Let me know what you think of the color difference :)

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Applications.xml
Sereza7: Irrelevant as of now.

PR #2266, opened 2023-07-21
XWIKI-20979: Add a quick action for inserting an icon

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-20979

This pull request adds a new Quick Action to quickly search for and insert an icon in the WYSIWYG editor.

preview

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/iconService.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/iconService.js
@@ -1,0 +1,67 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+define('xwiki-iconService', [
+  'jquery',
+], function ($) {
+
+  /**
+   * Generate an URL for getting JSON resources about icons.
+   */
+  const getResourceURL = function (action, parameters) {
+    const params = Object.assign({outputSyntax: "plain", action: action}, parameters);
+    return (new XWiki.Document('IconPicker', 'IconThemesCode')).getURL('get', $.param(params), true);
+  };
+
+  let cachedIconThemes = false;
+  const getIconThemes = function () {
+    return new Promise(function (resolve, reject) {
+
+      if (cachedIconThemes) {
+        return resolve(cachedIconThemes);
+      }
+
+      $.getJSON(getResourceURL('data_iconthemes'), function (data) {
+        cachedIconThemes = data;
+        resolve(data);
+      }).fail(reject);
+    });
+  };
+
+  const cachedIcons = {};
+  const getIcons = function (iconTheme) {
+    return new Promise(function (resolve, reject) {
+
+      if (cachedIcons[iconTheme]) {
+        resolve(cachedIcons[iconTheme]);
+      }
+
+      $.getJSON(getResourceURL('data_icons', {iconTheme}), function (dataIcons) {
+        cachedIcons[iconTheme] = dataIcons;
+        resolve(dataIcons);
+      }).fail(reject);
+    });
+  };
+
+  return {
+    getIconThemes,
+    getIcons,
+  };
+});
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/plugin.js
@@ -1,0 +1,79 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+(function () {
+  'use strict';
+
+  CKEDITOR.plugins.add('xwiki-icon', {
+    init: function (editor) {
+      editor.config.mentions = editor.config.mentions || [];
+      // Icon Autocomplete
+      editor.config.mentions.push({
+        marker: `icon::`,
+        pattern: new RegExp('icon::\\S{0,30}$'),
+        minChars: 0,
+        itemsLimit: 0,
+        itemTemplate: `<li data-id="{id}">
+            <div class="ckeditor-autocomplete-item-head">
+              <span class="ckeditor-autocomplete-item-icon-wrapper">` +
+                // We use aria-hidden="true" and role="presentation" for the icon display
+                // because it is redundant with the label that is shown next to it.
+                `<span class="{iconClass}" aria-hidden="true"></span>
+                <img src="{imgSrc}" role="presentation"></img>
+              </span>
+              <span class="ckeditor-autocomplete-item-label">{label}</span>
+            </div>
+          </li>`,
+        feed: function (opts, callback) {
+          require(['xwiki-iconService'], function (iconService) {
+            iconService.getIconThemes().then(function (iconThemes) {
+              // Retreive the list of available icons.
+              iconService.getIcons(iconThemes.currentIconTheme).then(function (icons) {
+                callback(icons
+                  .filter(icon => icon.name.toLowerCase().startsWith(opts.query.toLowerCase()))
+                  .map(icon => ({
+                    id: icon.name,
+                    label: icon.name,
+                    imgSrc: icon.metadata.url,
+                    iconClass: icon.metadata.cssClass
+                  }))
+                );
+              }).catch(function () {
+              editor.showNotification(editor.localization.get('xwiki-icon.iconsFetchFailed'), 'warning', 5000);
+            });
+            }).catch(function () {
+              editor.showNotification(editor.localization.get('xwiki-icon.iconThemesFetchFailed'), 'warning', 5000);
+            });
+          });
+        },
+        outputTemplate: function (item) {
+          editor.execCommand('xwiki-macro-insert', {
+            name: "displayIcon",
+            parameters: {
+              name: item.id,
+            },
+            // Displaying the macro inline allows to write text around the icon after insertion.
+            inline: "enforce",
+          });
+          return "";
+        }
+      });
+    },
+  });
+})();
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
@@ -724,6 +724,13 @@
           outputHTML: '/img::'
         }, {
           group: 'Content',
+          id: 'icon',
+          name: editor.localization.get('xwiki-slash.action.icon.name'),
+          iconClass: 'fa fa-image',
+          description: editor.localization.get('xwiki-slash.action.icon.hint'),
+          outputHTML: 'icon::'
+        }, {
+          group: 'Content',
           id: 'mention',
           name: editor.localization.get('xwiki-slash.action.mention.name'),
           iconClass: 'fa fa-user-o',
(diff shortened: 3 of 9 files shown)


vmassol:
Thans Dorian. I see you differentiate "icon" and "icon image". I think there's a problem of terminology: one is using the Font Awesome Icon set, the Silk icon set and there are even other icon sets we could use (this is extensible in XWiki). See https://design.xwiki.org/xwiki/bin/view/Proposal/XWiki%20Icon%20Set

So I think only the default icon set should be pushed forward in quick actions (it's the FA one ATM but it can be changed by the user, see https://extensions.xwiki.org/xwiki/bin/view/Extension/Icon%20Theme%20Application#HSelectingtheIconSet). 

WDYT?

Thanks

manuelleduc:
Thanks @douakli !

I agree with @vmassol, the quick action should only suggest and insert icons of the current icon set.
And in this case you can always insert the `displayIcon` macro, which will select the right way to display the icon according to the currently selected icon set (which also makes switching to another icon set quite easy).

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/iconService.js
```diff
@@ -0,0 +1,71 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+define('iconService', [
```
manuelleduc, line 21: We might want to move the icons out of the `xwiki-image` plugins as they are not strictly speaking images (they can be displayed using CSS in some cases).
WDYT of having a `xwiki-icon` plugin? (cc @mflorea)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/iconService.js
```diff
@@ -0,0 +1,71 @@
...
+
+      if (cachedIconThemes) {
+        return resolve(cachedIconThemes);
+      }
+
+      $.getJSON(getResourceURL('data_iconthemes'), function (data) {
+        cachedIconThemes = data;
+        resolve(data);
+      });
+    });
```
manuelleduc, line 49: What happens if the query fails for some reason? Does the user have a visual feedback (e.g., a notification) indicating that something went wrong?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -377,6 +377,88 @@
...
+            '<img src="{imgSrc}" aria-hidden="true"></img>' +
+            '</span>' +
+            '<span class="ckeditor-autocomplete-item-label">{label}</span>' +
+            '</div>' +
+            '</li>',
+          feed: function (opts, callback) {
+            require(['iconService'], function (iconService) {
+              iconService.getIconThemes().then(function (iconThemes) {
+
+
```
manuelleduc, line 401:
```suggestion

```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -377,6 +377,88 @@
...
+                        label: icon.name,
+                        imgSrc: icon.metadata.url,
+                        reference: icon.metadata.value,
+                        iconClass: icon.metadata.cssClass
+                      });
+                    }
+                  });
+                  callback(items);
+                });
+
```
manuelleduc, line 424:
```suggestion
```

I tend to remove those extra spaces :)

douakli:
@vmassol 
> So I think only the default icon set should be pushed forward in quick actions (it's the FA one ATM but it can be changed by the user

Currently, the icon tab in the image selection modal inserts an image reference, which only supports the Silk icon theme.
Since the JIRA issue mentioned using the suggest for it, I thought we should provide a way to insert an image reference.

I didn't know the image reference icons were not pushed anymore.
I think only pushing the displayIcon macro makes sense.
I updated the PR accordingly.

Thanks.

douakli:
Here is an updated preview of the feature:
icon-quick-action-demo

Thanks.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/iconService.js
mflorea: > WDYT of having a xwiki-icon plugin?

Makes sense.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/iconService.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const cachedIcons = {};
+  const getIcons = function (iconTheme) {
+    return new Promise(function (resolve, reject) {
+
+      if (cachedIcons[iconTheme]) {
+        resolve(cachedIcons[iconTheme]);
+      }
+
+      $.getJSON(getResourceURL('data_icons', 'iconTheme=' + iconTheme), function (dataIcons) {
```
mflorea, line 60:
```suggestion
      $.getJSON(getResourceURL('data_icons', {iconTheme}), function (dataIcons) {
```

See above.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/iconService.js
```diff
@@ -0,0 +1,71 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+define('iconService', [
```
mflorea, line 21:
```suggestion
define('xwiki-iconService', [
```

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/plugin.js
```diff
@@ -0,0 +1,80 @@
...
+ */
+(function () {
+  'use strict';
+
+  CKEDITOR.plugins.add('xwiki-icon', {
+    init: function (editor) {
+      editor.config.mentions = editor.config.mentions || [];
+      // Icon Autocomplete
+      editor.config.mentions.push({
+        marker: `/icon::`,
```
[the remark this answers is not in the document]
douakli: I removed the conflict [when introducing the image insertion quick action](https://github.com/xwiki/xwiki-platform/commit/1eed1c849aa852824e8e3f4961fcc472e4e8992d#diff-4060dab34e60de96a5790b5872408cc687019aab0dd4ebb874121c5215c7f096R350) (similar `/img::`). My idea was that it would make it part of Quick Actions and not crowd the shortcut space.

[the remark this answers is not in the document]
mflorea: Yes, but I don't see the benefit. It's still an additional shortcut, and it's complex enough without the slash prefix so that it doesn't get triggered accidentally. What if I really want to search for "img::" in quick actions? Quick actions are extensible, so we can't know what use cases there might be in the future that might require a search for "img::". I don't think it's right to prevent this search.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/plugin.js
```diff
@@ -0,0 +1,80 @@
...
+        pattern: new RegExp('/icon::\\S{0,30}$'),
+        minChars: 0,
+        itemsLimit: 0,
+        itemTemplate: '<li data-id="{id}">' +
+          '<div class="ckeditor-autocomplete-item-head">' +
+          '<span class="ckeditor-autocomplete-item-icon-wrapper">' +
+          '<span class="{iconClass}" aria-hidden="true"></span>' +
+          '<img src="{imgSrc}" aria-hidden="true"></img>' +
```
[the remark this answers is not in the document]
douakli: From my understanding, aria-hidden should be here when a screen reader would see duplicated content, in this case, we have the icon followed by its name.
@Sereza7, am I approaching this correctly, should I implement this differently?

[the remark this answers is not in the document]
Sereza7: So, first: you shouldn't use aria-hidden both on the image and the span describing it, because you don't have anything to read anymore (no problem of duplicate, but nothing is read out, that's no good). I recommend to hide the image and leave the span as is.
Let me know if I misunderstood it, I didn't dig deep to check what those are used for exactly.

___

Then: Check out the HTML code style I updated after a proposal a few days ago ;)
https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/

Seems like it's a case of "visually showed, semantically hidden" element.
I recommend to use `role='presentation'` on the image.
Using 'aria-hidden' is not wrong, but for the sake of consistency in xwiki-platform you should use the presentation role instead. They are [very similar solutions](https://timwright.org/blog/2016/11/19/difference-rolepresentation-aria-hiddentrue/), and in this case, they behave the same.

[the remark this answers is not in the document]
douakli: The span is not describing the image, it's here to hold the equivalent icon when it is rendered using a css class, and can be empty when an image rendered icon is used instead. The describing span is later in the template.

Thank you for the documentation link, I'm checking it out :)
I'll use the `role='presentation'` on the image and add a comment.

Thanks.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/plugin.js
```diff
@@ -0,0 +1,82 @@
...
+          <div class="ckeditor-autocomplete-item-head">
+          <span class="ckeditor-autocomplete-item-icon-wrapper">` +
+          // We use aria-hidden="true" and role="presentation" for the icon display
+          // because it is redundant with the label that is shown next to it.
+          `<span class="{iconClass}" aria-hidden="true"></span>
+          <img src="{imgSrc}" role="presentation"></img>
+          </span>
+          <span class="ckeditor-autocomplete-item-label">{label}</span>
+          </div>
+          </li>`,
```
mflorea, line 42: I would indent the HTML to make it easier to read.

```suggestion
        itemTemplate: `
          <li data-id="{id}">
            <div class="ckeditor-autocomplete-item-head">
              <span class="ckeditor-autocomplete-item-icon-wrapper">` +
                // We use aria-hidden="true" and role="presentation" for the icon display
                // because it is redundant with the label that is shown next to it.
                `<span class="{iconClass}" aria-hidden="true"></span>
                <img src="{imgSrc}" role="presentation"></img>
              </span>
              <span class="ckeditor-autocomplete-item-label">{label}</span>
            </div>
          </li>
        `,
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/iconService.js
```diff
@@ -0,0 +1,67 @@
...
+      $.getJSON(getResourceURL('data_icons', {iconTheme: iconTheme}), function (dataIcons) {
+        cachedIcons[iconTheme] = dataIcons;
+        resolve(dataIcons);
+      }).fail(reject);
+    });
+  };
+
+  return {
+    getIconThemes: getIconThemes,
+    getIcons: getIcons,
```
mflorea, line 65:
```suggestion
    getIconThemes,
    getIcons,
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-icon/plugin.js
```diff
@@ -0,0 +1,82 @@
...
+                  if (icon.name.toLowerCase().startsWith(opts.query.toLowerCase())) {
+                    items.push({
+                      id: icon.name,
+                      label: icon.name,
+                      imgSrc: icon.metadata.url,
+                      iconClass: icon.metadata.cssClass
+                    });
+                  }
+                });
+                callback(items);
```
mflorea, line 59:
```suggestion
                callback(icons
                  .filter(icon => icon.name.toLowerCase().startsWith(opts.query.toLowerCase()))
                  .map(icon => {
                    id: icon.name,
                    label: icon.name,
                    imgSrc: icon.metadata.url,
                    iconClass: icon.metadata.cssClass
                  })
                );
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
```diff
@@ -713,6 +713,13 @@
           iconClass: 'fa fa-image',
           description: editor.localization.get('xwiki-slash.action.img.hint'),
           outputHTML: '/img::'
+        }, {
+          group: 'Content',
+          id: 'icon',
+          name: editor.localization.get('xwiki-slash.action.icon.name'),
+          iconClass: 'fa fa-image',
```
mflorea, line 720: I'm wondering if we shouldn't use a different icon than the one used for Image quick action. There aren't any good options though. I can think only of https://fontawesome.com/v4/icon/file-image-o which is similar but different.

mflorea (commented):
Looks good. I left only some small suggestions.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
douakli: I wondered too, but I thought it wasn't appropriate for an icon. Maybe do the other way around and change the image Quick Action's icon?

mflorea: > Maybe do the other way around and change the image Quick Action's icon?

There aren't many alternatives for the Image Quick Action icon either. Let's keep it like this for now.

mflorea (approved):

PR #2268, opened 2023-07-25
XWIKI-21165: Heading 4-6 Quick Actions are missing

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-21165

This pull request adds headings 4-6 to the Quick Actions drop-down.

preview


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/Translations.xml
```diff
@@ -172,6 +172,9 @@ ckeditor.plugin.slash.group.Macro=Macro
 ckeditor.plugin.slash.action.h1.hint=Big section heading.
 ckeditor.plugin.slash.action.h2.hint=Medium section heading.
 ckeditor.plugin.slash.action.h3.hint=Small section heading.
+ckeditor.plugin.slash.action.h4.hint=Smaller section heading.
```
Sereza7, line 175: Shouldn't we update the values for the keys from h1 to h3 to avoid having all headers from h3 being described as small?
I think something like Largest > Large > Big > Medium > Small > Smallest would fit more section headings.

By the way, we should not encourage users to use headings just for their font size, but for their meaning (so that they write content better for accessibility and SEO :) ).
I would rather use the 'height' image to compare them:
Highest level of heading
Second level of heading
...
Fifth level of heading
Lowest level of heading

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements for more info about the semantics of the section headings

mflorea: @Sereza7 neither "Largest section heading" nor "Highest level of heading" sound good to me. I haven't seen anywhere section headings described as large or having height. I'd keep it simple:

* First level heading / Top level heading
* Second level heading
* Third level heading
* Forth level heading
* Fifth level heading
* Sixth level heading

@douakli WDYT?

Sereza7: Sounds good to me. 

The height was used in its non physical definition, see definition 9 of the Merriam-Webster online dictionary: https://www.merriam-webster.com/dictionary/high
I think it's a bit better than using 'large', but I think your solution is still simpler and more elegant. :)

douakli: It sounds good to me too,
Thanks for the suggestion! @Sereza7 :)

PR #2271, opened 2023-07-26
XWIKI-20988: Livetable edit sheet available column select lacks a name

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20988
## PR Changes
* Changed the `for` attribute of the label.
* Linked the \<select> element with its description

## View
***No visual change***
Before the PR vvv
20988-StateBefore
After the PR vvv
20988-stateAfter

## Notes
The label used to be linked to the input that stays hidden no matter what AFAICS in the code and in tests. A hidden input doesn't need any label: 
> Inputs with the type attribute value of hidden (e.g., type="hidden"). These inputs are hidden and unavailable for user input. They therefore need no label. 

from [Axe documentation](https://dequeuniversity.com/rules/axe/4.4/label)

I decided to go a bit further than fix the error reported on Jira and link the description to the select too since it's already there.

## Tests
An axe core check on the page screenshoted above didn't return any WCAG violation related to the changes in this PR.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
@@ -42,7 +42,7 @@
 
 {{velocity output="false"}}
 #macro (listAvailableColumns $classReference)
-  &lt;select id="availableColumns"&gt;
+  &lt;select id="availableColumns" aria-describedby='availableColumnsDescription'&gt;
     #set ($classFields = $xwiki.getDocument($classReference).getxWikiClass().properties)
     #if ($classFields.size() &gt; 0)
       &lt;optgroup label="$escapetool.xml($services.localization.render(
@@ -150,8 +150,8 @@
     &lt;/dt&gt;
     &lt;dd&gt;#displayPropertyValue('description')&lt;/dd&gt;
     &lt;dt&gt;
-      &lt;label for="AppWithinMinutes.LiveTableClass_0_columns"&gt;$doc.displayPrettyName('columns', false, false)&lt;/label&gt;
-      &lt;span class="xHint"&gt;$services.localization.render('platform.appwithinminutes.liveTableEditorColumnsHint')&lt;/span&gt;
+      &lt;label for="availableColumns"&gt;$doc.displayPrettyName('columns', false, false)&lt;/label&gt;
+      &lt;span id='availableColumnsDescription' class="xHint"&gt;$services.localization.render('platform.appwithinminutes.liveTableEditorColumnsHint')&lt;/span&gt;
     &lt;/dt&gt;
     &lt;dd&gt;#displayPropertyValue('columns')&lt;/dd&gt;
   &lt;/dl&gt;
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
@@ -42,7 +42,7 @@
 
 {{velocity output="false"}}
 #macro (listAvailableColumns $classReference)
-  &lt;select id="availableColumns" aria-describedby='availableColumnsDescription'&gt;
+  &lt;select id="addAvailableColumn" aria-describedby='addAvailableColumnHint'&gt;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
```diff
@@ -150,8 +150,8 @@
     &lt;/dt&gt;
     &lt;dd&gt;#displayPropertyValue('description')&lt;/dd&gt;
     &lt;dt&gt;
-      &lt;label for="AppWithinMinutes.LiveTableClass_0_columns"&gt;$doc.displayPrettyName('columns', false, false)&lt;/label&gt;
-      &lt;span class="xHint"&gt;$services.localization.render('platform.appwithinminutes.liveTableEditorColumnsHint')&lt;/span&gt;
+      &lt;label for="availableColumns"&gt;$doc.displayPrettyName('columns', false, false)&lt;/label&gt;
+      &lt;span id='availableColumnsDescription' class="xHint"&gt;$services.localization.render('platform.appwithinminutes.liveTableEditorColumnsHint')&lt;/span&gt;
```
michitux, line 154: As discussed on another PR, I would rather use `Hint` instead of `Description` as this is the term we use here (the translation key is also named `liveTableEditorColumnsHint`) and switching between the two just creates a mess.

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
```diff
@@ -150,8 +150,8 @@
     &lt;/dt&gt;
     &lt;dd&gt;#displayPropertyValue('description')&lt;/dd&gt;
     &lt;dt&gt;
-      &lt;label for="AppWithinMinutes.LiveTableClass_0_columns"&gt;$doc.displayPrettyName('columns', false, false)&lt;/label&gt;
-      &lt;span class="xHint"&gt;$services.localization.render('platform.appwithinminutes.liveTableEditorColumnsHint')&lt;/span&gt;
+      &lt;label for="availableColumns"&gt;$doc.displayPrettyName('columns', false, false)&lt;/label&gt;
```
michitux, line 153: There is actually an input in the generated HTML that has the id `AppWithinMinutes.LiveTableClass_0_columns` and that is a much better fit for this label. Also, speaking about progressive enhancement (in the forum), this form actually uses progressive enhancement here, without JavaScript, this input is actually visible and the right target for this label. Only with JavaScript, the select that you target now is shown. So I think the correct solution here would be to change the `for`-attribute in JavaScript once the select has been made visible.

I'm also wondering if the select shouldn't get a different label as this select is just for adding new columns. There is another widget for arranging columns - making that widget accessible is probably a completely different challenge, though, as it cannot even be focused at the moment.

michitux (changes requested):

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2271/commits/81b4383343e9ab7846aca91b62ef38f62dd7f662 :+1:

xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableEditSheet.xml
Sereza7: > There is actually an input in the generated HTML that has the id AppWithinMinutes.LiveTableClass_0_columns and that is a much better fit for this label. Also, speaking about progressive enhancement (in the forum), this form actually uses progressive enhancement here, without JavaScript, this input is actually visible and the right target for this label. Only with JavaScript, the select that you target now is shown. So I think the correct solution here would be to change the for-attribute in JavaScript once the select has been made visible.

Addressed in https://github.com/xwiki/xwiki-platform/pull/2271/commits/377ca41ceb8269a9db8a857fbc03f1868024e4b5

> I'm also wondering if the select shouldn't get a different label as this select is just for adding new columns. There is another widget for arranging columns - making that widget accessible is probably a completely different challenge, though, as it cannot even be focused at the moment.

I think it's okay, because this select is actually just a select and the action link to validate the selection is properly labelled.
Labelled with "Add selected column".
20988-labelIsEnough

PR #2275, opened 2023-07-27
XWIKI-20995: AWM livetable entry sheet view uses silk icons whatever the icon theme

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20995
## PR Changes
* Replaced the background image LESS with displayIcon macros in the link

## View
20995-AfterPR
20955-WithSilkTheme

## Note
* I needed to change the values of the icons for it to fit into the standard set of icons supported in XWiki. I think all icons fit their purpose both on the fontawesome and silk iconthemes.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableViewSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/LiveTableViewSheet.xml
@@ -89,7 +89,7 @@
     (((
       = $services.localization.render('platform.appwithinminutes.appHomePageActionsHeading') =
       #if ($hasCreateData)
-        * [[$services.localization.render('platform.appwithinminutes.appHomePageAddEntryHint')&gt;&gt;||anchor="AddNewEntry" class="action add"]]##
+        * [[{{displayIcon name="add"/}} $services.localization.render('platform.appwithinminutes.appHomePageAddEntryHint')&gt;&gt;||anchor="AddNewEntry" class="action add"]]##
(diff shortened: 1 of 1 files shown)


michitux (approved):
This seems to reference an attachment (`locate.png`), would it make sense to remove that or is this attachment used elsewhere?

This looks good to me, but maybe it would make sense to get some more feedback first on this UI change.

Sereza7:
From what I could see ([github search](https://github.com/search?q=repo%3Axwiki%2Fxwiki-platform%20locate.png&type=code)), this attachment was not used anywhere else.
Fixed in https://github.com/xwiki/xwiki-platform/pull/2275/commits/84cd064fe5f38aca9753e9ab9c9c2a0aa049ecef :)

tkrieck:
The chosen icons are good approximations of the original Silk icons, so not a problem with this IMO. Some of them are hard to decipher because of the size (translate and edit application, for example). Increasing the size could help, but carefully to not waste space.

Now on to a more conceptual view:

Different icons for the same base action is not ideal (delete, for example). And to find 100% matching icons for each of the options is very difficult.  Icons should always reinforce the base meaning of labels, but they also need to be recognized by users quickly, usually this is done using a picture that is common in other places.

If I were to redesign this section, I would group the actions between application and entries and only use icons when we can absolutely align it conceptually with the label. The other icons I would drop altogether. Showing this section as a horizontal toolbar would also help to save space.

I will make a proposal to discuss this better. Probably I am missing some implementation details and restrictions.

michitux:
@tkrieck Not all users see all actions (depends on rights) and the text that is next to it is chosen by the user who creates the application, so this could be a longer text. I think the "Add new entry" action is the one action that should be easily available and that is frequently used while other actions should be rarely used and could thus also, e.g., be hidden in some menu.

tkrieck:
@michitux That's nice to know, I will keep in mind on the proposal. The user can also choose the icons for these actions?

PR #2282, opened 2023-08-08
XWIKI-18007: Drawer menu improvements for accessibility

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-18007
## Changes
* removed the dependency on the 'drawer' module
* Added proper keyboard navigation to the drawer
* Kept style intact despite those changes

## View
Video demo: https://up1.xwikisas.com/#6ngSajsdEGjTa2mmq83UiA

## Note
This implementation is based upon the discussion held on a previous PR for this ticket: https://github.com/xwiki/xwiki-platform/pull/2225


Sereza7:
TODO: start a topic about the changes proposed in this PR

Sereza7:
Topic started: https://forum.xwiki.org/t/proposal-remove-the-drawer-dependency/12972

TODO: implement an easy way to create secondary drawers.

Sereza7:
## PR Changes
* Modified the new implementation of the drawer to fit a use case with multiple drawers.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,88 @@ require(['jquery', 'bootstrap'], function($) {
...
+   ** a unique ID
+   ** the class 'drawer-nav' (mostly for style) 
+   ** the class 'closed' if the drawer is supposed to start in a closed state.
+   ** a label 
+   ** The content to display inside. It's asserted that at least one element in this content can receive focus.
+  For an example of drawer creation, see #tmDrawerActivator and #tmDrawer.
+ */
+require(['jquery'], function($) {
+  // The overlay is the same whatever the drawer opened.
+  let drawerOverlay = $(".drawer-overlay");
```
manuelleduc, line 65: Maybe just me but I prefer to explicitly state constant "variables" as constants.
```suggestion
  const drawerOverlay = $(".drawer-overlay");
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,88 @@ require(['jquery', 'bootstrap'], function($) {
...
+    // which prevents us from using them in automated tests. We need something more reliable so we listen to
+    // 'drawer1.opened' and 'drawer1.closed' events and add our own markers.
+    let openDrawer = () => drawerContainer.trigger('drawer'+index+'.opened');
+    let closeDrawer = () => drawerContainer.trigger('drawer'+index+'.closed');
+    drawerContainerToggler.on('click', openDrawer);
+    drawerOverlay.on('click', closeDrawer);
+
+    drawerContainer.on('drawer'+index+'.opened', function(event) {
+      drawerContainerToggler
+        .attr('aria-expanded', 'true');
```
manuelleduc, line 84:
```suggestion
      drawerContainerToggler.attr('aria-expanded', 'true');
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,88 @@ require(['jquery', 'bootstrap'], function($) {
...
+      // We need to set a timeout for the class update to finish properly before trying to focus an element that would
+      // have no visibility before the class change. We use an interval so that the focus is moved no matter the
+      // performance of the client.
+      let focusInterval = setInterval( ()=>{
+        if(drawerContainer.hasClass('opened') && focusableElements.length !== 0) {
+          focusableElements.first().trigger('focus');
+          clearInterval(focusInterval);
+          focusInterval = null;
+        }
+      },50);
```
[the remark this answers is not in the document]
manuelleduc: Yes, very poor choice of example... Thanks for the update :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+  For an example of drawer creation, see #tmDrawerActivator and #tmDrawer.
+ */
+require(['jquery'], function($) {
+  // The overlay is the same whatever the drawer opened.
+  const drawerOverlay = $(".drawer-overlay");
+  $('.drawer-toggle').each(function(index) {
+    // Setting up the drawer.
+    let drawerContainerToggler = $(this);
+    let drawerId = drawerContainerToggler.attr('aria-controls');
+    let drawerContainer = $('#'+drawerId);
```
michitux, line 70: As this is supposed a re-usable component, I think this should properly escape the id for the use in a selector. Alternatively, it could also use the native plain JavaScript method to get an element by id which then wouldn't require escaping. Also missing, spaces around `+`.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+
+    drawerContainer.on('drawer'+index+'.opened', function(event) {
+      drawerContainerToggler.attr('aria-expanded', 'true');
+      // We update the state of the drawer (using setAttribute since it's faster)
+      drawerContainer.get(0)
+        .setAttribute('class', 'drawer-nav opened');
+      // We need to set a timeout for the class update to finish properly before trying to focus an element that would
+      // have no visibility before the class change. We use an interval so that the focus is moved no matter the
+      // performance of the client.
+      let focusInterval = setInterval(()=>{
```
michitux, line 90:
```suggestion
      let focusInterval = setInterval(() => {
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+        .removeClass('opened')
+        .addClass('closed');
+      drawerContainerToggler.attr('aria-expanded', 'false');
+      // We remove the listeners that were created when the drawer opened up
+      $("body").off('keydown');
+      focusableElements.off('focusout');
+    });
+    
+    // When the drawer is closed, collapse sub items
+    $(body).on('drawer'+index+'.closed', function() {
```
michitux, line 121:
```suggestion
    $(body).on('drawer' + index + '.closed', function() {
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
```diff
@@ -20,87 +20,133 @@
...
     }
 
-    // Avatar styling
-    img {
-      border-radius: 50%;
-      float: left;
-      height: 50px;
-      width: 50px;
+    & .drawer-menu-item:hover {
+      color: #555;
```
michitux, line 78: Same here, why is this color hardcoded?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+require(['jquery'], function($) {
+  // The overlay is the same whatever the drawer opened.
+  const drawerOverlay = $(".drawer-overlay");
+  $('.drawer-toggle').each(function(index) {
+    // Setting up the drawer.
+    let drawerContainerToggler = $(this);
+    let drawerId = drawerContainerToggler.attr('aria-controls');
+    let drawerContainer = $('#'+drawerId);
+    let focusableElements = drawerContainer.find('button, a, input:not([type="hidden"]), ' +
+     'select, textarea, [tabindex]:not([tabindex="-1"])');
```
michitux, line 72: Shouldn't you check here that these elements are actually visible and that they aren't disabled in some way?

michitux (changes requested):
Overall, this looks good but in the details I have some concrete feedback and some more general questions.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
```diff
@@ -20,87 +20,133 @@
...
-    padding: 20px 16px;
+  & + .drawer-overlay {
+    position: fixed;
+    display: none;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    opacity: 0.2;
+    background-color: #fff - @xwiki-page-content-bg;
```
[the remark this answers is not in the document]
Sereza7: Usually I remove it / avoid it, but here I think it makes sense because:
* This is an overlay which is supposed to be transparent (on the opposite to most elements of XWiki that don't really 'hide' anything and therefore have no business prefering transparency over full colors)
* The content behind it is not clickable

The contrast is broken on the other elements, but I think it's an okay case where it's all inactive. Changing this would lead to difficult questions and I'm not sure there's a solution that is marginally better.

Making this overlay a full color is not a proper solution.
And AFAIK the only other solution here is to remove this overlay, but I think it's useful for understandability to have it when the drawer opens.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+      }, 50);
+      // The drawer can be closed by pressing the ESC key.
+      $("body").on('keydown', function (event) {
+        if (event.key === 'Escape') {
+          closeDrawer();
+        }
+      });
+      // The drawer can be closed by setting focus outside of it
+      focusableElements.on('focusout', function (event) {
+        if (event.relatedTarget != null && event.relatedTarget.closest('#'+drawerId) == null) {
```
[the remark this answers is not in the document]
Sereza7: Thank you! I didn't know about this JQuery method :eyes:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+        }
+      });
+    }).on('drawer'+index+'.closed', function(event) {
+      // We update the state of the drawer
+      drawerContainer
+        .removeClass('opened')
+        .addClass('closed');
+      drawerContainerToggler.attr('aria-expanded', 'false');
+      // We remove the listeners that were created when the drawer opened up
+      $("body").off('keydown');
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/ed5c78ff4e6a7da860958e14e8194b94a9328eae :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,87 @@ require(['jquery', 'bootstrap'], function($) {
...
+      // The drawer can be closed by pressing the ESC key.
+      $("body").on('keydown', function (event) {
+        if (event.key === 'Escape') {
+          closeDrawer();
+        }
+      });
+      // The drawer can be closed by setting focus outside of it
+      focusableElements.on('focusout', function (event) {
+        if (event.relatedTarget != null && event.relatedTarget.closest('#'+drawerId) == null) {
+          closeDrawer();
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/0690f88dc914739d5c4dbb7c6c47adb8468a03bb :+1: 
There were some complications with the style and focus, but it's working well now:
[Video demo](https://up1.xwikisas.com/#ZG-hyBopdIpFs6OKAaMmxw)

I had to use 'margin-right' intead of 'right' for positioning.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
Sereza7: :+1: Fixed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/3bea63ccc02cc7327faf52f4513ecb50dfcd1321

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
```diff
@@ -20,87 +20,133 @@
...
+    margin: 0;
+    padding: 0;
+    list-style: none;
+
+    & .drawer-menu-item {
+      font-size: 1rem;
+      display: block;
+      padding: .75rem;
+      text-decoration: none;
+      color: #222;
```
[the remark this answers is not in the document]
Sereza7: :+1: Fixed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/3bea63ccc02cc7327faf52f4513ecb50dfcd1321

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: :+1: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/1e58216ef0c9138855bfb832ef2a9f5ee482a767

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: AFAIK there's no easy way to do this. We'd need to access the value of the css property 'display'. We could rule out all elements that have the 'hidden' HTML attribute on, but this makes the request more complex and AFAIK all elements shown in the drawer are enabled.

Sereza7:
Answered all comments. Ready for review.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,90 @@ require(['jquery', 'bootstrap'], function($) {
...
+    // 'drawer1.opened' and 'drawer1.closed' events and add our own markers.
+    let openDrawer = () => drawerContainer.trigger('drawer' + index + '.opened');
+    let closeDrawer = () => drawerContainer.trigger('drawer' + index + '.closed');
+    drawerContainerToggler.on('click', openDrawer);
+    drawerOverlay.on('click', closeDrawer);
+
+    drawerContainer.on('drawer' + index + '.opened', function(event) {
+      drawerContainerToggler.attr('aria-expanded', 'true');
+      // We update the state of the drawer (using setAttribute since it's faster)
+      drawerContainer.get(0).setAttribute('open','open');
```
michitux, line 85: Shouldn't you use the [`.showModal()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal) method to trigger native focus handling?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,90 @@ require(['jquery', 'bootstrap'], function($) {
...
+      // The drawer can be closed by pressing the ESC key.
+      $("body").on('keydown.drawerClose', function (event) {
+        if (event.key === 'Escape') {
+          closeDrawer();
+        }
+      });
+      // The focus is set back to the start by setting focus outside of it
+      focusableElements.on('blur.drawerClose', function (event) {
+        if (event.relatedTarget != null && event.relatedTarget.closest('#' + $.escapeSelector(drawerId)) == null) {
+          focusDrawer();
```
michitux, line 105: Why is this necessary? The dialog element should have native focus handling.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,90 @@ require(['jquery', 'bootstrap'], function($) {
...
+        }
+      });
+      drawerContainer.on('blur.drawerClose', function (event) {
+        if (event.relatedTarget != null && event.relatedTarget.closest('#' + $.escapeSelector(drawerId)) == null) {
+          focusLastItem();
```
michitux, line 110: How is it possible to reach this code? Shouldn't the dialog element make it impossible for the focus to leave it?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/drawer.vm
```diff
@@ -99,7 +99,8 @@ aria-label="$escapetool.xml($services.localization.render('core.menu.drawer.labe
       $globalPart
     #end
   </ul>
-</nav>
+</dialog>
+<span role='presentation' class='drawer-overlay'></span>
```
michitux, line 103: A dialog element has a native backdrop, so there should be no need anymore for an overlay. See [this discussion](https://stackoverflow.com/questions/25864259/how-to-close-the-new-html-dialog-tag-by-clicking-on-its-backdrop) how to handle click events on it.

michitux (changes requested):
I have the impression you're not actually using the native behavior of `dialog` element as you're not actually opening the dialog. I would suggest changing that, so we get the benefits of the native dialog focus support. It might be a good idea to add an explicit close button to the drawer, though.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/8f6b02b56dcd0bde6c9f0fe4a568a34e4c976feb :)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: Without using .showModal, it did not handle it natively ... 
Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/8f6b02b56dcd0bde6c9f0fe4a568a34e4c976feb

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/8f6b02b56dcd0bde6c9f0fe4a568a34e4c976feb
See https://github.com/xwiki/xwiki-platform/pull/2282#discussion_r1371689643

Sereza7:
## PR Changes
* Used native dialog handling instead of custom dialog behaviour
* Fixed transitions that were not working with the dialog element
* Added a close drawer button
## Note
I think the .js is cleaner now. Unfortunately controlling the animations on the dialog element is a pain, so I added a CSS class to handle those transitions.
## View
Check out [this video for a demo of the updated changes](https://up1.xwikisas.com/#bZp0M51jok-TnAJ39Sh5LQ).
18007-afterPRnewButtonCloseDrawer

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,78 @@ require(['jquery', 'bootstrap'], function($) {
...
+require(['jquery'], function($) {
+  $('.drawer-toggle').each(function(index) {
+    // Setting up the drawer.
+    let drawerContainerToggler = $(this);
+    let drawerId = drawerContainerToggler.attr('aria-controls');
+    let drawerContainer = $(document.getElementById(drawerId));
+    
+    let openDrawer = () => drawerContainer.trigger('drawer' + index + '.opened');
+    let closeDrawer = () => drawerContainer.trigger('drawer' + index + '.closed');
+    drawerContainerToggler.on('click', openDrawer);
```
michitux, line 72: It confuses me when reading this code that a "toggler" just opens the drawer and I looked for the place in the code where this listener is replaced to close the drawer once it has been opened. It then occurred to me that this button cannot be clicked when the drawer is open, and thus it is really just an opener. Maybe this could be made more explicit?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,78 @@ require(['jquery', 'bootstrap'], function($) {
...
+    // Close drawer when clicking the backdrop (or any element outside of the drawer itself).
+    drawerContainer.on('click', (event) => {
+      let drawerzone = event.target.getBoundingClientRect();
+      if (drawerzone.left > event.clientX || drawerzone.right < event.clientX
+        || drawerzone.top > event.clientY || drawerzone.bottom < event.clientY) {
```
michitux, line 77: The [Stack Overflow discussion](https://stackoverflow.com/questions/25864259/how-to-close-the-new-html-dialog-tag-by-clicking-on-its-backdrop) around this solution mentions several problems, in particular that click events from form submissions might bubble up and might not have correct coordinates when they originate from pressing "enter". It is mentioned that as a workaround, it should be checked if the event target is actually the dialog. Is there a reason why you didn't add that check?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,78 @@ require(['jquery', 'bootstrap'], function($) {
...
+      function waitAnimation() {
+        drawerContainer.get(0).close();
+        drawerContainer.removeClass('drawer-transitioning');
+        drawerContainer.get(0).removeEventListener('transitionend', waitAnimation);
+      }
+      // We use the drawer-transitioning class to give time for the hide modal transition to play.
+      drawerContainer.addClass('drawer-transitioning');
+      drawerContainer.get(0).addEventListener('transitionend', waitAnimation);
+      // We remove the listener that was created when the drawer opened up
+      $("body").off('keydown.drawerClose');
```
michitux, line 108: I wonder how that works with multiple drawers if one drawer is triggered from another drawer. Afaik we plan this use case for What's New.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
```diff
@@ -20,87 +20,127 @@
...
 
-      .brand-user {
-        color: @xwiki-drawer-brand-user-color;
-        font-weight: 400;
+      .brand-links {
+        flex-grow: 1;
+        margin-left: 16px;
+
+        a {
+          font-size: 14px;
```
michitux, line 109: I see that a fixed font size was used before, but shouldn't this actually be relative to the base font size to also respect, e.g., accessibility settings for larger fonts?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,78 @@ require(['jquery', 'bootstrap'], function($) {
...
+  * A drawer container, that will have:
+   ** a unique ID
+   ** the class 'drawer-nav' (for style) 
+   ** an accessible name 
+   ** The content to display inside.
+  In addition, the drawer can contain a `.drawer-close` button that will close the drawer on click.
+  For an example of drawer creation, see #tmDrawerActivator and #tmDrawer.
+ */
+require(['jquery'], function($) {
+  $('.drawer-toggle').each(function(index) {
```
michitux, line 64: Shouldn't this whole code be wrapped inside `$(function() {...});`?

michitux (changes requested):
The focus handling looks good now, but I have some comments regarding some details after doing a closer review of the code.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: It's an habit I got from working on dropdowns on menus. You're right, it doesn't make the most sense here.
Fixed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/e5d890aac88a76a839ef974d01309931c972a9e5 :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
Sereza7: :+1:  Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/9064120d8d06bf71d7f2921d3707cbd74399af23

Sereza7: Note that accessibility settings for larger fonts as of now are in a CSS file, so using `@font-size-base` wouldn't fix it.

I think this is a change I'll make on the new accessibility settings

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: I added the index to the event name, so we're sure there's no collisions between drawers :)
Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/911f5323562fce20e818135e8cbd4c1500e04629

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/9f73fb6eb79cb5bee45ffba33a898008ebb15028 :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: So far there's only links in the drawer. But if we want to generalize their use we need to be ready for it to contain a form.

I addressed this issue in https://github.com/xwiki/xwiki-platform/pull/2282/commits/b117cb9d7baa632bfcfd396ec6a13385b2ba38cd :+1:

Sereza7:
Addressed all comments so far :)

Sereza7:
[Here is a demonstration of working nested drawers](https://up1.xwikisas.com/#22qlH4j4OrkBEFFvb9Puqw)

In order to get this demo, I just:
* Added a drawer-opener button in the first drawer, with  aria-controls="extraDrawer"
* Added an `#extraDrawer` to the drawer.vm template (with content more or less copy pasted from the `#tmDrawer`)
* Updated one style selector in drawer.less from `#tmDrawer.drawer-nav` to `#tmDrawer.drawer-nav, #extraDrawer.drawer-nav`

Everything works as expected. Except for two aspects which could be considered as features:
* pressing down the `Escape` key closes all drawers, no matter the depth of drawers we're in.
* the drawer backdrops get on top of each other, which makes it darker and darker when navigating down multiple drawers.

michitux:
> pressing down the `Escape` key closes all drawers, no matter the depth of drawers we're in.

I don't think that's a feature, I think the `Escape` key should only close one drawer as this probably also what the close button does.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
```diff
@@ -40,14 +40,82 @@ require(['jquery', 'bootstrap'], function($) {
...
+        // We use the drawer-transitioning class to give time for the hide modal transition to play.
+        drawerContainer.addClass('drawer-transitioning');
+        drawerContainer.get(0).addEventListener('transitionend', waitTransition);
+        // We remove the listener that was created when the drawer opened up
+        $("body").off('keydown.drawer' + index + 'Close');
+      });
+
+      // When the drawer is closed, collapse sub items
+      $(body).on('drawer' + index + '.closed', function () {
+        $('.drawer-menu-sub-item').removeClass('in').attr('aria-expanded', 'false');
```
michitux, line 116: This seems wrong in the context of multiple drawers as this would remove the `in` class in all drawers whenever any drawer is closed. Wouldn't it be better to move that into the previous event handler and to only collapse sub items of that drawer?

michitux (approved):
Overall, this looks good now, but I have a note regarding the collapsing of sub-items when there are multiple drawers.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/dead0799794baaa46fe4ae9460c55e923d745b61 :+1: 
Leaving this listener on 'body' was important when the focus could leave the drawer, but as of now, it's trapped inside the container, so it's okay to have this listener on the container directly. Changing this scope allows to close items in one drawer only.

This solution does not work properly when drawer containers are nested. I think this is an okay tradeoff, AFAIK they shouldn't be. I still added a couple lines in the doc to indicate that :) https://github.com/xwiki/xwiki-platform/pull/2282/commits/2494a8e9f67ec1ea959e069b59e7a24f1026c18f

Sereza7:
> I don't think that's a feature, I think the Escape key should only close one drawer as this probably also what the close button does.

Addressed in https://github.com/xwiki/xwiki-platform/pull/2282/commits/dead0799794baaa46fe4ae9460c55e923d745b61 :+1: 

Same as the solution for subitems above, if the drawer containers are nested this does not work as intended since the event bubbles up and causes multiple drawers to close.

__
For both, even if it does not work as intended when drawer containers are nested, it doesn't break anything.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/flamingo.js
michitux: I'm not so sure about the nesting, but when they are nested, I would expect that outer drawers are always open when inner drawers are opened so in this case, there really shouldn't be any problem because of this.

michitux:
> Same as the solution for subitems above, if the drawer containers are nested this does not work as intended since the event bubbles up and causes multiple drawers to close.

Couldn't you prevent the bubbling up? Nevertheless, this is a detail that could be fixed later when we know better how/if containers are nested.

PR #2297, opened 2023-08-16
XWIKI-21216: Lack of contrast on hover of lightbox buttons when expanded

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21216
## PR Changes
* Replaced hard coded colors with panel values from the color theme.
## View
Hereafter are four screenshots of the involved changes. They come in pairs, the first being the state before the changes proposed here, and the second being the state after the changes proposed here. On all of those screenshot, the :hover style has been applied on the leftmost button. The green highlight is created by an extension.

(a) Modal with the Iceberg color theme, before
21216-icebergBefore
(b) Modal with the Iceberg color theme, after
21216-IcebergAfterPR

(c) Modal with the Darkly color theme, before
21216-DarklyBefore
(d) Modal with the Darkly color theme, after
21216-DarklyAfterPR


xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/less/lightbox.less
```diff
@@ -52,12 +54,13 @@
...
   .action:hover {
-    color: @text-color;
+    color: @panel-default-text;
     opacity: 1;
   }
 
   .action {
-    color: @xwiki-page-content-bg;
+    color: @panel-default-text;
+    opacity: 0.8;
```
michitux, line 63: I thought it was a bad idea to use opacity with colors?

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/less/lightbox.less
```diff
@@ -25,8 +26,9 @@
   }
 
   .lightboxDescription {
-    background-color: rgba(0, 0, 0, 0.6);
-    color: #FFFFFF;
+    background-color: @panel-bg;
+    opacity: 0.6;
```
michitux, line 30: Isn't the opacity dangerous in terms of contrast?

michitux (commented):
From a technical point of view I agree. However, making the background white/gray instead of black is quite a change and not typical for this kind of lightbox, so I think we should have a forum proposal for this.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/less/lightbox.less
Sereza7: It was already there in the previously hard coded color. I think this is too much of a change to remove it and there is a genuine point to opacity here (we want to still see content behind).

I insisted to remove it on some other PRs, because transparency was not necessary but just a mean to change the color, and we had easy alternative with full semantically appropriate colors from the theme.

xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/less/lightbox.less
Sereza7: I chose to use an opacity of 0.8 here since we don't have a variation of `@text-muted` that would match the color in `@panel-default-text` (something like `@panel-muted-text`). See [this ticket](https://jira.xwiki.org/browse/XWIKI-21105) and [this topic on the forum](https://forum.xwiki.org/t/color-theme-panel-content/12818) for a discussion about the state of panel colours in the colorthemes.

Opacity can reduce the importance of a text on both light and dark theme, which is not true for the LESS color operation functions lighten/darken.

This is not a perfect solution, but I think this is the one with the least drawbacks.

Sereza7:
Started a forum topic about the solution proposed here

https://forum.xwiki.org/t/lightbox-modal-background/13114

Sereza7:
Updated the lightbox UI to avoid breaking contrast with some color themes. Now all colors are hard coded.
Here are the new proposed looks for the lightbox, with the Iceberg color theme and the Darkly color theme:
21216-afterPRIceberg2
21216-afterPRDarkly2

PR #2298, opened 2023-08-16
XWIKI-21062: Pickers are not displayed properly on some Quick Actions on Windows

douakli:
JIRA: https://jira.xwiki.org/browse/XWIKI-21062
This PR fixes a bug where clicking (rather than pressing enter) on a drop-down triggering Quick Action would
only show the triggering text without opening the new drop-down.

Preview of the fix:
quick-click

Thanks.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
@@ -256,6 +256,15 @@
       return withStrictHTMLEncoding(() => {
         return originalAutoCompletePrototype.getHtmlToInsert.apply(this, args);
       });
+    },
+    getTextWatcher: function(...args) {
+      // When pressing enter to select a shortcut Quick Action, the shortcut's textWatcher catches the event
+      // and opens the new drop-down. It however doesn't catch click events by default, which is an other
+      // way to select a Quick Action. This mitigation ensures the textWatchers are updated after
+      // a shortcut Quick Action's html is inserted.
+      const textWatcher = originalAutoCompletePrototype.getTextWatcher.apply(this, args);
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-slash/plugin.js
```diff
@@ -256,6 +256,12 @@
       return withStrictHTMLEncoding(() => {
         return originalAutoCompletePrototype.getHtmlToInsert.apply(this, args);
       });
+    },
+    getTextWatcher: function(textTestCallback) {
+      // Update the drop-down on click for shortcut triggering Quick Actions, see XWIKI-21062.
+      const textWatcher = new CKEDITOR.plugins.textWatcher(this.editor, textTestCallback, this.throttle);
+      this.editor.on("afterInsertHtml", function () {textWatcher.check(false);});
```
mflorea, line 263: The comment above is not very clear in explaining **why** we need to listen to ``afterInsertHtml``. AFAIU (based on https://github.com/ckeditor/ckeditor4/blob/4a1fb11f44b5b34f4834472d5717d03bfb2f9915/plugins/textwatcher/plugin.js#L195 ) the reason is that currently the text watcher listens mostly to keyboard events and command execution. When we use the Enter key to trigger a quick action the text watcher catches that keyboard event and if the quick action inserts some text then it reacts to that text (e.g. identifies that the inserted text matches the link shortcut). The same happens if the quick action triggers a command: even if we use the mouse click to trigger the quick action, the fact that the action triggers a command it enough to trigger the text watcher. The problem is when the quick action doesn't use a command and is activated with a click. For this we need to trigger the text watcher ourselves, after the HTML is inserted.

mflorea (changes requested):
I think the proposed solution is good, but I'd improve the comment and the fix (to wrap the original method call).

mflorea:
I'm also wondering if it would be possible to write a functional test for this use case (i.e. use click instead of Enter key to select a quick action).

mflorea (approved):

PR #2303, opened 2023-08-18
XWIKI-21182: The Image Quick Action doesn't use a relative reference for attachments from the current page

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-21182

This PR fixes a bug where the image insertion Quick Action always inserted absolute references.
Now the references are relative.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
@@ -266,8 +266,8 @@
             return "img::" + item.query;
           }
 
-          if (item.id === "_uploadImage") {
-            require(['jquery'], function ($) {
+          require(['jquery', 'resource'], function ($, resource) {
+            if (item.id === "_uploadImage") {
 
               // Reuse attachment suggest code to show the file picker.
               // Provides xwiki-attachments-store and xwiki-file-picker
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -326,19 +326,17 @@
                       // Clear the cache so the new image can appear on next usage of the image quick action
                       attachmentService.clearCache();
 
+                      const resourceReference = {...resource.convertEntityReferenceToResourceReference(
+                        XWiki.Model.resolve(attachment.id, XWiki.EntityType.ATTACHMENT)),
```
mflorea, line 330: I'm wondering if we shouldn't use here the ``editor.config.sourceDocument.documentReference`` as base reference, instead of the default current document, because the edited (source) document is not always the current document.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-image/plugin.js
```diff
@@ -351,26 +349,22 @@
...
                 });
+              return;
+            }
 
-            });
-
-            return "";
-          }
+            const resourceReference = {...resource.convertEntityReferenceToResourceReference(
+              XWiki.Model.resolve(item.reference, XWiki.EntityType.ATTACHMENT)),
```
mflorea, line 356: Same as above.

mflorea (approved):

PR #2307, opened 2023-08-21
XWIKI-21212: Cannot export a page to RTF anymore

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21212

**TODO**
- [ ] Add a RTF functional test
- [ ] check if the `TEXT` `storeProperties` section should be kept and if the label of the `FilterName` of `WEB` is ok


surli:
LGTM. 

I'm trying to understand why the regression is happening, AFAICS we used to only have the TEXT property before the upgrade so it's not related to my changes on the file format JSON file. I'm wondering if it couldn't be related to the actual support adding of WEB format in jodconverter as it matches the version: https://github.com/jodconverter/jodconverter/commit/71c4a8bd3f6d72381c50d4dfbef54ab03d5972ca. Now when looking at the commit I don't see that rtf should use WEB (see: https://github.com/jodconverter/jodconverter/commit/71c4a8bd3f6d72381c50d4dfbef54ab03d5972ca#diff-780a5d22c09d10d08e20ad495315a0b011411a83868fd9bf8aee535b64c1e175R138) so it's a bit weird.  Would need to check with using only jodconverter if we reproduce.

manuelleduc:
I also found https://github.com/jodconverter/jodconverter/compare/v4.4.2...v4.4.3#diff-00fdd04a365e5ec430627fa9b9634ce87c981abebd04ea2bb469c894ed491a12R251 where they now return "WEB" first all the time. And I can't see how to fallback except by adding some config in the json (as I did).

PR #2310, opened 2023-08-25
XWIKI-20947: Mention dropdown is not keyboard accessible 

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20947
## PR Changes
* Added an accessible name to the listbox

## Note
See the comments on the jira ticket to understand how this is the only fix needed for this ticket

## View
No changes in view.


michitux (commented):
Why is this fix for the mention dropdown in the slash plugin? Does this mean that the fix also applies to all other plugins? If yes, will it stop working when the slash plugin is disabled? Also, to fix the mention dropdown, didn't you mention in the issue that the label should say that the list lists users?

Sereza7:
> Why is this fix for the mention dropdown in the slash plugin? Does this mean that the fix also applies to all other plugins? If yes, will it stop working when the slash plugin is disabled?

Sorry, I wasn't explicit enough about the solution.
The mention dropdown relies on the slash plugin for its UI, like a couple other dropdowns (e.g. quick actions).

> Also, to fix the mention dropdown, didn't you mention in the issue that the label should say that the list lists users?

It was difficult for me to understand everything that happened in the slash plugin, and I figured that one label ''Select the item to autocomplete." was enough for every dropdown. This is more vague than a label for each dropdown, but I do not know how to extend the slash module with an extra parameter 'hint' on the dropdown and had a hard time figuring things out.

PR #2311, opened 2023-08-25
XWIKI-20931: Solr 'media type' search facet has unaccessible dropdowns

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20931
## PR Changes
* Added two levels of buttons for dropdowns in the facet select menu
* Slight changes to UI
* Added aria attributes to the dropdowns

## View
Video demo at https://up1.xwikisas.com/#FolqI1Mfg24isbfSXYdWXg

## Tests
Successfully passed 
* `mvn clean install -f xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr -Pdocker,quality -Dxwiki.test.ui.wcag=true -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false`


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrMediaTypeFacet.xml
```diff
@@ -290,10 +290,18 @@
...
-    !selected &amp;&amp; category.addClass('closed');
+    if (selected) {
+      category.addClass('expanded');
+      subtypes.attr('aria-expanded', 'true');
+    }
     subtypes.toggle(selected);
-    category.find('.itemCount').on('click', function(event) {
-      category.toggleClass('closed');
+    var togglerButton = category.find('.facet-value-toggler');
+    var dropdownID = category.attr('id')+'-dropdown';
```
michitux, line 299:
```suggestion
    var dropdownID = category.attr('id') + '-dropdown';
```

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
```diff
@@ -796,8 +759,31 @@ a.search-result-highlightAll:after {
...
+button.facet-value-toggler:not(:active) {
+  color: $theme.btn-primary-color;
+}
+
+.search-facet .search-facet-header .facet-toggler:active, button.facet-value-toggler:active {
+  box-shadow: unset;
+}
+
+.search-facet .search-facet-header .facet-toggler span, button.facet-value-toggler span {
+  transform: rotate(90deg);
```
michitux, line 776: Have you tried this with the silk icon theme? Not every icon theme uses a span as container for the icon.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrMediaTypeFacet.xml
```diff
@@ -151,15 +151,15 @@
               $facetRequestValuesBackup.size())))
             #set ($discard = $facetRequestValues.removeAll($category.mediaTypes))
           #end
-          &lt;div class="mediaTypeCategory"&gt;#displaySearchFacetValue($category {} 'displaySearchFacetValue_mediaType')&lt;/div&gt;
+          &lt;div id="mediatype-$stringtool.substringBefore($category.name, '\')" class="mediaTypeCategory"&gt;#displaySearchFacetValue($category {} 'displaySearchFacetValue_mediaType' true)&lt;/div&gt;
```
michitux, line 154: Shouldn't this be a `/`? Or is `$category.name` not something like `image/png`?

michitux (changes requested):
I know this was already the case before, but I find it inconsistent that the number of documents is directly next to the value for the specific image types while it is aligned to the right for all other filters. Maybe this could also be fixed here? Of course only if it's easy, otherwise let's just open a new bug report for it.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearch.xml
Sereza7: I did not try this out.
Addressed in https://github.com/xwiki/xwiki-platform/pull/2311/commits/8706ef95cb59c12842618d1a2b2d257d5a7d89f5 :+1:

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrMediaTypeFacet.xml
Sereza7: Examples of the value in `$category.name` :
> (text\/* application\/xml* application\/javascript* application\/ecmascript* application\/json* application\/x\-sh* *\+xml)

>  (image\/*)

Here, taking the first substring is enough to get an unique ID, and AFAIK, we avoid any space (that would break the id parameter). It was not very clean (giving IDs like `mediatype-(image` ), so I cleaned it up in https://github.com/xwiki/xwiki-platform/pull/2311/commits/adabed9cbeb5208c5280e889848924b4f41700d2 :+1:

Sereza7:
> I know this was already the case before, but I find it inconsistent that the number of documents is directly next to the value for the specific image types while it is aligned to the right for all other filters. Maybe this could also be fixed here? Of course only if it's easy, otherwise let's just open a new bug report for it.

Note: this is not the case just for images but for all the kinds of media type values:
20931-notAligned

Here is the view after https://github.com/xwiki/xwiki-platform/pull/2311/commits/b84e20a6d7e7e79c100043d7075c9db0696aa3b6 :
20931-RightAlignedNumbers


Let me know if this looks okay :)

___
All the comments have been taken into account, PR is ready for review.

PR #2316, opened 2023-08-29
XWIKI-21253: Improve the Gradient of buttons

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21253
Forum discussion: https://forum.xwiki.org/t/gradient-on-buttons-minimalist-skin-design-1/12618/
Follow up work on image borders: https://github.com/xwiki/xwiki-platform/pull/2319

**Draft PR**

I will document below some before/after screenshot of some elements of XWiki Standard with and without gradients on buttons.

### Base

**before**

image

**after**

image

### Hover

**before**

image

**after**

image

### Active

**before**

image

**after**

image

### Disabled

image

**after**

image

## Edit page buttons

The state change are on `Save & View`.

### Base

**before**

image

**after**

image

### Hover

**before**

image

**after**

image

### Active

**before**

image

**after**

image

### Disabled

image

**after**

image


manuelleduc:
> Do you think we could remove the `.btn-gradient(@color)` mixin? I couldn't find any other occurence of it used in the xwiki organisation on github and I don't think it would make much sense for an extension or a custom script to use it because we already have 5 button variations that should be used instead of implementing a whole new button type.

I've considered it and:
1. it's been there since so long I would say it's part of the API (but we could agree to break it)
1. this can still be useful for whoever creates a theme with gradient buttons on top of flamingo

Sereza7 (approved):

PR #2319, opened 2023-08-30
XWIKI-21254: Improve the border on buttons

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21254
Forum discussion: https://forum.xwiki.org/t/borders-on-buttons-minimalist-skin-design-2/12619

Build on top of https://github.com/xwiki/xwiki-platform/pull/2316, please do not merge first.

All the buttons are borderless, except the default one (i.e., white) which has a border to distinguish it from the background.

Note: the screenshots below contains the changes on gradient on addition to the change on borders.

### Default

**before**

image

**after**

image

## Disabled

**before**

image

**after**

image

### Default

**before**

image

**after**

image

### Default

**before**

image

**after**

image

## Hovers

https://github.com/xwiki/xwiki-platform/assets/327856/f6355ac4-3175-4448-8244-a926959bce3a

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
@@ -15,18 +15,13 @@
 
 .btn-border(@color) {
   border: 1px solid @color;
-  border-bottom-color: darken(@color, 10%);
-}
-
-.btn-no-border(@color) {
-  border: 1px solid @color;
 }
 
 /* &-.button selectors are added in order to override the mapping and legacy usage of .button class */
 .btn {
   &-default {
     .btn-background(lighten(@btn-default-bg, 5%));
-    .btn-border(@btn-default-bg);
+    .btn-border(darken(@btn-default-bg, 5%));
 
     &:hover, 
     &:focus {
@@ -37,11 +32,11 @@
 
   &-primary {
     .btn-background(@btn-primary-bg);
-    .btn-no-border(@btn-primary-bg);
+    .btn-border(@btn-primary-bg);
 
     &:hover {
       .btn-background(darken(@btn-primary-bg, 5%));
-      .btn-no-border(darken(@btn-primary-bg, 5%));
+      .btn-border(darken(@btn-primary-bg, 5%));
     }
   }
 
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
@@ -18,6 +18,10 @@
   border-bottom-color: darken(@color, 10%);
 }
 
+.btn-no-border(@color) {
+  border: 1px solid @color;
+}
+
 /* &-.button selectors are added in order to override the mapping and legacy usage of .button class */
 .btn {
   border-radius: 8px;
@@ -34,11 +38,11 @@
 
   &-primary {
     .btn-background(@btn-primary-bg);
-    .btn-border(@btn-primary-bg);
+    .btn-no-border(@btn-primary-bg);
 
     &:hover {
       .btn-background(darken(@btn-primary-bg, 5%));
-      .btn-border(@btn-primary-bg);
+      .btn-no-border(@btn-primary-bg);
     }
   }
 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
```diff
@@ -9,72 +9,79 @@
...
-      .btn-border(@btn-primary-bg);
+      .btn-background(darken(@btn-primary-bg, 5%));
+      .btn-no-border(darken(@btn-primary-bg, 5%));
     }
   }
 
   &-success, &-success.button {
-    .btn-gradient(@btn-success-bg);
+    .btn-background(@btn-success-bg);
     .btn-border(@btn-success-bg);
```
Sereza7, line 50: Why doesn't this button follow the `btn-no-border` mixin like btn-primary?
I can see a small border on the bottom of all buttons except primary on hte screenshot after the PR.

(the same applies for btn-info, btn-warning and btn-danger at least)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
```diff
@@ -9,72 +9,79 @@
...
+}
+
 .btn-border(@color) {
   border: 1px solid @color;
   border-bottom-color: darken(@color, 10%);
-  .box-shadow(~"inset 0 1px 0 rgba(255,255,255,.2), 0 1px 2px rgba(0,0,0,.05)");
+}
+
+.btn-no-border(@color) {
+  border: 1px solid @color;
```
Sereza7, line 22:
```suggestion
  border: 0;
```
Is there a point to keeping this solid border even in the no-border mixin? From what I understood, it's always the same color as the background, but I don't think we should need a color argument to define a no-border element

manuelleduc: I should indeed document that, but without having a border the same color as the background, the buttons without border are 2px smaller than the ones with a border.
I felt like the easiest way was to have a border the same color as the background.

Sereza7: I don't know this property too well, but maybe you should consider using a different `box-sizing` value to solve this inconsistency :)

https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
```diff
@@ -9,72 +9,74 @@
...
   &-primary {
-    .btn-gradient(@btn-primary-bg);
+    .btn-background(@btn-primary-bg);
     .btn-border(@btn-primary-bg);
 
     &:hover {
-      .btn-gradient(darken(@btn-primary-bg, 5%));
-      .btn-border(@btn-primary-bg);
+      .btn-background(darken(@btn-primary-bg, 5%));
+      .btn-border(darken(@btn-primary-bg, 5%));
```
manuelleduc, line 39: I'm now using `darken` here because I was finding it hard to see the primary button borders.

manuelleduc:
New version with darken 5% on primary buttons.

image

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less
```diff
@@ -9,72 +9,74 @@
...
-  .box-shadow(~"inset 0 1px 0 rgba(255,255,255,.2), 0 1px 2px rgba(0,0,0,.05)");
 }
 
 /* &-.button selectors are added in order to override the mapping and legacy usage of .button class */
 .btn {
   &-default {
-    .btn-gradient(lighten(@btn-default-bg, 5%));
-    .btn-border(@btn-default-bg);
+    .btn-background(lighten(@btn-default-bg, 5%));
+    .btn-border(darken(@btn-default-bg, 5%));
```
manuelleduc, line 24: I'm now using darken here because I was finding it hard to see the default button borders.

Sereza7 (approved):
Tested it out and couldn't figure out a solution with `box-sizing` only.

I could get a solution without borders that involved removing the border width from the padding of the default button. 

I'm not sure this would be better than what you propose because it's probably less computationally heavy but more prone to breaking unexpectedly.

So :+1: for the solution you're proposing here :)

manuelleduc:
Note: I tested those changes on all the color themes we provide by default. While some of them are not free of usability/ux issues, I couldn't find any new ones.

PR #2320, opened 2023-08-30
XWIKI-21242: Login form is submitted twice if the login button is double clicked - and may create user profiles twice depending on the authenticator

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21242


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/form/form.css
```diff
@@ -0,0 +1,7 @@
+/* Visually convey that the form is being submitted once a first submission was done on a form. */
+form.xwiki-form-submitting,
+form.xwiki-form-submitting input,
+form.xwiki-form-submitting select,
+form.xwiki-form-submitting button {
+  cursor: wait;
```
manuelleduc, line 6: Currently the only visual clue is the spinning cursor. Feels enough to me but let me know if you think we should improve this.

michitux: This means there won't be any visible feedback on touchscreen devices which doesn't sound good.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,43 @@ document.observe("xwiki:dom:loaded", function() {
...
+/**
+ * Initializes the submit-once protection on the current page.
+ * Search for all form elements with a .xwiki-submit-once class one the document (or on an updated sub-part of the dom).
+ * For each form, once the form is submitted, mark it with the .xwiki-form-submitting class and prevent new submissions.
+ *
+ *  @since 14.10.17
+ *  @since 15.5.3
+ *  @since 15.8-rc-1
+ */
+document.observe('xwiki:dom:loaded', function () {
```
michitux, line 2005: Shouldn't we use modern JavaScript APIs or at least jQuery instead of Prototype.js?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,43 @@ document.observe("xwiki:dom:loaded", function() {
...
+  /**
+   * Initializes the root element by adding event listeners to all forms with the class 'form.xwiki-submit-once'.
+   * When a form is submitted, it prevents the submission if it is already being submitted and adds a class to visually
+   * convey that the form was submitted.
+   *
+   * @param {Element} rootElement - The root element where the forms are located.
+   */
+  function init(rootElement) {
+    Array.from(rootElement.getElementsByClassName('xwiki-submit-once')).forEach(form => {
+      form.addEventListener("submit", event => {
```
michitux, line 2015: Isn't there a danger of adding the event listener twice? According to my understanding of [the documentation](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), anonymous functions could be added twice so I guess it would be good to either a) note on the form that the function has already been added or b) make the function non-anonymous such that it won't be added twice. Adding the function twice is dangerous from what I understand as it would prevent submitting the form.

michitux (changes requested):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/form/form.css
manuelleduc: Hum indeed. I'll look for another visual clue that does not involve a cursor.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
manuelleduc: I'm not sure it it would have negative consequences, but it's indeed cleaner to make sure to only have one listener of this kind. Changing that.

manuelleduc: ok, I've improved the javascript code to prevent the handler to be registered several times on the same element.
In addition I removed any use of prototype.

manuelleduc:
@michitux @mflorea I tried on my smartphone, and I don't think a visual clue is required on devices with a touch screen since double-clicking is not a common concept there.
So:
1. smashing the submit button is not a common behavior
2. the submit button becomes active which is a weak but existing visual clue

Therefor,e  I would be fine for me to keep it simple with only a cursor change on mouse based devices.
WDYT?

michitux:
I was more thinking about a case where maybe the form submission is aborted, e.g., due to a timeout/the user clicking cancel in the browser and then form submission not working anymore without any explanation.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,45 @@ document.observe("xwiki:dom:loaded", function() {
...
+ * Search for all form elements with a .xwiki-submit-once class one the document (or on an updated sub-part of the dom).
+ * For each form, once the form is submitted, mark it with the .xwiki-form-submitting class and prevent new submissions.
+ *
+ *  @since 14.10.17
+ *  @since 15.5.3
+ *  @since 15.8RC1
+ */
+require(['jquery'], ($) => {
+  /**
+   * Initializes the root element by adding event listeners to all forms with the class 'form.xwiki-submit-once'.
```
mflorea, line 2007:
```suggestion
   * Initializes the root element by adding event listeners to all forms with the class 'xwiki-submit-once'.
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,45 @@ document.observe("xwiki:dom:loaded", function() {
...
+   * When a form is submitted, it prevents the submission if it is already being submitted and adds a class to visually
+   * convey that the form was submitted.
+   *
+   * @param {Element} rootElement - The root element where the forms are located.
+   */
+  function init(rootElement) {
+    Array.from(rootElement.getElementsByClassName('xwiki-submit-once')).forEach(form => {
+      const submittingClass = 'xwiki-form-submitting';
+      form.classList.remove(submittingClass);
```
mflorea, line 2016: Is there a real use case where the form was already submitted when this code is executed? And if the form was submitted, why remove the marker? AFAIU we don't know when the form can be resubmitted, so we can't remove the flag ourselves.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,45 @@ document.observe("xwiki:dom:loaded", function() {
...
+          e.preventDefault();
+        } else {
+          // Add a class to visually convey that the form was submitted
+          form.classList.add(submittingClass);
+        }
+      })
+    });
+  }
+
+  $(document).on('xwiki:dom:loaded xwiki:dom:updated', (event, data) => {
```
mflorea, line 2031: I don't think you need to listen to ``xwiki:dom:loaded`` if you call the ``init`` function below using ``$(...)``. Basically you need to initialize the forms available on page load and those added dynamically afterward, for which a ``xwiki:dom:updated`` is triggered.

mflorea (commented):
Looks good overall.

manuelleduc:
> I was more thinking about a case where maybe the form submission is aborted, e.g., due to a timeout/the user clicking cancel in the browser and then form submission not working anymore without any explanation.

hum indeed that's a legit usecase I did not consider. That's an annoying one... Checking if we can accurately monitor the status of the form submission.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,54 @@ document.observe("xwiki:dom:loaded", function() {
   interceptXMLHttpRequest();
   interceptFetch();
 })();
+
+/**
+ * Initializes the submit-once protection on the current page.
+ * Search for all form elements with a .xwiki-submit-once class one the document (or on an updated sub-part of the dom).
+ * For each form, once the form is submitted, mark it with the .xwiki-form-submitting class and prevent new submissions.
```
lucaa, line 1999: Can't we just disable the form / buttons to prevent re-submission? (back in the day this used to be the classical solution for this problem)

And then, the visual feedback that a user would get about this would be the visual feedback of "this is not clickable", which would be coherent with all the other forms that would not be clickable (regardless of the reason for which they're not clickable...).

manuelleduc: > This solution looks a little complex to me, can't we just disable the form / buttons to prevent re-submission? (back in the day this used to be the classical solution for this problem)

You can't disable the fields, their values wouldn't be sent server side.
I proposed to mark them are `readonly` which would lead more or less to the same result, but does not prevent buttons to be clicked.
Also, @mflorea rightfully suggested that it was best if we could avoid updating the whole form, and only update a single class at the root of the form.

For the UX pov I agree with you but I have no idea how to achieve this.

lucaa: > You can't disable the fields, their values wouldn't be sent server side.

the fields don't need to be disabled. The submit button needs to be disabled.

Since it's just a submit button, the server side management of the form should not depend on the presence of the button itself on the request.

I had the feeling that disabling submit buttons when a POST form is submitted was a sort of a best practice since forever (there are a bunch of examples in this direction on the internet, but indeed there are also many in the direction you implemented, with classes & all).

manuelleduc: > Since it's just a submit button, the server side management of the form should not depend on the presence of the button itself on the request.

This is not exactly correct. With the form below, the value of `a` will depend on the button that trigger the form submit:
```html
<form method="get" action="index.html">
  <input type="submit" name="a" value="1">
  <input type="submit" name="a" value="2">
  <button type="submit" name="a" value="3">XXX</button>
</form>
```

You can try with and without the following js code. The `a` param is missing when the js is here.
```
$(() => {
      let form = $("form");
      form.submit(() => {
        form.find(':input').prop('disabled', true);
      });
    });
```

Then, we can decide to provide a first solution that would be limited to form where the button values can be safely ignored. But that would be less general.

lucaa: > This is not exactly correct. With the form below, the value of a will depend on the button that trigger the form submit:

yes, I didn't mean for the general case where there are multiple submit buttons with different values, I meant for this specific form and generally for all forms with a single submit button - even if I know that there _may_ be forms with a single submit button where the submit value is used on the server side, I consider them rare.

> Then, we can decide to provide a first solution that would be limited to form where the button values can be safely ignored. But that would be less general.

I think it would make sense to try to go for a solution that is less generic (at least for the forms that have a single submit button whose value is not needed on the server), as going for the very generic case makes it _too_ generic and may make it complex and long to implement way beyond the need of the login form (it's a typical case where "best" kills "better").
For example, if we want to take into account the _very_ generic usecase of a form with multiple submit buttons, we'd need to also define what blocking a double submit would mean: would it mean that the same button cannot be clicked twice or would it mean that no other button can be clicked a second time?

manuelleduc: I have updated the PR with a simpler version where the submit buttons of the submitted form are disable during  the first submit.

mflorea: It's not clear to me what was wrong with the initial solution and how is the new solution better.

> The complexity, for me, comes from the fact that we take the responsibility to prevent clicks in our own code, while click preventing is something that already exists as a feature of html, it's called disabled forms / buttons.

@lucaa maybe I missed something, but I don't remember seeing any **click** listener. There was just a **submit** listener which is still present. The only difference I see is that now we also disable **all** the submit buttons. What's the benefit of this? I can only see the downside: the submit button value won't be sent to the server.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
```diff
@@ -1992,3 +1992,48 @@ document.observe("xwiki:dom:loaded", function() {
...
+   */
+  function init(rootElement) {
+    Array.from(rootElement.getElementsByClassName('xwiki-submit-once')).forEach(form => {
+      // Use a submit event handler with namespace 'xwiki-submit-once'. Before defining a handler, any potentially 
+      // registered listener is removed to prevent double registration of the same handler.
+      $(form).off("submit.xwiki-submit-once").on("submit.xwiki-submit-once", (e) => {
+        const submittingClass = 'xwiki-form-submitting';
+        // Prevent if already submitting.
+        if (form.classList.contains(submittingClass)) {
+          e.preventDefault();
```
lucaa, line 2023: To me this logic is superfluous: we're marking the form as submitting as soon as a button is clicked so that we can then catch any other submit that may happen to stop it because it's already submitting.

This is what button disabling is supposed to handle / replace: the possibility for the form to be submitted after it has been submitted once, so all this logic here is completely unnnecessary (unless I missed something).

lucaa: We _may_ keep the xwiki-form-submitting class, if we really want to, but just as a tool to give to devs to make styling, not more.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/xwiki.js
lucaa: > There was just a submit listener which is still present. 

Yeah, that's what I was talking about. There was some logic associated to preventing it, which is still indeed there in the new PR and it's redundant from my pov. I commented on the commit.

lucaa: > It's not clear to me what was wrong with the initial solution and how is the new solution better.

to clarify this too: what I found "too much" in the initial solution is that we were coding the logic of keeping track of the state of the form (submitted, not submitted yet) and also accepting additional form submissions in the logic of our code, only to stop them (also in our code) if the state of the form says that it was already submitted. 
What I was expecting by using the standard disabling mechanism was something along the lines of: when a submit happens, "turn off" the possibility to submit form and just forget about it, nothing else can happen further, we don't need to worry about what will happen further. 

More like instead of allowing an event to happen more than once and then determining, every time it happens, whenther we should let it happen or not based on a state we stored, we just "turn off" its source when it happens first and we don't need to think about to do when it happens again.

PR #2323, opened 2023-09-04
XWIKI-21272: History navigation arrows are not perceived as links and lacking contrast

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21272
@surli reported this issue
## PR Changes
* Replaced the `<` and `>` symbols used for buttons with proper icontheme arrow icons.
* Updated the order of the nodes and structure to fit improvements to the interface
* Updated style to improve usability: action links styled as button, clearer structure of the section, and explicit name on some buttons.
* Updated test selectors

## View
Video demo: https://up1.xwikisas.com/#MxPdvwAXwzg3ketQazbVMw

Before the changes in this PR:
21272-beforePR

After the changes in this PR:
21272-afterPR

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesdoc.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesdoc.vm
@@ -25,22 +25,22 @@
 #if ("$!previousOrigdocVersion" != '')
   #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${previousOrigdocVersion}&rev2=${rev2}")})
   #set ($changesTitle = $services.localization.render('core.viewers.diff.previousVersion'))
-  #set ($previousOrigdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&lt;</a>")
+  #set ($previousOrigdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('left')</a>")
 #end
 #if ("$!nextOrigdocVersion" != '')
   #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${nextOrigdocVersion}&rev2=${rev2}")})
   #set ($changesTitle = $services.localization.render('core.viewers.diff.nextVersion'))
-  #set ($nextOrigdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&gt;</a>")
+  #set ($nextOrigdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('right')</a>")
 #end
 #if ("$!previousNewdocVersion" != '')
   #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev1}&rev2=${previousNewdocVersion}")})
   #set ($changesTitle = $services.localization.render('core.viewers.diff.previousVersion'))
-  #set ($previousNewdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&lt;</a>")
+  #set ($previousNewdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('left')</a>")
 #end
 #if ("$!nextNewdocVersion" != '')
   #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev1}&rev2=${nextNewdocVersion}")})
   #set ($changesTitle = $services.localization.render('core.viewers.diff.nextVersion'))
-  #set ($nextNewdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&gt;</a>")
+  #set ($nextNewdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('right')</a>")
 #end
 ## Display the previous change arrow only if there is a previous version of the original document and there is no extension version
 #if ("$!previousOrigdocVersion" != '' && "$!previousNewdocVersion" != '')
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
@@ -416,6 +416,7 @@ dl.export-tree-legend dt + dd {
 .changes-info-arrow {
   font-size: 3em;
   line-height: 1;
+  align-self: center;
 }
 
 .changes-body > .changes-body-header {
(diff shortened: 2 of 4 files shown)


surli:
@Sereza7 thanks for working on this! Note that, it's to be discussed but IMO we should also remove the central arrow: I don't think it's useful as there is the "From..." and "To.." that are explicit enough, and it makes the UI bad on small screen (arrow on top of the "to" version). 

Now, sorry but I think I still have a problem with that UI :) So first I thought that the look&feel of the buttons with only the arrows is weird. Now I realize that it's not clear to me the purpose of the "Previous change" button, while we also have a left arrow on from version: it feels redundant. 

So we could apply your changes at least to fix the accessibility problems, but I do think we should revamp this UI to makes it easier to understand especially where the "Previous change" is bringing you to.

surli:
> I think the way it is now is 'good enough', and we can apply those changes. We can open a new UX issue if you think there's still work to do on this section :)

Thanks for your work! From a usability point of view, yes I think it's way better. From a look&feel point of view I'm not a big fan of the arrow buttons for the quick navigation between versions, but I don't have a better proposal right now so we can keep it like that for now.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1709,17 +1709,18 @@ web.history.changes.restrictedInfo=For security reasons, at least one of the com
...
 core.viewers.diff.editComment=Change comment:
 core.viewers.diff.noEditComment=There is no comment for this version
-core.viewers.diff.nextChange=Next change
-core.viewers.diff.previousChange=Previous change
-core.viewers.diff.nextVersion=Next version
-core.viewers.diff.previousVersion=Previous version
+core.viewers.diff.nextVersion.origin=Increase the origin version to {0}
+core.viewers.diff.previousVersion.origin=Decrease the origin version to {0}
+core.viewers.diff.nextVersion.target=Increase the target version to {0}
+core.viewers.diff.previousVersion.target=Decrease the target version to {0}
```
michitux, line 1723: I'm not super happy with the terms "origin version" and "target version". What do you think about "base version" and "comparison version"? Alternatively, we could also keep "target" and just change "origin" to "base". Any opinions?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesdoc.vm
```diff
@@ -24,40 +24,44 @@
...
-  #set ($previousOrigdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&lt;</a>")
+  #set ($changesTitle = $services.localization.render('core.viewers.diff.previousVersion.origin',[${previousOrigdocVersion}]))
+  #set ($previousOrigdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('left')</a>")
 #end
 #if ("$!nextOrigdocVersion" != '')
   #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${nextOrigdocVersion}&rev2=${rev2}")})
-  #set ($changesTitle = $services.localization.render('core.viewers.diff.nextVersion'))
-  #set ($nextOrigdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&gt;</a>")
+  #set ($changesTitle = $services.localization.render('core.viewers.diff.nextVersion.origin',[${nextOrigdocVersion}]))
+  #set ($nextOrigdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('right')</a>")
```
michitux, line 33: I think `$changesTitle` is missing escaping here. Same for the other titles.

Sereza7: :+1: Addressed in https://github.com/xwiki/xwiki-platform/pull/2323/commits/1278c35906dd44ccc2a6b131fd6dcf957dabf35f

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
surli: I personally would keep it like that for now. Maybe it's just me but I do find it more clear than "base version" and "comparison version" to know which is which without viewing the UI.

Sereza7: I agree that 'origin' was an awkward way to put a name on this concept.

I think the 'base' version could be thought of as being the 1.0 . I wanted to get names that conveyed their role in calculating the diff. 

Thinking about it now, it seems that 'source' is better than 'origin':
 source version -> target version

Do you think it's better than 'base'?
__
I don't like too much 'comparison version', because the comparison word does not give an idea of the role in this comparison: base (base of comparison), origin (origin of the comparison), and target (target for the comparison) all bring more information that is useful.

In https://github.com/xwiki/xwiki-platform/pull/2323/commits/ba9d2007631b73a1d961ccd0b06042fa3427a6ae I updated ~~origin~~ --> base

Sereza7:
@surli Following your comment on `#xwiki` this morning, I updated the structure. See https://github.com/xwiki/xwiki-platform/pull/2323/commits/43b7abd207218eff7c50b104edc6dd5bdad0681b .

21272-afterPRbothComments

____

@michitux 
> Why did you remove the background? Is this still needed for contrast? I'm wondering if the background doesn't help to group these navigation links? Or do you think it is clear enough even without background?

It's not needed for contrast. I do not think it was very important here. IMO there's enough blank space under this section to see the separation with the other elements. Maybe a background around each version block would be useful though. [Here is a video of the prototype I got after tweaking a bit with a couple styles](https://up1.xwikisas.com/#G5w34IPHNKheQcBrX4nukA)
21272-afterPRchangeBackground

I think it looks better than without it (see above), so I committed it in https://github.com/xwiki/xwiki-platform/pull/2323/commits/7612bd14b84ab371353215a1c8b37566acaad85b . We can revert it before merging if there's an issue with this UI change :)

surli:
> @surli Following your comment on #xwiki this morning, I updated the structure. See https://github.com/xwiki/xwiki-platform/commit/43b7abd207218eff7c50b104edc6dd5bdad0681b .

Thanks it's making it way more clear.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/changesdoc.vm
```diff
@@ -24,40 +24,44 @@
...
-  #set ($changesTitle = $services.localization.render('core.viewers.diff.nextVersion'))
-  #set ($nextOrigdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&gt;</a>")
+  #set ($changesTitle = $escapetool.html($services.localization.render('core.viewers.diff.nextVersion.origin', [${nextOrigdocVersion}])))
+  #set ($nextOrigdocLink = "<a class='btn btn-default changes-arrow' href='$changesLink' title='$changesTitle'>$services.icon.renderHTML('right')</a>")
 #end
 #if ("$!previousNewdocVersion" != '')
   #set ($changesLink = ${xwiki.getURL($tdoc, 'view', "viewer=changes&rev1=${rev1}&rev2=${previousNewdocVersion}")})
-  #set ($changesTitle = $services.localization.render('core.viewers.diff.previousVersion'))
-  #set ($previousNewdocLink = "<a class='changes-arrow' href='$changesLink' title='$changesTitle'>&lt;</a>")
+  #set ($changesTitle = $escapetool.html($services.localization.render('core.viewers.diff.previousVersion.target', [${previousNewdocVersion}])))
```
michitux, line 37: I think it should be fine here, but `$escapetool.html` doesn't escape `{` which might be dangerous if this template is included in a Velocity macro inside a nested HTML macro as it doesn't prevent closing the HTML macro. That's why we normally use `$escapetool.xml`.

Sereza7: :+1: Addressed in https://github.com/xwiki/xwiki-platform/pull/2323/commits/082bbf5d74b6d34d5c99a4e90fe1eebca9a3bc28

PR #2327, opened 2023-09-06
XWIKI-20409: Insert a LiveData as velocity macro

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20409

**TODO**
- [x] validate that the revapi breakage is ok
- [x] see with @surli which versions he needs this improvement on

**Note**
I successfully improved `children.vm` with this improvement locally:

```diff
Index: xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/children.vm
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/children.vm b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/children.vm
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/children.vm	(revision a9ad624af0e648091830a9d6421f6df011a7e6f1)
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/children.vm	(date 1704808196594)
@@ -85,17 +85,15 @@
       'childrenOf': $doc.space,
       '$doc': $doc.documentReference
     }))
-    {{/html}}
 
-    {{liveData
-      id='childrenIndex'
-      source='liveTable'
-      sourceParameters="$sourceParameters"
-      properties='doc.title,doc.location,doc.date,doc.author,_actions'
-      sort='doc.location:asc'
-      }}$jsontool.serialize($liveDataConfig){{/liveData}}
-    
-    {{html clean='false'}}
+    $services.liveData.execute({
+      'id':'childrenIndex',
+      'source': 'liveTable',
+      'sourceParameters': $sourceParameters,
+      'properties': 'doc.title,doc.location,doc.date,doc.author,_actions',
+      'sort': 'doc.location:asc'
+    }, $liveDataConfig)
+
   #end
 #end
 #######################################################
```


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/pom.xml
```diff
@@ -31,7 +31,7 @@
   <artifactId>xwiki-platform-livedata-api</artifactId>
   <name>XWiki Platform - Live Data - API</name>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.77</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.64</xwiki.jacoco.instructionRatio>
```
surli, line 34: Not great, LiveDataRenderer at least is missing a test no?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/internal/LiveDataRendererParameters.java
```diff
@@ -0,0 +1,258 @@
...
+    {
+        return this.id;
+    }
+
+    /**
+     * Sets the live data instance id.
+     *
+     * @param id the live data instance id
+     */
+    @PropertyDescription("The live data instance id.")
```
tmortagne, line 67: Is this a leftover ?

If not, then I don't understand why there and not on the other setters.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/templates/liveData/macro.vm
```diff
@@ -0,0 +1,32 @@
...
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+## Lesser General Public License for more details.
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+## Since: 16.0RC1
```
surli, line 20: Wrong format.

surli (approved):
Looks good except the minor comments

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/internal/LiveDataRendererParameters.java
manuelleduc: Yes, that a miss. Thanks.
@tmortagne do you agree with the revapi exclude https://github.com/xwiki/xwiki-platform/pull/2327/files/46bda7abc69c944390012b9fc5072261a34d2e20#diff-2f81214bd08060b9f2f1bdd7cb0dda0e7109ccadbf852cda6ddb864abc29c8e3R196-R208 ?

tmortagne: Yes, according to Revapi's own documentation, it's not a runtime breakage. It's only a "potentially breaking", it report it just in case basically.

tmortagne (approved):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/internal/LiveDataRenderer.java
```diff
@@ -0,0 +1,118 @@
...
+     * Render the Live Data content.
+     *
+     * @param parameters the parameters used to render the Live Data
+     * @param advancedParameters a json string containing the advanced parameters of the Live Data
+     * @param restricted when {@code true}, the Live Data is contained in a restricted environment and its content
+     *     will be considered as untrusted
+     * @return the {@link Block}s to render for this Live Data
+     * @throws LiveDataException in case of issue when resolving the Live Data configuration
+     */
+    public List<Block> render(LiveDataRendererParameters parameters, String advancedParameters, boolean restricted)
```
michitux, line 85: Why is this returning a list of blocks? From what I can see this only complicates things as the return value is always a list with a single item, and it also has the consequence that any caller now needs to iterate over the blocks. If there shall be several blocks in the feature, a composite block could always be used to group them transparently.

manuelleduc: Good catch. This will allow to simplify the macro code as well. Thanks.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/templates/liveData/macro.vm
```diff
@@ -0,0 +1,32 @@
...
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+## Since: 16.0.0RC1
+## Provide the macro to display a Live Data using Velocity.
+## $parameters the same parameters are the live data macro, but as an object instead of a set of macro parameters
+## $advancedParameters the advanced parameters usually located in the macro body, but as an object
+## $syntax the syntax to use for the rendering of the result
+#macro (liveData $parameters $advancedParameters $syntax)
```
michitux, line 25: I wonder if it wouldn't be cleaner to provide this simple helper macro directly in the script service.

michitux: Actually, I just noticed that you're serializing an object to JSON here and then later this is deserialized again. Couldn't we avoid that JSON roundtrip here?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/script/LiveDataScriptService.java
```diff
@@ -152,6 +161,64 @@ public String effectiveConfig(String liveDataConfigJSON)
...
+    /**
+     * Renders a Live Data as a {@link String}.
+     * 
+     * @param parameters the parameters to pass to the Live Data executor
+     * @return the result of {@link #render(Map)} in the current syntax
+     * @throws LiveDataException in case of error when rendering the Live Data
+     * @since 16.0.0RC1
+     */
+    @Unstable
+    public String execute(Map<String, Object> parameters) throws LiveDataException
```
tmortagne, line 202: It would be much more consistent with other API to do what I suggested:
* render -> string
* execute -> Block

PR #2328, opened 2023-09-07
XWIKI-21290: Dashboard edit title can be wrong if the content contains an element with a title class

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21290

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/dashboard/dashboard.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/dashboard/dashboard.js
@@ -375,24 +375,35 @@ XWiki.Dashboard = Class.create( {
   onEditGadgetClick : function(event) {
     var gadget = event.element().up('.gadget');
 
+    /**
+     * Finds the first direct child element of the given element with the specified class name.
+     *
+     * @param {Element} element - the element to search for the child in
+     * @param {string} className - the class name of the child element to find
+     * @returns {Element|undefined} the direct child element with the specified class name, or undefined if not found
+     */
+    function findDirectChildWithClass(element, className) {
+      return element.immediateDescendants().filter((e) => e.hasClassName(className)).first();
+    }
+
     if (gadget) {
       // check if it is a macro
-      var gadgetMetadata = gadget.down('.metadata');
+      const gadgetMetadata = findDirectChildWithClass(gadget, 'metadata');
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/dashboard/dashboard.js
```diff
@@ -375,24 +375,35 @@ XWiki.Dashboard = Class.create( {
...
 
+    /**
+     * Finds the direct child element of the given element with the specified class name.
+     *
+     * @param {Element} element - the element to search for the child in
+     * @param {string} className - the class name of the child element to find
+     * @returns {Element|undefined} the direct child element with the specified class name, or undefined if not found
+     */
+    function findDirectChildWithClass(element, className) {
+      return element.immediateDescendants().filter((e) => e.hasClassName(className)).first();
```
manuelleduc, line 386: Couldn't find a simpler way to write this, but if anybody has a suggestion I take it.

Sereza7: I think you could use Jquery, but it's barely used in this script, so it seems to me like your way is correct
It would be something like :

> $(element).children('.'+className).first().get()

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/dashboard/dashboard.js
```diff
@@ -375,24 +375,35 @@ XWiki.Dashboard = Class.create( {
   onEditGadgetClick : function(event) {
     var gadget = event.element().up('.gadget');
 
+    /**
+     * Finds the direct child element of the given element with the specified class name.
```
Sereza7, line 379: Maybe it would be worth indicating that this always returns the first matching element.

Sereza7 (approved):

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/dashboard/dashboard.js
manuelleduc: The best practice is definitely to switch to jquery/vanillajs to remove prototype. But I would say just changing this single selector wouldn't be beneficial.

PR #2329, opened 2023-09-07
XWIKI-20827: Inline links must be distinguishable without relying on color

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20827

## PR Changes
* Darkened color of the regular text on a themeless wiki and Iceberg colortheme.
* Added a rule to underline anchors in `muted-text` nodes.

## Notes
* This PR follows [a discussion on the forum](https://forum.xwiki.org/t/accessible-inline-links-and-the-extra-accessibility-features/12381) and [another PR trying to solve this issue another way](https://github.com/xwiki/xwiki-platform/pull/2313).
* It was not possible to just make the muted text color darker, because it would have been too similar to the regular text color.
* From my manual tests, this solved the `link-in-text-block` in all situations. Tested on the Home, Sandbox and Help pages.

## View
First, here is a comparison of the Home page with the Iceberg color theme, with and without the changes in this PR:
vvv before the PR
20827-beforePR
vvv after the PR
20827-afterPR2
vvv Side by side view
20827-comparison

Hereafter are views of other pages in a default distribution, with the changes in this PR. Note that some anchors are underlined when used in blocks with the .muted-text mixin.

20827-History
20827-Help
20827-sandbox

At last, a full color comparison between `#333` (former text-color value) and `#222` (new text-color value):
paletteTextColor20827


michitux:
My two cents but I think this is really a design and not a code review questions:

* :+1: for the darker text, I think this is much less intrusive than inconsistent underlining with complicated CSS.
* I'm not sure regarding the underline with muted-text. I understand the reasoning which absolutely makes sense but again this seems inconsistent.

Overall, to me this seems better than the other proposed solutions. Maybe it would be a good opportunity to question the use of muted-text. I'm not saying we should remove it, I'm just saying that we should check where it is used and if it really makes sense there.

I would suggest updating the forum thread with this solution to get some more feedback.

Sereza7:
Thank you for your feedback @michitux !

> I'm not sure regarding the underline with muted-text. I understand the reasoning which absolutely makes sense but again this seems inconsistent.

For muted-text, there is no 'easy' solution by just changing color. Therefore we need to make a selector (just like the solution we had to abandon in https://github.com/xwiki/xwiki-platform/pull/2313) to detect inline links. However, here the selector is easier, since we can assume that everything that uses the `.muted-text` mixin is a `text` element. It's not perfect but I think it's better than the solution we had before, and this drawback is limited to the scope of elements using .muted-text, which is way less than `any element` in the former PR.

Let me know if you think of a better solution, but even though we discussed a lot on the topic I couldn't think of a better one yet.

> Overall, to me this seems better than the other proposed solutions. Maybe it would be a good opportunity to question the use of muted-text. I'm not saying we should remove it, I'm just saying that we should check where it is used and if it really makes sense there.

Note that I took care of some edge cases in https://github.com/xwiki/xwiki-platform/pull/2329/commits/e19d0569c434dbf1241237734e11cd425f74edad. Before that, all anchors in the breadcrumbs would be underlined even though they are not inline. This commit limits the use of underline to the most straightforward kind of use of the mixin: muted textual element with an anchor node inside.

I think it would be difficult to find anywhere where the use of .muted-text is not appropriate:
* It's a pretty straightforward semantic: 'this element should not receive too much user attention'
* From experience, it's not used enough rather than too much (and the UI breaks in niche UC / secondary color themes because this semantic is conveyed through hard coded colors or non full opacity).

> I would suggest updating the forum thread with this solution to get some more feedback.

I reopened the topic and posted:
https://forum.xwiki.org/t/accessible-inline-links-and-the-extra-accessibility-features/12381/12 :+1:

Sereza7:
Note that Github seems to be heading towards a new option to underline links:
https://ericwbailey.website/published/github-now-has-a-setting-to-underline-links/

vmassol:
> Note that Github seems to be heading towards a new option to underline links:
https://ericwbailey.website/published/github-now-has-a-setting-to-underline-links/

Yes, and they say what we said, i.e. not everyone wants underline so it needs to be an option.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/misc.less
```diff
@@ -520,3 +520,13 @@ a.html-diff-context-toggle {
 .xwiki-select-option-selected {
   background-color: @nav-link-hover-bg;
 }
+
+// Underline the links in nodes that use the mixin text-muted.
+// This allows to keep a difference between anchors and regular text in those nodes where the
+// contrast between anchors and regular text is too low.
+.text-muted {
+  color: @text-muted;
+  & > a {
```
michitux, line 529: Isn't this selector too restrictive? For example, from wiki syntax, we frequently wrap links in a span.

Sereza7: > Note that I took care of some edge cases in https://github.com/xwiki/xwiki-platform/commit/e19d0569c434dbf1241237734e11cd425f74edad. Before that, all anchors in the breadcrumbs would be underlined even though they are not inline. This commit limits the use of underline to the most straightforward kind of use of the mixin: muted textual element with an anchor node inside.

____

It's not a perfect selector unfortunately... :/
From my understanding, if we wrap the link in a span, then it's correct to assume the '.text-muted' is only applied on this span. From what I could see, this selector was appropriate and allowed to avoid getting too many underlines everywhere (the breadcrumb ones were not that good).

From what I could see it worked well on the few pages I tested on (Home, Help, Sandbox and change history).

Sereza7:
Removed the changes related to the `.muted-text` mixin to provide an almost complete solution without the complexities brought by conditional inline underlining.

PR #2281, opened 2023-08-04
XWIKI-21096: Add keyboard navigation to the Lightbox feature

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21096
## PR Changes
* Added a button for toggling the lightbox that gets visible on keyboard navigation

## Notes
* I chose not to use the attributes `aria-controls` and `aria-expanded` for this toggler since the element that is controled is generated only when it appears.
* I added an invisible tabstop after the lightbox popup to be able to focus back correctly to the main content and not get the focus out of the page.

## View
Video demo: https://up1.xwikisas.com/#oOU64PbniOtkFIaKgH3LuA

PR #2330, opened 2023-09-08
XWIKI-21096: Add keyboard navigation to the Lightbox feature

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21096
## PR Changes
* Fixed an inconsistency in a translation key introduced in https://github.com/xwiki/xwiki-platform/pull/2281

### DIFF
--- a/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
+++ b/xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
@@ -22,7 +22,7 @@ define('xwiki-lightbox-messages', {
   keys: [
     'date',
     'author',
-    'toggle.label'
+    'toggler.label'
   ]
 });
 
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-image/xwiki-platform-image-lightbox/xwiki-platform-image-lightbox-webjar/src/main/webjar/lightbox.js
```diff
@@ -22,7 +22,7 @@ define('xwiki-lightbox-messages', {
   keys: [
     'date',
     'author',
-    'toggle.label'
+    'toggler.label'
```
michitux, line 25: Where are you using this key? As far as I can see, this key isn't used in JavaScript and should be removed.

Sereza7: I just remembered to check that this key is actually useful after sending the message, way too late ^^'
Sorry, this is addressed in https://github.com/xwiki/xwiki-platform/pull/2330/commits/b25c4bf8b3be2cf2e79a1a58820946ad229df6fa.

PR #2331, opened 2023-09-08
Cache template and compile Velocity template

tmortagne:
Before that change, templates are loaded and executed from scratch every time. The ideas is to:
* keep the used `Template`s in a cache
* for those templates which are Velocity template also cache the compiled version of the content (depends on https://github.com/xwiki/xwiki-commons/pull/508)

My first quick test on the standard home page with the skin gives:
* 5x faster to GET
* the cost being a new ~8MB cache containing about 230 templates (yes, we have quite a few of those)

In terms of cache invalidation, there are two mechanisms in this version:
* when accessing a template, the date of the cache template is compare to the current date of the content (or if it exist at all)
* just in case, the template cache also reacts to `CacheControl#setCacheReadAllowed` (in short, it means you can force refresh it the same way you would for the browser cache and async rendering cache: CTRL+SHIT+R, CTRL+F5, etc.)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractResourceSkin.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractResourceSkin.java
@@ -20,10 +20,8 @@
 package com.xpn.xwiki.internal.skin;
 
 import java.net.URL;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.attribute.FileTime;
 
 import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
@@ -48,8 +46,6 @@ public abstract class AbstractResourceSkin extends AbstractSkin
     protected static final Logger LOGGER = LoggerFactory.getLogger(AbstractResourceSkin.class);
 
     private Configuration properties;
-
-    private FileTime filetime;
 
     /**
      * Default constructor.
@@ -136,14 +132,6 @@ public abstract class AbstractResourceSkin extends AbstractSkin
 
                 this.properties = new BaseConfiguration();
             }
-        } else {
-            try {
-                Path resourcePath = Paths.get(getResourceURL(getPropertiesPath()).toURI());
-                FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
-            } catch (Exception e) {
-                
-            }
-
         }
 
         return this.properties;
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractResourceSkin.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractResourceSkin.java
@@ -20,8 +20,10 @@
 package com.xpn.xwiki.internal.skin;
 
 import java.net.URL;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.FileTime;
 
 import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
@@ -46,6 +48,8 @@ public abstract class AbstractResourceSkin extends AbstractSkin
     protected static final Logger LOGGER = LoggerFactory.getLogger(AbstractResourceSkin.class);
 
     private Configuration properties;
+
+    private FileTime filetime;
 
     /**
      * Default constructor.
@@ -132,6 +136,14 @@ public abstract class AbstractResourceSkin extends AbstractSkin
 
                 this.properties = new BaseConfiguration();
             }
+        } else {
+            try {
+                Path resourcePath = Paths.get(getResourceURL(getPropertiesPath()).toURI());
+                FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
+            } catch (Exception e) {
+                
+            }
+
         }
 
         return this.properties;
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/ClassLoaderSkinResource.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/ClassLoaderSkinResource.java
@@ -20,6 +20,7 @@
 package com.xpn.xwiki.internal.skin;
 
 import java.io.InputStream;
+import java.time.Instant;
 
 import javax.inject.Provider;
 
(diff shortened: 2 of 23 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractResourceSkin.java
```diff
@@ -132,6 +136,14 @@ protected Configuration getProperties()
 
                 this.properties = new BaseConfiguration();
             }
+        } else {
+            try {
+                Path resourcePath = Paths.get(getResourceURL(getPropertiesPath()).toURI());
+                FileTime lastModifiedTime = Files.getLastModifiedTime(resourcePath);
+            } catch (Exception e) {
```
surli, line 143: Not even a log or something?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/ClassLoaderSkinResource.java
```diff
@@ -51,11 +52,23 @@ public class ClassLoaderSkinResource extends AbstractSkinResource
...
 
+    private static String createId(ClassLoader classloader, String path)
+    {
+        return "classloader:" + classloader.getName() + ':' + path;
+    }
+
+    @Override
+    public Instant getInstant() throws Exception
+    {
+        // TODO: return the date of the package file holding the resource
```
surli, line 68: Future improvment or you forgot it?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/template/InternalTemplateManager.java
```diff
@@ -204,32 +223,55 @@ public String getPath()
...
+
+            return getContentInternal(strinContent);
+        }
+
         protected abstract T getContentInternal(String content) throws Exception;
 
+        /**
+         * @return the instant the cache content of that template was last modified or null if it's unknown
+         * @since 15.8RC1
+         */
```
surli, line 268: The javadoc should be on the interface no? Not sure why you need it here.

tmortagne: Yes. It initially was only exposed here, and I forgot to remove the javadoc when I exposed it also in the interface.

xwiki-platform-core/xwiki-platform-velocity/xwiki-platform-velocity-webapp/src/main/java/org/xwiki/velocity/XWikiWebappResourceLoaderInitializer.java
```diff
@@ -48,8 +48,5 @@ public void initialize(Properties velocityProperties)
         velocityProperties.setProperty(
             RuntimeConstants.RESOURCE_LOADER + '.' + RESOURCE_LOADER_ID + '.' + RuntimeConstants.RESOURCE_LOADER_CLASS,
             XWikiWebappResourceLoader.class.getName());
-
-        // Add macros.vm as default template
-        velocityProperties.put(RuntimeConstants.VM_LIBRARY, "/templates/macros.vm");
```
surli, line 53: Why is it not needed anymore?

surli: Ah found it, you put it in https://github.com/xwiki/xwiki-platform/pull/2331/files#diff-c35980baa623937175da7e26c81623cdd604dabae2f84d1810f58fae00b17be6R311

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/EnvironmentSkin.java
```diff
@@ -52,7 +52,7 @@ public class EnvironmentSkin extends AbstractResourceSkin
      * @param logger a logger used to log warning in case of error when parsing a skin's syntax
      * @param environment the wiki environment, this is where this skin load its resources from
      * @param xcontextProvider a wiki context provide, used to give access to the context when resolving the skin's
-     *     rsources.
+     *            rsources.
```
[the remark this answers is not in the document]
tmortagne: This was actually automatic formatting, did not really notice the change. Since there was no real modification to this file in the end, the best from the point of view of this PR might be to completely revert those changes.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/AbstractResourceSkin.java
tmortagne: Looks like I forgot (otherwise I would have put a comment).

tmortagne: Actually that code has no reason to exist at all (and it does nothing). I must have forgot to remove it.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/skin/ClassLoaderSkinResource.java
tmortagne: Yes, more a future improvement since it's not really needed yet, but could become a nice to have.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-api/src/main/java/org/xwiki/skin/Resource.java
```diff
@@ -57,6 +60,17 @@ public interface Resource<I extends InputSource>
      */
     I getInputSource() throws Exception;
 
+    /**
+     * @return the last modified time of that resource
```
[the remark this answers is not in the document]
tmortagne: Yes, I will replace this javadoc with something similar to Template#getInstant.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-api/src/main/java/org/xwiki/skin/Resource.java
```diff
@@ -57,6 +60,17 @@ public interface Resource<I extends InputSource>
...
+     * @throws Exception when failing to get the instance of the resource
```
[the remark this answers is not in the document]
tmortagne: You are thinking about servlet resource based Resource implementations (because I was too lazy to try to make the difference between a failure caused by a type of URL for which it's expected to not be able to get the instance and a type for which the problem is different), but wiki entity based implements (in the case of a wiki skin for example) can throw an exception (when failing to access the document).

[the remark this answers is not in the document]
surli: Ok and you really prefer to throw the exception there, than to fallback on a null value?

xwiki-platform-core/xwiki-platform-velocity/xwiki-platform-velocity-webapp/src/main/java/org/xwiki/velocity/XWikiWebappResourceLoaderInitializer.java
tmortagne: Yes, I have to completely think the support of macros.vm, as macros provided at Runtime init have priority over those passed as libraries (does not make much sense to be...). The good thing is that it will be much easier to support modifying macros.vm at runtime (right now, if you modify macros.vm and refresh, it's still using the old one).

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-api/src/main/java/org/xwiki/skin/Resource.java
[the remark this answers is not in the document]
tmortagne: I prefer to make the distinction between:
* we cannot know the instant of this resource, but there is nothing wrong with it
* there is a real problem with the resource (and trying to access the content will most probably fail with a similar error), and we should be able to know the instant of this resource

PR #2335, opened 2023-09-11
XWIKI-18880: Allow to decide where the Mark as read button is put in notifications template

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-18880

  * Insert the read button in the template
  * Enable it in the notification macros if needed

Note that right now these change might have an impact on backward compatibility if an extension uses a customized template for macro and more specifically if they override `displayNotificationEventSkeleton`: they would lose the "Mark as read" button since it was injected dynamically. Now I really doubt we have much extensions overriding it: at least I don't find any occurrence of such override in xwiki-contrib.
Also note that the provided changes have no impact at all on the UI look&feel.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm
@@ -114,7 +114,7 @@
               aria-label="$escapetool.xml($services.localization.render('notifications.macro.markEventAsRead'))"
               title="$escapetool.xml($services.localization.render('notifications.macro.markEventAsRead'))"
               disabled="disabled">
-        <span class="fa fa-check"></span>
+        $services.icon.renderHTML('check')
       </button>
       $content
       #if ($stringtool.isNotBlank($details))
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm
```diff
@@ -110,6 +110,12 @@
       #end
     </div>
     <div class="col-xs-9 notification-content">
+      <button class="notification-event-read-button btn btn-xs hidden"
+              aria-label="$services.localization.render('notifications.macro.markEventAsRead')"
+              title="$services.localization.render('notifications.macro.markEventAsRead')"
```
manuelleduc, line 115: Shouldn't those be xml escaped?

surli: Maybe, here I reuse same structure as lines below (for the details button).

michitux: Definitely, this needs escaping, imagine a translation `Mark this "event" as read`. Existing uses aren't good as examples, we have even translations that contain XML-escaping... Note that previously, escaping was automatic as JavaScript APIs were used for setting the attributes.

PR #2338, opened 2023-09-11
XWIKI-20904: The generated translation key for standard notification message is too strict

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-20904

This work provides a new macro in the notifications macro to compute the description to display for a composite event. The computation is done as before, based on the number of users concerned by the event. The main change is that the macro performs several fallbacks to try to always display something.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm
@@ -65,16 +65,7 @@
       #end
     </div>
     <div class="notification-description">
-      #if ($compositeEvent.users.size() == 1)
-        $services.localization.render("notifications.events.${compositeEvent.type}.description.by.1user", ["#displayNotificationEventUsers($compositeEvent.users, true, false)"])
-      #else
-        <div>
-          $services.localization.render("notifications.events.${compositeEvent.type}.description.by.users", [$compositeEvent.users.size()])
-        </div>
-        <div>
-          #displayNotificationEventUsers($compositeEvent.users, true, true)
-        </div>
-      #end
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm
```diff
@@ -243,3 +234,45 @@ $xwiki.getPlainUserName($user)##
...
+  <div>
+    #displayNotificationEventUsers($compositeEvent.users, true, true)
+  </div>
+  #end
+#end
+
+#macro(getNotificationTranslation $prefix $suffix $fallbackPrefix $parameters)
+  #set ($translationResult = $NULL)
+  #set ($translationKey = "${prefix}.${suffix}")
+  #set ($translationResult = $services.localization.render($translationKey, $parameters))
```
tmortagne, line 273: You can pass directly several keys to #render instead of taking care of the fallback yourself.

surli: Wow, I wasn't aware of that API... definitely easier to implement my solution.

surli: If we pass a parameter to a translation that doesn't use it, it's just ignoring the parameter, right?

tmortagne: > If we pass a parameter to a translation that doesn't use it, it's just ignoring the parameter, right?

Yes, the only constraint when you use this API is that all the keys must expect compatible parameters but too many parameters is not a problem.

PR #2342, opened 2023-09-14
XWIKI-18838: Use ARIA landmarks for main page regions

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-18838
## PR Changes
* Updated test selector for the page background in regards to the changes in layout.less
## Note
This is a patch for [a regression](https://ci.xwiki.org/job/XWiki/job/xwiki-platform/job/master/6877/testReport/org.xwiki.flamingo.test.ui/FlamingoThemeIT/Platform_Builds___main___integration_tests___IT_for_xwiki_platform_core_xwiki_platform_flamingo_xwiki_platform_flamingo_theme_xwiki_platform_flamingo_theme_test_xwiki_platform_flamingo_theme_test_docker___Build_for_IT_for_xwiki_platform_core_xwiki_platform_flamingo_xwiki_platform_flamingo_theme_xwiki_platform_flamingo_theme_test_xwiki_platform_flamingo_theme_test_docker___validateColorThemeFeatures_TestUtils__TestInfo_/) introduced in [this previous commit](https://github.com/xwiki/xwiki-platform/commit/22eee6a3f51d8af14357255596ee78644c5a26a1#diff-769e5c7e3d9bd07eeac5769262c5767b3ddb283a4d3e3bd06be2e8b1b56a4438R293).
The element the style was applied on changed, but the selector to check that the correct style was applied did not get updated.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/ViewPage.java
+++ b/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/ViewPage.java
@@ -288,7 +288,7 @@ public class ViewPage extends BasePage
 
     public String getPageBackgroundColor()
     {
-        return getElementCSSValue(By.id("mainContentArea"), "background-color");
+        return getElementCSSValue(By.className("main"), "background-color");
     }
 
     public String getTitleFontFamily()


manuelleduc:
I'm not sure to understand why this changed in the initial XWIKI-18838 commit.
And why we now need to search for the `main` class, which seems to be on the parent of the `#mainContentArea` elements instead.

Sereza7:
@manuelleduc 
We wanted:
* `#mainContentArea` to be consistently the `<main>` element.
* The breadcrumb to not be in the `<main>` (`#mainContentArea`)
* Minimal changes to the architecture of the page


I had to move the `breadcrumb` next to `#mainContentArea` (which makes sense semantically, because the breadcrumb is not the main content of the page). This broke the style, because the properties of `#mainContentArea` would not apply around the breadcrumb anymore (no top padding, looked awful). So I moved the few style rules from `#mainContentArea` to `.main`. This created a couple regressions on style (AFAIK this one, and the one you reported this morning), but I think that this is the best solution for this issue.

AFAIK the parent `.main` had only `#mainContentArea` as a child, so this style swap from child to parent does not change much since we also changed the position of the breadcrumb with it.

Sorry if it's not very clear, it's a summary of [a discussion we had with Michael on the previous PR](https://github.com/xwiki/xwiki-platform/pull/2166#issuecomment-1569829471), and the way I converged to a solution for it.

Here is a quick drawing to see how this works (left is before the PR, right is after the PR):
18838-styleCHange

manuelleduc:
Thanks for the explanations @Sereza7 ! I'll merge this PR for now as without it the CI does not pass.
@michitux will be able to confirm tomorrow :)

PR #2343, opened 2023-09-15
XWIKI-21310: Generalize on the pre-edit document check

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21310

**before**

image

Note that here, the document is also locked by U1. But, since only one error is displayed, when forcing the edit because the page belongs to an extension, the user also force for the locked page without being aware of it.

**after**

image
image

TODO:
- [ ] check functional tests


tmortagne:
I feel this message could benefit from a generic message related to the fact that the page is protected before listing those protections.

surli:
I agree with @tmortagne here, also from a pure look&feel point of view, I'm not a big fan of having several warning box. I mean the whole modal (I think here it's a modal) is a warning, so it's not really relevant to put each check in a warning box IMO.

manuelleduc:
@tmortagne +1 for the generic message. I'll propose something.

@surli 
> I agree with @tmortagne here, also from a pure look&feel point of view, I'm not a big fan of having several warning box. I mean the whole modal (I think here it's a modal) is a warning, so it's not really relevant to put each check in a warning box IMO.

Another option is to have a error box to list of the error messages first, and a warning box to list all the warning message next.
In this case, we still need a separator between messages from different checkers. I'll try with a horizontal (`hr`) bar but let me know if you have something else in mind.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * This class provides the script services for handling document edit confirmation.
+ *
+ * @version $Id$
+ * @since 15.8RC1
+ */
+@Component
+@Singleton
+@Named("editConfirmation")
```
surli, line 53: I'm wondering if we want `$services.editConfirmation` or `$services.modelValidation.edit`?

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
```diff
@@ -0,0 +1,172 @@
...
+            this.logger.warn("Failed to resolve the list of [{}]. Cause: [{}]", EditConfirmationChecker.class,
+                getRootCauseMessage(e));
+            return new CheckResults();
+        }
+    }
+
+    /**
+     * Represents the aggregated result of a check operation.
+     */
+    public static class CheckResults
```
surli, line 91: Not a huge fan of having this class nested in the script service...

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param editForced a boolean value indicating whether the check is forced
+     * @return a {@link CheckResults} object containing the results of the check
+     */
+    public CheckResults check(Boolean editForced)
+    {
+        try {
+            CheckResults result = new CheckResults();
+            boolean unboxedEditForced = Optional.ofNullable(editForced).orElse(false);
+            this.componentManager.<EditConfirmationChecker>getInstanceList(EditConfirmationChecker.class)
```
surli, line 76: Feels like you might want to document that those components should use priorities so that there's a controlled order of the displayed messages.

surli:
> > I agree with @tmortagne here, also from a pure look&feel point of view, I'm not a big fan of having several warning box. I mean the whole modal (I think here it's a modal) is a warning, so it's not really relevant to put each check in a warning box IMO.

> Another option is to have a error box to list of the error messages first, and a warning box to list all the warning message next.
In this case, we still need a separator between messages from different checkers. I'll try with a horizontal (hr) bar but let me know if you have something else in mind.

I actually missed that there could have been errors and warnings. And it might lead to a very bad UI right now, as there's no priorities and you're displaying the messages based on the execution order of the checkers. So you might have first a warning box, then an error box, then a warning, etc. So yes grouping all errors first, and then all warnings would be probably better.

tmortagne:
> have a error box to list of the error messages first, and a warning box to list all the warning message next

+1

> In this case, we still need a separator between messages from different checkers.

+1

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
manuelleduc: Hum, that's a good question. The think is, `ModelValidationScriptService` is in `xwiki-platform-model-validation-default` and `EditConfirmationScriptService` in `xwiki-platform-model-validation-api`.
So, if we name it `modelvalidation.edit`, we need to assume that `xwiki-platform-model-validation-default` is always installed (or that its replacement also provides a `modelvalidation` with a `get(String)` method).

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
manuelleduc: Done + defined priority on existing components.

manuelleduc:
> @tmortagne +1 for the generic message. I'll propose something.
> 
> @surli
> 
> > I agree with @tmortagne here, also from a pure look&feel point of view, I'm not a big fan of having several warning box. I mean the whole modal (I think here it's a modal) is a warning, so it's not really relevant to put each check in a warning box IMO.
> 
> Another option is to have a error box to list of the error messages first, and a warning box to list all the warning message next. In this case, we still need a separator between messages from different checkers. I'll try with a horizontal (`hr`) bar but let me know if you have something else in mind.

done

tmortagne:
Would be interesting to have an example where there is both warning and error messages.

manuelleduc:
> Would be interesting to have an example where there is both warning and error messages.

I haven't manage to get an error + a warning using a real world configuration, but with the browser debugger tweaking, this would look like this:
image

> I feel this message could benefit from a generic message related to the fact that the page is protected before listing those protections.

Regarding your comment for a generic message at the top. Do you still feel like it is needed?

tmortagne:
> this would look like this

Well, since there is an error, there should not be any force button. Also, the "Warning" title does not make much sense when there is an error, maybe it should be something closer to the actual problem like "The page is protected".

> Regarding your comment for a generic message at the top. Do you still feel like it is needed?

Yes, I feel it would be nice to have a short message explaining that the page is protected but that it can be forced, or that it cannot be forced depending on the various levels we end up with.

manuelleduc:
> > this would look like this
> 
> Well, since there is an error, there should not be any force button. Also, the "Warning" title does not make much sense when there is an error, maybe it should be something closer to the actual problem like "The page is protected".

My fault, as I said I produced the screenshot by manually editing the html, but the title would be "Error" and the "force edit" button would not be here either whenever there is at least one error.
> 
> > Regarding your comment for a generic message at the top. Do you still feel like it is needed?
> 
> Yes, I feel it would be nice to have a short message explaining that the page is protected but that it can be forced, or that it cannot be forced depending on the various levels we end up with. But this is in no way a blocker.

ok, I'll add that as it's simple to do :)

tmortagne:
You can control the protection level for extension pages using `extension.xar.protection` in xwiki.properties (the default is `warning`, and you can switch to `deny` for example).

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationChecker.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provides the operation that a pre-edit checker must provide. The checks are called by the
+ * {@link EditConfirmationScriptService} and aggregated in a {@link EditConfirmationCheckerResults}. The components
+ * implementing this role are called in the order of their priorities. It is advised to define a priority to make the
+ * order of the resulting messages deterministic. The two components provided by {@code xwiki-platform}:
+ * {@code SecurityLevelEditConfirmationChecker} and {@code XWikiDocumentLockEditConfirmationChecker} have respectively
```
surli, line 32: I wouldn't put that doc here: it's likely going to be an outdated doc very fast...

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
surli: Hmm indeed that's an issue, and I imagine we cannot move ModelValidationScriptService to the API module?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/edit_confirmation/lock.vm
```diff
@@ -0,0 +1,27 @@
...
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+##
+## Since 15.9RC1
+## The template used by XWikiDocumentLockEditConfirmationChecker, used in case of page locked by another user.
+##
+#set ($lockingUser = "#displayUser($tdoc.lockingUser {'useInlineHTML': true})")
+<p class="xwikimessage">
+  $escapetool.xml($services.localization.render('doclockedby')) ${lockingUser.trim()}.
```
surli, line 26: Isn't $lockingUser a parameter of the translation?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/edit_macros.vm
```diff
@@ -32,53 +32,50 @@
...
+        $escapetool.xml($services.localization.render($messageKey))
       </p>
+      #if (!$checkResult.errorMessages.isEmpty())
+      <div class="box errormessage">
+        #foreach ($block in $checkResult.errorMessages)
+          $services.rendering.render($block, 'html/5.0')
+          #if ($foreach.hasNext)
+            <hr/>
+          #end
+        #end
```
surli, line 62: You forgot to close the div no?

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-handlers/xwiki-platform-extension-handler-xar/src/main/resources/templates/extension/handler/xar/securityLevelEditConfirmationChecker.vm
```diff
@@ -0,0 +1,42 @@
...
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+##
+## Since 15.9RC1
+## The template used by SecurityLevelEditConfirmationChecker, used in case of raised error or warning before 
+## editing a page. 
+##
+#macro (documentProtectedMessage $protectionLevel)
```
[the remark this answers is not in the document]
manuelleduc: yes, indeed not really useful :) I'll simplify that!

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/edit_confirmation/lock.vm
manuelleduc: It's not, but indeed could be interesting to introduce a better translation with params during the refactoring.

manuelleduc: I've introduced `platform.core.editConfirmation.lock.lockedBy` with a param. Note that `doclockedby` is still used in several placed. I haven't updated as some of the required to be careful with parameters escaping.

xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-validation/xwiki-platform-model-validation-api/src/main/java/org/xwiki/model/validation/edit/EditConfirmationScriptService.java
manuelleduc: I see two options:
1. introduce a "second" `ModelValidationScriptService` in `xwiki-platform-model-validation-api` with a lower priority and that would just provide a `get(String)` method, and make the current `ModelValidationScriptService` inherit from it
2. move `ModelValidationScriptService` in `xwiki-platform-model-validation-api`. In this case we need to:
  - add a dependency to `xwiki-platform-security-authorization-api` which is fine
  - introduce an define an internal role `ReplaceCharacterEntityNameValidationConfiguration` in `xwiki-platform-model-validation-api`  and rename the current `ReplaceCharacterEntityNameValidationConfiguration` to `DefaultReplaceCharacterEntityNameValidationConfiguration` in `xwiki-platform-model-validation-api`, and make it inherit from the newly introduce role instead of using `roles = ReplaceCharacterEntityNameValidationConfiguration.class`
  
Feels like the second option is cleaner.

manuelleduc: Commit of option 2 https://github.com/xwiki/xwiki-platform/pull/2343/commits/cde389081427e60c7aebf64155ab8afe556ece5b

surli: > Feels like the second option is cleaner.

That's indeed cleaner, thanks!

surli (approved):

PR #2350, opened 2023-09-18
XWIKI-19490: Resizing columns in Live Data makes the column title too small

michitux:
* Resize both column and name to properly resize columns.
* Remove the forced width of 100% on the table to make sure column width
  are actually respected and also set all other column width to prevent
  weird resizing effects.
* Resize only the column left of the resizer to make it easier to resize
  single columns. This is also more consistent with other applications
  where you can resize columns like LibreOffice.

Jira issue: https://jira.xwiki.org/browse/XWIKI-19490

This is quite a big UX change, would be nice to get some feedback if we want to change it that much or only a little. We could, e.g., also just resize both column and name which would already be a big improvement and basically prevent the actually reported issue. When I tested it, though, the behavior just felt wrong, still, and you could easily reach states where just clicking the resize handle results in a jump in column size because all columns were shrunk or expanded to reach the 100% width. Also, this changes the reset (when double-clicking on a resize handle) to reset all column widths as to make this work I had to fix all column widths and not just some and resetting one width wouldn't have the effect you would expect I think. Here is a video of the new behavior:

[Screencast_20230918_153744-1.webm](https://github.com/xwiki/xwiki-platform/assets/198317/cb6c99db-7eaa-4ada-a4fc-b4345b6fa3cd)

As you can see in the video, there are still cases where the column can move away from the pointer position, but this now only happens when the table scrolls horizontally due to needing less space than before. I don't know if we could (or would want to) prevent this somehow.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
@@ -105,7 +105,7 @@
 <script>
 import LivedataEntrySelectorAll from "../../LivedataEntrySelectorAll.vue";
 import XWikiDraggable from "../../utilities/XWikiDraggable.vue";
-import { mousedownmove } from "../../directives.js";
+import {mousedownmove} from "../../directives.js";
 import XWikiIcon from "../../utilities/XWikiIcon";
 
 export default {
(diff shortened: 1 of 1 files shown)


lucaa:
The size of the columns of a table is not stored, at all, nor for the current user nor for other users. 

Thus, if we assume that the purpose of this resize columns feature is to fix whatever autolayout of tables in the browser didn't get right (although it should be good in 90% of cases for 90% of the users needs).

>  Remove the forced width of 100% on the table to make sure column width are actually respected and also set all other column width to prevent weird resizing effects.

I'm not a big fan of this. I think this table should always stay at its initial size (100% or less if some other container constraints it) and the tool to resize columns should be more of a tool for adjusting how the available space is divided between the columns (which is kind of what the autolayout is doing, is dividing available space between the columns).
The reason for which I prefer it to stay at its original size is because the size of the block should not be in the hands of the user that is using the table, just as the size of any other block of XWiki is not in the hands of the user using the wiki.

> This is also more consistent with other applications where you can resize columns like LibreOffice.

I am not sure that I would compare with _editing_ tools, because this resize is not a resize in the context of an edit (as it's not stored).
However, I would compare with visualisation tools / screens, for example, the network tab of browser's dev tools, where the context is more or less similar: 
* there is some available space given - all browser window width
* there is a proportion of columns proposed by default by the tool
* the user can use the column resize to adjust those proportions, but in the same available space.

If we really worry about the table being too wide, one can always make a very large last column, for example, by pulling everything to the left, or resize their browser window.

michitux:
I've replaced the fix by a simple one-line CSS fix that just fixes the original issue. I'll commit a separate fix for the right column resizing issue, and I'll propose a new PR with a (much simpler) resizing fix.

PR #2351, opened 2023-09-19
XWIKI-21209: Delete filter is slow (seems to load all filters)

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-21209

  * Refactor NotificationFilterPreferenceStore to ensure to use the right wiki in context when the local store is used
  * Use a query based on the filter internal identifier to perform single deletion, to prevent having to load it
  * Refactor the way NotificationFilterPreferenceDeletedEvent is used to not send the entire filter instance when it's triggered
  * Improve unit tests

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
@@ -24,7 +24,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 
 import javax.inject.Inject;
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/internal/NotificationFilterPreferenceStore.java
```diff
@@ -430,19 +436,21 @@ private void saveFilterPreferences(EntityReference entityReference,
...
+        } else if (wikiReference != null) {
+            context.setWikiReference(wikiReference);
         }
-
         try {
             return supplier.get();
         } finally {
-            if (!Objects.equals(currentWikiId, context.getWikiId())) {
-                context.setWikiId(currentWikiId);
+            if (wikiReference != null) {
```
tmortagne, line 452: Not sure what is the point of this if, the context should get back to whatever was the context wiki and what's in this if is not related.

surli: You're right, I don't know why I wrongly assumed that if there was no wikiReference parameter there was no point to reset it. It's only true with the localStore...

tmortagne (approved):

PR #2353, opened 2023-09-19
XWIKI-19449: Livedata table overflows horizontally for container sizes in which livetable fits perfectly fine

michitux:
* Remove margin between column title and sort icon
* Put the drag handle on top of the column title with an opacity of 0.8

Jira issue: https://jira.xwiki.org/browse/XWIKI-19449 (with screenshots of the before state)

Here is a screenshot of the new version:

image

And here another one to demonstrate how the handle for dragging columns is displayed (in the "Release Date" column):

XWIKI-19449-LD-with-column-hover

Live Data still needs more space, but the difference should be much smaller now. I think the main differences are:

1. Livetable has a smaller padding in the title cells, 8px (default of XWiki) vs. 3px.
2. The sort icon has still less padding (I already reduced it with this commit)

I'm not sure that it is worth giving up this extra space as this space is consistent with other parts of the UI.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
@@ -231,18 +231,22 @@ export default {
   /* Ensure that the name is never smaller than the width of the column, i.e., it always fills the available space even
    when the column has been resized to a smaller width that is prevented by some table cell. */
   min-width: 100%;
+  position: relative;
 }
 
 .layout-table .handle {
   height: 100%;
   margin-left: -@table-cell-padding;
   padding: 0 @table-cell-padding;
+  color: @text-color;
+  background-color: @xwiki-page-content-bg;
   cursor: pointer; /* IE */
   cursor: grab;
   opacity: 0;
+  position: absolute;
 }
 .layout-table .column-name:hover .handle {
-  opacity: 1;
+  opacity: 0.8;
   transition: opacity 0.2s;
 }
 .layout-table .handle .fa {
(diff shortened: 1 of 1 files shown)


lucaa:
I'm definitely for displaying the drag handle on top of the title and not reserve explicit room for it, as it's a feature that I'm expecting to use only sometimes (especially as it's not a saved setting).
We may want to double check consistency with the sort handle, which also appears only on hover but has its own reserved space (does it make sense?).

I need to check the padding a little more before speaking more about it. The trick with the padding here is that paddings of 2 adjacent columns would sum. 
So the question is: do we want each individual padding to be consistent with other paddings of XWiki or do we need the sum (the total spacing between the texts of two adjacent columns) to be consistent with other paddings? Since livedata has no borders, we could actually choose to go for the total padding and save some room, but we need to see how consistent is this with other paddings...

michitux:
> We may want to double check consistency with the sort handle, which also appears only on hover but has its own reserved space (does it make sense?).

The thing here is that the sort icon is displayed permanently when a column is sorted. Not reserving space for it would thus either mean that we would display it permanently on top (seems strange) or that column widths would change when you change which column is sorted.

> So the question is: do we want each individual padding to be consistent with other paddings of XWiki or do we need the sum (the total spacing between the texts of two adjacent columns) to be consistent with other paddings? Since livedata has no borders, we could actually choose to go for the total padding and save some room, but we need to see how consistent is this with other paddings...

For the content cells, the padding is 8px in both cells for adjacent cells both in Livetable and Live Data. Livetable is only different for the title cells (which is actually inconsistent). I'm not saying that we cannot remove some padding, I just want to give some more data points what we do in other places.

lucaa:
> The thing here is that the sort icon is displayed permanently when a column is sorted. 

ok then, we need to leave space for it for sure, at least until we have time to look for some creative solution :) 

> For the content cells, the padding is 8px in both cells for adjacent cells both in Livetable and Live Data. Livetable is only different for the title cells (which is actually inconsistent).

Ok for the current paddings, then, fine with me. I'll need to test the final result to see how it feels before having more opionion about it.
The objective of the ticket was not to reach the _exact_ size of the livetable, but to at least have less of a difference (currently the differenece is rather large).

manuelleduc (approved):

PR #2354, opened 2023-09-19
XWIKI-21329: Dragging the resize handle of a Live Data column leads to a jump in size

michitux:
* Set the size of all columns when resizing starts.
* Reset all sizes when resetting column sizes.

Jira issue: https://jira.xwiki.org/browse/XWIKI-21329

Here a video demonstrating both the issue and the solution (note the text at the top):

https://github.com/xwiki/xwiki-platform/assets/198317/20b3aeca-d84a-4d54-ae15-537eccaf985f

This is basically a less drastic version of what I proposed in #2350, taking the feedback from there into account - it still resizes both columns now, and you cannot resize below 100% width.


lucaa:
From a functional point of view it feels very natural, at least in the demo that you're doing in the video.
It's a good solution, from my point of view, I'm all for it (along with the fix for XWIKI-19449).

manuelleduc (approved):

PR #2357, opened 2023-09-20
XWIKI-21331: The document cache silently accepts a failed database load when it loads an existing document

michitux:
* Log an error when a document load fails that shouldn't have failed.

Jira issue: https://jira.xwiki.org/browse/XWIKI-21331

The intention would be to deploy this on xwiki.org with the next update the latest to hopefully get a log when a document disappears.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiCacheStore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiCacheStore.java
@@ -400,6 +400,15 @@ public class XWikiCacheStore extends AbstractXWikiStore
 
                     LOGGER.debug("Document [{}] was retrieved from persistent storage", key);
 
+                    // If the exists cache indicated that the document should exist, log an error with a
+                    // stacktrace.
+                    if (cachedoc.isNew() && result != null) {
+                        LOGGER.error("Document [{}] was not found in persistent storage but the exists cache "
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiCacheStore.java
```diff
@@ -400,6 +400,15 @@ public XWikiDocument loadXWikiDoc(XWikiDocument doc, XWikiContext inputxcontext)
 
                     LOGGER.debug("Document [{}] was retrieved from persistent storage", key);
 
+                    // If the exists cache indicated that the document should exist, log an error with a
+                    // stacktrace.
+                    if (cachedoc.isNew() && result != null) {
+                        LOGGER.error("Document [{}] was not found in persistent storage but the exists cache "
+                                + "indicated that it should exist. Unless that document has just been deleted, this "
+                                + "is a bug and should be reported.", key,
```
surli, line 408: Not sure key is best here to identify the doc by humans, I might have used the ref as it's more human readable.

surli: Also I'd probably have put in the message that this log is for identifying the root cause of XWIKI-21331: it might help if some users see those in their console.

michitux:
Closing as I believe this isn't needed anymore, and we can't add protections/logs for every impossible situation.

PR #2359, opened 2023-09-21
XWIKI-21322: CVE IDs column of the security list does not contain CVE IDs

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21322

**Main changes**
- OSV source now returned CVEs as IDs (and so should any other source)
- Results without CVE are ignored
- Nothing of aliases introduced to prevent re-notifying previously stored vulnerabilities with an outdated id

### DIFF
--- a/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/security/SecurityVulnerabilityDescriptor.java
+++ b/xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/security/SecurityVulnerabilityDescriptor.java
@@ -19,6 +19,8 @@
  */
 package org.xwiki.extension.index.security;
 
+import java.util.Set;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
@@ -38,6 +40,8 @@ import us.springett.cvss.Cvss;
 public class SecurityVulnerabilityDescriptor
 {
     private String id;
+
+    private Set<String> aliases;
 
     private String url;
 
@@ -184,6 +188,33 @@ public class SecurityVulnerabilityDescriptor
         return this;
     }
 
+    /**
+     * Retrieves the set of aliases associated with the current object.
+     *
+     * @return a set of aliases, used for vulnerability comparison, two vulnerabilities sharing an alias or an id are
+     *     considered equals
+     * @since 15.9RC1
+     */
+    @Unstable
+    public Set<String> getAliases()
+    {
+        return this.aliases;
+    }
+
+    /**
+     * Sets the set of aliases associated with the current object.
+     *
+     * @param aliases a set of aliases, used for vulnerability comparison, two vulnerabilities sharing an alias or
+     *     an id are considered equals
+     * @return the current object
+     * @since 15.9RC1
+     */
+    public SecurityVulnerabilityDescriptor setAliases(Set<String> aliases)
+    {
+        this.aliases = aliases;
+        return this;
+    }
+
     @Override
     public boolean equals(Object o)
     {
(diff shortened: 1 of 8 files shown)


xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-security/xwiki-platform-extension-security-index/src/main/java/org/xwiki/extension/security/internal/analyzer/osv/model/response/VulnObject.java
```diff
@@ -154,6 +156,24 @@ public Optional<Version> getMaxFixVersion(Version currentVersion)
             .min(Comparator.naturalOrder());
     }
 
+    /**
+     * @return the list of aliases associated with this vulnerability
+     */
+    public List<String> getAliases()
+    {
+        return this.aliases;
```
tmortagne, line 164: It would simply a lot of code to return an empty set when `this.alias` is null (i.e. never return null).

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/security/SecurityVulnerabilityDescriptor.java
```diff
@@ -184,6 +188,33 @@ public SecurityVulnerabilityDescriptor setReviews(String reviews)
...
+     * Retrieves the set of aliases associated with the current object.
+     *
+     * @return a set of aliases, used for vulnerability comparison, two vulnerabilities sharing an alias or an id are
+     *     considered equals
+     * @since 15.9RC1
+     */
+    @Unstable
+    public Set<String> getAliases()
+    {
+        return this.aliases;
```
tmortagne, line 201: It would simply a lot of code to return an empty set when this.alias is null (i.e. never return null, which by the way is not documented in the javadoc currently).

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-security/xwiki-platform-extension-security-index/src/main/java/org/xwiki/extension/security/internal/analyzer/osv/OsvResponseAnalyzer.java
```diff
@@ -78,17 +81,44 @@ public ExtensionSecurityAnalysisResult analyzeOsvResponse(String extensionId, St
...
 
-    private SecurityVulnerabilityDescriptor convert(VulnObject vulnObject, Version currentVersion)
+    private Optional<SecurityVulnerabilityDescriptor> convert(VulnObject vulnObject, Version currentVersion)
     {
-        return new SecurityVulnerabilityDescriptor()
-            .setId(vulnObject.getId())
-            .setURL(vulnObject.getMainURL())
-            .setSeverityScore(vulnObject.getSeverityCCSV3())
-            .setFixVersion(vulnObject.getMaxFixVersion(currentVersion).orElse(null));
+        // If we are unable to find a CVE, we ignore the vulnerability.
```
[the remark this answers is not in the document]
manuelleduc: We could accept vulnerabilities without CVE, but in this case we can't guarantee that we won't send duplicate notifs for already know vulns.

xwiki-platform-core/xwiki-platform-extension/xwiki-platform-extension-index/src/main/java/org/xwiki/extension/index/security/SecurityVulnerabilityDescriptor.java
```diff
@@ -39,6 +42,8 @@ public class SecurityVulnerabilityDescriptor
 {
     private String id;
```
tmortagne, line 43: Should we also change all the reference to the very generic "id" naming to something closer to "cve" or "cveId" ?

PR #2361, opened 2023-09-25
Improvements to user listing in the rights UI

aubincleme:
Related JIRA issues : 
* [XWIKI-21305](https://jira.xwiki.org/browse/XWIKI-21305) - Use user first name or last name when filtering in the rights livetable
* [XWIKI-21306](https://jira.xwiki.org/browse/XWIKI-21306) - In the user rights UI, display the user first name and last name
* [XWIKI-21349](https://jira.xwiki.org/browse/XWIKI-21349) - Provide parameters to enable or disable linking to user profiles in the user displayer

This PR aims at improving the experience of administrators when administering rights for instances with users who have usernames that do not correspond to their actual first name and last name.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getusersandgroups.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getusersandgroups.vm
@@ -171,9 +171,11 @@ $response.setContentType("application/json")
   #set($discard = $row.put("allows"  , $allows.keySet()))
   #set($discard = $row.put("denys"   , $denys.keySet()))
   #if ($uorg == "users")
-    #set($discard = $row.put("userDisplayer", "#displayUser($user.documentReference, {'displayLink': $isLocalUser})"))
+    #set($discard = $row.put("userDisplayer",
+      "#displayUser($user.documentReference, {'displayLink': $isLocalUser, 'showAlias': true})"))
   #else
-    #set($discard = $row.put("userDisplayer", "#displayGroup($user.documentReference, {'displayLink': $isLocalUser})"))
+    #set($discard = $row.put("userDisplayer",
+      "#displayGroup($user.documentReference, {'displayLink': $isLocalUser, 'showAlias': true})"))
   #end
   #if ($uorg == "groups")
     #set($discard = $row.put("isuseringroup", $xwiki.getUser().isUserInGroup($user.fullName)))
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getusersandgroups.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getusersandgroups.vm
@@ -162,20 +162,19 @@ $response.setContentType("application/json")
     #end
   #end ## foreach rights object
   #set($row = {})
-  #set($userFirstName = $user.getValue('first_name'))
-  #set($userLastName = $user.getValue('last_name'))
-  #if($stringtool.isNotBlank($userFirstName) || $stringtool.isNotBlank($userLastName))
-    #set($userDisplayName = "$!{userFirstName} $!{userLastName} (${user.name})")
-  #else
-    #set($userDisplayName = $user.name)
-  #end
+  #set($isLocalUser = ($wikiname == 'local'))
   #set($discard = $row.put("username", $user.documentReference.name))
   #set($discard = $row.put("fullname", $username))
-  #set($discard = $row.put("title"   , $userDisplayName))
+  #set($discard = $row.put("title"   , $user.title))
   #set($discard = $row.put("wikiname", $wikiname))
   #set($discard = $row.put("userurl" , $xwiki.getURL($user.fullName)))
   #set($discard = $row.put("allows"  , $allows.keySet()))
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getusersandgroups.vm
```diff
@@ -162,9 +162,16 @@ $response.setContentType("application/json")
     #end
   #end ## foreach rights object
   #set($row = {})
+  #set($userFirstName = $user.getValue('first_name'))
+  #set($userLastName = $user.getValue('last_name'))
+  #if($stringtool.isNotBlank($userFirstName) || $stringtool.isNotBlank($userLastName))
+    #set($userDisplayName = "$!{userFirstName} $!{userLastName} (${user.name})")
```
aubincleme, line 168: Note : I believe it's interesting to display the username here in case the username is significantly different from the first name / last name, however it may make the displayed names a bit too long sometimes. WDYT ?

tmortagne: Ideally, this column's display should be based on the standard user displayer. That's its job, after all and it would be much more consistent with other areas of the wiki.

aubincleme: Indeed, updating

aubincleme: @tmortagne fixed as part of [f0e088e](https://github.com/xwiki/xwiki-platform/pull/2361/commits/f0e088e35d8466ee4e0e176019998306c6ee5012). Here is how it looks : 

On a subwiki : 

Groups : 

image

Users : 

image

On the main wiki : 

Groups : 

image

Users : 

image

I don't think we have functional tests for that, can you confirm by any chance ?

tmortagne: I thought your initial goal was to show the alias, but it looks like you did not enable this option (which is false by default from what I see in macros.vm). It's enabled in the group, for example.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -241,7 +241,10 @@ $xwiki.getUserName("xwiki:${username}")
  *   showAlias: false, // whether to show the user alias or not
  *   useInlineHTML: false, // whether to use in-line HTML elements to display the user or not
  *   wrapAvatar: false // whether to wrap the avatar image with a span or not
+ *   displayLink: true // whether to wrap the user name in a link to the user profile
  * }
+ *
+ * @since 15.9
```
tmortagne, line 247: I don't understand, this macro definitely not been introduced in 15.9 and the introduction of this new option haven't changed its "signature". You might want to document this at the level of the displayLink option instead of here, as it's very missleading.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -332,7 +343,10 @@ $xwiki.getUserName("xwiki:${username}")
  *   showAlias: false, // whether to show the group alias or not
  *   useInlineHTML: false, // whether to use in-line HTML elements to display the group or not
  *   wrapAvatar: false // whether to wrap the avatar image with a span or not
+ *   displayLink: true // whether to wrap the group name in a link to the group profile
  * }
+ *
+ * @since 15.9
```
tmortagne, line 349: Same comment as above.

aubincleme: Fixed in a0a237b52f7

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/getusersandgroups.vm
aubincleme: Indeed, displaying the alias gives a much nicer result, updating the PR with `showAlias` to `true` for both groups and users.

aubincleme: Fixed in fcc103b6f29 ; here is how it looks : 

image

tmortagne: > I don't think we have functional tests for that, can you confirm by any chance ?

Not sure, but that area is not my strong suit. If there was something I would expect it to be in xwiki-platform-administration-test-docker or xwiki-platform-distribution-flavor-test-ui.

aubincleme: Oki, I've checked and did not find any tests related to the rights matrix. I'll merge and then keep an eye on the CI to see how it evolves.

PR #2394, opened 2023-10-02
XWIKI-21343: Localization preference select has no accessible name

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21343

## PR Changes
* Added an aria-label to the aforementioned select node

## View
21343-View

## Note
The xHint above contains a text similar to the aria-label, but I judged its value in english was too long for a label so I chose to introduce a new key.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminLocalizationSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminLocalizationSheet.xml
@@ -395,9 +395,9 @@
     </property>
     <property>
       <code>define('xwiki-administration-localization', {
-  prefix: 'administration.localization.',
+  prefix: 'XWiki.',
   keys: [
-    'localepicker.languages.label'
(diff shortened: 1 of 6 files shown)


michitux:
From the UI, it seems to me that the label should be "Supported Languages", so the translation of `XWiki.XWikiPreferences_languages`. Any reason not to use this? Also, ideally with an `aria-labelledby` or a native label relationship (but I guess that's not possible as the text is probably already a label for the text input). The hint could be used via `aria-describedby`.

Sereza7:
> Also, ideally with an aria-labelledby or a native label relationship (but I guess that's not possible as the text is probably already a label for the text input).

:+1: 
21343-labelAlreadyAssigned

> From the UI, it seems to me that the label should be "Supported Languages", so the translation of 
> XWiki.XWikiPreferences_languages. Any reason not to use this? 
> [...]
> The hint could be used via aria-describedby.


I decided to reuse the value already used in the input label as an 'aria-label' value, and to give an ID to the xHint in the field above to use it in 'aria-describedby'.
21343-viewAfterPR

Those changes are committed in https://github.com/xwiki/xwiki-platform/pull/2394/commits/ab5400439c80f299a7495dd7f7dbb9b00baf2b83 :)

PR #2439, opened 2023-10-03
XWIKI-19948: Bad diff information when xobject is deleted

surli:
* Ensure to retrieve the xclass even when the xobject has been deleted

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/BaseObject.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/BaseObject.java
@@ -308,8 +308,9 @@ public class BaseObject extends BaseCollection<BaseObjectReference> implements O
             BaseProperty newProperty = (BaseProperty) this.getField(propertyName);
             BaseProperty oldProperty = (BaseProperty) oldObject.getField(propertyName);
             BaseClass bclass = getXClass(context);
-            // if the object was removed current instance doesn't hold any information about the xclass
-            // however the old object will hold the information.
+            // Bulletproofing: in theory the BaseObject is defined with a xclass reference allowing to resolve it
+            // however, it's possible that the reference is not set, in which case we might still find the info
+            // in the old object.
             if (bclass == null) {
                 bclass = oldObject.getXClass(context);
             }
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/BaseObject.java
```diff
@@ -308,6 +308,11 @@ public List<ObjectDiff> getDiff(Object oldEntity, XWikiContext context)
             BaseProperty newProperty = (BaseProperty) this.getField(propertyName);
             BaseProperty oldProperty = (BaseProperty) oldObject.getField(propertyName);
             BaseClass bclass = getXClass(context);
+            // if the object was removed current instance doesn't hold any information about the xclass
+            // however the old object will hold the information.
+            if (bclass == null) {
+                bclass = oldObject.getXClass(context);
```
tmortagne, line 314: This gets the current class, not the class with which the xobject was serialized (that's the reason why this method takes a XWikiContext: to load the current class from the database).

surli: Sure, but here the UC is about an xobject being deleted, so the xclass is still correct. We need to fix UC when the xobject was using another version of the xclass but that's another issue IMO.

surli: So here the root cause is https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java#L7120 which creates a new BaseObject without specifying the xclass reference. It's probably better to fix that, but I do feel we should keep the bulletproof I'm proposing here since we might have other part of the code doing this kind of call with a BaseObject without any reference.

surli: @tmortagne so as announced I pushed also a fix in XWikiDocument, but I kept this bulletproof. Let me know if you think it's ok.

tmortagne (approved):

PR #2441, opened 2023-10-04
XWIKI-21386: Change viewer meta data lacks contrast

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21386
## PR Changes
* Replaced the half opacity colors with their full opacity alternatives from the color themes (the same as the ones used in .text-muted)
## View
With the Iceberg color theme:
21386-IcebergAfter
With the Darkly color theme (the text colors are hard coded, which make them look off, however I can't see an easy solution to get all of these colors from the color theme, and it's not the focus of this PR)
21386-DarklyAfterPR

We can see that with both color themes, the text passes contrast checks, and is overall more readable. However, it's also less subtle and the eye gets caught on it more easily.

## Note
AFAIK this style rule was created by @mflorea in [this commit](https://github.com/xwiki/xwiki-platform/commit/4ec10a623ed99d887340ddb719d395ebe945dddc#diff-cba9da1f0167da707d85622373578942fbb16bed22d59b271a4ffeea2ae5b9d1). 
Do you think reverting those changes on color is correct? Should we change something else to reduce emphasis on this text?

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/viewers/diff.css
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/viewers/diff.css
@@ -66,7 +66,7 @@ td.diff-line.diff-line-decision span.diff-decision {
 td.diff-line-meta {
   background-color: $theme.backgroundSecondaryColor;
   color: $theme.textSecondaryColor;
-  padding: .4em .5em;
+  padding: .8em .5em;
 }
 
 .diff-line del {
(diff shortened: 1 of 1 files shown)


mflorea (approved):
I think it looks OK. If you check here on GitHub, they also use the "text muted color" for the diff meta. You could maybe increase the line height of the diff meta (e.g. increase the padding) to add more spacing and thus to separate it a bit from rest of the diff.

Sereza7:
> You could maybe increase the line height of the diff meta (e.g. increase the padding) to add more spacing and thus to separate it a bit from rest of the diff.

:+1: Updated in https://github.com/xwiki/xwiki-platform/pull/2441/commits/ff1d43d740cf49d659e4e9ef7185dcd6d595bbe8
Here is the new look in Iceberg:
21386-afterPRchangemargin

mflorea (approved):

PR #2470, opened 2023-10-13
XWIKI-21311: Introduce a required rights analyzer

michitux:
* Introduce an analyzer for rights that are required by macros, XObjects or Velocity code in the title
* Introduce a pre-edit check to warn users about things they might break by editing or additional code that might be executed.

This is joint work with @manuelleduc.

There is still a lot of work to do, not limited to:

* Introduce additional analyzers. We need at least:
  * [ ] HTML macro
  * [ ] Context macro
  * [ ] Raw macro
  * [ ] Content macro (need to check at least)
  * [ ] Cache macro
  * [ ] UI Extension
  * [ ] Translations
  * [ ] Icon themes
* If possible, the different analyzers should be moved into the various modules where the functionality is, and we should use the analyzers in the actual right checks there to avoid code duplication.

We most likely won't have everything ready for 15.9, but I nevertheless suggest we merge this and add additional analyzers in 15.10. It might be good to have a warning that this analysis might not be complete, and most likely will never be complete as custom macros can do arbitrary things with parameters and sheets can do arbitrary things with XObjects.

The goal of this work is to introduce a basis that can be used for different purposes, including at least:

* The edit check that we're introducing now to warn users both when they might break stuff and when they might be about to enable the execution of dangerous code.
* An admin tool to help admins identify documents that don't have enough rights (e.g., due to right changes or users being deleted) and might thus be broken.
* The semi-automated enabling of "document rights" as discussed in the [original proposal for required rights](https://forum.xwiki.org/t/document-rights-management-improvement/12177), both for existing and new content.

Due to this wide range of applications, we decided to let the analyzers just report the required right(s) and leave it to the code running the analysis to check the rights against the desired author.

Jira issue: https://jira.xwiki.org/browse/XWIKI-21311 (note that we plan to create additional issues for the missing functionality once the first part has been released)

Here is a screenshot of the edit warning for the LiveTable view sheet of AWM, I have expanded some sections to show how macro parameters and content as well as XObject parameters and content are rendered:

image


manuelleduc:
New presentation following my last commit
image

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRight.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * @return the level at which the right is required (e.g., document, space, wiki)
+     */
+    public EntityType getEntityType()
+    {
+        return this.entityType;
+    }
+
+    /**
+     * @return whether the right is optional or not, i.e., if the entity also works without the right or not
```
surli, line 73: Why would we need a required right for an entity that would work without it? It could be useful to have a bit more doc / example for this.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRightAnalysisResult.java
```diff
@@ -0,0 +1,116 @@
...
+ * Represents the result of a required right analysis. This class provides getters and setters for various properties
+ * related to the analysis result.
+ *
+ * @version $Id$
+ * @since 15.9RC1
+ */
+@Unstable
+public class RequiredRightAnalysisResult
+{
+    private EntityReference entityReference;
```
surli, line 40: It feels weird that the entity reference is the only field not final here... I would almost have assumed the contrary.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/XWikiDocumentRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,111 @@
...
+                for (List<BaseObject> baseObjects : document.getXObjects().values()) {
+                    for (BaseObject object : baseObjects) {
+                        result.addAll(this.objectRequiredRightAnalyzer.analyze(object));
+                    }
+                }
+
+                return result;
+            }, document);
+        } catch (Exception e) {
+            throw new RequiredRightsException("Error...", e);
```
surli, line 108: Missing a proper message.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.platform</groupId>
+    <artifactId>xwiki-platform-security</artifactId>
+    <version>15.9-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-platform-security-requiredrights</artifactId>
```
surli, line 28: Should this be a submodule of `xwiki-platform-security-authorization`?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsChangedFilter.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Takes a list of {@link RequiredRightAnalysisResult} and filter them according to a set of
+     * {@link DocumentAuthors}.
+     *
+     * @param authors the document authors to use for the filtering
+     * @param resultList the list of required rights analysis result to filter
+     * @return the filtered list of required rights results, according to the provided document authors
+     */
+    public RequiredRightsChangedResult filter(DocumentAuthors authors,
```
surli, line 74: So that one is not easy to read, it's very compact might worth a bit of refactoring / comment. Also I'm not sure to understand why it's not part of the API? Sounds like a common need no ?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/AbstractBlockSupplierProvider.java
```diff
@@ -0,0 +1,157 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.platform.security.requiredrights.internal;
```
surli, line 20: The package contains already lots of classes and it mixes utility classes (e.g. this one, VelocityUtil, TranslationMessageSupplier, etc) and real required rights analyzers. Feels like it would worth having 2 subpackages to better organize those.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/VelocityUtil.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Utility class for velocity. This class is used to check if a string contains a velocity script.
+ *
+ * TODO move this class to a more appropriate place
+ * @version $Id$
+ */
+public final class VelocityUtil
```
surli, line 30: Outch looks like it needs a big fixme...

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/VelocityUtil.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Checks if a string contains a velocity script.
+     *
+     * @param input the string to check
+     * @return true if the string contains a velocity script, false otherwise
+     */
+    public static boolean containsVelocityScript(String input)
+    {
+        return StringUtils.containsAny(input, "#", "$");
```
surli, line 44: Sounds like it should be a component so we can override it easily if we find better heuristic / regex ?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/SkinExtensionObjectRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.xpn.xwiki.objects.BaseObject;
+
+/**
+ * Required right analyzer for skin extensions.
+ *
+ * @version $Id$
+ * @since 15.9RC1
+ */
+@Component(hints = { "object/XWiki.JavaScriptExtension", "object/XWiki.StyleSheetExtension" })
```
surli, line 45: Those hints are surprising: you never documented that for BaseObject it should have a specific name to be automatically taken into account.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsEditConfirmationChecker.java
```diff
@@ -0,0 +1,97 @@
...
+        XWikiDocument tdoc = (XWikiDocument) context.get("tdoc");
+        if (!this.authorization.hasAccess(Right.EDIT, tdoc.getDocumentReferenceWithLocale())) {
+            return Optional.empty();
+        }
+        if (tdoc.isNew()) {
+            return Optional.empty();
+        }
+        try {
+            RequiredRightsChangedResult analysisResults =
+                this.requiredRightsChangedFilter.filter(tdoc.getAuthors(), this.analyzer.analyze(tdoc));
```
surli, line 87: Ok so AFAIU it's the main entry point right now for performing the analysis, and the analysis is never performed by retrieving all analyzers it's done when traversing the document. Those info should be provided in the javadoc to better understand how to implement analyzers. Also I'm wondering how much the result here should be cached: I mean it feels a bit expansive, and the result is easy to invalidate at each update no?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRight.java
michitux: Yes, good point. I'll add it to the documentation, the use cases are as follows:

* A title that contains `#` or `$` (like "HashTag #XWiki") written by user without script right doesn't necessarily require script right but when saved by a user with script right, we cannot exclude that some code would be executed. So we say the right is optional. Note that there is also no way to escape this `#` or `$` that doesn't require script right, so there is absolutely no other way to write a title "HashTag #XWiki" if the user doesn't have script right.
* Any script might or might not need programming right, this is impossible to analyze. Even if there are checks for programming right in the code, the script might work just fine. In this case, we're also saying that programming right is optional.

Maybe "optional" is the wrong word, if you have a better idea let me know. I think the UI also needs to show these "optional" required rights differently.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRightAnalysisResult.java
michitux: I think we can now make it final, we had some analyzers before that didn't set the entity reference as they didn't know it and the idea was to make it easy to set it later at a different place in the code (think, e.g., macro analyzers being called from an XObject analyzer). The code has been changed in the meantime to ensure that all analyzers know the entity reference.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRight.java
surli: Ok so it feels more like "undecided" than "optional" as "impossible to decide whether or not the right is needed".

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/VelocityUtil.java
tmortagne, line 30: I would move this tool to `xwiki-commons-velocity` (but still internal for now).

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRight.java
surli: or we want to be more explicit something like "needHumanAnalysis"

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRightAnalysisResult.java
manuelleduc: Confirmed, `setEntityReference` is currently unused, making it final.

surli: Now honestly I'd probably only have kept `entityReference` and `requiredRight` final, and I'd have put setters for the others.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsEditConfirmationChecker.java
```diff
@@ -0,0 +1,97 @@
...
+            this.scriptContextManager.getCurrentScriptContext()
+                .setAttribute("analysisResults", analysisResults, GLOBAL_SCOPE);
+            return Optional.of(new EditConfirmationCheckerResult(
+                this.templateManager.executeNoException("security/requiredrights/requiredRightsConfirmationChecker.vm",
+                    false), false));
+        } catch (RequiredRightsException e) {
+            throw new RuntimeException(e);
```
surli, line 94: I think I would have handled it with an actual EditConfirmationCheckerResult returning an error with the root cause of the exception. Basically saying that it could be dangerous to edit since the analysis stopped.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/XWikiDocumentRequiredRightAnalyzer.java
surli: I added a comment on https://github.com/xwiki/xwiki-platform/pull/2470/files#r1363467928 on the topic. It looks good to me to have a defined exception and that each analyzer might throw it. Now I think by default when traversing the doc you should catch those and not stop at first analyzer error: it would be a pity that if adding a custom buggy analyzer it would break the whole analysis.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsEditConfirmationChecker.java
manuelleduc: indeed, good idea, thanks

surli: Check also the discussion on https://github.com/xwiki/xwiki-platform/pull/2470#discussion_r1363447951

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsChangedFilter.java
manuelleduc: Regarding the API, I'm not sure. For now this is only used by internal component and I'm not sure we want to expose this part to extensions.

surli: Ok so better keep it internal indeed for now.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/AbstractBlockSupplierProvider.java
```diff
@@ -0,0 +1,157 @@
...
+            .flatMap(this::renderProperty)
+            .collect(Collectors.toList());
+        // Add the xform class such that hints work.
+        return new DefinitionListBlock(propertyBlocks, Map.of(CLASS_ATTRIBUTE, "xform"));
+    }
+
+    protected Block getCodeBlock(String value)
+    {
+        if (StringUtils.isNotBlank(value)) {
+            return new GroupBlock(List.of(getStringBlock(value)), Map.of(CLASS_ATTRIBUTE, "code box"));
```
tmortagne, line 75: From what I understand #getStringBlock return inline content, so it's a bit strange to put it directly in a GroupBlock which is supposed to contain non inline blocks generally.

vmassol:
> New presentation following my last commit image

It's better, thanks. I still feel that we could do a bit better. Some ideas:
* The fact that page belongs to an extension should be listed at the same level as the security issues. I don't see any reason why the security issues are nested under it. For me, they're both issues to review at the same level.
* The text is bit too generic and I think we can do better. For example, we know here that the issue is that the current user doesn't have script rights while the page's elements listed require script rights. So the text could say something like "Several elements of the page currently have script rights and since you don't have script rights, saving the page would remove script rights from these elements, and prevent executing them, thus loosing features.". More generally, you should gather all types of problems and have an easy to read/understand summary of each type of problem.
* In LiveTableSheetView the first warning says the opposite to what it should say, I think. Since the current user doesn't have script rights, the title content should not have velocity executed on it (while it was the case before).
* The first paragraph is not good because it's way too vague and doesn't help at all the user.
* The displayed UI is still way too complex and technical to be shown to any user with edit rights. As I already commented, I believe we need to show only the summary and recommend an action to the user, and hide all the technical stuff behind an advanced button/link. In short the majority of users should not have to read nor understand the advanced part to take a decision (which is to cancel the edit). Only very advanced users with lots of tech knowledge of XWiki should be tempted to read the advanced part and go ahead with the edit.

For me the UI is not yet ready for the 15.x branch as it's still too complex. We need to put ourselves in the shoes of a simple user seeing this (BTW it's likely that this screen will be shown to simple users more than advanced ones simply by the fact that it's more likely that simple users won't have script scripts than advanced ones).

Thanks!

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsEditConfirmationChecker.java
```diff
@@ -0,0 +1,98 @@
...
+    private ContextualAuthorizationManager authorization;
+
+    @Inject
+    private ScriptContextManager scriptContextManager;
+
+    @Override
+    public Optional<EditConfirmationCheckerResult> check()
+    {
+        XWikiContext context = this.xcontextProvider.get();
+        XWikiDocument tdoc = (XWikiDocument) context.get("tdoc");
```
tmortagne, line 79: Note that we now have `XWikiDocument#CKEY_TDOC` (this code was probably written before).

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsEditConfirmationChecker.java
```diff
@@ -0,0 +1,98 @@
...
+        if (!this.authorization.hasAccess(Right.EDIT, tdoc.getDocumentReferenceWithLocale())) {
```
tmortagne, line 80: The security system does not take into account the locale.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/RequiredRightsEditConfirmationChecker.java
```diff
@@ -0,0 +1,98 @@
...
+            return Optional.empty();
+        }
+        try {
+            RequiredRightsChangedResult analysisResults =
+                this.requiredRightsChangedFilter.filter(tdoc.getAuthors(), this.analyzer.analyze(tdoc));
+            this.scriptContextManager.getCurrentScriptContext()
+                .setAttribute("analysisResults", analysisResults, GLOBAL_SCOPE);
+            return Optional.of(new EditConfirmationCheckerResult(
+                this.templateManager.executeNoException("security/requiredrights/requiredRightsConfirmationChecker.vm",
+                    false), false));
```
tmortagne, line 93: Note that `inline=false` is the default.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/AbstractMacroBlockRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,174 @@
...
+                Block.Axes.ANCESTOR);
+
+        if (metaDataBlock != null && metaDataBlock.getMetaData()
+            .getMetaData(XDOMRequiredRightAnalyzer.ENTITY_REFERENCE_METADATA) instanceof EntityReference) {
+            result = (EntityReference) metaDataBlock.getMetaData()
+                .getMetaData(XDOMRequiredRightAnalyzer.ENTITY_REFERENCE_METADATA);
+        } else {
+            metaDataBlock = source.getFirstBlock(new MetadataBlockMatcher(MetaData.SOURCE), Block.Axes.ANCESTOR);
+            if (metaDataBlock != null) {
+                // FIXME: the local is lost here as the metadata source does not keep the locale when serializing the
```
tmortagne, line 165: "the locale" ?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/DefaultMacroBlockRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,110 @@
...
+import org.xwiki.rendering.macro.MacroExecutionException;
+import org.xwiki.rendering.macro.descriptor.ContentDescriptor;
+import org.xwiki.rendering.macro.script.ScriptMacro;
+import org.xwiki.rendering.transformation.MacroTransformationContext;
+
+/**
+ * Default analyzer for macro blocks. Recurses into the macro content if it is wiki syntax.
+ *
+ * @version $Id$
+ * @since 15.8RC1
```
tmortagne, line 48: Wrong version.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/DefaultMacroBlockRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+@Component
+@Singleton
+@Named(DefaultMacroBlockRequiredRightAnalyzer.ID)
+public class DefaultMacroBlockRequiredRightAnalyzer extends AbstractMacroBlockRequiredRightAnalyzer
+{
+    /**
+     * The id of this analyzer.
+     */
+    public static final String ID = "macro";
```
tmortagne, line 58: Isn't it obvious for something associated to a MacroBlock ? Maybe you should have this one simply being the default `RequiredRightAnalyzer<MacroBlock>` and the other one have @Named("script").

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/DefaultObjectRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,208 @@
...
+ */
+@Component
+@Singleton
+@Named(DefaultObjectRequiredRightAnalyzer.ID)
+public class DefaultObjectRequiredRightAnalyzer implements RequiredRightAnalyzer<BaseObject>
+{
+    /**
+     * The id of this analyzer.
+     */
+    public static final String ID = "object";
```
tmortagne, line 70: Same comment as for `RequiredRightAnalyzer<MacroBloc> having a "macro" role hint`.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/DefaultObjectRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,208 @@
...
+
+    private List<RequiredRightAnalysisResult> analyzeVelocityScriptValue(String value, EntityReference reference,
+        String translationMessage)
+    {
+        List<RequiredRightAnalysisResult> result;
+        if (VelocityUtil.containsVelocityScript(value)) {
+            result = List.of(new RequiredRightAnalysisResult(reference,
+                this.translationMessageSupplierProvider.get(translationMessage),
+                this.stringCodeBlockSupplierProvider.get(value),
+                List.of(
```
tmortagne, line 184: This List should probably be a constant to avoid recreating it all the time.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/XDOMRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+@Component
+@Singleton
+@Named(XDOMRequiredRightAnalyzer.ID)
+public class XDOMRequiredRightAnalyzer implements RequiredRightAnalyzer<XDOM>
+{
+    /**
+     * The id of this analyzer.
+     */
+    public static final String ID = "xdom";
```
tmortagne, line 53: Same "is this role hint really needed" comment.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/XWikiDocumentRequiredRightAnalyzer.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+@Component
+@Singleton
+@Named(XWikiDocumentRequiredRightAnalyzer.ID)
+public class XWikiDocumentRequiredRightAnalyzer implements RequiredRightAnalyzer<XWikiDocument>
+{
+    /**
+     * The id of this component.
+     */
+    public static final String ID = "document";
```
tmortagne, line 56: Same "is this role hint really needed" comment.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/resources/templates/security/requiredrights/requiredRightsConfirmationChecker.vm
```diff
@@ -0,0 +1,94 @@
...
+      </div>
+    </div>
+      #set ($counter = $counter + 1)
+    #end
+  {{/html}}
+  #end
+#end
+#set ($counter = 0)
+#initRequiredSkinExtensions()
+#set ($discard = $xwiki.ssx.use('XWiki.Security.RequiredRights.Code.RequiredRightsResultStyle'))
```
tmortagne, line 74: Would it make more sense to store this css in the JAR, like the template ?

lucaa:
> * The fact that page belongs to an extension should be listed at the same level as the security issues. I don't see any reason why the security issues are nested under it. For me, they're both issues to review at the same level.

Regarding this issue of the UI, we could use 2 separate warning boxes, once for each problem we have, this way the 2 issues are shown on the same level.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/SkinExtensionObjectRequiredRightAnalyzer.java
michitux: I added documentation to `RequiredRightAnalyzer` about these hints.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-default/src/main/java/org/xwiki/platform/security/requiredrights/internal/analyzer/DefaultObjectRequiredRightAnalyzer.java
michitux: I've added constants to `RequiredRight` for common cases.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-requiredrights/xwiki-platform-security-requiredrights-api/src/main/java/org/xwiki/platform/security/requiredrights/RequiredRight.java
michitux: I've renamed it to `manualReviewRequired` with a longer documentation. I hope this is clearer now.

PR #2471, opened 2023-10-14
XWIKI-21433: Generate URLs for editable properties from the property …

AndreasJonsson:
…reference rather than XWiki.currentDocument

  * Resolve the data property reference in the attribute data-property and use this for generating save and get URLs.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/edit/editableProperty.js
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/edit/editableProperty.js
@@ -110,10 +110,21 @@ define('editableProperty', ['jquery', 'xwiki-meta'], function($, xcontext) {
 
   // Actions
 
+  var _propURLGet = function _propURLGet(propertyData, urlType) {
+    var ref = XWiki.Model.resolve(propertyData, XWiki.EntityType.OBJECT_PROPERTY);
+    if (ref) {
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/edit/editableProperty.js
```diff
@@ -110,10 +110,21 @@ define('editableProperty', ['jquery', 'xwiki-meta'], function($, xcontext) {
 
   // Actions
 
+  var _propURLGet = function _propURLGet(propertyData, urlType) {
+    var ref = XWiki.Model.resolve(propertyData, XWiki.EntityType.OBJECT_PROPERTY);
+    if (ref) {
+      var doc = new XWiki.Document(ref.extractReference(XWiki.EntityType.DOCUMENT));
```
vmassol, line 116: This might need a comment to explainwhy you need to check if there's an xobject before extracting the document, and why the fallback is to use the current document's url, wdyt? Thx!

AndreasJonsson: I was just being being defensive not wanting to break anything in the event that the parsing of the reference would fail for some reason that I have overlooked.

But if the attribute does not contain a valid property reference the edit is likely to fail anyway, so in the interest of failing early it is maybe better to just let it fail here?

PR #2473, opened 2023-10-16
XWIKI-21376: Live validation prompts should be presented to users without obtaining focus

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21376
## PR Changes
* Updated live validation to make use of 'aria-live'
* Updated the register form to work properly with the new live validation
## View
[Here is a video demo of the changes brought in this PR](https://up1.xwikisas.com/#BM8CU9zVmcNXnt19z4n09A)
This video was made using NVDA and displays a written down version of the screen reader output. We can see that with those changes, the screen reader gives frequent indications on the form while the user is filling it up.

Without those changes, a screen reader user would not hear any validation (unless they searched for it) to help them fill up the form.

## Notes
This attribute only works correctly when the element is generated from the start. Adding it afterwards would not work. In order to make it work, now, instead of generating and inserting a new span on every update, the only thing we change now on update is the content of this span, and some of its classes.
I decided to use aria-live='polite' instead of aria-live='assertive' since it's not critical to display the changes as fast as possible. We at least wait for the user to stop writing in the field to give the current validation state of this field.


michitux (changes requested):
You're just changing `register_macros.vm` for the output. However, Live Validation is used everywhere. Check for example the title input when creating a page. I think that a) these use cases aren't fixed and b) they might actually be broken.

Sereza7:
Thanks for your help @michitux !

I checked the xwiki organization on GitHub to see where it's used:
* https://github.com/search?q=org%3Axwiki%20LV_validation_message&type=code -> Nothing unexpected
* https://github.com/search?q=org%3Axwiki+LiveValidation&type=code -> Used in the location picker macro (which is itself used in multiple places in the wiki, e.g. page creation)

I updated this last file in https://github.com/xwiki/xwiki-platform/pull/2473/commits/a460be36f30366384fd293e98eeaf92d6de6f3c2 . Here is what the results looks like:
21376-locationPicker

___
Unfortunately, such changes to the templates where livevalidation is used are necessary AFAIK, since the 'aria-live' attribute doesn't work properly on nodes that are dynamically added to the page. We can't just update the .js . Ref. [this stackoverflow topic](https://stackoverflow.com/a/68492641/14899545)

___
Since:
* the changes needed are pretty straight forward (add an aria-live span just after the field to validate)
* AFAIK LiveValidation is not being used in plenty of places and when no changes are made
* only the live validation is broken (no additional issue with the input or page)

I think what's proposed in this PR is an okay solution.

Let me know if you have a better idea about this, or notice more uses of LiveValidation that I missed and would break if we merge this too fast.

michitux (changes requested):
Live Validation is a [documented](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/LiveValidation/) API and I think it is very bad to break this documented API and the [extension](https://extensions.xwiki.org/xwiki/bin/view/Extension/LiveValidation%20Macros) that uses this API. Instead, I think the solution is to still support the old way of not having the message holder in the output. If there is no element for the message in the HTML, the JavaScript code should just create it as before.

How did you reach the conclusion that `aria-live` doesn't work when it isn't part of the static HTML? From reading the [documentation on live regions on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions). Specifically, it says:

> Including an aria-live attribute or a specialized live region role (such as [role="alert"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)) on the element you want to announce changes to works as long as you add the attribute before the changes occur — either in the original markup, or dynamically using JavaScript.

So basically as long as the aria-live attribute is set before the content is updated, it should work. I don't mind having a placeholder in the HTML if you think this is safer for whatever reason, but please don't break the documented Live Validation behavior. Even if it shouldn't be possible to make it fully accessible (which is something I highly doubt), having a non-accessible validation message is better than no validation message.

Sereza7:
> How did you reach the conclusion that aria-live doesn't work when it isn't part of the static HTML? From reading the [documentation on live regions on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) I don't see why this should be the case.

After the [first commit](https://github.com/xwiki/xwiki-platform/pull/2473/commits/00288c44265a95b7c1a6ea05bf4404468a0e13f6) I tried this PR out with a screen reader and validation wouldn't be read properly. I thought this was an issue of accessibility tree updating time, so instead of furthering in a hacked solution, I decided to change the way the macro worked.

However, I tested it again this afternoon and somehow this time it reacts correctly. I don't know where I made a mistake in my previous tests.
I updated the PR, to include only non breaking changes.

Here is a new quick demo I got from those changes, displaying the screen-reader properly working.
https://up1.xwikisas.com/#nUCmf_pSpv6YrPTikcgsyg

PR #2475, opened 2023-10-17
XWIKI-21009: LiveData alternative to drag controls

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21009

## PR Changes
* Added listeners on livedata controls to make them keyboard compatible
* Improved semantics inside the livedata headers
* Updated style to improve readability and avoid 'appear on hover' buttons
## View
21009-afterPR1[
Here is a video preview highlighting the keyboard interactions](https://up1.xwikisas.com/#yOyBJMf2iJoi8lAVrKYDwA).

## Notes
* I decided to use the `btn-default` and `btn-xs` mixins to style the livedata column controls. This way, they are small, clearly grouped together (like the action buttons at the top of documents) and deemphasized with low saturation colors. 
* The choice for the icon in the resize-handle is not perfect, let me know if you can find an icon that's more appropriate for this use case in the icon themes :)
* Even though I made some effort to make those buttons not stand out, they are still quite cumbersome. Maybe a more in depth solution would be to add a toggle for column controls in the livedata options.

\<EDIT> **Added this last section on the 12th of Jan 2024**\</EDIT>
# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * stable-15.10.x (The scope of the changes is quite limited, for the most part it's just style. I already went through a lot of manual testing so I think it's pretty safe to backport)


Sereza7:
Need to make those changes to the minimum column width only affect some users.
An idea is to take advantage of the extra accessibility user preference.

Sereza7:
[Here is a video preview](https://up1.xwikisas.com/#xF5g0nTr9mZzyaNn9qvCGw) of the changes introduced in https://github.com/xwiki/xwiki-platform/pull/2475/commits/868e29f9fd1d4660c3f9b1c54bfa83d19bf52962 :)
Now the resize handle is a small bar at the end of each column.

tkrieck:
Moving my comment from Jira to here for better visibility 

I quite liked this last preview. The division between headers in tables is a very common affordance for resizing columns in other apps. On the visual aspect, we could have round corners and a hover style for better presentation, but I know you are still working on it.

You've mentioned the target size of this element, is it possible to have a target size and visual sizes separate? This way we could have something close to 24 units of target size but without a 24 bar or icon? 

For the move handle, what do you think of this icon https://fontawesome.com/v4/icon/arrows or this one https://fontawesome.com/v4/icon/arrows-h ? The bars can convey the notion that there's a menu there.

Sereza7:
Thanks for your quick feedback! :sparkles: 
> On the visual aspect, we could have round corners and a hover style for better presentation, but I know you are still working on it.

I thought round corners on a vertical bar would make it look messy. This bar is currently made with the border of an empty div, changing border-radius on it doesn't do anything here. I can update it to allow border but I think in any case it'll be barly noticeable since the rectangle is already just a couple pixels wide. IMO it's quite difficult to see/understand enough so making it pop out a bit with a sharp border is a plus.

> You've mentioned the target size of this element, is it possible to have a target size and visual sizes separate? This way we could have something close to 24 units of target size but without a 24 bar or icon?

We don't want the target areas to overlap, so if we have a 24px target size, it wouldn't allow to save space compared to a 24px size button. Maybe we could hack the area so that it goes over right next to the next column name but I don't think this'd be a clean solution. I think leaving a target size that's too low is better than have hacks that blur column delimitations (even if not visually). It's a complex subject and it's not as critical as the main point of this PR, so I don't mind if we overlook it for now. Note that grabbing this bar with a mouse can be difficult for pretty much anyone since it's so small in width.

> For the move handle, what do you think of this icon https://fontawesome.com/v4/icon/arrows or this one https://fontawesome.com/v4/icon/arrows-h ? The bars can convey the notion that there's a menu there.

:+1: I don't like those bars either.
It'd be nice, however as of now, those two are not in [our standard iconset for xwiki-platform](https://design.xwiki.org/xwiki/bin/view/Proposal/XWiki%20Icon%20Set) so we should avoid using them. There's [an idea to add some icons to the theme](https://design.xwiki.org/xwiki/bin/view/Proposal/FontAwesomeNewIcons), but I think it's at a stall for now. I'll ping Adina who's the creator of this idea, and ask if I can make it into a proposal so that eventually we can use those icons :)
If you find another icon that would suit this button more than the menu bars from the icon theme, let me know ;)

Sereza7:
I updated the presentation to have a cleaner and more consistent style for both buttons in https://github.com/xwiki/xwiki-platform/pull/2475/commits/6cee4b4aab1362c61644e7e5d677134b242c1eb1 .
[Here is a video demo of the updated display](https://up1.xwikisas.com/#unlX-a6ohZUY0_AO1NGbXw).

I decided to dumb-down the visuals of the button when the column is not focused, and make it more visible when it's focused. This way, the buttons respect contrast when the user is close to it.
I also replaced the former icon `text_align_justify` to `right`, which at least provides the intention of movement a bit better.

michitux (changes requested):
The resize handle seems to be in the wrong place as it isn't centered between the columns and the space between it and the reorder handle is way too small (or, to be precise, there seems to be no space). 

Also, this re-introduces [XWIKI-19449](https://jira.xwiki.org/browse/XWIKI-19449). In my opinion, column re-ordering is an absolutely minor feature that is probably almost never used. We shouldn't reserve so much space for such a minor feature. I think I would be in favor of just removing the reorder handles as it seems hard to find a proper place for them. Columns can also be reordered from the properties panel. Of course, it would be interesting to understand if column reordering is in fact used before we remove the easy way to access it.

I just got an idea for an alternative: What about just re-ordering columns when you drag the column name without having an extra icon, just mentioning it in the title? And of course then also with the keyboard events on the column name. In case you're worried this is not discoverable enough, I think this is okay. First, this is a feature most users don't need, they don't really loose anything. Second, it is also available in the properties panel and there it is easier to discover, and we could also permanently display the handles there.

In case you're worried that your work on the reorder handle was in vain: you can probably port your code for the reorder handle to the properties panel that has exactly the same reorder handles, also not accessible from what I can see.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -58,45 +58,49 @@
...
           :title="$t('livedata.action.reorder.hint')"
-          @click.stop
+          @keydown.left="keyboardDragNDropLeft($event)"
+          @keydown.right="keyboardDragNDropRight($event)"
         >
-          <XWikiIcon :icon-descriptor="{name: 'text_align_justify'}"/>
-        </div>
-        <!-- Property Name -->
-        <span class="property-name">{{ property.name }}</span>
+          <XWikiIcon :icon-descriptor="{name: 'right'}"/>
```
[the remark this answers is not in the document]
Sereza7: I proposed introducing the arrows icon as a part of https://github.com/xwiki/xwiki-platform/pull/2593 .
I proposed to name it `reposition` (since `move` was already taken for another definition of move).
I updated this PR to match these changes in https://github.com/xwiki/xwiki-platform/pull/2475/commits/2aae176b07f6b78d2cb95f73237eccab0c9ee512

Sereza7:
> I just got an idea for an alternative: What about just re-ordering columns when you drag the column name without having an extra icon, just mentioning it in the title? And of course then also with the keyboard events on the column name. In case you're worried this is not discoverable enough, I think this is okay. First, this is a feature most users don't need, they don't really loose anything. Second, it is also available in the properties panel and there it is easier to discover, and we could also permanently display the handles there.

I didn't think about it because it seemed to me like there's clearly be overlapses in listeners and behaviors (the column name is already clicked to filter with that column), but from a first test, it seems like the bahavior is working fine (ofc some style is messed up but that's only a couple trivial fixes) Thanks for this idea! :D

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -181,6 +185,34 @@ export default {
...
     },
+    
+    keyboardDragNDropRight (e) {
+      let handles = e.currentTarget.closest('tr').querySelectorAll('.handle');
+      let oldIndex = Array.from(handles).indexOf(e.currentTarget);
+      let newIndex = oldIndex + 1;
+      if (newIndex < handles.length) {
+          this.logic.reorderProperty(oldIndex, newIndex);
+      } else {
+          this.logic.reorderProperty(oldIndex, 0);
```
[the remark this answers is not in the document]
Sereza7: I don't think having content wrap around is unexpected behavior. When I am selecting the rightmost column and ask to move it to the right, I expect to have a shortcut to put it back to the leftmost position. This is just a convenience that can benefit power users: when I have 20 columns and I want to put the 13th in position 1, I could just go right 8 times and loop around, instead of going left 12 times. 
If I want to put the 20th column in 1st place, instead of going 20 times to the left, I can move it once to the right. Personally I think this is great QoL for this feature.

It's not the main point of the PR so I don't mind reverting it if we don't agree :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -181,6 +185,34 @@ export default {
...
+    },
+
+    keyboardDragNDropLeft (e) {
+      let handles = e.currentTarget.closest('tr').querySelectorAll('.handle');
+      let oldIndex = Array.from(handles).indexOf(e.currentTarget);
+      let newIndex = oldIndex - 1;
+      if (newIndex > -1) {
+          this.logic.reorderProperty(oldIndex, newIndex);
+      } else {
+          this.logic.reorderProperty(oldIndex, handles.length - 1);
```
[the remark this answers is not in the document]
Sereza7: Same as [above](https://github.com/xwiki/xwiki-platform/pull/2475#discussion_r1428066291).

Sereza7:
## PR Changes
* Removed the reorder button, and put its behavior on the property name button
* Updated style to fit this new solution
* Updated translations to fit this new solution
* Factorized some code between mouse and keyboard events
## View
21009-compactVersion

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -91,8 +91,8 @@ livedata.pagination.last=Last page
 
 livedata.action.refresh=Refresh
 livedata.action.addEntry=Add entry
-livedata.action.reorder.hint=Drag to reorder
-livedata.action.resizeColumn.hint=Drag to resize. Double click to reset.
+livedata.action.columnName.hint=Click to sort the entries based on this property. Drag or use arrow keys to reorder.
```
michitux, line 94: I think we should have two titles: One when the column is sortable, and another one when it isn't.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -182,13 +176,32 @@ export default {
       }
     },
 
-    resizeColumnInit (e) {
+    keyboardDragNDrop(e,deltaIndex) {
```
michitux, line 179:
```suggestion
    keyboardDragNDrop(e, deltaIndex) {
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -218,6 +250,39 @@ export default {
...
+      const offsetX = amount;
+      // Resize left column
+      const leftColumnWidth = leftColumnBaseWidth + offsetX;
+      leftColumn.style.width = `${leftColumnWidth}px`;
+
+      // Resize right column
+      if (rightColumn) {
+        const rightColumnWidth = rightColumnBaseWidth - offsetX;
+        rightColumn.style.width = `${rightColumnWidth}px`;
+      }
```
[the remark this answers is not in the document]
Sereza7: Removed duplicated code in https://github.com/xwiki/xwiki-platform/pull/2475/commits/6cedc2ee738ede3116d5b93fd57856d87e8eb9cc

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
```diff
@@ -236,44 +301,62 @@ export default {
...
+  border: 0;
+  color: @btn-default-border;
 }
+
+.layout-table .column-name:focus-within .handle,
 .layout-table .column-name:hover .handle {
-  opacity: 0.8;
-  transition: opacity 0.2s;
+    color: @text-muted;
+    border: 1px solid @text-muted;
```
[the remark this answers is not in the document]
Sereza7: In the latest version, this button is not a part of the UI anymore.

michitux (changes requested):
That looks much better in my opinion, thank you! Regarding the look, I have the impression that the sort icons are now aligned to the bottom of the column name while previously they were a bit more centered. Is that correct? I think centered looks better.

Also, a comment regarding the title, not all columns are sortable.

Apart from that, this looks good to me.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2475/commits/ff4685a27336fd8a3a49d875055980339f354d42 :+1:

Sereza7:
> Regarding the look, I have the impression that the sort icons are now aligned to the bottom of the column name while previously they were a bit more centered. Is that correct? I think centered looks better.

Yup, good catch, this was the result of a collision with handle style from another file.
Addressed in https://github.com/xwiki/xwiki-platform/pull/2475/commits/ce25b2ee06bc5304539a87e585e95bd321a11a74 :+1: 

Here is the look afterwards:
21009-centeredSortIcons

michitux:
Great. I just measured again with a (virtual) [ruler](https://apps.kde.org/kruler/) on the screen and I noticed that these resizes handles are still not fully centered between the columns. Do you believe this could be improved a little? Apart from that I'm now really happy with this, I think this makes Live Data much cleaner while at the same time more accessible. :+1:

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTableHeaderNames.vue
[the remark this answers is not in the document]
michitux: To me, it matches the mouse actions, you also cannot move the column further to the right with the mouse. But I don't mind that much.

michitux (approved):

Sereza7:
> Great. I just measured again with a (virtual) [ruler](https://apps.kde.org/kruler/) on the screen and I noticed that these resizes handles are still not fully centered between the columns. Do you believe this could be improved a little? Apart from that I'm now really happy with this, I think this makes Live Data much cleaner while at the same time more accessible. 👍

Thanks! :D
I intended for them not to be centered, but after looking at it from a user perspective, it does not make sense indeed. I supposed we wanted to tie them to the left column. Semantically it's a child of the left column, but functionally, it's the same role for both left and right column. Except for the highlighting triggered only on the rightside resize-handle, a regular user wouldn't know that the implementation is assymetrical, and it's a pointless information for every use case.

I added a negative margin in https://github.com/xwiki/xwiki-platform/pull/2475/commits/71822677a12403ef75b8e79072bc120c0365a5ed, and here is what it looks like:
21009-centeredResizeHandles
Then noticed that there was an issue with the last one getting over the expected space, and corrected it in https://github.com/xwiki/xwiki-platform/pull/2475/commits/a4ff595fa2f332c1f4b60d5d78916f0cf740b816, which gave the final result:
21009-lastResizehandleCorrectmargin

Sereza7:
[Here is a video preview of the updated livedata drag and drop feature](https://up1.xwikisas.com/#p6kEchxZb2t39obQGwSyug).

lucaa:
The last version looks good (better than the previous one).

However, the vertical controls for handling column width are a little misleading as they look a lot like separators and the livedata cells themselves don't have separators. Also, because they look like separators, they may not be discoverable...

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/resources/ApplicationResources.properties
```diff
@@ -91,8 +91,9 @@ livedata.pagination.last=Last page
 
 livedata.action.refresh=Refresh
 livedata.action.addEntry=Add entry
-livedata.action.reorder.hint=Drag to reorder
-livedata.action.resizeColumn.hint=Drag to resize. Double click to reset.
```
michitux, line 95: Sorry for noticing this only now, but this is not how we handle translation updates, see https://dev.xwiki.org/xwiki/bin/view/Community/L10N/Conventions/#HUpdatingtranslations - the old keys need to be deprecated instead of being removed, and new keys need to be added for the updated messages.

Sereza7: I updated translation to follow our translation update practices. 
I tried to be as thorough as possible, but I can't provide correct translations for most languages (just gave some for french). Instead I just took the former value, which had part of what should be in the value, and went with that. This is not perfect but I guess it's still better than putting english.

I looked at https://github.com/xwiki/xwiki-platform/blob/539f97b0d6f8f9e049abe2a0c5321e41a137ffa0/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties#L21C1-L43C80 and tried to follow the pattern described and used there.

Changes were made in https://github.com/xwiki/xwiki-platform/pull/2475/commits/e9a58de555032dd6282d1361895eb99b644b6d3a (changing the translations) and https://github.com/xwiki/xwiki-platform/pull/2475/commits/932801e2e7440450c6c4c1c02f42f64ade0df843 (adding the commentary section to explain organization of the translations).

michitux: You should normally not update other languages. I think the sentence 

> When deprecating a key in favor of a different one, you should try to see if some of the existing translations can be copied over as well.

Just says that if you, e.g., renamed an XClass property and now need to rename the translation keys as well, you should also see if you can copy the translations such that they don't need to be translated again. However, I think you shouldn't copy "wrong" translations as then translators won't see the strings as missing and cannot update them.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2475/commits/8fb8cb945fa970eda65207236cf5885eab05d5fd :+1:

PR #2485, opened 2023-10-20
XWIKI-21105: Drop the panel-default-text color from the color theme and use the content colors for all panel content, to facilitate contrast

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21105
## PR Changes
* Removed `panel-default-text` from the color theme template and from all the bundled color themes.
* Updated the value of this variable in the advanced section of the color themes where it was used, in order for it to follow regular text color

## Notes
The variable is still available in the color theme, however this PR makes it impossible to change it from the edit UI of a theme, and removes default values (most of the time hard coded for no reason).

See https://forum.xwiki.org/t/color-theme-panel-content/12818 for a discussion about removing this variable from the UI.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-theme/xwiki-platform-flamingo-theme-bootswatch/src/main/resources/FlamingoThemes/Cerulean.xml
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-theme/xwiki-platform-flamingo-theme-bootswatch/src/main/resources/FlamingoThemes/Cerulean.xml
@@ -753,7 +753,6 @@
         <validationRegExp/>
         <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
       </panel-bg>
-      
       <text-color>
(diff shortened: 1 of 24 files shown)


lucaa:
thanks for the work, @Sereza7 .

However, do you think it's possible to also remove the usages of this color (replace them with @text-color)? are there many?

lucaa:
Also,  I think it would be interesting to have some note on the color theme editor about the colors that will be used for the foreground items of the panels (now that the "foreground color" that was next to the panel-bg has disappeared from the UI).

Sereza7:
```
thanks for the work, @Sereza7 .

However, do you think it's possible to also remove the usages of this color (replace them with @text-color)? are there many?
```
From what I understood from our discussion, we want to leave this color in the theme, even if there's no UI to change it, so that users can still use it from advanced parameters.

That being said, with https://github.com/search?q=org%3Axwiki+%40panel-default-text%3B&type=code we see that there is only a couple uses of this variable, and completely removing it would not make it much more difficult to update style from advanced settings IMO.

PR #2487, opened 2023-10-20
XWIKI-21447: Incorrect results when typing slowly in filter fields

raphj:
This PR is somewhat difficult to test, but you can "feel" the code by setting a breakpoint on the abort line and setting network throttling to GPRS or something badly slow.

Convenient test link: http://localhost:8080/xwiki/bin/view/Main/AllDocs?minify=false&useLiveData=true#|t=alldocs&p=1&l=15&s=doc.location&d=asc&doc.title=f

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
@@ -21,14 +21,14 @@
 define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
   'use strict';
 
-  var baseURL = module.config().contextPath + '/rest/liveData/sources/';
+  const baseURL = module.config().contextPath + '/rest/liveData/sources/';
 
   let entriesRequest;
 
-  var getEntries = function(liveDataQuery) {
-    var entriesURL = getEntriesURL(liveDataQuery.source);
+  function getEntries(liveDataQuery) {
+    const entriesURL = getEntriesURL(liveDataQuery.source);
 
-    var parameters = {
+    const parameters = {
       properties: liveDataQuery.properties,
       offset: liveDataQuery.offset,
       limit: liveDataQuery.limit
@@ -68,7 +68,7 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
       .finally(cleanupRequest.bind(null, entriesRequest));
   };
 
-  var cleanupRequest = function(requestToClean) {
+  function cleanupRequest(requestToClean) {
     // We reset the request object to null for two reasons:
     // - avoid keeping an object we don't need anymore in memory, preventing it from being GC'd
     // - make sure we don't attempt to abort a request that already terminated.
@@ -82,13 +82,13 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
     }
   };
 
-  var getEntriesURL = function(source) {
-    var entriesURL = baseURL + encodeURIComponent(source.id) + '/entries';
+  function getEntriesURL(source) {
+    const entriesURL = baseURL + encodeURIComponent(source.id) + '/entries';
     return addSourcePathParameters(source, entriesURL);
   };
 
   function addSourcePathParameters(source, url) {
-    var parameters = {
+    const parameters = {
       // Make sure the response is not retrieved from cache (IE11 doesn't obey the caching HTTP headers).
       timestamp: new Date().getTime(),
       namespace: `wiki:${XWiki.currentWiki}`
@@ -104,7 +104,7 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
     const url = `${baseURL}${encodedSourceId}/entries/${encodedEntryId}/properties/${encodedPropertyId}`
     return addSourcePathParameters(source, url);
   }
-  
+
   function getEntryURL(source, entryId) {
     const encodedSourceId = encodeURIComponent(source.id);
     const encodedEntryId = encodeURIComponent(entryId);
@@ -112,7 +112,7 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
     return addSourcePathParameters(source, url);
   }
 
-  var addSourceParameters = function(parameters, source) {
+  function addSourceParameters(parameters, source) {
     $.each(source, (key, value) => {
       if (key !== 'id') {
         parameters['sourceParams.' + key] = value;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -51,7 +53,16 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
     parameters.sort = liveDataQuery.sort.map(sort => sort.property);
     parameters.descending = liveDataQuery.sort.map(sort => sort.descending);
 
-    return Promise.resolve($.getJSON(entriesURL, $.param(parameters, true)).then(toLiveData));
+    if (entriesRequest) {
+      // Avoid ordering issues by cancelling outdated requests
+      // See https://jira.xwiki.org/browse/XWIKI-21447
```
manuelleduc, line 58: Probably better to explain with a few sentences the issue here, instead of pointing to a jira id.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -93,6 +104,7 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
   };
 
   var toLiveData = function(data) {
+    entriesRequest = null;
```
manuelleduc, line 107: I'm not sure to understand why a reset to null is needed here. It might deserve a comment.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
vmassol: Yes, our comments are supposed to be self standing. An external link can be provided but if that links becomes broken in the future, the reader of the code need to understand what it was about and especially WHY that code was needed. Thx!

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -102,7 +123,7 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
   var addEntry = function(source, entry) {
     return Promise.resolve($.post(getEntriesURL(source), entry).then(e => e.values));
   };
-  
+
```
manuelleduc, line 126: Those small changes can be part of another "misc" commit.

manuelleduc (approved):
lgtm. We'll need to wait for 15.9-rc-1 to be released before merging on master.

manuelleduc:
> This needs to be merged without squash (because there are two distinct commits, one that removes useless spaces).

Nice :) 

Small remark though. The message of the misc commit must be `[Misc] Clean up spaces` and not `Misc: clean up spaces`

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -93,6 +110,10 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
   };
 
   var toLiveData = function(data) {
+    // We reset the request object to null for two reasons:
+    // - avoid keeping an object we don't need anymore in memory, preventing it from being GC'd
+    // - make sure we don't attempt to abort a request that already terminated.
+    entriesRequest = null;
```
mflorea, line 116: This is not the right place to reset the request. ``toLiveData`` should not be aware that there is a request. Its sole responsibility is to adapt the given JSON to the JSON expected by Live Data. The request should be reset in ``getEntries`` which should be the only function handling the request.

Moreover, ``toLiveData`` is called only on a successful request. It probably makes sense to reset even the failed requests, which is another argument for moving this to ``getEntries`` in a request "complete" callback.

Finally, note that promise / deferred callbacks are **always async**. Try this in the JavaScript console for instance:

```
jQuery.Deferred().resolve().then(function() {console.log('after')});console.log('before');
```

This means that when ``toLiveData`` is called you could have another (different) request already sent, and thus you would nullify this new request. Even if you move this to ``getEntries`` you should make sure that you reset the request only if it matches the request on which the callback was added. This can be done for instance by keeping a request count (ID) and resetting the request only if it's the last one (its ID equals the request count).

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -23,6 +23,8 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
 
   var baseURL = module.config().contextPath + '/rest/liveData/sources/';
 
+  var entriesRequest = null;
```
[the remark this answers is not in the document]
mflorea: > Is it fine to use let here where all the other declarations use var?

We should move to ``let`` / ``const`` progressively. So whenever we touch code that use ``var`` we should refactor it.

> (I also set null explicitly because I like that it's explicit but I don't care much either way - let vs var does not bother me too much, though I indeed prefer let usually)

I'm fine with this.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
```diff
@@ -53,22 +53,33 @@ define('xwiki-livedata-source', ['module', 'jquery'], function(module, $) {
...
+    // See https://jira.xwiki.org/browse/XWIKI-21447
+    entriesRequest?.abort();
     entriesRequest = $.getJSON(entriesURL, $.param(parameters, true));
 
-    return Promise.resolve(entriesRequest.then(toLiveData));
+    return Promise.resolve(entriesRequest.then(toLiveData))
+      .finally(cleanupRequest.bind(null, entriesRequest));
+  };
+
+  var cleanupRequest = function(requestToClean) {
```
mflorea, line 71:
```suggestion
  const cleanupRequest = function(requestToClean) {
```

or simply

```suggestion
  function cleanupRequest(requestToClean) {
```

mflorea (approved):
Looks good to me. Only one minor comment.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/webjar/liveDataSource.js
mflorea: > I figured I could bind a clean up function to the current request object, and if it is equal to the stored request by ref, there's no in flight request coming and we can nullify it.

Clever! I like it.

PR #2494, opened 2023-10-23
XWIKI-21451: The logging form bottom rounded corners have a blank gap

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21451

- Technically a (very old) regression between 7.1 and 7.2, by https://github.com/xwiki/xwiki-platform/commit/8b065192c4f26e13f9d4bbbdf34090055c18e218

**Before**

image

**After**

image


Sereza7:
This solution seems okay :+1: 

Is there a specific reason why you preferred this over `overflow:hidden` on the parent?
See this [stack overflow discussion](https://stackoverflow.com/questions/65498372/background-color-overflowing-from-parent)
I checked and it seems to be working just as well :)

PR #2541, opened 2023-10-26
XWIKI-21475: Hide font awesome icons from screen readers

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21475
## PR Changes
* Added aria-hidden attribute to Font Awesome icons
## Tests
AFAIK there's not much tests on nodes `.fa`, and the only test does not fail because of the changes in this PR
## View
This PR does not change the visuals, but on the screenshot below, we can see that the accessibility tree displayed by the Firefox inspector is clean (on the opposite to before this PR where it had a node with the code of the icon inside).
21475-afterPR

## Note
Related to the PR: https://github.com/xwiki/xwiki-platform/pull/2540


xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
```diff
@@ -37,370 +37,370 @@
   <syntaxId>plain/1.0</syntaxId>
   <hidden>true</hidden>
   <content>## General settings
-xwiki.iconset.type = font
-xwiki.iconset.ssx = IconThemes.FontAwesome
-xwiki.iconset.render.wiki = {{html clean="false"}}&lt;span class="fa fa-$icon"&gt;&lt;/span&gt;{{/html}}
-xwiki.iconset.render.html = &lt;span class="fa fa-$icon"&gt;&lt;/span&gt;
-xwiki.iconset.icon.cssClass = fa fa-$icon
+    xwiki.iconset.type = font
```
michitux, line 40: The new spaces in front of every line look wrong to me, is this really a good idea? Won't this space end up in the page content?

michitux (changes requested):
In general, this looks good, but the whitespace changes seem wrong.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
Sereza7: :+1: Even if this did not change page content, we should avoid changing the codestyle of the whole file when updating just a line. I should have checked the diff properly before opening the PR.
I fixed it in https://github.com/xwiki/xwiki-platform/pull/2541/commits/5cb34b044d7925768a37134d8f13ea30b8495dab and https://github.com/xwiki/xwiki-platform/pull/2541/commits/7894f798ff141f7113aeee52a1345c2a711e1b33
The diff looks clean now. :)

michitux (approved):

PR #2580, opened 2023-11-06
[Misc] Fix double quote in French startup message

raphj:


xwiki-platform-tools/xwiki-platform-tool-jetty/xwiki-platform-tool-jetty-listener/src/main/resources/org/xwiki/tools/jetty/listener/messages_fr.properties
```diff
@@ -17,7 +17,7 @@
 # Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 # 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 # ---------------------------------------------------------------------------
-jetty.startup.notification=Le serveur est d\u00E9marr\u00E9. Vous pouvez maintenant acc\u00E9der \u00E0 votre wiki en entrant l''adresse {} dans votre navigateur internet.
+jetty.startup.notification=Le serveur est d\u00E9marr\u00E9. Vous pouvez maintenant acc\u00E9der \u00E0 votre wiki en entrant l'adresse {} dans votre navigateur internet.
```
surli, line 20: You're sure about that one? AFAIR we escape those characters when there's a parameter.

surli: Now I have a doubt indeed for this one as it's a translation for jetty, and it might not use MessageFormat.

surli: btw @raphj you could have fixed it directly through l10n fyi: https://l10n.xwiki.org/translate/xwiki-platform/jetty-listener/fr/?checksum=d905050c2cd2c75e

tmortagne: > you could have fixed it directly through l10n

Yes the resource seems to have the right type (meaning no the XWiki properties file type) on l10n, so I would assume it works well (and if it does not then we need to fix it).

surli: Done with https://github.com/xwiki/xwiki-platform/pull/2588 I merged adding you as co-author @raphj.

PR #2581, opened 2023-11-06
XWIKI-21523: Fields in form of WYSIWYG configuration are not accessible

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21523
## PR Changes
* Added labels for the three elements in the form that failed because no accessible name.

## View
21523-afterPR

We can see that this PR properly adds the relation between the label and the field (here, only the 'advanced' textarea field is showcased).


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/ConfigSheet.xml
```diff
@@ -218,15 +218,19 @@ require(['jquery', 'xwiki-ckeditor', 'bootstrap-select'], function($, ckeditorPr
   var commaSeparator = /\s*,\s*/;
   var enhanceConfig = function(editor) {
     var availablePlugins = editor.config.plugins.split(commaSeparator);
-    enhanceList($('#CKEditor\\.ConfigClass_0_removePlugins'), availablePlugins);
+    enhanceList($('#CKEditor\\.ConfigClass_0_removePlugins'), availablePlugins, 'availablepluginsselect');
```
michitux, line 221: I'm not aware of any naming convention for ids, but I think it would be nice to use at least some way to differentiate words like camelCase.

Sereza7: You are right. I don't know of one either.
When checking the string `id="` in the codebase, it prompted a lot of things, but the most regularly used convention for IDs was camelCase.

I created a forum proposal to add a HTML codestyle appropriate for this situation: https://forum.xwiki.org/t/html-codestyle-proposal-conventions-for-ids-and-classes/13619

surli: So apparently the proposal went for the kebab-case for ids so I guess you need to fix that now :)

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2581/commits/8599be3eb3639c067095378f06bbdd7431bf0844 :+1:

PR #2579, opened 2023-11-06
XWIKI-21475: Hide font awesome icons from screen readers

douakli:
Jira: https://jira.xwiki.org/browse/XWIKI-21475

This PR fixes a Syntax Error in the `TourCode/TourJS` and `XWiki/Notifications/Code/NotificationsDisplayerUIX` jsxs.
These jsxs expect no single quote in the rendered icon code but allows double quotes.

See the following tracebacks:

```
nov. 06, 2023 3:19:50 PM com.google.javascript.jscomp.LoggerErrorManager printSummary

AVERTISSEMENT: 1 error(s), 0 warning(s)

2023-11-06 15:19:50,104 [qtp422392391-607 - http://localhost:8081/xwiki/bin/jsx/XWiki/Notifications/Code/NotificationsDisplayerUIX?language=en&docVersion=1.1] ERROR c.x.x.w.s.JsExtension          - Error at line [61], column [52]: [Pars
e error. ',' expected]  
nov. 06, 2023 3:19:51 PM com.google.javascript.jscomp.LoggerErrorManager println

GRAVE: xwiki:TourCode.TourJS:44:120: ERROR - [JSC_PARSE_ERROR] Parse error. ',' expected

  44|     var button = $('<button id="tourResume" class="btn btn-default btn-xs"><span class="fa fa-info-circle" aria-hidden='true'></span> Show tour</button>').appendTo(buttonContainer);
```

```
nov. 06, 2023 3:19:51 PM com.google.javascript.jscomp.LoggerErrorManager printSummary

AVERTISSEMENT: 1 error(s), 0 warning(s)

2023-11-06 15:19:51,771 [qtp422392391-606 - http://localhost:8081/xwiki/bin/jsx/TourCode/TourJS?language=en&docVersion=1.1] ERROR c.x.x.w.s.JsExtension          - Error at line [44], column [120]: [Parse error. ',' expected]  
nov. 06, 2023 3:19:51 PM com.google.javascript.jscomp.LoggerErrorManager println

GRAVE: xwiki:XWiki.Notifications.Code.NotificationsDisplayerUIX:61:52: ERROR - [JSC_PARSE_ERROR] Parse error. ',' expected

  61|       .html('<span class="fa fa-trash" aria-hidden='true'></span>&nbsp;Clear All')```

PR #2582, opened 2023-11-07
XWIKI-21475: Hide font awesome icons from screen readers

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21475
## PR Changes
* Added escaping in scripts that failed because of missing escaping of the icons
## Note
This is a complementary fix to the one proposed in https://github.com/xwiki/xwiki-platform/pull/2579. It can still be merged independantly though.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
@@ -322,7 +322,7 @@ Get the actual notifications.
     }
     var markAllReadButton = $('&lt;a href="#"&gt;')
       .addClass('notification-event-clean')
-      .html('$services.icon.renderHTML("trash")&amp;nbsp;$escapetool.xml($escapetool.javascript($services.localization.render('notifications.menu.clear')))')
+      .html('$escapetool.xml($services.icon.renderHTML("trash"))&amp;nbsp;$escapetool.xml($escapetool.javascript($services.localization.render('notifications.menu.clear')))')
       .on('click', function (event) {
         // Avoid the menu closing
         event.preventDefault();
--- a/xwiki-platform-core/xwiki-platform-tour/xwiki-platform-tour-ui/src/main/resources/TourCode/TourJS.xml
+++ b/xwiki-platform-core/xwiki-platform-tour/xwiki-platform-tour-ui/src/main/resources/TourCode/TourJS.xml
@@ -172,7 +172,7 @@ require(['jquery', 'xwiki-meta'], function ($, xm) {
     // Create the popover
     var popover = $('&lt;div class="popover-content"&gt;#tr("tour.popover.show.hint")&lt;/div&gt;').appendTo(buttonContainer);
     // Create the button that will start the tour again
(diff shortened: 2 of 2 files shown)


manuelleduc:
Afaiu https://github.com/xwiki/xwiki-platform/pull/2579 is not needed with your fix

Sereza7:
I think it doesn't hurt to merge https://github.com/xwiki/xwiki-platform/pull/2579 , just to make sure we don't break custom code and extensions where the same pattern without escaping might have been used.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
```diff
@@ -322,7 +322,7 @@ Get the actual notifications.
     }
     var markAllReadButton = $('&lt;a href="#"&gt;')
       .addClass('notification-event-clean')
-      .html('$services.icon.renderHTML("trash")&amp;nbsp;$escapetool.xml($escapetool.javascript($services.localization.render('notifications.menu.clear')))')
+      .html('$escapetool.xml($services.icon.renderHTML("trash"))&amp;nbsp;$escapetool.xml($escapetool.javascript($services.localization.render('notifications.menu.clear')))')
```
mflorea, line 325: If we're rendering the icon as HTML it means we need that HTML. Moreover, we're injecting that rendered icon HTML using jQuery, so my understanding is that the rendered icon is supposed to be used as HTML not as plain text, so I don't see why you would escape the HTML. What you need to JavaScript escaping, to make sure the output of the Velocity code doesn't break the JavaScript code.

```suggestion
      .html($jsontool.serialize("$services.icon.renderHTML('trash')&amp;nbsp;$escapetool.xml($services.localization.render('notifications.menu.clear'))"))
```

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationsDisplayerUIX.xml
manuelleduc: +1
I've tested the PR on a running instance and this is currently not fixing the issues with the notif and the tour.
Also, I don't know if this is related but the document tree can display strangely from time to time:
image

Sereza7: The tree fails to open at

> +1 I've tested the PR on a running instance and this is currently not fixing the issues with the notif and the tour. Also, I don't know if this is related but the document tree can display strangely from time to time: image

This UI issue is not related to this regression, see https://github.com/xwiki/xwiki-platform/pull/2583 :)

Sereza7: > If we're rendering the icon as HTML it means we need that HTML. Moreover, we're injecting that rendered icon HTML using jQuery, so my understanding is that the rendered icon is supposed to be used as HTML not as plain text, so I don't see why you would escape the HTML. What you need is JavaScript escaping, to make sure the output of the Velocity code doesn't break the JavaScript code.

Indeed, with proper tests, it's easy to see the solution I proposed before was not good:
wrongEscaping__21475

Proposed changes merged to the PR :+1:

mflorea (approved):

PR #2586, opened 2023-11-07
XWIKI-21376: Live validation prompts should be presented to users without obtaining focus

surli:
* Fix AbstractRegistrationPage page object which is now broken since the introduction of a new CSS class in live validation elements


Sereza7 (approved):
:+1: 
I'm pretty sure we could remove 'LV_validation_message' from these selectors, but it should work nonetheless.
Thanks! :)

PR #2590, opened 2023-11-08
XWIKI-21452: Macros info, success, warning and error are only distinguished by colors

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21452
## PR Changes
* Updated style to take into account the addition of an icon in xwiki-rendering

## Tests
No specific test since this is just style changes

## View
21452-platformPRafter
[Video demo](https://up1.xwikisas.com/#gZy-GH4zy0ICnHaXAqBfiA) to check out responsivisity of these style changes

## Note
This PR should be merged on main, only after the appropriate changes have been merged on xwiki-rendering -> https://github.com/xwiki/xwiki-rendering/pull/286 .

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/messages.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/messages.less
@@ -2,10 +2,36 @@
 // Messages
 // --------------------------------------------------
 
-.successmessage, .errormessage, .warningmessage, .infomessage,
 span.successmessage, span.errormessage, span.warningmessage, span.infomessage, 
-span.box { // Used by: Inline boxes and messages
-  padding: floor(@font-size-base * 0.2);
+span.box { // Used by: Inline boxes
+  padding: 0 floor(@font-size-base * 0.2);
+  & > img {
+    // Style the icons of the message boxes
+    margin: 0 .5rem;
+    vertical-align: sub;
+  }
+}
+
+div.successmessage, div.errormessage, div.warningmessage, div.infomessage {
+  // Used by: message boxes
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  
+  & > p {
+    // Content of the message box
+    flex-grow: 1;
+    flex-basis: 98%;
+  }
+  
+  & > img {
+    // We need to do a hack with the flex because the title is not contained in its own node, 
+    // it's just there in a textnode.
+    min-width: 2%;
+    max-height: 16px;
+    object-fit: contain;
+  }
+  
 }
 
 .box, .plainmessage, // Used by: Code Macro, Success Macro, etc.


vmassol:
Thanks for the updated screenshot. The separation between the title and the text is not very clear.  I think we already have a jira issue about this (for the box macro since this is what is used underneath). I know this not your topic though and not really related to accessibility ;)

tmortagne:
Yes, the styling of the title itself is the subject of another issue, I mainly wanted to see the placement of the icon compared to the title. Thanks, @Sereza7.

Sereza7:
I can add a couple CSS rules to fix that in this PR ^^
https://github.com/xwiki/xwiki-platform/pull/2590/commits/12d5de087e7f7df63ab0f6c6e6d98285a976dc08 sets a `font-weight: bold` on the title. This isn't much but I think it's enough:
21452-emboldenedTitles
This is a screenshot displaying the changes in the last commit, we can see that the title font is bolder, which differentiates it slightly from the content of the box.

tkrieck:
Hi, I've done some mockups to test a better separation between elements. I don't know yet the difference between the small version and the big one on your screenshots, but I called them big and compact :) Some callouts have titles, others don't, just to check different versions.

normal

I think it's important to maintain the horizontal padding consistent, so we can have better alignment between the two versions, as we can see below. Also, from the screenshot you can see that maintaining the text in its own box can help improve legibility as the beginning of the text line always start at the same vertical point. 

redlines

If we could use Font Awesome it would be... well... awesome. The Silk iconset is pretty dated and, being a PNG, it does not render very well on a Hi Dpi monitor;

font awesome

tmortagne:
> the difference between the small version and the big one

It's inline (when the macro is in the middle of a paragraph, for example) and standalone (completely isolated).

Sereza7:
@tkrieck 
> I don't know yet the difference between the small version and the big one on your screenshots, but I called them big and compact

Big -> Regular macro, box with a message inside. You need to make sure you have an empty line before and after the macro.
Compact -> Inline mode for the macro, span with the message inside (and no title). This one is used if you didn't leave an empty line before/after the macro. We can't really play too much with the padding here since we want it to not increase line height.

For reference [here is the xwiki 2.1 syntax I used to generate the page in my screenshot](https://up1.xwikisas.com/#DYpXVTTrL15yT85UZOKnDA).

>I think it's important to maintain the horizontal padding consistent, so we can have better alignment between the two versions, as we can see below. Also, from the screenshot you can see that maintaining the text in its own box can help improve legibility as the beginning of the text line always start at the same vertical point.

How did you manage to get that style with the current HTML? I'm only seeing more and more style hacks we need to perform because there's a lack of container for the title. I'll add a container and update this PR soon so the style implementation can be more straightforward/stable.

> If we could use Font Awesome it would be... well... awesome. The Silk iconset is pretty dated and, being a PNG, it does not render very well on a Hi Dpi monitor;

Unfortunately in this context we can't use font-awesome without a lot more work (see https://github.com/xwiki/xwiki-rendering/pull/286 ). It's actually not that bad here because the icons have approximately the same colors as the blocks. IMO it's way worst with [transformation icons](https://extensions.xwiki.org/xwiki/bin/view/Extension/Icon%20Transformation) for example.

___
Setting this PR as a draft until I add a container for the title, set up vertical alignment and update padding :)

tkrieck:
> How did you manage to get that style with the current HTML?
Those are only mockups, but the alignment should be doable with an absolute positioning of the icon inside a relative positioned container for the message or even better with a flex layout;

Regarding the title, yeah, a container would help to achieve this effect without hacks on the CSS. 

> Unfortunately in this context we can't use font-awesome without a lot more work

No problem.

michitux:
> I think it's important to maintain the horizontal padding consistent, so we can have better alignment between the two versions, as we can see below. Also, from the screenshot you can see that maintaining the text in its own box can help improve legibility as the beginning of the text line always start at the same vertical point.
> 
> redlines

What about long messages without title? Would you indent them the same way as with title?

tkrieck:
> What about long messages without title? Would you indent them the same way as with title?

If they have icons before them, then yes.  

I don't know if there will be support from now on for messages without the icons, but if that's the case, then the text aligns normally with the message box, as it is today;

Example expanded with more options

redlines (1)

Sereza7:
## PR Changes
* Added style to fit the new `box-title` class
* Replaced the flexbox display with a grid to support vertical alignment of title and content.
## Note
We don't know the exact height we want for this box (depends on content), so we can't use a column flexbox.
## View
21452-fullSizeBoxesAfterPR

Sereza7:
Fixed the presentation for inplace editing in https://github.com/xwiki/xwiki-platform/pull/2590/commits/d6ed654e75edcd89361904d7997c2dd4ec3b3655 : 
21452-inplaceStyle

Inplace editing wraps the nodes we expected to be in the message box in a layer of CKEditor nodes. I added a selector to apply the same style to those CKEditor nodes as the one applied on the base nodes.

michitux:
> ## PR Changes
> 
>     * Added style to fit the new `box-title` class
> 
>     * Replaced the flexbox display with a grid to support vertical alignment of title and content.
> 
> 
> ## Note
> 
> We don't know the exact height we want for this box (depends on content), so we can't use a column flexbox.
> ## View
> 
> 21452-fullSizeBoxesAfterPR

This doesn't follow the styles proposed by @tkrieck - is there any reason for it? In particular, the icons have way less spacing to the border and are not on the same line as the text when there is no title. This looks very strange in my opinion. What @tkrieck looks much better in my opinion.

Also, I'm really not happy with introducing silk icons here. Even if we replace these icons by SVG icons from Font Awesome, this won't look good as they won't have the text color. I'm really wondering if it wouldn't be better to either add better icon support to `xwiki-rendering` or to override these macros in `xwiki-platform`. I think we should first discuss how to continue with icons before merging this.

Sereza7:
After these commits, here is what it looks like:
Silk vvv
21452-replacingIcons-Silk
Font Awesome vvv
21452-replacingIcons-FA

Sereza7:
@manuelleduc 
We discussed it with michitux a few days ago, it looked okay to me at first, but the conclusion was that I needed to review some of the style. I'll try to push a fix before the end of the day and share screenshots of the updated display (wanted to focus on the 'difficult' part which was creating a rendering API for icons).

Sereza7:
Here is the updated style (with font awesome):
21452-updatedStyle-FA

I replaced the grid layout with a flex, making use of an update to the DOM structure pushed on rendering. 
I increased the padding around the icon, and evened out left/right and top/bottom margins in the block.
Now, the icon is aligned with the first line of content. Note that alignment is different for silk icons, but the looks are similar (the 'base' icon is different so we need different rules to align them properly).

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
```diff
@@ -288,7 +288,11 @@ disk = floppy-o
 drink = glass
 drive = hdd-o
 email = envelope-o
-emoticon_smile=smile-o
+emoticon_smile= smile-o
+emoticon_grin = smile-o
+emoticon_unhappy = frown-o
+emoticon_tongue = smile-o
+emoticon_wink = smile-o
```
michitux, line 295: This looks wrong. The same icon for several emoticons?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-transformations/xwiki-platform-rendering-transformation-icon/src/main/resources/META-INF/components.txt
```diff
@@ -1 +1,2 @@
 500:org.xwiki.rendering.internal.transformation.icon.XWikiIconTransformationConfiguration
+org.xwiki.rendering.internal.transformation.icon.XWikiIconProvider
```
michitux, line 2: Don't you also need a `500:` here to safely override the other implementation?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-transformations/xwiki-platform-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/XWikiIconProvider.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return the block containing an icon.
+     */
+    @Override
+    public Block get(String iconName)
+    {
+        IconSet iconSet = null;
+        try {
+            iconSet = getIconSet(iconName);
+            String iconContent = this.iconRenderer.renderHTML(iconName, iconSet);
+            return new RawBlock(iconContent, Syntax.HTML_5_0);
```
michitux, line 71: In the icon macro, I'm using the rendered wiki syntax in order to have a chance to have, e.g., an actual image block. Is it intentional that you didn't use/move that code here? I know that FontAwesome actually just produces the same with a lot more effort, but in the case of Silk, for example, you would get an actual image block.

Again, I'm a bit worried also about export here as it is quite hard to support raw HTML blocks in an export.

michitux (changes requested):
Thank you also here for updating this both for using FA and updating the alignment. Just a few comments and when `IconProvider` is moved, the implementation here needs to be moved, too. The canonical place for things related to `xwiki-rendering-api` would be `xwiki-platform-rendering-xwiki`.

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
Sereza7: I made the choice to map these even though there is no correct alternative in FA.
Considering those icons are really far away from the initial commit, I'll revert those changes. I'd prefer having all icon transformations mapped to FA to not have silk in there anymore, but I guess it can't be done.
See https://jira.xwiki.org/browse/XRENDERING-734 , we discussed a bit about this IRL with manuel and using UTF-8 for emoticons (like those used in CKEditor) might be a solution.
Reverted the approximative icon mappings in https://github.com/xwiki/xwiki-platform/pull/2590/commits/2caaadf479c90111fb501c6714e4fb0b02663a6b :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-transformations/xwiki-platform-rendering-transformation-icon/src/main/resources/META-INF/components.txt
Sereza7: Sure :+1:  I wasn't sure wether to put it or not. When testing out my changes, it worked without it, and I figured I shouldn't add it if it's unnecessary.
I didn't find the [XWiki component documentation](https://extensions.xwiki.org/xwiki/bin/view/Extension/Component%20Module#HOverrides) when coding this part and it was a bit fuzzy still...

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-transformations/xwiki-platform-rendering-transformation-icon/src/main/java/org/xwiki/rendering/internal/transformation/icon/XWikiIconProvider.java
Sereza7: The code of the icon macro is woven with conditions related to the macro parameters. For this use case, I have no parameters so I don't need everything. 
Moreover, if I understood correctly I don't have access to the DisplayIconMacro class in `org.xwiki.icon.macro.internal` (and its `execute` method which is the one interesting to me) from the XWikiIconProvider in `org.xwiki.rendering.internal.transformation.icon`.

Sereza7:
> The canonical place for things related to xwiki-rendering-api would be xwiki-platform-rendering-xwiki.

Addressed in https://github.com/xwiki/xwiki-platform/pull/2590/commits/d9063155e070d99a9dd7ab9b175fc298339ff9fb :+1: 

I rebuilt all of the necessary modules successfully, and here is the result:
image
There is no UI change compared to before, except that we can see here icon transformation provided by two iconsets: font awesome when the mapping makes sense ( `:)` and `(y)`) and the default theme/Silk when it doesn't make much sense in other icon themes (`:P` and `:D`).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return the block containing an icon.
+     */
+    @Override
+    public IconBlock get(String iconName)
+    {
+        IconSet iconSet = null;
+        try {
+            iconSet = getIconSet(iconName);
+            String iconContent = this.iconRenderer.renderHTML(iconName, iconSet);
+            return new IconBlock(List.of(new RawBlock(iconContent, Syntax.HTML_5_0)));
```
michitux, line 70: If you should adopt my suggestion to remove `IconBlock` from the API, this needs to be adjusted.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
```diff
@@ -0,0 +1,92 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.rendering.internal.transformation;
```
michitux, line 20: Why is this in a `transformation` package? This is not really related to transformations anymore I think.

michitux (changes requested):
Thank you, overall this looks good just some updates to be done after the changes to the corresponding PR in xwiki-rendering.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
Sereza7: The `DefaultIconProvider` from `xwiki-rendering` is in a transformation package. I want to inherit this class which is internal, so I figured I'd put it in the same package.
Moreover, there's no real place where this would suit perfectly, the best I could find from a quick look would be `org.xwiki.rendering.internal.util`.

For reference, the classes `XWikiHTMLRawBlockFilter` and `XWikiErrorBlockGenerator` are in a similar situation.

I'm not very confident in this choice, let me know where you'd put it instead.

___
Taking a look to see if it would make sense to fit the `DefaultIconProvider` in utils, and move this one afterwards. :eyes:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2590/commits/341b4d976b3269115b4bb1b330b158d9d55bdbea :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
```diff
@@ -0,0 +1,92 @@
...
+            return new IconBlock(List.of(new RawBlock(iconContent, Syntax.HTML_5_0)));
+        } catch (IconException e) {
+            return super.get(iconName);
+        }
+    }
+
+    /**
+     * @return the java class of the icon block created.
+     */
+    public List<Class> getIconClass()
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2590/commits/341b4d976b3269115b4bb1b330b158d9d55bdbea :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
```diff
@@ -0,0 +1,92 @@
...
+@Component
+@Singleton
+public class XWikiIconProvider extends DefaultIconProvider
+{
+    /*
+    We need to update the icon class in order to take into account the situations
+    where we fall back on the default behaviour (the current icon theme does not have
+    the requested icon).
+     */
+    private static final List<Class> ICON_CLASS = new ArrayList<Class>(List.of(RawBlock.class, ImageBlock.class));
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2590/commits/341b4d976b3269115b4bb1b330b158d9d55bdbea :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/transformation/XWikiIconProvider.java
Sereza7: From what I can see, there's no dependency anymore to anything in the transformation package, so I'll be moving those classes to `util` and updating the `components.txt` files :)

Sereza7: Move done in https://github.com/xwiki/xwiki-platform/pull/2590/commits/e1dfba516e54c9fc40d6d909e2781097761b3a33 :+1:

Sereza7:
Changes since https://github.com/xwiki/xwiki-platform/pull/2590#pullrequestreview-1823547051:
* Updated the content and position of the XWikiIconProvider class to fit the changes on `xwiki-rendering`. It's now in the `util` package instead of being in the `transformation` package.

## Tests
After https://github.com/xwiki/xwiki-platform/pull/2590/commits/e1dfba516e54c9fc40d6d909e2781097761b3a33 and making sure all the changes in `xwiki-rendering` were built, successfully built:
* `mvn clean install -f xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-transformations/xwiki-platform-rendering-transformation-icon` 
* `mvn clean install -f xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki`

michitux (approved):

Sereza7:
TODO: Add a sr-only alternative for the icon See https://forum.xwiki.org/t/updating-the-presentation-of-message-macros/13872/15?u=charpentierlucas

Sereza7:
Added a text alternative for icons in the messages. Those text alternatives are not needed for any visual users, so I decided to leave them as `.sr-only` alternatives. Here is what the structure and display looks like after these changes to the XWikiIconProvider : 
21452-afterPRwithTextAlt

Waiting for the CSS update to set back this PR status as `ready for review`.

tkrieck:
I did the changes discussed[ in this topic](https://forum.xwiki.org/t/updating-the-presentation-of-message-macros/13872?u=charpentierlucas) with the proposed solution.

Screenshot 2024-01-29 at 11 10 23

tkrieck:
> @tkrieck I brought small codestyle fixes in [4e5bee5](https://github.com/xwiki/xwiki-platform/commit/4e5bee54295d7b3ce082341ad030bd8c8568283f) :
> 
> * One of our CSS practices is that there's a space before all properties. You had some lines with `border:none` that did miss this space. Just realized that this was not yet in [our codestyle](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/#HCSS)... Adding it right now.
> * Factorized some of the styles in the above selectors that already existed. As a rule of thumb, if you copy/paste some code you should ask yourself if you can factorize it. This is important for long term: if we need to update this bit of code, instead of changing 4 lines with the same content, we'll just be able to change one line (and avoid creating inconsistencies by forgetting to change some among the 4 lines...) :)

Got it, thank you!

PR #2593, opened 2023-11-09
XWIKI-21236: Object edit mode does not use current icon theme

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21236
## PR Changes
* Updated the icons on the object and class editors to use current icontheme
## Notes
* There's definitely some accessibility improvements to do left on this interface, e.g. the move 'button' on class editing that is instead just a span with an event listener...
* This was made during the BFD
* I did not change the property icons on the class editor, since there's no "close enough" replacement in the xwiki icon theme icons. E.g. we don't have a proper icon for a password field.
## View
Here is a couple screenshot of the interface before this PR, given for reference:
21236-beforePR2
21236-beforePR
Here are the same interfaces, updated (EDIT 24-06-24: those are screenshots from the latest version of this PR, on XWiki 16.6.0-snapshot): 
21236-afterPR2
21236-afterPR


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editobject.vm
```diff
@@ -39,10 +39,11 @@
...
   #getFieldSummary($obj $firstField $firstFieldSummary)
   <div id="xobject_${escapetool.xml($class.name)}_${obj.number}_title" class="xobject-title">
+    <span class="toggle-collapsable" role="presentation">$services.icon.renderHTML('caret-down')</span>
     <h3>#cleanClassname(${class.name}) <span class="editor-objectNumber">${obj.number}</span>#if ($firstField): $firstFieldSummary#end
-        <a href="$doc.getURL('objectremove', "form_token=$!{services.csrf.getToken()}&amp;classname=${escapetool.url($class.name)}&amp;classid=${obj.number}&amp;xredirect=${escapetool.url($doc.getURL('edit', 'editor=object'))}")" class="xobject-action delete" title="$services.localization.render('core.editors.object.removeObject.tooltip')">$services.localization.render('core.editors.object.removeObject')</a>
+        <a href="$doc.getURL('objectremove', "form_token=$!{services.csrf.getToken()}&amp;classname=${escapetool.url($class.name)}&amp;classid=${obj.number}&amp;xredirect=${escapetool.url($doc.getURL('edit', 'editor=object'))}")" class="xobject-action delete" title="$services.localization.render('core.editors.object.removeObject.tooltip')">$services.icon.renderHTML('cross')</a>
```
michitux, line 44: Why are you removing the text here? Isn't this bad for accessibility?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editobject.vm
```diff
@@ -39,10 +39,11 @@
...
         #if (!$hasCustomObject)
-          <a href="$doc.getURL('edit', "editor=object&amp;classname=${escapetool.url($class.name)}&amp;object=${obj.number}")" class="xobject-action edit" title="$services.localization.render('core.editors.object.editSingleObject.tooltip')">$services.localization.render('core.editors.object.editSingleObject')</a>
```
michitux, line 45: Same, why are you removing the text here? Isn't this bad for accessibility?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/dataeditors.js
```diff
@@ -753,7 +753,9 @@ editors.XDataEditors = Class.create({
       var movebutton = new Element('span', {
         'class': 'tool move',
         title: 'Drag and drop to change the order'
-      }).update('move');
+      }).update('<span class="presentation">' +
+        $jsontool.serialize($services.icon.renderHTML('arrows')) + 
+        '</span><span class="sr-only">move</span>');
```
michitux, line 758: Do we have an issue that this should be translated? If not, could you create one?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/dataeditors.js
```diff
@@ -753,7 +753,9 @@ editors.XDataEditors = Class.create({
       var movebutton = new Element('span', {
         'class': 'tool move',
         title: 'Drag and drop to change the order'
-      }).update('move');
+      }).update('<span class="presentation">' +
+        $jsontool.serialize($services.icon.renderHTML('arrows')) +
```
michitux, line 757: I think arrows is the wrong icon here. I don't know if we have a better icon, but this looks way too similar to the cross for removing the field and in some icon themes this is the icon for triggering fullscreen view. We might need to add a new icon for this.

michitux (changes requested):
Apart from the individual comments, I'm also wondering about the icon color. Previously, we had a red delete icon for example which is visually way clearer. Also, I know these are links and that's why they are blue. But as they're not actual links in the sense that you can visit a page but rather buttons I'm not sure that it is a good idea that all these icons get link color.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editobject.vm
Sereza7: 1. There is already a text alternative in the `title` field
2. This text was not visible, so it didn't help with partial vision users.

AFAIK this text was not useful for accessibility and an inconsistency between visuals and what was exposed to assistive techs.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editobject.vm
Sereza7: Same as https://github.com/xwiki/xwiki-platform/pull/2593#discussion_r1411848993

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/dataeditors.js
Sereza7: There's no icon in silk that would be appropriate for this use case IMO
See https://design.xwiki.org/xwiki/bin/view/Proposal/AddiconstotheXWikiIconSet and https://peacocksoftware.com/silk

That's a bit of a deadlock: to deprecate Silk, we need to replace its uses, but to replace its uses, we would need to have innacurate icons in some places until we deprecate Silk...

michitux: But couldn't we just use the existing icon for "silk", so basically adding a new icon to silk?

Sereza7: The icon we used for this element was a resource which was not part of Silk:
`icons/datamodel/move.png`
It's an extra resource provided in the XWiki web war, in the 'datamodel' folder https://github.com/xwiki/xwiki-platform/blob/bc6fb452ef10414c1756b2b547e8129ec1e35c9d/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/datamodel/move.png
It looks pretty similar to the silk style, but AFAIK it cannot be mapped using the icontheme...
And it respects its own naming convention, so there's no way to use it in the silk icontheme without a complete overhaul: `move` in datamodel is something different from `move` in silk or FA.

michitux: We can just copy the icon into the silk folder and then add a mapping in the silk icon theme. See for example #2645 which also adds a new icon to silk. In particular if we have an icon that can be added to silk, this shouldn't hold us back from using proper icons from FA.

Sereza7: Changes pushed in https://github.com/xwiki/xwiki-platform/pull/2593/commits/2138cd940c9b04a25e55a15a4fd617f9a88b47ce

michitux: Note that I'm also pretty sure that we fall back on the default icon theme which currently silk icons when an icon is not found, meaning that even if the user is using an icon theme that doesn't include a mapping for the new icon there is no regression compared to the current state - so this is not an argument for not adding a new icon.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/editclass.vm
```diff
@@ -71,8 +72,10 @@ $xwiki.jsfx.use('js/xwiki/editors/dataeditors.js', true)##
           'form_token': $services.csrf.token
         })))
         <a class="tool delete" href="$deletePropURL" title="$services.localization.render(
-          'core.editors.class.deleteProperty.tooltip', [$field.name])">$services.localization.render(
-          'core.editors.class.deleteProperty.text')</a>
+          'core.editors.class.deleteProperty.tooltip', [$field.name])">
+          <span role="presentation">$services.icon.renderHTML('cross')</span>
+          <span class="sr-only>">$services.localization.render(
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2593/commits/08e8010df2a5de63551c1cffc60edb412379abef :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/dataeditors.js
Sereza7: I figured the sr-only span is not necessary. I removed it and added a translation to the 'button' title where this icon appears. :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/editors/dataeditors.js
Sereza7: I just added the icon `reposition` to the iconTheme (default, silk and FA). I copied the resource from the datamodel folder to the silk one (while renaming it properly). Eventually we'll need to remove it from datamodel, but this resource might still be used in extensions so I don't think we should break this.
I'd prefer to use `move` to name icons for this, and use something along the line of `goto` for the icons we currently have mapped for `move`. However, this is an important breakage of the mapping, and doing this change would increase drastically the scope of this PR. 

See https://github.com/xwiki/xwiki-platform/pull/2593/commits/f3bf12105646ed809d25da90a4a7c7f76d5306e4 for the changes

Sereza7:
I updated the color of the delete action 'buttons' to `$theme.notificationErrorColor` in https://github.com/xwiki/xwiki-platform/pull/2593/commits/f85798ab5417b74b8505097dcd9dc867cd050a80.

michitux (changes requested):
From a technical point of view, this looks okay to me. However, I have a complaint regarding the visual changes in the object editor:

1. The space between the icon and the object name is too large.
2. The non-expanded icon seems to be a bit too low. I'm not 100% sure about this as putting a ruler on it, it doesn't seem really off but I have the feeling like the right corner of the triangle should be in the middle of the "P" instead of being aligned to the bottom. In the expanded state, the icon is also not aligned to the bottom but more centered.

Here a zoomed in version of the screenshot with a ruler:

image

Sereza7:
> 1. The space between the icon and the object name is too large.

This was caused because there is a default padding on this element, which a H3. Similarly, the padding on the H2 for the class title was removed. I removed it in https://github.com/xwiki/xwiki-platform/pull/2593/commits/609dab51596ddff647ed50eca626cf49ee18b957 :+1: 

> 2. The non-expanded icon seems to be a bit too low. I'm not 100% sure about this as putting a ruler on it, it doesn't seem really off but I have the feeling like the right corner of the triangle should be in the middle of the "P" instead of being aligned to the bottom. In the expanded state, the icon is also not aligned to the bottom but more centered.

When looking at it, it's not obvious to me :thinking:
When zooming in a lot, the icon ended up too high.
I'm not sure where this slight shift comes from. 
In https://github.com/xwiki/xwiki-platform/pull/2593/commits/609dab51596ddff647ed50eca626cf49ee18b957, I tried to correct the position by applying the rotation on icon itself instead of its container. I'm not sure this helps much. Here is a demo video with it properly shown in context and with a higher zoom level:

https://github.com/xwiki/xwiki-platform/assets/28761965/a01304c4-96f3-4067-baca-bcd3f06149a8

On the video, we can see that it respects two alignements:
* Its center does not move when rotating
* It's in the vertical center of the grey highlighted zone

Let me know if there's still an issue with the way things are now. I'm not sure how to make things better here :/

michitux (approved):

PR #2606, opened 2023-11-13
XWIKI-21545: Preview layout is broken

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21545
## PR Changes
* Fixed the broken preview
## Note
* This was caused by an oversight when implementing the changes for XWIKI-18838.
* No test was broken by the changes in XWIKI-18838, and I can't find a reference to ".bottombuttons" in the tests, so I suppose this PR will not break any tests.
## View
vvv Before this PR
21545-beforePR
vvv After this PR
21545-afterPR

We can see that the bottom button block was out of the mainContent 'white' block, which was unexpected. This is fixed by this PR.


manuelleduc (approved):
lgtm but might be safer to have a second look by someone else. Maybe @michitux as you probably followed the initial work.

PR #2609, opened 2023-11-14
XWIKI-21506: Both customize buttons in XWiki Preferences > Themes don't have top padding

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21506


surli:
Would be nice to have screenshots :)

PR #2613, opened 2023-11-15
[Misc] Change 'does not exists' to 'does not exist'

raphj:
I ran:

```
git grep -l  "does not exists"  | xargs sed -i 's/does not exists/does not exist/g'
```

### DIFF
--- a/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/Translations.xml
+++ b/xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/Translations.xml
@@ -244,7 +244,7 @@ admin.wikis.rights=Creation Right
 ## XWikiServer Sheet
 platform.wiki.sheet.title=Descriptor of the wiki \"{0}\"
 platform.wiki.sheet.erroraliasalreadynotexists=Wiki alias \"{0}\" already exists.
-platform.wiki.sheet.erroraliasdoesnotexists=Wiki alias \"{0}\" does not exists.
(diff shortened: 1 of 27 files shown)


xwiki-platform-core/xwiki-platform-wiki/xwiki-platform-wiki-ui/xwiki-platform-wiki-ui-mainwiki/src/main/resources/WikiManager/Translations.xml
```diff
@@ -244,7 +244,7 @@ admin.wikis.rights=Creation Right
 ## XWikiServer Sheet
 platform.wiki.sheet.title=Descriptor of the wiki \"{0}\"
 platform.wiki.sheet.erroraliasalreadynotexists=Wiki alias \"{0}\" already exists.
-platform.wiki.sheet.erroraliasdoesnotexists=Wiki alias \"{0}\" does not exists.
+platform.wiki.sheet.erroraliasdoesnotexists=Wiki alias \"{0}\" does not exist.
```
surli, line 247: Note that the key itself is "doesnotexists" :)

tmortagne: > This would be a breaking change, no?

Yes it would. The best you can do is move the current key in the deprecated section and introduce a new one used in the code (and bonus points if you also copy the translation to the new key).

surli: Not sure it worths it tbh, nobody really look at it and it wouldn't be the first time a key has a typo.

surli (approved):

PR #2627, opened 2023-11-20
XWIKI-21579: Invalid lang HTML Tag when "German (Germany)" locale is set

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21579
## PR Changes
* Used the method from the java Locale object that's appropriate for our use case
## Notes
* I searched all occurences for `lang="$xcontext` in the xwiki-platform codebase and replaced the `Locale` .toString call to get the return value from the appropriate method instead. The `Locale` class from java.util has a function just for this use case: [toLanguageTag()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html?is-external=true#forLanguageTag(java.lang.String)). This is enough of a fix for the accessibility test that initiated the reporting of this issue to pass and should be enough for long term.
* This fix was not found earlier, because most of our testing and every default instance is configured on the `en` locale, which has no subtag (the `Locale` `toString()` and its `forLanguageTag()` return the same value). 

## View
On the screenshot below, we can see the return values of both the former and new values proposed for the lang attribute.
21579-demo
On the screenshot below, we can see that the `<html>` block has an attribute `lang` which has a correct value (for the sake of the demo, default language has been updated to en-GB).
21579-demo2

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/ContentSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/ContentSheet.xml
@@ -59,7 +59,7 @@
   #else
     &lt;!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"&gt;
   #end
-  &lt;html xmlns="http://www.w3.org/1999/xhtml" lang="$xcontext.locale" xml:lang="$xcontext.locale"&gt;
+  &lt;html xmlns="http://www.w3.org/1999/xhtml" lang="$xcontext.locale.toLanguageTag()" xml:lang="$xcontext.locale"&gt;
     &lt;head&gt;
       &lt;meta http-equiv="Content-Type" content="text/html; charset=$xwiki.encoding" /&gt;
       ## Add HTML tags inside the page's head element, provided by UI Extensions implementing the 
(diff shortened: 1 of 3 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-ui/src/main/resources/CKEditor/ContentSheet.xml
```diff
@@ -59,7 +59,7 @@
   #else
     &lt;!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"&gt;
   #end
-  &lt;html xmlns="http://www.w3.org/1999/xhtml" lang="$xcontext.locale" xml:lang="$xcontext.locale"&gt;
+  &lt;html xmlns="http://www.w3.org/1999/xhtml" lang="$xcontext.locale.toLanguageTag()" xml:lang="$xcontext.locale"&gt;
```
tmortagne, line 62: `xml:lang` is not supposed to use the same format ?

Sereza7: https://www.w3.org/WAI/standards-guidelines/act/rules/5b7ae0/
You're right :+1: 
Addressed in https://github.com/xwiki/xwiki-platform/pull/2627/commits/2e4507395f94079f0c7735709ffb7b297681d82a

tmortagne (approved):
You know much more than me about HTML specification but at least the change make sense to me.

PR #2637, opened 2023-11-20
XWIKI-21583: Invitation decline link lacks contrast

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21583
## PR Changes
* Updated variablesInit to overwrite the default value got from bootstrap even when there's no theme
* Updated InvitationCommon.xml to match the changes made previously in InvitationConfig.xml
## Tests
Running `mvn clean install -f xwiki-platform-core/xwiki-platform-invitation/xwiki-platform-invitation-test -Dxwiki.test.ui.wcag=true -Pdocker` now only returns two WCAG violations, that are unrelated to contrast, and not the 9 or so extra that are currently returned in CI.

## View
Here is a demonstration of this link on an instance without a theme. We can see that unlike before, the link is correctly initialized to the expected value and not to the one from bootstrap.
21583-demoOnInstance

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/variablesInit.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/variablesInit.vm
@@ -261,7 +261,7 @@
   @state-danger-bg:                     #f8ecec;
   @state-success-bg:                    #e5f3df;
   @state-info-bg:                       #e1f1f9;
-  ## We want to keep overwritting the @brand-danger value from bootstrap as it does not pass contrast checks on #fff
+  ## We override the @brand-danger value from bootstrap as it does not pass contrast checks on #fff
   @brand-danger:                        #cc3333;
 #end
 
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/variablesInit.vm
```diff
@@ -261,6 +261,8 @@
   @state-danger-bg:                     #f8ecec;
   @state-success-bg:                    #e5f3df;
   @state-info-bg:                       #e1f1f9;
+  ## We want to keep overwritting the @brand-danger value from bootstrap as it does not pass contrast checks on #fff
```
michitux, line 264:
```suggestion
  ## We override the @brand-danger value from bootstrap as it does not pass contrast checks on #fff
```

There is at least a typo in "overwri**tt**ing", but for me, it is also not clear what this "want to keep" means, is there any reason to stop overriding this value that I missed?

michitux (approved):
Looks good to me, just a suggestion how to improve the wording of a comment.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/variablesInit.vm
Sereza7: You're right, this formulation was too wordy and can be abridged :+1:

PR #2638, opened 2023-11-20
XWIKI-21584: Mentions do not have enough contrast

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21584
## PR Changes
* Updated the default style of mentions
## View
vvv Before the PR
21584-beforePR
vvv After the PR
21584-afterPR

## Note
This is a style change, but I also took the opportunity to remove hard-coded transparent colors. Hard coded is against best practice, and adding transparency does not make sense in this context, the mention is too small to cover anything.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
@@ -374,12 +374,15 @@ require(['deferred!ckeditor', 'xwiki-suggestUsers', 'jquery', 'xwiki-meta'], fun
     <property>
       <code>.xwiki-mention {
   background-color: $services.mentions.mentionsColor;
-  border-radius: 10px;
-  padding: 2px 5px 2px 5px;
+  border-radius: @border-radius-small;
+  padding: 1px 5px 1px 5px;
+  border: 1px solid @dropdown-divider-bg;
 }
 
 .xwiki-mention.user.self {
   background-color: $services.mentions.selfMentionsColor;
+  color: @btn-danger-color;
+  border: 0;
 }
 
 .xwiki-mention.removed {
@@ -391,7 +394,7 @@ blockquote.mention-quote {
 }</code>
     </property>
     <property>
-      <contentType>CSS</contentType>
+      <contentType>LESS</contentType>
     </property>
     <property>
       <name/>
--- a/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/Configuration.xml
+++ b/xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/Configuration.xml
@@ -355,13 +355,13 @@
       </selfMentionsColor>
     </class>
     <property>
-      <mentionsColor>rgba(194, 194, 194, 0.8)</mentionsColor>
+      <mentionsColor>@breadcrumb-bg</mentionsColor>
     </property>
     <property>
       <quoteActivated>1</quoteActivated>
     </property>
     <property>
-      <selfMentionsColor>rgba(255, 0, 1, 0.5)</selfMentionsColor>
+      <selfMentionsColor>@btn-danger-bg</selfMentionsColor>
     </property>
   </object>
 </xwikidoc>


xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
```diff
@@ -374,12 +374,15 @@ require(['deferred!ckeditor', 'xwiki-suggestUsers', 'jquery', 'xwiki-meta'], fun
     <property>
       <code>.xwiki-mention {
   background-color: $services.mentions.mentionsColor;
-  border-radius: 10px;
-  padding: 2px 5px 2px 5px;
+  border-radius: @border-radius-small;
```
manuelleduc, line 377: A you rush 10px is `border-radius-small` ? Feels like a change in radius in the screenshots.

manuelleduc: I'm not against changing the border ratio, but it feels unrelated to the current fix.
Feel free to open a separate improvement for it though :slightly_smiling_face:

Sereza7: Changes reverted in https://github.com/xwiki/xwiki-platform/pull/2638/commits/60023d9bf464aeceb15f2fe4155b8d321ccd791d

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
```diff
@@ -375,11 +375,14 @@ require(['deferred!ckeditor', 'xwiki-suggestUsers', 'jquery', 'xwiki-meta'], fun
...
   background-color: $services.mentions.mentionsColor;
   border-radius: 10px;
-  padding: 2px 5px 2px 5px;
+  padding: 1px 5px 1px 5px;
+  border: 1px solid @dropdown-divider-bg;
 }
 
 .xwiki-mention.user.self {
   background-color: $services.mentions.selfMentionsColor;
+  color: @btn-danger-color;
```
surli, line 384: It doesn't feel good to use that here: what if an admin configures the mention to use the same color as `@btn-danger-color` for background? Feels like this color should also be configurable...

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
Sereza7: I just tried creating a ticket, and noticed the recommendation pointing to https://jira.xwiki.org/browse/XWIKI-21502 ...
Adding a note on this ticket :D

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
Sereza7: > what if an admin configures the mention to use the same color as @btn-danger-color for background? 

That's a mistake, similar to setting text in white in Iceberg, we can't do much to prevent that kind of error... As of now `*-color` are for foreground colors, and `*-bg` are for background colors. The first one is not really obvious.

The naming convention in this module are not making things easier... they use a `*Color` structure for the parameter names for backgrounds.

I don't know why only the background was made configurable :thinking: 
Updated those color parameters in https://github.com/xwiki/xwiki-platform/pull/2638/commits/ae70d914396f28a86088b248d6ce056878d8abb4 :+1:

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
```diff
@@ -373,13 +373,16 @@ require(['deferred!ckeditor', 'xwiki-suggestUsers', 'jquery', 'xwiki-meta'], fun
     </property>
     <property>
       <code>.xwiki-mention {
-  background-color: $services.mentions.mentionsColor;
+  background-color: $services.mentions.mentionsBackground;
```
surli, line 376: It's using a script service so you should probably also check the Java API and probably makes change there too... That kind of change is not that easy :)

Sereza7: I reverted this change in service name. From what I udnerstand, we'd need an API breakage for this, and I'm not sure renaming one property is enough of a reason.

Sereza7:
* Reverted most of the changes to allow a better backwards compatibility
* Used a `foreground` variable instead of a `background` one so that the meaning and content of the `color` would stay the same and we don't break them.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
```diff
@@ -375,11 +375,14 @@ require(['deferred!ckeditor', 'xwiki-suggestUsers', 'jquery', 'xwiki-meta'], fun
...
   background-color: $services.mentions.mentionsColor;
   border-radius: 10px;
-  padding: 2px 5px 2px 5px;
+  padding: 1px 5px 1px 5px;
+  border: 1px solid @dropdown-divider-bg;
 }
 
 .xwiki-mention.user.self {
   background-color: $services.mentions.selfMentionsColor;
+  color: $services.mentions.selfMentionsForeground;
```
surli, line 384: Looks like you forgot to add this as a new API in the script service.

Sereza7: @surli 
Sorry for the delay,

I'm not very confident on script services yet. 
I added an API in https://github.com/xwiki/xwiki-platform/pull/2638/commits/e87afbf1a2d663125c07c90fa33eb19bf11ca0c4 
It definitely needs some review.

I added the unstable annotation onto this API since this is new.
From what I understand, the mention configuration source is registered under the same service name, we get the value of the mentionForeground color directly when using `$services.mentions.selfMentionsForeground`, but adding the `getSelfMentionsForeground` method onto the MentionScriptService allows us to:
* Properly make it an API
* Access it from other XWiki components if needed
* Access it from velocity script with `$services.mentions.getSelfMentionsForeground()`

Why did we not use this method for `selfMentionsColor` and we get the value directly from the configuration source with the variable name `$services.mentions.selfMentionsColor` ? Since we have a proper API getter I would have expected us to use it.

Let me know if I misunderstood something (which I probably did ^^`)

Sereza7:
Setting this PR as a draft until resolution of https://github.com/xwiki/xwiki-platform/pull/2638#discussion_r1732234656

Sereza7:
@surli This is the last contrast violation found by automated tests, it would be great if we could progress on this. What do you think of the solution I proposed [here](https://github.com/xwiki/xwiki-platform/pull/2638#discussion_r1803016592)? Can you spot anything to change before a merge?

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/script/MentionsScriptService.java
```diff
@@ -74,6 +75,17 @@ public String getSelfMentionsColor()
         return this.configuration.getSelfMentionsColor();
     }
 
+    /**
+     *
+     * @see MentionsConfiguration#getSelfMentionsForeground()
+     * @return the mentions foreground color configuration value for the current user.
+     */
+    @Unstable
```
surli, line 83: Missing a since in the doc

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsConfiguration.java
```diff
@@ -48,15 +49,21 @@ public interface MentionsConfiguration
...
-     * @return the color for the mentions to the current user.
+     * @return the background color for the mentions to the current user.
      */
     String getSelfMentionsColor();
 
+    /**
+     * @return the foreground color for the mentions to the current user.
+     */
+    @Unstable
+    String getSelfMentionsForeground();
```
surli, line 65: Missing since info

surli: Also this should probably be a default method to not break compatibility.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-ui/src/main/resources/XWiki/Mentions/MentionsMacro.xml
surli: > From what I understand, the mention configuration source is registered under the same service name

I don't understand what you mean there. `MentionsConfiguration` is not implementing any script service role, so it cannot be called by using any `$services.mention`. 

> Why did we not use this method for selfMentionsColor and we get the value directly from the configuration source with the variable name $services.mentions.selfMentionsColor ?

I also don't understand what you mean here: we use same method for `$services.mentions.selfMentionsColor ` it's a call to the script service.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsConfiguration.java
Sereza5: Addressed both of those in https://github.com/xwiki/xwiki-platform/pull/2638/commits/9a1b31b22efd939059e5e9c8555b028cb14539b2 :+1:  . I picked the default value from the default Xwiki Standard flavor configuration.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/script/MentionsScriptService.java
Sereza5: Addressed in https://github.com/xwiki/xwiki-platform/pull/2638/commits/9a1b31b22efd939059e5e9c8555b028cb14539b2 :+1:

Sereza7:
All comments have been addressed, this PR is ready for further review :)

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsConfiguration.java
```diff
@@ -56,6 +56,13 @@ public String getSelfMentionsColor()
         return this.configuration.getProperty("selfMentionsColor", "rgba(255, 0, 1, 0.5)");
     }
 
+    @Override
+    public String getSelfMentionsForeground()
+    {
+        // default color is rgba, not hexa to allow setting opacity and keep working on IE11
+        return this.configuration.getProperty("selfMentionsForeground", "rgba(255, 255, 255, 1)");
```
surli: The default should match what's in MentionsConfiguration default return.

michitux:
Wouldn't it make more sense to move the color configuration to the color theme and use CSS variables to access them?

I'm not saying this isn't okay as a "quick fix", I have just the feeling that from a technical/implementation point of view this is going in the wrong direction.

In any case, it's good to address those last remaining contrast issues, thank you very much for that!

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-default/src/main/java/org/xwiki/mentions/internal/DefaultMentionsConfiguration.java
Sereza5: Addressed in https://github.com/xwiki/xwiki-platform/pull/2638/commits/68048517ca59f7122fe76a0c97929ca08e9d2a4e :+1:

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/script/MentionsScriptService.java
```diff
@@ -74,6 +75,18 @@ public String getSelfMentionsColor()
         return this.configuration.getSelfMentionsColor();
     }
 
+    /**
+     *
+     * @see MentionsConfiguration#getSelfMentionsForeground()
+     * @return the mentions foreground color configuration value for the current user.
+     * @since 18.0.0RC1
```
surli, line 82:
```suggestion
     * @since 18.0.0RC1
     * @since 17.10.3
```

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-api/src/main/java/org/xwiki/mentions/MentionsConfiguration.java
```diff
@@ -48,15 +49,27 @@ public interface MentionsConfiguration
...
-     * @return the color for the mentions to the current user.
+     * @return the background color for the mentions to the current user.
      */
     String getSelfMentionsColor();
 
+    /**
+     * The default value is white (@btn-danger-color value from the Iceberg FlamingoTheme)
+     * that is meant to have enough contrast on the saturated red of the background.
+     * @return the foreground color for the mentions to the current user.
+     * @since 18.0.0RC1
```
surli, line 65:
```suggestion
     * @since 18.0.0RC1
     * @since 17.10.3
```

PR #2642, opened 2023-11-21
XWIKI-21521: DefaultSkin form should be accessible

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21521
## PR Changes
* Wrapped the pretty name in a label which is bound to the form field with its `for` attribute
## Note
* This is an improvwement upon https://github.com/xwiki/xwiki-platform/pull/2584 

## View
We can see on the screenshot below that an axe-core analysis of the page does not return any violation anymore.
21521-afterPR


xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-ui/src/main/resources/SkinsCode/XWikiSkinsSheet.xml
```diff
@@ -149,6 +149,7 @@
     (((
       {{html}}&lt;button class="btn btn-default btn-xs"&gt;$services.icon.renderHTML('cross')&lt;/button&gt;{{/html}}
     )))
+    )))
```
michitux, line 152: I cannot see any code that adds the corresponding `(((`, was this missing before?

michitux (approved):
Looks good apart from the added `)))` where I'm not sure why it was added.

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-ui/src/main/resources/SkinsCode/XWikiSkinsSheet.xml
Sereza7: Sorry, it was a mistake, I don't know what I did here...
Addressed in https://github.com/xwiki/xwiki-platform/pull/2642/commits/18cc897f5255e678a50fb4f70fa31e914338ed7d :+1:

xwiki-platform-core/xwiki-platform-skin/xwiki-platform-skin-ui/src/main/resources/SkinsCode/XWikiSkinsSheet.xml
```diff
@@ -80,7 +80,7 @@
   #set ($class = $doc.getObject($className).xWikiClass)
   #foreach ($prop in $class.properties)
     #if ($prop.classType != 'TextArea')
-    ; $services.rendering.escape($prop.prettyName, 'xwiki/2.1')
+    ; {{html}}&lt;label for="${className}_0_$escapetool.xml($prop.name)"&gt;$escapetool.xml($prop.prettyName)&lt;/label&gt;{{/html}}
```
surli, line 83: It's not necessarily the object at index 0 is it? Chances are low but we could imagine that the retrieved object has another index no?

Sereza7: I just tried using the `$prop.number` attribute to fill up this id.
Seems like the number used in the ids is not the same, and so it doesn't match.
image
From what I could see in this form, all properties used `0` (because they were the only ones with their name, no duplicate names? ).
I'll be leaving the PR as is for now.

surli: > I just tried using the $prop.number attribute to fill up this id.

Yes it won't work with `$prop.number`: as the name suggest it's the index of the property. Here the 0 is the index of the `xobject` itself.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2642/commits/afe5ee95e8f445fff2d8b56c0f551d70b615f6fe :+1: 

Here is the result with the updated template:
21521-wihtoutHardcodedNumber
As we can see, all attribute values reference to the right nodes.

PR #2648, opened 2023-11-23
XWIKI-11043: Use Flamingo-like button styling when editing Dashboard

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-11043
## PR Changes
* Updated the presentation of buttons
* Removed the use of hard coded silk references in dashboard.css, replkaced them with calls to the icon theme
## View
After the PR
11043-afterPR
We can see that the new link buttons follow the icon and color theme styles.
Additionally, actions for the widgets are updated to use the current icon theme too.
## Note
I decided to replace the old style based on floats with some flex.


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/dashboard/dashboard.js
```diff
@@ -75,7 +80,12 @@ XWiki.Dashboard = Class.create( {
     // find out all the gadget-containers in element and add them ids
     this.containers = element.select(".gadget-container");
     this.createDragAndDrops();
-    this.addGadgetsHandlers();
+    this.addGadgetsHandlers();``
+    // Create the section to contain add buttons
+    var sectionAddButtons = new Element('section', {
+      'class': 'containeradd'
+    })
+    this.element.insert({'top' : sectionAddButtons});
```
surli, line 88: You couldn't actually put that directly in the velocity template? From what I understand this DOM element would always be there no?

Sereza7: From what I understand, tehre is no velocity template for the edit part of the dashboard. Everything added to make it nice to edit the dashboard is done so through this .js script.
Those elements are only there in edit mode.
Most of the dashboard is handled in [its own macro](https://github.com/xwiki/xwiki-platform/blob/9a4e759f3189b2125a5d13af6ab32ce24515f5f9/xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-macro/src/main/java/org/xwiki/rendering/internal/macro/dashboard/DashboardMacro.java#L279).

surli:
@Sereza7 I think you should check that DashboardEditPage doesn't need to be fixed too: I'm afraid that your changes would break it and the integration test relying on it.

Sereza7:
> @Sereza7 I think you should check that DashboardEditPage doesn't need to be fixed too: I'm afraid that your changes would break it and the integration test relying on it.

Right now the Solr upgrade makes the enforcer fail.
However I could pass 
`mvn clean install -f xwiki-platform-core/xwiki-platform-dashboard/xwiki-platform-dashboard-test/xwiki-platform-dashboard-test-docker -Pquality,docker,integration-tests -Dit.test=DashboardIT -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false  -Dxwiki.enforcer.skip=true` successfully. This is the only place where the `DashboardEditPage` pageobject is used in xwiki-platform.
I checked the selectors in the DashboardEditPage and from what I could see, selectors are resilient to DOM changes:

* `By.cssSelector(".dashboard .gadget")` doesn't break since we keep the .gadget class, and additional wrapping doesn't affect it.
* `By.cssSelector(".dashboard .addgadget")` similar

PR #2649, opened 2023-11-23
XWIKI-21597: Make the rights UI use icon themes

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21597
## PR Changes
* Updated icons in the rights UI
## View
After the PR
21597-afterPR
We can see that this PR updates the icons both in the right assignment field and the fields under. UInder, this is just a two state checkbox, which can be handled by the default HTML, so I did not use the icon theme. However in the right assignment fields, I changed the icon for the three states. For the none state, which was previously shown with an empty box, I decided to leave the button visuals empty.


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.js
```diff
@@ -57,44 +57,42 @@ window.MSCheckbox = Class.create({
...
     this.states = [0,1,2]; // 0 = undefined; 1 = allow, 2 = deny
     this.nrstates = this.states.length;
-    this.images = [
-      "$xwiki.getSkinFile('js/xwiki/usersandgroups/img/none.png')",
-      "$xwiki.getSkinFile('js/xwiki/usersandgroups/img/allow.png')",
-      "$xwiki.getSkinFile('js/xwiki/usersandgroups/img/deny1.png')"
+    this.icons = [
+      "",
+      "$escapetool.javascript($services.icon.renderHTML('check'))",
+      "$escapetool.javascript($services.icon.renderHTML('cross'))"
```
surli, line 63: Would be cleaner to move out the velocity calls, as it's recommended in https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HJavaScriptBestPractices (and what you used in https://github.com/xwiki/xwiki-platform/pull/2648/files#diff-14e05ae329b087de0d0f45b460cdb1b33da4382f3d9c806125daf81cfac152bbR51 for example)

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2649/commits/78ca48e9fb3ed17d074ec0d0aefd8f615db47672 :+1: 
I tested this change manually and as far as I could see it did not break anything (at least I fixed mistakes until I could notice it working again :) ).

Sereza7:
Faisl webstandard tests, `Bad value “checkbox” for attribute “type” on element “input”.`. I think this is because I left an alt on it.

Sereza7:
With the changes from https://github.com/xwiki/xwiki-platform/pull/2649/commits/b2f2a7a69e78ee3d533bfe4b27eea204f13b02d6, testing 
`mvn clean install -amd -f xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards  -Pquality,integration-tests,docker -Dxwiki.test.startXWiki=false` on a live distribution didn't fail for the `Rights` section. Other fails were found, but expected in the context of this one week old distribution (e.g. without fixes for https://github.com/xwiki/xwiki-platform/pull/2796 but the test suite itself was updated).

Everything good, opening this PR back for reviews :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.css
```diff
@@ -51,10 +51,11 @@
   width: 15% !important;
 }
 
-/* This button has no border or background, but wraps an image to allow keyboard navigation. */
-#usersandgroupstable button.rights-edit {
-  border-color: transparent;
-  background: transparent;
+#usersandgroupstable button.rights-edit.yes {
+  color: $theme.notificationSuccessColor;
```
surli, line 55: Feels a bit weird here to use `notification(Success|Error)Color`, now maybe that's the best practice in themes?

surli (approved):
LGTM

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/usersandgroups/usersandgroups.css
Sereza7: That's some of the old colortheme variables. I'm not sure we have a best practice for this. I think I did use those old variables since others are used all throughout the file and I'd rather keep consistent inside one file.

To be honest this whole file looks quite old and breaks current CSS codestyles in multiple places.
I opened a topic on the forum about introducing a new [type of label for codestyle](https://forum.xwiki.org/t/jira-label-proposal-codestyle/14167), this file could really use a ticket :)

I just checked our [HTML and CSS best practices](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/) and the entry `don't hard code colors in properties - use [ColorTheme variables](https://extensions.xwiki.org/xwiki/bin/view/Extension/Color%20Theme%20Application#HUsingColorThemesvariables)` points to a doc that uses the old colortheme variables. We might want to update the link or the doc.

I'll look a bit more into the state of deprecation of this old colortheme, it's not clear to me what it is yet. Using the old colortheme everywhere does not look like something we should recommend...

Sereza7: I updated the doc to include a link to how to use the new variables.

PR #2650, opened 2023-11-23
XWIKI-21599: Theme creation form uses a silk icon

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21599
## PR Changes
* Replaced the silk icon reference with a tangible iconset icon
## View
Before the PR
ThemeCreationFormSilkIcon
After the PR
21599-afterPR


michitux (commented):
This also removes the background color of the form. Is this intentional? Apart from that this looks good.

Sereza7:
This was not intentional. However, I don't think this background provides much for the form. This reduces visual clutter a bit. Following this change, I also updated the border color to remove the "form box" completely from the visuals in https://github.com/xwiki/xwiki-platform/pull/2650/commits/d35f91e8d1ea5120c3892688c963a5512e011dcc. IMO it was barely noticeable before, and contained only one line with two fields so it didn't play much of a role anyways.
Here is what the form looks like without the border:
21599-afterUpdatePR
Note that I could barely notice the border even at full resolution. This doesn't look different from the screenshot for the initial proposal.

surli (approved):

PR #2655, opened 2023-11-24
XWIKI-21608: XWIKI-21509 Broke Upgrade if Tomcat Service is Running, …

gohrner:
…But Not Enabled

If the service is running, we need to restart it in any case, no matter if it's enabled or not.
However is-active does not always seem to work if the package just gets installed (see XWIKI-18624), so also check if it's only enabled and try to restart it then.

See https://forum.xwiki.org/t/no-tomcat-restart-after-xwiki-lts-debian-package-upgrade/13505 for the discussion.


tmortagne:
Thanks for contribution !

PR #2667, opened 2023-11-28
XWIKI-13999: Use the available horizontal space for page titles

michitux:
* Remove fixed columns from the document header.
* Use flexbox to dynamically grow the page title while still wrapping it when it doesn't fit.

I added a padding to the buttons as otherwise there was no more space between the buttons and the title. I'm not sure that this is the correct padding.

Jira issue: https://jira.xwiki.org/browse/XWIKI-13999

View mode with long title that wraps:

image

View mode with long title that doesn't wrap:

image

Edit mode:

image

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
@@ -260,6 +260,8 @@
 .document-header {
   margin-top: -@line-height-computed;
   margin-bottom: @line-height-computed;
+  display: flex;
+  flex-wrap: wrap;
   #document-title h1 {
     word-wrap: break-word;
   }
@@ -269,6 +271,14 @@
   }
   .document-menu, .document-info {
     margin-top: @line-height-computed;
+  }
+  .document-info {
+    flex-grow: 1;
+  }
+  .document-menu {
+    margin-left: auto;
+    padding-left: @grid-gutter-width;
+    white-space: nowrap;
   }
 }
 
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentheader.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/contentheader.vm
@@ -17,8 +17,8 @@
 ## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
 ## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 ## ---------------------------------------------------------------------------
-<div class="row document-header">
-  <section class="document-info col-xs-12 #if($displayContentMenu)col-md-7#end"
+<div class="document-header">
+  <section class="document-info"
     aria-label="$escapetool.xml($services.localization.render('core.document.header.info.label'))">
     ## --------------------------------------------------------
     ## Display UI Extensions before the title element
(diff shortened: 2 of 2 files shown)


michitux:
With ratings:

image

image

image

michitux:
Things I don't like after testing a bit more:

View mode looks nice:

image

Edit mode looks totally different:

image

Maybe we could use some [JavaScript code](https://stackoverflow.com/questions/65575027/how-to-set-html-input-tag-width-fit-content-in-css) to dynamically adjust the width of the input based on its content with some minimum width?

At the moment, the content menu that is inside the `div.document-menu` has a `pull-right` class that has a `float: right` style. While it seems to work, I think it would be better to remove that to avoid surprises. I used `white-space: nowrap` to avoid a line break between the different menus, but this might also have unintended side effects, maybe we should unset that again inside the menu? I expect that this line break was also a consequence of the floating, but I didn't check. 

Also, my changes currently don't apply to the "inline" edit (with sheet) screen as this is another template. However, it also doesn't break it. Removing the floating on the document menu currently no effect as there is another style with text-alignment that also brings the menu to the right side. However, the menus are a UIXP which might be affected. I think for consistency and also to avoid issues when removing the floating we should also adapt that template.

surli:
> View mode with long title that wraps:

I don't understand the screenshot below that one: shouldn't the title be wrapped and not the buttons? I mean I'd expect the button to not move but instead the title to be on 2 lines there.

michitux:
> I don't understand the screenshot below that one: shouldn't the title be wrapped and not the buttons? I mean I'd expect the button to not move but instead the title to be on 2 lines there.

The current behavior on small screens is that the buttons are below the title as the space next to the buttons could be too small for the title. I basically implemented this behavior regardless of the screen width but based on title width. I agree that this might not always be desirable. A good option that should also be easy to implement would be to disable this wrapping above a certain screen width (e.g., above the screen width at which we currently wrap). I think for small screen widths there is no alternative to wrapping.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
```diff
@@ -270,6 +272,14 @@
   .document-menu, .document-info {
     margin-top: @line-height-computed;
```
tkrieck, line 273: Since the proposal is to use flex (which is the right way IMO), we could also start using the `gap` property. With this, we can remove the individual margins and paddings from `.document-menu, .document-info`. To achieve the same results, the gap should be `gap: @line-height-computed @grid-gutter-width ;`

With this implementation, the buttons also come to the left side of the screen when breaking, making for a more streamlined appearance.

To test it (from line 260 to 279):
```
.document-header {
  display: flex;
  flex-wrap: wrap;
  gap: @line-height-computed @grid-gutter-width ;

  #document-title h1 {
    word-wrap: break-word;
  }
  #document-title-input {
    // Override the fixed height as the content is a large heading that would otherwise overflow.
    height: auto;
  }

  .document-info {
    flex-grow: 1;
  }
  .document-menu {
    white-space: nowrap;
  }
}
```

Here are the expected results:

https://github.com/user-attachments/assets/af3c7a6d-7fc1-4b7b-aee0-0a62870ac72f

In edit mode, the title takes all the available space:

Screenshot 2025-02-19 at 09 40 24

tkrieck:
Thanks for working on this problem! It bugged me for quite some time now =)

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/layout.less
tkrieck: https://github.com/xwiki/xwiki-platform/pull/2667#discussion_r1961629818 The side effect of this is that the buttons wrap when the title is too long, so perhaps is important to make a media query on `flex-wrap` to only wrap when a certain breakpoint is reached (@dl-horizontal-breakpoint, for example).

PR #2680, opened 2023-12-01
XWIKI-21659: Provide entry index in data attribute of LiveData

surli:
* Add `data-livedata-entry-index` and `data-livedata-entry-id` attributes on the parent DOM node of the entries, for both cards and table layouts

See: 
Capture d’écran du 2023-12-04 11-42-27
Capture d’écran du 2023-12-04 11-42-05

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
@@ -75,8 +75,7 @@
             v-for="(entry, idx) in entries"
             :key="`table-${logic.getEntryId(entry)}-${idx}`"
             :entry="entry"
-            :data-livedata-entry-index="idx"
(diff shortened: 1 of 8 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
```diff
@@ -75,7 +75,8 @@
             v-for="(entry, idx) in entries"
             :key="`table-${logic.getEntryId(entry)}-${idx}`"
             :entry="entry"
-            :data-index="idx"
+            :data-livedata-entry-index="idx"
```
manuelleduc, line 78: I know we really have clear practice on this, but wdyt of adding a comment with the "since" version for the two values, as they will become part of the API?

surli: Any advice on the format to use? AFAICS I cannot add a comment in the middle of the properties here?

PR #2692, opened 2023-12-05
XWIKI-21448: Notification filter preferences are not cleaned when a page is deleted

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-21448

This PR contains different works: 
  1. a renaming of `ModelBridge` class contained in `xwiki-platform-notifications-filters-api` to `FilterPreferencesModelBridge` to clarify what it's used for: this leads to renaming lots of stuff
  2. a new API in `NoficationFilterPreferenceManager` (and lower level components) to allow deleting multiple filters at once
  3. a new component `DeletedDocumentCleanUpFilterProcessingQueue` aiming at processing the clean up of notifications filters in a dedicated low priority thread, a listener `DeletedDocumentCleanUpFilterListener` calling it whenever a doc is deleted in a folded event, and UserEventDispatcher calling it also whenever a doc is deleted in an unfolded event
  4. a migration to clean up filters that are not related to an existing documents when upgrading

### DIFF
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterPreferenceManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterPreferenceManager.java
@@ -30,6 +30,7 @@ import org.xwiki.model.reference.DocumentReference;
 import org.xwiki.model.reference.WikiReference;
 import org.xwiki.notifications.NotificationException;
 import org.xwiki.notifications.NotificationFormat;
+import org.xwiki.stability.Unstable;
 
 /**
  * Provide an interface for interacting with user notification filters preferences.
@@ -117,6 +118,22 @@ public interface NotificationFilterPreferenceManager
      * @since 9.11.9
      */
     void deleteFilterPreference(DocumentReference user, String filterPreferenceId) throws NotificationException;
+
+    /**
+     * Delete a set of filter preferences.
+     * @param user the user to use
+     * @param filterPreferenceIds ids of the filter preferences
+     * @throws NotificationException if an error happens
+     *
+     * @since 15.10.2
+     * @since 16.0.0RC1
+     */
+    @Unstable
+    default void deleteFilterPreferences(DocumentReference user, Set<String> filterPreferenceIds)
+        throws NotificationException
+    {
+        // Do nothing
+    }
 
     /**
      * Delete a filter preference.
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterPreferenceManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterPreferenceManager.java
@@ -170,15 +170,21 @@ public class DefaultNotificationFilterPreferenceManager implements NotificationF
     @Override
     public void deleteFilterPreference(DocumentReference user, String filterPreferenceId) throws NotificationException
     {
+        deleteFilterPreferences(user, Set.of(filterPreferenceId));
+    }
+
+    @Override
+    public void deleteFilterPreferences(DocumentReference user, Set<String> filterPreferenceIds)
+        throws NotificationException
+    {
         try {
             for (NotificationFilterPreferenceProvider provider
-                    : componentManager.<NotificationFilterPreferenceProvider>getInstanceList(
-                    NotificationFilterPreferenceProvider.class)) {
-                provider.deleteFilterPreference(user, filterPreferenceId);
+                : componentManager.<NotificationFilterPreferenceProvider>getInstanceList(
+                NotificationFilterPreferenceProvider.class)) {
+                provider.deleteFilterPreferences(user, filterPreferenceIds);
             }
-
         } catch (ComponentLookupException e) {
-            logger.info("Failed to remove the user filter preference [{}].", filterPreferenceId, e);
+            logger.info("Failed to remove the user filter preference [{}].", filterPreferenceIds, e);
         }
     }
 
--- a/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterPreferenceManager.java
+++ b/xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterPreferenceManager.java
@@ -158,7 +158,7 @@ public class DefaultNotificationFilterPreferenceManager implements NotificationF
                 provider.saveFilterPreferences(user, preferencesMapping.get(providerHint));
 
             } catch (ComponentLookupException e) {
-                logger.error("Unable to retrieve the notification filter preference provider for hint [{}]: [{}]",
+                logger.error("Unable to retrieve the notification filter preference provider for hint [{}]:",
                         providerHint, e);
             } catch (NotificationException e) {
                 logger.warn("Unable save the filter preferences [{}] against the provider [{}]: [{}]",
@@ -184,7 +184,7 @@ public class DefaultNotificationFilterPreferenceManager implements NotificationF
                 provider.deleteFilterPreferences(user, filterPreferenceIds);
             }
         } catch (ComponentLookupException e) {
-            logger.info("Failed to remove the user filter preference [{}].", filterPreferenceIds, e);
+            logger.info("Failed to remove the user filter preferences {}.", filterPreferenceIds, e);
         }
     }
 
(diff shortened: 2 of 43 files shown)


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/UserEventManager.java
```diff
@@ -302,4 +315,30 @@ private boolean isEventFiltered(List<NotificationFilter> filters, Event event, D
...
         return false;
     }
+
+    public void cleanUpFilters(DocumentDeletedEvent event, DocumentReference user)
+    {
+        DocumentReference deletedDocumentReference = event.getDocumentReference();
+        String serializedReference = this.entityReferenceSerializer.serialize(deletedDocumentReference);
+        try {
+            Set<NotificationFilterPreference> matchingPreferences =
+                this.notificationFilterPreferenceManager.getFilterPreferences(user)
```
surli, line 325: Here it means we won't get only ScopeNotificationFilterPreferences, but could also be other kind of preferences: I guess it makes sense?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/UserEventManager.java
```diff
@@ -302,4 +315,30 @@ private boolean isEventFiltered(List<NotificationFilter> filters, Event event, D
...
+                    // We only clean up the filters matching exactly the page reference
+                    // we never clean up filters for space / wiki there
+                    .stream().filter(pref -> StringUtils.equals(pref.getPageOnly(), serializedReference))
+                    .collect(Collectors.toSet());
+            if (!matchingPreferences.isEmpty()) {
+                CleaningFilterEvent cleaningFilterEvent = new CleaningFilterEvent();
+                this.observationManager.notify(cleaningFilterEvent, deletedDocumentReference, matchingPreferences);
```
surli, line 332: I'm not sure what's best between sending a single event with all the preferences about to be deleted, or multiple events for each one. I feel like it's better like this for scalability... 
Also I guess we need to be careful to document that any listener on `CleanFilterEvent` shouldn't perform any operation leading to a delete to not enter in loops...

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-notifiers/xwiki-platform-notifications-notifiers-default/src/main/java/org/xwiki/notifications/notifiers/internal/UserEventManager.java
```diff
@@ -302,4 +315,30 @@ private boolean isEventFiltered(List<NotificationFilter> filters, Event event, D
...
+                if (!cleaningFilterEvent.isCanceled()) {
+                    for (NotificationFilterPreference matchingPreference : matchingPreferences) {
+                        this.notificationFilterPreferenceManager.deleteFilterPreference(user,
```
surli, line 335: Would that be ok or do we need an API for performing the delete in batch? I have the feeling it should be ok: in theory we shouldn't get that many filters preferences to delete...

surli: I actually provided the API to delete in batch.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R160000000XWIKI21448DataMigration.java
```diff
@@ -0,0 +1,222 @@
...
+import com.xpn.xwiki.store.migration.hibernate.AbstractHibernateDataMigration;
+
+/**
+ * Migration for cleaning up remaining notification filters that concern deleted documents.
+ *
+ * @version $Id$
+ * @since 16.0.0RC1
+ */
+@Component
+@Named("R160000000XWIKI21448")
```
surli, line 60: I'll change the name / version whenever I'm sure which version we target.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterPreferenceManager.java
```diff
@@ -169,16 +169,22 @@ public void saveFilterPreferences(DocumentReference user, Set<NotificationFilter
...
-                    NotificationFilterPreferenceProvider.class)) {
-                provider.deleteFilterPreference(user, filterPreferenceId);
+                : componentManager.<NotificationFilterPreferenceProvider>getInstanceList(
+                NotificationFilterPreferenceProvider.class)) {
+                provider.deleteFilterPreferences(user, filterPreferenceIds);
             }
-
         } catch (ComponentLookupException e) {
-            logger.info("Failed to remove the user filter preference [{}].", filterPreferenceId, e);
+            logger.info("Failed to remove the user filter preference [{}].", filterPreferenceIds, e);
```
tmortagne, line 187: You might want to update a bit the message (plural, and you already get `[]` around a `Set.toString`).

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DeletedDocumentCleanUpFilterProcessingQueue.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private final BlockingQueue<CleanUpFilterData> cleanupQueue = new LinkedBlockingQueue<>();
+
+    private boolean disposed;
+
+    @Override
+    public void dispose() throws ComponentLifecycleException
+    {
+        this.disposed = true;
+        this.cleanupQueue.clear();
```
tmortagne, line 99: Doesn't this mean that some filters which should be deleted will never be ?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterPreferenceManager.java
```diff
@@ -118,6 +119,22 @@ Stream<NotificationFilterPreference> getFilterPreferences(
...
+     * Delete a set of filter preferences.
+     * @param user the user to use
+     * @param filterPreferenceIds ids of the filter preferences
+     * @throws NotificationException if an error happens
+     *
+     * @since 15.10.2
+     * @since 16.0.0RC1
+     */
+    @Unstable
+    default void deleteFilterPreferences(DocumentReference user, Set<String> filterPreferenceIds)
```
tmortagne, line 132: Wouldn't `Collection<String>` be enough for the `filterPreferenceIds` parameter?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DeletedDocumentCleanUpFilterProcessingQueue.java
surli: That's correct, now not sure how to handle this without redoing what we have for TaskManager.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/NotificationFilterPreferenceManager.java
surli: Hmmm I prefer that the API shows that there shouldn't be any duplicate of ids there: I doubt it would be really an issue, but frankly not entirely sure how a `IN ()` clause for a delete would handle duplicate ids.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DeletedDocumentCleanUpFilterProcessingQueue.java
tmortagne: Why not use TaskManager ?

tmortagne: Could worth putting a comment about those design choices.

surli: I also need a ticket for the actual improvment of TaskManager as IMO it would be useful (would allow also to get rid of the dedicated mechanism for live notification emails: we could use it too, and we wouldn't need a dedicated mechanism for missing emails). 
I guess whenever we do that we should again perform a cleanup of possibly remaining filters though, since as you mentioned we could have some never clean up.

tmortagne: >  and we wouldn't need a dedicated mechanism for missing emails

I don't think it makes that much sense as this "dedicated mechanism" is not that dedicated and is relying on the same boolean used by the non-live emails notifications (hourly, daily, etc.) stored in the events core to know if a notification has been sent already or not.

surli: I opened https://jira.xwiki.org/browse/XWIKI-21692 so we could move that discussion there

PR #2694, opened 2023-12-06
XWIKI-21492: Underline inline links

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21492
## PR Changes
* Removed the 'extra accessibility' setting
* Split it into two more refined accessibility settings: underlining links and default-size
* Updated the default value for underlining links from 'No' to 'Inline only'

## Notes
* This PR implements the changes as discussed on the forum. This means giving users a new preference to toggle this change.
* Adding LESS dynamically (based on user) is not possible as of now (we'd need to recompile the whole skin for each new user, or expand the cache to contain all combinations of preferences...) so I relied on the former way we had to change the size of text. This way is flawed because some text size is 'hard-coded' in pixels and this method relies on inheritance. For example, this won't work on headings (see the screenshots below).
* The selector for inline links underlining is not perfect, but from testing it out on a few pages, it worked properly. Unfortunately we couldn't find a perfect selector for this yet.Some fine tuning might be needed later for some niche UIs and use cases.
* Some of the CSS in accessibility.css looked deprecated since we don't support Colibri anymore. I cleaned it up and kept only what was necessary in the new stylesheet extension. E.g. it made superfluous use of `!important` that are now against our codestyle.
* The commits in this PR can be squashed
* The solution proposed in this PR, if accepted, should be cherry-picked for 15.10.x, as this ticket is a part of the cycle 15 roadmap

## View
Here is the new default: Only inline links underlined and regular font size
21492-defaultHome
Hereafter are a few of the variations:
Only inline links underlined and large font size
21492-largeHome
Only inline links underlined and larger font size
21492-largerHome
Only inline links underlined and largest font size
21492-largestHome
No link underlined and regular font size (same as the default before this PR)
21492-noUnderlineHome
All links underlined and regular font size
21492-allUnderlineHome

Hereafter are the view and edit modes of the updated user preference section:
21492-viewPreferences
21492-editPreferences


michitux (changes requested):
Thank you very much for working on this. Unfortunately, I think this needs significantly more work before it can be merged. Apart from the comments I already added, there are the following issues:

1. We need a migration to enable both options by default for users who had the old option enabled. I see two options how this could be done:
   1. By reading the value of the old option (the value won't be removed when it has been modified) and enabling the CSS also in this case and also setting the options by default to a value that corresponds to the old behavior when editing the profile.
   2. By using a proper migration that migrates the configuration on the upgrade. I don't know if we have any examples/best practices for this.
2. Inline links are still ill-defined and won't lead to a consistent experience. For example, consider this example content: image 
here, the first link would be underlined, but the second won't, unless I'm mistaken. I'm really not happy with this inconsistency, in particular if this is our default behavior.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -702,10 +702,11 @@ platform.core.profile.phone=Phone
 platform.core.profile.address=Address
 platform.core.profile.editor=Default editor to use
 platform.core.profile.userType=User Type
-platform.core.profile.enableAccessibility=Enable extra accessibility features
 platform.core.profile.displayHiddenDocuments=Display hidden pages
 platform.core.profile.timezone=Timezone
 platform.core.profile.extensionConflictSetup=Enable extension conflict setup
+platform.core.profile.accessibility.underline=Underlining
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/344d906ec30c9bb5a6409119e715a1d233aa5931 :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/stylesheets.vm
```diff
@@ -46,13 +46,8 @@ $originalUrl?$queryString## prevent line break in output
...
 <link href="#concatenateQueryParameter($escapetool.xml($xwiki.getSkinFile('print.css', true)), $urlParameters)" rel="stylesheet" type="text/css" media="#if ($printss)all#{else}print#{end}" />
-#set ($a11y = "$!{request.getCookie('a11y').getValue()}")
-#if ($a11y == '')
-  #set ($a11y = "$!{xwiki.getUserPreference('accessibility')}")
-#end
-#if ($a11y == '1')
-  <link href="$escapetool.xml($xwiki.getSkinFile('css/accessibility.css', true))" rel="stylesheet" type="text/css" media="all" />
-#end
+## User preferences stylesheet
+#set($discard = $xwiki.ssx.use('XWiki.XWikiUserPreferencesSheet'))
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/commit/344d906ec30c9bb5a6409119e715a1d233aa5931 👍

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/internal/mandatory/XWikiUsersDocumentInitializer.java
```diff
@@ -101,7 +101,8 @@ protected void createClass(BaseClass xclass)
         xclass.addTextField("imaccount", "imaccount", 30);
         xclass.addStaticListField("editor", "Default Editor", "Text|Wysiwyg");
         xclass.addStaticListField("usertype", "User type", "Simple|Advanced", "Simple");
-        xclass.addBooleanField("accessibility", "Enable extra accessibility features", "yesno");
+        xclass.addStaticListField("font_size", "Font size", "Regular|Large|Larger|Largest", "Regular");
+        xclass.addStaticListField("underline", "Underline links", "Yes|Only inline links|No", "Only inline links");
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/commit/344d906ec30c9bb5a6409119e715a1d233aa5931 👍

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
```diff
@@ -37,7 +37,8 @@
   <syntaxId>plain/1.0</syntaxId>
   <hidden>true</hidden>
   <content>XWiki.XWikiUsers_displayHiddenDocuments.hint=The wiki contains documents that are not displayed by default. These hidden documents represent technical content, like application classes, configuration pages, macros, styles, scripts, etc.
-XWiki.XWikiUsers_accessibility.hint=Extra accessibility will enable various visual enhancements like: bigger fonts, underlined links, etc.
+XWiki.XWikiUsers_font_size.hint=Change the font size for the main text (in px).
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/344d906ec30c9bb5a6409119e715a1d233aa5931 and https://github.com/xwiki/xwiki-platform/pull/2694/commits/acb4768891ac511b1e472bd2ca606e5f1eaa6545 :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
```diff
@@ -37,7 +37,8 @@
...
+XWiki.XWikiUsers_underline.hint=Set your underline preferences. Default is to underline only inline links.
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/344d906ec30c9bb5a6409119e715a1d233aa5931 and https://github.com/xwiki/xwiki-platform/pull/2694/commits/acb4768891ac511b1e472bd2ca606e5f1eaa6545 :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,169 @@
...
+#if ($underlining == 'No')
+  a {
+    text-decoration: none;
+  }
+#end
+
+#set ($fontSizePreference = $!{xwiki.getUserPreference('font_size')})
+## Default font size is 14px, which is equivalent to 140%
+#if ($fontSizePreference == 'Large')
+  .skin-flamingo {
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/b4e00a6f4664b1fd538098ab9d5625e3e78f1ccf :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,169 @@
...
+  }
+#end
+#if ($underlining == 'Only inline links')
+  p, span, div {
+    &amp; &gt; .wikilink,
+    &amp; &gt; .wikiexternallink,
+    &amp; &gt; .wikiinternallink,
+    &amp; &gt; .wikicreatelink {
+      &amp; &gt; a {
+        text-decoration: underline;
```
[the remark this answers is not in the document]
Sereza7: I reused the selector we converged towards at the end of this previous PR. It can definitely still be improved, I'll try to make it a bit better, but from what I can remember there's no perfect solution because we can't 'see' text nodes in CSS.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,169 @@
...
+        <separator> </separator>
+        <separators>|, </separators>
+        <size>1</size>
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>default</cache>
```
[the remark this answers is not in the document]
Sereza7: The line at https://github.com/xwiki/xwiki-platform/pull/2694#discussion_r1440556427 was in charge of reloading the SSX no matter the cache. I updated the cache policy to account for removing this line, disabling cache for this file.
Ideally we should only invalidate cache for this skin extension when the user preference change, but technically it's not that easy.
Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/6b9d08c8463c9d101f8f0bf968f926e481edc0dd :+1: 
See https://up1.xwikisas.com/#i_R7F8zyuo9LADkeBg-vgw for a demo. In this demo, I just changed the font size since it's the easiest to spot out quickly, but it works the same for the link underlining.

Sereza7:
@michitux 
> We need a migration to enable both options by default for users who had the old option enabled.

I'm not sure this is so important, from my understanding, barely anyone used it, and those who did can easily change it back to what they used to have (and that's an incentive to find out about the new granularity in the preferences).

Nevertheless, I added logic to take into account the old parameter if it's set in https://github.com/xwiki/xwiki-platform/pull/2694/commits/7e5d8f0c05250722a2cb873d059987656bf0e493.

I'm wondering what would happen for users who had it enabled though:
They had it mostly for the increased font-size, so they reconfigure and now they have 'No' to underline links. However there's no editor for the `accessibility` property, so they get stuck with underlining despite their change. Would the `accessibility` preference be discarded at some point? I'm worried that this change might bring more complications than what it's worth...

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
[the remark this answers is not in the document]
Sereza7: See https://github.com/xwiki/xwiki-platform/pull/2694#issuecomment-1878752565

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -287,7 +450,19 @@
       &lt;h1&gt;$escapetool.xml($services.localization.render('platform.core.profile.section.displayPreferences'))&lt;/h1&gt;
       &lt;dl&gt;
         #displayField('displayHiddenDocuments', 'platform.core.profile.displayHiddenDocuments', '0')
-        #displayField('accessibility'         , 'platform.core.profile.enableAccessibility'   , '0')
+      &lt;/dl&gt;
+      &lt;h1&gt;$escapetool.xml($services.localization.render('platform.core.profile.section.accessibilityPreferences'))&lt;/h1&gt;
+      &lt;dl&gt;
+        #set ($defaultValue = $xwiki.getXWikiPreference('font_size'))
```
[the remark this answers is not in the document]
michitux: >  From what I understand, the value comes from the edited user preferences.

No, this loads a value from the wiki configuration, i.e., it reads values from the `XWiki.XWikiPreferences` document - which has an `accessibility` field, but no `font_size` or `underline`. Actually, I think `$!{xwiki.getUserPreference('accessibility')}` reads this `accessibility` field from the wiki (or main wiki) configuration, too, when it is not set in the user profile/subwiki.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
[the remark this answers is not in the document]
michitux: I haven't checked what kind of elements we have normally as parents of links, but in general this approach seems very fragile and error-prone to me. I would have preferred an approach where we underline links with a broad, generic selector (to, e.g., generically match content and panels) and then use specific selectors to remove the underline where we don't want it. Just to avoid that we have cases where users don't understand why one link in the content is underlined and another isn't.

This is my opinion, I don't know how others see this and if they would prefer either that potentially some links in the content aren't underlined or that potentially too many links are underlined, but that underlining is mostly consistent.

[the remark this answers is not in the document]
Sereza7: I updated the selector in https://github.com/xwiki/xwiki-platform/pull/2694/commits/d2941bf81fd0aa5a64a36b44d014569f87dac5c7 . 
I've split it in different sections with easy objectives:
1. Ruleset for all links in content.
2. Exceptions to add underlining outside of content
3. Exceptions to remove underlining inside content
4. Helper classes to "force" the decoration on an anchor. Note that this class only makes a difference with the `underline inline link` preference chosen. With `all links underlined` or `no link underlined`, the user preference is predominant.

The advantage of this is that it's more consistent and easier to understand than the previous one. The disadvantage is that it's not really consistent with the `inline link` definition. We might need to update this selector soon to address leftover accessibility violations.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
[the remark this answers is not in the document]
Sereza7: In regards to that, I added the values in the Xwiki and space preferences in https://github.com/xwiki/xwiki-platform/pull/2694/commits/32a7cca54e4e86417d7ece0695acca9ef097ae25 .
:+1:

Sereza7:
Addressed the topics: [value origin](https://github.com/xwiki/xwiki-platform/pull/2694#discussion_r1440695874) and [inline link selector](https://github.com/xwiki/xwiki-platform/pull/2694#issuecomment-1878752565).

There's still the [backward compatibility](https://github.com/xwiki/xwiki-platform/pull/2694#issuecomment-1878774819) discussion that would need an approval from @michitux before considering merging this PR.

michitux:
There was an [interesting talk about accessibility at FOSDEM](https://fosdem.org/2024/schedule/event/fosdem-2024-2638-breaking-barriers-content-management-systems-and-accessibility/) that argued that websites should respect the user's settings in the browser instead of introducing new controls. There was an explicit question at the end if it would be bad for a web application to introduce a font size control and the presenter kind of agreed that it would be better to just respect the font size that the user has configured in the web browser. Also, all web browsers allow changing the font size/zooming the website. Thinking about it again, I wonder if it wouldn't really be better to just respect what the user set in the browser's settings and make sure that the browser's zoom feature works properly in XWiki.

Sereza7:
I updated the base font-size strategy in https://github.com/xwiki/xwiki-platform/pull/2694/commits/7bf473e9b73151cdc7e18894bc4305d9b3eca237

Here is what it looks like in a few situations (Sandbox page with a bit of custom content, 2560*1440 display, 100% zoom level, Firefox):
Regular xwiki user preference, default (16px) browser font-size
21452-DWDB
Regular xwiki user preference, 22px browser font-size
21492-DWLB
Largest xwiki user preference, default (16px) browser font-size
21492-LWDB
Largest xwiki user preference, 22px browser font-size
21492-LWLB

Sereza7:
Discussion started on https://forum.xwiki.org/t/deprecating-the-font-size-base-colortheme-variable/14005 , opening back this PR for reviews when the discussion ends.

Sereza7:
@michitux 
Following the discussion we had at FOSDEM, I took your idea and implemented preferences as a `body` class.
Now, instead of re compiling and loading an extra LESS at each page load, the logic is used directly in the velocity template, and style is part of the style.css bundle.
Those changes were made in https://github.com/xwiki/xwiki-platform/pull/2694/commits/3ae794da3f9ce848adddb013e35ae76f052e8d32

Here is a demo of the updated code. As expected, this works as well as before, with the added advantage that we don't have to mess around with the cache of a SSX, so presentation reliably updated.
https://youtu.be/q1satswfwbY

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
```diff
@@ -129,7 +129,42 @@
 #if($xcontext.action == "admin" || $xcontext.action == "import" || $xcontext.action == "export")
   #set($bodyAction = "view")
 #end
+## Query user preferences for UI customization
+## Get the legacy preference for backward compatibility
+#set ($a11y = "$!{xwiki.getUserPreference('accessibility')}")
```
michitux, line 134: Is there any code to clear that legacy preference? I wonder if the code shouldn't consider this setting only if the user preferences have never been saved with the new settings and make sure that the user profile takes the old value into account for initializing the preference.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -287,7 +287,19 @@
...
+        #if ("$!defaultValue" == '')
+          #set ($defaultValue = 'Regular')
+        #end
+        #displayField('font_size', 'platform.core.profile.accessibility.fontSize', $defaultValue)
+        #set ($defaultValue = $xwiki.getXWikiPreference('underline'))
+        #if ("$!defaultValue" == '')
+          #set ($defaultValue = 'OnlyInlineLinks')
+        #end
+        #displayField('underline', 'platform.core.profile.accessibility.underline', $defaultValue)
```
michitux, line 302: What I would suggest here is a) an initialization of the new settings if the legacy property is set and b) a hidden input to clear the legacy property.

xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/WebPreferences.xml
```diff
@@ -1232,6 +1256,30 @@
...
+        <separators>|, </separators>
+        <size>1</size>
+        <sort>none</sort>
+        <tooltip/>
+        <unmodifiable>0</unmodifiable>
+        <validationMessage/>
+        <validationRegExp/>
+        <values>OnlyInlineLinks|Yes|No</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </underline>
```
michitux, line 1282: I wonder if we shouldn't also remove the legacy accessibility property here.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
```diff
@@ -131,4 +131,81 @@ img {
...
   hyphens: auto;
+}
+
+// User preferences ===========================================================
+body.content {
+  /* We apply various user preference related styles to the content of the page.
+   We use classes that are added to the body itself in htmlheader.vm */
+  // XWiki font size preference
+  &.preference-font-size-regular {
+    font-size: 140%;
```
michitux, line 142: I would have left the font size unchanged in this case, also to avoid introducing any regressions. For example, I see that when setting the font size in the theme customization, this value is also applied to the body element. So when you set the font size, e.g., to `20px` in the theme customization, you'll have a `font-size: 20px` on `body`. This CSS would completely break this. Further, also all other font sizes are not relative to the font size set in the theme customization. When the font size in the theme is set to `20px`, all these settings apart from the `largest` setting would actually decrease the font size.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
```diff
@@ -131,4 +131,81 @@ img {
...
+  &.preference-font-size-large {
+    font-size: 160%;
+  }
+
+  &.preference-font-size-larger {
+    font-size: 180%;
+  }
+
+  &.preference-font-size-largest {
+    font-size: 200%;
```
michitux, line 154: I'm wondering if 200% isn't breaking the layout too much currently.

michitux (changes requested):
Overall, this looks much better already but I have some more comments in particular regarding the backwards compatibility handling.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
Sereza7: > Is there any code to clear that legacy preference?

Not as of now. I decided to leave it in so that nothing changes for users if they never update their accessibility preferences.

> I wonder if the code shouldn't consider this setting only if the user preferences have never been saved with the new settings and make sure that the user profile takes the old value into account for initializing the preference.

When the user updates preferences, we can't be sure that the accessibility ones are the ones updated. I don't think we want to fill in the accessibility fields with values different from default when the user edits its default editor ... IMO this would be a pretty deceptive pattern and leaving the old value in makes it easy both user experience and for development. We just need to set a note in documentation about this difference in behaviour in between seemingly and things should be okay.

xwiki-platform-distribution/xwiki-platform-distribution-ui/xwiki-platform-distribution-ui-base/src/main/resources/XWiki/WebPreferences.xml
Sereza7: To be honest I'm not sure what's the exact purpose of this file (seems like a repeat with the content of XWikiPreferences.xml ), so I'd rather not break unknown things by only adding what seems to make sense adding and leaving the legacy content unchanged.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: > I would have left the font size unchanged in this case, also to avoid introducing any regressions. For example, I see that when setting the font size in the theme customization, this value is also applied to the body element. 

Okay, so I just checked, and bootstrap uses two default hard coded values at different levels :/
There's a `10px` on the `html` node, and `font-size: @font-size-base;` on the body node(with its value expected to be in px).
This makes little sense to me and IMO that's an issue with bootstrap 3.
The way things are now, the 10px on the html node is completely useless...

> So when you set the font size, e.g., to 20px in the theme customization, you'll have a font-size: 20px on body.

With the solution above, the theme customization value would be overriden by the percentage which has more specificity. 
Our font-sizes are defined in multiple ways:
* percentage, rems and ems: depend on the relative font size (native CSS)
* `0.9*@font-size-base` : depend on the bootstrap value.

The only point of setting `@font-size-base` on the body is to allow using relative units without breaking presentation.
So I guess the way I should implement this is with a percent on the bootstrap value... However I feel like even this solution would break some UIs somehow...

This is an overly complex solution and I dislike it a lot. I'll make a proposal to start uniformizing our code base font-sizes so that we don't get stuck in an interwoven web of dependencies going through ten style files and two fundamentally incompatible models next time we try to make a system change relative to font sizes.

I think the best solution for long term would be to override bootstrap code with something that makes sense with the modern CSS capabilities, this would provide code that's easier to understand and maintain on the long term.

---
In my opinion we should aim at:
* The only absolute units for font sizes are used on `:root` (or `html`).
* Every other font size should be defined using percentages , ems or rems

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: I updated it to `1.4*@font-size-base`. 
Note that this 200% wasn't actually 200% of the default size but 200/140, which was only a 42% increase.

With a `@font-size-base` of 20px (above default), zoom level 100%, on a 2k display I get:
21492-largestNew
I used the Firefox device size emulation dev tool to get this display in the same conditions for a 1920*1080 screen:
21492-largest19201080


IMO this looks okay, a bit off in some places but not breaking the layout.
Note that for most users, the text will be smaller than that, this is a worst case scenario.

Sereza7: This was done with the changes pushed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/09af3b4516658f4e8b3356a9c64fce626a0d607b

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: The changes described here have been pushed on this PR in https://github.com/xwiki/xwiki-platform/pull/2694/commits/09af3b4516658f4e8b3356a9c64fce626a0d607b :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: See https://github.com/xwiki/xwiki-platform/pull/2694#discussion_r1521821029 :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
michitux: I absolutely agree with you that we need to get rid of the current font size system and use one that is based on either a single font size or even the browser's font size. However, the problem is that this font size that is set by bootstrap is not only used for computing other font sizes but also for paddings and margins (search for `@line-height-computed` in the source code).

However, this will be a lot of work and I think we should first decide if we want to stay with Bootstrap 3 or not before we put more work into modernizing Bootstrap 3.

Ideally, the font size settings shouldn't be part of this PR. However, I see the problem that you need to split the accessibility settings in order to offer several options for underlining links.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
michitux: To me the whole spacing just looks completely wrong. Also, the fact that not all text is scaled and in particularly the smallest text is not scaled at all makes this feature kind of useless to me.

I'm wondering if we shouldn't just have a vote to drop this font size increase feature in favor of pointing users to the browser's zoom functionality until we have proper font scaling support.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
michitux: > Ideally, the font size settings shouldn't be part of this PR. However, I see the problem that you need to split the accessibility settings in order to offer several options for underlining links.

As I wrote in my previous comment, maybe it would make sense to just drop the font size feature for now (requires a vote) and to simply replace the accessibility setting by the underline links setting.

Sereza7: > However, the problem is that this font size that is set by bootstrap is not only used for computing other font sizes but also for paddings and margins (search for @line-height-computed in the source code).

Thank you for highlighting this, I did not connect the dots here. It will be a pain moving, since the basic semantics of this system look skewed. This is kind of similar to a regression we had earlier this month because we used a foreground color as a background :) .

> However, this will be a lot of work and I think we should first decide if we want to stay with Bootstrap 3 or not before we put more work into modernizing Bootstrap 3.

From what I understood, the only reason why we stay on bootstrap 3 is the cost of migration (without it I'm pretty sure we'd have moved to the latest bootstrap in an heartbeat :) ). My idea would be to be less reliant on bootstrap 3 for font-size, which would make it easier to move to another system later.

> Ideally, the font size settings shouldn't be part of this PR. However, I see the problem that you need to split the accessibility settings in order to offer several options for underlining links.

:+1: It's a bit of a pain but it's something I would have added not long after anyways.

Sereza7: > I would have left the font size unchanged in this case, also to avoid introducing any regressions. For example, I see that when setting the font size in the theme customization, this value is also applied to the body element. So when you set the font size, e.g., to `20px` in the theme customization, you'll have a `font-size: 20px` on `body`. This CSS would completely break this. Further, also all other font sizes are not relative to the font size set in the theme customization. When the font size in the theme is set to `20px`, all these settings apart from the `largest` setting would actually decrease the font size.

By the way I just realized that the legacy way to implement the `accessible` mode also overrode the `@font-size-base` defined in the colortheme and bootstrap :)

michitux: My comment here was only regarding the "regular" size, "this case" meant the default "regular" font size. My worry here was primarily that the configured font size should be respected and left unchanged as much as possible when the font size setting is on the default value.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: I figured out why... The values with `0.9*@font-size-base` would not be modified by this...
E.g. the `last modified` is the same size whatever the preference.

We'd need to have the value of font-size-base change based on the preferences, but AFAIK this is not something we want to do because we would need to recompile all the LESS of the codebase for each user... which was the exact thing we tried to avoid by using a class on the body node...


So :+1: to start a vote, and implement this font size feature back when we have a codebase consistent enough to make it possible.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: The point that the accessible font size could end up smaller than the regular one was already a thing though

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
michitux: I think what might be worth investigating is if we could replace the less variable `@font-size-base` by a CSS variable without changing the whole logic too much. This might allow easily changing the base font size with full effects everywhere without needing any less compilation.

Sereza7: So, I just checked it out, and LESS is not really made with that in mind...
First, I came up with escaping issues, the `--` used to define a CSS variable were interpreted as the start of `-->` and failed the LESS compilator, so I had to escape all of those characters:
```
#elseif($property == 'font-size-base')
  ## Special case, we rely on a CSS variable to control the value so that we can update it with more ease.
  :root {\-\-$property: $value;}
  @$property: var(\-\-$property);
```
in variablesInit.vm
and
```
&.preference-font-size-regular {
  \-\-font-size-base: calc(1*var(\-\-font-size-base));
}
```
in general.less

However, this workaround was not enough, because after that compilation failed with the message:


> Caused by: com.github.sommeri.less4j.Less4jException: Could not compile less. 270 error(s) occurred:
> ERROR less/bootstrap/variables.less 67:34 Wrong argument type to function 'floor', expected number saw LIST_EXPRESSION.
>  66: //** Computed "line-height" (`font-size` * `line-height`) for use with `margin`, `padding`, etc.
>  67: @line-height-computed:    floor((@font-size-base * @line-height-base)); // ~20px
>  68: 

Bootstrap uses the value of the @font-size-base to calculate the value of @line-height-computed, but uses LESS syntax to do so. This syntax is not compatible with CSS variables :disappointed:

Sereza7:
Started a vote at https://forum.xwiki.org/t/drop-the-font-size-increase-preference/14241

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
michitux: > Bootstrap uses the value of the @font-size-base to calculate the value of @line-height-computed, but uses LESS syntax to do so. This syntax is not compatible with CSS variables 😞

My idea would have been to also replace `@line-height-computed` by a variable that is initialized as CSS `calc()` expression.

Sereza7:
Vote ended with two +0.
Opening this PR back up with the font-size preference removed.
If anyone disagrees with the content of the changes here after they are merged, we should revert them.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -3462,6 +3463,15 @@ XWiki.XWikiUsers_usertype_Advanced=Advanced
...
 XWiki.XWikiUsers_extensionConflictSetup=Enable extension conflict setup
+XWiki.XWikiUsers_underline=Underline links
+XWiki.XWikiUsers_underline_Yes=Yes
+XWiki.XWikiUsers_underline_OnlyInlineLinks=Only Inline Links
+XWiki.XWikiUsers_underline_No=No
+XWiki.XWikiUsers_fontSize=Font size
+XWiki.XWikiUsers_fontSize_Regular=Regular
+XWiki.XWikiUsers_fontSize_Large=Large
+XWiki.XWikiUsers_fontSize_Larger=Larger
+XWiki.XWikiUsers_fontSize_Largest=Largest
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/674b74fa79f5182e1f8b6bafe37bde3ebd08a507 :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
```diff
@@ -37,7 +37,7 @@
   <syntaxId>plain/1.0</syntaxId>
   <hidden>true</hidden>
   <content>XWiki.XWikiUsers_displayHiddenDocuments.hint=The wiki contains documents that are not displayed by default. These hidden documents represent technical content, like application classes, configuration pages, macros, styles, scripts, etc.
-XWiki.XWikiUsers_accessibility.hint=Extra accessibility will enable various visual enhancements like: bigger fonts, underlined links, etc.
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/5c627bce591d4da1abbec9af0a71fd6ee4695c8a :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
```diff
@@ -129,7 +129,28 @@
...
+#set ($underlining = "$!{xwiki.getUserPreference('underline')}")
+## Underlining of links
+## This preference will take the values: preference-underlining-yes, preference-underlining-only-inline-links and
+## preference-underlining-no
+#set ($preferenceUnderlining = "preference-underlining-")
+#if ($underlining == 'Yes' || $a11y == '1')
+  #set ($preferenceUnderlining = $preferenceUnderlining + "yes")
+#elseif($underlining == 'OnlyInlineLinks' && $a11y != '1')
```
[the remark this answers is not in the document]
Sereza7: Why couldn't it be 1 here?

From what I understand this is pretty much just a repeat of what was in `xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/stylesheets.vm` .
We didn't remove the accessibility parameter but only made it disappear from the UI, so it cannot be updated anymore, but it'll still be here as far as I can understand.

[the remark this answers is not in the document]
michitux: I'm referring to this specific line 142 in my comment. If `$a11y` is `1`, the first branch is taken. Thus, in the `#elseif` the condition `$a11y != '1'` is always `true` and can be removed.

Sereza7:
@michitux 

> I wonder if we shouldn't really remove the accessibility preference from XWiki.XWikiPreferences. 

Removing it could be considered a backward compatibility breakage. If a custom content relies on this preference to provide its own accessibility hacks, it would completely break it. IMO we should at least keep it in the backend for a while, even if we remove all the UI related to it and override its default behavior with new preferences. Eventually if we find out that this first step didn't stir any trouble, we could consider removing it completely. 
See https://github.com/xwiki/xwiki-platform/pull/2694#issuecomment-1878774819

> And then of course this probably needs more testing to find all the edge cases where links are underlined that shouldn't be underlined.

Some additionnal manual testing is necessary :+1: I looked at a few pages but for obvious reasons I couldn't try all use cases. I'll keep an eye out for those once it's merged, and put details about this situation in a release note for devs. IMO what we have now is good enough to solve the ticket and anything we report afterward can have its own issue (hopefully nothing critical enough to be categorized as a bug :) ).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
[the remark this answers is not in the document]
Sereza7: Okay, I think I left it there for the sake of a better understanding, but this can be achieved with a comment instead of actual code :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/htmlheader.vm
```diff
@@ -129,7 +129,28 @@
...
+  #set ($preferenceUnderlining = $preferenceUnderlining + "only-inline-links")
+#elseif($underlining == 'No' && $a11y != '1')
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2694/commits/df185ba1ffcb8b72aba5ce62836549cfdae66929 :+1:

michitux (approved):

PR #2695, opened 2023-12-06
XWIKI-21670: The versionPreserved instance output filter stream property defaults to false

raphj:
Discussed in an internal chat with @tmortagne 

Not sure about two things:
 - should we put EntityInstanceOutputProperties in an internal package?
 - this is not actually an api breakage, as I stated in the revapi.differences breakage. I think I'm in the same case as reported by Vincent at https://github.com/revapi/revapi/issues/235 but a double check on this would be nice.

Please do not squash, the two commits are separated intentionally.

### DIFF
--- a/xwiki-platform-core/pom.xml
+++ b/xwiki-platform-core/pom.xml
@@ -166,6 +166,18 @@
                 </item>
               </differences>
             </revapi.differences>
+            <revapi.differences>
+              <differences>
+                <item>
+                  <criticality>allowed</criticality>
+                  <code>java.class.nonFinalClassInheritsFromNewClass</code>
+                  <old>class org.xwiki.filter.instance.output.DocumentInstanceOutputProperties</old>
+                  <new>class org.xwiki.filter.instance.output.DocumentInstanceOutputProperties</new>
+                  <superClass>org.xwiki.filter.instance.output.EntityInstanceOutputProperties</superClass>
+                  <justification>Not a breakage, the old DocumentInstanceOutputProperties content has been moved to the new EntityInstanceOutputProperties class and old class now inherits from the new one</justification>
+                </item>
+              </differences>
+            </revapi.differences>
           </analysisConfiguration>
         </configuration>
       </plugin>
--- a/xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/src/main/java/org/xwiki/filter/instance/output/EntityInstanceOutputProperties.java
+++ b/xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/src/main/java/org/xwiki/filter/instance/output/EntityInstanceOutputProperties.java
@@ -1,0 +1,120 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.filter.instance.output;
+
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.properties.annotation.PropertyDescription;
+import org.xwiki.properties.annotation.PropertyName;
+import org.xwiki.stability.Unstable;
+
+/**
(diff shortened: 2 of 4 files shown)


xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/pom.xml
```diff
@@ -40,4 +40,28 @@
...
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.revapi</groupId>
+        <artifactId>revapi-maven-plugin</artifactId>
+        <configuration>
+          <analysisConfiguration>
+            <revapi.differences>
```
surli, line 50: We don't put revapi ignores there, every ignore is going to `xwiki-platform-core` pom.xml: the reason is that we get them in the RN so we need them to be all at the same place.

surli: Also I think we don't use `ignore` anymore but we put a criticality see the example: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/pom.xml#L107

surli: See also: https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HRevapiIgnores

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/src/main/java/org/xwiki/filter/instance/output/EntityInstanceOutputProperties.java
```diff
@@ -0,0 +1,121 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.filter.instance.output;
+
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.properties.annotation.PropertyDescription;
+import org.xwiki.properties.annotation.PropertyName;
+
+/**
+ * @version $Id$
```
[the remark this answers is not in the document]
surli: Also it's unstable.

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/src/main/java/org/xwiki/filter/instance/output/EntityInstanceOutputProperties.java
```diff
@@ -0,0 +1,121 @@
...
+ * @since 16.0
```
tmortagne, line 28: 16.0.0RC1 (which reminds me I need to conclude https://forum.xwiki.org/t/always-have-3-parts-in-the-xwiki-standard-version/13577 and update the documentation).

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/src/main/java/org/xwiki/filter/instance/output/EntityInstanceOutputProperties.java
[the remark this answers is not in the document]
surli: > @surli does the documentation I added look good to you?

So generally since and version tags are put below the doc. Now for the content itself, well it's a bit repeating the name of the class but better than nothing :)

xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/pom.xml
surli: Oups, it seems that's for me to fix :)

surli: Ok you can rebase and fix the conflict you should be able to test it.

tmortagne: > [ERROR] - Component [org.xwiki.configuration.internal.PermanentConfigurationSource] is not declared in [/home/raph/dl/xwiki/xwiki-platform/xwiki-platform-core/xwiki-platform-configuration/xwiki-platform-configuration-default/target/classes/META-INF/components.txt]! Consider adding it or if it is normal use the "staticRegistration" parameter as in "@Component(staticRegistration = false)"

Definitely unrelated but also not true so it's strange you get that.

xwiki-platform-core/pom.xml
```diff
@@ -164,6 +164,14 @@
                   <code>java.method.removed</code>
                   <old>method java.lang.String com.xpn.xwiki.doc.XWikiDocument::displayTooltip(java.lang.String, com.xpn.xwiki.objects.BaseObject, com.xpn.xwiki.XWikiContext)</old>
                 </item>
+                <item>
+                  <criticality>allowed</criticality>
```
surli, line 168: That's not good, you put your breaking justification under my own justification. You need to add a `revapi.differences` block here and provide your own explanation.

tmortagne:
Merged, thanks !

PR #2700, opened 2023-12-07
XWIKI-15065: Replace Silk icons with Icon Themes values in attachments Upload component

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-15065
## PR Changes
* Added icons to the generated nodes and removed their unsemantic counterpart from css backgrounds

## Note
I decided to go with a CSS only logic: instead of creating and deleting the icons based on the status, we use the logic already in place to hide/show the icons (all created at the start). There's only 4 of them, and they are not generated on every page load, so I think this is an appropriate solution.

## View
Here is a screenshot from the issue report on Jira:
upload
Hereafter is a screeenshot of the same UI after the changes proposed in this PR
15065-afterPR

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.css
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.css
@@ -8,19 +8,28 @@
   background: $theme.highlightColor none 2px .8em no-repeat;
   margin: .5em 0;
   clear: both;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 0.2em 1em;
 }
-.upload-inprogress {
+
+.upload-status .status-icon {
+  display: none;
+}
+
+.upload-status.upload-inprogress .status-icon.icon-inprogress,
+.upload-status.upload-done .status-icon.icon-done,
+.upload-status.upload-canceled .status-icon.icon-canceled,
+.upload-status.upload-error .status-icon.icon-error {
+  display: inherit;
+}
+.status-icon.icon-inprogress {
+  height: 14px;
+  width: 17px;
   background-image: url("$xwiki.getSkinFile('icons/xwiki/spinner.gif')");
 }
-.upload-done {
-  background-image: url("$xwiki.getSkinFile('icons/silk/tick.png')");
-}
-.upload-canceled {
-  background-image: url("$xwiki.getSkinFile('icons/silk/cancel.png')");
-}
-.upload-error {
-  background-image: url("$xwiki.getSkinFile('icons/silk/delete.png')");
-}
+
 .progress-info, .file-info {
   font-size: .91em;
 }
@@ -39,7 +48,7 @@
   margin: 0;
 }
 .progress-info {
-  clear: right;
+  flex-basis: 100%;
   opacity: 0.5;
 }
 .upload-inprogress .progress-info {
@@ -71,7 +80,13 @@
   margin: 0.3em 0;
   width: 100%;
 }
-.upload-canceled .file-info .file-name, .upload-error .file-info .file-name {
+.upload-done .status-icon.icon-done {
+  color: $theme.notificationSuccessColor;
+}
+.upload-canceled .file-info .file-name,
+.upload-canceled .status-icon.icon-canceled,
+.upload-error .file-info .file-name,
+.upload-error .status-icon.icon-error {
   color: $theme.notificationErrorColor;
 }
 .upload-canceled .progress-info .progress-container {
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.css
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.css
@@ -13,11 +13,9 @@
   align-items: center;
   gap: 0.2em 1em;
 }
-
 .upload-status .status-icon {
   display: none;
 }
-
 .upload-status.upload-inprogress .status-icon.icon-inprogress,
 .upload-status.upload-done .status-icon.icon-done,
 .upload-status.upload-canceled .status-icon.icon-canceled,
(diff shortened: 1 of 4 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.js
```diff
@@ -121,7 +121,23 @@ var XWiki = (function(XWiki) {
...
+      statusUI.STATUS_ICON_INPROGRESS.insert(statusUI.STATUS_ICON_INPROGRESS_ALTERNATIVE);
+      statusUI.STATUS_ICON_DONE = UploadUtils.createDiv('status-icon icon-done', "$!escapetool.javascript($services.icon.renderHTML('check'))");
+      statusUI.STATUS_ICON_DONE_ALTERNATIVE = UploadUtils.createSpan('sr-only',"$services.localization.render('core.widgets.html5upload.status.icon.done')");
+      statusUI.STATUS_ICON_DONE.insert(statusUI.STATUS_ICON_DONE_ALTERNATIVE);
+      statusUI.STATUS_ICON_CANCELED = UploadUtils.createDiv('status-icon icon-canceled', "$!escapetool.javascript($services.icon.renderHTML('remove'))");
+      statusUI.STATUS_ICON_CANCELED_ALTERNATIVE = UploadUtils.createSpan('sr-only',"$services.localization.render('core.widgets.html5upload.status.icon.canceled')");
+      statusUI.STATUS_ICON_CANCELED.insert(statusUI.STATUS_ICON_CANCELED_ALTERNATIVE);
+      statusUI.STATUS_ICON_ERROR = UploadUtils.createDiv('status-icon icon-error', "$!escapetool.javascript($services.icon.renderHTML('error'))");
+      statusUI.STATUS_ICON_ERROR_ALTERNATIVE = UploadUtils.createSpan('sr-only',"$services.localization.render('core.widgets.html5upload.status.icon.error')");
```
michitux, line 137: All these localization messages needs escaping, otherwise this breaks as soon as somebody uses `"` in a translation message. Also, please always put a space after `,`.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2700/commits/3f54043c5f7bb3791aff186cd1e894842e4f28f9 :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.js
```diff
@@ -121,12 +121,28 @@ var XWiki = (function(XWiki) {
...
+      statusUI.STATUS_ICON_DONE.insert(statusUI.STATUS_ICON_DONE_ALTERNATIVE);
+      statusUI.STATUS_ICON_CANCELED = UploadUtils.createDiv('status-icon icon-canceled', "$!escapetool.javascript($services.icon.renderHTML('remove'))");
+      statusUI.STATUS_ICON_CANCELED_ALTERNATIVE = UploadUtils.createSpan('sr-only', "$!escapetool.javascript($services.localization.render('core.widgets.html5upload.status.icon.canceled'))");
+      statusUI.STATUS_ICON_CANCELED.insert(statusUI.STATUS_ICON_CANCELED_ALTERNATIVE);
+      statusUI.STATUS_ICON_ERROR = UploadUtils.createDiv('status-icon icon-error', "$!escapetool.javascript($services.icon.renderHTML('error'))");
+      statusUI.STATUS_ICON_ERROR_ALTERNATIVE = UploadUtils.createSpan('sr-only', "$!escapetool.javascript($services.localization.render('core.widgets.html5upload.status.icon.error'))");
+      statusUI.STATUS_ICON_ERROR.insert(statusUI.STATUS_ICON_ERROR_ALTERNATIVE);
+      statusUI.UPLOAD_STATUS.insert(statusUI.STATUS_ICON_INPROGRESS).insert(statusUI.STATUS_ICON_DONE).insert(statusUI.STATUS_ICON_CANCELED).insert(statusUI.STATUS_ICON_ERROR);
+
```
surli, line 140: Not great to mix velocity and javascript for translations. You should use the bestpractice documented in https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HJavaScriptBestPractices

Sereza7: So, I tried applying the best practice, but this is some Prototype.js and it broke things. 
Instead, I decided to group up all of these translations at the start as seen on https://github.com/xwiki/xwiki-platform/blob/6bbb277528374f801a3c256fd281234db844d9bb/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/importer/import.js#L24C1-L43 . See https://github.com/xwiki/xwiki-platform/pull/2700/commits/6af9113d21e1d16df648c8c06223be129326ef3e

Note that most of those translations were already there before the PR, they are changed to make escaping consistent in this file after https://github.com/xwiki/xwiki-platform/pull/2700#discussion_r1419340603

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.css
```diff
@@ -8,19 +8,25 @@
...
-.upload-inprogress {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 0.2em 1em;
+}
+.upload-status .status-icon {
+  display: none;
+}
+.upload-status.upload-inprogress .status-icon.icon-inprogress,
```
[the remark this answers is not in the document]
Sereza7: I did build the module successfully, but there's no integration tests in `xwiki-platform-web-war`. 

I'm pretty sure this CSS could not break any automatic test we have as of now, escept for the `display: none` part. But it's on a precise enough selector and it's only for non clickable elements so I'm pretty sure it won't break anything.

PR #2701, opened 2023-12-07
XWIKI-18971: Number property input does not take into account the current locale to parse floating numbers

surli:
Use a NumberFormat instance based on the locale of the owner document for parsing number values.
I also simplified a bit the code of NumberClass#fromString, started refactoring of NumberClassTest, but it's still missing lots of test cases.

Also it's still a WIP: right now we've several unit tests failing in oldcore because of this: probably because the owner document is now mandatory for this method. We might need a fallback in case no owner document is specified.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/NumberClass.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/NumberClass.java
@@ -19,6 +19,10 @@
  */
 package com.xpn.xwiki.objects.classes;
 
+import java.text.NumberFormat;
+import java.text.ParseException;
+
+import org.apache.commons.lang3.StringUtils;
 import org.apache.ecs.xhtml.input;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/objects/classes/NumberClass.java
```diff
@@ -126,28 +132,30 @@ public BaseProperty fromString(String value)
...
-                    nvalue = Double.valueOf(value);
-                }
-            } else {
-                if ((value != null) && (!value.equals(""))) {
-                    nvalue = Long.valueOf(value);
+        boolean valueBlank = StringUtils.isBlank(value);
+
+        if (!valueBlank) {
+            try {
+                NumberFormat numberFormat = NumberFormat.getNumberInstance(getOwnerDocument().getLocale());
```
tmortagne, line 139: Using `#getLocale()` does not really help as the xobject are located on the default document, which is always Local.ROOT.

`#getRealLocale()` should give better results

tmortagne:
Well, actually the safest would probably be to not change anything to the server side and only have the client side do number format conversion from `,` based to `.` based, depending on the browser locale.

PR #2704, opened 2023-12-08
XWIKI-14088: Ensure that implied rights are consistent

michitux:
* Add a new concept of rights that are needed for another right.
* Deny rights that are missing a needed right.
* Add unit tests.

Forum discussion: https://forum.xwiki.org/t/should-denied-view-right-deny-other-rights-or-be-overridden-by-other-rights/13557

I'm marking this as a draft as long as the forum discussion hasn't been concluded.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
+++ b/xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
@@ -33,6 +33,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.xwiki.model.EntityType;
+import org.xwiki.stability.Unstable;
 
 import static org.xwiki.security.SecurityReference.FARM;
 import static org.xwiki.security.authorization.RuleState.ALLOW;
@@ -99,7 +100,7 @@ public class Right implements RightDescription, Serializable, Comparable<Right>
         = EnumSet.of(EntityType.WIKI, EntityType.SPACE, EntityType.DOCUMENT);
 
     /** Serialization identifier. */
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 2L;
 
     /** Internal list of existing instances. */
     private static final List<Right> VALUES = new CopyOnWriteArrayList<>();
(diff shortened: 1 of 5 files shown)


xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
```diff
@@ -285,6 +295,32 @@ private Right(String name, RuleState defaultState, RuleState tieResolutionPolicy
         }
     }
 
+    /**
+     * @param neededRights the rights required for this right to be granted
+     * @since 15.10.2
+     */
+    @Unstable
+    public void setNeededRights(Set<Right> neededRights)
```
surli, line 303: I find it a bit weird that it's a setter and not part of the constructor. What does it mean to call that method after the rights are computed? E.g. I could have in theory a component performing `Right.EDIT.setNeededRights(Set.of())` at any moment and calling a cache invalidation to recompute the rights: would that have an impact? 
Looks like the set of needed rights should be final no?

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/internal/AbstractAuthorizationSettler.java
```diff
@@ -245,6 +246,37 @@ protected XWikiSecurityAccess applyDefaults(UserSecurityReference user, Security
...
+        // Check needed rights.
+        boolean rightsModified;
+        do {
+            rightsModified = false;
+            for (Right right : Right.values()) {
+                // For every right that is allowed, check if all needed rights are allowed.
+                if (access.get(right) == RuleState.ALLOW) {
+                    for (Right neededRight : right.getNeededRights()) {
+                        if (access.get(neededRight) != RuleState.ALLOW) {
+                            access.set(right, RuleState.DENY);
```
surli, line 267: What about adding a debug log here to help in case of problem?

surli: I'm also wondering if there shouldn't be an hardcoded mechanism for PR right. Let's imagine that for some reason PR right is configured to need VIEW. Right now, you completely lock down a wiki from anyone including PR admins, with this loop.

xwiki-platform-core/xwiki-platform-security/xwiki-platform-security-authorization/xwiki-platform-security-authorization-api/src/main/java/org/xwiki/security/authorization/Right.java
tmortagne: I'm also really not a fan of making Right publicly mutable, especially when I don't see any use case for it.

surli: > My point was to not add another constructor parameter, in particular as we already have too many.

I agree that it's not nice to add another parameter to the constructor: what we can maybe do is adding a new constructor with a map String object which would contain all properties that are not mandatory.

tmortagne: Usually, a constructor becoming too big is when RightBuilder tends to appear.

PR #2715, opened 2023-12-14
XWIKI-21670: The versionPreserved instance output filter stream prope…

raphj:
I forgot to fix some descriptions when doing the refactoring for XWIKI-21670. They will be incorrect for documents if we don't fix this.


xwiki-platform-core/xwiki-platform-filter/xwiki-platform-filter-instance/xwiki-platform-filter-instance-document/src/main/java/org/xwiki/filter/instance/output/EntityInstanceOutputProperties.java
```diff
@@ -59,7 +59,7 @@ public class EntityInstanceOutputProperties extends InstanceOutputProperties
      * @return the comment to use when saving users
```
michitux, line 59: This comment isn't consistent with the property's description.

tmortagne: Done.

PR #2725, opened 2023-12-18
XWIKI-21702: Too big button for closing the drawer in a minimal distribution

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21702
## PR Changes
* Refined the selector for style of the avatar in order to avoid unexpected collision with the new icon in the close drawer button (when it's an image, using Silk)
## View
Button with the silk icon theme, before the PR
21702-beforePR
Button with the silk icon theme, after the PR
21702-afterPR
We can see that the button is much smaller after the PR, it's indeed of a similar size to the button when using font awesome:
21702-withFA

## Test
Since this is just a style change on a very specific element, I did not proceed to further testing than the checks available in the screenshots above.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
@@ -96,7 +96,7 @@
       }
 
       // Avatar styling
-      img {
+      img.avatar {
         border-radius: 50%;
         height: 50px;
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less
```diff
@@ -96,7 +96,7 @@
       }
 
       // Avatar styling
-      img {
+      img.avatar {
```
manuelleduc, line 99: The `avatar` class is only applied for the default avatar, but not when an actual avatar image is uploaded (https://jira.xwiki.org/browse/XWIKI-21743)

PR #2726, opened 2023-12-18
XWIKI-17243: Migrate WatchListClass objects to the new Notifications filter storage

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-17243

  * Add a new migration to properly save the old WatchListClass objects as filter preferences using the NotificationFilterPreferenceStore
  * Reuse the logic provided in the old bridge for performing the migration
  * Get rid of the the watchlistbridge module as it was only containing the component for performing the bridge
  * Remove the fallback on the WatchListClass for the automatic watch mode and create a new automatic watch mode xobject in the migration if that one was missing 
  * Integration tests are passing as well as manual test to migrate a 8.3 wiki with watches to 16.0.0


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R160000000XWIKI17243DataMigration.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        try {
+            this.notificationFilterPreferenceStore.saveFilterPreferences(userDocReference, results);
+        } catch (NotificationException e) {
+            throw new DataMigrationException(String.format("Error while trying to save [%s] filter preferences",
+                results.size()), e);
+        }
+        if (document.removeXObjects(classReference)) {
+            context.getWiki().saveDocument(document, "Migration of watchlist preferences", context);
```
tmortagne, line 145: Modifying and saving document during a migration have a tendency to cause problem with the listener which listen to those events because XWiki is not fully initialized. Generally, the migration in such a case is only used to register things to do after the XWiki init is finished.

surli: Ok I guess in this case I can use a dedicated TaskConsumer for cleaning up the docs afterwards.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R160000000XWIKI17243DataMigration.java
```diff
@@ -0,0 +1,198 @@
...
+    protected void hibernateMigrate() throws DataMigrationException, XWikiException
+    {
+        String statement = "from doc.object(XWiki.WatchListClass) as user";
+        int offset = 0;
+        List<String> results = null;
+        do {
+            try {
+                results = this.queryManager.createQuery(statement, Query.XWQL)
+                    .addFilter(uniqueFilter)
+                    .setOffset(offset)
```
surli, line 106: Outch I forgot to increment the offset in the loop...

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/R160000000XWIKI17243DataMigration.java
```diff
@@ -0,0 +1,198 @@
...
+import com.xpn.xwiki.objects.ListProperty;
+import com.xpn.xwiki.objects.PropertyInterface;
+import com.xpn.xwiki.objects.classes.ListClass;
+import com.xpn.xwiki.store.migration.DataMigrationException;
+import com.xpn.xwiki.store.migration.XWikiDBVersion;
+import com.xpn.xwiki.store.migration.hibernate.AbstractHibernateDataMigration;
+
+import static org.apache.commons.codec.digest.DigestUtils.sha256Hex;
+
+@Component
```
manuelleduc, line 48: Missing javadoc?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-watch/src/main/java/org/xwiki/notifications/filters/watch/internal/DefaultWatchedEntitiesConfiguration.java
```diff
@@ -94,12 +91,6 @@ public AutomaticWatchMode getAutomaticWatchMode(DocumentReference user)
             return AutomaticWatchMode.valueOf((String) value);
         }
 
-        // Fallback to the value of the Watchlist
```
surli, line 97: Note that this fallback is removed because we handle that in the `WatchListObjectsRemovalTaskConsumer` by actually setting a new automatic watch xobject if it's missing before removing the WatchListClass xobject. Now since it's done in a TaskConsumer there might be a few delay before it's getting into account leading to the miss of a few filters creation. IMO it's acceptable as it's not something that should happen a lot.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/WatchListObjectsRemovalTaskConsumer.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @version $Id$
+ * @since 16.0.0RC1
+ */
+@Component
+@Named(WatchListObjectsRemovalTaskConsumer.TASK_NAME)
+@Singleton
+public class WatchListObjectsRemovalTaskConsumer implements TaskConsumer
+{
+    static final String TASK_NAME = "watchlist-xobject-removal";
```
manuelleduc, line 57: Maybe prefix this with the name of the module `notifications-watchlist-xobject-removal`?

surli: I sure can, but to my knowledge we don't really have a best practice for those hints / names?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/WatchListObjectsRemovalTaskConsumer.java
```diff
@@ -0,0 +1,100 @@
...
+            BaseObject watchXObject = document.getXObject(classReference);
+            if (watchXObject != null && document.getXObject(autoWatchClassReference) == null) {
+                String automaticwatch = watchXObject.getStringValue("automaticwatch");
+                if (!StringUtils.isBlank(automaticwatch) && !"default".equals(automaticwatch)) {
+                    BaseObject autowatchXObject = document.newXObject(autoWatchClassReference, context);
+                    autowatchXObject.setStringValue("automaticWatchMode", automaticwatch);
+                }
+            }
+
+            if (document.removeXObjects(classReference)) {
```
manuelleduc, line 93: Is it expected that the new XObject created line 88 won't be saved if `removeXObjects` returns false?

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/WatchListObjectsRemovalTaskConsumer.java
manuelleduc: yes!

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-default/src/main/java/org/xwiki/notifications/filters/migration/WatchListObjectsRemovalTaskConsumer.java
surli: `removeXObjects` returns false only if there's no object with the given reference, so the if `watchXObject != null` would have been true. So basically you cannot enter in previous if without entering in that if too

manuelleduc: The if might be superfluous in this case?

surli: Which one? I need the first if, since it's really an edge case. This if is less likely to return false, since in theory I enter the TaskConsumer if I have an XWikiListClass xobject to remove, now since the method does return a boolean better safe and use the result for saving IMO.

PR #2728, opened 2023-12-19
XWIKI-20435: Pencil icon in user profile can be displayed white over white and thus not be visible

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-20435
## PR Changes
* Replaced the custom color mappings with a bootstrap standard class for the action button
## Notes
* It's still possible to display the pencil icon in white over white, but at least now it breaks contrast with the same conditions as any other `btn-default` (e.g. action buttons for page editing and creation).
* I also took the opportunity to remove a duplicated non color related style: ` div.editProfileCategory { float: right;}` and fix the codestyle in the CSS block where those changes are done


Sereza7:
Sorry! I did prepare those screenshots but forgot to link them here:

before the PR vvv
20435-beforePR
after the PR vvv
20435-afterPR

PR #2729, opened 2023-12-19
XWIKI-21370: Header structure uses too many H1

Sereza7:
## Jira
https://jira.xwiki.org/browse/XWIKI-21370
## PR Changes
* Added a sr-only title for wiki and wysiwyg pages
* Removed two empty pages from the distribution flavor tests...
## Note
Looking at the failure reports on CI https://ci.xwiki.org/job/XWiki/job/xwiki-platform/job/stable-15.10.x/34/testReport/junit/org.xwiki.test.webstandards/CustomDutchWebGuidelinesValidationTest/Platform_Builds___main___distribution___flavor_test_webstandards___Build_for_Flavor_Test___Webstandards_______22/ and https://ci.xwiki.org/job/XWiki/job/xwiki-platform/job/stable-15.10.x/34/testReport/junit/org.xwiki.test.webstandards/CustomDutchWebGuidelinesValidationTest/Platform_Builds___main___distribution___flavor_test_webstandards___Build_for_Flavor_Test___Webstandards_______53/ ), it seems like there is no content on those pages (search for `mainContentArea`). Therefore I don't see a problem in removing them from those tests.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
@@ -1257,6 +1257,7 @@ platform.web.editors.class.pageTitle=Editing class {0}
 platform.web.editors.rights.pageTitle=Editing access rights for {0}
 platform.web.editors.unknown.pageTitle=Editing {0}
 
+core.editors.content.title=Editing title and content of <a href="{1}">{0}</a>
 core.editors.content.parentField.label=Parent
 core.editors.content.parentField.edit=(edit)
 core.editors.content.parentField.edit.title=Edit parent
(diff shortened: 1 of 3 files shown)


Sereza7:
If this PR is accepted, the following commit should be backported to the `15.10.x` branch similarly to both of the previous commits for XWIKI-21370 :)

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1257,6 +1257,7 @@ platform.web.editors.class.pageTitle=Editing class {0}
 platform.web.editors.rights.pageTitle=Editing access rights for {0}
 platform.web.editors.unknown.pageTitle=Editing {0}
 
+core.editors.content.title=Editing title and content of <a href="{1}">{0}</a>
```
surli, line 1260: Why providing a link there? I mean you're already on the page when you edit it...

Sereza7: It's the same pattern as the one used for other edit modes:
* `core.editors.object.title=Editing objects of <a href="{1}">{0}</a>`
* `core.editors.class.title=Editing class <a href="{1}">{0}</a>`
* `core.editors.rights.title=Editing rights of <a href="{1}">{0}</a>`

The links provided redirect to the view mode. I don't think it's really necessary but we shouyld be consistent and change the others too if we don't want a link here.

surli: Yep agreed for consistency

PR #2748, opened 2023-12-28
XWIKI-21743: Non-default avatars are not rounded nor resized

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-21743

Warning, the currently proposed solution has some broad impact than just the drawer.
The image inserted through `largeUserAvatar`/`mediumUserAvatar`/`smallUserAvatar` or any other code relying internally on `resizedUserAvatar` will now have classes `avatar` and `avatar_${size}`.
This has for instance an impact on the user directory where non-default images where not rounded previously, and will now be rounded.

## Before

image

## After

image

It's in my opinion an interesting fix.

PS: I've used this screen as an example but globally, the fix is making the use of rounded corners for avatar more uniform across the wiki.

But, if somebody feels like this is too impactful, another option is to add an optional parameter `additionalClasses` to `largeUserAvatar`, `mediumUserAvatar`, `smallUserAvatar`, and `resizedUserAvatar` , allow to define additional classes, and it use it in `drawer.vm` to see the style from [drawer.less](https://github.com/xwiki/xwiki-platform/blob/5cdd0ad615640017e381a1d4d48dcb5024e536cc/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/drawer.less#L99) applied again on the user's avatar.


Sereza7:
Looks good to me :+1: 

Thanks for working on this :)

surli (approved):

PR #2752, opened 2023-12-30
XWIKI-21712: In create screen, set focus to the title input field by default

wthrajat:
Jira: https://jira.xwiki.org/browse/XWIKI-21712

**Live changes: [Video demo](https://up1.xwikisas.com/#0KmSUenKWkmlbbpgKfEmZg)**

This PR implements the use of `autofocus` attribute for simplicity purposes (instead of going with the JS solutions), see more at: [autofocus - MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)

This has the following accessibility compromises though (as also described in the MDN article):

1. When autofocus is assigned, screen-readers "teleport" their user to the form control without warning them beforehand.

2. Automatically focusing on a control can cause the page to scroll on load.
**(Not relevant since the `TITLE` field is at the top).**

3. The screen reader **might** not announce anything before the label.

(CC: @slauriere)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/locationPicker_macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/locationPicker_macros.vm
@@ -47,7 +47,7 @@
       </dt>
       <dd>
         <input type="text" id="$escapetool.xml($!{options.id})Title" name="$escapetool.xml($titleField.name)" value="$!escapedValue"
-          class="location-title-field" placeholder="$!escapetool.xml($services.localization.render($titleField.placeholder))" />
+          class="location-title-field" placeholder="$!escapetool.xml($services.localization.render($titleField.placeholder))" autofocus />
       </dd>
     #elseif ($titleField)
       <dt class="hidden"></dt>


Sereza7:
@wthrajat Thanks for working on this! 
Nice demo of the changes :)
I agree with using `autofocus`, it's better (at least regarding a11y) than any workaround we could make with Javascript :+1: 

Regarding the accessibility compromises, I think they are worth it, in fact I think we actually make it more accessible :)

1. We assign autofocus at page load, so the screen-reader is only teleported at page load. I think this sentence on MDN meant to discourage adding autofocus on the go, which could create focus traps and disorientate users. Here, the user click on `Create`, a new page loads and puts them on the start of the main content of the page, so in my opinion they shouldn't be too lost.
3. It's actually a nice feature for this use case. I think the label provides enough context. The non-sighted user will directly get access to the form field `Title of the new page` after clicking the `Create` button. In my opnion it's not necessary to repeat anything even for first time non-sighted users.


PS: We shouldn't put autofocus on every "form page" because some might need more context. Creating a page is a pretty common and simple action, that most users would expect to start by giving the page name. It's just like logging in (which already has autofocus), a simple form for which most users would expect to start by giving their username.

slauriere:
That's really cool! Thumbs up @wthrajat  :+1:  + very interesting comments @Sereza7!  Looking forward to using this improvement soon :-)  / cheers ++ (and thanks for adding me in cc as i'm not receiving notifications from Jira at the moment)

wthrajat:
> i'm not receiving notifications from Jira at the moment

@slauriere, haha I remember this from the GSoC times, hence tagged you :-)

PR #2770, opened 2024-01-04
XWIKI-13967: Confusion between Supported Languages and Default Languages when multilingual is set to no

Sereza7:
# Jira
https://jira.xwiki.org/browse/XWIKI-13967
# PR Changes
* Added a CSS ruleset to remove the supported languages field from the UI when it's not useful.
# Note
This makes use of the new CSS feature `:has`, which is only available in the latest stable release of Firefox. For now this release is not used by all users but in less than 30 days we can count on 70% of users moving on this version. All of the other supported browser for XWiki have supported `:has` for a while. It's a correct change in regards to our browser support policy.
\<EDIT> On older versions of Firefox, the rule is ignored and the supported language field is always shown. \</EDIT>

# View
[Here is a demo of the changes brought in this PR](https://up1.xwikisas.com/#8W-FfnSRNQebgqXtIEmfeQ). We can see that the `support languages` field disappears when the wiki is not set as multilingual.

## Tests
Successfully passed `mvn clean install -f xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-test/xwiki-platform-administration-test-docker -Dxwiki.enforcer.skip=true -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false`.


surli (approved):

manuelleduc (approved):

manuelleduc:
Out of curiosity, why do you need those options in the maven command: `-Dxwiki.enforcer.skip=true -Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false` ?

Sereza7:
@manuelleduc 
* `-Dxwiki.enforcer.skip=true` at this time, I didn't upgrade my maven yet, so the enforcer failed for any build and I disabled it
* `-Dgradle.cache.local.enabled=false -Dgradle.cache.remote.enabled=false` to make sure I run everything and don't rely on the cache. I don't understand those options very well but IMO they just disable some build cache, which can not be a bad thing for tests. It's slower but at least I'm sure to test the actual changes brought in the PR.

PR #2771, opened 2024-01-04
XWIKI-14054: Don't have the option to Edit users avatar from view mode as Admin (only my own)

Sereza7:
# Jira
https://jira.xwiki.org/browse/XWIKI-14054
# PR Changes
* Added an alternative condition to allow direct edit of the profile picture to improve consistency of the UI for admins and people with right on profile pages
# View
14054-afterPR
We can see on this screenshot that the PR allows the Administrator to edit directly the avatar of another user from the view mode, just like what is possible when viewing your own profile.


manuelleduc (approved):

Sereza7:
Note that this boolean is used throughout the multiple tabs. 
I'll be checking all of them and propose a fix in here or open a proposal to change this if this was discussed before (aka if it's an intended feature and not just an arbitrary decision).

Sereza7:
So it seems it was not a discussed change (or that we lost its history): https://github.com/xwiki/xwiki-platform/commit/947ea7473cb698fb355ddd12277c935698b1bcb3 .
IMO removing this makes sense in regards to the current state of rights in XWiki, done in https://github.com/xwiki/xwiki-platform/pull/2771/commits/170635b5aee4653e50314f5baa3087f8443c9c27 . 

Seems like [all the other uses of this condition](https://github.com/search?q=repo%3Axwiki%2Fxwiki-platform%20%24isMyProfile&type=code) are legitimate. Here, it was not legitimate because we used it in place of `$hasEdit`.

PR #2774, opened 2024-01-05
XWIKI-14873: Accordeon in the Administration misses an indicator to show they are expandable

Sereza7:
# Jira
https://jira.xwiki.org/browse/XWIKI-14873
# PR Changes
* Added carets to the accordeon headers
# Note
I decided to make the caret orientation  consistent with the one already used for dropdowns in the horizontal menus.
# View
[Here is a demo of the UI with the carets.](https://up1.xwikisas.com/#m9rBSekAaak8frDkgVrHzg)

Before the PR
14873-beforePR
After the PR
14873-afterPR

### DIFF
--- a/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminSheet.xml
@@ -304,9 +304,12 @@
     #sortCollectionOfMapsByField($item.children, 'order', 99999, 'asc', $children)
     &lt;div class="panel panel-default"&gt;
       &lt;a class="panel-heading#if (!$isActive) collapsed#end" id="panel-heading-$escapedId"
-        href="$!item.url" data-toggle="collapse"#if ("$!options.id" != '') data-parent="#$options.id" #end
-        data-target="#panel-body-$escapedId" aria-expanded="$isActive" aria-controls="panel-body-$escapedId"
-        title="$!escapetool.xml($item.description)"&gt;$!services.icon.renderHTML($item.icon)$escapetool.xml($name)&lt;/a&gt;
+      href="$!item.url" data-toggle="collapse"#if ("$!options.id" != '') data-parent="#$options.id" #end
+      data-target="#panel-body-$escapedId" aria-expanded="$isActive" aria-controls="panel-body-$escapedId"
+      title="$!escapetool.xml($item.description)"&gt;
+        &lt;span&gt;$!services.icon.renderHTML($item.icon)$escapetool.xml($name)&lt;/span&gt;
+        &lt;div&gt;$services.icon.renderHTML('caret-down')&lt;/div&gt;
+      &lt;/a&gt;
       &lt;section class="panel-collapse collapse#if ($isActive) in#end" id="panel-body-$escapedId"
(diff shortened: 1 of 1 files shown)


manuelleduc:
I would have expected the caret to be left aligned. Did you try that?

Sereza7:
You mean something like that?
image

Ruleset changes: 
* (+) .admin-menu a.panel-heading {  gap: 0.2em;}
* (-) .admin-menu a.panel-heading span {  flex-grow: 1;}

___
IMO it looks cleaner in the version I proposed first. Moreover, so far carets we use for dropdowns are not left aligned:
* Search Facets 
image
* Notification preferences
image
* Menu is a bit of an in-between, it's left aligned but the size of the container changes based on content so it's also on the right of the container. However sub-menu are left-aligned. 
image

From what I remember thost are the main spots where we use carets for dropdowns. IMO we should have consistency between all of those since they hold the same function. Should we left align them?

Thanks for your feedback!

manuelleduc:
> You mean something like that?

I meant at  the left of the elements (just before the icons).
But if we already have other UI elements with the same layout, I guess it's fine.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminSheet.xml
```diff
@@ -967,20 +970,35 @@ require(['jquery'], function($) {
...
+}
+
 .admin-menu a.panel-heading:not(.collapsed) {
   background-color: $theme.menuBackgroundColor;
   color: $theme.menuLinkColor;
 }
+
+/* Rotate the carret on collapse. */
+.admin-menu a.panel-heading.collapsed &gt; *:last-child {
+  transform: rotate(90deg);
```
surli, line 988: Interesting, I thought you were using two different icons, but it's same with rotation. Do you know if that's what we're using in other places?

surli (approved):

surli:
>  IMO we should have consistency between all of those since they hold the same function.

+1 and so you should keep them aligned to the right as in first screenshot. Also we probably need a ticket for the ones on notification settings as the icons are always pointing down.

xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminSheet.xml
Sereza7: We don't have a caret-left icon in the IconTheme, so it's pretty common to do so.

There's some places where we use two icons, but that's with `caret-right` and `caret-down` together.

I checked a bit further than before and found:
* This `more option` dropdown at the bottom of the admin export tab, IMO we should update this :)
image
* This `Local history` dropdown in extension history
image
* In the `displayJobStatus` velocity macro, on a dropdown for the logs:
image

... and in a few other places.
Most of those places have their own classes.
I'll spend some time comparing them and proposing a design pattern to make it easier to maintain on the long run :)

Sereza7:
I created a ticket for the one in Notification settings: https://jira.xwiki.org/browse/XWIKI-21766

After [looking at the other dropdowns](https://github.com/xwiki/xwiki-platform/pull/2774#discussion_r1444573768), it seems to me like the best solution would be to provide a template so that we don't have 20 different implementations of the same component.

PR #2778, opened 2024-01-09
XWIKI-21757: Support for Live Data description

manuelleduc:
# Jira URL
https://jira.xwiki.org/browse/XWIKI-21757

# Changes
## Description

For discussion: https://forum.xwiki.org/t/live-data-description/13787
Provide a way to add a textual description to the Live Data content.

# Screenshots & Video

The example below use the migration from Livetable to Live Data of the page index.
Previously, the description was in the table caption, with a `sr-only` class, making the description only visible to screen readers.

## Before

image

## After

image

# Executed Tests
* xwiki-platform-livedata-macro
  * LiveDataMacroTest.java updated
* xwiki-platform-livedata-webjar
  * LivedataLayout.spec.js introduced
# Expected merging strategy
* Prefers squash: Yes
* Backport on branches: none, this is an improvement

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataMeta.java
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataMeta.java
@@ -63,6 +63,8 @@ public class LiveDataMeta
 
     private LiveDataSelectionConfiguration selection;
 
+    private String description;
+
     /**
      * @return the default layout used to display the live data
      */
@@ -285,6 +287,24 @@ public class LiveDataMeta
     public void setSelection(LiveDataSelectionConfiguration selection)
     {
         this.selection = selection;
+    }
+
+    /**
+     * @return an optional textual description of the Live Data
+     * @since 16.0.0RC1
+     */
+    public String getDescription()
+    {
+        return this.description;
+    }
+
+    /**
+     * @param description an optional textual description of the Live Data
+     * @since 16.0.0RC1
+     */
+    public void setDescription(String description)
+    {
+        this.description = description;
     }
 
     /**
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataMeta.java
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataMeta.java
@@ -25,6 +25,7 @@ import java.util.Objects;
 
 import org.xwiki.livedata.LiveDataPropertyDescriptor.DisplayerDescriptor;
 import org.xwiki.livedata.LiveDataPropertyDescriptor.FilterDescriptor;
+import org.xwiki.stability.Unstable;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -293,6 +294,7 @@ public class LiveDataMeta
      * @return an optional textual description of the Live Data
      * @since 16.0.0RC1
      */
+    @Unstable
     public String getDescription()
     {
         return this.description;
@@ -302,6 +304,7 @@ public class LiveDataMeta
      * @param description an optional textual description of the Live Data
      * @since 16.0.0RC1
      */
+    @Unstable
     public void setDescription(String description)
     {
         this.description = description;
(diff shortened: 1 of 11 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataMeta.java
```diff
@@ -287,6 +289,24 @@ public void setSelection(LiveDataSelectionConfiguration selection)
         this.selection = selection;
     }
 
+    /**
+     * @return an optional textual description of the Live Data
+     * @since 16.0.0RC1
+     */
+    public String getDescription()
```
manuelleduc, line 296: Forgot the unstable annotation, adding it...

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/LivedataLayout.vue
```diff
@@ -34,6 +34,9 @@
 -->
 <template>
   <div class="livedata-layout">
+    <span class="description" :id="descriptionId" v-if="hasDescription">
```
michitux, line 37: It looks odd to me to place a `span` inside a `div`, this is not something we usually do in XWiki. In particular, I think this element should be a block element so a `p` would be much more appropriate. Also, in the proposal you mentioned using a `caption` element for the table layout, why did you decide to deviate from your proposal?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/java/org/xwiki/livedata/macro/LiveDataMacroParameters.java
```diff
@@ -133,4 +134,15 @@ public void setPageSizes(String pageSizes)
     {
         super.setPageSizes(pageSizes);
     }
+
+    /**
+     * {@inheritDoc}
```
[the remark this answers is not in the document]
manuelleduc: Good point, the `Override` is indeed missing.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/LivedataLayout.vue
manuelleduc: This is making the location of the description more stable.
I realized that if the location of the description was up to the layout, switching layout could mean having the description moving quite radically to a new location, which can be confusing.

manuelleduc: Also, this is less work for implementors of new layout, as they don't have to remember to take into account the description.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/LivedataLayout.vue
```diff
@@ -34,6 +34,9 @@
 -->
 <template>
   <div class="livedata-layout">
+    <div class="description" :id="descriptionId" v-if="hasDescription">
```
michitux, line 37: Note that `div` has no spacing etc. by default while `p` has which is why I suggested using a `p` - in the screenshot you show, for example, I miss spacing above the caption. I don't know what styling you get by `&:extend(caption);`, though.

Also, I'm wondering if using such generic classes like `description` is a good idea as this way, styles could easily clash with similar elements on other parts of the UI. But I don't know if we have any guidelines on that.

manuelleduc: Hum indeed, while having short names is fine in combination with vue scope style for local definitions, this is not ok for external css that might have unexpected impacts on the descriptions.

Sereza7: > Also, I'm wondering if using such generic classes like description is a good idea as this way, styles could easily clash with similar elements on other parts of the UI. But I don't know if we have any guidelines on that.

From what I can see, this is barely used yet (see https://github.com/search?q=org%3Axwiki++.description+language%3ACSS+OR+language%3ALESS+&type=code) so there shouldn't be any collision. Fields use the `hint` vocabulary instead with classes like `.xHint`.

However I agree that adding more specificity to [the style](https://github.com/xwiki/xwiki-platform/pull/2778/files#diff-85e5e59addf6ae705f1cafde1728bb052d6f4ca0e6e022e10ab3998e162d7c2fR170-R174) will avoid unexpected issues if we ever use this class again later on :+1:

manuelleduc: What about `livedata-layout-description` or `xwiki-livedata-layout-description`?

manuelleduc: Using `livedata-layout-description` for now as it is consistent with the `livedata-layout` class of the parent element.

PR #2782, opened 2024-01-10
XWIKI-21783: Add a check to ensure we don't have two migrations with same DB version

surli:
JIRA: https://jira.xwiki.org/browse/XWIKI-21783

  * Fail the data migration manager initialization if two migrations are found with same version.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/AbstractDataMigrationManager.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/AbstractDataMigrationManager.java
@@ -49,6 +49,7 @@ import com.xpn.xwiki.XWikiConfig;
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.internal.store.hibernate.HibernateConfiguration;
+import com.xpn.xwiki.store.migration.hibernate.HibernateDataMigration;
 
 /**
  * Template for {@link DataMigrationManager}.
@@ -338,9 +339,9 @@ public abstract class AbstractDataMigrationManager implements DataMigrationManag
     @Override
     public void initialize() throws InitializationException
     {
+        checkMigrationsVersions();
         try {
             SortedMap<XWikiDBVersion, XWikiMigration> availableMigrations = new TreeMap<>();
-
             Map<XWikiDBVersion, XWikiMigration> forcedMigrations = getForcedMigrations();
             if (!forcedMigrations.isEmpty()) {
                 availableMigrations.putAll(forcedMigrations);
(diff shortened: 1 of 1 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/AbstractDataMigrationManager.java
```diff
@@ -347,12 +347,18 @@ public void initialize() throws InitializationException
...
+                        || ignoredMigrations.contains(migratorVersion.toString())) {
                         continue;
                     }
                     XWikiMigration migration = new XWikiMigration(migrator, false);
-                    availableMigrations.put(migrator.getVersion(), migration);
+                    if (!availableMigrations.containsKey(migratorVersion)) {
+                        availableMigrations.put(migratorVersion, migration);
+                    } else {
+                        throw new InitializationException(
+                            String.format("A migration with version [%s] already exists.", migratorVersion));
```
tmortagne, line 360: Giving a bit more information, like the classes of the two colliding migrations, would be nice.

surli: Ok will improve the message. Besides that this check is enough you think? Was wondering if checking versions of forced migration was needed too?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/AbstractDataMigrationManager.java
```diff
@@ -338,26 +339,26 @@ protected String getMainXWiki()
...
                     }
                     XWikiMigration migration = new XWikiMigration(migrator, false);
-                    availableMigrations.put(migrator.getVersion(), migration);
+                    availableMigrations.put(migratorVersion, migration);
                 }
             }
 
             this.targetVersion =
-                (availableMigrations.size() > 0) ? availableMigrations.lastKey() : new XWikiDBVersion(0);
+                (!availableMigrations.isEmpty()) ? availableMigrations.lastKey() : new XWikiDBVersion(0);
```
manuelleduc, line 361:
```suggestion
                !availableMigrations.isEmpty() ? availableMigrations.lastKey() : new XWikiDBVersion(0);
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/AbstractDataMigrationManager.java
surli: So I actually changed a bit the logic to perform a check of all migrations directly at the beginning, by getting list of components. It involves one more loop over the list of components that I didn't have before, but it allows to check globally the migrations.

surli: Also I'm just realizing that previous implem was wrong: in case of forced migration you'd have had an exception thrown actually...

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/migration/AbstractDataMigrationManager.java
```diff
@@ -366,6 +367,49 @@ public void initialize() throws InitializationException
...
+    {
+        try {
+            List<HibernateDataMigration> migrationList =
+                this.componentManager.getInstanceList(HibernateDataMigration.class);
+            Set<String> ignoredMigrations = new HashSet<>(this.hibernateConfiguration.getIgnoredMigrations());
+            Map<XWikiDBVersion, String> hintMap = new HashMap<>();
+            for (HibernateDataMigration dataMigration : migrationList) {
+                XWikiDBVersion version = dataMigration.getVersion();
+                if (hintMap.containsKey(version)) {
+                    if (isMigrationIgnored(dataMigration, ignoredMigrations)) {
```
manuelleduc, line 391: What if the ignored migration is processed second?
iiuc the first ignored migration is going to be added to `hintMap`, and the second migration will trigger a `InitializationException` instead of a warn log.
Is my understanding correct?

surli: Yes and for me it's ok: the only reason why I make exception here for ignored migration is to not block users: i.e. they can use the ignored migration feature to skip this check if they need to. Ignored migration can use either the version of the migration, or the name. In theory this should never be used, but I imagine weird UC where there's a migration coming from an extension which have same DB version: we probably don't want to block the initialization of the wiki.

manuelleduc: ok, make sense. Thanks
Might be worth having a comment, to avoid confusion when reading this code later.

PR #2788, opened 2024-01-11
XWIKI-19150: The Page Like UI is not following the common template style for a page

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-19150
# Changes
## Description
* Added a proper content header to the liker page.
* Added an anchor in the title
* Updated the id of the livetable to avoid a collision with an item in the `#tmMoreActions` menu

# Screenshots & Video

Before PR vvv
beforePR
After PR vvv
19150-afterPR

# Executed Tests

Succcessfully built with `mvn clean install -f xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-api`
# Expected merging strategy

* Prefers squash: Yes
* Backport to 15.10.X : localized changes on the content of a template, low risk of breaking something.


xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-api/src/main/resources/templates/likers.vm
```diff
@@ -118,5 +119,5 @@
         'translationPrefix' : "xe.admin.users."
     })
 
-    #livetable("likers" $columns $columnsProperties $options)
+    #livetable("likersTable" $columns $columnsProperties $options)
```
manuelleduc, line 122: Why do you need to change the live table id?

manuelleduc:
Also, note that I'm working on https://jira.xwiki.org/browse/XWIKI-18992 which is also going to impact `likers.vm`.

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-api/src/main/resources/templates/likers.vm
Sereza7: Because it collides with the item in the #tmMoreAction menu.

Sereza7: Without this change the livetable wouldn't work properly anymore when adding the contentheader

manuelleduc: ok, maybe add a comment explaining that the is a risk of "id collision".

Sereza7: I wasn't sure such a comment would really be helpful on the long term / outside of the context of this PR (not really any reason to change this livetable ID again). 
I added a comment in https://github.com/xwiki/xwiki-platform/pull/2788/commits/95a0d0fb0037158ec197d0b1464f66a76d919595 . :+1:

xwiki-platform-core/xwiki-platform-like/xwiki-platform-like-api/src/main/resources/templates/likers.vm
```diff
@@ -99,7 +99,8 @@
     #livetable_filterObfuscated($map)
     $jsontool.serialize($map)
 #else
-    <h1>$escapetool.xml($services.localization.render('like.likers.title', [$doc.plainTitle]))</h1>
+    #set($titleToDisplay = $services.localization.render('like.likers.title', ["<a href='$doc.getURL()'>$escapetool.xml($doc.plainTitle)</a>"]))
```
manuelleduc, line 102: To be tested but escaping html properties is probably safer.
```suggestion
    #set($titleToDisplay = $services.localization.render('like.likers.title', ["<a href='$escapetool.xml$doc.getURL())'>$escapetool.xml($doc.plainTitle)</a>"]))
```

Sereza7: I'm pretty sure we shouldn't escape this URL: https://github.com/search?q=repo%3Axwiki%2Fxwiki-platform+%2Fhref%3D%27.*getURL%2F&type=code

Most of those are not escaped ...

surli:
@Sereza7 you have conflicts to fix on this one.

PR #2790, opened 2024-01-15
XWIKI-21553: Do not override existing documents on user action

michitux:
# Jira URL

* https://jira.xwiki.org/browse/XWIKI-21553

# Changes

## Description

* Mark a document as not new if the document already exists, but a user action saves a new document.
* Add a unit test.

## Clarifications

It was never meant to happen that a user can save a new document if the document already exists, but this is possible due to my fixes for [XWIKI-19549](https://jira.xwiki.org/browse/XWIKI-19549) in the case that a user has edit but not view right. In this case, a new document can be in context, e.g., in the save action. I've started a discussion to change the right resolution to make this possible on the [forum](https://forum.xwiki.org/t/should-denied-view-right-deny-other-rights-or-be-overridden-by-other-rights/13557), but there has been very little feedback and such a change seems dangerous on a LTS version. This change is therefore meant either as bulletproofing or also as an alternative fix that should hopefully be safer.

# Executed Tests

I've manually tested this with the scenario described in the Jira issue and run the tests in `xwiki-platform-oldcore` and `xwiki-platform-legacy-oldcore`. Test coverage for this scenario seems highly unlikely.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 14.10.x
  * 15.5.x
  * 15.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
@@ -33,10 +33,17 @@ import com.xpn.xwiki.internal.event.UserUpdatingDocumentEvent;
 /**
  * Cancel any save that tries to save a new document if the document already exists.
  *
+ * <p>
+ *     This can happen when the context document is set to an empty document because the user doesn't have view right,
+ *     but the user still has, e.g., edit right. In this situation, the user would still be able to to save but it is
+ *     hard to imagine a scenario where this would make sense. This listener therefore cancels the save in this
+ *     situation.
+ * </p>
+ *
  * @version $Id$
  * @since 14.10.21
  * @since 15.5.5
- * @since 15.10.5
+ * @since 15.10.6
  */
 @Component
 @Singleton
@@ -60,12 +67,9 @@ public class DocumentOverrideListener extends AbstractEventListener
     public void onEvent(Event event, Object source, Object data)
     {
         XWikiDocument document = (XWikiDocument) source;
-
         XWikiDocument originalDocument = document.getOriginalDocument();
 
-        if (event instanceof CancelableEvent && document.isNew() && originalDocument != null
-            && !originalDocument.isNew())
-        {
+        if (document.isNew() && originalDocument != null && !originalDocument.isNew()) {
             ((CancelableEvent) event).cancel(
                 "The document already exists but the document to be saved is marked as new.");
         }
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
@@ -1,0 +1,73 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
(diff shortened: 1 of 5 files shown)


michitux:
As discussed with @tmortagne, the first version had problems like not preserving the creator. We arrived at the conclusion that it might actually be best to just cancel such a save action as this should never intentionally happen, and an admin would want to undo this action in basically all cases, anyway. I've therefore implemented a new event listener that implements this behavior. The advantage of this approach is also that if there should be any issue with this approach, the event listener can relatively easily be either disabled or overridden by an extension without having to patch XWiki itself.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
```diff
@@ -0,0 +1,73 @@
...
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.internal.event.UserUpdatingDocumentEvent;
+
+/**
+ * Cancel any save that tries to save a new document if the document already exists.
+ *
+ * @version $Id$
+ * @since 14.10.21
+ * @since 15.5.5
+ * @since 15.10.5
```
surli, line 39: Probably need to be moved to 15.10.6

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+@Component
+@Singleton
+@Named(DocumentOverrideListener.NAME)
+public class DocumentOverrideListener extends AbstractEventListener
```
surli, line 44: Not sure but I feel like it should be a local event listener only? What's happening in case of remote event?

michitux: I thought this kind of event cannot be remote?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        XWikiDocument document = (XWikiDocument) source;
+
+        XWikiDocument originalDocument = document.getOriginalDocument();
+
+        if (event instanceof CancelableEvent && document.isNew() && originalDocument != null
```
tmortagne, line 66: UserUpdatingDocumentEvent is cancellable by definition, and you are not listening to any other event so in theory you don't need the `event instanceof CancelableEvent`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
tmortagne: Yes, I don't see how you could end up with this event sent to another cluster member.

surli: Ok then good.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/org/xwiki/internal/document/DocumentOverrideListener.java
michitux: Okay, I've removed it.

tmortagne (approved):

PR #2803, opened 2024-01-17
XWIKI-21807: Error when adding two properties of the same name not properly displayed in class editor

manuelleduc:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21807

# Changes

* Don't rely on the status message which is not supported since jetty 10.x (https://github.com/jetty/jetty.project/issues/3225)
* Instead, set the status message as the response body and stop the action there (return false)

## Description

* [x] Add a functional tests

# Executed Tests

```bash
mvn clean install -Pquality,integration-tests,docker,legacy -pl :xwiki-platform-oldcore,:xwiki-platform-xclass-test-docker,:xwiki-platform-legacy-oldcore -Dit.test=ClassSheetIT
```

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * stable-15.5.x
  * stable-15.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
@@ -19,6 +19,7 @@
  */
 package com.xpn.xwiki.web;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -43,6 +44,9 @@ import com.xpn.xwiki.util.Util;
 @Singleton
 public class PropAddAction extends XWikiAction
 {
+    private static final String PROPERTY_ERROR_ALREADY_EXISTS_ERROR_MESSAGE =
+        "action.addClassProperty.error.alreadyExists";
+
     @Override
     protected Class<? extends XWikiForm> getFormClass()
     {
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
```diff
@@ -80,13 +84,19 @@ public boolean action(XWikiContext context) throws XWikiException
...
             parameters.add(propName);
             context.put("messageParameters", parameters);
-            response.setStatus(HttpServletResponse.SC_BAD_REQUEST,
-                localizePlainOrKey("action.addClassProperty.error.alreadyExists", parameters.toArray()));
-            return true;
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            try {
+                String localizedMessage =
+                    localizePlainOrReturnKey(PROPERTY_ERROR_ALREADY_EXISTS_ERROR_MESSAGE, propName);
+                response.getWriter().write(localizedMessage);
```
michitux, line 95: This sounds dangerous, when writing arbitrary content to the output, always also set the mime type to `text/plain` to avoid XSS attacks (no idea if this would be possible here, but better stay on the safe side).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
```diff
@@ -80,13 +84,19 @@ public boolean action(XWikiContext context) throws XWikiException
         BaseClass bclass = doc.getXClass();
         bclass.setName(doc.getFullName());
         if (bclass.get(propName) != null) {
-            context.put("message", "action.addClassProperty.error.alreadyExists");
+            context.put("message", PROPERTY_ERROR_ALREADY_EXISTS_ERROR_MESSAGE);
             List<String> parameters = new ArrayList<String>();
             parameters.add(propName);
             context.put("messageParameters", parameters);
```
michitux, line 90: I don't understand why you're still adding the message and the parameters to the context. Has this any use when the action is aborted?

michitux (changes requested):
Thank you, overall this looks good if this fixes the issue, I just think that the code could be simplified and the content type should be set.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
manuelleduc: Good point, the context is not used afterward when false is returned, I can remove that code.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
```diff
@@ -80,13 +81,16 @@ public boolean action(XWikiContext context) throws XWikiException
         BaseClass bclass = doc.getXClass();
         bclass.setName(doc.getFullName());
         if (bclass.get(propName) != null) {
```
tmortagne, line 83: The content of this if seems to be mostly duplicating `writeAjaxErrorResponse`.

tmortagne:
Looks good, just a codestyle comment.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/PropAddAction.java
manuelleduc: Thanks, I couldn't find it. This is definitely doing the same thing.

PR #2812, opened 2024-01-22
XWIKI-21778: Admin section: make the Extension section pass webstandard tests

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21778

# Changes

## Description

* Replaced inline styling with a use of the meter element
* Added the webstandard test for the extension section.

## Clarifications

* The meter style options are not standard HTML yet, so multiple style need to be used for webkit based browsers and Firefox. Changes have been tested manually with Firefox 121 and Chrome 120
* Regarding accessibility, it seems like the doc about meter on the Aria Practice Guide is [not up to date](https://github.com/w3c/aria-practices/issues/2905)

# Screenshots & Video

No UI Change expected.
before vvv
21778-beforePR
after (Firefox) vvv
21778-afterPRfirefox
after (Chrome) vvv
21778-afterPRChrome

# Executed Tests

Successfully passed `mvn clean install -f xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards -Dxwiki.test.startXWiki=false` for the added page `/xwiki/bin/admin/XWiki/XWikiPreferences?editor=globaladmin&section=XWiki.Extensions`.
Manual tests for Firefox 121 and Chrome 120. Browser support is pretty approximative, it might be necessary to check it on Safari and Edge too, even though it should be okay (based on Webkit like Chrome).

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/rating_macros.vm
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/rating_macros.vm
@@ -36,7 +36,9 @@
     <div class="rating-header $cssclass">$ratinglabel</div>
     <div class="rating-stars">
         <ul class="star-rating ${cssclass}-star #if($locked) locked #end">
-            <li class="current-rating" style="width:${width}%;"></li>
+            <li class="current-rating">
+                <meter class="average-rating" min="0" max="5" value="$rating"/>
+            </li>
             #set($cls = ["one-star", "two-stars", "three-stars", "four-stars", "five-stars"])
             #foreach($r in [1..5])
                 #set($i = $r - 1)
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/rating/rating.css
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/rating/rating.css
@@ -22,8 +22,7 @@
   display: inline;
 }
 
-.star-rating a,
-.star-rating .current-rating {
+.star-rating a {
   position: absolute;
   top: 0;
   left: 0;
@@ -66,7 +65,7 @@
   z-index: 2;
 }
 
-.star-rating .current-rating {
+.star-rating .average-rating {
   z-index: 1;
   background-position: left center;
 }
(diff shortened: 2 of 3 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/rating_macros.vm
```diff
@@ -36,7 +36,9 @@
     <div class="rating-header $cssclass">$ratinglabel</div>
     <div class="rating-stars">
         <ul class="star-rating ${cssclass}-star #if($locked) locked #end">
-            <li class="current-rating" style="width:${width}%;"></li>
+            <li class="current-rating">
+                <meter class="average-rating" min="0" max="5" value="$rating"/>
+            </li>
```
surli, line 41: Very interesting case of "is this a potentially breaking DOM change or not?" :) I mean: here the chance that anyone is impacted is absurdly low so I don't see any reason to not accept the changes. But, in theory we could imagine somehow relying on some code to check the width of a `current-rating` class in javascript to perform some stuff...

surli: So as @Sereza7 mentioned in the chat: 
> It's been a while that we have don't use inline style declarations in [our front end code style](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/#HCSS)

so then it's clearly not an API we're breaking :)

Sereza7: (more discussion from the chat, moved here for the sake of keeping long term traces)

> I think this rating_macros.vm is also used to display ratings in pages, so it's not only for the admin
it's just that you don't have the Ratings Extension bundled by default so you need to install it first: you might want to check that

I just finished checking simple interactions, and it seems to behave like the one in extension behaved. I made minimal modifications, only removing the default inline styling and add a meter node in the list item. The part that would be difficult to handle for backward compatibility (handling how we update this at runtime) is not changed. It's not perfect, but the goal of this ticket is to avoid breaking web standard tests (only run at page load) and backward compatibility.

PR #2816, opened 2024-01-23
XWIKI-21779: Admin section: make the search suggest section pass webstandard tests

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21779

# Changes

## Description

* Updated the template to avoid duplicating IDs
* Updated the template to avoid incorrect checkbox in a label.
* Removed the `for` attribute for the labels in the template, because they didn't point towards any input.
* Added an empty dl after checkboxes, in order to keep a correct dd structure.
* Removed the comment from the test for searchSuggest

## Clarifications

* ID duplication: the Config and SourceSheet files share the same property name, which makes it weird with the template structure used in velocity. Somehow adding the empty object as a parameter to the `display` method changes the ID (adds a {} at the start, still a correct ID, and different from the one used in Config), but does not change how the template is used to create new sources. New sources are still correctly generated.

# Screenshots & Video

The only UI change is a small empty space after each checkbox in SourceSheets, that display the empty dl.
beforePR  vvv
21779-beforePR
afterPR vvv
21779-afterPR

# Executed Tests

Successfully passed `mvn clean install -f xwiki-platform-distribution/xwiki-platform-distribution-flavor/xwiki-platform-distribution-flavor-test/xwiki-platform-distribution-flavor-test-webstandards -Dxwiki.test.startXWiki=false` for the added page `/xwiki/bin/admin/XWiki/XWikiPreferences?editor=globaladmin&section=searchSuggest`.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestSourceSheet.xml
```diff
@@ -54,10 +54,10 @@
...
   #set ($isCheckbox = $editing &amp;&amp; $property.classType == 'Boolean' &amp;&amp; (!$displayFormType || $displayFormType.value == 'checkbox'))
   ; {{html wiki="true"}}
-    &lt;label#if ($editing) for="${prefix}$property.name"#end&gt;
-      #if ($isCheckbox)
-        $doc.display($property.name, 'edit')
-      #end
+    #if ($isCheckbox)
+      $doc.display($property.name, 'edit', $object)
+    #end
+    &lt;label#if ($editing &amp;&amp; $object!={}) for="${prefix}$property.name"#end&gt;
```
tmortagne, line 60: It's not very clear to me what `$object!={}` exactly does (feel like a comparison with an empty map, but `$object` is not a map, so I guess I'm missing something.).

Sereza7: @tmortagne It's not easy to understand what happens here in fact :)

The issue is that this content is called in two ways:
* multiple times to generate the search suggest source panel content (regular - visible use)
* once without a proper $object on the seach suggest page load (border - not visible use).
This secondary use stores the template in a hidden DOM element that can be used further down the line to generate the DOM for new sources without reloading the page.
This content is rendered by https://github.com/xwiki/xwiki-platform/blob/742cd4591642be4cdcaf68325f17540e0934e64e/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfigSheet.xml#L89 . The `$object` takes its value from the `$source` parameter. The value of this `$source` parameter is, in most cases, [a regular object](https://github.com/xwiki/xwiki-platform/blob/742cd4591642be4cdcaf68325f17540e0934e64e/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfigSheet.xml#L160). However, for the template, which has no object, an empty map is given as a parameter instead: https://github.com/xwiki/xwiki-platform/blob/742cd4591642be4cdcaf68325f17540e0934e64e/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/SearchSuggestConfigSheet.xml#L168 . If I understand well, we'd expect an empty object instead. I didn't want to create too much changes in this content, so I just rolled with it and added a condition that works with the empty map.

This condition on the object makes sure we don't add a `for` attribute to labels that would have no input bound to them.
In a few words, `$object` is not hard typed to an object, it's also sometimes a map, henceforth this weird looking condition.

Those changes are contained in the commit https://github.com/xwiki/xwiki-platform/pull/2816/commits/f73bcc7fd4d00f1a0138bd58688d4a704d3d08d2 .

Sereza7: Added  a comment in https://github.com/xwiki/xwiki-platform/pull/2816/commits/8031c992b289871e936d70de3542ab02d29130d7 :+1:

PR #2817, opened 2024-01-24
XWIKI-19965: Enable the images from new Image dialog Gallery to be selected on double click or with Enter key

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-19965
# Changes

## Description

* Added the listeners for double click and enter key pressed down
* Limited the listener for simple click to avoid collision with the behaviour of the double click one.

## Clarifications

* A lot of places recommend to avoid using `click` and `dblclick` events on the same element. Here I just used click and separated those events based on the details property. It seems to be more consistent and behaves just as expected.
* This implementation removes behaviours for click chains higher than 2 (event.detail > 2). I think it's okay to do so because it's rarely even a use case, and removing interaction at this point can actually help users.
* I decided not to implement the side changes proposed in the ticket description, and instead open [a new ticket](https://jira.xwiki.org/browse/XWIKI-21828) so that the scope of this PR doesn't grow too wide.

# Screenshots & Video

[Here is a demonstration video](https://up1.xwikisas.com/#qber9GZgFrkqrn42BFaFAQ), with an input display, of the changes added in this PR.
We can see that now, double clicking or pressing enter while focusing an image directly validates the selection.
# Executed Tests

Successfully built: `mvn clean install -amd -f xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker -pl :xwiki-platform-attachment-picker-ui,xwiki-platform-attachment-picker-webjar -Pquality,integration-tests,docker` and manual testing similar to the one in the demo video.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * N/A


xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-ui/src/main/resources/Attachment/Picker/Code/CKEditorImagePlugin.xml
```diff
@@ -134,6 +134,29 @@
     picker.on('xwiki:attachmentGalleryPicker:unselected', function() {
       imageSelector.updateSelectedImageReferences([], $(this));
     });
+    // We extend the attachment picker to allow quick form submission
+    let results = picker.find('.attachmentPickerResults');
```
manuelleduc, line 138: General remake, I tend to prefer `const` over `let` when the value is not modified elsewhere.
```suggestion
    const results = picker.find('.attachmentPickerResults');
```

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
```diff
@@ -179,16 +179,18 @@ define('xwiki-attachment-picker',
       initSelectedAttachment() {
         this.resultsBlock.on('click', 'a', (event) => {
           event.preventDefault();
-          const parent = $(event.currentTarget).parents('.attachmentGroup');
+          if (event.detail === 1) {
```
manuelleduc, line 182: Same here, I'm not sure to understand why the rest of this method needs to be wrapped in this if.

manuelleduc (commented):
Thanks for working on this. Looks good except form some code style improvements suggestions.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
Sereza7: I added more explanation in https://github.com/xwiki/xwiki-platform/pull/2817/commits/2368ef30d1e20bedcae42b80bc8f4fc5408fffd2.
The issue I was facing was that a click toggles selection of the image.
When double clicking, it created unexpected behaviour where the image would be selected only in between the two clicks. With this condition, we make sure we select (or unselect) only on the first click.

We can use the detail attribute of the click event to get the `number` of the click in the clickchain, see [this paragraph about it on MDN Webdocs](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#usage_notes).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-ui/src/main/resources/Attachment/Picker/Code/CKEditorImagePlugin.xml
```diff
@@ -134,6 +134,29 @@
...
+      event.preventDefault();
+      // Make sure we have a selected picture before going forward
+      let selected = $(event.currentTarget).parents('.attachmentGroup').data('id');
+      imageSelector.updateSelectedImageReferences([selected], picker);
+      // Validate the selection of the double click picture.
+      selectButton.trigger('click');
+    }
+    // Submit form on double click
+    results.on('click', 'a', (event) => {
+      if (event.detail === 2) {
```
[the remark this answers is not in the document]
Sereza7: I added more explanation in https://github.com/xwiki/xwiki-platform/pull/2817/commits/2368ef30d1e20bedcae42b80bc8f4fc5408fffd2.

We can use the detail attribute of the click event to get the `number` of the click in the clickchain, see [this paragraph about it on MDN Webdocs](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#usage_notes).

The event.which value is the id of the enter key. A hard coded value like that is used in [multiple places in the codebase](https://github.com/search?q=repo%3Axwiki%2Fxwiki-platform+%22%3D%3D+13%22&type=code) already. I don't like hardcoding it, but it's a very stable id, still supported by jquery, so I don't think this is an issue. See https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which , it's not standard anymore but still supported anywhere. From what I can remember, `jstree` also heavily relies on those integer key ids.
We might want to migrate to a more up to date attribute, but IMO it's not a priority and doesn't fall into this PR's scope.

Sereza7:
All comments have been addressed :+1:

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-ui/src/main/resources/Attachment/Picker/Code/CKEditorImagePlugin.xml
```diff
@@ -134,6 +134,33 @@
...
+      // The `detail` event property for a click indicates the position of this click in a chain of click.
+      // detail === 2 means that this click is the second click of a double click.
+      if (event.detail === 2) {
+        validateForm(event);
+      }
+    });
+    // Submit form on Enter key pressed
+    results.on('keypress', 'a', function(event) {
+      // 13 is the key id of the Enter key
+      if(event.which === 13) {
```
michitux, line 160:
```suggestion
      if (event.which === 13) {
```

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-ui/src/main/resources/Attachment/Picker/Code/CKEditorImagePlugin.xml
```diff
@@ -134,6 +134,33 @@
...
     });
+    // We extend the attachment picker to allow quick form submission
+    const results = picker.find('.attachmentPickerResults');
+    let selectButton = element.parents('.modal-content').find('.modal-footer .btn-primary');
+    function validateForm(event) {
+      event.preventDefault();
+      // Make sure we have a selected picture before going forward
+      let selected = $(event.currentTarget).parents('.attachmentGroup').data('id');
+      imageSelector.updateSelectedImageReferences([selected], picker);
+      // Validate the selection of the double click picture.
```
michitux, line 145: This comment is not correct, this code is also used for the enter key.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-webjar/src/main/webjar/attachmentGalleryPicker.js
michitux: Apparently, [click events](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) are also sent when pressing space or enter while the element is focused. Did you check that this change doesn't interfere with this?

michitux (commented):
In general, this looks fine, but I'm wondering if we shouldn't add an integration test to verify the new behavior?

Also, does this change the behavior of the enter key from selection to directly submitting the form? If yes, do you think that this is expected by users?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-picker/xwiki-platform-attachment-picker-ui/src/main/resources/Attachment/Picker/Code/CKEditorImagePlugin.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2817/commits/9f172766d215ed68b89ceb9bdbe883a4cc50c599 :+1:

Sereza7:
> Apparently, [click events](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) are also sent when pressing space or enter while the element is focused. Did you check that this change doesn't interfere with this? [Edit] Looking at the video suggests you checked that selection with the space key still works.

Note that on the video, the space key does nothing, but I pressed enter right after it and it does "everything" in one keypress (selection + form submission) :)
See the green note in https://w3c.github.io/uievents/#event-type-click .
It's true that browsers now support additional buttons on some elements. From what I know, this is `Space` on a button and `Enter` on an anchor. Here, we use anchors so we only need to care about the Enter presses. I tested it out and as expected pressing `Space` on those anchors does nothing. The first `Enter` press will already validate the form so we don't need to care about a double `Enter` click. I didn't find browser doc where they explicitely list those alternative ways to trigger a click though :/


> Also, does this change the behavior of the enter key from selection to directly submitting the form? If yes, do you think that this is expected by users?

It does change the behavior. I think it might be a bit disorientating for a few users at first, but easy to understand and integrate into new habits. In my opinion, pressing enter on an item is already a commitment (usually it'd take us to a new page). With this new behavior, we skip the extra steps in the procedure to make sure that the user wanted to go along with that picture.

Before: `I press enter on a picture, it is selected. There's only three actions from now on:
A. Submit the form and move to the next  step
B. Cancel image insertion (which will close the modal)
C. Close the modal
To continue with option A (what I do most of the time) I need to tab through all pictures and press space on the submit button.`
After: `I press enter on a picture, it is selected. The interface assumes I'm going with option A and submits the form right now. If I wanted option B I can just go back one more step in the procedure. Option C is done exactly the same as before.`
IMO most keyboard users would understand and appreciate the shortcut. The setback for activating it when they didn't mean to is small.
We could make this whole process easier to understand by putting the attachment name somewhere in the UI of the `Edit Image` step (as a reminder of the decision made in the first step), but I believe that's another issue :)

PR #2823, opened 2024-01-25
XWIKI-14342: Make Extension Manager use IconThemes

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-14342
# Changes

## Description

* Removed the Hardcoded silk icons from the CSS
* Fixed style to fit actual icon nodes in the same place as they used to be as background.
* Added them in the velocity template instead
* Fixed the behavior of status switch when loading dependencies.

## Clarifications

* Adding a spinner dynamically was already difficult enough, so I decided to just use the default xwiki file instead of going for an approach that would react to the current iconTheme, like the one in the velocity template.

# Screenshots & Video

[Here is a demo of the changes brought in this PR.](https://up1.xwikisas.com/#kyOdRVcs2yeSnSHfP9aBzQ) In the first part, we can see that the looks with the silk iconTheme are the same as they used to be. However in the second part, we can see that now Font Awesome icons are correctly used when FontAwesome is set as the iconTheme.
# Executed Tests

No test executed exept manual tests similar to the one in the demo.
# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 15.10.X


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.js
```diff
@@ -495,6 +495,12 @@ XWiki.ExtensionBehaviour = Class.create({
       parameters : formData,
       onCreate : function() {
         dependency.removeClassName('extension-item-unknown').addClassName('extension-item-loading');
+        // Remove the unknown icon.
+        dependency.removeChild(dependency.childNodes[1]);
+        // Add the load icon
+        let loadPath = '$xwiki.getSkinFile("icons/xwiki/spinner.gif")';
+        let loadIcon = '<img  src="'+ loadPath + '" alt=""/>';
```
manuelleduc, line 502:
```suggestion
        const loadPath = '$xwiki.getSkinFile("icons/xwiki/spinner.gif")';
        const loadIcon = '<img  src="'+ loadPath + '" alt=""/>';
```

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.js
```diff
@@ -495,6 +495,12 @@ XWiki.ExtensionBehaviour = Class.create({
       parameters : formData,
       onCreate : function() {
         dependency.removeClassName('extension-item-unknown').addClassName('extension-item-loading');
+        // Remove the unknown icon.
```
manuelleduc, line 498: This is not really clear why we need to do all this.
Also, I don't know this code much but why do you need to manipulate Velocity code from javascript?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm
```diff
@@ -1038,6 +1069,24 @@ $namespace##
     </div>
     <div class="extension-header">
       <h2 class="extension-title">
+        #if($!extensionStatus == 'core' ||
```
manuelleduc, line 1072: I don't like too much this duplicated logic. @tmortagne do you know if we have velocity services for this?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm
```diff
@@ -576,6 +576,14 @@ $namespace##
   </dd>
 #end
 
+#macro (spinner)
```
manuelleduc, line 579: Duplicate of 
- https://github.com/xwiki/xwiki-platform/blob/be3b9476e9bb02ccccbd1edeaf8e7bf6d058e916/xwiki-platform-core/xwiki-platform-export/xwiki-platform-export-pdf/xwiki-platform-export-pdf-ui/src/main/resources/XWiki/PDFExport/AdminSection.xml#L90
- https://github.com/xwiki/xwiki-platform/blob/466b48df3d2dab96b1bffda9e6fce90be848a4bf/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm#L579

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm
tmortagne: AFAIK, this concept of status is something @mflorea introduced in `#determineExtensionStatus`, and it does not exist on Java side.

Sereza7: Note that the logic is not duplicated, it looks very similar, but e.g. there's also the `incompatible` ones for dependencies.
This solution is very similar to what was done in CSS before (see the css removed from extension.css, it wasn't really refined either ^^' ). Instead of having a switch in the rulesets, I pretty much brought it in the template. I'd need to get back to it, but a more elegant solution would be to store the icon mapping in a unified table. This was just a minimal working alternative with icons actually in the HTML template and not only in the CSS.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.js
Sereza7: > Also, I don't know this code much but why do you need to manipulate Velocity code from javascript?

We could make it work with smart CSS relying on logic classes to hide and show icons. I tried to keep the changes to a minimum to avoid making this more complex. Keeping things simple means also that it's less likely to break things in a usecase/customization unknown to me. 

This code is quite old and large. A complete overhaul would be the best solution quality wise, but it's beyond the goal of this PR.

> This is not really clear why we need to do all this.

Before, the icon would swap directly when we swapped the classes. Here, I added a few lines to swap the icon DOM elements at the same time as they did before.

Sereza7: TODO: At least add a comment to explain the HTML manipulations. Should probably add CSS logic to avoid this  HTML manipulations. Do leave things like this if extensive testing show that it's not possible to have a simple CSS logic in place...

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm
Sereza7: :+1: 

TODO provide this spinner as a macro in https://github.com/xwiki/xwiki-platform/blob/29111f02b38e529f44ff88aaf4ec67af92fcf3df/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm (which from my experience is the main place where we provide API-ish velocimacros).

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.js
```diff
@@ -495,6 +495,12 @@ XWiki.ExtensionBehaviour = Class.create({
...
+        dependency.removeChild(dependency.childNodes[1]);
+        // Add the load icon
+        const loadPath = '$xwiki.getSkinFile("icons/xwiki/spinner.gif")';
+        const loadIcon = '<img  src="'+ loadPath + '" alt=""/>';
+        dependency.update(loadIcon + dependency.innerHTML);
```
mflorea, line 503: I'd rather use https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceWith to replace the unknown icon with the loading icon.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/extension/extension.js
```diff
@@ -495,6 +495,12 @@ XWiki.ExtensionBehaviour = Class.create({
       parameters : formData,
       onCreate : function() {
         dependency.removeClassName('extension-item-unknown').addClassName('extension-item-loading');
+        // Remove the unknown icon.
+        dependency.removeChild(dependency.childNodes[1]);
+        // Add the load icon
+        const loadPath = '$xwiki.getSkinFile("icons/xwiki/spinner.gif")';
+        const loadIcon = '<img  src="'+ loadPath + '" alt=""/>';
```
mflorea, line 502: This doesn't match the behavior of the spinner Velocity macro. A possible solution is to always output the spinner from ``extension.vm``, but hidden by default. It would be made visible by the ``extension-item-loading`` CSS class, which means no change in needed in this JavaScript file.

mflorea (changes requested):

Sereza7:
Since there's a project of reworking the UI completely, and this PR is still open as a draft, I'm closing it. If the changes here become relevant again, feel free to use them as a basis for a new PR :)

PR #2827, opened 2024-01-26
XWIKI-18998: Replace the Livetable of the System Filters of the Notifications Administration with a Live Data macro

manuelleduc:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-18998

# Changes

## Description

* Replace the System Filters Live Table with a Live Data

Bonus: better handling of the toggles on error

## Clarifications

* Introduction of a ToggleDisplayer Vue component for the display of the togglers
* All the Livetable hacking logic replaced by a listener to the toggles events

# Screenshots & Video

## Before

image

## After

image

## Error handling

**TODO**

# Executed Tests

```
mvn clean install \
  -pl :xwiki-platform-livedata-webjar,:xwiki-platform-notifications-ui,:xwiki-platform-notifications-test-docker \
  -Pquality,integration-tests,docker
```

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * N/A


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-ui/src/main/resources/XWiki/Notifications/Code/NotificationFilterPreferenceLivetableResults.xml
```diff
@@ -108,11 +108,11 @@
         #set ($objectNumber = $!obj.reference.objectNumber)
       #end
       #if ($checked)
-        #set ($checked = 'checked = "checked"')
+        #set ($checkedAttr = 'checked = "checked"')
```
surli, line 111: I'm honestly wondering if it wouldn't be better for the long term maintenance to get rid of that, and provide a dedicated Java source for the LD. Feels like it would be easier to test / maintain, no?

manuelleduc: I kept that one because I considered it could be used by extensions. But yes, might not be a bad idea to provide a clean source and to move that whole one to legacy.

manuelleduc: Well, actually there is 3 other livetable using this results page, so we'll need to migrate them as well before being able to depracate.

surli: > I kept that one because I considered it could be used by extensions.

I honestly doubt it but yes it makes sense to be conservative. 

> Well, actually there is 3 other livetable using this results page, so we'll need to migrate them as well before being able to depracate.

I was thinking about doing that for the custom notification filters: I don't recall but I guess it's same page used for those too. Actually I'm surprised: you do have filtering / sorting with that page result? Or the filtering / sorting only work on frontend with the live data here?

surli: If you think it's ok maybe we can do the move by steps: 
  1. merge that,
  2. create an improvment ticket to have dedicated source for LD for those
  3. open a vote to get rid of this LT results (would avoid having to maintain it on the long run) 
  4. migrate other LT to LD using the source from 2 and migrate again that one with the source too

manuelleduc: +1, I'm taking a close look at the migration, and it's not "too" difficult (modulo the random bumps you can get when migrating messy Velocity script), but this PR is already quite large, so I'm not against delaying this to another step.

manuelleduc: > Actually I'm surprised: you do have filtering / sorting with that page result? Or the filtering / sorting only work on frontend with the live data here?

No, I just realized the mistake when looking at the migration to a LD source. Fixing that.

manuelleduc: Should be better with https://github.com/xwiki/xwiki-platform/pull/2827/commits/75fd7efd7cf494a690beff424356b4514abe1c0d

manuelleduc: > 2\. create an improvment ticket to have dedicated source for LD for those

https://jira.xwiki.org/browse/XWIKI-21848

PR #2835, opened 2024-02-01
XWIKI-21729: Bad interaction between quick search and the drawer

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21729
# Changes

## Description

* Fired an event that was overwise unused
* Fixed its listener
* Added an optional argument for the suggest resetTimeout function (to allow a more precise control of the suggest panel)

# Screenshots & Video

Video demo: https://youtu.be/_XcDOmUG870
# Executed Tests

None, because 'cause changes' in [XWIKI-16140](https://jira.xwiki.org/browse/XWIKI-16140) and/or [XWIKI-20733](https://jira.xwiki.org/browse/XWIKI-20733) did not impact any tests (except webstandards, for reasons unrelated to the changes in this PR).
# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 15.10.X to match with XWIKI-16140 and XWIKI-20733. IMO this is quite a low risk backport.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-ui/src/main/resources/XWiki/QuickSearchUIX.xml
@@ -136,6 +136,7 @@
       globalSearch.addClass('globalsearch-close');
       globalSearchButton.attr('aria-expanded', 'false');
       globalSearchInput.attr('disabled','');
+      document.fire('xwiki:suggest:collapsed');
     }
 
     // Open the global search when the user click on the global search button
--- a/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-webjar/src/main/webjar/searchSuggest.js
+++ b/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-webjar/src/main/webjar/searchSuggest.js
@@ -127,7 +127,8 @@ var XWiki = (function (XWiki) {
      * Callback triggered when the suggest element is collapsed, because of a focusout event for example.
      */
     onSuggestCollapsed: function(event) {
-      /* We match the collapse of the input field timing to close the suggest panel. */
+      /* We match the collapse of the input field timing to close the suggest panel.
+         As of 16.2.0-RC1, this value is defined in action-menus.less -> #headerglobalsearchinput */
       this.suggest.resetTimeout(300);
     },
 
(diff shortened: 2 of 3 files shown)


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/search/searchSuggest.js
```diff
@@ -118,7 +118,8 @@ var XWiki = (function (XWiki) {
      * Callback triggered when the suggest element is collapsed, because of a focusout event for example.
      */
     onSuggestCollapsed: function(event) {
-      this.suggest.clearSuggestions();
+      /* We match the collapse of the input field timing to close the suggest panel. */
```
surli, line 121: Where is located the duration of the input field timing? Not clear to me when reading the changes where I could find this `300` value.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/search/searchSuggest.js
```diff
@@ -118,7 +118,9 @@ var XWiki = (function (XWiki) {
      * Callback triggered when the suggest element is collapsed, because of a focusout event for example.
      */
     onSuggestCollapsed: function(event) {
-      this.suggest.clearSuggestions();
+      /* We match the collapse of the input field timing to close the suggest panel.
+         As of 16.2.0-RC1, this value is defined in action-menus.less -> #headerglobalsearchinput */
```
surli, line 122: Thanks for the doc improvment. Just to be sure: there's no way to actually extract the data from the CSS rule with some javascript? e.g. in xwiki.org I'm able to get it using `jQuery('#headerglobalsearchinput').css('transition-duration')` (not sure it works with all browsers tbh). 

My point here is I'm afraid we forget to update this value if we ever change the less file and it can also be modified for having a custom skin, which would make that code not efficient.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/search/searchSuggest.js
Sereza7: Updated the comment to be more explicit :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/search/searchSuggest.js
Sereza7: Thanks for the comment! I didn't know JQuery offered this feature. It seems to work well on recent FF and Chromium.
I updated it in https://github.com/xwiki/xwiki-platform/pull/2835/commits/0531d28a63b1222a2fdc3e27d2ae2953466a42d7 

I never even looked for such a feature because it looked to me like a violation of the MVC principle. Having the object behave in a different way only based on its style didn't sound right to me. In this use case it's convenient and somewhat appropriate :+1:

PR #2836, opened 2024-02-01
XWIKI-21730: Delete own comments should not require edit rights on page

Sereza5:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21730

# Changes

## Description

* Changed the permission needed for the delete comment action to be displayed

## Clarifications

* [The code responsible for allowing or not to delete the comment](https://github.com/xwiki/xwiki-platform/blob/5be65a1621598b4d2b407197fb50d827c61e15a2/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm#L187) was introduced in version 3.0-rc1 (see [the commit](https://github.com/xwiki/xwiki-platform/commit/8255d1990233c8bfca8115d6e0db828d93eb4dfb)), and apparently it was not a use case considered by those changes. So I don't see an issue with tying the comment deletion action to the comment right instead of the edit right.
* The template for generating these action buttons is a bit of a mess. I considered for a moment changing the order of the buttons to make it cleaner. However after considering stakes, I realized the UI inconsistency between versions would be worst for users than the small code quality/optimization we would have gotten.

# Screenshots & Video

In order to test those changes, I registered as a regular user (named '123456'). With the admin user, I changed regular user rights at the wiki level: removed the Edit right and kept the Comment right.
We can see that the `delete` button only appears after the changes in this PR. This is the behavior expected by the reporter of this ticket.
Before PR:
21730-beforePR
After PR:
21730-afterPR

# Executed Tests

None, this change only changes the presence of the button in specific right configurations (No edit right but comment right, different from the default Edit+comment rights). I could only see one reference to this [in a pageobject](https://github.com/xwiki/xwiki-platform/blob/69ac280d43999572e6a3ad54a05151769111c5e9/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/CommentsTab.java#L132-L140). This function is used only once with regular logged in user rights (in [here](https://github.com/xwiki/xwiki-platform/blob/69ac280d43999572e6a3ad54a05151769111c5e9/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker/src/test/it/org/xwiki/flamingo/test/docker/CommentsIT.java#L110)).
Manual tests were conducted successfully, see the screenshots above.
# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 15.10.X - minor change with very low risk


Sereza7:
Added backend support for deletion of comments. This new action is very similar to regular object removal (which was the action used before), but is based on the Comment right instead of the Edit right.

Changes are mostly contained in the oldcore module, because that's where all the backend logic for actions used in those buttons was already contained.

See [a demo of this updated action](https://youtu.be/LwnLWfZFYZw).
This demo worked with just creating two symlinks on the updated jars for xwiki-oldcore and xwiki-security-authorization-bridge , and updating the commentsinline.vm template.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -184,7 +184,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
       <a class="permalink btn btn-default btn-xs" data-toggle="modal" data-target="#permalinkModal" rel="nofollow"
         href="$doc.getURL('view', 'viewer=comments')#xwikicomment_${comment.number}"
         title="$services.localization.render('core.viewers.comments.permalink')">$services.icon.renderHTML('link')</a>
-      #if ($hasAdmin || ($hasEdit && $isUserComment))
+      #if ($hasAdmin || ($xwiki.hasAccessLevel('comment') && $isUserComment))
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2836/commits/f536a96693cf815abc7e26f7433f2c318f1bc268 and https://github.com/xwiki/xwiki-platform/pull/2836/commits/f943ae67afdabe4b2491b00d3e0a1812770f21ed . :+1: 


This selector was a bit more restrictive, actually checking for the comment right. Now, any user can delete their comments even if they don't have any right on a page.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -184,7 +184,7 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
       <a class="permalink btn btn-default btn-xs" data-toggle="modal" data-target="#permalinkModal" rel="nofollow"
         href="$doc.getURL('view', 'viewer=comments')#xwikicomment_${comment.number}"
         title="$services.localization.render('core.viewers.comments.permalink')">$services.icon.renderHTML('link')</a>
-      #if ($hasAdmin || ($hasEdit && $isUserComment))
+      #if ($hasAdmin || $isUserComment)
```
surli, line 187: What about introducing a new variable to use in both if for edit and delete? That way we ensure that we'll keep them consistent.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,159 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xpn.xwiki.web;
```
surli, line 20: Hmmm normally we shouldn't introduce new stuff in this package, we should instead put stuff in `org.xwiki...` package, even in oldcore... Now it would make this class far from the other actions, so I'm not sure here...

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,159 @@
...
+    public String render(XWikiContext context) throws XWikiException
+    {
+        if (Utils.isAjaxRequest(context)) {
+            XWikiResponse response = context.getResponse();
+            response.setStatus(HttpServletResponse.SC_CONFLICT);
+            response.setContentType("text/plain");
+            try {
+                response.getWriter().write("failed");
+                response.setContentLength(6);
+            } catch (IOException e) {
```
surli, line 152: Nothing to handle the exception?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    public String render(XWikiContext context) throws XWikiException
+    {
+        if (Utils.isAjaxRequest(context)) {
+            XWikiResponse response = context.getResponse();
+            response.setStatus(HttpServletResponse.SC_CONFLICT);
+            response.setContentType("text/plain");
+            try {
+                response.getWriter().write("failed");
+                response.setContentLength(6);
```
surli, line 151: Hmm really not a fan of this piece of code: at the very least "failed" should be in a String variable and you should eruse it also for the length. Now I'm not sure why you just answer "failed"?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Action used to remove a comment from a page, requires comment right but not edit right.
+ *
+ * @version $Id$
+ * @since 16.1.0RC1
+ */
+@Component
+@Named("commentdelete")
+@Singleton
+public class CommentDeleteAction extends XWikiAction
```
surli, line 53: A test class should be provided too.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
Sereza7: Done in https://github.com/xwiki/xwiki-platform/pull/2836/commits/9f841ae94d968251156fe49ca69e011e87cf20cb :+1:

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
Sereza7: Note that most parts are the same as `ObjectRemoveAction`. This empty exception handle has been in this other file since 2008 so it seems like it's not a huge problem.
This is bad quality I agree.
I added a log similar to what was done in XWikiAction.java
https://github.com/xwiki/xwiki-platform/blob/4e2efc811fd359eaf87f1bc1e43e78d54777a801/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/XWikiAction.java#L385 

Addressed in https://github.com/xwiki/xwiki-platform/pull/2836/commits/767daaca4e1b6aa5e143e872c725c742e83c8bbb :+1:

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
Sereza7: Note that this part is pretty much the same as `ObjectRemoveAction`. 
I updated it and also changed `ObjectRemoveAction` to match these changes.

See https://github.com/xwiki/xwiki-platform/pull/2836/commits/6edbec17009036c8375bf8d60f5ca3dc128279e8 :+1:

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,159 @@
...
+            return true;
+        }
+
+        doc.removeObject(obj);
+        doc.setAuthorReference(userReference);
+
+        String changeComment = localizePlainOrReturnKey("core.comment.deleteComment");
+
+        // Make sure the user is allowed to make this modification
+        context.getWiki().checkSavingDocument(userReference, doc, changeComment, true, context);
```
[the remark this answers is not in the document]
Sereza7: From what I can understand of this code I reused from ObjectRemoveAction, this does not check the rights of the current user but the state of the document (whether it's in a saveable state or not). I'm retesting things manually to make sure everything is alright.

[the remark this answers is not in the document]
Sereza7: Okay, I finally figured it out. The authorisation to do this action is handled at the level of `XWikiCachingRightService`. This is also the reason why we need an independant action even thought it's pretty much the same as object deletion. From what I understand, one action can only be mapped to one right. Setting the other object editions to depend on the Comment right is wrong, we need to separate this into two actions.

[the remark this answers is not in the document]
Sereza7: I can confirm this. I forgot to put the security .jar on my local instance and the action wouldn't go through (with edit rights disabled but comment rights enabled). Updating it as expected fixed the backend permission issue.

Sereza7:
This PR is ready for further reviewing.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
```diff
@@ -177,14 +178,14 @@ $xwiki.ssfx.use('uicomponents/viewers/comments.css', true)
...
-        #if($hasAdmin || $isUserComment)
+        #if($hasCommentEditionRight)
           <a class="edit btn btn-default btn-xs" rel="nofollow" href="$doc.getURL('view', "viewer=comments&amp;number=${comment.number}&amp;xredirect=$doc.getURL('view')")" title="$services.localization.render('core.viewers.comments.edit')">$services.icon.renderHTML('pencil')</a>
         #end
       #end
       <a class="permalink btn btn-default btn-xs" data-toggle="modal" data-target="#permalinkModal" rel="nofollow"
         href="$doc.getURL('view', 'viewer=comments')#xwikicomment_${comment.number}"
         title="$services.localization.render('core.viewers.comments.permalink')">$services.icon.renderHTML('link')</a>
-      #if ($hasAdmin || ($hasEdit && $isUserComment))
+        #if ($hasCommentEditionRight)
```
surli, line 188: Hmmm I'm not so sure about that change... Yeah I know I commented to have consistency in an early review :) 
Have you tested if it actually works to edit a comment for a page in which the comment user doesn't have edit right?

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/CommentDeleteActionTest.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for {@link com.xpn.xwiki.web.CommentDeleteAction}.
+ *
+ * @version $Id$
+ * @since 17.0.0RC1
+ */
+@ComponentList
```
surli, line 59: Hmm wrong annotation I think: this one takes as values a list of component. I guess you wanted to use `@ComponentTest`?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
tmortagne, line 188: Is the indentation change on purpose ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
tmortagne: Honestly it does not make much sense to have CommentAddAction here and CommentDeleteAction elsewhere so +1 to keep it here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/ObjectRemoveAction.java
```diff
@@ -41,6 +43,8 @@
 @Singleton
 public class ObjectRemoveAction extends XWikiAction
 {
+    private static final String FAIL_MESSAGE = "failed";
+    private static final Logger LOGGER = LoggerFactory.getLogger(ObjectRemoveAction.class);
```
tmortagne, line 47: Same comment as above about injected Loggers.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/CommentDeleteActionTest.java
tmortagne: Note that you should not use `@ComponentTest` either since you have `@OldcoreTest`.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/CommentDeleteActionTest.java
```diff
@@ -0,0 +1,157 @@
...
+@ReferenceComponentList
+@OldcoreTest(mockXWiki = false)
```
tmortagne, line 61: `mockXWiki = false` sounds a bit strange, this was basically introduced for test of XWiki

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Action used to remove a comment from a page, requires comment right but not edit right. Note that this class is
+ * largely inspired by ObjectRemoveAction and Comment
+ *
+ * @version $Id$
+ * @since 17.0.0RC1
+ */
+@Component
+@Named("commentdelete")
```
tmortagne, line 54: Nothing critical but `commentremove` feel more consistent with `objectremove`, especially since it's reusing `ObjectRemoveForm`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,163 @@
...
+@Singleton
+public class CommentDeleteAction extends XWikiAction
```
tmortagne, line 56: Might worth moving most of this class in some `AbstractObjectRemoveAction` that would also be extended by `ObjectRemoveAction`.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,163 @@
...
+            getCurrentScriptContext().setAttribute(attributeName,
+                localizePlainOrReturnKey("platform.core.action.commentRemove.noClassnameSpecified"),
+                ScriptContext.ENGINE_SCOPE);
+        } else if (classId < 0) {
+            getCurrentScriptContext().setAttribute(attributeName,
+                localizePlainOrReturnKey("platform.core.action.commentRemove.noCommentSpecified"),
+                ScriptContext.ENGINE_SCOPE);
+        } else {
+            // Comment class reference
+            DocumentReference commentClass = new DocumentReference(context.getWikiId(), XWiki.SYSTEM_SPACE,
```
tmortagne, line 89: Why do you need to create a DocumentReference ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return ObjectRemoveForm.class;
+    }
+
+    protected BaseObject getObject(XWikiDocument doc, XWikiContext context)
+    {
+        ObjectRemoveForm form = (ObjectRemoveForm) context.getForm();
+        BaseObject obj = null;
+
+        String className = form.getClassName();
```
tmortagne, line 76: Manipulating the class name does not make sense for this action, since it's always about comments (or only to return an error if someone tried to be clever and use this action to delete another type of xobject with just comment right).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/commentsinline.vm
Sereza7: https://github.com/user-attachments/assets/3062fcfc-e464-4a2c-b2da-29dbaedcdc70

I'm not sure why this wouldn't work anymore. The boolean logic is the same as it was before isn't it?
`$hasCommentEditionRight` is always set/updated a few lines above and takes the current context values for rights.

Sereza7: This was a mistake. We don't indent for the div a few lines above, this is probably what threw me off here. The `#end` tag is not intended, I removed this indentation change in https://github.com/xwiki/xwiki-platform/pull/2836/commits/c0db94eb7fe64a370e3d5da9e765a917a4b82a4e :+1:

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
Sereza7: Because we shouldn't use `doc,getObject` anymore, it's deprecated. AFAIR, getXObject only accepts reference parameters and not Strings.

xwiki-platform-core/xwiki-platform-oldcore/src/test/java/com/xpn/xwiki/web/CommentDeleteActionTest.java
Sereza7: Okay! Removed it in https://github.com/xwiki/xwiki-platform/pull/2836/commits/a71eee7b1ea33a6b7333521dc3feacc29ca0272e :+1:

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
Sereza7: :+1: Updated the action name in https://github.com/xwiki/xwiki-platform/pull/2836/commits/a71eee7b1ea33a6b7333521dc3feacc29ca0272e :+1:

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
Sereza7: @tmortagne  I tried factorize this code by moving it to an abstract class like you recommended. However, I hit the revapi fail:

```
[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.15.0:check (revapi-check) on project xwiki-platform-legacy-oldcore: The following API problems caused the build to fail:
[ERROR] java.class.nonFinalClassInheritsFromNewClass: class com.xpn.xwiki.web.ObjectRemoveAction: Non-final class now inherits from 'com.xpn.xwiki.web.AbstractObjectRemoveAction'. https://revapi.org/revapi-java/differences.html#java.class.nonFinalClassInheritsFromNewClass
```

AFAIU, there was nothing final in this class, so everything is fine. Do you know how we should ignore this warning or find an alternative solution to avoid triggering it?
For now I just added a revapi.skip in the legacy POM, but it's just a temporary solution to be able to build and test locally.

Sereza7:
WIP, still need to handle the className differently in both.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/web/CommentDeleteAction.java
```diff
@@ -0,0 +1,163 @@
...
+ * @since 17.0.0RC1
+ */
+@Component
+@Named("commentdelete")
+@Singleton
+public class CommentDeleteAction extends XWikiAction
+{
+    private static final String FAIL_MESSAGE = "failed";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CommentDeleteAction.class);
```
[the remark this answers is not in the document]
Sereza7: Okay! Thanks for the clarification :)

I updated it in https://github.com/xwiki/xwiki-platform/pull/2836/commits/a71eee7b1ea33a6b7333521dc3feacc29ca0272e :+1:

PR #2880, opened 2024-02-13
PNPM Experiment

manuelleduc:
# Changes

## Description

Demo of the use of https://github.com/xwiki/xwiki-commons/pull/772 in `xwiki-platform`
Two `webjar-node` modules are defined, `xwiki-platform-lib-a` and `xwiki-platform-lib-b`.
`xwiki-platform-lib-b` depends on `xwiki-platform-lib-a`.
Both projects are defined using typescript and are type checked.

This PR is currently mainly a support for the forum proposal that will follow.


pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - "./**/src/main/node"
```
manuelleduc, line 2: Any project containing a `src/main/node` directory is considered as a pnpm package.

xwiki-platform-core/xwiki-platform-lib-b/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+  <packaging>webjar-node</packaging>
+  <artifactId>xwiki-platform-lib-b</artifactId>
+  <name>Lib B NPM</name>
+  <properties>
+    <xwiki.extension.name>Lib B NPM</xwiki.extension.name>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-lib-a</artifactId>
```
manuelleduc, line 39: Runtime dependency to lib-a for the webjar level.
As of now, I don't think we have a better solution than to manually maintain modules dependencies at maven and npm levels (i.e., so code duplication).

manuelleduc:
Closing because the ideas of this PoC are not applied through other commits.

PR #2885, opened 2024-02-14
XWIKI-21884: Uploading files during page creation step will mark XWikiGuest as creator

oanalavinia:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21884

# Changes

## Description

* check if the document where the current attachment is created is actually a new document and set the creator in this case

## Clarifications

* for more context on when the issue can occur, there is https://jira.xwiki.org/browse/XWIKI-16476 and https://jira.xwiki.org/browse/XWIKI-16476 . These introduced the bug on Office Viewer Macro, but it is the reproducible for other macros that use the attachment picker as a parameter

# Screenshots & Video

**Before:**

createdByGuestUser

**After:**

guestUserCreatorAfter

# Executed Tests

Run  `mvn clean install -Pquality,integration-tests,docker` in module `xwiki-platform-rest`.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x


manuelleduc (commented):
The code looks good, but I'd suggest adding a unit test along your change.

oanalavinia:
I adjusted an existing unit test to verify the creator of the page as well. It was assuming XWikiGuest before

PR #2886, opened 2024-02-14
XWIKI-20477: Upgrade to Solr 9

tmortagne:
# Jira URLs

[XWIKI-20477](https://jira.xwiki.org/browse/XWIKI-20477): Upgrade to Solr 9.4.1 and Lucene 9.8.0
[XWIKI-12955](https://jira.xwiki.org/browse/XWIKI-12955): Upgrade to JAX-RS 2.1.6 (JSR370)
[XCOMMONS-2407](https://jira.xwiki.org/browse/XCOMMONS-2407): Upgrade to Antlr 4.13.1
[XWIKI-21709](https://jira.xwiki.org/browse/XWIKI-21709): Upgrade to Jetty Client 10.0.19
[XWIKI-12674](https://jira.xwiki.org/browse/XWIKI-12674): Replace Restlet with Jersey
[XWIKI-19813](https://jira.xwiki.org/browse/XWIKI-19813): Automatically migrate Solr cores when the major version changes

# Changes

## Description

The main goal was to upgrade to Solr 9.

But upgrading Solr also requires:
* Upgrade to JAX-RS 2.1
* Upgrade to Jetty Client 10.0.19+
* Upgrade to Antlr 4.11.1+

To make migration of some cores from Solr 8 to Solr 9 easier (which means having both versions of the core usable at the same time) the naming of the cores changed a bit: they are now all suffixed with the major version of Solr. 

A new XWikiSolrCore was introduced to hold various information about the core. We used to pass around the SolrClient, but it's missing some information like the core name, and it will be interesting to have a more obvious location to put some Solr core related tool.

Anything that was using Restlet classes (quite a lot more than expected...) moved to using pure JAX-RS APIs (and in some rare cases Servlet API).

### JAX-RS and Jersey

Solr moved to JAX-RS 2. Problem is that Restlet is not compatible with it so we had to find another implementation and since Solr uses Jersey now as JAX-RS implementation, moving to Jersey ourselves is what made the most sense (even if there are other implementations of JAX-RS 2.1).

Restlet had quite a lot of non JAX-RS standard hacks and while I kept some that I have hit during tests (for example forcing a specific media type through the URL parameter `?media=json`) I did not keep everything. Those are mainly located in `org.xwiki.rest.internal.PreMatchingRequestFilter`.

While JAX-RS fixed quite a few Restlet bugs in the implementation of `javax.ws.rs.core.UriBuilder` it's very possible that someone might have relied on one of those bugs. For example, I found code in `org.xwiki.test.ui.TestUtils` which was expecting `/` to not be escaped in elements passed to `UriBuilder#build` (which does not really make any sense).

We are proxying the internal tool used by Jersey to do injection because XWiki JAX-RS components (anything implementing `org.xwiki.rest.XWikiRestComponent`) are expected to be XWiki components too with full support of the XWiki component system (and especially injection of other XWiki component and other XWiki component related tools and metadata). Another consequence is that we had to take care of the injection of JAX-RS `javax.ws.rs.core.Context` annotated fields on XWiki side.

### Antlr 4.10+

Any lexer which was built with Antlr <4.10 won't work anymore (for example, it's the case of the [dokuwiki syntax parser](https://extensions.xwiki.org/xwiki/bin/view/Extension/DokuWiki/DokuWiki%20Syntax%20Parser/)).

# Executed Tests

Not much in terms of new tests, but those changes impact most integration tests (anything which manipulates an XWiki REST endpoint or Solr).

There is still one planned integration test: add an XWiki upgrade test with custom data containing Solr 8 cores.

# Expected merging strategy

* Prefers squash: No
* No backport


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/ImageIT.java
```diff
@@ -65,7 +65,14 @@
...
  * @since 14.7RC1
  */
-@UITest
+@UITest(
+    extraJARs = {
+        // It's currently not possible to install a JAR contributing a Hibernate mapping file as an Extension. Thus
+        // we need to provide the JAR inside WEB-INF/lib. See https://jira.xwiki.org/browse/XWIKI-8271
+        "org.xwiki.platform:xwiki-platform-notifications-filters-default"
+    },
+    resolveExtraJARs = true
```
surli, line 74: Not specifically related to your changes, is it?

vmassol:
> Any lexer which was built with Antlr <4.10 won't work anymore (for example, it's the case of the [dokuwiki syntax parser](https://extensions.xwiki.org/xwiki/bin/view/Extension/DokuWiki/DokuWiki%20Syntax%20Parser/)).

So this is a breaking change we need to document in the backward-compat section of the RN, right?

Are there any other breaking changes you know of? (for example does the move to Jersey has impacts that we need to mention in the RN?).

Thanks for your work on this! Looks complex

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
```diff
@@ -138,25 +143,34 @@ public abstract class AbstractSolrCoreInitializer implements SolrCoreInitializer
...
 
     @Inject
-    protected SolrSchemaUtils solrSchemaUtils;
+    protected SolrUtils solrUtils;
 
-    protected SolrClient client;
+    protected XWikiSolrCore core;
 
-    protected Map<String, FieldTypeRepresentation> types;
+    @Deprecated(since = "16.1.0RC1")
```
surli, line 156: Missing some doc to explain what to use now it's deprecated: I saw in one of the initializer that the call should now be `this.core.getClient()` right?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/DefaultJAXRSUtils.java
```diff
@@ -0,0 +1,142 @@
...
+    @Inject
+    private Container container;
+
+    @Override
+    public Charset getCharset(MediaType mediaType)
+    {
+        if (mediaType != null) {
+            String name = mediaType.getParameters().get(MediaType.CHARSET_PARAMETER);
+
+            if (name != null) {
```
surli, line 66: The name can never be empty as ""?

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/JAXRSUtils.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@Role
+@Unstable
+public interface JAXRSUtils
+{
+    /**
+     * Extract the charset from the media type.
+     * 
+     * @param mediaType the media type of the request
+     * @return the charset of the media type
```
surli, line 49: missing doc to specify the default returned value when it cannot be guessed from the media type.

xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-rest/src/main/java/org/xwiki/notifications/rest/NotificationsResource.java
```diff
@@ -126,9 +127,9 @@ String getNotificationsRSS(
      * Get notifications matching the given parameters. The POST method is used to allow large content in the
      * parameters. For example, the parameter "blackList" could be very long, and the associated URL with the GET method
      * would be too long (generating HTTP 414 error).
-     *
-     * Note: in the interface, we do not list the parameters, because Restlet does not support @FormParam...
-     * See: https://github.com/restlet/restlet-framework-java/issues/1120
+     * <p>
+     * While the signature of the method changed from Java point of view in 16.0.0, it's still exactly the same API from
```
michitux, line 131:
```suggestion
     * While the signature of the method changed from Java point of view in 16.1.0, it's still exactly the same API from
```

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
```diff
@@ -342,6 +342,129 @@ protected void initializeCoreSchema() throws SolrException
...
+            } catch (Exception e) {
+                throw new SolrException("Failed to save the document", e);
+            }
+
+            // Add it to the list of documents to delete
+            delete(source, sourceCore);
+        }
+
+        // Delete the documents
+        delete(source, targetCore);
```
michitux, line 440: Why are the documents deleted from the target core?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
```diff
@@ -138,25 +143,34 @@ public abstract class AbstractSolrCoreInitializer implements SolrCoreInitializer
 
     private static final String SOLR_VERSIONFIELDTYPE_VALUE = "defVal";
 
+    private static final int MIGRATION_BATCH_ROWS = 100;
```
michitux, line 146: This seems quite low, couldn't we set this to a higher value like 1000? Or do we really have Solr cores with large documents?

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
```diff
@@ -342,6 +342,129 @@ protected void initializeCoreSchema() throws SolrException
...
+                throw new SolrException("Failed to save the document", e);
+            }
+
+            // Add it to the list of documents to delete
+            delete(source, sourceCore);
+        }
+
+        // Delete the documents
+        delete(source, targetCore);
+        commit(sourceCore);
```
michitux, line 441: Why is the source but not the target committed? Wouldn't it make sense to first commit the target and then the source to ensure no data is lost?

xwiki-platform-distribution/xwiki-platform-distribution-debian/pom.xml
```diff
@@ -63,7 +63,8 @@
     <debian.homepage>http://www.xwiki.org</debian.homepage>
     <debian.controlDir>${project.build.directory}/maven-deb-control-directory</debian.controlDir>
 
-    <xwikiDataDir>/var/lib/xwiki/data</xwikiDataDir>
+    <debian.xwiki.var>/var/lib/xwiki</debian.xwiki.var>
+    <xwikiDataDir>${debian.xwiki.var}/data</xwikiDataDir>
```
michitux, line 67: Is this really related to this PR?

michitux:
Given my comments on the Solr core migration, is there any test that implicitly or explicitly tests Solr core migration? I didn't see any, but maybe I missed it.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-docker/src/test/it/org/xwiki/ckeditor/test/ui/ImageIT.java
tmortagne: No, it's just that without it the generated instance is unusable so I tough I would fix that.

tmortagne:
> Are there any other breaking changes you know of? (for example does the move to Jersey has impacts that we need to mention in the RN?).

Yes, I mentioned stuff that Restlet was adding (like some other special URL parameters) which I did not keep because it was not documented or used anywhere in XWiki.

tmortagne:
> Given my comments on the Solr core migration, is there any test that implicitly or explicitly tests Solr core migration? I didn't see any, but maybe I missed it.

Not yet, I plan to add an XWiki upgrade test based on a custom data which contains a Solr 8 with cores from back then.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/internal/DefaultJAXRSUtils.java
tmortagne: Does not sound like a valid use case to me, in this case you will get the same UnsupportedCharsetException as you would with any other bad charset.

xwiki-platform-core/xwiki-platform-rest/xwiki-platform-rest-server/src/main/java/org/xwiki/rest/JAXRSUtils.java
tmortagne: Indeed, better be explicit, as changing the default charset might be an API breakage.

> when it cannot be guessed from the media type

When no charset is provided at all.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
tmortagne: Indeed, this is wrong. I remember testing a migration by hand, but I think I then simplified the code and looks like I forgot to test again.

xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-api/src/main/java/org/xwiki/search/solr/AbstractSolrCoreInitializer.java
tmortagne: Indeed, the cores which are actually going to be migrated in XWiki Standard are not supposed to have very big document, but it might be the case later or in some extension.

To be safe, I'm going to keep 100 by default and allow AbstractSolrCoreInitializer to provide their own values (could even be more than 1000 for ratings for example, since those are supposed to be very small documents).

tmortagne:
> I plan to add an XWiki upgrade test based on a custom data which contains a Solr 8 with cores from back then

Done, migrated from a custom 12.10.11.

PR #2895, opened 2024-02-15
XWIKI-21894: JSC_PARSE_ERROR in Tour.WebHome

amottier:
# Changes

## JIRA

https://jira.xwiki.org/browse/XWIKI-21894

## Description

Use double quotes for `aria-hidden` attribute value. I.e.: `aria-hidden="true"`.

## Clarifications

The single quotes leads to error for example with tour extension:
```
xwiki:Tour.WebHome:17:78: ERROR - [JSC_PARSE_ERROR] Parse error. ',' expected
  17|       $('<a class="action actionLaunch"><span class="fa fa-play" aria-hidden='true'></span> Launch</a>')
                                                                                    ^
1 error(s), 0 warning(s)
```

# Executed Tests

`./mvnw -e install -f xwiki-platform-core/xwiki-platform-icon/pom.xml`

### DIFF
--- a/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
+++ b/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
@@ -39,8 +39,8 @@
   <content>## General settings
 xwiki.iconset.type = font
 xwiki.iconset.css = $services.webjars.url('org.webjars:font-awesome', 'css/font-awesome.min.css')
-xwiki.iconset.render.wiki = {{html clean="false"}}&lt;span class="fa fa-$icon" aria-hidden='true'&gt;&lt;/span&gt;{{/html}}
-xwiki.iconset.render.html = &lt;span class="fa fa-$icon" aria-hidden='true'&gt;&lt;/span&gt;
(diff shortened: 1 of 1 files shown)


surli:
@amottier Hi, thanks for the contribution, if it's a real bug could you open a ticket on jira.xwiki.org for it so that we can check for backporting it?

amottier:
Sure @surli I opened https://jira.xwiki.org/browse/XWIKI-21894

amottier:
@surli don't hesitate to let me know if I need to do anything else to get the PR merged :slightly_smiling_face:

surli:
> @surli don't hesitate to let me know if I need to do anything else to get the PR merged 🙂

Sorry I forgot to check it in last few days. It looks good to me. @Sereza7 apparently the  regression is related to a commit of yours in same file, wdyt?

Sereza7:
It' okay for me too :+1: 

Cause PR: https://github.com/xwiki/xwiki-platform/pull/2541#pullrequestreview-1701873141 
From what I remember, when merging this commit we had a couple regressions in CI due to poor escaping (that failed because the content became actually escape sensitive).
See https://github.com/xwiki/xwiki-platform/pull/2582 and https://github.com/xwiki/xwiki-platform/pull/2585 which contains some fixes.

I did forget to address the root inconsistency and I agree with these changes :+1: 
It would be interesting to see why it's not escaped before parsing here. Maybe we can open an improvement issue to avoid breaking this again later?

surli:
> It would be interesting to see why it's not escaped before parsing here. Maybe we can open an improvement issue to avoid breaking this again later?

probably yes. 

>  I agree with these changes 👍

OK I'll wait for today's release to be done and then I'll merge it.

PR #2897, opened 2024-02-15
XWIKI-21886: No contrast on Readable colorTheme message boxes

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21886
# Changes

## Description

* Updated the no colortheme initialization of variables to match more closely what is needed for box contrast
* Removed the override of the style itself. We just need to override the variable value.

## Clarifications

* There's no place in the codebase where I can see that those `state` colors are used, but better safe than sorry. This solution seems the safest to me.

# Screenshots & Video

The `Readable` colortheme looks okay after this PR:
21886-afterPR

# Executed Tests

Style only changes, very unlikely to break any tests. Tested them and got expected results on Readable, Iceberg, notheme and Darkly. Note that the Darkly color scheme does not fulfill accessibility color contrast requirement on boxes.
21886-afterPRDarkly
21886-afterPRIceberg
21886-afterPRnotheme

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 15.10.X , issue reported by LTS users.


michitux (commented):
I don't understand the diff. Why is this changing `messages.less` to set the border color instead of the background color? Further, why is this changing `@alert-danger-bg` instead of `@state-danger-bg` now?

Sereza7:
> I don't understand the diff. Why is this changing `messages.less` to set the border color instead of the background color? 

This was a leftover from a wrong merge conflict resolution. Fixed in https://github.com/xwiki/xwiki-platform/pull/2897/commits/a749e678928e3c8739532b265d3267eec684cda6

> Further, why is this changing `@alert-danger-bg` instead of `@state-danger-bg` now?

Because `@state-danger-bg` is never used as is (does not break anything except alerts to remove it), and `@alert-danger-bg` is the one used in bootstrap. This allows to not have to override the bootstrap style like what used to be done. We override the variable instead of the style.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/variablesInit.vm
```diff
@@ -265,9 +265,9 @@
   ## from an orangish yellow to complete brown. Instead, we change the text color to black.
   @btn-warning-color:                   #000;
   ## The boxes should have enough contrast by default, we make it lighter than bootstrap.
-  @state-danger-bg:                     #f8ecec;
-  @state-success-bg:                    #e5f3df;
-  @state-info-bg:                       #e1f1f9;
+  @alert-danger-bg:                     #f8ecec;
+  @alert-success-bg:                    #e5f3df;
+  @alert-info-bg:                       #e1f1f9;
```
michitux, line 270: I'm somehow missing "warning" in this list. According to the styles above, there should have been a similar `@state-warning-bg` variable. Was it missing before?

Sereza7: It wasn't a violating value from bootstrap, so there was no need to update it :)
Maybe I should add a comment here?

Sereza7: Added a comment in https://github.com/xwiki/xwiki-platform/pull/2897/commits/e1e311523b8bf28dc8bbacf365b0143aa418555c :+1:

PR #2920, opened 2024-02-22
XWIKI-9759: The javascript for the "Display annotation by default" feature gets cached too eagerly

manuelleduc:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-9759

# Changes

## Description

* Remove `parse=true` from `AnnotationCode.Script`, as a consequence:
  * All the translation are now loaded using a l10n Javascript module
  * the dependency to fast-diff is declared using a new `org.xwiki.platform.requirejs.module` UIX
  * the config is provided through a `org.xwiki.platform.template.header.after` UIX, providing the config in the form of serialized json in a `#annotation-config` script tag

# Executed Tests

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * stable-15.10.x
  * stable-16.1.x if merged early enough


xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -560,1365 +560,1365 @@ return XWiki;
...
+    'annotations.annotated.loading',
+    'annotations.annotated.loaderror',
+    'annotations.annotated.loaderror.wrongresponse',
+    'annotations.action.delete.confirm',
+    'annotations.action.delete.inProgress',
+    'annotations.action.delete.done',
+    'annotations.action.delete.failed',
+    'annotations.action.validate.success',
+    'annotations.action.validate.loaderror',
+    'annotations.action.edit.form.loaderror',
```
manuelleduc, line 582: Note: this key was introduced by https://github.com/xwiki/xwiki-platform/commit/3847fd1008ad648a9a1f8cf21b732afea0035c23 in 2010 and never had a corresponding translation key.
We need to define one.

xwiki-platform-core/xwiki-platform-annotation/xwiki-platform-annotation-ui/src/main/resources/AnnotationCode/Script.xml
```diff
@@ -560,70 +560,90 @@ return XWiki;
...
+    'annotations.action.create.selection.invalid',
+    'annotations.action.create.form.loaderror',
+    'annotations.action.create.success',
+    'annotations.action.create.loaderror'
+  ]
+})
 
-var XWiki = (function (XWiki) {
+require(['xwiki-l10n!xwiki-annotation-messages'], function(l10n) {
+  XWiki = (function (XWiki) {
```
mflorea, line 596: Why keep the anonymous function? Its purpose was to create a namespace so that we don't pollute the global namespace, but the require callback function does this now.

manuelleduc: But don't we still want to expose `XWiki.Annotation` publicly?

manuelleduc: ok, now I get what you mean, indeed I can drop a level of wrapping. Thanks!

PR #2925, opened 2024-02-23
XWIKI-21922: Introduce methods to fetch a subset of revisions in XWikiVersioningStoreInterface

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21922

# Changes

## Description

* Introduce two new methods to `XWikiVersioningStoreInterface`: `getXWikiDocVersions` and `getXWikiDocVersionsCount`, both taking an instance of `RevisionCriteria` as parameter. They both have a default implementation that work by using the existing `getXWikiDocVersions` and filtering the results.
* Introduce an implementations for both methods in `XWikiHibernateVersioningStore`.
* Add the methods `XWikiDocument.getRevisionsCount` and `Document.getRevisionsCount` to expose the new count method.
* Edit `XWikiDocument.getRevisions(RevisionCriteria)` to use the new fetch method. The existing filtering was used as a basis for the default implementation of `XWikiVersioningStoreInterface.getXWikiDocVersions`.
* Move hibernate queries from `XWikiHibernateVersioningStore` to a new `VersioningStoreQueryFactory` class to even out class fan-out complexity.

## Clarifications

* https://forum.xwiki.org/t/introduce-methods-to-fetch-a-subset-of-revisions-in-xwikiversioningstoreinterface/14078/8

# Screenshots & Video

N/A

# Executed Tests

A XWiki instance was run with the applied patches and everything was working as expected. A substantial speed-up was noticed in Velocity code using `XWikiDocument.getRevisions` for documents with more than 200k revisions.

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * stable-15.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -2650,7 +2650,7 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
      * @param criteria criteria used to match versions
      * @param context the XWiki context
      * @return the number of matching versions
-     * @since 15.10.7
+     * @since 15.10.8
      * @since 16.2.0RC1
      */
     @Unstable
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -52,6 +52,7 @@ import java.util.Vector;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -2643,6 +2644,21 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
         return getVersioningStore(context).getXWikiDocVersions(this, context);
     }
 
+    /**
+     * Counts the number of document versions matching criteria like author, minimum creation date, etc.
+     *
+     * @param criteria criteria used to match versions
+     * @param context the XWiki context
+     * @return the number of matching versions
+     * @since 15.10.7
+     * @since 16.2.0RC1
+     */
+    @Unstable
+    public long getRevisionsCount(RevisionCriteria criteria, XWikiContext context) throws XWikiException
+    {
+        return getVersioningStore(context).getXWikiDocVersionsCount(this, criteria, context);
+    }
+
     public String[] getRecentRevisions(int nb, XWikiContext context) throws XWikiException
     {
         try {
@@ -2668,52 +2684,15 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
     }
 
     /**
-     * Get document versions matching criterias like author, minimum creation date, etc.
+     * Gets document versions matching criteria like author, minimum creation date, etc.
      *
      * @param criteria criteria used to match versions
      * @return a list of matching versions
      */
     public List<String> getRevisions(RevisionCriteria criteria, XWikiContext context) throws XWikiException
     {
-        List<String> results = new ArrayList<String>();
-
-        Version[] revisions = getRevisions(context);
-
-        XWikiRCSNodeInfo nextNodeinfo = null;
-        XWikiRCSNodeInfo nodeinfo;
-        for (Version revision : revisions) {
-            nodeinfo = nextNodeinfo;
-            nextNodeinfo = getRevisionInfo(revision.toString(), context);
-
-            if (nodeinfo == null) {
-                continue;
-            }
-
-            // Minor/Major version matching
-            if (criteria.getIncludeMinorVersions() || !nextNodeinfo.isMinorEdit()) {
-                // Author matching
-                if (criteria.getAuthor().equals("") || criteria.getAuthor().equals(nodeinfo.getAuthor())) {
-                    // Date range matching
-                    Date versionDate = nodeinfo.getDate();
-                    if (versionDate.after(criteria.getMinDate()) && versionDate.before(criteria.getMaxDate())) {
-                        results.add(nodeinfo.getVersion().toString());
-                    }
-                }
-            }
-        }
-
-        nodeinfo = nextNodeinfo;
-        if (nodeinfo != null) {
-            if (criteria.getAuthor().equals("") || criteria.getAuthor().equals(nodeinfo.getAuthor())) {
-                // Date range matching
-                Date versionDate = nodeinfo.getDate();
-                if (versionDate.after(criteria.getMinDate()) && versionDate.before(criteria.getMaxDate())) {
-                    results.add(nodeinfo.getVersion().toString());
-                }
-            }
-        }
-
-        return criteria.getRange().subList(results);
+        return getVersioningStore(context).getXWikiDocVersions(this, criteria, context).stream()
+            .map(Version::toString).collect(Collectors.toCollection(ArrayList::new));
     }
 
     public XWikiRCSNodeInfo getRevisionInfo(String version, XWikiContext context) throws XWikiException
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
@@ -2691,8 +2691,10 @@ public class XWikiDocument implements DocumentModelBridge, Cloneable
      */
     public List<String> getRevisions(RevisionCriteria criteria, XWikiContext context) throws XWikiException
     {
+        // We collect explicitly into an ArrayList to ensure mutability of the resulting list.
         return getVersioningStore(context).getXWikiDocVersions(this, criteria, context).stream()
-            .map(Version::toString).collect(Collectors.toCollection(ArrayList::new));
+            .map(Version::toString)
+            .collect(Collectors.toCollection(ArrayList::new));
     }
 
     public XWikiRCSNodeInfo getRevisionInfo(String version, XWikiContext context) throws XWikiException
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
@@ -39,6 +39,7 @@ import com.xpn.xwiki.XWiki;
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.criteria.impl.RevisionCriteria;
+import com.xpn.xwiki.criteria.impl.RevisionCriteriaFactory;
 import com.xpn.xwiki.doc.XWikiAttachment;
 import com.xpn.xwiki.doc.XWikiDocument;
 import com.xpn.xwiki.doc.XWikiDocumentArchive;
@@ -108,7 +109,8 @@ public class XWikiHibernateVersioningStore extends XWikiHibernateBaseStore imple
     @Override
     public Version[] getXWikiDocVersions(XWikiDocument doc, XWikiContext context) throws XWikiException
     {
-        return getXWikiDocVersions(doc, new RevisionCriteria(null, null, null, true), context).toArray(new Version[0]);
+        return getXWikiDocVersions(doc, new RevisionCriteriaFactory().createRevisionCriteria(true), context)
+            .toArray(new Version[0]);
     }
 
     @Override
@@ -138,7 +140,7 @@ public class XWikiHibernateVersioningStore extends XWikiHibernateBaseStore imple
     public XWikiDocumentArchive getXWikiDocumentArchive(XWikiDocument doc, XWikiContext inputxcontext)
         throws XWikiException
     {
-        return getXWikiDocumentArchive(doc, new RevisionCriteria(null, null, null, true), inputxcontext);
+        return getXWikiDocumentArchive(doc, new RevisionCriteriaFactory().createRevisionCriteria(true), inputxcontext);
     }
 
     private XWikiDocumentArchive getXWikiDocumentArchive(XWikiDocument doc, RevisionCriteria criteria,
@@ -172,7 +174,7 @@ public class XWikiHibernateVersioningStore extends XWikiHibernateBaseStore imple
     public void loadXWikiDocArchive(XWikiDocumentArchive archivedoc, boolean bTransaction, XWikiContext context)
         throws XWikiException
     {
-        loadXWikiDocArchive(archivedoc, new RevisionCriteria(null, null, null, true), context);
+        loadXWikiDocArchive(archivedoc, new RevisionCriteriaFactory().createRevisionCriteria(true), context);
     }
 
     private void loadXWikiDocArchive(XWikiDocumentArchive archivedoc, RevisionCriteria criteria, XWikiContext context)
@@ -312,7 +314,7 @@ public class XWikiHibernateVersioningStore extends XWikiHibernateBaseStore imple
     protected List<XWikiRCSNodeInfo> loadAllRCSNodeInfo(XWikiContext context, final long id, boolean bTransaction)
         throws XWikiException
     {
-        return loadRCSNodeInfo(context, id, new RevisionCriteria(null, null, null, true));
+        return loadRCSNodeInfo(context, id, new RevisionCriteriaFactory().createRevisionCriteria(true));
     }
 
     /**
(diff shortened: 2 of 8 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -2668,52 +2684,15 @@ public String[] getRecentRevisions(int nb, XWikiContext context) throws XWikiExc
...
-                Date versionDate = nodeinfo.getDate();
-                if (versionDate.after(criteria.getMinDate()) && versionDate.before(criteria.getMaxDate())) {
-                    results.add(nodeinfo.getVersion().toString());
-                }
-            }
-        }
-
-        return criteria.getRange().subList(results);
+        return getVersioningStore(context).getXWikiDocVersions(this, criteria, context).stream()
+            .map(Version::toString).collect(Collectors.toCollection(ArrayList::new));
```
manuelleduc, line 2695: The formatting is a matter of taste, but using `Collectors.toList()` to convert the steam to a list if more standard.
```suggestion
        return getVersioningStore(context)
            .getXWikiDocVersions(this, criteria, context)
            .stream()
            .map(Version::toString)
            .collect(Collectors.toList());
```

vmassol: Actually sonarqube will complain if you don't use it ;)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
```diff
@@ -351,11 +372,39 @@ public XWikiRCSNodeContent loadRCSNodeContent(final XWikiRCSNodeId id, boolean b
...
+     */
+    protected long getRCSNodeInfoCount(XWikiContext context, final long id, RevisionCriteria criteria)
+        throws XWikiException
+    {
+        return executeRead(context, session -> {
+            try {
+                return VersioningStoreQueryFactory.getRCSNodeInfoCountQuery(session, id, criteria).getSingleResult();
+            } catch (IllegalArgumentException e) {
+                // This happens when the database has wrong values...
+                LOGGER.error("Encountered invalid history when computing archive size for document [{}]", id, e);
```
manuelleduc, line 404: I'm not sure if we want to capture and log this exception. Shouldn't we re-throw it in a `XWikiException`?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
```diff
@@ -351,11 +372,39 @@ public XWikiRCSNodeContent loadRCSNodeContent(final XWikiRCSNodeId id, boolean b
...
+
+                return Long.valueOf(0);
```
manuelleduc, line 406:
```suggestion
                return 0L;
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
pjeanjean: I thought about that, but using `toList()` does not guarantee the mutability of the returned list. Having an immutable list here would affect some current usages of this method (e.g., https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm#L53)

manuelleduc: This is fine but in this case you need to make it clear with some doc ;)

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
vmassol: The comment is more cryptic than the error message below it! If it doesn't bring anything more then it should be removed IMO

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
```diff
@@ -111,18 +108,23 @@ public XWikiHibernateVersioningStore()
...
+    @Override
+    public Collection<Version> getXWikiDocVersions(XWikiDocument doc, RevisionCriteria criteria, XWikiContext context)
+        throws XWikiException
     {
         try {
-            XWikiDocumentArchive archive = getXWikiDocumentArchive(doc, context);
+            XWikiDocumentArchive archive = getXWikiDocumentArchive(doc, criteria, context);
             if (archive == null) {
-                return new Version[0];
+                return Collections.emptyList();
```
manuelleduc, line 122: Slight shorted version based on Java standard APIs.
```suggestion
                return List.of();
```

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiVersioningStoreInterface.java
```diff
@@ -47,6 +55,79 @@ void saveXWikiDocArchive(XWikiDocumentArchive archivedoc, boolean bTransaction,
...
+     *
+     * @param doc the document
+     * @param criteria criteria used to match versions
+     * @param context the XWiki context
+     * @return the matching versions
+     * @since 15.10.7
+     * @since 16.2.0RC1
+     */
+    @Unstable
+    default Collection<Version> getXWikiDocVersions(XWikiDocument doc, RevisionCriteria criteria, XWikiContext context)
```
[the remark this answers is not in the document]
manuelleduc: Also, wouldn't it be interesting to refactor a little this method to call `stream().count()` [here](https://github.com/xwiki/xwiki-platform/pull/2925/files#diff-7675da62f39751ae6a949342a3ef04af1d629d3c9bcc39313fb955dd1093ad91R111), allowing to avoid aggregating the steam to a list when we just want a count (e.g., [here](https://github.com/xwiki/xwiki-platform/pull/2925/files#diff-7675da62f39751ae6a949342a3ef04af1d629d3c9bcc39313fb955dd1093ad91R128)).
It's a small perf improvement, but since we are focusing on versions perfs here.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
pjeanjean: I agree, but I was concerned about https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java#L320 (which is why I copied the comment). Now, this comment has been introduced in https://github.com/xwiki/xwiki-platform/commit/725f4594285667b7dc777a874156d43b5458f1b8 and the code has been refactored a lot since then so it might no longer be relevant (but I can't be sure).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
```diff
@@ -2668,52 +2684,17 @@ public String[] getRecentRevisions(int nb, XWikiContext context) throws XWikiExc
...
-                    results.add(nodeinfo.getVersion().toString());
-                }
-            }
-        }
-
-        return criteria.getRange().subList(results);
+        // We collect explicitly into an ArrayList to ensure mutability of the resulting list.
+        return getVersioningStore(context).getXWikiDocVersions(this, criteria, context).stream()
+            .map(Version::toString)
+            .collect(Collectors.toCollection(ArrayList::new));
```
surli, line 2697: There's a Collectors.toList() AFAIK.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
```diff
@@ -111,18 +107,23 @@ public XWikiHibernateVersioningStore()
 
     @Override
     public Version[] getXWikiDocVersions(XWikiDocument doc, XWikiContext context) throws XWikiException
+    {
+        return getXWikiDocVersions(doc, new RevisionCriteria(null, null, null, true), context).toArray(new Version[0]);
```
surli, line 111: Feels like it might almost worth having an empty constructor for `RevisionCriteria` which would fallback with this parameters no? I'm seeing that you use same signature below.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
```diff
@@ -288,39 +301,44 @@ public void updateXWikiDocArchive(XWikiDocument doc, boolean bTransaction, XWiki
...
+    /**
+     * Loads a part of the RCS nodes present in the archive of a given document, based on criteria.
+     *
+     * @param context the XWiki context
+     * @param id {@link XWikiRCSNodeContent#getId()}
+     * @param criteria the criteria matching the nodes to load
+     * @return loaded RCS nodes content
+     * @throws XWikiException if any error
+     */
+    protected List<XWikiRCSNodeInfo> loadRCSNodeInfo(XWikiContext context, final long id, RevisionCriteria criteria)
```
surli, line 327: Missing since

surli: Not sure if we put unstable on protected methods of public APIs ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/hibernate/query/VersioningStoreQueryFactory.java
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Helper class to build Hibernate queries for a VersioningStore.
+ *
+ * @param <T> the type of element returned by the query
+ * @version $Id$
+ * @since 15.10.7
+ * @since 16.2.0RC1
+ */
+public final class VersioningStoreQueryFactory<T>
```
surli, line 47: Missing unstable

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
pjeanjean: There is actually a `RevisionCriteriaFactory`, I could add a method over there.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiVersioningStoreInterface.java
```diff
@@ -47,6 +55,74 @@ void saveXWikiDocArchive(XWikiDocumentArchive archivedoc, boolean bTransaction,
...
+        throws XWikiException
+    {
+        return getXWikiDocStringVersions(doc, criteria, context).stream()
```
surli, line 104: Honestly not sure I'm a big fan of having a whole implemented method in the interface just for a fallback that will be very unlikely used: I mean we provide a default just in case someone would have maybe re-implemented this interface. Which might be very very rare. So I would only return an empty collection here. And same below I'd only return 0. 
Providing an implem here means that we'll need to maintain it on the long run, while we don't really want to: it should be maintained in the actual implem, not in the interface IMO.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocument.java
surli: Indeed my bad I haven't read the comment above. Now there's a reason for wanting to preserve mutability here?

pjeanjean: This instruction would not work if the list was immutable: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/historyinline.vm#L53
Now, this could be rewritten, but it might not be the only one to expect a mutable output.

surli: Ok so indeed it's probably safer to keep it like that.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiVersioningStoreInterface.java
surli: Check the warning in https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility it says:
> Note that the default method should not throw an exception in general (like throw new UnsupportedOperationException("Not supported");), since that would mean that the code calling the default method will fail. As such, it cannot be considered as backward compatibility.

pjeanjean: Indeed. I'm still unsure on what to do then, I understand your point but at the same time I'm not a fan of returning empty values.

tmortagne: I prefer the current version (real retro compatibility). The only alternative if to not provide any default implementation so that this breakage is documented in the release note (but I'm really not a fan since we have a retro compatible solution).

michitux: :+1: for the real retro compatibility. What we should make sure is that it is covered by a unit test (I haven't checked if it is) to ensure that it is actually working as this is code that will never be executed in any real use and it would be quite pointless if it should turn out that it doesn't actually work as it was never tested.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
tmortagne: Anything no private is a public API so yes missing `@Unstable` too.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiHibernateVersioningStore.java
```diff
@@ -351,11 +369,37 @@ public XWikiRCSNodeContent loadRCSNodeContent(final XWikiRCSNodeId id, boolean b
...
+    /**
+     * Counts the number of RCS nodes present in the archive of a given document, based on criteria.
+     *
+     * @param context the XWiki context
+     * @param id {@link XWikiRCSNodeContent#getId()}
+     * @param criteria the criteria matching the nodes to count
+     * @return the number of matching RCS nodes
+     * @throws XWikiException if any error
+     */
+    protected long getRCSNodeInfoCount(XWikiContext context, final long id, RevisionCriteria criteria)
```
tmortagne, line 393: Same question as above: does it really need to be protected ?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiVersioningStoreInterface.java
```diff
@@ -47,6 +55,74 @@ void saveXWikiDocArchive(XWikiDocumentArchive archivedoc, boolean bTransaction,
...
+            .map(Version::new)
+            .toList();
```
tmortagne, line 106: Be careful as this is not going to work in 15.10.x (Java 11), I generally try to use the same code for both branches for code which is intended to be cherry-picked (make future bugfixing easier).

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/XWikiVersioningStoreInterface.java
pjeanjean: There are still no tests, as I wanted feedback on the implementation first. I'll add them tomorrow.

PR #2926, opened 2024-02-23
XWIKI-9046: Renaming a document holding a lot of revisions can lead to an OutOfMemory exception

pjeanjean:
This requires merging https://github.com/xwiki/xwiki-platform/pull/2925 first.

# Jira URL

https://jira.xwiki.org/browse/XWIKI-9046

# Changes

## Description

* Only load the last node when updating the document archive instead of the whole history.

## Clarifications

N/A

# Screenshots & Video

N/A

# Executed Tests

Renaming a document with around 200k revisions was tested on an instance running this patch and it worked fine. Also, the saving process only took around 150ms instead of 1s without the patch applied (with HSQLDB).

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * stable-15.10.x


tmortagne:
> Renaming a document with around 200k revisions

Might be interesting to have such an integration test, a lot of pieces are involved here, and it would be easy to break this use case again.

pjeanjean:
I added a very basic test for now.

tmortagne (approved):

PR #2929, opened 2024-02-26
XWIKI-21809: ClassCastException in DefaultEventConverterManager on a cluster environment

surli:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21809

# Changes

## Description

Some task output is Optional and we never want the Optional in the notification data: we do want what the Optional contains.
This PR provides a test to expose the bug and a naive fix.

## Clarifications

I've not been able to provide a better fix as it would involve breaking APIs to not return Optional. On the long run I'm not sure it was really necessary to have Optional for the APIs here: it's basically only used to support backward compatibility with the legacy store AFAICS, so the Optional is never empty with the new store... 

# Executed Tests

`mvn clean install -Pquality` on `xwiki-platform-eventstream`

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches:
  * 15.10.x
  * 14.10.x (?)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/AbstractAsynchronousEventStore.java
+++ b/xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/AbstractAsynchronousEventStore.java
@@ -439,21 +439,25 @@ public abstract class AbstractAsynchronousEventStore implements EventStore, Init
         task.future.complete(output);
 
         // Notify event listeners
+        Object notificationOuput = task.output;
+        if (task.output instanceof Optional<?>) {
+            notificationOuput = ((Optional<?>) task.output).orElse(null);
+        }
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-eventstream/xwiki-platform-eventstream-api/src/main/java/org/xwiki/eventstream/internal/AbstractAsynchronousEventStore.java
```diff
@@ -439,37 +439,41 @@ private <O, I> void complete(EventStoreTask<O, I> task, O output)
         task.future.complete(output);
 
         // Notify event listeners
+        Object notificationOuput = task.output;
+        if (task.output instanceof Optional<?>) {
+            notificationOuput = ((Optional<?>) task.output).orElse(null);
```
tmortagne, line 444: I'm not sure it's enough.

We should probably not send any notification if the optional is not present (meaning nothing was in fact deleted).

tmortagne (approved):

PR #2940, opened 2024-02-27
XWIKI-21879: $xwiki.getURLContent() does not forward credentials

lrlunin:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21879
# Changes

## Description

* Added `setAuthenticationPreemptive(true)` to `HttpClient` parameters. Otherwise the given credentials for basic auth were ignored.

## Clarifications

* I guess, there is no additional explanations needed since the PR is 2 lines long.

# Screenshots & Video

# Executed Tests

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  *

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
@@ -6679,6 +6679,9 @@ public class XWiki implements EventListener
         throws IOException
     {
         HttpClient client = getHttpClient(timeout, userAgent);
+        
+        // otherwise there is no authentification credentials will be sent
+        client.getParams().setAuthenticationPreemptive(true);
 
         // pass our credentials to HttpClient, they will only be used for
         // authenticating to servers with realm "realm", to authenticate agains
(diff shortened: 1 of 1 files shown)


lrlunin:
Any Updates?

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -6679,6 +6679,9 @@ public String getURLContent(String surl, String username, String password, int t
         throws IOException
     {
         HttpClient client = getHttpClient(timeout, userAgent);
+        
+        // otherwise there is no authentification credentials will be sent
+        client.getParams().setAuthenticationPreemptive(true);
```
tmortagne, line 6684: This change might cause some problem to some use case. I assume forcing `Basic` (which is what you are doing here) could cause a problem when, for example, the endpoint actually requires something else like `Digest` (which seems supported by HttpClient 3 too and should work with the current code).

What might be safer to to introduce new #getURLContent API with a parameter to force Basic auth.

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/XWiki.java
```diff
@@ -6679,6 +6679,9 @@ public String getURLContent(String surl, String username, String password, int t
         throws IOException
     {
         HttpClient client = getHttpClient(timeout, userAgent);
+        
+        // otherwise there is no authentification credentials will be sent
```
tmortagne, line 6683: This is not fully accurate. By default, HttpClient only send credentials when asked for it by the server (using [a 401 response|https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401])

PR #2959, opened 2024-03-05
XWIKI-21972: FilterList shouldn't force using a contains operator in LiveData

surli:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21972

# Changes

## Description

* Fallback on  the default operator in FilterList instead of forcing using contains.
* Provide default operator for all base filters
* Prevent LiveTable source to override operators for the filter list defined in a property descriptor: we only want to override the operator in the specific case of a StringList for which we only equals and empty

## Clarifications

* It's possible that a filter descriptor defines another default operator than contains (e.g. equals) and that the custom LD entry store expects the operator to never be contains: it's quite surprising to always obtain a contains whenever it's not even described in the filter descriptor... 

# Executed Tests

Ran `mvn clean install -Pquality,integration-tests,docker` on xwiki-platform-livedata

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 15.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
@@ -14,7 +14,8 @@
     "filters": [
       {
         "id": "text",
-        "operators": ["contains", "startsWith", "equals"]
+        "operators": ["contains", "startsWith", "equals"],
+        "defaultOperator": "contains"
       },
       {
         "id": "number",
@@ -22,20 +23,24 @@
           {"id": "equals", "name": "="},
           {"id": "less", "name": "<"},
           {"id": "greater", "name": ">"}
-        ]
+        ],
+        "defaultOperator": "equals"
       },
       {
         "id": "boolean",
-        "operators": ["equals"]
+        "operators": ["equals"],
+        "defaultOperator": "equals"
       },
       {
         "id": "date",
         "operators": ["between", "before", "after", "contains"],
-        "dateFormat": "yyyy/MM/dd HH:mm"
+        "dateFormat": "yyyy/MM/dd HH:mm",
+        "defaultOperator": "between"
       },
       {
         "id": "list",
-        "operators": ["equals", "startsWith", "contains", "empty"]
+        "operators": ["equals", "startsWith", "contains", "empty"],
+        "defaultOperator": "contains"
       }
     ],
 
--- a/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataPropertyStore.java
+++ b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataPropertyStore.java
@@ -71,6 +71,8 @@ import com.xpn.xwiki.objects.classes.PropertyClass;
 @InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
 public class LiveTableLiveDataPropertyStore extends WithParameters implements LiveDataPropertyDescriptorStore
 {
+    private static final String EQUALS_OPERATOR = "equals";
+
     @Inject
     @Named("current")
     private DocumentReferenceResolver<String> currentDocumentReferenceResolver;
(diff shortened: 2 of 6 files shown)


xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
```diff
@@ -14,28 +14,33 @@
...
         "operators": ["between", "before", "after", "contains"],
-        "dateFormat": "yyyy/MM/dd HH:mm"
+        "dateFormat": "yyyy/MM/dd HH:mm",
+        "defaultOperator": "between"
       },
       {
         "id": "list",
-        "operators": ["equals", "startsWith", "contains", "empty"]
+        "operators": ["equals", "startsWith", "contains", "empty"],
+        "defaultOperator": "contains"
```
surli, line 43: Note that contrarily to the other, it's not the first operator, but it matches the old behaviour of always using `contains` in the FilterList code. Changing the default to equals would be a breaking change.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
```diff
@@ -93,7 +93,11 @@ export default {
             this.removeFilter();
           } else if (value !== this.value) {
             // If the selected value has an empty value, then use the empty filter, otherwise use the contains filter.
-            this.applyFilter(value, value === '' ? 'empty' : 'contains');
+            if (value === '') {
+              this.applyFilter(value, 'empty');
```
surli, line 97: Note that here we assume there will always be an empty operator in any LD source for the list type. We need to document this.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
```diff
@@ -93,7 +93,11 @@ export default {
             this.removeFilter();
           } else if (value !== this.value) {
             // If the selected value has an empty value, then use the empty filter, otherwise use the contains filter.
```
mflorea, line 95: This comment needs to be updated.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/filters/FilterList.vue
```diff
@@ -93,7 +93,11 @@ export default {
...
-            this.applyFilter(value, value === '' ? 'empty' : 'contains');
+            if (value === '') {
+              this.applyFilter(value, 'empty');
+            } else {
+              this.applyFilter(value);
```
mflorea, line 99: I'd mention in a comment that here we're applying the **default** operator.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataPropertyStore.java
```diff
@@ -155,6 +154,7 @@ private LiveDataPropertyDescriptor getLiveDataPropertyDescriptor(PropertyClass x
                 // The default live table results page currently supports only exact matching for list properties with
                 // multiple selection and no relational storage (selected values are stored concatenated on a single
                 // database column).
+                descriptor.getFilter().addOperator("empty", null);
```
mflorea, line 157: I take it that the default ``LiveTableResults`` page supports empty list filter only for ``StringListProperty``? What operators are available for ``DBStringListProperty`` (relational storage)?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/resources/liveDataConfiguration.json
surli: I agree, now problem is that JSON doesn't allow to put comments...

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataPropertyStore.java
surli: > I take it that the default LiveTableResults page supports empty list filter only for StringListProperty?
No, that's not really what this change means. 

Previously to my change the `empty` operator was added to any descriptor with type list, and when we're getting the default operator we're checking first the operators defined in the descriptor: *this list is not merged*. Which means that in case of a list which wasn't a StringList we were always getting only `empty`... 

Now for a  `StringListProperty` we do want to override the list of operators to only use equals because others are not supported (according to the comment), but we still also need to support empty, so I just moved it there. 
The result is that for all list properties there no operator defined in the descriptor: we entirely rely on the type filters which are defining all the operators (including empty). And for `StringListProperty` we rely on the descriptor which defines the two operators `equals` and `empty`.

Now when writing this I'm still wondering if there's not a hidden bug here with `StringListProperty` because 1. the operators defined in the descriptor are only used in the column filter, not in the custom filters from the panels (which are always coming directly from the type) 2. I have a doubt on the code of applyFilter here: I think it would default on the descriptor operator first, but we probably need to also define the default operator in the descriptor, to ensure it doesn't fallback on empty operator, but on equals. @manuelleduc wdyt?

PR #2970, opened 2024-03-07
XWIKI-21824: The enter key in any field does not validate the macro editor

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21824
# Changes

## Description

* Changed the nature of the macro edit modal to be based on a form (forms can use the native implicit form submission on inputs).

## Clarifications

* First step is to replace a `div` node with a `form`
* Then we remove the `type="button"` fromt he submit button in the modal footer
* Then we move the listener on the submit button onclick event to the form onsubmit event.
* At last, we limit propagation of the event. We want to keep the same behaviour as before and we need to make sure none of the native behaviours would not happen.
* In this implementation, I tried to rely as much as possible on native javascript to avoid increasing our dependency on prototype or jquery too much. 

# Screenshots & Video

[Video demo](https://youtu.be/0zcucVvR650)

# Executed Tests

From what I could see in [the MacroDialogEditModal pageobject](https://github.com/xwiki/xwiki-platform/blob/9aeae7bdf5d40648ec675a40d182a1853081b837/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-test/xwiki-platform-ckeditor-test-pageobjects/src/main/java/org/xwiki/ckeditor/test/po/MacroDialogEditModal.java#L43), no selector is broken.
After building ckeditor-plugins and ckeditor-webjar, I tested them manually on my local instance (see video demo above). It behaved according to expectations and did not seem to create any regression. 
Docker tests on xwiki-platform-core/xwiki-platform-ckeditor/ failed on `AllIT » NoClassDefFound Failed to get fields for class [org.xwiki.extension.script.AbstractExtensionScriptService] because the class [org/xwiki/script/safe/ScriptSafeProvider] couldn't be found in the ClassLoader.` . Everything before that in `mvn clean install -f xwiki-platform-core/xwiki-platform-ckeditor/ -Pdocker,integration-tests,quality ` passed successfully.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No (low test coverage, DOM changes that might break some extensions)


manuelleduc:
`ckeditor-webjar` does not need to be rebuild on recent XWiki versions (i.e., all but 14.10.x)
The `ScriptSafeProvider` issue is temporary and should be fixed by now.

manuelleduc:
@Sereza7 From your comments, it feels like all you do is a consequence of limitations in [modal.js](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js)
Wouldn't a better approach be to improve modal.js, and use the improvement in the macros plugin?

Sereza7:
TLDR: Improving modal.js would be too dangerous.

This is a pretty big change in architecture, and I fear changing modal.js could break a lot of things. From what I saw, it could make sense to include the modal architecture in our frontend API. :)
I think the safest solution is to limit the changes to what was reported in the ticket. The scope of the changes proposed here are much more controlled: we already know what kind of nodes are inside of a macro editor dialog, on the opposite of a generic modal, which can have pretty much anything and everything.

An example of dangerous regression that might happen is that any button that does not have `type="button"` would submit the form and close the modal...

I'm pretty sure there shouldn't be too much of those when editing macro fields, but I'm almost certain there's some at least one modal that would break (if not in xwiki-platform, one from an extension or some created in user custom code).

Updating every modal could be an improvement we make later on, but IMO we'd need to vote it on the forum as an API breakage. If we decide to apply this PR as it is right now, it would give some facts to support or go against the possibility of updating modal.js later on.

The scope of the changes in this PR are pretty limited (15 lines in 2 blocks in 1 file), so it would be straightforward to revert them if we decide to change this in the modal API.

manuelleduc:
> This is a pretty big change in architecture, and I fear changing modal.js could break a lot of things. From what I saw, it could make sense to include the modal architecture in our frontend API. :)

I'm not suggesting to change it all the time. But wouldn't it be possible to provide a parameter "form: true" that would only change the DOM conditionally?

Sereza7:
In https://github.com/xwiki/xwiki-platform/pull/2970/commits/c5d975acfcac4960cab1221355ade347b2e0d14b I moved the new feature from the macroEditor to the modal definition, so it can be easily reused. This addressed Manuel's comment at https://github.com/xwiki/xwiki-platform/pull/2970#issuecomment-1985673830.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroEditor.js
```diff
@@ -723,7 +724,10 @@ define(
           load.call(this, input, macroEditor);
         }
       });
-      submitButton.on('click', function(event) {
+      modal.find('.modal-content')[0].addEventListener('submit', function(event) {
```
surli, line 727: why not using jQuery style here? `.on('submit', callback)`?

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js
```diff
@@ -58,8 +59,20 @@ define('modal', ['jquery', 'l10n!modal', 'bootstrap'], function($, translations)
     });
     modal.find('.modal-title').text(definition.title);
     modal.find('.modal-body').html('').append(definition.content);
-    modal.find('.modal-footer .btn-primary').text(definition.acceptLabel);
+    const submitButton = modal.find('.modal-footer .btn-primary');
+    submitButton.text(definition.acceptLabel);
     modal.find('.modal-footer .btn[data-dismiss="modal"]').text(definition.dismissLabel);
+    if(definition.form) {
```
mflorea, line 65:
```suggestion
    if (definition.form) {
```

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js
```diff
@@ -58,8 +59,20 @@ define('modal', ['jquery', 'l10n!modal', 'bootstrap'], function($, translations)
...
+      // We want the submit button to actually submit the form
+      submitButton.removeAttr('type');
+      // This modal should act and look like a form. We replace its content node by a form, so we can benefit from
+      // native form utilities, such as implicit form validation (pressing Enter on a text field)
+      const modalContent= modal.find('.modal-content');
+      const modalContentInnerHTML = modalContent.html();
+      const formModalContent = $('<form class="modal-content">');
+      formModalContent.html(modalContentInnerHTML);
+      modalContent.replaceWith(formModalContent);
```
mflorea, line 74:
```suggestion
      const formModalContent = $('<form class="modal-content">');
      formModalContent.append(modalContent.contents());
      modalContent.replaceWith(formModalContent);
```

It's safer and faster to move the existing nodes rather than using ``innerHTML``.

mflorea (changes requested):

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroEditor.js
manuelleduc: While I agree moving to vanillajs could be nice, it needs to be discussed on a proposal first.
For now, I'm +1 to keep using jQuery for consitency.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2970/commits/81baeca8ac3a25b91ccd020aa807399c5e8d7a25 :+1:

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroEditor.js
```diff
@@ -723,7 +724,10 @@ define(
           load.call(this, input, macroEditor);
         }
       });
-      submitButton.on('click', function(event) {
+      modal.find('.modal-content').on('submit', function(event) {
```
michitux, line 727: I wonder if it wouldn't make sense for the modal to explicitly expose a form (or `formElement`) attribute to easily get the form without relying on, e.g., the fact that the form has the `model-content` class.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js
```diff
@@ -58,8 +59,19 @@ define('modal', ['jquery', 'l10n!modal', 'bootstrap'], function($, translations)
...
     modal.find('.modal-footer .btn[data-dismiss="modal"]').text(definition.dismissLabel);
+    if (definition.form) {
+      // We want the submit button to actually submit the form
+      submitButton.removeAttr('type');
+      // This modal should act and look like a form. We replace its content node by a form, so we can benefit from
+      // native form utilities, such as implicit form validation (pressing Enter on a text field)
+      const modalContent= modal.find('.modal-content');
+      const formModalContent = $('<form class="modal-content">');
+      formModalContent.append(modalContent.contents());
+      modalContent.replaceWith(formModalContent);
```
michitux, line 73: Are we sure that there is no code targeting `div.modal-content`? I'm just asking here, it's probably fine, this is just the main place here where I could see problems with backwards compatibility.

michitux (approved):
Looks good to me, just two comments about possible improvements that aren't mandatory in any way.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js
```diff
@@ -49,7 +49,8 @@ define('modal', ['jquery', 'l10n!modal', 'bootstrap'], function($, translations)
       title: '',
       content: '',
       acceptLabel: translations.get('ok'),
-      dismissLabel: translations.get('cancel')
+      dismissLabel: translations.get('cancel'),
+      form: false
```
manuelleduc, line 53: Some doc of what `form` does, and when it should be used, here and in xwiki.org would be nice.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-macro/macroEditor.js
Sereza7: If the `form` parameter is set to false (default), this element would not be a form but just a regular container for content, so the `formElement` name would not be appropriate. 
We can propose to expose a `modalContent`element as an improvement in another ticket, but from what I can see on https://github.com/search?q=org%3Axwiki%20.modal-content&type=code I doubt this would be very useful.

Let me know what you think about this. I can handle creating a small improvement ticket if you think it's worth it :)

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js
Sereza7: From what I can see on https://github.com/search?q=org%3Axwiki%20.modal-content&type=code this is okay :+1: 

It could be interesting to find out when / why we target `div` . `div` is not semantic, so IMO it would be good for it to be replaceable without worrying about breaking anything but also pointless to put it in a CSS selector or an xpath (if that's needed, it means that there's not enough semantics in the first place :) ).

Unfortunately there's [plenty of examples](https://github.com/search?q=repo%3Axwiki%2Fxwiki-platform%20%22div.%22&type=code) of this in the codebase.

xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-dialog/modal.js
Sereza7: Added a comment about it in https://github.com/xwiki/xwiki-platform/pull/2970/commits/90cd576e17180c0707650aaccfd087d274daba85 .
I'll provide some documentation and a small release note for devs when we merge the PR :ok_hand:

PR #2971, opened 2024-03-07
XWIKI-21848: Migrate NotificationFilterPreferenceLivetableResults to a Live Data source

surli:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21848

# Changes

## Description

* Provide 2 different LD custom source: one for Custom Filter Preferences and one for System Filter Preferences: the reason for not going to a single LD source is that those LT are now very different and we use to hack the data source to provide different information. Having 2 different sources make things very more clear
* Provide ordering / filtering on the Custom Filter source as it's the one with most entries
* Don't provide any kind of ordering / filtering for System filter preferences: in XS we only have 6 items, and it's rare to have custom filters
* Provide some new APIs where needed

WIP: 
  - [x] need to fix enabling / disabling system filters
  - [x] Really remove the old NotificationFilterPreferenceLivetableResults
  - [x] Fix doc
  - [x] Fix checkstyle
  - [x] Fix bug with boolean filter displaying translation keys
  - [x] Fix coverage
  - [x] Deprecate some translations
  - [x] Check tests
  - [x] Run integration tests with other DBs
  - [x] Manual tests on subwiki and local users in subwiki
  - [x] Scenario to test: add custom filter for an event type coming from an extension and then uninstall the extension
  - [x] Provide integration tests
  - [x] Fix since
  - [x] Check cards layout
  - [x] Ensure to have a default order for custom filters (sort by ID)

## Clarifications

*

# Screenshots & Video

System filters:
image

Custom filters:
image

# Executed Tests

   * All integration tests of xwiki-platform-notification-test-docker also run with database mysql
   * Ran manual checks of the live data in main wiki and subwiki with global and local users and also with wikis administration
   * Played with filtering / sorting / enabling / disabling / deleting / creating 
   * Tried combination of filtering and pagination (> 300 filters)

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None


xwiki-platform-core/xwiki-platform-notifications/xwiki-platform-notifications-filters/xwiki-platform-notifications-filters-api/src/main/java/org/xwiki/notifications/filters/internal/DefaultNotificationFilterManager.java
```diff
@@ -207,6 +214,29 @@ public Stream<NotificationFilter> getToggleableFilters(Collection<NotificationFi
...
+
+    @Override
+    public List<ToggleableNotificationFilter> getToggleableFilters(WikiReference wikiReference)
+        throws NotificationException
+    {
+        List<ToggleableNotificationFilter> result = new ArrayList<>();
+        getAllFilters(wikiReference)
+            .stream().filter(filter -> filter instanceof ToggleableNotificationFilter)
+            .forEach(item -> result.add((ToggleableNotificationFilter) item));
+        return result;
```
manuelleduc: Slightly more idiomatic (can be applied on the next method as well).
```suggestion
        DocumentReference userDoc = this.documentReferenceUserReferenceSerializer.serialize(userReference);
        return getAllFilters(userDoc, false)
            .stream()
            .filter(filter -> filter instanceof ToggleableNotificationFilter)
            .map(filter -> (ToggleableNotificationFilter) filter)
            .toList();
```

surli: Thanks I missed your comment but I actually applied same solution in new version :)

PR #2981, opened 2024-03-13
XWIKI-16216: Allow users to rebind the shortcuts from their user profile UI

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-16216

# Changes

## Description

* Added user preferences for most shortcuts
* Added tests for this new feature
* Only display view shortcuts in the preferences when the user can actually use them (usertype = advanced)

## Clarifications

* Some shortcuts are `hard-coded`, making it impossible to change them easily. They aren't supported in our documentation about changing shortcuts, so I made the choice to not include them in this solution either. They can be added later as an improvement :)
* Tests don't check every single shortcut, but only the most important use cases on a couple of them.
* For now, only a simple shortcut syntax is supported. I don't think it's essential to provide a way to input key combo shortcuts (such as x+x+x+a). However this could be an improvement implemented later on.
* If this PR is merged as is, I'll create tickets for both improvements mentionned above.

# Screenshots & Video

https://github.com/xwiki/xwiki-platform/assets/28761965/b10b12e6-9d9d-4abe-9509-41a5c8669508

Side note: this video was taken before the last commit that fixed the title attribute on edit action buttons to match the shortcut value whatever the situation.

# Executed Tests

Extensive manual tests (see video above).
Passed `mvn clean install -f xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/ -Pquality,integration-tests,docker`
Passed regular builds for the various packages where there's changes.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None, changes are pretty widespread and not the safest.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
@@ -105,6 +105,11 @@ platform.core.profile.shortcut.edit.saveandview=Save and view
 platform.core.profile.shortcut.dev.usertype=Switch the current user type
 platform.core.profile.shortcut.dev.hiddendocs=Toggle hidden documents
 
+platform.core.profile.shortcut.notifications.update.successful=Updated shortcut preference
+platform.core.profile.shortcut.notifications.update.failed=Failed to save shortcut preference
+platform.core.profile.shortcut.notifications.reset.successful=Reset shortcut preference
+platform.core.profile.shortcut.notifications.reset.failed=Failed to reset shortcut preference
+
 ## Used to indicate where deprecated keys start
 #@deprecatedstart
 
--- a/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
+++ b/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
@@ -125,7 +125,17 @@
       <cache>forbid</cache>
     </property>
     <property>
-      <code>require(['jquery', 'xwiki-meta'], function ($, xm) {
+      <code>define('xwiki-profile-preferences', {
+  prefix: 'platform.core.profile.',
+  keys: [
+    'shortcut.notifications.update.successful',
+    'shortcut.notifications.update.failed',
+    'shortcut.notifications.reset.successful',
+    'shortcut.notifications.reset.failed'
+  ]
+});
+
+require(['jquery', 'xwiki-meta', 'xwiki-l10n!xwiki-profile-preferences'], function ($, xm, l10n) {
   const SHORTCUT_XOBJ_ACTION = Object.freeze({
     CREATE: 'create',
     DELETE: 'delete',
@@ -196,10 +206,18 @@
         let data = {
           'property#value': newValue
         };
-        $.ajax({url: updateUrl,
+        Promise.resolve($.ajax({url: updateUrl,
           type: "PUT",
           data: data
-        });
+        })).then(() => {
+          editor.trigger('xwiki:document:saved');
+          notification.replace(new XWiki.widgets.Notification(
+            l10n['platform.core.profile.shortcut.notifications.update.successful'], 'done'));
+        }).catch(response => {
+          editor.trigger('xwiki:document:saveFailed');
+          notification.replace(new XWiki.widgets.Notification(
+            l10n['platform.core.profile.shortcut.notifications.update.successful'], 'error'));
+          $('#saveFailureReason').text(response.statusText);});
       }
     }
   };


xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
```diff
@@ -60,7 +60,76 @@ user.profile.network.saving=Saving...
 user.profile.network.error=Error saving the notification setting
 user.profile.status.done=Done
 user.profile.status.saving=Saving...
-user.profile.status.error=Error saving the user status</content>
+user.profile.status.error=Error saving the user status
+
+platform.core.profile.section.shortcutPreferences=Keyboard shortcut Preferences
+platform.core.profile.section.shortcutPreferences.hint=Configure what keys to press to quickly operate common actions. The preferences defined in this section override the default shortcuts used on the wiki.When leaving the value of this preference empty, the default shortcut will be used.
```
michitux, line 66:
```suggestion
platform.core.profile.section.shortcutPreferences.hint=Configure what keys to press to quickly operate common actions. The preferences defined in this section override the default shortcuts used on the wiki. When leaving the value of this preference empty, the default shortcut will be used.
```

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/UserProfileTranslations.xml
```diff
@@ -60,7 +60,76 @@ user.profile.network.saving=Saving...
 user.profile.network.error=Error saving the notification setting
 user.profile.status.done=Done
 user.profile.status.saving=Saving...
-user.profile.status.error=Error saving the user status</content>
+user.profile.status.error=Error saving the user status
+
+platform.core.profile.section.shortcutPreferences=Keyboard shortcut Preferences
```
michitux, line 65:
```suggestion
platform.core.profile.section.shortcutPreferences=Keyboard Shortcut Preferences
```

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +247,29 @@
...
+        #set ($escapedHintId = $escapedSelectId + "_hint")
+        aria-describedby="$escapedHintId"
+      #end
+      #if ($value)
+        value="$value"
+      #elseif ($defaultValue)
+        value="$defaultValue"
+      #end&gt;
+  #else
+    &lt;span&gt;$!value&lt;/span&gt;
```
michitux, line 269: Missing escaping, this allows XSS.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +247,29 @@
...
+  #set ($defaultValueHint = "($escapetool.xml($services.localization.render('userprofile.default')))")
+  #if ($isEdit)
+    #set ($escapedId = $escapetool.xml("XWiki.XWikiUsers_0_${fieldName}"))
+    &lt;input type="text" name="$escapedId" id="$escapedId" size="30"
+      #if ($services.localization.get("XWiki.XWikiUsers_${fieldName}.hint"))
+        #set ($escapedHintId = $escapedSelectId + "_hint")
+        aria-describedby="$escapedHintId"
+      #end
+      #if ($value)
+        value="$value"
```
michitux, line 264: Missing escaping.

michitux (changes requested):
I'm wondering if it wouldn't be better to introduce a script service or at least a Velocity macro to get a keyboard shortcut to avoid these many if/else statements in Velocity.

Regarding the translation keys, all of them start with "Key" even though some of them could clearly be several keys. I would try to find a more generic formulation for this.

For all keyboard shortcuts, the UI should display the default values both in view and edit mode, this would dramatically improve the usability and value of this UI.

I'm wondering if it really makes sense to introduce so many properties in the user class. Sure, it seems like the easiest solution for now, but it hardly scales if, e.g., an extension wants to add another keyboard shortcut that should be configurable. I wonder if it wouldn't make more sense to have one XObject per customized keyboard shortcut on the user profile and a management Java class that provides the list of available keyboard shortcuts and allows querying their values (both with/without user customization to display the global default that could also be customized through a custom translation document) and that provides a way to extend them.

Was there any proposal on the forum for this feature?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -74,7 +74,12 @@
  * @param class The class to use.
  *#
 #macro(editActionButton $action $resourceIdentifier $class)
-#submitButton("action_${action}", $services.localization.render("core.shortcuts.edit.${resourceIdentifier}"), $services.localization.render($resourceIdentifier), $class)
+#if($!xwiki.getUserPreference("shortcut_edit_${resourceIdentifier}") != '')
+  #set ($shortcut = $xwiki.getUserPreference("shortcut_edit_${resourceIdentifier}"))
+#else
+  #set ($shortcut = $services.localization.render("core.shortcuts.edit.${resourceIdentifier}"))
+#end
+#submitButton("action_${action}", $shortcut, $services.localization.render($resourceIdentifier), $class)
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2981/commits/72af9633ce65cfc64171030cae704c87c6e61501 :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2981/commits/72af9633ce65cfc64171030cae704c87c6e61501 :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
```diff
@@ -155,8 +155,11 @@
         ## Override keyboard shortcut (if any)
         ##
         #if ($keyboardShortcutsEnabled && "$!extraShortcut" != "")
-          shortcut.remove("$extraShortcut");
-          shortcut.add("$extraShortcut", function() { XWiki.displayDocExtra("${extraAnchor}", "${extraTemplate}", true); }, { 'disable_in_input':true });
+          #set ($userPreferenceName = "shortcut_view_${extraName}")
+          #set ($shortcut_view_extra = $!{xwiki.getUserPreference($userPreferenceName)})
+          #set ($extraShortcutWithUserPreferences = "#if($shortcut_view_extra!='')$shortcut_view_extra#{else}$extraShortcut#end")
+          shortcut.remove("$extraShortcutWithUserPreferences");
+          shortcut.add("$extraShortcutWithUserPreferences", function() { XWiki.displayDocExtra("${extraAnchor}", "${extraTemplate}", true); }, { 'disable_in_input':true });
```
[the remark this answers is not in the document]
Sereza7: I'm not sure what kind of escaping is necessary here. There's definitely some needed escaping since the userPreference value is something any user types in their profile. However the `getShortcutValue` macro itself is velocity templating, which is used inside a js file (actionButtons.js), a js script block in some html template (shortcuts.vm), and a html template (editActionButton > submit button > name attribute of an input).
The context is varied so I'm not sure what escaping would be appropriate...
Should we just add two escaping, XML and javascript ones?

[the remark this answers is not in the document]
Sereza7: I added both escaping to the result of the velocity macro, better safe than sorry.

Sereza7:
@michitux 
I updated the gestion of the shortcuts. Now it's a specific class with a couple properties instead of being individual properties of each user. This means that any extension can easily add its own UI and use the shortcut preference system with their own shortcuts and values. This is still not very easy to do but it can be done with some velocity and javascript only (opposed to before where it wasn't possible).

The current system makes it look like they are regular fields in the user preference, but a small script creates, updates and deletes the objects associated to the preferences. This is not the easiest solution, but it scales relatively well on the number of shortcuts (not 200 empty objects on every user, objects are only created when there's an actual preference that's different from the default set at the wiki level).

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +389,50 @@
...
+#macro (displayShortcut $fieldName $label $translationKey)
+  #set ($hintKey = "XWiki.UserPreferenceShortcutsClass_${fieldName}.hint")
+  #set ($escapedSelectId = $escapetool.xml("XWiki.UserPreferenceShortcutsClass_${fieldNumber}"))
+  #set ($escapedHintId = $escapedSelectId + "_hint")
+  #set ($defaultValue = $escapetool.xml($services.localization.render($translationKey)))
+  &lt;dt&gt;
+    &lt;label #if($isEdit)for="$escapedSelectId"#end&gt;$!escapetool.xml($services.localization.render($label))&lt;/label&gt;
+    #if ($services.localization.get($hintKey))
+      &lt;span id="$escapedHintId" class='xHint'&gt;$!escapetool.xml($services.localization.render($hintKey)) ##
+      $escapetool.xml($services.localization.render('userprofile.default')): $defaultValue##
```
michitux, line 403: The default value should be put as placeholder inside the `userprofile.default` key (which also seems a bit short/non-specific).

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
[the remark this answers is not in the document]
michitux: Please don't just escape twice as escaping causes real bugs. The best practice for escaping is to put it where you put the output into the context which then also tells you what escaping is needed.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,148 @@
...
+          data: data
+        });
+      }
+    }
+  };
+  var timeoutId = 0;
+  $(document).on('input.shortcutPreference', '.shortcutPreference', function () {
+    if (timeoutId != 0){clearTimeout(timeoutId)}
+    timeoutId = setTimeout(updateShortcut, 500, this);
+  });
```
michitux, line 175: Isn't it a bit surprising to save during editing? Wouldn't it be better to have a dedicated save button?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,148 @@
...
+        $.ajax({url: updateUrl,
+          type: "PUT",
+          data: data
+        });
+      }
+    }
+  };
+  var timeoutId = 0;
+  $(document).on('input.shortcutPreference', '.shortcutPreference', function () {
+    if (timeoutId != 0){clearTimeout(timeoutId)}
```
michitux, line 173:
```suggestion
    if (timeoutId !== 0) {
      clearTimeout(timeoutId);
    }
```

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,148 @@
...
+    <property>
+      <cache>forbid</cache>
+    </property>
+    <property>
+      <code>require(['jquery', 'xwiki-meta'], function ($, xm) {
+  /**
+   * Handle the shortcut preference modifications
+   */
+  $(document).off('input.shortcutPreference');
+  let updateShortcut = function( shortcutPreferenceDOM) {
```
michitux, line 133:
```suggestion
  let updateShortcut = function (shortcutPreferenceDOM) {
```

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +389,50 @@
...
+      &lt;input type="text" class="shortcutPreference" name="$escapedSelectId" id="$escapedSelectId" size="30"
+        data-translatekey="$translationKey"
+        data-defaultvalue="$defaultValue"
+        #if ("$!shortcutPreference" != '')data-shortcutprefnumber='$shortcutPreference.getReference().getObjectNumber()'#end
+        #if ($services.localization.get("XWiki.XWikiUsers_${fieldName}.hint"))
+          #set ($escapedHintId = $escapedSelectId + "_hint")
+          aria-describedby="$escapedHintId"
+        #end
+        #if ($value)
+          value="$value"
```
michitux, line 421: `$value` needs XML escaping.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +389,50 @@
...
+        #elseif ($defaultValue)
+          value="$defaultValue"
+        #end&gt;
+    #else
+      &lt;span&gt;
+        #if("$!value" != '') $!value##
+        #else $defaultValue##
+          &lt;small&gt;($escapetool.xml($services.localization.render('userprofile.default')))&lt;/small&gt;##
```
michitux, line 429: As before, `$defaultValue` should be inserted into the translation with a placeholder so translators can place it freely.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -65,6 +65,21 @@
...
+ * @param userPreferenceName The id of the user preference that is listed in XWikiUsersDocumentInitializer.java
+ *        e.g. `shortcut_developer_usertype`
+ * @param translationKey  The key of the translation containing the wiki level defined shortcut.
+ *        e.g. `core.shortcuts.developer.user.type`
+ * @since 16.9.0-rc-1
+ *#
+#macro(getShortcutValue $translationKey)
+#set ($shortcutPreferenceObject = $xwiki.getDocument($xcontext.getUser()).getObject('XWiki.UserPreferenceShortcutsClass', 'translationKey', $translationKey))
+#if("$!shortcutPreferenceObject"!='')$escapetool.javascript($escapetool.xml($shortcutPreferenceObject.value))##
+#{else}$escapetool.javascript($escapetool.xml($services.localization.render($translationKey)))##
```
michitux, line 80: Escaping should be done by the caller, in particular if the value is used in different contexts.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -65,6 +65,21 @@
...
+ * Returns the escaped value of the shortcut. This value can be set in two ways: via user preferences and wiki level translation
+ * keys. If there are both, the user preference will take priority.
+ * @param userPreferenceName The id of the user preference that is listed in XWikiUsersDocumentInitializer.java
+ *        e.g. `shortcut_developer_usertype`
+ * @param translationKey  The key of the translation containing the wiki level defined shortcut.
+ *        e.g. `core.shortcuts.developer.user.type`
+ * @since 16.9.0-rc-1
+ *#
+#macro(getShortcutValue $translationKey)
+#set ($shortcutPreferenceObject = $xwiki.getDocument($xcontext.getUser()).getObject('XWiki.UserPreferenceShortcutsClass', 'translationKey', $translationKey))
```
michitux, line 78: This call is quite expensive. This gets the document from the cache and performs a full clone of it. This seems particularly dangerous as this macro is likely called in a loop with all shortcuts and this might happen in every request (not sure if there is further caching).

I wonder if it wouldn't make sense to introduce a script service for it. This could help in two ways:

1. It could avoid the clone of the document.
2. It/a separate component could contain a cache of the shortcuts, making it even cheaper to get all shortcuts.

An alternative to an extra cache could also be a way to simply get all shortcuts in the script service to avoid repeatedly calling the method when a larger number of shortcut values is needed.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/shortcuts.vm
```diff
@@ -131,16 +131,16 @@
...
     };
 
-    // Append developer shortcuts for toggeling userType and hiddenDocuments in the current user profile
-    shortcut.add("$services.localization.render('core.shortcuts.developer.user.type')", function() {
+    // Append developer shortcuts for toggling userType and hiddenDocuments in the current user profile
+    shortcut.add("#getShortcutValue('core.shortcuts.developer.user.type')", function() {
         developerShortcutsRestCall("${request.contextPath}/rest/currentuser/properties/usertype/next",
                 "$escapetool.javascript($services.localization.render('core.shortcuts.developer.user.type.error'))");
-    }, {'type': shortcut.type.SEQUENCE, 'disable_in_input': true });
+    }, {'type': #if("$!xwiki.getDocument($xcontext.getUser()).getObject('XWiki.UserPreferenceShortcutsClass', 'translationKey', 'core.shortcuts.developer.user.type')"!='')shortcut.type.SIMPLE#{else}shortcut.type.SEQUENCE#end, 'disable_in_input': true });
```
michitux, line 138: This call seems quite expensive. See also my comment about performance above.

michitux (changes requested):
Thank you very much for your work! I like the new design with the extra objects. Has there been a forum proposal for this storage design already?

I've added some comments, I think the main points are:

1. When are the preferences saved?
2. Escaping.
3. Performance of fetching shortcuts.

I'm also wondering if the user profile UI module is the right place for the shortcut preference class as it seems independent of the UI and the code in the Flamingo skin now basically depends on it. If the logic for getting the shortcuts should be moved into a script service/Java component, maybe it would even make sense to add a mandatory document initializer for it.

tkrieck:
Hi @Sereza7 as requested by @michitux here my review of this PR. It is not a code review, so the "Add Your Review" button is not making much sense, that's why I am leaving this as a comment instead.

---

First, I am sorry I missed the original forums post about the UI, I will post my review here as requested, but if they go against what was decided in the post, please feel free to ignore them.

- Too many descriptions start with "Key to...." making it redundant. We can keep it short and only inform the command associated and place a more general help at the top of each section. Ex. "These are the keys and combinations associated with each command. You can change each using the appropriate field. You can also return them to their default designations."
- Space utilization seems to be very inefficient, is a long list of commands and a line for each label and input. The input takes 100% of the width, but it don't need to be as the shortcuts are small texts and letters. Could we use a table comprised of two columns, one for command description and another for the command itself?
- Do I need to type the name of the modifier?
	- If possible, suggest the modifier based on the modifier being pressed. 
	- Ex. If I press "alt" fill the field with alt, otherwise allow the selection of modifier based on a fixed list (select)
- The input accepts only one key press?
	- If I focus the field and hit V and then B, the input should change V into B and not display VB
- Always show the current assigned shortcut, even if it is the default.
- To make the "return to default" action faster, place a button beside the customized shortcut. It also serves as an indication of what was customized.

For a more visual approach to the feedback above, something like the Wireframe below. Note that the buttons are only links because it's faster to prototype this way, they can be buttons in the final version (albeit, small ones).

custom shortcut ui (1)

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: I've been following the pattern already used in https://github.com/xwiki/xwiki-platform/blob/6bb5cf74c28b0fe12e29117aaa3f1f39a7b9ded1/xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml#L356-L375 

Should we replace all the uses of this translationKey with a new one that'd take a parameter and deprecate the old one in this PR?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2981/commits/74f90d2e6fb8da0fee19b1d1256e637f135e9b87 :+1:

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: Same discussion as https://github.com/xwiki/xwiki-platform/pull/2981#discussion_r1787411826

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2981/commits/67e337c00aa79e9cc213efac408d7a23099c79c5 :+1:

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
Sereza7: Updated this for https://github.com/xwiki/xwiki-platform/pull/2981#discussion_r1783076538 in https://github.com/xwiki/xwiki-platform/pull/2981/commits/74f90d2e6fb8da0fee19b1d1256e637f135e9b87

Sereza7:
@tkrieck 

> First, I am sorry I missed the original forums post about the UI, I will post my review here as requested, but if they go against what was decided in the post, please feel free to ignore them.

It was mostly for where to put the UI for this feature (more semantics and not so much about design) https://forum.xwiki.org/t/shortcut-settings-position/14116

> Too many descriptions start with "Key to...." making it redundant. We can keep it short and only inform the command associated and place a more general help at the top of each section. Ex. "These are the keys and combinations associated with each command. You can change each using the appropriate field. You can also return them to their default designations."

IMO adding details like this made sense with the form presentation. If we go with a more dense grid presentation like you proposed after, I agree that we should reduce the hints used for each shortcut. :+1: 

> Space utilization seems to be very inefficient, is a long list of commands and a line for each label and input. The input takes 100% of the width, but it don't need to be as the shortcuts are small texts and letters. Could we use a table comprised of two columns, one for command description and another for the command itself?

I went with the standard XWiki form design to match the other sections on the page. It should be pretty easy to make it into a table. I'll look into it before reopening this PR for reviews :+1: 

> Do I need to type the name of the modifier?

Currently the only accepted shortcuts are simple ones no modifiers. It's written in the `Keyboard Shortcut Preferences` hint, but it's quite small. Reworking the organization of the UI would probably help a bit with getting this info to the user.

> The input accepts only one key press?
> If I focus the field and hit V and then B, the input should change V into B and not display VB

It's a good idea but I think it's a bit extra for a first implementation. If that's okay with you, I'll open an improvement ticket to implement this later. The way I'm doing things now is just to use a regular text input.

> Always show the current assigned shortcut, even if it is the default.

It's already implemented :+1: see the second screenshot at https://github.com/xwiki/xwiki-platform/pull/2981#issuecomment-2386147786 

____

For reference, here is what the UI looks like right now. I'll be updating it to get the columns going on :+1: 
Screenshot from 2024-10-04 14-51-52

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
Sereza7: > I wonder if it wouldn't make sense to introduce a script service for it. This could help in two ways:
>    * It could avoid the clone of the document.
>    * It/a separate component could contain a cache of the shortcuts, making it even cheaper to get all shortcuts.


Isn't the main reason for it to be cheaper that we avoid cloning the doc? AFAIK 1. is not an issue except for the performance gain described in 2. 

> An alternative to an extra cache could also be a way to simply get all shortcuts in the script service to avoid repeatedly calling the method when a larger number of shortcut values is needed.

:+1: If we set up a script service, might as well make the most of it and make ourselves some shortcuts :)
From what I understood, it could also be done in addition to a cache.

tkrieck:
> It's a good idea but I think it's a bit extra for a first implementation. If that's okay with you, I'll open an improvement ticket to implement this later. The way I'm doing things now is just to use a regular text input.

Yeap, that's ok 👍

> It's already implemented 👍 see the second screenshot at https://github.com/xwiki/xwiki-platform/pull/2981#issuecomment-2386147786

Ah nice, I based my feedback from the video you posted at the start of the PR and missed this implementation.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
michitux: Talking about performance without having actual numbers (from profiling) is hard. It's a feeling of me that this could impact the performance, but it would be best to perform some actual profiling to see how bad it really is. I assume that adding a script service to get all shortcuts should already be a huge win. Adding a cache then adds quite some complexity that might not be warranted. A possibility could also be to cache shortcuts just for the current request in the context.

tkrieck:
@Sereza7 Thank you very much for the changes. From viewing the final layout, we can improve it a little bit to make reading the page faster. Could please test three other changes? 
1. **Provide a line between fields.** This way is easier for the eye to read the name of the command and follow the lines to its hotkey. Maybe you will need to space each line a little bit on view mode, otherwise it can look cramped. A rule of thumb you can follow for this is to make each line the same height on view and edit modes. 
2. **User bold text to separate sections ("Edit modes", "View Page Information", etc)**. This way, it is easier to scan the whole content and look for the section that might correlate with a given shortcut. Another alternative is to use the appropriate <H#>...</H#> tag.
3. **Give the edit input a narrower size.** When declaring sole keys (a, h, e...) the field is much bigger than it needs to be, but ofc we need to give it sufficient space to view a combination of keys also (alt+shift+e) so I guess a good compromise would be half of the size that you are currently using.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
Sereza7: When looking on making a script service, I found `$crtUserDoc`. It's in our velocity scripting doc and from what I understand, it would solve most of the performance issue with this. Currently, it's not used at all in XS, but it seems like our situation here is exactly why it's available.

https://github.com/xwiki/xwiki-platform/blob/bc80e8eac2dedfc11a4b9201b6517c50756dce1f/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/frequentlyUsedDocs.vm#L29

Testing it quickly.

This'd mean that there's still one `getObject` call for each shortcut, but as far as I understand, this is not as much of a performance issue as the `getDocument`.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/shortcuts.vm
Sereza7: Addressed partially in https://github.com/xwiki/xwiki-platform/pull/2981/commits/bbc63a04652dd86aeec049ed67bb2336ade88bb1 
Does this look like a good enough solution?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
Sereza7: Proposed the improvement in https://github.com/xwiki/xwiki-platform/pull/2981/commits/bbc63a04652dd86aeec049ed67bb2336ade88bb1 .
Do you think a script service is still useful in light of this new knowledge?

Sereza7:
@tkrieck 
Here's two quick screenshots of the updated UI at 100% zoom on a 2k monitor:
Screenshot from 2024-10-23 14-47-45
Screenshot from 2024-10-23 14-47-13


1. I wasn't sure what kind of line to go for. With the old colortheme variable as far as I can see (AFAICS) we don't have a very low contrast grey. I used a dashed line so that it wouldn't clutter the UI too much. I hard coded the height of the lines to be the exact same in view and edit mode.
2. :+1: It's a bit tricky since this part of the UI uses a bold that's not standard for XWiki. I tried to strike a better balance, but now it's a bit more difficult to distinguish between h2 "Keyboard shortcut preferences" and h3s.
3. I went with 10em. It sounded correct  to use it: 1em is the size of 1 character in the context of the block. 10em would fit 10 characters whatever the zoom level or user preferences.


All of those changes were brought in https://github.com/xwiki/xwiki-platform/pull/2981/commits/fbcd0f6d16a7a6273c405655da83d7c94997512d

Sereza7:
Both presentation and performance questions have been answered, opening this PR back for reviews.

tkrieck:
> I wasn't sure what kind of line to go for. With the old colortheme variable as far as I can see (AFAICS) we don't have a very low contrast grey. I used a dashed line so that it wouldn't clutter the UI too much. I hard coded the height of the lines to be the exact same in view and edit mode.

Looks good to me, eventually we will need a very low contrast for these situations, but it can be left for future work.

>👍 It's a bit tricky since this part of the UI uses a bold that's not standard for XWiki. I tried to strike a better balance, but now it's a bit more difficult to distinguish between h2 "Keyboard shortcut preferences" and h3s.

It is tricky indeed, you can leave it as it was then. For future work it would make sense to review the CSS of this whole UI. I can see it set its own custom CSS for the headings, I just don't know why.

> I went with 10em. It sounded correct to use it: 1em is the size of 1 character in the context of the block. 10em would fit 10 characters whatever the zoom level or user preferences.

Em is about only the height of the font, as far as I know. On non-monospaced fonts, the width of each character will vary ("X" is wider than "I"). Unfortunately, we can't really predict the width of the user chosen font, so we have to eyeball it. https://fonts.google.com/knowledge/glossary/monospaced

Sereza7:
Reverted the change on boldness of the subsection titles :+1:
image
> I can see it set its own custom CSS for the headings, I just don't know why.

It's a good example of [technical debt](https://en.wikipedia.org/wiki/Technical_debt) :) It was a good change back when it was implemented, but it didn't get properly updated, and now we have abstract styles that do not match our current system and cause extra bugs.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
Sereza7: It was not questionned on the forum topic. As proposed by Thiago, we could improve this by having "inline editing" for those properties. I think it goes beyong the scope of this first PR and I'd rather bring improvements like this in a different PR.

--> https://github.com/xwiki/xwiki-platform/pull/2981#issuecomment-2396554277

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
```diff
@@ -63,7 +63,22 @@
  * @param class The class to use.
  *#
 #macro(submitButton $name $shortcut $value $class)
-<input class="btn $!class" type="submit" name="$name"#if($keyboardShortcutsEnabled) title="$shortcut"#end value="$value"/>
+<input class="btn $!class" type="submit" name="$name"#if($keyboardShortcutsEnabled) title="$escapetool.xml($shortcut)"#end value="$value"/>
+#end
+#**
+ * Returns the escaped value of the shortcut. This value can be set in two ways: via user preferences and wiki level translation
```
michitux, line 69: There seems to be no more escaping (which is okay) but the comment needs to be updated.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
```diff
@@ -96,14 +96,14 @@ var XWiki = (function(XWiki) {
...
-        'action_edit' : "$services.localization.render('core.shortcuts.edit.backtoedit')",
-        'action_inline' : "$services.localization.render('core.shortcuts.edit.backtoedit')",
-        'action_save' : "$services.localization.render('core.shortcuts.edit.saveandview')",
-        'action_propupdate' : "$services.localization.render('core.shortcuts.edit.saveandview')",
-        'action_saveandcontinue' : "$services.localization.render('core.shortcuts.edit.saveandcontinue')"
+        'action_edit' : "$escapetool.javascript("#getShortcutValue('core.shortcuts.edit.backtoedit')")",
+        'action_inline' : "$escapetool.javascript("#getShortcutValue('core.shortcuts.edit.backtoedit')")",
+        'action_save' : "$escapetool.javascript("#getShortcutValue('core.shortcuts.edit.save')")",
+        'action_propupdate' : "$escapetool.javascript("#getShortcutValue('core.shortcuts.edit.saveandview')")",
+        'action_saveandcontinue' : "$escapetool.javascript("#getShortcutValue('core.shortcuts.edit.saveandcontinue')")"
```
michitux, line 106: Just as a suggestion, I think it would be much easier to read if this used `$escapetool.json()` to serialize one big map from Velocity to a JSON object.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -578,6 +578,40 @@ div.editProfileCategory {
...
+  list-style: none;
+  padding-left: unset;
+  &amp; &gt; li {
+    display: flex;
+    flex-direction: row;
+    gap: 1em;
+    justify-content: space-between;
+    align-items: center;
+    /* Keep the line large enough to fit an input in both view and edit mode. */
+    height: 38px;
```
michitux, line 597: Doesn't this clash with the idea that font size is variable?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
michitux: While a script service could certainly make things a bit cleaner, the performance should indeed be okay that way. The clone happens only once this way the first time an object is accessed.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/shortcuts.vm
michitux: Yes, as written above, this looks okay from a performance point of view.

tkrieck:
> It's a good example of [technical debt](https://en.wikipedia.org/wiki/Technical_debt) :) It was a good change back when it was implemented, but it didn't get properly updated, and now we have abstract styles that do not match our current system and cause extra bugs.

I see, it will be nice to revisit these styles in the future. I will keep a tab on my notes about it. Thank you!

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
Sereza7: Somehow I just noticed when testing that the crtUser is not properly set when in the context of velocity evaluation of this script. The shortcuts are not properly set here since we use `$crtUserDoc`... Looking further into it.

My guess is that we need to make sure `frequentlyUserDoc.vm` is loaded even for this velocity template evaluation. I need to find the place responsible for evaluating this .js file.

Sereza7: Addressed the main topic of this comment and the bug I found in https://github.com/xwiki/xwiki-platform/pull/2981/commits/150f1e6fc7948400cffb459ae1910e484b34151a :+1: 

Note that the solution to the bug is the low level solution. It's not a solution to the fact that the frequentlyUsedDocs template is not in the context of `$xwiki.jsfx.use`.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
Sereza7: Since this is a CSS file, I decided to hardcode this instead of relying on LESS variables.
Using LESS variables, it would have been quite a lot of operations to get this result:
max(@line-height-base*@font-size-base+ 2*@padding-base-vertical + [2*2](https://github.com/xwiki/xwiki-platform/blob/0fbc4ccbae83e56f8fbff12964f53a97b6c0ecda/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/buttons.less#L92-L93) + 2*1 px, @input-height-base + 2*1 px)
We want to fit both the button and the input in Edit mode, that both use different variables for their height...

The only impact of defining this height is the min-height. With different sizes of font sizes:
* Larger font sizes: the edit mode lines are larger than the view mode ones, in order to fit the large inputs.
* Smaller font sizes: the input and buttons have more space around them in edit mode.

I [changed the style](https://github.com/xwiki/xwiki-platform/pull/2981/commits/72e7f894a2e4f081c40d4ee916f351a0e591bef3) to make it clearer that this is not an upper limit on the height of the lines. If the content is large, the line height will increase to fit this content :) 

Those are very small drawbacks to hardcoding. I'd rather keep it hardcoded with a comment rather than spend time trying to get something to work with our current LESS and _need_ to rework it soon.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/macros.vm
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2981/commits/90f8096a8de8cc1f5f547026a343e05486503e3c :+1:

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/actionbuttons/actionButtons.js
Sereza7: Seems like this error I made is related to https://jira.xwiki.org/browse/XWIKI-19455

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
michitux: Can't you use a height based on `em`?

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +405,51 @@
...
+        #end&gt;
+      &lt;button class="btn btn-default reset-shortcut-preference" type="button&gt;
+        $services.icon.renderHTML('repeat') ##
+        $escapetool.xml($services.localization.render('platform.core.profile.shortcut.button.reset'))##
+      &lt;/button&gt;
+    #else
+      &lt;span&gt;
+        #if("$!shortcutPreference" != '') $!value##
+        #if("$!value" == '')&lt;small&gt;($escapetool.xml($services.localization.render('platform.core.profile.shortcut.hint.unbound')))&lt;/small&gt;#end
+        #else $defaultValue##
```
michitux, line 445: To me, it looks wrong to have the `(` directly after the value, I suggest inserting a space before it:
```suggestion
        #else $defaultValue ##
```

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
michitux: We usually use the approach to follow best practices in new code without changing everything else. Now if this exact pattern is already used, I'm fine to keep it here.

michitux (changes requested):
We're getting there, thank you for the updates! I think the big issues have been solved, there are just a few details left.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
michitux: My point is mostly that these inputs are presented in a page that has explicit "Save" and "Cancel" buttons at the bottom - not applying these buttons to the keyboard shortcut preferences seems weird from a usability point of view. Anyway, I'm not going to insist on this if others think that this is okay.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -37,6 +37,164 @@
...
+      } else {
+        // Update the existing preference
+        let updateUrl = url + "/objects/XWiki.UserPreferenceShortcutsClass/" + oldItemNumber +"/properties/value";
+        let data = {
+          'property#value': newValue
+        };
+        $.ajax({url: updateUrl,
+          type: "PUT",
+          data: data
+        });
```
michitux, line 169: It might be good to handle errors here and maybe also display a notification if the changes have been saved. Otherwise, users have no way to understand that there was an error saving the keyboard shortcuts. See, e.g., https://github.com/xwiki/xwiki-platform/blob/05f0ba0694a3e5e4395f493c3388ac152d218ad4/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/edit/editableProperty.js#L156-L178 for an example how this is handled in other places.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
```diff
@@ -155,8 +155,9 @@
         ## Override keyboard shortcut (if any)
         ##
         #if ($keyboardShortcutsEnabled && "$!extraShortcut" != "")
```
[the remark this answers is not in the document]
Sereza7: I updated this in https://github.com/xwiki/xwiki-platform/pull/2981/commits/ac1e5f68c1de201a9b150fbf96eac545cf79abcf :+1: 
Thank you for noticing, the logic here was off.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/docextra.vm
```diff
@@ -155,8 +155,9 @@
...
-          shortcut.remove("$extraShortcut");
-          shortcut.add("$extraShortcut", function() { XWiki.displayDocExtra("${extraAnchor}", "${extraTemplate}", true); }, { 'disable_in_input':true });
+          #set ($extraShortcutWithUserPreferences = "#getShortcutValue($extraShortcut)")
```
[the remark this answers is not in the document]
Sereza7: Thank you for noticing this :+1: 
I fixed this implementation in https://github.com/xwiki/xwiki-platform/pull/2981/commits/ac1e5f68c1de201a9b150fbf96eac545cf79abcf

I tested this new version locally and it worked properly with unset shortcuts (empty) and regular edited values

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +405,51 @@
...
+      &lt;span id="$escapedHintId" class='xHint'&gt;##
+      $escapetool.xml($services.localization.render('userprofile.default')): $defaultValue##
+      &lt;/span&gt;
+    &lt;/label&gt;
+    #set ($shortcutPreference = '')
+    #set ($shortcutPreference = $doc.getObject('XWiki.UserPreferenceShortcutsClass', 'translationKey', $translationKey))
+    #set ($value = $doc.getValue('value', $shortcutPreference))
+    #if ($isEdit)
+      &lt;input type="text" class="shortcutPreference" name="$escapedSelectId" id="$escapedSelectId" size="30"
+        data-translatekey="$translationKey"
```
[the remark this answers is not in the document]
Sereza7: Wow, that's some nice QoL I wasn't aware of! :D

I'm not very knowledgeable on custom data fields, thanks for the help :)

In the javascript I wrote to leverage those custom data fields, I used JQuery.
I updated it to work with those new fields in https://github.com/xwiki/xwiki-platform/pull/2981/commits/9b4344ccc2c3dd38b2430c512bd4eae931da9e04

With a quick Ctrl+F on this PR's diff files, I checked that those were the only places where those custom properties for the inputs were used.

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserPreferencesSheet.xml
```diff
@@ -247,6 +405,51 @@
...
+        data-defaultvalue="$defaultValue"
```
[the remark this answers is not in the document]
Sereza7: Updated :+1: 
https://github.com/xwiki/xwiki-platform/pull/2981#discussion_r1834695759

xwiki-platform-core/xwiki-platform-user/xwiki-platform-user-profile/xwiki-platform-user-profile-ui/src/main/resources/XWiki/XWikiUserSheet.xml
```diff
@@ -586,6 +586,40 @@ div.editProfileCategory {
   float: right;
 }
 
+/* Preferences */
+/* Shortcut preferences */
+span.xHint:has(+ ul.shortcut-list) {
+  margin-bottom: 10px;
```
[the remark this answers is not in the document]
Sereza7: Changed to an `em` value in https://github.com/Sereza7/xwiki-platform/commit/3edd13a216c122a18eddb1e5560b9ab8a10bee7b :+1:

PR #2986, opened 2024-03-14
XWIKI-22504: Update icon names to respect our icon naming convention

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22504
# Changes

## Description

* Reordered the icon mappings for the three icon sets in the default flavor (default-- Silk, Silk and Font Awesome) to be alphabetical and not an approximative time order (that didn't help in any way when looking for a mapping).
* Changed some icon names to fit the [new icon naming convention](https://forum.xwiki.org/t/name-convention-for-icons-in-the-xwiki-icon-set/14032)
* Added the legacy mappings in a special `deprecated` section.

## Clarifications

* These changes do not break any backwards compatibility. The introduction of the versioning for mappings will make it easier to maintain on the long term :)

# Screenshots & Video

N/A
# Executed Tests

N/A

A typo might break a mapping in one of the three themes. I don't think we have auto tests for those mappings.
I did my best to stay focused and provide the changes on the three files at the same time, but the risk for breaking a mapping is still here. Note that breaking a mapping shouldn't be a large regression because:
* Iconthemes can still fall back on the default mapping  (if it's not broken too) to provide an icon
* Worst case scenario, the icon is just not displayed and it doesn't break any more of the interface

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None, it's dev only changes and they are a bit dangerous because of the lack of tests.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/default.iconset
+++ b/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/default.iconset
@@ -1167,7 +1167,7 @@ zoom = zoom
 zoom_in = zoom_in
 zoom_out = zoom_out
 
-## Deprecated since 16.3.0 to conform to the new naming convention.
+## Deprecated since 16.8.0 to conform to the new naming convention.
 bullet_black = bullet_black
 credit-card = creditcards
 emoticon_smile = emoticon_smile
--- a/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
+++ b/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
@@ -407,7 +407,7 @@ zoom = search
 zoom_in = search-plus
 zoom_out = search-minus
 
-## Deprecated since 16.3.0 to conform to the new naming convention.
+## Deprecated since 16.8.0 to conform to the new naming convention.
 bullet_black = circle
 credit-card = credit-card
 emoticon_smile=smile-o
--- a/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-ui/src/main/resources/IconThemes/Silk.xml
+++ b/xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-ui/src/main/resources/IconThemes/Silk.xml
@@ -1205,7 +1205,7 @@ zoom = zoom
 zoom_in = zoom_in
 zoom_out = zoom_out
 
-## Deprecated since 16.3.0 to conform to the new naming convention.
+## Deprecated since 16.8.0 to conform to the new naming convention.
 bullet_black = bullet_black
 credit-card = creditcards
 emoticon_smile = emoticon_smile


xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/default.iconset
```diff
@@ -1158,6 +1165,14 @@ xhtml_valid = xhtml_valid
 zoom = zoom
 zoom_in = zoom_in
 zoom_out = zoom_out
-filter = find
+
+## Deprecated in 16.2.0 to fit the new name convention
```
manuelleduc, line 1169:
```suggestion
## Deprecated since 16.2.0 to conform to the new naming convention.
```

Sereza7: :+1:
https://github.com/xwiki/xwiki-platform/pull/2986/commits/08c4d53e32e1017950ae0518bcea4f4473369abd

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/default.iconset
```diff
@@ -1158,6 +1165,14 @@ xhtml_valid = xhtml_valid
...
+bullet_black = bullet_black
+credit-card = creditcards
+emoticon_smile = emoticon_smile
 fast-backward = control_rewind
-arrow_in = arrow_in
+fast-forward = control_fastforward
+floppy-disk = disk
+log-in=user_go
+log-out=user_go
+shopping-cart = cart
```
manuelleduc, line 1178: Preferably those would be moved in a legacy module. But, I'm not sure we have an easy way to do that.

manuelleduc (approved):
lgtm

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/default.iconset
Sereza7: There's the same kind of deprecated block in all icon themes. I'd rather have the same structure as the one we already have in translation files. It keeps things easy to maintain and understand, even if it might not the most optimized.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/icons/default.iconset
```diff
@@ -1159,6 +1166,14 @@ xhtml_valid = xhtml_valid
 zoom = zoom
 zoom_in = zoom_in
 zoom_out = zoom_out
-filter = find
+
+## Deprecated since 16.3.0 to conform to the new naming convention.
```
surli, line 1170: I guess the version should be changed

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/2986/commits/2469e4d7dc0eb0353216c28cd8a1c2d34bbeab35 :+1:

michitux:
I think this should have a Jira issue as introducing new icon names and deprecating old ones affects users and developer. Further, I wonder if we should make any efforts to not list deprecated icon names, e.g., in the icon picker. Currently, both will be listed I think which will confuse users.

Sereza7:
Setting this PR as a draft until a solution for the iconPicker "deprecated" icons have been found and implemented here.

PR #2993, opened 2024-03-15
XWIKI-21922: Introduce methods to fetch a subset of revisions in XWikiVersioningStoreInterface

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21922

# Changes

## Description

* Fix date filtering for MySQL, MariaDB, PostgreSQL and Oracle.

## Clarifications

The previous implementation of date filtering relied on timestamps stored as doubles, which can be too large for most of the DB backends we target. The queries now enforce timestamps smaller than 2038 which is the maximum supported in these databases.

# Screenshots & Video

N/A

# Executed Tests

N/A

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/hibernate/query/VersioningStoreQueryFactory.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/hibernate/query/VersioningStoreQueryFactory.java
@@ -93,11 +93,16 @@ public final class VersioningStoreQueryFactory<T>
             }
 
             Date minDate = criteria.getMinDate();
+            Date maxDate = criteria.getMaxDate();
             // Hibernate requires positive timestamps.
             if (minDate.getTime() < 0) {
                 minDate = new Date(0);
             }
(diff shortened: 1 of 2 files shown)


surli (approved):

xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/store/hibernate/query/VersioningStoreQueryFactory.java
```diff
@@ -93,11 +93,16 @@ private void applyCriteria(final long id, RevisionCriteria criteria)
             }
 
             Date minDate = criteria.getMinDate();
+            Date maxDate = criteria.getMaxDate();
             // Hibernate requires positive timestamps.
             if (minDate.getTime() < 0) {
                 minDate = new Date(0);
             }
-            predicates.add(this.builder.between(this.root.get(FIELD_DATE), minDate, criteria.getMaxDate()));
+            // Most databases store timestamps as seconds, using integers.
```
vmassol, line 101: This comment isn't clear. It should be improved as it's not obvious what the code is doing and why it works for all the databases we support

PR #2997, opened 2024-03-18
XWIKI-12674: Replace Restlet with Jersey

michitux:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-12674

# Changes

## Description

* Set a default encoding for the response as Jersey doesn't seem to set one.

## Clarifications

* Tomcat uses ISO-8859-1 by default which means that any response that doesn't explicitly set the character set will have broken UTF-8 characters. As XWiki basically always uses UTF-8 it makes sense to set a default to UTF-8.

# Screenshots & Video

# Executed Tests

`mvn clean install -U -Pdocker,legacy,integration-tests,snapshotModules -pl :xwiki-platform-container-servlet,:xwiki-platform-office-test-docker -Dmaven.test.failure.ignore -Dmaven.build.dir=target/mysql-8.3-pom-tomcat-9-jdk17-chrome -Dxwiki.checkstyle.skip=true -Dxwiki.surefire.captureconsole.skip=true -Dxwiki.revapi.skip=true -Dxwiki.spoon.skip=true -Dxwiki.enforcer.skip=true -Dxwiki.test.ui.database=mysql -Dxwiki.test.ui.databaseTag=8.3 -Dxwiki.test.ui.jdbcVersion=pom -Dxwiki.test.ui.servletEngine=tomcat -Dxwiki.test.ui.servletEngineTag=9-jdk17 -Dxwiki.test.ui.browser=chrome -Dxwiki.test.ui.wcag=true -DjenkinsAgentName="a4-004bxqhg9po1p`

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
   * To be decided, maybe if we discover that this also affects resources outside Jersey.

### DIFF
--- a/xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/java/org/xwiki/container/servlet/filters/internal/SetCharacterEncodingFilter.java
+++ b/xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/java/org/xwiki/container/servlet/filters/internal/SetCharacterEncodingFilter.java
@@ -104,12 +104,14 @@ public class SetCharacterEncodingFilter implements Filter
     {
         // Conditionally select and set the character encoding to be used
         if (ignore || (request.getCharacterEncoding() == null)) {
-            String encoding = selectEncoding(request);
-            if (encoding != null) {
-                request.setCharacterEncoding(encoding);
+            String selectedEncoding = selectEncoding(request);
+            if (selectedEncoding != null) {
+                request.setCharacterEncoding(selectedEncoding);
             }
         }
-        // Set the default encoding for the response.
+        // Set the default encoding for the response. Use the same encoding as the request as this is supposed to be
+        // the encoding that XWiki uses in general. It is not clear that XWiki in general supports any encoding besides
+        // UTF-8, though.
         response.setCharacterEncoding(this.encoding);
         // Pass control on to the next filter
         chain.doFilter(request, response);
--- a/xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/java/org/xwiki/container/servlet/filters/internal/SetCharacterEncodingFilter.java
+++ b/xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/java/org/xwiki/container/servlet/filters/internal/SetCharacterEncodingFilter.java
@@ -30,14 +30,16 @@ import javax.servlet.ServletResponse;
 
 /**
  * <p>
- * Example filter that sets the character encoding to be used in parsing the incoming request,
- * either unconditionally or only if the client did not specify a character encoding. Configuration
- * of this filter is based on the following initialization parameters:
+ * Filter that sets the character encoding to be used in parsing the incoming request,
+ * either unconditionally or only if the client did not specify a character encoding. Further, the default encoding of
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-container/xwiki-platform-container-servlet/src/main/java/org/xwiki/container/servlet/filters/internal/SetCharacterEncodingFilter.java
```diff
@@ -107,6 +109,8 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 request.setCharacterEncoding(encoding);
             }
         }
+        // Set the default encoding for the response.
```
vmassol, line 112: Could be interesting to document why we set the response to use the same encoding as the request. I don't necessarily see the relationship. You can send a request for some document but encode your request with a given encoding like ISO8859 but wan to receive the response with a UTF-8 encoding. Also shouldn't the response encodi be set only if the request encoding is set? What does it do to set a response encoding to null?

Thanks for working on this!

michitux: As far as I understand, the parameter is mandatory and thus cannot be `null`. Also, this is supposed to be the encoding that is used by XWiki. Unless the `ignore` parameter is set, the encoding of the request is honored and the encoding of the response is set independent of the request encoding.

We probably don't really support setting any encoding but UTF-8. We could also introduce a new filter to just set the response encoding to UTF-8. It just seemed less invasive to re-use the existing filter, in particular as it name doesn't imply in any way that it only affects the request. My understanding is that in theory you can set XWiki to use another encoding than UTF-8 for everything. I don't know what happens in this case, i.e., if in this case it would be correct to use this encoding for all responses or if in this case some code would convert the content to UTF-8 at some point and the response would still need to be UTF-8. I think in practice UTF-8 is the only encoding XWiki actually supports. We should probably remove the configuration and replace all references to it by UTF-8.

vmassol: > As far as I understand, the parameter is mandatory and thus cannot be null

I've mixed `encoding` vs `this.encoding`. Would be nice to rename `encoding` as it's a bad practice (reported by SonarQube BTW).

> We probably don't really support setting any encoding but UTF-8. We could also introduce a new filter to just set the response encoding to UTF-8. It just seemed less invasive to re-use the existing filter, in particular as it name doesn't imply in any way that it only affects the request. My understanding is that in theory you can set XWiki to use another encoding than UTF-8 for everything. I don't know what happens in this case, i.e., if in this case it would be correct to use this encoding for all responses or if in this case some code would convert the content to UTF-8 at some point and the response would still need to be UTF-8. I think in practice UTF-8 is the only encoding XWiki actually supports. We should probably remove the configuration and replace all references to it by UTF-8.

I don't have problem with that but it would be nice to explain in a comment.  

Also, in parallel, it would be nice to start a discussion thread on the forum with a proposal such as "Only support UTF-8 encoding" and get an agreement there. 

Thanks

tmortagne (approved):

PR #3006, opened 2024-03-21
XWIKI-22010: The rights buttons for Unregistered Users do not display anymore the value set after page refresh

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22010

# Changes

## Description

* Renamed the r variable to a more explicit name : guestRights
* Replaced the string value when calling MSCheckbox for guest users with an int value, like it is done for any other user
* Indented blocks with a wrong indentation

## Clarifications

This regression is caused because the MSCheckbox instantiation is done with a state of "1" (string) instead of 1 (int) for unregistered users.
This is inconsistent with the regular user/group MSCheckbox instantiation that used a numeric value. Beforehand it passed successfully because we used it as a list index. Now the check on the value doesn't pass and it's initialized with the default state : none.

___

I could think of 2 solutions:
1. Fix the state selector to accept int and string as a state.
2. Fix the MSCheckbox call to provide only ints for the state.

2. could be considered as leaving a breaking change in. However the MSCheckbox class is [documented nowhere](https://www.xwiki.org/xwiki/bin/view/Main/Search?text=MSCheckbox&f_type=DOCUMENT&f_locale=en&f_locale=&r=1). Because of that I suppose it's not part of our public API, so as long as we fix all internal uses, it's okay to update it.

I'd rather keep variable's type consistent here, so I'll go with 2.

# Screenshots & Video

https://github.com/xwiki/xwiki-platform/assets/28761965/7a35502b-79e2-4890-97ce-a469b01e862e

We can see on this video that unlike before, the state of the unregistered user checkboxes are correctly initialized after a refresh.
# Executed Tests

Only manual tests, see demo above. No automatic tests were broken by this regression so I suppose fixing it does not break anything. Most of the changes in this PR are codestyle only.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No backport (same as https://github.com/xwiki/xwiki-platform/pull/2649)

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/rightsUI.vm
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/rightsUI.vm
@@ -129,7 +129,7 @@ $xwiki.ssfx.use('js/xwiki/table/livetable.css', true)
     #end
   #end
 #end
-#set ($r = [])
+#set ($guestRights = [])
 #foreach ($i in [0..$maxlevel])
   #set ($value = 0)
   #if ($allowWins.contains($i))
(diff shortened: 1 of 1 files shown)


manuelleduc (approved):
lgtm. @Sereza7 do you mind splitting this is two commits?
- one for the actual fix
- one `[Misc]` for the code style changes
Thanks!

Sereza7:
I reverted the previous commit and added back only the necessary change here :+1:

manuelleduc:
> Started a PR with the misc changes at #3009 :)

It's also acceptable to rework the history locally and "force-with-lease", I would I use "rebase and merge", but that works too. Thanks a lot!

PR #3015, opened 2024-03-22
XWIKI-18356: "aria-labelledby" for navigation panel entries and corresponding ids contain illegal whitespaces

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-18356

# Changes

## Description

* Added escaping for the spaces in the IDs of the jstree nodes.
* Unescaped this when using the ID to query new nodes for the tree

## Clarifications

* With the changes in this PR, we make sure IDs from xtrees do not contain any spaces. This is a HTML standard rule that should be followed.
* We need to unescape the ID when reusing it to make a query for child elements.
* This change would break most features relying on the value of these IDs to work. AFAICS the only one is only the child query and it's fixed.
* Because this value was not conform to HTML standards (and prevented some browser features like label association to work properly...), I deem it correct to assume we wouldn't include it in our front-end API, and breaking backwards dependency on this is not a matter worth of a vote. 

# Screenshots & Video

https://github.com/xwiki/xwiki-platform/assets/28761965/5fa0a69b-f4fa-430e-8faf-0729aca9c25a

# Executed Tests

Manual tests on the Navigation panel tree (see demo video) and manual verification on a couple other document trees went well.
Successfully built `mvn clean install -f xwiki-platform-core/xwiki-platform-tree/ -Pdocker,quality,integration-tests`, however there is no tests in this module itself.
I could not find tests that would use this ID elsewhere, except in the jasmine specs for xwiki-platform-web-war. `mvn clean install -Pintegration-tests-chrome -f xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war` passes without a fail but I'm not sure this is even a relevant test for what we changed...

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No backport. This is quite a dangerous PR because it changes an important attribute (the id) on multiple nodes on pretty much every page (most pages include at least one tree somewhere).


xwiki-platform-core/xwiki-platform-tree/xwiki-platform-tree-webjar/src/main/webjar/tree.js
```diff
@@ -71,7 +71,15 @@ define([
         urlParams[pair[0]] = pair[1];
       }
     }
-    return $.post(new URL('?', url), $.param($.extend(urlParams, data), true));
+    return $.post(new URL('?', url), $.param($.extend(urlParams, data), true))
+      .then((nodes) => {
+        nodes.forEach((object) => {
+          // Each node will be rendered as an HTML element.
+          // As such, we want to make sure we escape space characters in their ids
+          object.id = object.id.replaceAll(' ','%20');
```
michitux, line 79:
```suggestion
          object.id = object.id.replaceAll(' ', '%20');
```

Are we sure that there is no `%20` in the id attribute? If not, shouldn't we also replace `%` by `%25`? Alternatively, would it be possible to store the original id, e.g., in a data attribute and use that as the id parameter in the request?

michitux (changes requested):
The escaping doesn't seem very robust to me, see also my other comment.

xwiki-platform-core/xwiki-platform-tree/xwiki-platform-tree-webjar/src/main/webjar/tree.js
Sereza7: Somehow testing manually this pitfall doesn't bring up any error... I don't understand why so there's probably something that needs to be fixed somewhere anyway :thinking:

Sereza7: Okay, the test I conducted yesterday did not fail because I misremembered what I should test. After taking another look, the getChildren can get broken for page titles with a `%20` :+1:

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3015/commits/31648cabdc7ad4ec206cb7ae48b954c8d78dad2c :+1: 

18356-afterChanges
We can see on the screenshot above (taken with the latest version of this PR) that page titles with `%20` and `%25` as well as other page titles all fulfill the following:
* The IDs of the `a` and `li` elements displaying them are properly escaped: there's no space inside.
* The request to get their children is successful: the tree is generated properly and entirely.

Sereza7:
All comments have been addressed. This PR is ready for further reviews.

michitux (approved):
Looks good apart from that comment.

xwiki-platform-core/xwiki-platform-tree/xwiki-platform-tree-webjar/src/main/webjar/tree.js
```diff
@@ -101,9 +109,11 @@ define([
     parameters = parameters || {};
     if (!childrenURL) {
       childrenURL = this.element.attr('data-url');
+      // We make sure to unescape the parent node id because we use it as content of a POST request,
+      // and not the URL of a GET.
```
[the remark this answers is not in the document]
Sereza7: :+1: I was confused at the start because this looked quite similar to an URL. With hindsight this comment is more confusing than anything else.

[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3015/commits/46170ae8afde9034dc383d03dacb6bd56eb1ef81 :+1:

PR #3023, opened 2024-03-27
XWIKI-21452: Macros info, success, warning and error are only distinguished by colors

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21452
# Changes

## Description

* Updated style to take into account the addition of an icon in xwiki-rendering
* Fixed tests

## Clarifications

Most of these changes were initially merged with https://github.com/xwiki/xwiki-platform/pull/2590, but after a passage through CI we noticed a lot of tests broke because of the changes. So we reverted it and I updated the failing tests to fit the new architecture of the message macros.

# Screenshots & Video

image

# Executed Tests

First, I built all changes with :
```
mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-message -Pquality,integration-tests,docker;mvn clean install -f xwiki-rendering-macros/xwiki-rendering-macro-box -Pquality,integration-tests,docker; mvn clean install -f xwiki-rendering-api -Pdocker,integration-tests,quality; spd-say "DONE"
```
```
mvn clean install -f xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki -Pdocker,quality,integration-tests; mvn clean install -f xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-transformations/xwiki-platform-rendering-transformation-icon -Pdocker,integration-tests,quality;spd-say "DONE"
```
and
```
mvn clean install -f xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources 
```
___
Then I checked all places that failed on the CI when we merged https://github.com/xwiki/xwiki-platform/pull/2590:
platform-panels-test-docker checked on 18/03 17:00
platform-administration-test-docker checked on 25/03
platform-invitation-test-docker checked on 18/03 14:50
platform-livedata-test-docker checked on 18/03 14:40
platform-flamingo-skin-test-docker checked on 21/03 15:00
platform-realtime-wysiwyg-test-docker checked on 18/03 14:10
platform-extension-test-tests checked on 18/03 morning
platform-repository-test-tests checked on 18/03 morning
platform-rendering-xwiki checked on 18/03 13h40

I tested all of those with a command similar to:
```
mvn clean install -f xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki -Pdocker,quality,integration-tests`
```

All of those needed a slight change to the tests themselves. Most of the changes only needed to take into account the new layout inside the blocks, which ended up adding a few newlines in the computed text.
Administration tests were a bit complex to fix, because some cases used the `#info` velocity template while others used the `{{info}}` XWiki macro. In the end I added a formatting step when retrieving the content of the block that changed.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No backport. Dangerous and not critical.
* Should only be merged after https://github.com/xwiki/xwiki-rendering/pull/296


xwiki-platform-core/xwiki-platform-realtime/xwiki-platform-realtime-wysiwyg/xwiki-platform-realtime-wysiwyg-test/xwiki-platform-realtime-wysiwyg-test-docker/src/test/it/org/xwiki/realtime/wysiwyg/test/ui/RealtimeWYSIWYGEditorIT.java
```diff
@@ -707,7 +707,7 @@ void editSameMacro(TestReference testReference, TestUtils setup)
         // Focus the rich text area.
         secondTextArea.click();
         // Move the caret inside the information box.
-        secondTextArea.sendKeys(Keys.ARROW_UP, "two ");
+        secondTextArea.sendKeys(Keys.ARROW_UP, Keys.TAB, "two ");
```
michitux, line 710: This doesn't seem good. Why do you need to press tab now? Where is the focus before pressing tab?

xwiki-platform-core/xwiki-platform-realtime/xwiki-platform-realtime-wysiwyg/xwiki-platform-realtime-wysiwyg-test/xwiki-platform-realtime-wysiwyg-test-docker/src/test/it/org/xwiki/realtime/wysiwyg/test/ui/RealtimeWYSIWYGEditorIT.java
```diff
@@ -761,7 +761,7 @@ void editSameMacro(TestReference testReference, TestUtils setup)
         firstMacroEditModal.clickCancel();
 
         firstEditPage.clickSaveAndView();
-        assertEquals("{{info cssClass=\"bar\" title=\"Some cool title\"}}\ntwo one\n{{/info}}\n\n ",
+        assertEquals("{{info cssClass=\"bar\" title=\"(% class=~\"box-title box-title~\" %)Some cool title\"}}\ntwo one\n{{/info}}\n\n ",
```
michitux, line 764: Why do we get this strange wiki syntax? This seems wrong.

Sereza7: I fixed it [in rendering](https://github.com/xwiki/xwiki-rendering/pull/296/commits/d64771c76e26747f0b529b43f7257a2de447873b). Somehow the FormatBlock was interacting weirdly with its child Metadata when editing inline. The GroupBlock doesn't have this issue and behaves nicely. This also fixes the extra newlines that were added in the parsing of the block content.
WIP: reverting the changes on most tests and making sure they still pass successfully

xwiki-platform-core/xwiki-platform-realtime/xwiki-platform-realtime-wysiwyg/xwiki-platform-realtime-wysiwyg-test/xwiki-platform-realtime-wysiwyg-test-docker/src/test/it/org/xwiki/realtime/wysiwyg/test/ui/RealtimeWYSIWYGEditorIT.java
Sereza7: The focus was on the title so an extra tab was needed. With the change described in https://github.com/xwiki/xwiki-platform/pull/3023#discussion_r1561239990, there's no need for this change anymore :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/resources/ApplicationResources.properties
```diff
@@ -42,6 +42,12 @@
 #     old.key.name=Some translation
 ###############################################################################
 
+# Those entries are used to label the message boxes of different types.
```
michitux, line 45: This text contradicts the generality of the icon provider. The icon provider should not just be usable for message boxes. Also, it is not clear to me why the icon provider is the one that is responsible for providing an alternative text. Wouldn't it make more sense that the code that calls the icon provider should also add the alternative text?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/test/java/org/xwiki/rendering/internal/util/XWikiIconProviderTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.xwiki.test.junit5.mockito.MockComponent;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for {@link XWikiIconProvider}.
+ *
+ * @version $Id
+ *  @since 16.2.0-rc1
```
michitux, line 44: We normally don't put since versions on tests.

michitux (changes requested):
The test changes look good now. What I'm not yet happy with is how the alternative text is added. We [agreed](https://forum.xwiki.org/t/accessibility-html-codestyle-icons-along-text-alternatives/13378) that the icon renderer should not provide an alternative text for icons, but that instead icons should only be used alongside alternative text. To me, the new icon provider is not different from the icon renderer here.

We could use a similar mechanism as for the icon to get a `Block` with an alternative text that is then added to the output, though in this case the API could be fully internal and would just be used for the message box macro.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
```diff
@@ -952,15 +952,15 @@ $html
...
 #end
 
 #macro(error $text)
-<div class="errormessage"><span class="messagetype">$services.localization.render('error'): </span>$text</div>
+<div class="box errormessage"><span class="messagetype">$services.localization.render('error'): </span>$text</div>
 #end
 
 #macro(info $text)
-<div class="infomessage">$text</div>
+<div class="box infomessage">$text</div>
```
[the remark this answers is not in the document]
Sereza7: We applied changes to the style of the boxes. When generated using the macro, all the message types have the box type by default so I assumed I could factorize some of the style to get onto the `.box` class. It makes sense to do so IMO. See xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/messages.less

These velocity macro did not generate the boxes exactly the same as the XWiki macro and the lack of the `box` class would break the style. With this change the few use of those velocity macros should match the normal style :)

Another solution would be to rework the selectors in `messages.less` to take into account the cases where there is no `box` class. I'd rather have simpler LESS and more consistency.

Let me know if you think we should move onto the other solution :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/util/XWikiIconProvider.java
```diff
@@ -0,0 +1,95 @@
...
+            IconSet iconSet = getIconSet(iconName);
+            String iconContent = this.iconRenderer.renderHTML(iconName, iconSet);
+            Block iconAlternative;
+            try {
+                // Try to retrieve a translation for the icon.
+                iconAlternative = new FormatBlock(List.of(l10n.getTranslation(
+                    String.format("rendering.icon.provider.icon.alternative.%s", iconName)).render()), Format.NONE);
+            } catch (Exception e) {
+                // As a fallback, we just get the english name of the icon as an alternative.
+                iconAlternative = new FormatBlock(List.of(new WordBlock(iconName)), Format.NONE);
```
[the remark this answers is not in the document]
Sereza7: It's probably a bias from my knowledge in linguistics. I went to check and the documentation of the WordBlock says that it shouldn't contain any special character: https://github.com/xwiki/xwiki-rendering/blob/7dc092d4ce5c2c5955cba17a7982f96fd4628727/xwiki-rendering-api/src/main/java/org/xwiki/rendering/block/WordBlock.java#L40-L41 I agree with your comment, I'll propose a change.

michitux:
Did you perform any tests how the icon and the CSS changes in general affect the optional image that can be specified in the macro parameters?

Sereza7:
@michitux 

> Did you perform any tests how the icon and the CSS changes in general affect the optional image that can be specified in the macro parameters?

https://forum.xwiki.org/t/updating-the-presentation-of-message-macros/13872/9?u=charpentierlucas
I can make sure it still behaves this way, but it shouldn't have changed, since the only change since then that is not dev only (for tests) is the nature of the title block (now parsed to a HTML div instead of a span).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/util/XWikiIconProvider.java
```diff
@@ -0,0 +1,95 @@
...
+import org.xwiki.rendering.syntax.Syntax;
+
+import java.util.List;
+
+/**
+ * Component to use the icon theme to provide a proper block for displaying an icon.
+ *
+ * @version $Id$
+ * @since 15.10.6
+ * @since 16.1.0RC1
```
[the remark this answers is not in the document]
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3023/commits/1ff7bb4f2591e05a3ae7b52b620a441b31f76f0c :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/test/java/org/xwiki/rendering/internal/util/XWikiIconProviderTest.java
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3023/commits/1ff7bb4f2591e05a3ae7b52b620a441b31f76f0c :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiErrorMessageMacro.java
```diff
@@ -0,0 +1,52 @@
...
+    @Inject
+    private ContextualLocalizationManager l10n;
+
+    /**
+     * Create and initialize the descriptor of the macro.
+     */
+    public XWikiErrorMessageMacro()
+    {
+        super();
+        this.iconPrettyName = l10n.getTranslation(iconPrettyNameKey).toString();
```
michitux, line 50: You should use `getTranslationPlain` here to get the translation as plain text.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiErrorMessageMacro.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @version $Id$
+ * @since 16.4.0RC1
+ */
+@Component
+@Named("error")
+@Singleton
+public class XWikiErrorMessageMacro extends ErrorMessageMacro
+{
+    private static String iconPrettyNameKey = "rendering.macro.message.icon.alternative.error";
```
michitux, line 40: This should be `private static final String ICON_PRETTY_NAME_KEY` (Checkstyle definitely won't be happy with the naming, otherwise).

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
```diff
@@ -288,7 +288,8 @@ disk = floppy-o
 drink = glass
 drive = hdd-o
 email = envelope-o
-emoticon_smile=smile-o
+emoticon_smile= smile-o
+emoticon_unhappy = frown-o
```
michitux, line 292: I think this change is unrelated to this PR.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/util/XWikiIconProvider.java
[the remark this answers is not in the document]
michitux: I think this whole code for the alternative text needs to be removed again.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/resources/ApplicationResources.properties
michitux: I think this should be removed again now that the translations are handled by the overridden macros.

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm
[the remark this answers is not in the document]
michitux: Sounds good.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.xwiki.platform</groupId>
+        <artifactId>xwiki-platform-rendering-macros</artifactId>
+        <version>16.3.0-SNAPSHOT</version>
```
michitux, line 31: This version needs to be updated.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/resources/META-INF/components.txt
```diff
@@ -0,0 +1,4 @@
+org.xwiki.rendering.internal.macro.message.XWikiInfoMessageMacro
+org.xwiki.rendering.internal.macro.message.XWikiWarningMessageMacro
+org.xwiki.rendering.internal.macro.message.XWikiErrorMessageMacro
+org.xwiki.rendering.internal.macro.message.XWikiSuccessMessageMacro
```
michitux, line 4: I think these components need to be declared with a lower priority (I think we typically use 500) to actually reliably override the components from `xwiki-rendering`. Further, the module needs to be added as a dependency somewhere, the most relevant place probably being https://github.com/xwiki/xwiki-platform/blob/3d4b2e28553b594b1ddb255de69c3b645b551e54/xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml#L301-L306. Now we have lots of modules that also directly depend on the macros from `xwiki-rendering`, I don't know if we should all change them.

michitux (changes requested):
Some more feedback on what remains to be done from my point of view.

In general, I'm wondering if overriding the macros is the best solution. It certainly does the job, though. I'm a bit worried, though, that now some UI tests might not use the correct version of the message macros and won't get the translations so we won't test the translations at all.

An alternative would be to have something like the icon provider but for the message names, basically a kind of mini translation service just for the macros. But it also doesn't feel right as a solution. @tmortagne do you have any idea how we could best provide translated strings to the message macros in `xwiki-rendering`?

tmortagne:
> In general, I'm wondering if overriding the macros is the best solution.

Yeah, feels like a hack to me too.

> An alternative would be to have something like the icon provider but for the message names, basically a kind of mini translation service just for the macros. But it also doesn't feel right as a solution. @tmortagne do you have any idea how we could best provide translated strings to the message macros in xwiki-rendering?

Not really a better idea than you (there are not many other choices than overwriting, or using a component implemented in xwiki-platform when you need to manipulate xwiki-platform concepts, which ultimately is an overwrite too, just a more reusable one). I feel like the cleanest might be to "simply" move the localization API to xwiki-rendering (it does not depend on anything from platform), but it's starting to be a bit big for this pull request :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/resources/META-INF/components.txt
tmortagne: Yes, any component designed to overwrite another should be prefixed with `500:` (or something else for more specific cases).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/pom.xml
Sereza7: Done in 

I think this whole code for the alternative text needs to be removed again.
 :+1:

xwiki-platform-core/xwiki-platform-icon/xwiki-platform-icon-fontawesome/src/main/resources/IconThemes/FontAwesome.xml
Sereza7: Removed in https://github.com/xwiki/xwiki-platform/pull/3023/commits/9e7759387dc99ee81b038d9712712fd62ab6ada7 :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiErrorMessageMacro.java
Sereza7: Updated in https://github.com/xwiki/xwiki-platform/pull/3023/commits/9e7759387dc99ee81b038d9712712fd62ab6ada7 :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiErrorMessageMacro.java
Sereza7: Updated in https://github.com/xwiki/xwiki-platform/commit/9e7759387dc99ee81b038d9712712fd62ab6ada7 :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/resources/ApplicationResources.properties
Sereza7: The translation values are still needed in the overridden macros though. I kept them in, let me know if I misunderstood something here :)

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/resources/META-INF/components.txt
Sereza7: > I think these components need to be declared with a lower priority (I think we typically use 500) to actually reliably override the components from xwiki-rendering. 

Added the priority in https://github.com/xwiki/xwiki-platform/pull/3023/commits/9e7759387dc99ee81b038d9712712fd62ab6ada7 :+1: . 


> Further, the module needs to be added as a dependency somewhere, the most relevant place probably being [...]

I replaced the xwiki-rendering dependency here with the overriding module.
However I don't understand too well why we'd need to do it only here and in none of the other results of:
https://github.com/search?q=repo%3Axwiki%2Fxwiki-platform%20xwiki-rendering-macro-message&type=code

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/util/XWikiIconProvider.java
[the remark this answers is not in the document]
Sereza7: Done in https://github.com/xwiki/xwiki-platform/pull/3023/commits/40015cdd1798c3e736117671fa54029dafafe898 :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/resources/ApplicationResources.properties
michitux: As far as I can see, they are now provided with identical keys in two different files. Is that really intentional and necessary?

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/resources/META-INF/components.txt
michitux: If you don't change the other places, in these other places, you won't have the new macros. For an actual installation, the dependency in the war will be installed, and thus the macros are overridden. Most of these UI modules have an extra dependency to have the macro available in the page tests. In these page tests, the original macro will thus be used and not the overridden one.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiWarningMessageMacro.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @version $Id$
+ * @since 16.7.0RC1
+ */
+@Component
+@Named("warning")
+@Singleton
+public class XWikiWarningMessageMacro extends WarningMessageMacro
+{
+    private static final String ICON_PRETTY_NAME_KEY = "rendering.macro.message.icon.alternative.warning";
```
tmortagne, line 40: So you decided to make this pretty name related to the macro and not to the icon chosen for the macro, right ?

michitux: We decided in https://forum.xwiki.org/t/accessibility-html-codestyle-icons-along-text-alternatives/13378 that icons should always be used next to a text alternative. This is why I pushed for having the pretty names independent of the icons and to not provide them by the icon manager.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/resources/ApplicationResources.properties
tmortagne: > As far as I can see, they are now provided with identical keys in two different files. Is that really intentional and necessary?

Same here, those looks like leftovers to me. The keys make more sense in the module where they are used (which is already the case AFAICS).

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiWarningMessageMacro.java
tmortagne: Sure, I just wanted to be sure it was the intended goal given the handling of the translations.

tmortagne:
> I didn't implement the mini translation service because it adds a lot of architecture (and need for tests :))) ) for only four translations.

The minimum is not really a "translation service" (and certainly not a public one). All you need is an internal macro icon pretty name provider which is overwritten on platform. It's a pity to introduce an icon provider to avoid overwriting those macros and finally overwrite all those macros…

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/internal/util/XWikiIconProvider.java
```diff
@@ -0,0 +1,76 @@
...
+@Component
+@Singleton
+public class XWikiIconProvider extends DefaultIconProvider
+{
+    @Inject
+    private IconSetManager iconSetManager;
+    @Inject
+    private IconRenderer iconRenderer;
+    @Inject
+    private ContextualLocalizationManager l10n;
```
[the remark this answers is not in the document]
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/pull/3023/commits/4a8cecb09f1efb97179059ad60968a789de7147a :+1: 
Thanks

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiWarningMessageMacro.java
Sereza7: Yeah, because the alternative of the icon is context dependant. What is used as a `success` icon here could also be a `done` icon in another context.

Sereza7:
Ready for reviews.

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/resources/ApplicationResources.properties
```diff
@@ -0,0 +1,48 @@
+# ---------------------------------------------------------------------------
```
tmortagne, line 1: Should this file be removed ? All those translations are already provided by xwiki-rendering-macro-message.

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3023/commits/2df9eece0f9c674dc55d438e2de08637d387e01a :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * XWiki specific component to provide an alternative text for an icon.
+ *
+ * @version $Id$
+ * @since 16.7.0RC1
+ */
+@Component(roles = MacroIconPrettyNameProvider.class)
+@Singleton
+public class MacroIconPrettyNameProvider
```
tmortagne, line 36: This does not override anything, it should have extended the xwiki-rendering `MacroIconPrettyNameProvider` (and be named `XWikiMacroIconPrettyNameProvider` to be more clear).

xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
```diff
@@ -300,7 +300,7 @@
     </dependency>
     <dependency>
       <groupId>org.xwiki.rendering</groupId>
```
tmortagne, line 302: Wrong groupid.

xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
```diff
@@ -300,7 +300,7 @@
...
-      <artifactId>xwiki-rendering-macro-message</artifactId>
+      <artifactId>xwiki-platform-rendering-macro-message</artifactId>
       <version>${rendering.version}</version>
```
tmortagne, line 304: Wrong version.

tmortagne:
Sorry, I missed some mistakes before merging (but would have been nice to re-test all that after the refactoring).

vmassol:
@Sereza7 Should the following be changed too: 

* https://github.com/xwiki/xwiki-platform/blob/80609c31ad28e1b43436314df9fc2e3a44c418b1/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/extension.vm#L742
* https://github.com/xwiki/xwiki-platform/blob/bd419a583691c1156a38e44022781ae876932da0/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/distribution/macros.vm#L124

There might be other places like these.

vmassol:
Also, why is https://github.com/xwiki/xwiki-platform/blob/beaeb18d3aaf1028f40cbfcf2a958f0c1c459b8d/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm#L963 different in structure from the error or warning ones?

tmortagne:
It seems to be the same general structure, it's just that we don't show "Information:" (like we have "Error:" and "Warning:") in the case of an information message. It was only like this before.

vmassol:
> It seems to be the same general structure, it's just that we don't show "Information:" (like we have "Error:" and "Warning:") in the case of an information message. It was already like this before.

Doesn't seem very consistent to me and probably something to fix (possibly with a boolean parameter in the macro to decide to show the title or not).

vmassol:
@Sereza7  and all: There's a discussion to have which is about the alternate text to use for icons. We used to use "Information" before and it's now been changed to "Info". I personally find "Information" to be better ("Info" can mean several things, see https://www.thefreedictionary.com/words-that-start-with-info ;)). Also we've said that we should try to avoid shortcuts in XWiki and be explicit in sentences.

I know that we're now using the Macro name and that the macro is called `{{info}}` but when a screen reader or otherwise reads it, I'd find "Information" to be more explicit.

For example I've had to fix the Invitation tests, see: https://github.com/xwiki/xwiki-platform/commit/3050a0e6668f9faaffaa3f82905f1655474a500f

WDYT?

tmortagne:
> I know that we're now using the Macro name and that the macro is called {{info}} but when a screen reader or otherwise reads it, I'd find "Information" to be more explicit.

Actually, we don't really use the Macro name in XWiki. The macro name is just [the fallback in xwiki-rendering](https://github.com/xwiki/xwiki-rendering/blob/master/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java#L43), but on platform side, and so in an XWiki instance, [a translation is used](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/XWikiMacroIconPrettyNameProvider.java#L50), and [the translation text is "Information"](https://github.com/xwiki/xwiki-rendering/blob/master/xwiki-rendering-macros/xwiki-rendering-macro-message/src/main/resources/ApplicationResources.properties#L22).

vmassol:
@tmortagne interesting, thanks for the info. 

> So the question is more if we want to have `xwiki-platform-rendering-macro-message` in the minimal WAR.

Yes indeed (I see it's been added to the XS WAR: https://github.com/xwiki/xwiki-platform/blob/191e7fee453d59d9d376bbf91d1a5072d7cd850c/xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml#L301-L306), so either that or we add it to `WARBuilder` for the functional tests.

Not easy to decide but I think if we push it to the extreme, a minimal war should not have message macros (it could have `#info` velocity macros for templates though).

So FTM I'll add the dep to `WARBuilder` and revert my change in `InvitationIT`

vmassol:
hmm actually I wonder if we shouldn't just change our dep on the rendering message macro to a dep on `xwiki-platform-rendering-macro-message`.

xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/commit/ed0cd6cbbef0be26342da4722244e60a9544f422 Thanks :+1:

xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-message/src/main/java/org/xwiki/rendering/internal/macro/message/MacroIconPrettyNameProvider.java
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/commit/eb30e7074a52d8d60b2da00459d6a36b263a4607 :+1: Thank you

xwiki-platform-distribution/xwiki-platform-distribution-war-dependencies/pom.xml
Sereza7: Fixed in https://github.com/xwiki/xwiki-platform/commit/5b087f8307cb42f4879090e830147089a3af9f22 :+1: Thank you :)

PR #3030, opened 2024-03-28
XWIKI-21910: Display pagination of a LiveData only if number of entries exceed 10

Sereza5:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21910

# Changes

## Description

* Updated both layouts to not display the pagination tools if there's only one page of entries

## Clarifications

* Note that I went with the solution of hiding the pagination if the initial state of the livedata has only one page. IMO this is easier to understand for users than a cutoff at numberOfEntries=10. When merging this, I'll update the Livedata doc to make sure users can find out what's happening if they couldn't fitgure it out on their own.

# Screenshots & Video

Here are what a few default flavor livedatas look like after the changes in this PR:
Menu listing
21910-afterPRMenu
Notification preferences filters
21910-afterPRNotificationPreferences

# Executed Tests

Successfully built `mvn clean install -f xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar -Pdocker,integration-tests,quality`. Made sure to not add warnings to the npm run lint.
I doubt another module integration tests using a livedata would fail since this pagination element is only removed in case it's useless.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None, master only.


manuelleduc:
> * Note that I went with the solution of hiding the pagination if the initial state of the livedata has only one page. IMO this is easier to understand for users than a cutoff at numberOfEntries=10. When merging this, I'll update the Livedata doc to make sure users can find out what's happening if they couldn't fitgure it out on their own.

Imo it shouldn't be the default behavior. Or at least it should be configurable, i.e., provide an option to allow the pagination to be hidden when there is only one page.
I'd suggest making a proposal.

Sereza7:
Started a proposal to discuss this topic at https://forum.xwiki.org/t/deactivate-livedata-pagination-when-theres-only-one-page/14669 .

Sereza7:
In regards to the results of the proposal, I updated the implementation:

# Screenshots & Video


I ran the changes locally and got the following results:

https://github.com/user-attachments/assets/06bc233f-b41d-46fc-bb60-bf6511e8c195

We can see on this video that the parameter changes the display of the livedata in the expected way. Note that there's one edge case where this might be confusing for the user. When we don't want to show the pagination on a single page and there's initially multiple pages. If we increase the size of the pages to have only one, the livedata will be reloaded and the pagination will not be available anymore. We can address this possible confusion with a small note in the documentation, IMO this is better to stay consistent here and not introduce extra logic. Moreover, I expect this parameter to be used mostly on livedatas with a known number of elements that fills up only one page or less.

# Executed Tests


Successfully built `mvn clean install -f xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api; mvn clean install -f xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar`. Ran it locally with expected results on multiple livedata configurations (see manual test video above).

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataPaginationConfiguration.java
```diff
@@ -157,6 +159,24 @@ public void setShowPageSizeDropdown(Boolean showPageSizeDropdown)
         this.showPageSizeDropdown = showPageSizeDropdown;
     }
 
+    /**
+     * @return whether to show or not the pagination should be shown when there's only one page.
+     */
+    public Boolean getShowPaginationOnSinglePage()
```
manuelleduc, line 165: missing since and unstable annotations.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataPaginationConfiguration.java
```diff
@@ -157,6 +159,24 @@ public void setShowPageSizeDropdown(Boolean showPageSizeDropdown)
...
+    {
+        return showPaginationOnSinglePage;
+    }
+
+    /**
+     * Sets whether to show or not the pagination should be shown when there's only one page.
+     *
+     * @param showPaginationOnSinglePage {@code true} to always show the pagination, {@code false} otherwise
+     */
+    public void setShowPaginationOnSinglePage(Boolean showPaginationOnSinglePage)
```
manuelleduc, line 175: missing since and unstable annotations.

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/internal/LiveDataRendererParameters.java
```diff
@@ -276,6 +278,27 @@ public void setDescription(String description)
...
+        return this.showPaginationOnSinglePage;
+    }
+
+    /**
+     * Sets whether to show or not the pagination list when there is only one page.
+     *
+     * @param showPaginationOnSinglePage {@code true} to always show the page list, 
+     * {@code false} only show it when there's multiple pages;
+     * leave {@code null} to inherit from the default configuration
+     * @since 16.10.3
```
manuelleduc, line 295: Reminder: since versions to be updated

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/LivedataPagination.vue
```diff
@@ -296,6 +296,9 @@ export default {
     },
     showEntryRange() {
       return this.data.meta.pagination.showEntryRange
+    },
+    showPagination() {
+      return (this.logic.getPageCount() > 1) || this.data.meta.pagination.showPaginationOnSinglePage;
```
manuelleduc, line 301: ymmv but the parenthesis as not required here
```suggestion
      return this.logic.getPageCount() > 1 || this.data.meta.pagination.showPaginationOnSinglePage;
```

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/vue/layouts/table/LayoutTable.vue
```diff
@@ -34,7 +34,7 @@
     -->
     <LivedataTopbar>
       <template #left>
-        <LivedataPagination />
+        <LivedataPagination v-if="isMoreThanOnePage" />
```
manuelleduc, line 37: Unless I'm missing something, `isMoreThanOnePage` does not exists.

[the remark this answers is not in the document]
Sereza7: My bad, it was an artifact left from a previous implementation.
Removed in https://github.com/xwiki/xwiki-platform/pull/3030/commits/bf1abe12608e6c0aceded424b0d40a3e7c990221 :+1:

Sereza7: My bad, it was an artifact left from a previous implementation.
Removed in https://github.com/xwiki/xwiki-platform/pull/3030/commits/bf1abe12608e6c0aceded424b0d40a3e7c990221 :+1:

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/java/org/xwiki/livedata/macro/LiveDataMacroParameters.java
```diff
@@ -158,4 +158,17 @@ public void setDescription(String description)
...
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @since 16.10.3
+     */
+    @Override
+    @Unstable
+    @PropertyDescription("Show or hide the page list that allows the user to switch pages when there is only one page.")
```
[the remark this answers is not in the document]
Sereza7: Added them in https://github.com/xwiki/xwiki-platform/pull/3030/commits/bf1abe12608e6c0aceded424b0d40a3e7c990221 :+1:

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataPaginationConfiguration.java
Sereza7: I added them on all new APIs in https://github.com/xwiki/xwiki-platform/pull/3030/commits/52828d9d4e0f66741d00b6db825d7f711b472625 :+1:

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/LiveDataPaginationConfiguration.java
Sereza7: Updated them in https://github.com/xwiki/xwiki-platform/pull/3030/commits/52828d9d4e0f66741d00b6db825d7f711b472625 :+1:

Sereza7:
@manuelleduc Sorry for the long time with unsolved merge conflicts. 
This PR should be ready for further review or a merge :)

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/java/org/xwiki/livedata/macro/LiveDataMacroParameters.java
```diff
@@ -158,4 +158,17 @@ public void setDescription(String description)
...
+    public void setShowPaginationOnSinglePage(Boolean showPaginationOnSinglePage)
```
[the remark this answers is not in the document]
manuelleduc: I'm not sure of the conclusion of this discussion, do we want to 
1. have it as an advanced macro parameter
2. not as a macro parameter, and rely solely on configuration merging from the macro's body

I lean toward option two, wdyt @michitux ?

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-webjar/src/main/node/src/components/layouts/table/LayoutTable.vue
```diff
@@ -142,6 +142,9 @@ export default {
     canAddEntry() {
       return this.logic.canAddEntry();
     },
+    isMoreThanOnePage () {
+      return this.logic.getPageCount() > 1
```
manuelleduc: Does that mean that pagination is hidden 100% of the time when the page count is 0?
Furthermore, I would move this conditional inside `LivedataPagination` template.
With the current approach, we have to duplicate it for each layout, which makes implementing new layouts more difficult.

Sereza5: > Does that mean that pagination is hidden 100% of the time when the page count is 0?

Sorry, I made a mistake when trying to solve the merge too quickly and added back this code which was wrong. The up-to-date conditional is already in LivedataPagination ^^`
Fixed in https://github.com/xwiki/xwiki-platform/pull/3030/commits/bef787adcf40659829979b13e506b8d501623a6f :+1:

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-api/src/main/java/org/xwiki/livedata/internal/LiveDataRendererParameters.java
```diff
@@ -276,6 +279,30 @@ public void setDescription(String description)
         this.description = description;
     }
 
+    /**
+     * @since 17.10.0RC1
```
manuelleduc: Reminder to merger: update the since ;)

Sereza5: Thank you for the reminder :)
Updated all "since" to 18.3.0RC1 :+1:

xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-macro/src/main/java/org/xwiki/livedata/macro/LiveDataMacroParameters.java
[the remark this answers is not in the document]
michitux: To me, it depends a bit on how well this is supported. If we only support the use case where the total number of entries is small, I think we shouldn't have this as a macro parameter and support it only as configuration metadata in the macro body with some clear documentation that mentions these limits. The way this is phrased currently, as a user I would think I should always enable this as hiding pagination controls when there are no pages improves the usability.

Sereza5:
> Does this correctly handle cases like you're on page 2, apply a filter and number of entries drops below 10? Like that you're not left on an empty page 2 without any way to navigate to page 1?

Filtering resets the offset to 0, so you land back on page 1 and can't end up on an empty page 2. Same when you change the page size. I checked it on a local instance with the document index.

> If we don't have proper UX for such use cases, I think we should not expose this as a macro parameter as users won't understand these limitations and wrongly use this option in cases where it shouldn't be used as it doesn't cover all edge cases.

Fixed in 5ecafb4abf7.
You can still get an out-of-range offset another way: it isn't clamped when the data shrinks under you, and it's restored from the saved config. So rather than drop the parameter, 5ecafb4abf7 also keeps the pagination whenever the current page isn't the first one, which always leaves a page-1 button to click back to. New unit test covers it.

PR #3036, opened 2024-04-02
XWIKI-22036: Add hasNext check before call next.

Xiqinger:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22036

# Changes

## Description

Add hasNext check before call next.

## Clarifications

Difficulty in determining the evenness of the variable 'nodes' and the non-emptiness of 'tables' based on the code.

# Expected merging strategy

* Prefers squash: Yes

### DIFF
--- a/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
+++ b/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
@@ -263,6 +263,9 @@ public class XWikiDocumentArchive
             Collection nodes = archive.getNodes(getWikiReference(), getId());
             for (Iterator it = nodes.iterator(); it.hasNext();) {
                 XWikiRCSNodeInfo nodeInfo = (XWikiRCSNodeInfo) it.next();
+                if (!it.hasNext()) {
+                    throw new IllegalStateException("The number of nodes is an odd number.");
+                }
                 XWikiRCSNodeContent nodeContent = (XWikiRCSNodeContent) it.next();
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/XWikiDocumentArchive.java
```diff
@@ -263,6 +263,9 @@ public void setArchive(String text) throws XWikiException
             Collection nodes = archive.getNodes(getWikiReference(), getId());
             for (Iterator it = nodes.iterator(); it.hasNext();) {
                 XWikiRCSNodeInfo nodeInfo = (XWikiRCSNodeInfo) it.next();
+                if (!it.hasNext()) {
+                    throw new IllegalStateException("The number of nodes is an odd number.");
+                }
```
vmassol, line 268: This change is not correct: as you can see at https://github.com/xwiki/xwiki-platform/blob/5fa50d553ccd8c4f5ac7356fbc249df0add3ae17/xwiki-platform-core/xwiki-platform-oldcore/src/main/java/com/xpn/xwiki/doc/rcs/XWikiRCSArchive.java#L284-L285 there are always a pair number of nodes.

vmassol:
@Xiqinger thanks a lot for you PR. However, I don't think the main change is needed (I don't see how you could have reproduced the problem so I suspect you just tried to fix it based on the code without writing a test for it, is that so or were you able to reproduce the issue?). And the second change is not related to the JIRA issue mentioned for the PR (I'm also not sure if the problem re the table name is real or not).

FTM I'l close the issue but if you can bring arguments for it, we can reopen it. Thanks a lot!

PR #3042, opened 2024-04-04
XWIKI-17966: Mention events should display the name of the wiki where they took place in Notifications

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-17966

# Changes

## Description

* Added a wiki indicator to mention notifications similar to the one found in regular notifications in notification/macros.vm

## Clarifications

* Same template as https://github.com/xwiki/xwiki-platform/blob/69fe80c401fc1f542bda61bd2bdf1621f881e9f6/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/notification/macros.vm#L63-L65 so that we keep consistency with how and when this wiki indication is displayed.
* Instead of putting it after the page name, since there's no page name in the content we instead put it after the rest of the notification content 

# Screenshots & Video

Before the PR, there was no indication about the wiki in mention notifications (see examples on the jira ticket).
Here are two screenshots of my test instance after the PR. On this instance, I used a default user account to ping the Administrator on both the main wiki (prettyName = Home) and a sub wiki (prettyName = subwiki)
The first one shows mention notifications as the Admin would see them when in the main wiki:
17966-afterPR2
The second one shows mention notifications as the Admin would see them when in the sub wiki:
17966-afterPR1

We can see on both of these examples that the wiki name is displayed on mention notifications coming from another wiki than the one currently navigated. This message works whatever the plurality of mentions in the notification is.

# Executed Tests

Manual tests, see screenshots above.
Successfully passed `mvn clean install -f xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications -Pdocker,quality,integration-tests`. Tried to check the docker tests in `xwiki-platform-mentions-test-docker` too, but the build got stuck at `INFO  o.x.t.d.internal.junit5.WARBuilder - Resolving distribution dependencies ...`, from experience it's an issue with our nexus instance.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * 15.10.X, IMO it's a pretty safe change to backport and it's.


Sereza7:
With the current nexus issues, I don't know how I'd try out the changes to the docker tests. Leaving this as a draft until I commit a properly tested improvement to MentionPageTest.

manuelleduc:
> With the current nexus issues, I don't know how I'd try out the changes to the docker tests. Leaving this as a draft until I commit a properly tested improvement to MentionPageTest.

`MentionPageTest` page test is not a docker test, running `mvn clean install -pl :xwiki-platform-mentions-notifications` will work fine. And so will be running the test from your IDE.

Sereza7:
Added a test for the new feature, successfully passed: `mvn clean install -pl :xwiki-platform-mentions-notifications -Pquality,integration-tests,docker`.

xwiki-platform-core/xwiki-platform-mentions/xwiki-platform-mentions-notifications/src/test/java/org/xwiki/mentions/templates/MentionPageTest.java
```diff
@@ -158,4 +158,60 @@ void mentionNotificationTemplate() throws Exception
...
+        compositeEventParams.put(event1, new MentionView()
+                .setLocation("DOCUMENT")
+                .setAuthorURL("/U1")
+                .setDocumentURL("/page1")
+                .setDocument(xWikiDocument1));
+        velocityContext.put("compositeEventParams", compositeEventParams);
+        velocityContext.put("xcontext", this.context);
+        // Template rendering.
+        String actual = templateManager.render("mentions/mention.vm");
+        String expected = IOUtils.toString(getClass().getResourceAsStream("/templates/mentions/mention2.html"),
```
manuelleduc, line 212: Wdyt of giving a slightly more descriptive name to `mention2.html` (e.g., `mentionOtherWiki.html`)?

manuelleduc: Hum actually I'm not even sure comparing to a full html file is the way to go (I know I introduced the first test which was ok-ish to validate the general structure, but I'm not sure this is really the best approach).
What I suggest is to parse the content with jsoup, run selections on the resulting document and assert the existence and/or content of the selections.
I also suggest adding the "wiki" script services to the loaded components, it seems to be missing.
Of course, feel free to ping me if you need help.

Sereza7: > Hum actually I'm not even sure comparing to a full html file is the way to go (I know I introduced the first test which was ok-ish to validate the general structure, but I'm not sure this is really the best approach).
What I suggest is to parse the content with jsoup, run selections on the resulting document and assert the existence and/or content of the selections.

Addressed in https://github.com/xwiki/xwiki-platform/pull/3042/commits/8a9a01966a30a0c64c756373161e8b2d3f2782ed :+1: 

> I also suggest adding the "wiki" script services to the loaded components, it seems to be missing.

I tried it out, and it does not seem like the result rendered by the templateRenderer is impacted by adding the WikiManagerScriptServiceComponentList component list. Since it'd be the only use for it here, I did not add it in the end. 
Let me know if I misunderstood something. 

Thanks!
Lucas.

PR #3056, opened 2024-04-12
XWIKI-21938: Create a description field for page attachments

pjeanjean:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21938

# Changes

## Description

* Add an option `attachment.upload.enableSummaries` to indicate whether or not attachments should have specific change summaries set during upload, and displayed in the history.
* With the option enabled:
  * Add a text input in the attachments extra footer to set a change summary. In the case of a multi-file upload, the same summary is applied to all the new attachments;
  * Display the summaries for each revision in the attachment history.
* Add a UI test "AttachmentSummaryIT" checking that setting and displaying attachment summaries work as expected.

## Clarifications

https://forum.xwiki.org/t/create-a-description-field-for-page-attachments/14346

# Screenshots & Video

image
image
image

# Executed Tests

The behavior was tested both manually and through the new UI test.

# Expected merging strategy

* Prefers squash: Yes


xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-test/xwiki-platform-attachment-test-docker/src/test/it/org/xwiki/attachment/test/ui/docker/AttachmentSummaryIT.java
```diff
@@ -0,0 +1,90 @@
...
+})
+class AttachmentSummaryIT
+{
+    private static final String ATTACHMENT_FILENAME = "moveme.txt";
+
+    @Test
+    @Order(1)
+    void checkSummary(TestUtils setup, TestReference testReference, TestConfiguration testConfiguration)
+    {
+        DocumentReference attachmentsPage = new DocumentReference("Attachments", testReference.getLastSpaceReference());
```
vmassol, line 64: why not use `testReference` directly for the page containing the attachments?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-test/xwiki-platform-attachment-test-docker/src/test/it/org/xwiki/attachment/test/ui/docker/AttachmentSummaryIT.java
```diff
@@ -0,0 +1,90 @@
...
+
+        setup.loginAsSuperAdmin();
+        setup.deletePage(attachmentsPage);
+
+        setup.createPage(attachmentsPage, "");
+
+        String attachmentPath = new File(new File(testConfiguration.getBrowser().getTestResourcesPath(), "v0"),
+            ATTACHMENT_FILENAME).getAbsolutePath();
+
+        setup.gotoPage(attachmentsPage);
```
vmassol, line 74: I think `createPage()` returns a ViewPage already, so you're on the new page and no need to navigate

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationConfiguration.java
```diff
@@ -71,4 +71,10 @@ public interface AttachmentValidationConfiguration
      * @since 14.10.13
      */
     long getMaxUploadSize(EntityReference entityReference);
+
+    /**
+     * @return whether or not summaries for attachments should be settable and displayed from UI
+     * @since 16.3.0RC1
+     */
+    boolean areSummariesEnabled();
```
vmassol, line 79: Why are summaries config related to attachment validation? What could be related to validation is if we have a config saying that a summary is mandatory for accepting a new attachement but this is not the case here AFAICS.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1120,6 +1121,7 @@ core.viewers.attachments.showHistory=View attachment history
 core.viewers.attachments.author=Posted by {0}
 core.viewers.attachments.date=on {0}
 core.viewers.attachments.noAttachments=No attachments for this page
+core.viewers.attachments.upload.comment=Add summary...
```
vmassol, line 1124: comment or summary?

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
```diff
@@ -1105,6 +1105,7 @@ core.viewers.attachments.delete.done=Attachment deleted
 core.viewers.attachments.delete.failed=Failed to delete attachment:
 core.viewers.attachments.livetable.actions=Actions
 core.viewers.attachments.livetable.author=Posted by
+core.viewers.attachments.livetable.comment=Summary
```
vmassol, line 1108: comment or summary?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationConfiguration.java
pjeanjean: Creating a completely new script service just for that setting felt wrong somehow (so I went with `AttachmentValidationScriptService`), but I agree that the end result doesn't make a lot of sense semantically.
I guess a better option would be to find a more fitting configuration service to extend, but I'm not sure which one.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
pjeanjean: This is confusing, but since I use the field `comment`, and also existing code that explicitly mentions it this way, it's a bit hard to tell which one makes the most sense as a label. I guess this one could be summary, but `core.viewers.attachments.livetable.comment` very clearly references the comment field of an attachment.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationConfiguration.java
vmassol: It's fine to add missing classes (there was no need before). So right now we need a new configuration class for attachments AFAICS and a way to return that config from a new script service (might be better to return the full config class than duplicate the config methods, unless some methods are supposed to be not visible from a SS).

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
vmassol: Might be good to do a quick poll to the devs on the matrix chat for the terminology.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-validation/xwiki-platform-attachment-validation-api/src/main/java/org/xwiki/attachment/validation/AttachmentValidationConfiguration.java
tmortagne: +1 for moving that elsewhere

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-pageobjects/src/main/java/org/xwiki/flamingo/skin/test/po/AttachmentsPane.java
```diff
@@ -124,6 +124,20 @@ public void setFileToUpload(final String filePath)
         input.sendKeys(filePath);
     }
 
+    /**
+     * Fills the upload summary.
+     *
+     * @param summary the upload summary.
+     */
```
tmortagne, line 131: A `@since` would be nice.

xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/AttachmentHistoryPage.java
```diff
@@ -79,6 +81,15 @@ public String getAuthor(int rowIdx)
         return getCell(rowIdx, AUTHOR_COL_IDX).getText();
     }
 
+    /**
+     * @param rowIdx the index of the row of the attachment, starting at 1
+     * @return the summary of the requested attachment version
+     */
```
tmortagne, line 87: A `@since` would be nice.

xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties
tmortagne: I would simply follow the same as what we do on document side for the same need: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-oldcore/src/main/resources/ApplicationResources.properties#L940-L966

vmassol: What happened in the past is that we called it "comment" in the DB and in the code, hence the translation keys with "comments", the apis (e.g. `$xwiki.hasEditComment()`, `isEditCommentFieldHidden`) and configuration keys.  It was implemented in https://jira.xwiki.org/browse/XWIKI-114

Then in https://jira.xwiki.org/browse/XWIKI-5526 it was decided to rename "Comments" in "Summary" in the UI (too bad we can't see http://markmail.org/thread/y4gnmnwfdglpygkj anymore).

So, if we want to do the same, then we should keep the concept of "Comments" in the code and in the APIs/configuration and only use Summary in the UI translation keys.

vmassol:
@pjeanjean Maybe I've missed it but I haven't seen an alt text when hovering over the comments field to explain what it is to the user. I think that would help as "Add summary..." is not very explicit.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+public class DefaultAttachmentConfiguration implements AttachmentConfiguration
+{
+    private static final String ATTACHMENT_MIMETYPE_ENABLE_SUMMARIES_PROPERTY = "attachment.upload.enableSummaries";
+
+    @Inject
+    @Named("xwikiproperties")
+    private ConfigurationSource xWikiPropertiesConfigurationSource;
+
+    @Override
+    public boolean areSummariesEnabled()
```
michitux, line 47: Boolean getters should start with `is`, so maybe `isSummaryEnabled()`? Then you can also use `.summariesEnabled` on the Velocity side.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/configuration/AttachmentConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.xwiki.component.annotation.Role;
+
+/**
+ * General configuration values for attachments.
+ *
+ * @version $Id$
+ * @since 16.3.0RC1
+ */
+@Role
```
michitux, line 30: Missing `@Unstable`.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/script/AttachmentScriptService.java
```diff
@@ -156,4 +160,15 @@ public long backlinksCount(AttachmentReference attachmentReference) throws Attac
...
         }
     }
+
+    /**
+     * Gets the current configuration for attachments.
+     *
+     * @return the current configuration for attachments
+     * @since 16.3.0RC1
+     */
+    public AttachmentConfiguration getConfiguration()
```
michitux, line 170: Missing `@Unstable`.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/test/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfigurationTest.java
```diff
@@ -0,0 +1,62 @@
...
+    private DefaultAttachmentConfiguration attachmentConfiguration;
+
+    @MockComponent
+    @Named("xwikiproperties")
+    private ConfigurationSource xWikiPropertiesConfigurationSource;
+
+    private static final String ATTACHMENT_MIMETYPE_ENABLE_SUMMARIES_PROPERTY = "attachment.upload.enableSummaries";
+
+    @Test
+    void areSummariesEnabledInPropertiesConfigurationSource()
```
michitux, line 53: This could benefit from a [parameterized test](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) that checks both true and false.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-test/xwiki-platform-attachment-test-docker/src/test/it/org/xwiki/attachment/test/ui/docker/AttachmentSummaryIT.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.xwiki.attachment.internal.RedirectAttachmentClassDocumentInitializer.REFERENCE;
+
+/**
+ * Tests for setting and displaying attachment summaries.
+ *
+ * @version $Id$
+ * @since 16.3.0RC1
+ */
+@UITest(properties = {
+    "xwikiPropertiesAdditionalProperties=attachment.upload.enableSummaries = true"
```
michitux, line 52: Note that this enables summarizes for all tests in this module, which means that there are probably not many tests anymore where attachment summaries are disabled. Did you verify that there is indeed a test where attachment summaries are disabled to verify that there is no regression?

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsinline.vm
```diff
@@ -45,7 +45,13 @@ $xwiki.ssfx.use('js/xwiki/viewers/attachments.css', true)
     ## Temporarily disabled, until we fix attachment name handling
     ## <div><label id="xwikiuploadnamelabel" for="xwikiuploadname">$services.localization.render('core.viewers.attachments.upload.filename')</label></div>
     ## <div><input id="xwikiuploadname" type="hidden" name="filename" value="" size="40"/></div>
-    <div class="fileupload-field"><label class="sr-only" for="xwikiuploadfile">$services.localization.render('core.viewers.attachments.upload.file')</label><input id="xwikiuploadfile" type="file" name="filepath" size="40" class="uploadFileInput noitems" data-max-file-size="$!escapetool.xml($xwiki.getSpacePreference('upload_maxsize'))" /></div>
+    <div class="fileupload-field">
+      #if ($services.attachment.configuration.areSummariesEnabled())
+        <label for="xwikiuploadcomment" class="sr-only">$services.localization.render('core.viewers.attachments.upload.comment')</label>
```
michitux, line 50: Why is this label hidden?

Wouldn't it make sense to have a summary per uploaded file instead of one for all files?

xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/widgets/upload.js
```diff
@@ -202,6 +202,11 @@ var XWiki = (function(XWiki) {
         fields[key] && formData.append(key, fields[key]);
       });
 
+      if (this.formData.comment) {
+        var commentValue = this.formData.comment.value;
```
michitux, line 206:
```suggestion
        const commentValue = this.formData.comment.value;
```

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsinline.vm
```diff
@@ -45,7 +45,13 @@ $xwiki.ssfx.use('js/xwiki/viewers/attachments.css', true)
...
+        <input type="text" id="xwikiuploadcomment" class="uploadCommentInput" name="comment" placeholder="$services.localization.render('core.viewers.attachments.upload.comment')" title="$services.localization.render('core.comment.tooltip')" maxlength="1023" />
```
michitux, line 51: Placeholders must not be used as replacement for a label. See also [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder):

> Effective placeholder text includes a word or short phrase that hints at the expected data type, not an explanation or prompt. The placeholder must not be used instead of a [`<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label). As the placeholder is not visible if the value of the form control is not null, using placeholder instead of a `<label>` for a prompt harms usability and accessibility.

Instead, the placeholder could provide an example for the data that should be entered.

Further, I don't understand why you're using `core.comment.tooltip` as tooltip. "Enter a brief description of your changes" doesn't make sense here.

michitux (changes requested):
In general, this looks quite straightforward. However, I'm really wondering why we're hiding attachment summaries behind a configuration property. This has the effect of cluttering the code base with `if`s that increase the complexity for everybody (developers, administrators, users, support, …). We have a dedicated mechanism for implementing features that shouldn't be there by default: extensions.

Apart from that, I have a few small comments.

vmassol:
> In general, this looks quite straightforward. However, I'm really wondering why we're hiding attachment summaries behind a configuration property. This has the effect of cluttering the code base with ifs that increase the complexity for everybody (developers, administrators, users, support, …). We have a dedicated mechanism for implementing features that shouldn't be there by default: extensions.

@michitux This would be an interesting discussion to have on the forum. Some quick thoughts (to be discussed on the forum):
* If you do it with extension instead of `IF` then you need to add a UIXP which has the complexity of an `IF` (and makes it hard to implement another skin).
* An extension is a lot more complex for the end user: they need to know that a feature they need is inside an extension and which one, then install it. It's also harder to disable a feature since they need to understand the same and remove that extension with the extension manager (hoping, it won't also remove other dependent extensions that you need).

I don't think it's the right technical choice to have extensions for small variations in features. For a full feature it makes sense, but for sub-feature it feels too coarse-grained and IMO has too many drawbacks.

I think it's interesting to discuss it on the forum because we might not all agree and it would be interesting to define some best practices as to when to use an extension and UIXPs and when to use a configuration option.

pjeanjean:
> @pjeanjean Maybe I've missed it but I haven't seen an alt text when hovering over the comments field to explain what it is to the user. I think that would help as "Add summary..." is not very explicit, ie. explain that it's a message/comment explaining the reason for the attachment upload, and visible in the attachment history.

That's why the `title` is for on https://github.com/xwiki/xwiki-platform/pull/3056/files#diff-77ae3ef018401a68554439f6646326dd593f570246a90f490dc806a7c26a8870R51
But the current tooltip is not explicit enough so I'll replace it.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/test/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfigurationTest.java
```diff
@@ -0,0 +1,64 @@
...
+    private DefaultAttachmentConfiguration attachmentConfiguration;
+
+    @MockComponent
+    @Named("xwikiproperties")
+    private ConfigurationSource xWikiPropertiesConfigurationSource;
+
+    private static final String ATTACHMENT_MIMETYPE_ENABLE_COMMENTS_PROPERTY = "attachment.upload.enableComments";
+
+    @ParameterizedTest
+    @CsvSource({"true", "false"})
```
michitux, line 54: `@ValueSource(booleans = { true, false })` seems simpler to me.

pjeanjean:
I fixed most of the comments, and tried to improve the UX a bit with better localization strings. Right now it looks like this 
image
WDYT?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-test/xwiki-platform-attachment-test-docker/src/test/it/org/xwiki/attachment/test/ui/docker/AttachmentSummaryIT.java
pjeanjean: I did not know that, but this module only had a single test before (for moving attachments) so it should be fine. The other "basic" UI tests are in `flamingo` and `web`.

michitux:
I really find it strange to have this input without visible label. This is bad for accessibility and usability. Mobile users for example won't really be able to see the title and the placeholder is gone once you start typing. Accessibility guidelines like [this one](https://www.w3.org/WAI/tutorials/forms/labels/) clearly state that a visible label is always the best option and I really don't see why we can't use one here. Further, this should follow the [form standard of XWiki](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/VerticalForms/).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-test/xwiki-platform-attachment-test-docker/src/test/it/org/xwiki/attachment/test/ui/docker/AttachmentSummaryIT.java
michitux: Okay, that's good then.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/test/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfigurationTest.java
vmassol: As a general rule I'm not very fond of parameterized tests for functional tests since func tests are supposed to be scenarios and there's often some running duplications with parameterized tests (to reset the initial conditions). I'm not saying it's bad here, I haven't checked. But you need to keep in mind that we want the minimal number of functional tests needed (if some tests can be written as unit tests or page tests, it's better), and to be very conscious about the time it takes to execute them (our build is already very long and running on all supported configurations make it even slower). So when you make a parameterized test with N options, for example you add N times the number of environments.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsinline.vm
pjeanjean: I only added this label for accessibility purposes, the input itself should hopefully be self-explanatory through the placeholder and the tooltip (granted, it was not in the previous version).

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/test/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfigurationTest.java
michitux: This is not a functional test, this is a very simple unit test of a single component.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/attachmentsinline.vm
michitux: As I wrote, placeholders and titles are no replacement for a visible label unless the label is obvious (like a file picker with an upload button). Further, we have standards for forms in XWiki that clearly include a label.

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/test/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfigurationTest.java
vmassol: oops my bad then, that'll teach me to speak without re-checking first ;) (I had checked a test earlier and it was a functional tests, but I didn't recheck). Sorry guys (but what I mentioned is still valid and there are plenty of existing functional tests that take too much time)

pjeanjean:
Following the form standard, we end up with:
image

I think this is a big change, and I feel that this looks slightly out of place right now and would require a larger redesign of the attachments tab.

Note: this version also seems to break the upload script and I haven't investigated the reason yet.

michitux:
@tkrieck as you already commented on the forum post, maybe you could offer some inputs on the visual aspects of the input field? My opinion is that we should have a proper, visible label for the input, but I agree also that it looks a bit strange.

tkrieck:
> @tkrieck as you already commented on the forum post, maybe you could offer some inputs on the visual aspects of the input field? My opinion is that we should have a proper, visible label for the input, but I agree also that it looks a bit strange.

The overall look of the labels are a bit heavy in my opinion, but this about XS standards and should be in another discussion.

The main thing for me here is that IMO the label+hint are a little bit over descriptive now. It takes time from the user to read and understand all that is involved. Maybe we could try and reduce it to something like:

**UPLOAD SUMMARY**
This description will be shown in the attachment's history.

vmassol:
> This description will be shown in the attachment's history.

@tkrieck I feel it's less good than the proposed description above . It doesn't help to know what to put and it's too indirect IMO. I feel it's better to explain what it's about rather than where it's going to be displayed (or use both, which is even more clear).

Now the proposed description could be shortened a bit by removing "Enter a", "which will be" -> "Description of this upload (shown in the attachment history)". "Upload Summary" feels ok and clear enough.

I also don't feel that the real issue at hand is a question of length of text but more of the overall UI. For example "Attach files to this page" is displayed in a font smaller than the "ATTACHMENT UPLOAD SUMMARY" which is also in bold and more visible. That looks odd.

pjeanjean:
Updated to include the latest suggestions:
image

michitux:
What I just noticed is that we actually have [CSS](https://github.com/xwiki/xwiki-platform/blob/08190dd323118e16d833465ba343a4f5732ef9c1/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/js/xwiki/viewers/attachments.css#L30-L32) to decrease the size of the "Attach files to this page". If we remove that CSS, the heading is much bigger. Maybe this would be an option?

pjeanjean:
For reference, the output with the rule disabled:
image
And the same thing with "1.3em" which I think might be enough:
image

tkrieck:
> For reference, the output with the rule disabled:

I would go with the rule disabled, just to keep it standard and remove a specific style. But nothing wrong with 1.3em.

The label and description are also much better. All in all, looks ok to me. 👍

vmassol:
Can you also show a screenshot with the attachment comment feature disabled (since this is what we'll have most of the time)? The displayed UI needs to be better than what we have now as otherwise it'd be a regression (even if it's an improvement when the feature is enabled).

pjeanjean:
Good point:
image
image

I think it works in both cases with `1.3em`, but the default is way too large with the feature disabled.

michitux:
Then let's take `1.3em` and merge this?

xwiki-platform-core/xwiki-platform-attachment/xwiki-platform-attachment-api/src/main/java/org/xwiki/attachment/internal/configuration/DefaultAttachmentConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+@Component
+@Singleton
+public class DefaultAttachmentConfiguration implements AttachmentConfiguration
+{
+    private static final String ATTACHMENT_MIMETYPE_ENABLE_COMMENTS_PROPERTY = "attachment.upload.enableComments";
+
+    @Inject
+    @Named("xwikiproperties")
+    private ConfigurationSource xWikiPropertiesConfigurationSource;
```
vmassol, line 44: small typo, it should be `xwikiPropertiesConfigurationSource`.

pjeanjean:
Should be good now.

michitux (approved):

PR #3067, opened 2024-04-15
XWIKI-22073: Image specific syntax with data-widget causes CKEditor to freeze

manuelleduc:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22073

# Changes

## Description

* introduce a  `toHtml` event handler removing `data-widget` attributes from images

# Executed Tests

```
mvn clean install \
  -Pquality,integration-tests,docker \
  -pl :xwiki-platform-ckeditor-plugins,:xwiki-platform-ckeditor-test-docker
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x
  * stable-14.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
@@ -40,35 +40,15 @@
     __namespace: true
   };
 
-  /**
-   * Remove data-widget attributes on image as they can lead to CKEditor crashing.
-   * @param event a toHtml event
-   */
-  function cleanupImages(event) {
-    const domParser = new DOMParser();
-    let html = event.data.dataValue;
-    // Depending on the editor type (inline or iframe-based) and configuration (fullPage true or false) the HTML
-    // string can be a full HTML document or just a fragment (e.g. the content of the BODY tag).
-    const isFragment = !html.trimEnd().endsWith('</html>');
-    try {
-      const doc = domParser.parseFromString(html, 'text/html');
-      // We want to modify the input HTML string only if there is a style tag that need escaping.
-      let modified = false;
-      doc.querySelectorAll('img').forEach(img => {
-        const dataWidgetAttribute = 'data-widget';
-        const hasDataWidget = img.hasAttribute(dataWidgetAttribute);
-        if(hasDataWidget) {
-          img.removeAttribute(dataWidgetAttribute);
-        }
-        modified = modified || hasDataWidget;
-      });
-      if (modified) {
-        event.data.dataValue = isFragment ? doc.body.innerHTML : doc.documentElement.outerHTML;
+  const removeDataWidgetAttributeFromImage = {
+    elements: {
+      img: function (element) {
+        console.log('removeDataWidgetAttributeFromImage', element);
+        delete element.attributes['data-widget'];
+        return element;
       }
-    } catch (e) {
-      console.warn('Failed to cleanup the image attributes from the given HTML string: ' + html, e);
     }
-  }
+  };
 
   CKEDITOR.plugins.add('xwiki-filter', {
     init: function(editor) {
@@ -180,6 +160,7 @@
       // Filter the editor input.
       var dataFilter = editor.dataProcessor && editor.dataProcessor.dataFilter;
       if (dataFilter) {
+        dataFilter.addRules(removeDataWidgetAttributeFromImage, {priority: 5});
         dataFilter.addRules(replaceEmptyLinesWithEmptyParagraphs, {priority: 5});
         if (editor.config.loadJavaScriptSkinExtensions) {
           dataFilter.addRules(unprotectAllowedScripts, {priority: 5});
@@ -197,7 +178,6 @@
       // parser is called in both cases. Priority 1 is needed to ensure our listener is called before the HTML parser
       // (which has priority 5).
       editor.on('toHtml', escapeStyleContent, null, null, 1);
-      editor.on('toHtml', cleanupImages, null, null, 1);
       editor.on('toDataFormat', escapeStyleContent, null, null, 1);
 
       // Transform <font color="..." face="..."> into <span style="color: ...; font-family: ...">.
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
@@ -40,16 +40,6 @@
     __namespace: true
   };
 
-  const removeDataWidgetAttributeFromImage = {
-    elements: {
-      img: function (element) {
-        console.log('removeDataWidgetAttributeFromImage', element);
-        delete element.attributes['data-widget'];
-        return element;
-      }
-    }
-  };
-
   CKEDITOR.plugins.add('xwiki-filter', {
     init: function(editor) {
       var replaceEmptyLinesWithEmptyParagraphs = {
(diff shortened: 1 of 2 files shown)


xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
```diff
@@ -167,6 +197,7 @@
       // parser is called in both cases. Priority 1 is needed to ensure our listener is called before the HTML parser
       // (which has priority 5).
       editor.on('toHtml', escapeStyleContent, null, null, 1);
+      editor.on('toHtml', cleanupImages, null, null, 1);
```
manuelleduc, line 200: I couldn't make the "Filter the editor output." `htmlFilter` above to work, so I'm sharing what I got so far:
1. the editor is not freezing anymore
2. `data-widget` are effectively removed from images (and I hope the removal is not too aggressive), and as you can see in the docker test, is not persisted after save

Part of `cleanupImages` could be factored with `escapeStyleContent` if we agree that this is the right approach.

mflorea: Have you tried to use the ``dataFilter`` to filter the editor **input**? If the editor is freezing when loading then I think you need to worry first about filtering the input HTML, and then the output HTML.

manuelleduc: Yes, I tried with the patch below, the `img` function is never called.

The reproduction steps are:
1. replace the content of `Main.WebHome` with `[[image:image.gif||data-widget='uploadimage']]` using the wiki editor
2. go back to `Main.WebHome` in view mode, then edit in-place
3. I'm expecting the `img` function to be called, but it is not

```
Index: xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js
--- a/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js	(revision 2aef8e2f9925fdb5bd4cd1b5fba207f5f0222ea3)
+++ b/xwiki-platform-core/xwiki-platform-ckeditor/xwiki-platform-ckeditor-plugins/src/main/webjar/xwiki-filter/plugin.js	(date 1713250081873)
@@ -40,35 +40,15 @@
     __namespace: true
   };
 
-  /**
-   * Remove data-widget attributes on image as they can lead to CKEditor crashing.
-   * @param event a toHtml event
-   */
-  function cleanupImages(event) {
-    const domParser = new DOMParser();
-    let html = event.data.dataValue;
-    // Depending on the editor type (inline or iframe-based) and configuration (fullPage true or false) the HTML
-    // string can be a full HTML document or just a fragment (e.g. the content of the BODY tag).
-    const isFragment = !html.trimEnd().endsWith('</html>');
-    try {
-      const doc = domParser.parseFromString(html, 'text/html');
-      // We want to modify the input HTML string only if there is a style tag that need escaping.
-      let modified = false;
-      doc.querySelectorAll('img').forEach(img => {
-        const dataWidgetAttribute = 'data-widget';
-        const hasDataWidget = img.hasAttribute(dataWidgetAttribute);
-        if(hasDataWidget) {
-          img.removeAttribute(dataWidgetAttribute);
-        }
-        modified = modified || hasDataWidget;
-      });
-      if (modified) {
-        event.data.dataValue = isFragment ? doc.body.innerHTML : doc.documentElement.outerHTML;
+  const removeDataWidgetAttributeFromImage = {
+    elements: {
+      img: function (element) {
+        console.log('removeDataWidgetAttributeFromImage', element);
+        delete element.attributes['data-widget'];
+        return element;
       }
-    } catch (e) {
-      console.warn('Failed to cleanup the image attributes from the given HTML string: ' + html, e);
     }
-  }
+  };
 
   CKEDITOR.plugins.add('xwiki-filter', {
     init: function(editor) {
@@ -180,6 +160,7 @@
       // Filter the editor input.
       var dataFilter = editor.dataProcessor && editor.dataProcessor.dataFilter;
       if (dataFilter) {
+        dataFilter.addRules(removeDataWidgetAttributeFromImage, {priority: 5});
         dataFilter.addRules(replaceEmptyLinesWithEmptyParagraphs, {priority: 5});
         if (editor.config.loadJavaScriptSkinExtensions) {
           dataFilter.addRules(unprotectAllowedScripts, {priority: 5});
@@ -197,7 +178,6 @@
       // parser is called in both cases. Priority 1 is needed to ensure our listener is called before the HTML parser
       // (which has priority 5).
       editor.on('toHtml', escapeStyleContent, null, null, 1);
-      editor.on('toHtml', cleanupImages, null, null, 1);
       editor.on('toDataFormat', escapeStyleContent, null, null, 1);
 
       // Transform <font color="..." face="..."> into <span style="color: ...; font-family: ...">.
```

mflorea: When the editor is loaded the editor data (HTML) is set (e.g. based on the value of the textarea element that is replaced). This data is:

* first parsed and filtered [using the data processor](https://github.com/ckeditor/ckeditor4/blob/master/plugins/wysiwygarea/plugin.js#L468)
  * filtering is done in [multiple stages](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-toHtml):
  * widgets are **upcasted** (but not yet initialized!) with [priority 8](https://github.com/ckeditor/ckeditor4/blob/master/plugins/widget/plugin.js#L3145)
  * ``dataProcessor.dataFilter`` (i.e. your filter) is called [with priority 10](https://github.com/ckeditor/ckeditor4/blob/master/core/htmldataprocessor.js#L153)
* then ["written" (appended) to the editable area](https://github.com/ckeditor/ckeditor4/blob/master/plugins/wysiwygarea/plugin.js#L585) -> this triggers the initialization of all the widgets found in the content -> the failure happens when the ``uploadimage`` widget is initialized

The problem is that before your filter is called the widgets are upcasted which means the image is wrapped with a span that has ``contenteditable=false``. Before ``dataProcessor.dataFilter`` applies a filter rules it [checks that rule can be applied](https://github.com/ckeditor/ckeditor4/blob/master/core/htmlparser/filter.js#L392) in the current "context". Your rule is not applied because the context has the ``nonEditable`` flag set (because of the parent span element that has ``contenteditable=false``). You can force this though using ``applyToAll``:

    editor.dataProcessor.dataFilter.addRules(removeDataWidgetAttributeFromImage, {priority: 5, applyToAll: true});

With this your filter is now applied, but it doesn't fix the problem because the widgets are already upcasted when your filter is called, and this image is upcasted as an "uploadimage" not "image" (because ``data-widget="uploadimage"``) and the initialization of the ``uploadimage`` widget fails because the HTML structure is not the expected one.

The actual fix is to filter the attribute before widgets are upcasted, i.e. to use a ``toHtml`` listener with priority less than 8. In your code you use 1, but this is too early, when the HTML is in string format. It's easier to filter after the HTML is parsed:

> 5-9: Data is available in the parsed format, but CKEDITOR.htmlDataProcessor.dataFilter is not applied yet.

```
editor.on('toHtml', (event) => {
  // CKEDITOR.htmlParser.fragment instance
  event.data.dataValue.filter(new CKEDITOR.htmlParser.filter({
    elements: {
      img: function (element) {
        delete element.attributes['data-widget'];
      }
    }
  }));
// We must use a priority below 8 to make sure our filter is executed before widgets are upcasted.
}, null, null, 7);
```

mflorea (approved):

PR #3072, opened 2024-04-18
[Misc] Expose test alert problem with firefox

surli:
# Jira URL

N/A

# Changes

## Description

* This PR only intends at exposing a bug with selenium firefox, it should not be merged.


surli:
I can't reproduce anymore the failure: the test is passing fine now it seems.

PR #3073, opened 2024-04-18
XWIKI-22108: Extra empty paragraph and spacing in the information panel in edit mode

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22108

Related to the failure https://ci.xwiki.org/job/XWiki%20Environment%20Tests/job/xwiki-platform/job/master/526/testReport/org.xwiki.flamingo.test.docker/AllIT/MySQL_8_3__Tomcat_9_jdk21__Chrome___Docker_tests_for_xwiki_platform_flamingo_skin___Build_for_MySQL_8_3__Tomcat_9_jdk21__Chrome___Docker_tests_for_xwiki_platform_flamingo_skin______/ on CI.
# Changes

## Description

* Added `wiki="true"` to the html macro so that we can use it right under the velocity macro. Therefore all the velocity content is already encapsulated in some blocks (the ones created in the HTML macro ) and velocity does not need to wrap it up any more.

# Screenshots & Video

Before the PR:
Before PR
After the PR:
After PR
We can see that unlike before the PR, the dt and dd are not wrapped in a div and siblings of empty `<p>`s.

# Executed Tests

Manual tests :ok_hand: 
Successfully built changes with `mvn clean install -f xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui -Pquality,integration-tests`. And successfully passed the test that failed on CI with `mvn clean install -f xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-test/xwiki-platform-flamingo-skin-test-docker -Pdocker -Dit.test=SectionEditIT#sectionEditInWikiEditorWhenSyntax2x -Dxwiki.test.ui.wcag=true`.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None, cannot reproduce this change on the latest 15.10.X snapshot

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
@@ -213,7 +213,7 @@
       &lt;dt&gt;$escapetool.html($services.localization.render('panels.documentInformation.includesCount', [$pages.size()]))&lt;/dt&gt;
       #foreach ($page in $pages)
         #set ($alt = $escapetool.xml($services.localization.render('panels.documentInformation.editIncluded', [$page])))
-        &lt;dd&gt;&lt;a href="$xwiki.getURL($page, 'view')"&gt;$escapetool.xml($page)&lt;/a&gt; &lt;a href="$xwiki.getURL($page, 'edit')"&gt;$services.icon.renderHTML('edit')&lt;span class="sr-only"&gt;$alt&lt;/span&gt;&lt;/a&gt;&lt;/dd&gt;
+        &lt;dd&gt;&lt;span class="wikilink"&gt;&lt;a href="$xwiki.getURL($page, 'view')"&gt;$escapetool.xml($page)&lt;/a&gt;&lt;/span&gt; &lt;span class="wikiinternallink"&gt;&lt;a href="$xwiki.getURL($page, 'edit')"&gt;$services.icon.renderHTML('edit')&lt;span class="sr-only"&gt;$alt&lt;/span&gt;&lt;/a&gt;&lt;/span&gt;&lt;/dd&gt;
       #end
     #end
   {{/html}}
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
@@ -210,11 +210,11 @@
   {{html clean="false" wiki="true"}}
     #set ($pages = $tdoc.includedPages)
     #if($pages.size() != 0)
-      &lt;dt&gt;$services.localization.render('panels.documentInformation.includesCount', [$pages.size()])&lt;/dt&gt;
+      &lt;dt&gt;$escapetool.html($services.localization.render('panels.documentInformation.includesCount', [$pages.size()]))&lt;/dt&gt;
       #foreach ($page in $pages)
-        #set ($pageLink = $services.rendering.escape($page, 'xwiki/2.1'))
-        #set ($pageName = $services.rendering.escape($services.rendering.escape($page, 'xwiki/2.1'), 'xwiki/2.1'))
-        #set ($alt = $services.rendering.escape($services.localization.render('panels.documentInformation.editIncluded', [$page]), 'xwiki/2.1'))
+        #set ($pageLink = $escapetool.html($services.rendering.escape($page, 'xwiki/2.1')))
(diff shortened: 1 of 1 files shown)


michitux:
I think this PR introduces an XSS vulnerability as the page name will be interpreted as HTML. In HTML macros with `wiki="true"`, escaping needs to also escape for HTML.

Sereza7:
Addressed in https://github.com/xwiki/xwiki-platform/pull/3073/commits/127423681edb802bcfd6ab2d4afe6b2d6ef9c377 :+1:

michitux:
`$pageName` is still vulnerable. I'm not sure though, maybe I'm also wrong, in general the HTML macro with `wiki="true"` should be avoided because of such problems. If possible, I would try to find another fix that doesn't involve the `wiki="true"`.

michitux:
I wonder if it wouldn't be cleaner to just use plain HTML links and the icon script service to produce the icon HTML (which would then also follow the icon theme).

manuelleduc:
We do we still need this PR now that XCOMMONS-2276 was fixed (by https://github.com/xwiki/xwiki-commons/commit/fd730e0a40d677788fd24dd14392598eb3ea12ed)?

Sereza7:
See [the Jira issue](https://jira.xwiki.org/browse/XWIKI-22108) Michael created, this is a regression and should be handled before the release..

vmassol:
Note that plain HTML links is a bit dangerous and not the best because it means they won't have the proper classes as generated by the rendering. Ofc you could add these classes manually but then that's code duplication and whenever we will change what the rendering generates we will also need to update these manual links...

Sereza7:
I updated with plain HTML links.

This is what it looks like in the latest version: 
image

I made sure the escaping changes did not break the hrefs and text content of the links.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
```diff
@@ -207,15 +207,14 @@
...
     #if($pages.size() != 0)
       &lt;dt&gt;$escapetool.html($services.localization.render('panels.documentInformation.includesCount', [$pages.size()]))&lt;/dt&gt;
       #foreach ($page in $pages)
-        #set ($pageLink = $escapetool.html($services.rendering.escape($page, 'xwiki/2.1')))
-        #set ($pageName = $services.rendering.escape($pageLink, 'xwiki/2.1'))
-        #set ($alt = $escapetool.html($services.rendering.escape($services.localization.render('panels.documentInformation.editIncluded', [$page]), 'xwiki/2.1')))
-        &lt;dd&gt;[[$pageName&gt;&gt;$pageLink]] [[[[image:icon:page_white_edit||alt="$alt"]]&gt;&gt;path:$xwiki.getURL($page, 'edit')]]&lt;/dd&gt;
+        #set ($pageRef = $escapetool.html($page))
+        #set ($alt = $escapetool.html($services.localization.render('panels.documentInformation.editIncluded', [$page])))
+        &lt;dd&gt;&lt;a href="$xwiki.getURL($pageRef, 'view')"&gt;$pageRef&lt;/a&gt; &lt;a href="$xwiki.getURL($pageRef, 'edit')"&gt;$services.icon.renderHTML('edit')&lt;span class="sr-only"&gt;$alt&lt;/span&gt;&lt;/a&gt;&lt;/dd&gt;
```
michitux, line 217: Passing escaped references to `getURL` doesn't work. Also, we normally use `$escapetool.xml`, not `$escapetool.xml`. See also https://jira.xwiki.org/browse/XCOMMONS-2828 and the surrounding discussion.

vmassol:
> I updated with plain HTML links.
> 
> This is what it looks like in the latest version: image
> 
> I made sure the escaping changes did not break the hrefs and text content of the links. Note that there is no `<p>` before the `<dl>`

See comment at https://github.com/xwiki/xwiki-platform/pull/3073#issuecomment-2066448066

I see for example that you don't have the `<span class="wikilink">` around the A tag that we get when using the rendering. Which means that if we have CSS or javascript code that do something, it won't apply here.

xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3073/commits/421738ba5320f4bea951c056442850a3f830d061 :+1:

vmassol:
> Addressed this specific issue in https://github.com/xwiki/xwiki-platform/commit/5dd78a70482a0561fd411e19a7c74f7f2049c04f

Note that you haven't addressed my comment about the issue of using HTML links directly. See https://github.com/xwiki/xwiki-platform/pull/3073#issuecomment-2066448066

Sereza7:
See https://github.com/xwiki/xwiki-platform/pull/3078 for a more robust fix :)

PR #3076, opened 2024-04-19
XWIKI-22082: Some LiveData tables flicker on page refresh

manuelleduc:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22082

# Changes

## Description

* always add enough padding on the selectize fields to avoid a change of width on load

# Screenshots & Video

**Limitation**

Empty
image

With content
image

Vertical space is always reserved for the spinner, even when the field is not loading.
As a consequence, the red square above is always there, preventing the field to get any smaller.
Note: as you can see the reserved space is also visible when some content is selected, making the minimal space between the selected value and the right caret always at least what you can see in the "with content" screenshot above.
I don't think this is an issue in practice as it only noticeable when the field is really constrained in its allowed horizontal space. 

# Executed Tests

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x


manuelleduc:
@mflorea do you remember why those right paddings were required initially?

PR #3077, opened 2024-04-19
XWIKI-21492: Underline inline links

Sereza7:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-21492
# Changes

## Description

* Fixed spacing for hidelefthideright
* Added a comment
* Improved the inline link selector by extending the exceptions
* Added a specific rule on the SolrSearch page

## Clarifications

* The selector for SolrSearch is specific, so I added it to this page's SSX instead of including it in the general LESS. 

# Screenshots & Video

Here is a series of links that were updated with these changes. Always the picture with the state of things before this PR comes first and is followed by the looks after this PR.
image
image
___
image
image
___
image
image
___
image
image
____
image
No specific change, should be fixed with https://jira.xwiki.org/browse/XWIKI-15069
____
image
image

# Executed Tests

Manual tests only, because the changes are only on style
# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * No

### DIFF
--- a/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
+++ b/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
@@ -187,9 +187,11 @@ body.content {
         .not-inline-link();
       }
 
-      // Remove the underlining from links with an icon. In the vast majority of cases, these are already spottable.
-      & a:has(img[alt=""]), // With the silk iconTheme
-      & a:has(span.fa) { // With the Font Awesome iconTheme
+      /* Remove the underlining from links with an image or an icon.
+        In the vast majority of cases, these links are already easy to find.
+       */
+      & a:has(img), // With an image or an icon from the Silk iconTheme
+      & a:has(span.fa) { // With an icon from the Font Awesome iconTheme
         .not-inline-link();
       }
     }


xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
```diff
@@ -168,13 +177,21 @@ body.content {
...
       }
+
+      /* Remove the underlining from links with the btn class.
+       Those are already styled differently from their neighboring text. */
+      & a.btn {
+        .not-inline-link();
+      }
+
+      // Remove the underlining from links with an icon. In the vast majority of cases, these are already spottable.
+      & a:has( img[alt=""]), // With the silk iconTheme
```
michitux, line 191: I wonder if this should really filter on the alt tag. I think it makes sense that in general all links that contain images aren't underline as otherwise the image is underlined which looks strange.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
```diff
@@ -168,13 +177,21 @@ body.content {
...
+      & a:has( span.fa) { // With the Font Awesome iconTheme
```
michitux, line 192:
```suggestion
      & a:has(span.fa) { // With the Font Awesome iconTheme
```

Also, at least in Java, we put comments on extra lines, I don't know if such a rule also exists for CSS?

michitux (approved):
Looks good in general, just a few comments.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: Change made in https://github.com/xwiki/xwiki-platform/pull/3077/commits/0764066aace1eadc145603f45d2aa7fe312148c2 (with a change on the line above too).

> Also, at least in Java, we put comments on extra lines, I don't know if such a rule also exists for CSS?

We have [some commenting code style for CSS](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/XhtmlCssCodeStyle/#HCSS), but AFAIK they don't force a position for the comment. They are not even respected in this LESS file. I kept the commenting style already here.

About it being on the exact line, I think it's useful here because each comment is meant to describe exactly the line it's on. They wouldn't make sense at the level of the ruleset, that's why I separated them from the `//Remove the underlining from [...]` comment.

xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less
Sereza7: I added this filter when thinking that we need to select icons. We ruled that icons do not have semantics so they should always have an empty alt. An other possibility would have been to select by `[data-xwiki-lightbox="false"]` but IMO this is worse (we don't want our anchor display to rely on the lightbox feature).

Sereza7: Soooo...

I was testing it out and it happens that any semantically relevant image in content still hits this selector :)
Any of those images can be opened with  lightbox. Recently I added a hidden button under this image, with an icon inside it. This icon makes it so that this selector is hit no matter the state of the attached image...

michitux: My argument is that the same that applies to icons also applies to all images. But maybe we don't all agree on that?

Sereza7: Addressed in https://github.com/xwiki/xwiki-platform/pull/3077/commits/126d0bea1e51be688061d09407d7f0918899d6dd :+1:

PR #3078, opened 2024-04-22
XWIKI-22108: Extra empty paragraph and spacing in the information panel in edit mode

michitux:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-22108

# Changes

## Description

* Make all HTML macros standalone to avoid wrapping in paragraphs
* Replace multiple dd tags by a single dd with a nested unordered list as otherwise we cannot avoid that the XWiki syntax content is wrapped inside a paragraph

## Clarifications

* An alternative is to just move all HTML macros on their own lines to be standalone and accept that there is now some space between the items. The spacing could also be removed with custom CSS but to me this doesn't seem clean.
* I did not change the edit icon to use the icon theme to not introduce additional changes apart from what's really required to fix this issue.

# Screenshots & Video

clipboard

Alternative with spacing between items:

rhmozHykEHmDSHwLYYoBpTwV

Previous state:

UuigjQkXhNuXYeKmxgxPwWSI

# Executed Tests

```
mvn clean install -Pintegration-tests,docker,legacy,standalone,quality -pl :xwiki-platform-panels-ui
```

There do not seem to be any UI tests targeting this element, the [page object](https://github.com/xwiki/xwiki-platform/blob/c7a69d2835f83db1694fbb88dae9a2055dc3309e/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-test/xwiki-platform-panels-test-pageobjects/src/main/java/org/xwiki/panels/test/po/DocumentInformationPanel.java#L31) for this panel doesn't offer access to the included pages, neither do [EditPage](https://github.com/xwiki/xwiki-platform/blob/b646b33583cbb28b134352af4083daefc6588b85/xwiki-platform-core/xwiki-platform-test/xwiki-platform-test-ui/src/main/java/org/xwiki/test/ui/po/editor/EditPage.java#L47) or any of the subclasses for the wiki or WYSIWYG editor.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x

### DIFF
--- a/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
+++ b/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/IncludedPagesDocumentInformation.xml
@@ -209,13 +209,18 @@
 ##--------------------------------------------------------------------------
 #set ($pages = $tdoc.includedPages)
 #if($pages.size() != 0)
-  {{html clean="false"}}&lt;dt&gt;{{/html}}$services.localization.render('panels.documentInformation.includesCount', [$pages.size()]){{html clean="false"}}&lt;/dt&gt;{{/html}}
+  {{html clean="false"}}&lt;dt&gt;$escapetool.xml($services.localization.render('panels.documentInformation.includesCount', [$pages.size()]))&lt;/dt&gt;{{/html}}
+
+  {{html clean="false"}}&lt;dd&gt;{{/html}}
+
   #foreach ($page in $pages)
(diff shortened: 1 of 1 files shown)


Sereza7 (approved):
Looks good to me :+1: 
I assume you checked the DOM and that it's still correct HTML :)

tkrieck:
Looks good. The alternative seems odd with the extra spacing so I would stick with the proposed change.

michitux:
> I assume you checked the DOM and that it's still correct HTML :)

I didn't perform any automatic checks, but I manually checked the HTML and there are no extra `p` tags anywhere in the panel, the `dt`/`dd` tags are there and the list is inside `dd` and `dd ` allows arbitrary flow content so I would be very surprised if this was invalid.

